PR #1139, opened 2019-02-17
[NETBEANS-2120] Placing synthetically generated classes in spi.java.h…

jlahoda:
…ints into an accessible package.


lkishalmi:
This issue has been reported against the recent 11.0 version, I guess it needs to be cherry-picked over when it is got approved and merged.

PR #1140, opened 2019-02-18
[NETBEANS-2121] Updated dependency version on our Java EE support modules.

lkishalmi:
The module we are depending on is org.netbeans.modules.j2ee.persistenceapi has a version 1.40.0 it has the include implementation version checkbox ticked in and its implementation version is 1.
Shall I mark the dependency as 1.40.0.1 instead?
Telling the truth I could not reproduce the CNFE locally, but I just have one project where I can test this an it works there.

The other part of the commit is just Maven -> Gradle renames.

### DIFF
--- a/groovy/gradle.persistence/nbproject/project.xml
+++ b/groovy/gradle.persistence/nbproject/project.xml
@@ -56,7 +56,7 @@
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>0</release-version>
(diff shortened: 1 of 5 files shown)


thuryn:
How can I test the result - should I install the build from here https://builds.apache.org/job/incubator-netbeans-linux/ ?

geertjanw:
Yes, once it has been merged, wait for the next build, and then try it.

PR #1142, opened 2019-02-20
[NETBEANS-2148] Maven dependency Check Out action is failing with "The goal you specified requires a project to execute ..."

herkrath:
Details specified at https://issues.apache.org/jira/browse/NETBEANS-2148


matthiasblaesing:
The reasoning in the issue is sound and the change looks good. Will merge.

PR #1143, opened 2019-02-20
[NETBEANS-1995] Exclude extra cluster when generating LICENSE file

dukescript:
According to [NETBEANS-1995](https://issues.apache.org/jira/browse/NETBEANS-1995) report the `LICENSE` file contains references to files from `extra` cluster. However that one isn't included in the final ZIP. As such there is no need to include those files in the NOTICE file.

Adjusting the build tasks and scripts to exclude `extra/**` files.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CreateLicenseSummary.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CreateLicenseSummary.java
@@ -55,6 +55,7 @@ import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
 import org.apache.tools.ant.types.FileSet;
+import org.apache.tools.ant.types.PatternSet;
 import org.apache.tools.ant.types.Resource;
 import org.apache.tools.ant.types.selectors.SelectorUtils;
 import org.netbeans.nbbuild.JUnitReportWriter;
@@ -144,6 +145,12 @@ public class CreateLicenseSummary extends Task {
     private FileSet moduleFiles;
     public FileSet createModuleFiles() {
         return (moduleFiles = new FileSet());
+    }
+
+    private PatternSet excludeFiles;
+    public void setExcludes(String str) {
+        excludeFiles = new PatternSet();
+        excludeFiles.setExcludes(str);
     }
 
     private Map<String, String> pseudoTests;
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
Thank you for tackling this. A bit more background information would be good though. If I read this correctly, the `nb-extra-files` are explicitly excluded from the zip file, so excluding them from the license file seems sane. This raises the question, why they are placed in the build directory in the first place. If they are not there, we don't need a hack to ignore them.

JaroslavTulach:
The `extra` cluster has traditionally been a place where additional modules (for example from original [contrib](http://hg.netbeans.org/main/contrib) repository) end up.

The purpose of the `extra` cluster is to support local development without influencing packaging of the distributions. One can build any module, and that module is available when one does `ant tryme`. However no such additional module influences content of ZIP produced by `ant build`.

> A bit more background information would be good though

The `extra` cluster is also a place where NBM files downloaded into installation location end up (unless they specify cluster by themselves). I suggested @eppleton to use the `extra` cluster _trashbin_ for #917. Unless I am mistaken the overall idea is [described in this comment](https://github.com/apache/incubator-netbeans/pull/917#issuecomment-439595927).

matthiasblaesing:
Thank you for the pointer and explanation Jaroslav, that helped me to understand. Maybe this could be considered as an enhancement to the change:

```diff
--- a/nbbuild/build.xml
+++ b/nbbuild/build.xml
@@ -79,6 +79,12 @@
             <include name="platform/javahelp/external/binaries-list"/>
         </manifest>
     </downloadbinaries>
+    <!--
+    nb-extra-files denotes directories/files, that are not packed for
+    distribution. The clusters are only used for local development/testing.
+    The files must not be included in distribution zip and also need to be
+    excluded from license analysis
+    -->
     <property name="nb-extra-files" value="test/**,extra/**"/>
 
     <echo message="Bootstrapping NetBeans-specific Ant extensions..."/>
```

Though either way, for a license perspective this looks good to me (I did a build and checked the LICENSE file of the result zip).

However, I don't see that it works. I used the above mentioned zip, extracted it and run it with a clean user+cache dir. I would now expect the "Embedded Browser UI - JavaFX WebView" plugin to be activeable (requirering me to install javafx), but this is not the case. I don't see an option to install JavaFX as a plugin. What am I missing?

JaroslavTulach:
I did some testing with [NETBEANS-2131](https://issues.apache.org/jira/browse/NETBEANS-2131) and [NETBEANS-2132](https://issues.apache.org/jira/browse/NETBEANS-2132) the scenarios described in there work (they were broken in NetBeans 10 on JDK11) to some extent.

I haven't considered the "Embedded Browser UI" usecase so far. It may need few more fixes to get it working, but it should be possible as far as I can tell.

PR #1145, opened 2019-02-21
[NETBEANS-326] fix for local variables not visible in autocomplete in…

sarveshkesharwani:
… lambda expressions

Added Unit Test for this scenario too.


lkishalmi:
Shall this one be part of NetBeans 11.0?

sarveshkesharwani:
This issue has been around since NetBeans 8.2, so I wouldn't say that it is necessary for this to be a part of NetBeans 11.0.

lkishalmi:
Let's have it on the master then...

PR #1146, opened 2019-02-25
[NETBEANS-2047] Maven/Java project template should be preferred

geertjanw:
Highlight the fact that the Maven/Java Application project template should be the first to use.


geertjanw:
Merging this simple HTML change.

PR #1147, opened 2019-02-26
[NETBEANS-2125] Enable Oracle JS Parser to be installed with Java EE

geertjanw:
Tried @JaroslavTulach instructions in https://issues.apache.org/jira/browse/NETBEANS-2125, and it works, so here it is.


geertjanw:
OK, merging.

PR #1141, opened 2019-02-19
[NETBEANS-2130] Add support to download/install/register GlassFish 5.1

geertjanw:
Did everything, except pointing to the Eclipse Licence, hope @peedeeboy or someone will take a look at this. Simply followed @peedeeboy's excellent steps in https://issues.apache.org/jira/browse/NETBEANS-2130, which works great.

PR #1148, opened 2019-02-27
[NETBEANS-2182] Display correct licence (Oracle / Eclipse Foundation) depending on version of GlassFish to download

peedeeboy:
What this pull request does in a nutshell
---------------------------------------------
Expands on #1141 by:

- Removing the JOptionDialog to select a version of GlassFish to download, and moving that functionality to the JPanel in the Wizard
- Amending the ServerDetails enums to include a URL to the appropriate license (Oracle / EF) for each GlassFish version supported
- Button to open license now checks selected version of GlassFish and opens browser pointing to correct license from ServerDetails
- Fixes a tiny typo in #1141 (typo was probably actually copied from my instructions in Jira!)

API Changes
--------------
org.netbeans.modules.glassfish.common.ServerDetails now has a new public method, getLicenseUrl() that returns the URL for a given version of GlassFish

I have incremented the API version number in the manifest file from 1.76 to 1.77.  Is there anything else I need to do?

Alternative to API change
----------------------------
If the API change is a bad thing, an alternative would be to add a couple of constants to the Glassfish Wizard, and if ServerDetails.getVersion() <= 500 then Oracle license, else Eclipse license.  That would feel hacky though, and one more random thing to miss when added support for a new version of GF.

Testing
--------
- Launch NB, Services tab -> Servers -> Right click -> Add Server
- Select GlassFish server -> Next
- After PR: JComboBox is now on this panel.  Select a version of GF and click on the license link - correct license for the version of GlassFish selected should be displayed
- Changing version of GlassFish selected, will uncheck the license checkbox

Dependencies
----------------
This is built on top of b141842 from #1141

### DIFF
--- a/enterprise/glassfish.common/manifest.mf
+++ b/enterprise/glassfish.common/manifest.mf
@@ -4,6 +4,6 @@ OpenIDE-Module: org.netbeans.modules.glassfish.common/0
 OpenIDE-Module-Install: org/netbeans/modules/glassfish/common/Installer.class
 OpenIDE-Module-Layer: org/netbeans/modules/glassfish/common/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/glassfish/common/Bundle.properties
-OpenIDE-Module-Specification-Version: 1.76
+OpenIDE-Module-Specification-Version: 1.77
 OpenIDE-Module-Provides: org.netbeans.modules.glassfish.common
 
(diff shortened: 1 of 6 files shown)


geertjanw:
Awesome!

enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/ServerDetails.java
```diff
@@ -319,4 +346,7 @@ public String getIndirectUrl() {
         return indirectUrl;
     }
 
+    public String getLicenseUrl() {
```
JaroslavTulach: Add `@since 1.77`

enterprise/glassfish.common/manifest.mf
```diff
@@ -4,6 +4,6 @@ OpenIDE-Module: org.netbeans.modules.glassfish.common/0
 OpenIDE-Module-Install: org/netbeans/modules/glassfish/common/Installer.class
 OpenIDE-Module-Layer: org/netbeans/modules/glassfish/common/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/glassfish/common/Bundle.properties
-OpenIDE-Module-Specification-Version: 1.76
+OpenIDE-Module-Specification-Version: 1.77
```
JaroslavTulach: Usually one also modifies `apichanges.xml`, but as there are none in the module, we can probably skip that step.

PR #1151, opened 2019-03-03
Fix ExplorerManager#setSelectedNodes()

sjaenick:
Use correct index variable when copying validated nodes; this was originally
reported by me in 2015, original issue is at

https://netbeans.org/bugzilla/show_bug.cgi?id=257181


geertjanw:
Great, makes sense.

matthiasblaesing:
I had to read the code multiple times, to see the logic, but I agree the change looks sane. Merging.

lkishalmi:
https://issues.apache.org/jira/browse/NETBEANS-2223

PR #1152, opened 2019-03-03
[NETBEANS-2201] GlassFish log view manager bundle missing

jGauravGupta:


matthiasblaesing:
The change looks good. You might want to "fix" the entry for `CONFIG` for consistency (see the leading space).

geertjanw:
Great!

PR #1154, opened 2019-03-05
Java Frontend Application wizard for Gradle

dukescript:
Introducing new module that provides a _Java Frontend Application_ wizard for Gradle project. The goal is to get the support for [Apache HTML/Java API](https://github.com/apache/incubator-netbeans-html4j) and the rest of the [DukeScript technologies](https://github.com/dukescript/dukescript-presenters/blob/master/README.md) into the hands of our Gradle users. They should be able to choose the project type and get a project with a sample application that can be executed, debugged and tested out of the box.

@lkishalmi, what is the right way to populate a project with predefined structure? There is nothing like Maven archetypes in the Gradle world, right? As such the IDE has to configure the project skeleton by own means. The first attempt extracts the sample from a ZIP file. That is probably not ideal. Are there some better ideas?

Thanks for the review. We'd like to take some time and polish this PR and integrate for NetBeans 12.


dukescript:
Thanks @lkishalmi for the review. Rather than copying the files from a ZIP file, let's copy them from the `layer.xml`! The 03c5e39 commit contains an example how that could be done (iOS files are still in the ZIP as there is too many of them). More work will be needed on processing the template files and placing them in the right packages...

junichi11:
@dukescript Missing license headers:
```
https://github.com/apache/incubator-netbeans/tree/master/groovy/gradle.htmlui/build.xml
https://github.com/apache/incubator-netbeans/tree/master/groovy/gradle.htmlui/nbproject/project.properties
https://github.com/apache/incubator-netbeans/tree/master/groovy/gradle.htmlui/nbproject/project.xml
https://github.com/apache/incubator-netbeans/tree/master/groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/Bundle.properties
```

dukescript:
Re. Review by @lkishalmi:

> The Gradle files are need some love as well. It would be better to review them as source.

Done. Most of the files is available as freemarker template.

> As of actions, the reload rules can be just DEFAULT (removed from the properties) 

Handled in 657726b

> Do not add wrapper related files to the zip as those can be created by 
> Gradle as a last step of installation.

There are no `gradlew` files, as far as I can tell, but please take another look now.

groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaApplicationProjectWizard.java
```diff
@@ -0,0 +1,114 @@
...
+import org.netbeans.api.templates.TemplateRegistration;
+import org.netbeans.modules.gradle.api.NbGradleProject;
+import org.openide.WizardDescriptor;
+import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
+import org.openide.util.Exceptions;
+import org.openide.util.NbBundle.Messages;
+
+@TemplateRegistration(folder="Project/Gradle", position=999, displayName="#template.htmlUIProject", iconBase="org/netbeans/modules/gradle/htmlui/DukeHTML.png", description="description.html")
+@Messages("template.htmlUIProject=Java Frontend Application")
```
lkishalmi: This could be quite confusing especially when this is in the main Gradle Project Wizard folder. At least if I'd like to do a Java Frontend Application, I would not think on bck2brwsr app on the first place (I'm sorry for that).

groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaApplicationProjectWizard.java
```diff
@@ -0,0 +1,114 @@
...
+public class HtmlJavaApplicationProjectWizard extends SimpleGradleWizardIterator {
+    private static final String DEFAULT_LICENSE_TEMPLATE = "/Templates/Licenses/license-default.txt"; //NOI18N
+    
+    @Messages("LBL_FrontendApplicationProject=Java Frontend Application with Gradle")
+    public HtmlJavaApplicationProjectWizard() {
+        super(Bundle.LBL_FrontendApplicationProject(), initParams());
+    }
+
+    private static Map<String, Object> initParams() {
```
lkishalmi: I guess this method is a copy paste and does not really needed.

groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaApplicationProjectWizard.java
dukescript: This template is not about bck2brwsr. The main focus is [DukeScript](http://dukescript.com) - e.g. combination of the best Java VM and portable frontend rendering. What exactly is the problem? The "Run" action? If so, it can be changed to do the same thing as "Debug".

The naming is consistent with the Maven category. There is a *Java Frontend Application* archetype in the Maven category which leads to the same/similar application, using the same libraries. Our intention is to deliver the same *Java Frontend Application* experience to Gradle users.

dukescript: Btw. there is also a way to run the application as [Android application](http://kt-mvvm.org/android.html) and as [iOS application](http://kt-mvvm.org/ios.html). They aren't advertised in the UI yet as it requires additional configuration of SDK, simulator, etc. We'd welcome coordinated support in this direction of generic *Java Frontend Applications*, but we think it is out of the scope of this PR.

groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaApplicationProjectWizard.java
dukescript: The method is used on line 47, but the properties are likely useless: ee258ebd978a

dukescript:
Are there any outstanding issues or can we move on and proceed with integration, please? @lkishalmi, @geertjanw - this PR is open to maintainers - feel free to do any renames that you wish. There is no need to block integration just because of the naming.

dukescript:
The [checks fails](https://api.travis-ci.org/v3/job/520098630/log.txt) when running `org.netbeans.modules.java.completion.JavaCompletionTask110FeaturesTest`. That is unlikely related to the gradle wizard changes, as `groovy` cluster isn't even present in such mode. Do you, @jlahoda, know something about that test? Can you address the failure and merge this PR when appropriate?

PR #1156, opened 2019-03-06
[NETBEANS-1936] Removing the SVN client version suffix before parsing the number

jan-tosovsky-cz:
If dev version of the SVN client is used together with Netbeans, many operations over SVN versioned projects fail because of NumberFormatException.


geertjanw:
I think this makes sense and is needed when dev versions of Subversion are used.

PR #1158, opened 2019-03-08
[NETBEANS-2220] Preventing NPE when there is no item at the specified…

jlahoda:
… cell in the table.


lkishalmi:
Harmless trivial NPE check. Merging...

PR #1159, opened 2019-03-11
[NETBEANS-2132] Sign wrapping NBMs bundled in the ZIP distribution

dukescript:
Polishing the #917 experience by signing the NBM wrappers that provide JavaFX libraries on JDK11. The NBMs are distributed as part of overall ZIP file, which is trusted - e.g. trusting the NBMs is the right thing to do.

The `nb/updatecenters` module already contained (an empty) `ide.ks` keystore. This change generates temporary signing key, signs the three NBMs files with it and imports the signing certificate into the newly created `ide.ks` store.

The temporary signing keystore is then destroyed, so nobody can use the private key to sign anything else.

CCing @lkishalmi, @matthiasblaesing, @jtulach, @jlahoda


jlahoda:
I think the approach seems good to me. Travis does not seem to like the patch, though:
---
 [nbmerge] Failed to build target: all-updatecenters

BUILD FAILED
/home/travis/build/apache/incubator-netbeans/nbbuild/build.xml:514: The following error occurred while executing this line:
/home/travis/build/apache/incubator-netbeans/nbbuild/build.xml:509: The following error occurred while executing this line:
/home/travis/build/apache/incubator-netbeans/nbbuild/build.xml:556: The following error occurred while executing this line:
/home/travis/build/apache/incubator-netbeans/nb/updatecenters/build.xml:51: exec returned: 1
---

Moreover, would it be possible to generate the keystore into the build directory (or into the cluster directly), so that the checkout is not polluted with additional files (even if they are .gitignored?).

dukescript:
The temporary keystore with the signing certificate is now generated into `build/ide.ks` and included at proper location via `release.*` property in the `nbproject/project.properties` file.

JaroslavTulach:
Yeah, the user experience is certainly better than it was before. I agree with the fix.

PR #1160, opened 2019-03-11
[NETBEANS-2160] Start on jdk11.0.2

JaroslavTulach:
Updates Felix to version `6.0.2` that supports the `jrt://` protocol.


jlahoda:
+1

PR #1162, opened 2019-03-13
[NETBEANS-2246] Strings for update centers

geertjanw:
See https://issues.apache.org/jira/browse/NETBEANS-2246.


lkishalmi:
Also should this change affect master as well or just specific to the release branch?

PR #1164, opened 2019-03-16
Place the internationalisation bundles from the 3rd donation in the A…

borisheithecker:
…pache NetBeans GitHub repository.

https://issues.apache.org/jira/browse/NETBEANS-2263


junichi11:
Could you please fix the commit message?
see: https://netbeans.apache.org/participate/submit-pr.html

junichi11:
Maybe, `*.mdd` files are not AL. see: https://github.com/apache/incubator-netbeans/blob/404eb9481beba31856b66b50fce440f18ae4a8a3/enterprise/j2ee.dd/licenseinfo.xml 
@matthiasblaesing Could you please take a look at this?

matthiasblaesing:
@junichi11 you are right, the license change is not correct and must not be committed (the files were not part of an oracle donation, but integrated from a 3rd party source).
@borisheithecker it needs to be established, that the files you want to be merged are all from a donation. Then the license change needs to be applied (again only on the donated files). This is easiest done by splitting it into two commits: commit 1) import the files commit 2) update the license headers. If you then provide some info how the contents can be compared with the donated files it would be great.

borisheithecker:
Okay, I made a mistake converting the license headers. Sorry for that. 
Maybe provide a new pull request? 
If this is to be done as @matthiasblaesing comments, the PR will have at least 2 commits. 
Also, the following changes are required: 
1. ".hgignore" in the donated zip needs to be rewritten immediately to a corresponding ".gitignore". 
2. Fix netbeans-l10n/build.xml to work with jdk8
Let it be 4 commits?

borisheithecker:
New PR
https://github.com/apache/incubator-netbeans/pull/1167

matthiasblaesing:
Thank you for providing an updated PR, I'll have a look at in the next few days. As a hint for the future: There is no need to create new PRs in these cases, at least github and gitlab support updating PRs in place by force-pushing to the tracked branch. Feel free to either close this PR or update this PR with your modified approach.

PR #1171, opened 2019-03-20
Integration of JEP-330

sarveshkesharwani:
A feature of JDK-11, here is the link for [JEP-330](https://openjdk.java.net/jeps/330).


java/java.source/src/org/netbeans/modules/java/JavaNode.java
```diff
@@ -213,6 +216,48 @@ protected final Sheet createSheet () {
...
+        // If any of the parent folders is a project, this menu won't appear.
+        if (parentProject == null) {
+            Node.Property arguments = new org.openide.nodes.PropertySupport.ReadWrite<String> (
+                    "runFileArguments", // NOI18N
+                    String.class,
+                    "Arguments",
+                    "Arguments passed to the main method while running the file."
+                ) {
+                    public String getValue () {
+                        return NbPreferences.forModule(JavaNode.class).get(fileName + "_SINGLE_FILE_RUN_ARGUMENTS", "");
```
JaroslavTulach: `dataObject.getPrimaryFile().setAttribute(...)` or `getAttribute()` would be my preferred location to store these metadata. These two methods were invented exactly for the purpose of associating additional metadata with files.

java/java.source/src/org/netbeans/modules/java/JavaNode.java
```diff
@@ -213,6 +216,48 @@ protected final Sheet createSheet () {
         });
         sheet.put(ps);
         
+        Project parentProject = FileOwnerQuery.getOwner(super.getDataObject().getPrimaryFile());
+        String fileName = super.getDataObject().getPrimaryFile().getName();
+        // If any of the parent folders is a project, this menu won't appear.
```
JaroslavTulach: "this menu" is misleading I think.

java/java.source/nbproject/project.xml
```diff
@@ -222,6 +222,15 @@
                         <specification-version>1.58</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.extexecution</code-name-base>
```
JaroslavTulach: Such dependency makes the `java.source` module more [heavyweight](http://wiki.apidesign.org/wiki/HeavyWeight). Is that OK, @jlahoda?

java/java.source/src/org/netbeans/api/java/source/support/SingleJavaSourceRunActionProvider.java
```diff
@@ -0,0 +1,126 @@
...
+public class SingleJavaSourceRunActionProvider implements ActionProvider {
+
+    @Override
+    public String[] getSupportedActions() {
+        return new String[]{ActionProvider.COMMAND_RUN_SINGLE};
+    }
+
+    @Override
+    public void invokeAction(String command, Lookup context) throws IllegalArgumentException {
+        FileObject fileObject = getJavaFileWithoutProjectFromLookup(context);
```
JaroslavTulach: `fileObject` can be `null`.

java/java.source/src/org/netbeans/api/java/source/support/SingleJavaSourceRunActionProvider.java
```diff
@@ -0,0 +1,126 @@
...
+        String arguments = NbPreferences.forModule(SingleJavaSourceRunActionProvider.class).get(fileObject.getName() + "_SINGLE_FILE_RUN_ARGUMENTS", "");
+        String vmOptions = NbPreferences.forModule(SingleJavaSourceRunActionProvider.class).get(fileObject.getName() + "_SINGLE_FILE_RUN_VM_OPTIONS", "");
+        ExecutionDescriptor descriptor = new ExecutionDescriptor().controllable(true).frontWindow(true).
+                preExecution(null).postExecution(null);
+        List<String> commandsList = new ArrayList<>();
+        if (Utilities.isUnix()) {
+            commandsList.add("bash");
+            commandsList.add("-c");
+        }
+        String javaPath = "\"" + System.getProperty("java.home") + File.separator + "bin" + File.separator + "java\"";
```
JaroslavTulach: I'd suggest: `new File(new File(new File(javaHome), "bin", "java")`

java/java.source/src/org/netbeans/api/java/source/support/SingleJavaSourceRunActionProvider.java
```diff
@@ -0,0 +1,126 @@
...
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ * This class provides support to run a single Java file without a parent
+ * project (JEP-330).
+ *
+ * @author Sarvesh Kesharwani
+ */
+@ServiceProvider(service = ActionProvider.class)
+public class SingleJavaSourceRunActionProvider implements ActionProvider {
```
JaroslavTulach: I am surprised this is an `ActionProvider` - that is for executing projects. I'd rather create  the action via `Actions.context` or annotations and linked to it in `Loaders/Actions/text/java` folder in the layer via `@ActionReference`.

sarveshkesharwani: Thanks for the review @JaroslavTulach . My first instinct was to add this via @ActionReference too, but here, I am making use of the "Run File" which is already added to Project UI's layer. (which is also present for a single file). So, now, if I create a new action with the annotation, wouldn't it mean that we would have two run menus (one for the new action, and the other disabled "Run File")?

java/java.source/nbproject/project.xml
sarveshkesharwani: The reason why I was doing this in the first place, is that I was using NbPreferences to store arguments, and VM Options. And NbPreferences require for the accessing class to be in the same module, but now, I am using dataobject's attribute, so I can move this class out of java.source.

java/java.source/src/org/netbeans/modules/java/JavaNode.java
sarveshkesharwani: Great tip! I wasn't aware of this.

java/java.api.common/src/org/netbeans/modules/java/api/common/SingleJavaSourceRunActionProvider.java
```diff
@@ -0,0 +1,132 @@
...
+    public void invokeAction(String command, Lookup context) throws IllegalArgumentException {
+        FileObject fileObject = getJavaFileWithoutProjectFromLookup(context);
+        if (fileObject == null) 
+            return;
+        String filePath = fileObject.getPath();
+        Object argumentsObject = fileObject.getAttribute(FILE_ARGUMENTS);
+        String arguments = argumentsObject != null ? (String) argumentsObject : "";
+        Object vmOptionsObject = fileObject.getAttribute(FILE_VM_OPTIONS);
+        String vmOptions = vmOptionsObject != null ? (String) vmOptionsObject : "";
+        ExecutionDescriptor descriptor = new ExecutionDescriptor().controllable(true).frontWindow(true).
```
JaroslavTulach: I would extract this code into a separate method and then call this method in the `testSingleJavaSourceRun`...

java/java.source/test/unit/src/org/netbeans/modules/java/JavaNodeTest.java
```diff
@@ -40,4 +48,30 @@ public void testSheetToArrayReturnsModifiableArray() {
...
+        try {
+            List<String> commandsList = new ArrayList<>();
+            if (Utilities.isUnix()) {
+                commandsList.add("bash");
+                commandsList.add("-c");
+            }
+            File f1 = new File(new File(new File (getDataDir().getAbsolutePath()), "files"), "TestSingleJavaFile.java");
+            File javaPathFile = new File(new File(new File(System.getProperty("java.home")), "bin"), "java");
+            commandsList.add(javaPathFile.getAbsolutePath() + " " + f1.getAbsolutePath());
+            ProcessBuilder pb = new ProcessBuilder(commandsList);
```
JaroslavTulach: ...rather than testing `ProcessBuilder` I suggest to test the code extracted from the `invokeAction`.

java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/TestJavaFile.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    public void testSingleJavaSourceRun() {
+        try {
+            
+            File f1 = new File(new File(new File (getDataDir().getAbsolutePath()), "files"), "TestSingleJavaFile.java");
+            FileObject javaFO = FileUtil.toFileObject(f1);
+            SingleJavaSourceRunActionProvider runActionProvider = new SingleJavaSourceRunActionProvider();
+            if (!runActionProvider.isActionEnabled("run.single", Lookup.EMPTY)) {
+                return;
```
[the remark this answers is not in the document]
sarveshkesharwani: Well, this test basically checks if the IDE is running on JDK 11 or above, and if it isn't, then this option is not given to the user. So, in such a case, there is no point running the test too, right?

java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/TestJavaFile.java
```diff
@@ -0,0 +1,73 @@
...
+import org.openide.filesystems.FileUtil;
+import org.openide.util.Exceptions;
+import org.netbeans.modules.java.api.common.util.RunProcess;
+import org.openide.util.Lookup;
+
+/**
+ *
+ * @author Sarvesh Kesharwani
+ */
+public class TestJavaFile extends NbTestCase {
```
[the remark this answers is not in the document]
sarveshkesharwani: Fixed this in the latest commit.

java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/TestJavaFile.java
```diff
@@ -0,0 +1,73 @@
...
+            BufferedReader reader =
+                    new BufferedReader(new InputStreamReader(process.getInputStream()));
+            StringBuilder builder = new StringBuilder();
+            String line = null;
+            while ( (line = reader.readLine()) != null) {
+                builder.append(line);
+            }
+            String result = builder.toString();
+            assertEquals("hello world", result);
+        } catch (IOException ex) {
```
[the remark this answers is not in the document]
sarveshkesharwani: Fixed this in the latest commit.

java/java.api.common/src/org/netbeans/modules/java/api/common/util/RunProcess.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.concurrent.Callable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.openide.util.Exceptions;
+
+/**
+ *
+ * @author Sarvesh Kesharwani
+ */
+public class RunProcess implements Callable<Process> {
```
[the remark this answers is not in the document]
sarveshkesharwani: Right. I have moved this file to a different package, so there is no longer a need for it to be public.

java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/TestJavaFile.java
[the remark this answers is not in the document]
JaroslavTulach: The following code "checks if the IDE is running on JDK 11 or above" and if not it asserts that the action is disabled and skips further testing.
```java
if (!isJDK11OrNewer()) {
  assertFalse("The action is only enabled on JDK11 and newer", runActionProvider.isActionEnabled(....));
  return;
}
```
I know the `run.single` action is disabled on JDKs older than 11, but that is actually something to test. Just my 2 Kč.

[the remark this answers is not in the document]
sarveshkesharwani: Got it. It does make sense to have test here to see whether the action is enabled or not based on the current version of JDK. I have added it in the latest commit.

junichi11:
Need squashing?

sarveshkesharwani:
> Need squashing?

Yes. We can squash while merging though, right?

junichi11:
Yes, it's just confirmation for a person who merges this :)

zimmi:
@sarveshkesharwani Thank you for doing this, really cool! :)

geertjanw:
It is a great honor for me to squash and merge this amazing enhancement for NetBeans. Thanks!

geertjanw:
Squashing and merging,

apws:
Hi all, I am now heavily experimenting with JEP-330 now as java beginner (coming from c#), using it for simple and also not so simple experiments with javafx too using "launch(AppClass.class, args)" trick and heavy usage of "multi-class" experiments, where also MORE public classes (reflected beans for FX TableView for example) are present (it seems supported and expected by JEP-330) but what is failing for me is "Debug file" feature in case there are MORE PUBLIC classes in single file - I know that Debug compiles into class-files and there may be some troubles with standard compilation, but helping to "tolerate" many public classes in single file would be GREAT. Netbeans is currently alone IDE which allows such JEP-330 debugging without ANY project settings - I just have mapped *.java (and my win-associated *.jx/jxc extensions too) to launch Netbeans, where I use "Run File" (in memory, OKAY with mutli-class and publics) and "Debug File" (complaining about MORE PUBLIC classes during compilation ...)

junichi11:
@apws Please join the mailing list: https://netbeans.apache.org/community/mailing-lists.html Then, please ask about it. Thanks.

PR #1173, opened 2019-03-23
maven project compile args not passed to editor javac

rtaneja1:


java/maven/src/org/netbeans/modules/maven/queries/PomCompilerOptionsQueryImpl.java
```diff
@@ -105,8 +105,11 @@ public Result getOptions(@NonNull final FileObject file) {
             String[] compilerArgs = PluginPropertyUtils.getPluginPropertyList(proj, Constants.GROUP_APACHE_PLUGINS, Constants.PLUGIN_COMPILER, "compilerArgs", "arg", null); // NOI18N
             if (compilerArgs != null) {
                 for (String compilerArg : compilerArgs) {
-                    if(compilerArg != null && ARG_PARAMETERS.equals(compilerArg.trim())) {
-                        return Arrays.asList(ARG_PARAMETERS);
+                    if (compilerArg != null)  {
+                        if (ARG_PARAMETERS.equals(compilerArg.trim())) {
+                            return Arrays.asList(ARG_PARAMETERS);//reqd?
```
matthiasblaesing: In the original code there are only two cases:
- if a compiler argument `-parameters` was provided, it will be returned
- else an empty list is returned

This makes sense reading the commit message of the addition:
http://hg.netbeans.org/main-golden/log/tip/maven/src/org/netbeans/modules/maven/queries/PomCompilerOptionsQueryImpl.java 
It is basicly a guard to limit the change to that single parameter.

rtaneja1: Thanks, removed special handling for -parameters here.

rtaneja1:
https://github.com/apache/incubator-netbeans/blob/master/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java

validateCompilerOptions removes any unsupported parameters.

PR #1174, opened 2019-03-24
[NETBEANS-2303] Update version number on Windows Launchers

lkishalmi:
Well in the unlikely event of releasing a 11.0-vc5 it could be useful to have a Windows Launcher version update as well.

### DIFF
--- a/harness/apisupport.harness/windows-launcher-src/app.exe.manifest
+++ b/harness/apisupport.harness/windows-launcher-src/app.exe.manifest
@@ -20,7 +20,7 @@
 
 -->
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
-<assemblyIdentity version="9.0.0.0"
+<assemblyIdentity version="11.0.0.0"
    processorArchitecture="x86"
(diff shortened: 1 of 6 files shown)


eirikbakke:
Looks good. There's also harness/apisupport.harness/windows-launcher-src/app.exe.manifest , which should ideally be updated for consistency as well.

More generally, is there a checklist somewhere for steps that must be taken whenever a new release is to be cut out? We should include the bumping of these version constants there.

geertjanw:
Here would be a good place:
https://cwiki.apache.org/confluence/display/NETBEANS/Apache+NetBeans+Release+README

lkishalmi:
The actual list is here: https://issues.apache.org/jira/browse/NETBEANS-2052
This item is new on the list as this come up the first time. Or maybe as we were working with pre-built launcher artifacts at that time.
Anyhow it is there now. The next RM just need to clone the issue and will have the todo-s for the next release.

eirikbakke:
Great! Could you update harness/apisupport.harness/windows-launcher-src/app.exe.manifest as well? Then it looks ready to merge.

PR #1176, opened 2019-03-28
Add the "main" abbreviation as the alias of "psvm"

junichi11:


junichi11:
@svatoun Thank you for reviewing it! Yes. I also think so. `set` and `map` are as well? They are already included by default. If some users don't expect it, they can remove it easily as a workaround.

sdedic:
Good examples (map, set). Actually I do not remember these abbreviations got in my way, so the same is hopefully true for `main` as well. The only possble clash is on class level, but I can't think of "regular" case when main+tab is used; even the user declares `main` field, the next typed character is most likely `;` space or `=`.

junichi11:
Merging.

PR #1177, opened 2019-03-28
[NETBEANS-2318] Remove synchronized elements from SQLEditorSupport

matthiasblaesing:


Chris2011:
I had a look into the code, but it means, that the dialog for a SQL connection is still there, right? This was not removed. The dialog comes up, when you hit ctrl + space inside a SQL file.

matthiasblaesing:
Sorry - I missed the description in this PR. The intention of this PR is to change nothing on the user facing side. All processes in the GUI and below stay the same. The issue is: When closing an SQL editor, there is a change, that there is a deadlock. This blocks the UI indefinitely and thus you have to kill netbeans.

The root of the deadlock lies in the fact, that a UI thread and a RequestProcessor both try to access synchronized functions in SQLEditorSupport. This PR removes the synchronized primitives and 
a) either replaces them with volatile variables (the changes around dbcon and executing) or
b) introduce a more finegrained lock, that has less potential for lock conflicts (logger).

I think you are thinking about a different issue (code completion without an sql connection).

Chris2011:
Thx, for the explanation. Now it is a bit more clear. Yes i thought it could be this issue but as i had a look into the code, i could answered it by myself.

matthiasblaesing:
Will merge by wednesday if noone objects till then.

PR #1178, opened 2019-03-29
Release100

PhilippBaumi:
Give us an option to create every time a project with different JDK versions in the "File->New project" menu in the
grafik

screen with the name "JDK version" and then you can select the JDK which you want to use from the installed JDKs.

That would be really helpful in creating a new project.

Thank you.


junichi11:
@PhilippBaumi "Pull requests" is not an issue tracker. If you have issues, please report them to the [JIRA](https://issues.apache.org/jira/projects/NETBEANS/issues).

PR #1179, opened 2019-03-30
NETBEANS-1820 Deleting images with licensing conerns

geertjanw:


matthiasblaesing:
Please remove the whole easeldemo project and the references to it and the images:
- in the licenseinfo.xml the file references need to removed
- in the build.xml the task "zip-examples" needs to be removed
- the jar target does not need to be overriden anymore and thus can be removed to

geertjanw:
Now we're talking, great, thanks a lot, done AFAIK those specific things too.

matthiasblaesing:
The update looks good so far, but needs on update:
_/webcommon/web.client.samples/build.xml_: Remove lines 25-32 (that are the two targets `jar` and `zip-samples`. Both are not needed, after the removal of the Easel sample
_/webcommon/web.client.samples/license.xml_: Remove all "<file>...</file>" entries, that start with "EaselDemo". 
That should fix the build.

geertjanw:
Sorry I thought I had done that, turns out I hadn't committed and pushed these two changed files.

webcommon/web.client.samples/build.xml
```diff
@@ -21,13 +21,4 @@
 -->
 <project basedir="." default="build" name="webcommon/web.client.samples">
     <description>Builds, tests, and runs the project org.netbeans.modules.web.client.samples</description>
-    <import file="../../nbbuild/templates/projectized.xml"/>
```
matthiasblaesing: line 24 needs to stay (this is the inclusion of the normal netbeans build harness)

matthiasblaesing:
The PR is failing in the ergonomic test, because the now removed zip is still referenced - in the layer the entry for "ResponsiveRabbits" needs to be removed and the corresponding entry in the bundle can be removed true.

This is the resulting diff:

```diff
diff --git a/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/layer.xml b/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/layer.xml
index 280c2c7..065d47c 100644
--- a/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/layer.xml
+++ b/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/layer.xml
@@ -27,14 +27,6 @@
                 <folder name="HTML5">
                     <attr name="SystemFileSystem.localizingBundle" stringvalue="org.netbeans.modules.web.client.samples.wizard.ui.Bundle"/>
                     <attr name="position" intvalue="210"/>
-                    <file name="ResponsiveRabbits" url="Easel.zip">
-                        <attr name="displayName" bundlevalue="org.netbeans.modules.web.client.samples.wizard.ui.Bundle#Templates/Project/Samples/HTML5/ResponsiveRabbits"/>
-                        <attr name="position" intvalue="100"/>
-                        <attr name="iconBase" stringvalue="org/netbeans/modules/web/client/samples/resources/HTML5_project_icon.png"/>
-                        <attr name="instantiatingIterator" newvalue="org.netbeans.modules.web.client.samples.wizard.iterator.SampleWizardIterator"/>
-                        <attr name="instantiatingWizardURL" urlvalue="nbresloc:/org/netbeans/modules/web/client/samples/resources/ResponsiveRabbits.html"/>
-                        <attr name="template" boolvalue="true"/>
-                    </file>
                  </folder>
             </folder>
         </folder>
diff --git a/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/wizard/ui/Bundle.properties b/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/wizard/ui/Bundle.properties
index be2f343..07fdb3c 100644
--- a/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/wizard/ui/Bundle.properties
+++ b/webcommon/web.client.samples/src/org/netbeans/modules/web/client/samples/wizard/ui/Bundle.properties
@@ -15,7 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 Templates/Project/Samples/HTML5=HTML5/JavaScript
-Templates/Project/Samples/HTML5/ResponsiveRabbits=Responsive Rabbits
 
 LBL_ProjectLocation=Project &Location:
 LBL_ProjectName=Project &Name:

```

Without this change, the PR still fails locally in the ergonomics test, with it, it runs cleanly.

geertjanw:
No worries, will do. (Note we have 4 +1s for 11.0 and this one will be in the next release. I.e., this is not a blocker for 11.0.)

arvindaprameya:
If this is not a blocker for 11, is the "Need cherry picking " label still needed ?

geertjanw:
True.

geertjanw:
But we were thinking about doing another voting candidate, i.e., vc5 , in which this one would be cherry picked, but vc5 will not be needed.

lkishalmi:
Just a note from Frankfurt before leaving Europe behind.
I think if we would have NB 11.1 in the summer, that would be branced/released from this branch, so cherry picking would be required.

matthiasblaesing:
Looks good to me. Please squash before merge though. Thank you for taking care.

PR #1181, opened 2019-03-31
[NETBEANS-1691] Add support for @supports at-rule (fix error in CSS)

peedeeboy:
**What this Pull-Request does**
Adds CSS support for the supports at-rule

**Files Changed**
*CSS Editor Module*
ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java - supports add rule will be recognised in code completion

*CSS Library Module*
ide/css.lib/manifest.mf - updated API version to 1.88
ide/css.lib/src/org/netbeans/modules/css/lib/Css3.g - added SUPPORTS_SYM as a lexer token, along with several new grammar rules for supports at-rules
ide/css.lib/src/org/netbeans/modules/css/lib/Css3Lexer.java - Regenerated by ANTLR
ide/css.lib/src/org/netbeans/modules/css/lib/Css3Parser.java - Regenerated by ANTLR
ide/css.lib/src/org/netbeans/modules/css/lib/api/CssTokenId.java - Added SUPPORTS_SYM as a lexer token
ide/css.lib/src/org/netbeans/modules/css/lib/api/NodeType.java - Add the new grammar rules as nodes
ide/css.lib/test/unit/src/org/netbeans/modules/css/lib/Css3LexerTest.java - Add unit tests for SUPPORTS_SYM lexer token
ide/css.lib/test/unit/src/org/netbeans/modules/css/lib/Css3ParserTest.java - Added several unit tests for the support rule syntax (mixture of AND and OR etc.)
ide/css.lib/test/unit/src/org/netbeans/modules/css/lib/api/properties/GrammarResolverListenerTest.java - Fixed an issue caused by a another PR due to tokens changing number
ide/css.lib/test/unit/src/org/netbeans/modules/css/lib/api/properties/PropertyValueTest.java - Fixed several tests that were failing due to another PR

*CSS Source Model Module*
ide/css.model/src/org/netbeans/modules/css/model/impl/ElementFactoryImpl.java - Returns an implementation of PlainElement where encountering a supports at-rule, this stops the "WARNING: created element by reflection for SupportsAtRule, update the ElementFactoryImpl.createElement() methods ugly switch!" being written to the NB log

**How to Test**
Download/checkout this PR
Rebuild NB with Ant
Create a new CSS file, copy and paste the examples from the W3C specification / MDN etc.
OR open the Bootstrap CSS file as this contains supports at-rules
Check to parsing errors on the supports rules

**Things I'm not 100% happy with**
1) I don't fully understand the purpose of the CSS Source Model Library.  To suppress the "WARNING: created element by reflection for SupportsAtRule, update the ElementFactoryImpl.createElement() methods ugly switch!" error I am returning a PlainElement implementation for supports at-rules.

I guess in an ideal world it should be possible to build an OO supports at-rule using this library, but I can't find any documentation on it, and was tying myself in knots trying.

If somebody that understands this library could point me in the right direction / have a crack at it, that would be super!

2) I am using the MediaBody grammar rule for the body of the Support at-rule too.  This is because they are the same rule.  Given the option of create a completely new grammar rule that was a copy of MediaBody and risk the two getting out of synch, it seemed prudent to just use the MediaBody rule.

Ideally, I think the MediaBody rule would be renamed to something more generic, but that would change the CSS Library API, so guess that's a no go...


matthiasblaesing:
Thank you - I'll try to have a look at the changeset in the next few days.

matthiasblaesing:
This PR looks very good.

I looked over the code changes and they look sane. The same goes for the new unittests. The fixed tests are at least partly the result of my changeset to support the CSS variables, sorry about that, I would have sworn I ran through the unittests. Thanks for fixing!

The bootstrap CSS indeed opens without error. On the MDN site however is one sample that fails:

```css
@supports selector(A > B) {}
```

This can be merged without support for `selector` feature, I just want to make sure that you know about it, so that you can decide if you want to update it or not.

peedeeboy:
Thanks for checking over the code!  Much appreciated :)  

The selector() function appears to be Firefox only, and still experimental, so maybe to implement support later when it is bit more fleshed out?
https://developer.mozilla.org/en-US/docs/Web/CSS/@supports#Browser_compatibility

matthiasblaesing:
Agreed. Merging.

PR #1182, opened 2019-03-31
[NETBEANS-2327] fix bug to avoid program crash

bd2019us:
Fix: [#NETBEANS-2327](https://issues.apache.org/jira/browse/NETBEANS-2327)

### DIFF
--- a/java/performance/src/org/netbeans/modules/performance/utilities/CommonUtilities.java
+++ b/java/performance/src/org/netbeans/modules/performance/utilities/CommonUtilities.java
@@ -279,7 +279,7 @@ public class CommonUtilities {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
                 if (entry.isDirectory()) {
-                    new File(System.getProperty("nbjunit.workdir") + File.separator + ".." + File.separator + "data" + File.separator + entry.getName()).mkdir();
+                    new File(System.getProperty("nbjunit.workdir") + File.separator + ".." + File.separator + "data" + File.separator + entry.getName()).mkdirs();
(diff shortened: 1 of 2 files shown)


java/performance/src/org/netbeans/modules/performance/utilities/CommonUtilities.java
```diff
@@ -279,7 +279,7 @@ public static String jEditProjectOpen() {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
                 if (entry.isDirectory()) {
-                    new File(System.getProperty("nbjunit.workdir") + File.separator + ".." + File.separator + "data" + File.separator + entry.getName()).mkdir();
+                    new File(System.getProperty("nbjunit.workdir") + File.separator + ".." + File.separator + "data" + File.separator + entry.getName()).mkdirs();
```
matthiasblaesing: This assumes, that the upper levels of the directory where not yet created. The order in the zip is off. If that is the case, it will still blow in line 287, when the content file is to be written to the not existing directory.

matthiasblaesing:
General: please use valid author information for your git commits. Neither the name part, nor the email part look right.

I think this needs some more comments: is this a real problem or a theoretic problem? Indeed mkdirs can be used as a drop-in replacement, but it needs to be seen in context (see inline comment).

matthiasblaesing:
@bd2019us I notice, that you are opening more PRs, without reacting here. The changes looks random, so I'd like to ask you to discuss changes first on the mailing list.

matthiasblaesing:
No reply in over 6 months.

PR #1186, opened 2019-04-06
[NETBEANS-2346] Fixed index out of bound exception in GAV parsing.

lkishalmi:


matthiasblaesing:
I'm not an expert in this, but the change looks sane and works with the supplied sample project. From that +1.

PR #1188, opened 2019-04-06
[NETBEANS-2348] Added support for 'Test Package'.

lkishalmi:


lkishalmi:
I have something else less intrustive.

PR #1189, opened 2019-04-07
[NETBEANS-1775] Fix for NPE

lbruun:
Simple typo but with devastating effect: `&` should have been `&&`. This caused an NPE when NetBeans was used with a VPN connection. 

This fixes [NETBEANS-1775](https://issues.apache.org/jira/browse/NETBEANS-1775).


matthiasblaesing:
Merged - thanks.

PR #1194, opened 2019-04-10
[NETBEANS-2389] Missing source level 11 and 12 in a NetBeans module project

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-2389


junichi11:
Umm, it seems that Travis CI doesn't work.

PR #1196, opened 2019-04-12
[NETBEANS-2381] JDK 12 ver string support in nbi native win launcher

rtaneja1:
Handled JDK 11, 12 release version string like 11+28, 12+33.
Also registry entry path has changed. 
icon.ico was still old NetBeans icon, copied from nb/ide.launcher/windows/netbeans.ico 
 
Build nlw.exe as per : http://wiki.netbeans.org/LauncherBuild
But removed -mno-cygwin arg as compilation fails if this is present.

Updated binary zip.


matthiasblaesing:
Looks sane - should be squashed on merge. (I just looked at the changes and can't comment on the compilation).

rtaneja1:
Part fix was done in this: https://github.com/apache/incubator-netbeans/pull/734

geertjanw:
Thanks, to me it makes sense.

nbi/engine/native/launcher/windows/src/JavaUtils.c
```diff
@@ -138,7 +139,7 @@ JavaVersion * getJavaVersionFromString(char * string, DWORD * result) {
             major = (major) * 10 + c - '0';
             if (major > 999) return vers;
             continue;
-        } else if(c=='.'){
+        } else if(c=='.' || c=='+'){
```
sdedic: Why is this ?

rtaneja1: Without this, method returns null for JDK 12 version string 12+33
$ java -version
java version "12" 2019-03-19
Java(TM) SE Runtime Environment (build 12+33)
Java HotSpot(TM) 64-Bit Server VM (build 12+33, mixed mode, sharing)

rtaneja1: This doesn't set the build number in version, I'll handle this.

rtaneja1:
Thanks Svata, created NETBEANS-2528 to track build number issue

PR #1197, opened 2019-04-12
NETBEANS-2408 use System.currentTimeMillis()

bd2019us:
Hello,
new Date() is just a thin wrapper around System.currentTimeMillis(). Using System.currentTimeMillis() can help speed up the system.

### DIFF
--- a/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/AppClientProvider.java
+++ b/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/AppClientProvider.java
@@ -427,7 +427,7 @@ public final class AppClientProvider extends J2eeModuleProvider
             
             // set timeout to suppress the same messages during next 20 seconds (feel free to adjust the timeout
             // using more suitable value)
-            notificationTimeout = new Date().getTime() + 20000;
+            notificationTimeout = System.currentTimeMillis() + 20000;
         }
     }
     
--- a/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/AppClientProvider.java
+++ b/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/AppClientProvider.java
@@ -410,7 +410,7 @@ public final class AppClientProvider extends J2eeModuleProvider
     
     private void showErrorMessage(final String message) {
         // only display the messages if the project is open
-        if(new Date().getTime() > notificationTimeout && isProjectOpen()) {
+        if(System.currentTimeMillis() > notificationTimeout && isProjectOpen()) {
             // DialogDisplayer waits for the AWT thread, blocking the calling
             // thread -- deadlock-prone, see issue #64888. therefore invoking
             // only in the AWT thread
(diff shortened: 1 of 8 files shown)


lkishalmi:
I have the feeling this guy just would like to gather some Open Source footprint for whatever reason, by 'traveling around' with the same kind of "harmless" stuff in a few dozen Apache (maybe other) repositories. He joined to GitHub two months ago and already had 140 "contributions" to 58 repositories. He usually offers File.mkdir() -> File.mkdirs() or new Date().getTime() -> System.currentTimeMillis() changes.

Lets give @bd2019us  a week to speak up for himself. Otherwise we shall close the PR-s and the JIRA tickets.

bd2019us:
Hi @matthiasblaesing and @lkishalmi, sorry for our late reply and the trouble to you due to our identification issue.

We are a research group majoring in Software Engineering from a university in America, and we are working on a project related to bug detection in open-source projects. However, because of the anonymous policy of paper publication, we did not disclose our real identification. We are really sorry for the inconvenience. Definitely, if you are still not convinced, we are also happy to provide our detailed information by some way. In all, thanks for your understanding and support.

As for the code changes, all PRs were carefully written and examined by experienced developers in our group. The target of our research is to improve the quality of open-source projects. Though the improvement of some code changes is not very high, we still think it can help in some degree. Therefore, we appreciate for your kind consideration for the PRs.

Looking forward to your reply. Thanks.

matthiasblaesing:
Sorry - I don't buy that argumentation. I'm not a researcher, but I doubt any researcher would quote a work or use foreign research without knowing the author (even if just to be able to quote him/her). You are responsible for your work and thus should stand up to it with your name.

In this case I wonder for example, why line 430 was not changed. You also did not address the question why the change in this slow path method is a good idea, but not other cases in the netbeans code base.

bd2019us:
Hello!
We are sorry that we missed other cases like in line 430. The reason is that all the changed lines are firstly suggested by an automatic bug detection tool and then we manually checked each of them. In this process, we may miss some of them. Thanks for your kind reminding. If possible, we can finish the PR with careful check of all locations that may be relevant.

Following example shows that using `currentTimeMillis` is faster than `new Date().getTime()` when it is called many times.

56081029-36782380-5dce-11e9-8a19-3b4fbee4ce00

Also, to be convinced, we have updated the personal information in the profile.
Sorry for the inconvenience again and thanks for your understanding!

lleming:
There is a special performance utility http://openjdk.java.net/projects/code-tools/jmh/. Anyway, this test is abstract it just shows the obvious difference due to the nature of current circumstances. The reality is different. The current line will not be called thousands of times, every second. Better to improve code readability, using java date time api.

Chris2011:
I think the code is still readable. It doesn't matter whether I use new Date().getTime(); or System.getCurrentTimeMillis. And for me, the System... is more readable thatn the others. getCurrentTimeMillis give us exact what it is there for. My 2c.

lkishalmi:
Just being curious is this the only one new Date().getTime() in the entire codebase? If this PR fixes all occurrences (even that this is the only one), then from me I can let it go, however let it be clear in the JIRA issue that we are doing this for readability not for performance.

bd2019us:
> Just being curious is this the only one new Date().getTime() in the entire codebase? If this PR fixes all occurrences (even that this is the only one), then from me I can let it go, however let it be clear in the JIRA issue that we are doing this for readability not for performance.

Hello, @lkishalmi.  I just updated all locations of new Date().getTime(). Please check if they are good for PR fixes.   Thanks!

lkishalmi:
This is Ok with me now, I changed my viewpoint to natural: +0. If anyone else have other feelings about this change I respect that.

wadechandler:
@matthiasblaesing are you satisfied with the changes now?

matthiasblaesing:
As no-one came up to merge this in one month, I'm closing this.

PR #1199, opened 2019-04-13
[NETBEANS-1696] PHP 7.4 Support

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1696

- Null Coalescing Assignment Operator part
- Fix the parser and the lexer
- Add unit tests for formatter, parser, and lexer
- Add 7.4 to PhpVersion

#### Not Yet

- https://wiki.php.net/rfc/typed_properties_v2


junichi11:
I've investigated a bit. It seems that unit tests fail after this commit: https://github.com/apache/incubator-netbeans/commit/3357fa5b18590908ac9de4ad7e55ab1fdf88cd50

Unit tests passed when I've checkout https://github.com/apache/incubator-netbeans/commit/123f616488cbd335d63da493bb19b1aaabd0ac2f (previous commit)
e.g. 
- https://github.com/apache/incubator-netbeans/blob/master/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/actions/ImportDataCreatorTest.java
- https://github.com/apache/incubator-netbeans/blob/master/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/indent/PHPFormatterHtmlTest.java

@sdedic 
Do you have any ideas to fix this problem?

junichi11:
I've submitted the problem to the JIRA: https://issues.apache.org/jira/browse/NETBEANS-2437

junichi11:
The unit test problem was fixed :)

PR #1201, opened 2019-04-16
Netbeans 11: "Organize Imports" does not import new Classes anymore

deratzmann:
Hey, guys.
First of all: Thank you for your great work on netbeans. It is a great tool (and always was).
Here is my found bug:
With the release of netbeans 11 it seems, that the source code function "organize imports" is not working as expected.
My expectation (from the previous version):
- delete unused classes
- import currently unknown but used classes (by best guess)

currently it is only deleting unused classes.

So I guess, this might be a bug?

Best regards

Bastian


Chris2011:
@deratzmann please create an issue at JIRA: https://issues.apache.org/jira/projects/NETBEANS, you created a PR which is not correct.

PR #1202, opened 2019-04-16
[NETBEANS-2418][NETBEANS-2298] update suggested MySQL Driver to latest

mcdonnell-john:
There was a defect related to the JDBC Connection String for MySQL, I took a look and noticed we were suggesting an older driver.

I've updated the suggested driver to the latest, and ran some CRUD DB tests with it with NB 11.0 with no issues


mcdonnell-john:
Not sure, but looking at Travis:
The Checks fail on the check: RUN_TESTS_JDK9PLUS

But theres no errors and the end of the logs say:
"The job exceeded the maximum log length, and has been terminated."

Are we running too many checks in these Travis builds?

wadechandler:
Could be John. The changes themselves look just fine. Will check the mailing lists to see if this is a known thing too. On mobile right now.

matthiasblaesing:
One think that the driver reminded me at work: The driver class was moved to `com.mysql.cj.jdbc.Driver`, should that also be updated?

mcdonnell-john:
I didn't come across a change to the class when testing but will double check before merging to be sure

wadechandler:
It is more likely a deprecation warning https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-api-changes.html

Notice the old class name is deprecated. I would expect tests to fail, but it makes sense to update.

matthiasblaesing:
@mcdonnell-john should we mark this as work in progress?

mcdonnell-john:
Yep. My laptop broke a while back and only got a replacement a while back. I'm looking at the changes to driver and datasource classes at the moment but won't make 11.1

mcdonnell-john:
Closing due to re-basing issue

PR #573, opened 2018-06-03
[NETBEANS-417] Updating Windows launchers binaries.

jlahoda:
General comments:
-I've created new makefiles, as my primary development machine is Linux, it is set-up to do cross-compilation on Linux to produce the Windws binaries. This uses MinGW. The "old" makefiles are retained, we may do some more thorough cleanup later.
-the launchers need to be built against jni.h; version from Apache Harmony appears to work OK, so include that one (it should be OK, I assume: it is an Apache source code)
-the resulting binaries are marked to be under Apache License, which should be (hopefully) OK based on the "GCC RUNTIME LIBRARY EXCEPTION":
https://gcc.gnu.org/onlinedocs/libstdc++/manual/license.html
-testing by Windows users (esp. non-x86-64 users) greatly appreciated!

PR #1203, opened 2019-04-17
[NETBEANS-2361] Fix app.conf and launcher on macOS after changes in #573

oyarzun:
#573 removed the variable `default_mac_userdir` from the app.conf, but the shell script was still using that variable which results in incorrect userdir. It also should check for the old variable in case the RCP app was build with a custom app.conf that has the old variable.

* Remove extra " in app.conf 
* Check for old `default_mac_userdir` variable in case RCP has a custom app.conf with old settings; otherwise use `default_userdir`

### DIFF
--- a/harness/apisupport.harness/release/etc/app.conf
+++ b/harness/apisupport.harness/release/etc/app.conf
@@ -54,7 +54,7 @@
 #
 
 default_userdir="${DEFAULT_USERDIR_ROOT}/dev"
-default_cachedir=""${DEFAULT_CACHEDIR_ROOT}/dev"
+default_cachedir="${DEFAULT_CACHEDIR_ROOT}/dev"
 
 # options used by the launcher by default, can be overridden by explicit
 # command line switches
--- a/harness/apisupport.harness/release/launchers/app.sh
+++ b/harness/apisupport.harness/release/launchers/app.sh
@@ -57,7 +57,11 @@ args=""
 
 case "`uname`" in
     Darwin*)
-        userdir="${default_mac_userdir}"
+        if [ ! -z "$default_mac_userdir" ]; then
+          userdir="${default_mac_userdir}"
+        else
+          userdir="${default_userdir}"
+        fi
         ;;
     *)
         userdir="${default_userdir}"


neilcsmith-net:
@sdedic under current draft release plan, NB 14 will be in April 2022! :smile: Maybe just use the date for now?

oyarzun:
We could just remove the workaround. Then add a [wiki](https://netbeans.apache.org/wiki/index.asciidoc) entry under "Development issues, module basics and classpath issues, and information about RCP/Platform application configuration" that explains that `default_mac_userdir` no longer exists.

branscha:
The change is still not in 11.1 which was released in July; is there some documentation on how the fixes reach production?

oyarzun:
@branscha This was merged in before 11.1 was released. It is included in the binaries from https://netbeans.apache.org/download/nb111/nb111.html. 

I just tried created an ant based platform application using 11.1, packaged it as a Mac OS X Application. The app.conf and the app.sh has the changes from this merge request.

Is your platform application maven based?

branscha:
@oyarzun I am creating a netbeans platform application on Linux (not on OSX), and  I still get the extra " in  the generated app.conf. The generated project is maven based. I thought that the fix would also have solved the Linux version, but probably it was only applicable for OSX.

oyarzun:
@branscha It looks like the problem is that nbm-maven-plugin needs to be updated and built with the new app.conf file. @ebarboni Can you confirm this?

branscha:
@oyarzun @ebarboni Yes that seems to be the case. I will try to find the correct location to mention the issue. Thanks for helping out.

neilcsmith-net:
@branscha yes, I noticed this issue too while testing Maven post-release, and commented "somewhere" - I think on a related JIRA ticket.  There are a few issues with Maven support for Apache releases that we can hopefully get fixed in 11.2.

In answer to your wider question about how fixes hit production - anything in master on the release date is in the release.

oyarzun:
@branscha FYI, the current workaround would be to add `etcConfFile` to your pom.

```xml
<configuration>
     <etcConfFile>src/main/resources/my.conf</etcConfFile>
</configuration>
```

See Geertjan's blog entry https://blogs.oracle.com/geertjan/nimbus-look-and-feel-for-maven-based-netbeans-rcp-applications

chema-valls:
> @branscha FYI, the current workaround would be to add `etcConfFile` to your pom.
> 
> ```
> <configuration>
>      <etcConfFile>src/main/resources/my.conf</etcConfFile>
> </configuration>
> ```
> 
> See Geertjan's blog entry https://blogs.oracle.com/geertjan/nimbus-look-and-feel-for-maven-based-netbeans-rcp-applications

@oyarzun , @branscha
That workaround did not work form me but adding the property `netbeans.conf.file` to the pom did:
```xml
<properties>
   <netbeans.conf.file>src/main/resources/my.conf</netbeans.conf.file>
</properties>
```

PR #1204, opened 2019-04-19
[NETBEANS-2409](partial): attempting to avoid FileUtil.normalizeFile …

jlahoda:
…inside FileObject.getFileObject in java.openjdk.project related cases.

I tried to avoid the normalizeFile call inside FolderObj.getFileObject (in masterfs), but I don't seem to have enough knowledge for that. But OpenJDK project is responsible for invocation of getFileObject("../<something>"), so avoid that call inside OpenJDK project should help. Unless someone knowledgeable of FileSystems wants to speed up them.

### DIFF
--- a/ergonomics/ide.ergonomics/src/org/netbeans/modules/ide/ergonomics/fod/FeatureProjectFactory.java
+++ b/ergonomics/ide.ergonomics/src/org/netbeans/modules/ide/ergonomics/fod/FeatureProjectFactory.java
@@ -140,7 +140,15 @@ implements ProjectFactory, PropertyChangeListener, Runnable {
         }
 
         final boolean hasFile(String relative) {
-            return dir.getFileObject(relative) != null;
+            FileObject d = dir;
+            int pos = 0;
+
+            while (relative.startsWith("../", pos) && d != null) {
+                d = d.getParent();
+                pos += 3;
+            }
+
+            return d != null && d.getFileObject(relative) != null;
         }
 
         final boolean isDeepCheck() {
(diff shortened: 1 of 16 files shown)


ergonomics/ide.ergonomics/src/org/netbeans/modules/ide/ergonomics/fod/FeatureProjectFactory.java
```diff
@@ -140,7 +140,15 @@ Document dom(String relative) {
         }
 
         final boolean hasFile(String relative) {
-            return dir.getFileObject(relative) != null;
+            FileObject d = dir;
+            int pos = 0;
+
+            while (relative.startsWith("../", pos) && d != null) {
```
JaroslavTulach: This is a fix that could go into `AbstractFileObject` and `masterfs` own `BaseFileObj`.

jlahoda: Yes. I think there's something like this in openide.filesystems, but not in masterfs. The reason I shied away from doing this in masterfs is that this actually issues the FileObjects, and could be slower in some cases (on Linux, AFAIK, normalizeFile is very cheap). I think here, the slowdown is probably OK, but in masterfs, not so sure. I tried to implement without issuing the FOs, but that seems to be beyond my knowledge.

geertjanw:
Any reason not to merge this?

lkishalmi:
I do not see any.

ebresie:
Is this a candidate for 11.1?

neilcsmith-net:
@ebresie at this stage, I would say probably not, although someone else might have a different opinion?  Somehow we've missed flagging this for merging prior to freeze or inclusion in a beta release.  And sorry, I also missed the link to a critical issue in JIRA for this, although it's only marked as a partial fix.

At this stage, I think we need to hold for NB11.2, and merge along with @lkishalmi NIO2 change and prioritise some of those performance issues with files early in the next release cycle?

lkishalmi:
Let's have this merged then.

PR #1206, opened 2019-04-20
Ensuring important DTDs are used locally when running java.completion…

jlahoda:
… tests.

Adding the DTDs to the web was nice, but didn't seem to help the builds much. So trying to resolve this for java.completion tests by using the local copies of the DTDs.


lkishalmi:
Looks sane. Merging.

vieiro:
Wow, that's a fast merge :-)
What about using netbeans.apache.org as the URL? That's already there (no redirections required)

lkishalmi:
That could work as well, though I really value unit test independence of external environment.

vieiro:
That's a sound reason to have all DTDs included in a directory somewhere :-)

lkishalmi:
I could really support that move! Shall I file a JIRA Improvement on that?

vieiro:
Definitely! Feel free to assign it to me.

jtulach:
Using DTDs locally is the best fix. Thanks @jlahoda.

PR #1208, opened 2019-04-22
[NETBEANS-2274] Remove special casing of JComboBox handling in language chooser

matthiasblaesing:
In the Options Dialog, the Panel "Fonts & Colors" and there the subpanel
Syntax Pane relies on undocumented ActionEvent#getModifiers behavior.

The currently choosen language is only updated when:

- the new language is acknowledged with a space/enter keypress
- or choosen with a left mouse click

Two problems:
1. The language in the combobox and the currently select language can
   drift apart. (use only cursor keys to change selection)
2. On Ubuntu Linux (OpenJDK 1.8u191) it was observed, that the
   undocumented behaviour, the action handler relied on, was not
   present.

Both problems are solved by using the default behaviour of the JComboBox:

Everytime the selection changes and action event is fired, regardless
of the source of the selection change.


matthiasblaesing:
I plan to merge this sometime next week.

PR #1210, opened 2019-04-23
[NETBEANS-2392] adjusting handling of erroneous types in TreeFactory …

jlahoda:
…to current javac model


lkishalmi:
LGTM, also has tests, so I've merged it. Just one question: is this issue fixed in nb-javac as well? Or it is totally irrelevant from the compiler being used?

jlahoda:
This is (should be) independent on which javac is used.

PR #1211, opened 2019-04-23
[NETBEANS-2350]NPE while doing auto complete in rule switch block

vikasprabhakar:
Null pointer exception while doing auto complete in rule switch block

Code scenario:

enum Days{MON, TUES, WED, THUR, FRI};

private static void test1(Days day) {
switch (day)

{ case FRI -> System.out.println("javaapplication2.JavaApplication2.test1()");  
<Caret position>   --> Auto complete press 'Ctrl+Space' throw Null pointer exception }
}


java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
```diff
@@ -2291,7 +2291,7 @@ private void insideSwitch(Env env) throws IOException {
                 }
                 lastCase = t;
             }
-            if (lastCase != null) {
+            if (lastCase != null && lastCase.getStatements() != null) {
```
arusinha: Can use
 caseBody = TreeShims.getBody(lastCase);
                if (caseBody instanceof StatementTree)
                    statements = Collections.singletonList((StatementTree) caseBody);

PR #1213, opened 2019-04-24
Making sure profiler.oql tests pass.

JaroslavTulach:
Adding a heap dump file omitted on the Apache donation. Enabling testing as part of the `scripting.sh` travis script.


profiler/profiler.oql/test/unit/src/org/netbeans/modules/profiler/oql/engine/api/impl/OQLEngineTest.java
```diff
@@ -515,6 +516,7 @@ public boolean visit(Object o) {
                 return false;
             }
         });
+        output[0] = output[0].replaceAll("UID=301077366599181[0-9]*", "UID=301077366599181567");
```
matthiasblaesing: Could you please add a comment what this making replacement does? Not so much for the change itself, as for future readers, that wonder what was this meant to say.

JaroslavTulach: According to @thurka there was some bug in the serialization that affects the `UID` value. Tomáši, can you share the JDK bug reference?

profiler/profiler.oql/test/unit/src/org/netbeans/modules/profiler/oql/engine/api/impl/OQLEngineTest.java
```diff
@@ -50,6 +50,7 @@ public OQLEngineTest() {
 
     @BeforeClass
     public static void setUpClass() throws Exception {
+        System.setProperty("polyglot.js.nashorn-compat", "true");
```
JaroslavTulach: This magic option will be moved to `Bindings` [once this change](https://github.com/graalvm/graaljs/commit/af68dca48fa16f7542e6f363ca8099d02a3d8ccd) gets in.

JaroslavTulach:
> add is a licenseinfo.xml file, so that the license info is still present 

Done in b4cdd1b71995da60355185724bb4a684f18feb49

profiler/profiler.oql/test/unit/src/org/netbeans/modules/profiler/oql/engine/api/impl/OQLEngineTest.java
thurka: https://netbeans.org/bugzilla/show_bug.cgi?id=256761. JDK bug is already fixed, there is no need for workaround.

JaroslavTulach:
The two failing checks are unrelated to `profiler.oql` changes. Merging anyway. CCing @jlahoda

PR #1214, opened 2019-04-24
uihandler/arch.xml → platform/uihandler/arch.xml

ooxi:
`uihandler/arch.xml` was moved in 3dfe01225c66a7b6438801621b4d111f946e64b5 and thus the instructions from `README.md` no longer work out of the box

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ you may want to merge the modern and ancient versions together:
 ```bash
 $ git clone https://github.com/apache/incubator-netbeans.git
 $ cd incubator-netbeans
-$ git log uihandler/arch.xml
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
The change itself looks sane - could you please update the commit with valid user data? ooxi looks like a nickname and not a realname. (see https://www.git-tower.com/learn/git/faq/change-author-name-email for a description, you will need to do a force-push after the update).

ooxi:
Thanks for the feedback @matthiasblaesing, force pushed commit to change username :-)

matthiasblaesing:
Thank you for the update. Merging.

PR #1216, opened 2019-04-24
NETBEANS-888 Open Recent file not working

michaelsembwever:
- update menuItems under RecentFileAction via a RecentFiles property change listener, rather than popup menu listener

### DIFF
--- a/ide/utilities/src/org/netbeans/modules/openfile/RecentFileAction.java
+++ b/ide/utilities/src/org/netbeans/modules/openfile/RecentFileAction.java
@@ -25,19 +25,18 @@ import java.awt.Point;
 import java.awt.PointerInfo;
 import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
 import java.io.File;
 import java.util.List;
 import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
-import javax.swing.event.PopupMenuEvent;
-import javax.swing.event.PopupMenuListener;
 import org.netbeans.modules.openfile.RecentFiles.HistoryItem;
 import org.openide.awt.*;
 import org.openide.filesystems.FileUtil;
 import org.openide.util.NbBundle;
 import org.openide.util.RequestProcessor;
-import org.openide.util.Utilities;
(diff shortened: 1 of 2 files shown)


geertjanw:
Wow, cool. Why was this needed, why did it work in earlier releases but not anymore requiring this PR? Thanks for this great work.

michaelsembwever:
@geertjanw ,

> Wow, cool. Why was this needed, why did it work in earlier releases but not anymore requiring this PR? Thanks for this great work.

I suspect it never worked on the combination of
 - macos
 - `apple.laf.useScreenMenuBar=true`
 - jdk >1.8 (currently experiencing it with 11.0.3)

and that these users using Java9 or above is a relatively new phenomena.

geertjanw:
So it was macos specific, didn't know that. And also related to some change in the JDK after 1.8, but what could it be (just curious really).

michaelsembwever:
@geertjanw ,

> So it was macos specific, didn't know that. And also related to some change in the JDK after 1.8, but what could it be (just curious really).

I didn't (don't) have the time to hunt that down.
The existing code was recreating the menuItems every time the popup menu listener notified the jmenuitem was made visible. The ScreenMenuItems was then without reference back to these menuItems, and so there's was no actionListener and the event was not enabled.

Changing it so that the menuItems were only recreated when the history in RecentFiles changed, solved the problem, probably because the ScreenMenuItems <-> menuItem referencing is done before the popup menu change support kicks in, but that's just a guess.

There's a bit more of my analysis info in https://issues.apache.org/jira/browse/NETBEANS-888

michaelsembwever:
@geertjanw ping.

GeertjanWielenga:
Sure, how ready is this for merging? Happy to merge if you’re confident about this pr.

michaelsembwever:
I've been using it now the past week.

geertjanw:
OK, merged.

matthiasblaesing:
@michaelsembwever please update your git settings. I just noticed, that the history does not reflect your real name. For commits valid realname and email information should be present.

michaelsembwever:
@matthiasblaesing , thanks for the heads up. Has been set to `Author: Mick Semb Wever <mck@apache.org>`

matthiasblaesing:
@michaelsembwever thank you, we welcome your future changesets :-)

PR #1219, opened 2019-04-26
Popupswitcher npe fixes

timboudreau:
The attached patch fixes two rare but irritating NPEs from the popup switcher: If enough files are open that the popup switcher has two columns, but there is an odd number of files open, then the final item in the right column of the table has no item - but the code assumed it would always be non-null, so hovering or clicking the close button area on that spot generates a NullPointerException.


geertjanw:
Great, thanks!

matthiasblaesing:
I was able to reproduce the problem and verified the fix. The code change looks sane. Will merge. Thank you!

PR #1223, opened 2019-04-27
[NETBEANS-2467] Using the correct jcodings dependency to avoid touch …

jlahoda:
…nonexisting field.


matthiasblaesing:
There is one unittest failing "validly":

> BUILD FAILED
> /home/travis/build/apache/netbeans/nbbuild/build.xml:1568: Failed VerifyLibsAndLicenses test(s):
> Some binaries are duplicated (edit nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ignored-overlaps as needed)
> **ide/libs.bytelist/external/jcodings-1.0.18.jar and ide/textmate.lexer/external/jcodings-1.0.18.jar are identical**

So we need either a libs.jcodings module, then we can share the depencency between the two modules or need to decleare and exception.

jlahoda:
Huh. I am not really eager to put a library I know very little of to our APIs. Adding exception.

matthiasblaesing:
I'm ok with this. Maybe we can create a library module, that is only used internally, in the future. We have other libraries, that are netbeans-only (require implementation dependencies or fried-status), so we can get away without API.

PR #1221, opened 2019-04-27
Remove incubator parts

junichi11:
- Remove "incubator"
- Fix the Travis CI and the Jenkins build links
- Remove DISCLAIMER

PR #1224, opened 2019-04-29
Removed incubator- references from README.md

lhochet:
Removed the incubator- or incubator. mentions in the README.md following the graduation of the project


junichi11:
Thank you! However, there is already the PR for it. #1221 
So will close this.

PR #1225, opened 2019-04-30
[NETBEANS-2119] Let the netbeans-dev Snapcraft package happen

lkishalmi:
Well, i did not give up on having a "rolling" release for testing from the master branch. It needs a different snap package name: netbeans-dev, so it could be private or public but hidden from the store (those who knew the name would be able to install it). Right now the package is being kept private.

So having these files in place and: https://builds.apache.org/job/netbeans-snap-weekly-master
We could have an auto updated test version from master by every Saturday morning. (Unfortunately Linux only)


geertjanw:
Excellent plan, merge it. :-)

PR #1226, opened 2019-04-30
Lazy parameter names

jlahoda:
Currently, when running without nb-javac, when a class is load from a classfile without debugging info (and without the MethodParameters attributes on its methods), the parameters of methods in the class will be autogenerated.

With nb-javac, these can be load lazily from sources, javadoc, or better names can be autogenerated.

Since (the upcoming) JDK 13, vanilla javac has ability that allows to lazily inject parameter names if they are missing:
http://hg.openjdk.java.net/jdk/jdk/rev/b0651dcc8d98

This patch strives to enhance NetBeans to use this new ability to lazily inject parameter names.


GeertjanWielenga:
Excellent.

lkishalmi:
How would this one affect pre-Java 13 operations?

jlahoda:
The intent is that it shouldn't affect pre-JDK 13 usage - if the ParameterNameProvider is not available, the registration should fail silently and everything should work as before.

PR #1227, opened 2019-05-01
Add position info to gradle extension registration

matthiasblaesing:
This fixes the startup message:

> WARNING [org.openide.filesystems.Ordering]: Not all children in Services/MIMEResolver/ marked with the position attribute: [org-netbeans-modules-gradle-GradleDataObject-Extension.xml], but some are: [org-openide-loaders-DataLoaderPool$InstanceLoaderSystem-Extension.xml, org-netbeans-core-startup-layers-SystemFileSystem-Extension.xml, org-netbeans-modules-java-JavaDataObject-Extension.xml, [...]


lkishalmi:
Thank you!

PR #1228, opened 2019-05-02
Syntax coloring for .kt files

JaroslavTulach:
Last week I had a presentation in Ostrava. The second speaker, Leoš, was talking about Kotlin. Hence I decided to demonstrate native image with the sieve of Eratosthenes written in Kotlin.

I had to write a NetBeans module that provides syntax coloring for the .kt files. Here is the code that I have. If OK, I'd like to push it to master branch.


GeertjanWielenga:
Excellent.

JaroslavTulach:
This PR can serve as an example how to integrate syntax coloring for any language into NetBeans:

1. find the TextMate grammar somewhere (mine is from [kotlin-sublime-package](https://github.com/vkostyukov/kotlin-sublime-package/blob/ce599a846ac9f86d4671a024d21220ca08e50f66/Kotlin.JSON-tmLanguage) project, ASF 2.0 licensed, so OK, right @matthiasblaesing?)
2. use the FileType wizard to generate `DataObject` representing the file in editor
3. use `@GrammarRegistration` annotation to connect the two

Chris2011:
Cool. I did it similar in my repo: https://github.com/Chris2011/TextmateBeans. Didn't know that a @GrammarRegistratrion exists, but I see, only for the NetBeans core, not usable for 3rd-party-plugins.

I had a look into the code and it looks straight forward :)

jtulach:
Re. TextMateBeans - wouldn't it be better to donate all of that to Apache? Then I wouldn't have to bother with Kotlin (only Kotlin) myself!

Chris2011:
Yes but first, I had an other intention to create such plugin, for this we can talk in private :) and second not all of that languages are working as expected. I created 4 tickets in case of problem with the textmate lexer. But also another topic.

JaroslavTulach:
> ... I had an other intention to create such plugin...

OK, I'll integrate my changes then. If you even consider contributing to Apache NetBeans, you can modify my commits.

PR #1229, opened 2019-05-03
Correct license information for kotlin icon/grammar

matthiasblaesing:
The kotlin grammar and icon are not donated to the ASF, so the correct license reference is "Apache-2.0" (without the "-ASF" suffix). The difference between the two:

Apache-2.0-ASF means the file could be relicensed (for example when an apache license 3 is released), as it is owned by the ASF. Apache-2.0 means the file is licenses ALv2, but is not owned by the ASF.

This PR adjusts the license reference for the kotlin grammar. The icons were updated from the kotline repository. The license in the repository clearly states, that the license is ALv2 and the exceptions don't match the image files. However, the kotlin license comes with a NOTICE file, which is now included in the notice.


JaroslavTulach:
Thanks for polishing the mess I had created.

PR #1231, opened 2019-05-04
[NETBEANS-2500] Java source file palette

geertjanw:
When Java source file is open, palette is available, with one demo item, and text from the Java Editor can be dragged into the palette.


java/java.editor/src/org/netbeans/modules/java/editor/palette/JavaSourceFileLayerPaletteFactory.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    
+    public static PaletteController createPalette() {
+        try {
+            if (null == palette)
+                palette = PaletteFactory.createPalette(
+                        JAVA_PALETTE_FOLDER, 
+                        new MyActions(), 
+                        null, 
+                        new JavaDragAndDropHandler());
```
sdedic: This can be `null`, the default handler is also no-op

java/java.editor/src/org/netbeans/modules/java/editor/palette/JavaPaletteCustomizerAction.java
```diff
@@ -0,0 +1,38 @@
...
+package org.netbeans.modules.java.editor.palette;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import org.openide.awt.ActionID;
+import org.openide.awt.ActionReference;
+import org.openide.awt.ActionRegistration;
+import org.openide.util.NbBundle.Messages;
+
+@ActionID(id = "org.netbeans.modules.html.palette.HtmlPaletteCustomizerAction", category = "Tools")
```
sdedic: html customizer ID in java-related module ? Copypaste error ?

java/java.editor/src/org/netbeans/modules/java/editor/palette/items/Item.java
```diff
@@ -0,0 +1,75 @@
...
+    private String comment = "";
+    
+    public Item() {
+    }
+    
+    private String createBody() {
+        String comment = getComment();
+        StringBuilder buffer = new StringBuilder();
+        buffer.append( "/** " );
+        buffer.append( comment );
```
sdedic: I guess because this is a javadoc, it could be better formatted so each line gets '*' prefix etc.

java/java.editor/src/org/netbeans/modules/java/editor/palette/items/resources/Item.xml
```diff
@@ -0,0 +1,32 @@
...
+-->
+<!DOCTYPE editor_palette_item PUBLIC "-//NetBeans//Editor Palette Item 1.0//EN" "http://www.netbeans.org/dtds/editor-palette-item-1_0.dtd">
+
+<editor_palette_item version="1.0">
+    <class name="org.netbeans.modules.java.editor.palette.items.Item" />
+
+    <icon16 urlvalue="org/netbeans/modules/java/editor/palette/items/resources/Item16.png" />
+    <icon32 urlvalue="org/netbeans/modules/java/editor/palette/items/resources/Item32.png" />
+    <description localizing-bundle="org/netbeans/modules/java/editor/palette/items/resources/Bundle"
+               display-name-key="NAME_html-ITEM"
```
sdedic: Better naming: _html seems as a copy-paste leftover.

geertjanw:
Thanks for the review.

I'm not sure if one item should be in the palette by default, maybe best to leave it completely empty. I would like to think more about this and I think we should not include this in 11.1, marking it for 11.2.

geertjanw:
I'd like to redo this, by providing an empty palette, that is not opened by default for the Java source files, via a new pull request, but not for 11.2,

PR #1236, opened 2019-05-07
[NETBEANS-2514] - remove sun.misc.Unsafe

BradWalker:
Remove usage of sun.misc.Unsafe in NoJavacHelper.java

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
@@ -73,14 +73,13 @@ int major = Integer.parseInt(javaVersionElements[1]);
 
                 // Java versioning is changing and has never really made sense.
                 // Look at JEP 322 for the current scheme. 
-                if (((javaVersionElements[0] == 1) && (javaVersionElements[1] >= 8)) ||
-                    (javaVersionElements[0] >= 10)) {
+                if (((Integer.parseInt(javaVersionElements[0]) == 1) && (Integer.parseInt(javaVersionElements[1]) >= 8)) ||
+                    (Integer.parseInt(javaVersionElements[0]) >= 10)) {
                     try {
                         Method defineClass = MethodHandles.Lookup.class.getDeclaredMethod("defineClass", Byte[].class); //NOI18N
                         defineClass.invoke(MethodHandles.lookup(), classData);
                     } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException ex) {
                         Logger.getLogger(NoJavacHelper.class.getName()).log(Level.FINE, null, ex);
-
                     }
                 } else {
                     // Flag an error here?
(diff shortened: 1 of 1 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
```diff
@@ -69,35 +68,20 @@ public void run() {
...
-                boolean isJdkVer8OrBelow = true;
-                if (!JavaVersion.startsWith("1.")) {   //NOI18N
-                    isJdkVer8OrBelow = false;
-                }
-                if (isJdkVer8OrBelow) {
-                    try {
-                        Field theUnsafe = Unsafe.class.getDeclaredField("theUnsafe"); //NOI18N
-                        theUnsafe.setAccessible(true);
-                        Unsafe unsafe = (Unsafe) theUnsafe.get(null);
+                String[] javaVersionElements = System.getProperty("java.version").split("\\.");
```
matthiasblaesing: Sorry - this will break with newer version of Java: `java.version = 13-ea`
`

java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
```diff
@@ -69,35 +68,20 @@ public void run() {
...
+                int major = Integer.parseInt(javaVersionElements[1]);
```
matthiasblaesing: This looks only sane on the old version scheme 1.X.Y. As shown above, the version scheme is shifting...

java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
BradWalker: Agreed..

jlahoda:
So, let us think on why there was a use of Unsafe in the first place. The reason was that, when the ability to use javac from JDK instead of the nb-javac fork was added, the use of javac's APIs that were JDK 9 specific (and were in nb-javac) was way too widespread, and it was simply unfeasible to make the Java support work using javac from JDK 8. So, the Java editing features in NetBeans can run with nb-javac installed on JDK 8+, or using javac from JDK (without nb-javac) on JDK 9+, but not using JDK's javac on JDK 8. But it was desirable that the IDE is able to start, and does not crash too hard when running without nb-javac on JDK 8. One step to achieve that was to inject the WriteableScope (which is in JDK 9, but not in JDK 8, and is used somewhere too hard/early in the Java support in NB, I don't recall where exactly). One thing to note is that this is neither necessary nor done on JDK 9+. So it seemed (and still seems to me) OK to use Unsafe.defineClass. After all, I don't think there is a better way to do that in JDK 8.

In any case, using MethodHandles.Lookup.defineClass(byte[]), which is (AFAIK) @since 9, feels useless to me.

matthiasblaesing:
@BradWalker ware you still interested in this PR? I think @jlahoda reply needs to be adressed.

PR #1241, opened 2019-05-09
[NETBEANS-2506] Added OpenJFX Samples module to JavaFX cluster.

jperedadnr:
This PR adds two Maven samples that show how to run JavaFX 11+ (with or without FXML), with the JavaFX Maven plugin.


JaroslavTulach:
Wouldn't it be better to extract the project skelton from a Maven archetype? CCing @eppleton ...

jperedadnr:
Actually the two samples in the PR are just the application of the [JavaFX Maven archetypes](https://github.com/openjfx/javafx-maven-archetypes), and these can be easily generated via `New Project -> Java with Maven ->  Project from Archetype`, selecting either of the two existing archetypes. This is also well documented here: https://openjfx.io/openjfx-docs/#IDE-NetBeans (Maven sections).

However, the purpose of this PR is to "bundle the Open JFX NetBeans samples", so while these two samples are the same anyone could generate via archetypes or via the [OpenJFX samples repository](https://github.com/openjfx/samples/blob/master/IDE/NetBeans/Modular/Maven), the user will get them ready to use from `New Project -> Samples`.

What's more,  these samples are a little bit more integrated with NetBeans (they contain custom nbactions that override `run` and `debug` default actions).

geertjanw:
Makes sense to me, and many thanks for this contribution from Gluon, may there be more!

jtulach:
> Makes sense to me, and many thanks for this contribution from Gluon, may there be more!

@geertjanw, I understand that you are excited seeing new contributor. Actually I am as well. Welcome José. 

That however doesn't mean that the contribution make sense to me.

> ... just the application of the [JavaFX Maven archetypes](https://github.com/openjfx/javafx-maven-archetypes), 

Good. There is an archetype. 

> via `New Project -> Java with Maven -> Project from Archetype`, 

I see. This is not the most straightforward way to integrate a technology into NetBeans. JavaFX indeed deserves a better wizard!

> However, the purpose of this PR is to "bundle the Open JFX NetBeans samples", so while these two samples are the same anyone could generate via archetypes or via the [OpenJFX samples repository](https://github.com/openjfx/samples/blob/master/IDE/NetBeans/Modular/Maven), the user will get them ready to use from `New Project -> Samples`.

There already is "New Project/Java with Maven/JavaFX Application" - why don't you rather fix/modify that one to be a nice UI over the archetypes? NetBeans has a nice support for creating wizards over Maven archetypes. I suggest to use it.

> What's more, these samples are a little bit more integrated with NetBeans (they contain custom nbactions that override `run` and `debug` default actions).

I'd suggest to put those files into the archetypes. Unless I am mistaken @eppleton does that for the DukeScript archetypes. Unacceptable? Even if hidden by a archetype property?

José, please don't let me discourage you. We need Gluon to polish the JavaFX integration in NetBeans. I just want to do it the right "Maven (UI) way". I'd like NetBeans to be the _UI for Maven_ - e.g. just a wizard that harnesses Maven and its archetypes.

geertjanw:
I propose we go ahead with this pull request and open a new issue for the better approach described by @jtulach. If no problem with this in 24 hours, I will merge this.

geertjanw:
OK, merging.

neilcsmith-net:
(as I was adding a similar comment on something else)

If these stay as is for 11.2 then we need to fix master in the same way as 11.1 for these to actually work - see mailing list thread at https://lists.apache.org/thread.html/69de8f6e1cdcfcc65b0c92a81ea9c3724058b89e5a56adea0b3d79cd@%3Cdev.netbeans.apache.org%3E  Or we replace as suggested above?

PR #1242, opened 2019-05-09
 [NETBEANS-2509] Bring back docker plugin support for unix sockets

Ihsahn:
Restore ability to connect to docker via unix socket, replaced no-longer existing socket access library with (wrapped) [com.kohlschutter.junixsocket](https://github.com/kohlschutter/junixsocket) (Apache License).

Checked on `linux/amd64` - I was able to connect and list images/containers.

_ps. according to junixsocket docs this should also work on `linux/aarch64` and `macos`, but since I don't have ability to test on these I'm not enabling them (yet)_

### DIFF
--- a/ide/docker.api/src/org/netbeans/modules/docker/api/DockerSupport.java
+++ b/ide/docker.api/src/org/netbeans/modules/docker/api/DockerSupport.java
@@ -24,7 +24,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.prefs.NodeChangeEvent;
 import java.util.prefs.NodeChangeListener;
@@ -129,12 +128,12 @@ public final class DockerSupport {
     }
 
     public boolean isSocketSupported() {
-        return false;
-//        if (BaseUtilities.getOperatingSystem() != BaseUtilities.OS_LINUX) {
-//            return false;
-//        }
-//        String arch = System.getProperty("os.arch"); // NOI18N
-//        return arch != null && (arch.contains("x86") || arch.contains("amd64")); // NOI18N
+        if (BaseUtilities.getOperatingSystem() != BaseUtilities.OS_LINUX
+                && BaseUtilities.getOperatingSystem() != BaseUtilities.OS_MAC) {
+            return false;
+        }
+        String arch = System.getProperty("os.arch"); // NOI18N
+        return arch != null && (arch.contains("x86") || arch.contains("amd64")); // NOI18N
     }
 
     private boolean isInitialized() {
--- a/ide/libs.c.kohlschutter.junixsocket/build.xml
+++ b/ide/libs.c.kohlschutter.junixsocket/build.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<project name="ide/libs.c.kohlschutter.junixsocket" default="build" basedir=".">
+    <import file="../../nbbuild/templates/projectized.xml"/>
+</project>
--- a/ide/docker.api/nbproject/project.xml
+++ b/ide/docker.api/nbproject/project.xml
@@ -26,6 +26,15 @@
             <code-name-base>org.netbeans.modules.docker.api</code-name-base>
             <module-dependencies>
                 <dependency>
+                    <code-name-base>libs.c.kohlschutter.junixsocket</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>1</release-version>
+                        <specification-version>2.20</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
                     <code-name-base>org.netbeans.api.annotations.common</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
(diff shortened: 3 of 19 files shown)


ide/docker.api/src/org/netbeans/modules/docker/api/DockerSupport.java
```diff
@@ -129,12 +128,11 @@ public void removeChangeListener(ChangeListener listener) {
     }
 
     public boolean isSocketSupported() {
-        return false;
-//        if (BaseUtilities.getOperatingSystem() != BaseUtilities.OS_LINUX) {
-//            return false;
-//        }
-//        String arch = System.getProperty("os.arch"); // NOI18N
-//        return arch != null && (arch.contains("x86") || arch.contains("amd64")); // NOI18N
+        if (BaseUtilities.getOperatingSystem() != BaseUtilities.OS_LINUX) {
```
sdedic: What about MacOS X ? `junixsocket`'s documenation says it supports mac too...

Ihsahn: I stated this at the very top already: I don't have ability to verify this on `linux/aarch64` and `macos`

matthiasblaesing:
The binaries should not be checked in to git. The netbeans build system has its own code to download binaries at build time. The mechanism was updated to also work with maven central, when netbeans was donated to apache.

For a sample have a look at the jna module: `platform/libs.jna`. There are two things you'll need:

- `external/binaries-list`: This lists the binaries with SHA1 sums
- `external/<binary-name>-<version>-license.txt` with the license information about the file

ide/libs.c.kohlschutter.junixsocket/build.xml
```diff
@@ -0,0 +1,24 @@
...
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<project name="ide/libs.c.kohlschutter.junixsocket" default="build" basedir=".">
+    <import file="../../nbbuild/templates/projectized.xml"/>
+</project>
```
JaroslavTulach: The library seems to be on Maven central. Make sure to create `external/binaries-files` to download the JARs from Maven central instead of having them in the Git repository.

ide/libs.c.kohlschutter.junixsocket/external/junixsocket-common-2.2.0-license.txt
```diff
@@ -0,0 +1,210 @@
+Name: junixsocket
+License: Apache-2.0
+OSR: 4300
```
matthiasblaesing: This is an artifact from the oracle/sun era. Please remove.

ide/libs.c.kohlschutter.junixsocket/external/junixsocket-common-2.2.0-license.txt
```diff
@@ -0,0 +1,210 @@
...
+Description: junixsocket is a Java/JNI library that allows the use of Unix Domain Sockets (AF_UNIX sockets) from Java.
+Origin: https://github.com/kohlschutter/junixsocket/
+Version: 2.2.0
```
matthiasblaesing: There is no need for three copies of the license file. Please remove two copies and rename the remaining copy to "junixsocket-license.txt". Add a new header line that references the three files:
```
Files: junixsocket-common-2.2.0.jar, junixsocket-core-2.2.0.jar, junixsocket-native-common-2.2.0.jar
```

What is more, the ALv2 requires to carry the NOTICE if the source package has one. In this case the original author failed to include it in his/her binaries, but we are good citizens of the ecosystem.

Here is the notice file:
https://github.com/kohlschutter/junixsocket/blob/master/NOTICE

Please add a file "junixsocket-notice.txt", with the contents of the above referenced file. The contents of that file will be picked up by the build system and integrated into the final notice file of NetBeans.

ide/docker.api/src/org/netbeans/modules/docker/api/DockerSupport.java
matthiasblaesing: Is there a reason to doubt, that this will work on Mac OS X? If not I would enable it there also (I'm assuming, that docker on Mac OS X also uses unix domain sockets).

Ihsahn: addressed (afaik MacOs does allow connection via unix domain sockets)

ide/libs.c.kohlschutter.junixsocket/external/junixsocket-common-2.2.0-license.txt
Ihsahn: removed

ide/libs.c.kohlschutter.junixsocket/external/junixsocket-common-2.2.0-license.txt
Ihsahn: licenses/notice changed/updated

Ihsahn:
```
ide/libs.c.kohlschutter.junixsocket/external/junixsocket-license.txt does not contain the version 2.2.0 in its name
```
and if I add version to this, than it states, that there are 3 files and only one license, and so on...

matthiasblaesing:
This looks reasonably sane. Your test-failure is explainable, as I gave you the wrong filenames, sorry about that. Please move the license and notice to `junixsocket-2.2.0-license.txt`  and `junixsocket-2.2.0-notice.txt`. I'd also make the junixsocket library friend-only for now. That way we know which modules to check when we update the library. This can be done in the GUI (Project Properties -> API Versionioning -> `Export Packages only to Friends`. Or directly in the project.xml:

```xml
<!-- ... -->
            <!-- Remove public-packages element -->
            <friend-packages>
                <friend>org.netbeans.modules.docker.api</friend>
                <package>org.newsclub.net.unix</package>
            </friend-packages>
<!-- ... -->
```

Ihsahn:
> This looks reasonably sane. Your test-failure is explainable, as I gave you the wrong filenames, sorry about that. Please move the license and notice to `junixsocket-2.2.0-license.txt` and `junixsocket-2.2.0-notice.txt`. I'd also make the junixsocket library friend-only for now. That way we know which modules to check when we update the library. This can be done in the GUI (Project Properties -> API Versionioning -> `Export Packages only to Friends`. Or directly in the project.xml:
> 

Done.

neilcsmith-net:
> I'd also make the junixsocket library friend-only for now. That way we know which modules to check when we update the library.

-0  That doesn't seem the best reason to do that?!  I'm not a fan of the friend concept in general, particularly with third-party libraries.  Anything adding in native libraries that way can be a real pain for platform / plugin developers.

matthiasblaesing:
Ok - I take your point Neil. To no play request ping-pong, I rebased the changeset onto master, reopened the library module and merged it to master (https://github.com/apache/netbeans/commit/0d3a5448b906c339e54323d972824d6dd36baab2).

Thank you for your contribution Bartosz.

PR #1243, opened 2019-05-10
Ensuring important DTDs are used locally when running java.completion…

lkishalmi:
This is just a test if this cherry-pick makes the release110 branch compilable again.
Do not merge, yet.


matthiasblaesing:
@lkishalmi is this still relevant? it seems to be stale, if so it would be good to remove.

lkishalmi:
Not relevant any more. Only test, so it is closed.

PR #1245, opened 2019-05-11
[NETBEANS-2455] splash-about screen generation

vieiro:
Splash and about images are now created at build time, and imprinted with ${buildnum} (or a modification of that).


vieiro:
Previous about dialog
netbeans-about-dialog-previous

New about dialog
netbeans-about-dialog

geertjanw:
So awesome.

vieiro:
NOTE: As you know the splash screen is cached in $HOME/.cache/netbeans, so it's better to run tests with a clean cache directory.

junichi11:
Cool :)

platform/core.startup/build.xml
```diff
@@ -21,4 +21,14 @@
 -->
 <project name="platform/core.startup" default="build" basedir=".">
     <import file="../../nbbuild/templates/projectized.xml"/>
+    <target name="build-init" depends="projectized.build-init">
+      <taskdef name="splash-screen" classname="org.netbeans.nbbuild.assets.GenerateSplashScreenTask" classpath="${nbantext.jar}" />
+      <echo message="Generating splash screens with build number ${buildnum}" />
+      <splash-screen 
+        font='../../nbbuild/external/montserrat-medium.ttf' 
+        output='src/org/netbeans/core/startup/splash.gif'>${buildnum}</splash-screen>
```
jlahoda: I wonder - is there a strong reason to generate the files into the source directories? Wouldn't it work to generate them to '${build.classes.dir}' (may require <mkdir dir="${build.classes.dir}"> as well)? That would keep the sources clean, and would avoid the need to tweak .gitignore.

nbbuild/antsrc/org/netbeans/nbbuild/assets/package.html
```diff
@@ -0,0 +1,31 @@
...
+- Apache NetBeans Platform splash image
+- Apache NetBeans IDE splash image
+- Apache NetBeans About image
+See <a href="https://issues.apache.org/jira/browse/NETBEANS-2455">NETBEANS-2455</a>.
+
+This directory contains also the following assets:
+
+- apache-netbeans-splash.svg - InkScape compatible SVG file.
+- apache-netbeans-splash.png - Template PNG file for screen image, with size 640x396
+- apache-netbeans-about.png - Template PNG file for about image, with size 529x327
```
jlahoda: I wonder if it would make sense to have an example/command line on how the SVG should be converted to the PNG here?

vieiro:
+1 Excellent! Generating in the ${build.classes.dir} is the way to go!
There's an alternate solution by @eirikbakke , so let's close this one and discuss his first.

PR #1246, opened 2019-05-11
[NETBEANS-378,NETBEANS-2455] HiDPI splash screen with product version

eirikbakke:
These commits (1) add the ability for the NetBeans splash screen to be automatically imprinted with the NetBeans product version and build number (2) makes the splash screen and "About" box look good on HiDPI displays, and (3) adds new simplified splash screen graphics.

(We could also ask David Schulz if he'd like to make an updated version of the old splash screen, making space for the version string. In that case the graphics uploaded here just serves as an example.)

This resolves https://issues.apache.org/jira/browse/NETBEANS-2455 and https://issues.apache.org/jira/browse/NETBEANS-378 . See attached screenshot.

Splash Screen HiDPI and Product Version

Old splash screen showing HiDPI scaling artifacts on Windows 10 at 150% DPI scaling:
old splash

Tested on Windows 10; it would be great if someone else can test it on MacOS (with a Retina screen) and Linux as well.


AlexFalappa:
On which JDKs these modifications are best supported?
Does this also show up nicely on JDK 8 (Oracle and AdoptOpen implementations)?

eirikbakke:
On Windows, HiDPI screens are supported from JDK 9 and upwards; on MacOS it might have been introduced earlier (not sure).

On Java versions that do not support HiDPI screens, the splash screen should show up in low resolution, like before. I have confirmed this on JDK8 on Windows (I don't have a Mac to test).

The way this works is that the splash screen PNG is now double resolution, but resized to 50% (to SPLASH_WIDTH*SPLASH_HEIGHT logical pixels) via a transform on the paint method's Graphics2D object. JDK versions that support HiDPI will in turn apply its own transform to scale the image back up to a higher device pixel resolution.

AlexFalappa:
@eirikbakke thanks for the details.

You answered to my concerns, that is your implementation also degrades gracefully on older JDKs.

eirikbakke:
The changes to ImageUtilities are now already applied as part of another PR ( https://github.com/apache/netbeans/pull/1273 ). Rebasing to exclude the relevant changes from this PR.

eirikbakke:
Adding API change label since this PR adds new properties that can be added to the branding file to set the location of the version string on the splash screen.

neilcsmith-net:
Added NB11.1 label.  Would love to get this merged this week so I don't have to faff around with splash screens during the release process! :smile:

eirikbakke:
Added a comment on https://issues.apache.org/jira/browse/NETBEANS-2455 to see if there's consensus on this approach.

GeertjanWielenga:
Just merge it unless there’d be any reason to disagree with this.

eirikbakke:
Will merge Friday, before feature freeze... just wanted to give David Schulz a chance to provide updated splash screen artwork, so I don't "overwrite" his work.

geertjanw:
Awesome, let's get it in for 11.1. :-)

geertjanw:
@eirikbakke, I leave this to you to merge, OK?

eirikbakke:
Done. If we get new splash screen artwork but it's after the feature freeze it can just go in the following version.

neilcsmith-net:
@eirikbakke great! :+1: Personally, I don't think there's a problem with merging new artwork after feature freeze if people want it.  At the same time, I prefer your artwork anyway - it's cleaner and more legible.

One possible enhancement for this might be to also consider screen resolution in deciding splash size?  I have a large 2560x1440 display, but with DPI scaling set to 1.  Still be nice to have larger splash on such a setup.

eirikbakke:
> One possible enhancement for this might be to also consider screen resolution in deciding splash size? I have a large 2560x1440 display, but with DPI scaling set to 1. Still be nice to have larger splash on such a setup.

That would be a simple change to make. Feel free to open a JIRA issue for it. It should be done after the PR at https://github.com/apache/netbeans/pull/1278 gets merged, because the latter can make all the caching logic in ScaledBitmapIcon.java go into a superclass.

PR #1247, opened 2019-05-12
Inline parameter name hints for Java

jlahoda:
For method calls, parameter names can be injected (in a virtual text) directly into the editor. Disabled by default, as this probably needs more experience tweaking. Enable using View/Inline Hints.


jlahoda:
> The feature has to be on by default, in my opinion.

Even though I agree in principle, I don't think the feature is in a state good enough to be enabled by default now. So either we wait with merging this until it is good enough, or we merge it as disabled by default, and flip the enabled/disabled by default switch when the time comes for that.

AlexFalappa:
Is it possible to have a small screenshot of this new editor feature?

wumpz:
Is this features included within the next release of netbeans?

jlahoda:
Screenshot:
inline-names

ide/editor.lib2/apichanges.xml
```diff
@@ -83,6 +83,21 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="PrependedTextOpt">
+            <summary>Prepended text for highlights</summary>
+            <version major="2" minor="24"/>
+            <date day="14" month="5" year="2019"/>
+            <author login="markiewb"/>
```
junichi11: Maybe, it seems that you forget to change the author.

ide/editor.lib2/src/org/netbeans/modules/editor/lib2/highlighting/HighlightsList.java
```diff
@@ -157,12 +159,14 @@ public AttributeSet cutSameFont(Font defaultFont, int maxEndOffset, int wsEndOff
 
         // Extends beyond first highlight
         Font firstFont = ViewUtils.getFont(firstAttrs, defaultFont);
+        Object firstPrependText = usePrependText && firstAttrs != null ? firstAttrs.getAttribute("virtual-text-prepend") : null;
```
junichi11: Is `// NOI18N` missing?
Should introduce the constant somewhere? But I could be wrong.

neilcsmith-net:
@jlahoda fantastic! :smile: I started looking at what was required to implement this myself a couple of months ago but got waylaid.  So, what are your thoughts on merging as off-by-default for NB11.1 to get some testing in the wild, with a view to on-by-default in NB12.x if not before?

svenreimers:
Agreed, off by default for 11.1 if possible and on by default for 12?

AlexFalappa:
I very much agree on having this in NB11.1, even off-by-default.

Parameter names and in general in-line dynamic editor annotations/helpers are hype in competitor IDEs. This would allow to show that NB is not lagging behind.

geertjanw:
Or, it seems to be off by default already? If so, let's merge this.

geertjanw:
OK, then. Merging.

Chris2011:
Wow, this is exactly what we need for each language. I wanted to try to achieve that, but I didn't know how to create virtual text or readonly text inside the editor. Is there an public API/documentation for creating such virtual text to makes it easier @jlahoda?

jlahoda:
@Chris2011, yes, there is an API. A short description is in the apichanges:
https://bits.netbeans.org/dev/javadoc/org-netbeans-modules-editor-lib2/apichanges.html#PrependedTextOpt

Basically, create a highlight that will have "virtual-text-prepend" as the highlight key, value is the virtual text. This is the place where it is used in Java, which may be useful as well:
https://github.com/apache/netbeans/blob/0c67f21385ea5ddc0cf29be5d959ae2897c47f49/java/java.editor/src/org/netbeans/modules/java/editor/semantic/SemanticHighlighter.java#L75

graben:
It seems this patch only affect literals? Shouldn't it fire for every kind of parameter? Maybe configurable for performance reasons? :)

sharpedavid:
> 
> 
> It seems this patch only affect literals? Shouldn't it fire for every kind of parameter? Maybe configurable for performance reasons? :)

I bet the thought is that if you're passing in an actual variable, you should just a good variable name. Otherwise you have lots of instances of:

`firstName.substring(beginIndex: beginIndex, endIndex: endIndex);`

neilcsmith-net:
@graben was trying to find time to look at enhancing similarly - been mentioned various times by users.  While I somewhat understand @sharpedavid comment, maybe there's a way to make it configurable?  Also would be great to cover parameters that are method return values.

One area that would be good to drop duplication though is on varargs.  Wonder how easy to just label the first?

graben:
Just copy the code (DRAFT :)) from visitliteral to visitidentifier does make the feature work for variables, too. I do not see any performance issues at the moment.

graben:
> > It seems this patch only affect literals? Shouldn't it fire for every kind of parameter? Maybe configurable for performance reasons? :)
> 
> I bet the thought is that if you're passing in an actual variable, you should just a good variable name. Otherwise you have lots of instances of:
> 
> `firstName.substring(beginIndex: beginIndex, endIndex: endIndex);`

It's just a simple if around the put of the hint :)

Chris2011:
I also had some troubles, that not always shows the inline hint. Will show an example later by reprodcution.

robinsonrk:
> Edit: Just to reword, I'm saying that if your variable names are good, these hints would add a lot of screen clutter. I am still interested in this feature, but I am thinking it has some interesting unintended consequences.
> 
> ```
> // Example 1: Clutter
> Car car = factory.buildCar(make: carMake, model: carModel);
> car.drive(speed: fastSpeed, destination: nearestArbys)
> 
> // Example 2: Cleaner
> Car car = factory.buildCar(carMake, carModel);
> car.drive(fastSpeed, nearestArbys);
> ```
> 
> Please don't get too caught up deciding if you prefer example 1 or 2, my point is that it's probably a preference.

Is there an option to configure the editor behavior in this case?

Chris2011:
There is just activating/deactivating afaik.

eirikbakke:
Just discovered this feature... very cool! It will be very useful during code review workflows.

PR #1251, opened 2019-05-14
Clean modules/licenses

matthiasblaesing:
The intention of this PR is to cleanup the licenses and modules in the netbeans
codebase. It was noticed, that `ant verify-all-libs-and-licenses` shows many problems
and that the problematic modules are distributed through the whole codebase.

This PR cleans up several license problems and moves plugins, that were removed
from the build into a contrib folder.

Please see the individual commits for details.


matthiasblaesing:
Hearing no objects, merging this.

PR #1260, opened 2019-05-22
Adds SHA-512 to external binaries

emilianbold:
The download task now supports lines with the format

SHA-512 artifact-name hash-value

following the original SHA1 hash line.

Right now the file name still needs the SHA1 value, so we keep it. In the future we could modify the task to fail the build if the SHA-512 is missing.


emilianbold:
I like the current format but it probably needs some README.

My changes bring it closed to a distinfo file from pkgsrc. See https://wiki.netbsd.org/pkgsrc/intro_to_packaging/ where the distinfo file looks like:

```
SHA1 (osxinfo/de74b8960f27844f7b264697d124411f81a1eab6.tar.gz) = 83a2838ad95ff73255bea7f496a8cc9aaa4e17ca
RMD160 (osxinfo/de74b8960f27844f7b264697d124411f81a1eab6.tar.gz) = 9102eb2a938be38c4adf8cfbf781c04d0844d09a
Size (osxinfo/de74b8960f27844f7b264697d124411f81a1eab6.tar.gz) = 5981 bytes
```

So if we use

```
sha-1 filename hashvalue
sha-256 filename hashvalue
```

it's close enough.

Note that I'm still supporting the old format which basically is

```
sha1-hash filename
```

but I didn't want to replace all those lines with the new format `sha-1 filename hashvalue` as it would have made the patch really hard to verify...

So, I don't really want to do the extra work for an XML. I don't believe there even is a standard we could follow to say we just copy the DTD... These files also quite rarely change.

matthiasblaesing:
I can live with the format.

vieiro:
pkgsrc also adds the file length. That's clever. I imagine that's to protect against the length extension attack https://en.wikipedia.org/wiki/Length_extension_attack. Maybe it's a good idea to add it as well, as it's cheap.

matthiasblaesing:
I would not add even more information to the file. The structure is difficult as it is. What is more, the referenced problem applies to MACs and not basic hashing. 

The length extensions attack works by given valid_hash1, that was constructed as:
hash(secret + m1) => valid_hash1
you can create valid_hash2 for a message m1 + m2 if you only know the length of m1 and valid_hash1.
hash(secret + m1 + m2) => valid_hash2
The target system receives valid_hash2 and the combined message m1+m2. This assumes, that valid_hash2 can only be constructed, if you know "secret", but this is not the case.

But for our case, you want:
hash(m1) => valid_hash1
hash(m2) => valid_hash1
That is, an attacker creates a jar (m2), that produces the same hash as our jar (m1). The whole TLS security is based on the assumption, that hash-collisions are hard. For SHA-1 it was shown, that hash collisions were doable, for SHA-256 to my knowledge not yet and SHA-512 should keep us save for a few yers.

emilianbold:
We can add the size check later on. SHA-512 is really strong compared to SHA-1. I think we'll be good.

vieiro:
SHA-512 is strong enough so we're good. +1 to not overcomplicate things.

The idea of the length check was to make it more difficult to have two different with the same SHA-1 _and_ the same length.

emilianbold:
I would prefer to merge this before we add some new libs and I have to re-add the SHA512 values.

emilianbold:
I think we can drop this PR, correct? Don't understand why it wasn't merged for 11.1...

neilcsmith-net:
@emilianbold by dropped you mean what / why?  It didn't get merged sufficiently before the 11.1 feature freeze for some reason, and it wasn't something I wanted to merge after.  But it sounds a good idea to me.  If we do keep it, fixed and merged sooner rather than later would be good!  Maybe bring up on the mailing list?

emilianbold:
By drop I mean abandon.

BTW, it's quite frustrating that as I committer I made a PR then nicely waited in line just so my PR does not get merged for the next release so I have to re-do it since it also has conflicts now...

neilcsmith-net:
I'm just explaining why I put the NB11.2 label on when it wasn't merged as we closed master. I don't know why it wasn't merged in time, nor why you didn't push for further review / merge yourself ahead of freeze.

At this point sounds like something more useful to fix than abandon?

emilianbold:
Wasn't blaming you. It happened for reasons, can still be frustrating.

neilcsmith-net:
As this hasn't been updated for NB 11.2 and is likely to drift further away from being mergeable I suggest we close this PR, and open another one as and when it's ready to be merged?

emilianbold:
Would be an interesting proof of concept to build a JAR with a colliding
SHA1 using the https://shattered.io paper and show that if one were to
control the network / proxy one could basically inject executable code into
a fresh NetBeans build.

Of course, it cost Google $100,000 beginning of year to find a collision
but perhaps in a few years it could become cheap enough to run on a GTX.

--emi

PR #1261, opened 2019-05-22
[NETBEANS-39] Fix for no code completion in Jade files while typing

peedeeboy:
*What this PR does*
Enables auto-completion pop-up when typing in a Jade template file

*Files Changed*
* webcommon/javascript2.jade/src/org/netbeans/modules/javascript2/jade/editor/JadeCodeCompletion.java
getAutoQuery() method updated to check that typed text is not null, is greater than length of 0 and does not have whitespace at the end (because the dialog popping up after a space, tab or newline would be annoying).  If these criteria are met, code completion dialog will pop up

* webcommon/javascript2.jade/test/unit/src/org/netbeans/modules/javascript2/jade/editor/JadeAutoCompletionTest.java
Tests that the code completion pop-up is only activated if the correct criteria are met (i.e. one or more characters are typed, and the final character isn't a whitespace character)


geertjanw:
Excellent, thanks a lot, looks good, merging it, keep them coming. :-)

PR #1262, opened 2019-05-24
Support for Debugging single source java files (JEP-330)

sarveshkesharwani:
This feature allows the user to debug any Java File which is not a part of any project.
Although it mentions JEP-330, this feature works fine for JDK versions lower than 11 too.

### DIFF
--- a/java/api.debugger.jpda/src/org/netbeans/api/debugger/jpda/LaunchingDICookie.java
+++ b/java/api.debugger.jpda/src/org/netbeans/api/debugger/jpda/LaunchingDICookie.java
@@ -124,11 +124,11 @@ public final class LaunchingDICookie extends AbstractDICookie {
         // TODO: This code is likely wrong, we need to run debuggee on JDK that
         //       is set on the project.
         // XXX: This method is likely not called from anywhere.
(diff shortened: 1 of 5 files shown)


geertjanw:
Looks great to me, thanks for this enhancement to your already great support for JEP 330.

jlahoda:
Not sure how exactly is this supposed to work, but I would expect this to work simply as:
-enhance the existing ActionProvider that implements Run, to also start the debugger in the listening mode, and then perform run with the additional parameter that attaches to the debugger. Besides the code to start the debugger, this should only be a few lines of code. Making another copy of JPDAStart is not nice, but opening a random package in a random module is much worse.

Of course, it might be that that the existing ActionProvider is not in a perfect module to touch debugger, but that IMO means it needs a better module, rather than modifying debugger.jpda stashing another ActionProvider into the OpenJDK project.

matthiasblaesing:
@sarveshkesharwani is this still relevant?

sarveshkesharwani:
@matthiasblaesing sorry for the late reply (had a busy month). I do intend to work on this feature soon. However, I think it would be better to start afresh, so, I'll close this PR and create a new one as soon as I'm done.

PR #1268, opened 2019-05-28
[NETBEANS-2609]:Provided support to build NB IDE specific to Java cluster

arusinha:


nbbuild/cluster.properties
```diff
@@ -97,6 +97,10 @@ clusters.config.java.list=\
         nb.cluster.3rdparty,\
         nb.cluster.nb
 
+clusters.config.standard-java.list=\
+        ${clusters.config.java.list},\
+        nb.cluster.apisupport
```
arusinha: nb.cluster.apisupport is required without which import of NB modules were failing with below error:

unrecognized project; missing plug-in?

JaroslavTulach: The support for NB modules in IDE is part of `apisupport` cluster. There already is a cluster config for that:
```
clusters.config.basic.list=
```
I suggest to use that one instead of `java` or `standard-java`.

java/build.xml
```diff
@@ -0,0 +1,45 @@
...
+      <ant dir="." target="build" inheritAll="false">
+          <property name="cluster.config" value="standard-java"/>
+          <property name="permit.jdk9.builds" value="true"/>
+      </ant>
+  </target>
+
+ <target name="test" description="Test the NetBeans Java modules.">
+      <mkdir dir="${netbeans.dest.dir}"/>
+      <ant dir="." target="test" inheritAll="false">
+          <property name="cluster.config" value="java"/>
```
JaroslavTulach: The value should be consistent with the value on line 29.

nbbuild/cluster.properties
arusinha: had used now basic cluster for building

java/build.xml
arusinha: Had added new axis for basic cluster in .travis.yml. Already there are entries in .travis.yml to run test cases for few java modules, so I haven't added any new entry for running test-cases in .travis.yml file

nbbuild/build.xml
```diff
@@ -470,6 +470,24 @@ Hg ID:    ${hg.id}
     </nbmerge>
   </target>
 
+  <target name="build-basic" description="Build the NetBeans Basic cluster.">
```
JaroslavTulach: Is there any reason for the `target`s to be in `nbbuild/build.xml`? I preferred when they were only in `java/build.xml`.

arusinha: To be able to call target "build-basic" from .travis.yml.
Not sure travis will accept something like  OPTS="-f java/build.xml"

arusinha: hi @JaroslavTulach , Can i merge the PR. there is an unresolved conversation?

PR #1034, opened 2018-11-23
NETBEANS-1499 - change package.json icon

Chris2011:
This PR will change the generic js icon for a specific file (package.json) with a specific mimetype (text/package+x-json) to the logo of npm, to make it more prominent.

Before merging, we have to check the licenses of the npm logo: https://www.npmjs.com/policies/trademark. I wrote them an e-mail, will let you know whether it is ok or not. Maybe a license guy can help here too.

This logo is also used inside of a VS Code plugin for better icons for specific filetypes.

PR #1269, opened 2019-05-29
[NETBEANS-2417] Update Gradle logo/icon

louhy:
Also may address NETBEANS-2608.


louhy:
Actually, I've tried out a lighter theme and I agree. The darker ones will probably look better on white. I'll prepare a second update.

louhy:
See how you like this one... at first I thought maybe the contrast was a little too much, but maybe it's okay after looking at it a bit more. Both versions will be in the repo, it can always be changed to the medium-blue if it's unpopular. Unfortunately their branding rules are very strict about color tweaking (i.e. none allowed at all).

matthiasblaesing:
> Unfortunately their branding rules are very strict about color tweaking (i.e. none allowed at all).

This raises the question: The logos, where do they come from and how are they licensed? For most of our use cases trademark law does not apply. Copyright does. Here is the apache legal faq, that list the acceptable licenses.

https://apache.org/legal/resolved.html

Also possible related the legal issue that was opened: 
https://issues.apache.org/jira/browse/LEGAL-353
and the TL;DR: https://github.com/apache/netbeans/pull/1034#issuecomment-462447508

When adding/changing graphics, please update the `licenseinfo.xml` file if necessary, that way it is tracked what license applies.

louhy:
IANAL so I'll just go with the flow here on that stuff. 

Icons are from the [Gradle branding site](https://gradle.com/brand/) referenced in the Jira ticket NETBEANS-2417, where a member of the Gradle team (I believe) requested the change. 
They have their own set of guidelines listed there. (I did do everything I could to ensure these were honored: no color changes, "distortions", rotations, etc.)

Given all that, I'm not sure what updates (if any) should be done on licenseinfo.xml. 
(groovy/gradle/licenseinfo.xml and groovy/gradle.java/licenseinfo.xml)

matthiasblaesing:
@louhy I respect what you did here. The problem is the base you worked on. The file have no license and as such are not usable. I put some more info to the issue, maybe the files can be provided with an appropriate license.

louhy:
Also replied on the issue, I guess we need help from Gradle folks.

aweida:
Hi all, Charles here from Gradle. Would it be possible to share a screenshot(s) of the icons in context so us design folks who are less technically inclined can take a look? 

Thank you

louhy:
@aweida Does this help? Projects and files have special icons which indicate project or file type: 

gradle

lkishalmi:
Closing this one in favor of #1557

PR #1270, opened 2019-05-29
[NETBEANS-2612] WIP apidoc frameless and branding agnostic

ebarboni:
This PR encompass several changes (currently putting effort on layout of web page)

### Remove branding properties
remove of properties in *nbbuild/build.properties*:
*config.javadoc.all,previous.release.year,previous.release.month,previous.release.day*
### Setup build from new parameter **build.branch**
**SetApidocClustersConfig.java** created to reconfigure the *config.javadoc.all* according to **build.branch**  
**ReleaseJsonProperties.java** created to retrieve some buildinfo including previous release from a
it also transorm in an xml file to have all history.
json file hosted here https://gitbox.apache.org/repos/asf?p=netbeans-jenkins-lib.git;a=blob;f=meta/netbeansrelease.json;
(this json file will also help @vieiro for building website)
### rework of the apidoc html
alter build of javadoc to remove time as set by **ReleaseJsonProperties.java**
alter build to have another template during work in progress.
use l&f of Apache NetBeans website, scrollabe menu to have all history
embedd arch + apichanges to have jdk like menu (try to help navigability)
### todo
nice looking, all classes display, jquery or script addition to ease searching filtering


ebarboni:
Hi @vieiro, and @neilcsmith-net (sorry I wouldn't distract you from  release process 11.1). This incomplete PR try to remove the apidoc reconfiguration of branding on branch cut but it may be extended for general branding.
I would know if it make sense and if there may be some missing branding we can put into json file https://github.com/apache/netbeans-jenkins-lib/blob/master/meta/netbeansrelease.json

vieiro:
Hi @ebarboni , I'm afraid I don't currently have time to go through this complex setup. I can't tell if "remove of properties in nbbuild/build.properties:" has implications in other parts of the NetBeans build infrastructure. Will this break project builds?

Currently the javadoc publishing is being done by adding entries by hand in a script file in netbeans-vm.apache.org. AFAICT the information in the JSON file seems enough to try to make this more automatic, but I can't currently tackle this. You can find the script file at @netbeans-vm.apache.org/usr/share/netbeans-javadoc.

ebarboni:
@vieiro, for now it did not hurt the other build target because it regenerate the previous properties from the json.
I also mainly try to do list of version somehow automatic also to avoid editing xsl on release. 

I will try to setup a webserver to get it live.

By the way I'm not sure that only javadoc date and javadoc category are the element to configure per release.

neilcsmith-net:
> By the way I'm not sure that only javadoc date and javadoc category are the element to configure per release.

@ebarboni not exactly sure what you mean by this question - maybe this is a helpful references? https://github.com/apache/netbeans/pull/1304

I think we should move to be able to configure all of that, not just the Javadoc, from a single configuration file for NB 11.2

ebarboni:
hi @neilcsmith-net and thanks ? it perfectly fit what I was asking. (I ad a limited javadoc centric view)
Some more work to do on the bundle.

I guess I will do as fast as I could maybe 11.3 (if too complex) but it's ok to do like we were doing for the first apidoc rewrite to be (incubating compliant) and commit change to all branch event after release. And hopefully we can branch future release without thinking on this anymore.

ebarboni:
Hi @vieiro @sdedic @JaroslavTulach, @geertjanw ,@neilcsmith-net 
As I can use the netbeans-vm, I created a live demo here, http://bits.netbeans.org/apidoctest/
I would appreciate review for the website, to see if this is usable, (not loosing info from previous apidoc system)
I try to unframe (because if we switch to jdk12 in the future frame are no more present) , add search.
 
As json contains all the Apache NetBeans Version it's easy to create a navbar to locate the current version of apidoc in the versions timeline. The data are taken from my local master branch, but I force a build using release90 data of the json to see if it works. (apichanges are not 9.0, but release date should match)
I remove the old oracle javadoc because we may have 404 issue and put a link to the bits.netbeans.org where I will later put the list here. (at least the last 8.2)

This build system will work on anybranch and we never have to care on apidoc date, apidoc stability.
This may ease a bit the release management, but I have no idea on how to take care of the BUNDLE file pointed in #1304 by @neilcsmith-net

neilcsmith-net:
Interesting!  Need to find some time to look more.

> This may ease a bit the release management, but I have no idea on how to take care of the BUNDLE file pointed in #1304 by @neilcsmith-net

With regard to this, I had some thoughts on templating via the build to have it all in one place.  We need version (for userdir, update centres, etc.) and optional version suffix (eg. -beta1) for use in display areas (eg. window, About).  Potentially that can be taken from the same place you're setting it?

JaroslavTulach:
The [demo](http://bits.netbeans.org/apidoctest/) is quite slick. I admire your courage to dig deep into all the `.xsl` transformations I have once written (I prefer to stay away). Looks like the integration with `arch.xml`, `apichanges.xml` and `@codesnippet` doclet still works (that makes me happy). Good job.

ebarboni:
@JaroslavTulach xsl were clear.

@neilcsmith-net data can be seen here: 
https://gitbox.apache.org/repos/asf?p=netbeans-jenkins-lib.git;a=blob;f=meta/netbeansrelease.json
Build is done according to branch.
(json is missing the extra term version like feature update 1 and LTS terms at least for website and but can be envisioned for about and welcome)

Rereading 10times but still complex sorry :p :
Assuming we  may produce the version json is set with/
0) json updated with data
1) releaseAAA branch is created on git, .
   We add an optional version -beta1 on releaseAAA with LTS as extra version term. 
   It will generate a releaseAAA-LTS-beta1 with everything in about,welcome....  
   But as beta1 is set in source+installers, once RM take and sign for a vote, it would be a vote for releaseAAALTS-beta1 isn't? 
   We have to wait until we set option-version to empty string to vote on releaseAAA-LST, assuming we consider out of beta.

-- Aside that, on the voting candidate if we want to releaseAAALTS we can put it on a distinctive folder and not renaming per Vote the sources zip
dist/dev/netbeans/AAA-VC1/releaseAAALTS-sources.zip

---
For the production, 
For apidoc (assuming we merge the pr)  the version read from json will be set in temp build files, easy move no source modification. 
For NetBeans bundle file,ant will modify a Bundle file on source repo, git will say uncommited change. We need a gentle way to do this modification.
(No sure a good idea, alter the build of concerned module to preprocess bundle file before compilation).

neilcsmith-net:
@ebarboni thanks!  Wondering whether some of this might be easier on email than here - maybe separate addition to this PR anyway?  But ...

> data can be seen here:
https://gitbox.apache.org/repos/asf?p=netbeans-jenkins-lib.git;a=blob;f=meta/netbeansrelease.json

On build / bundle parameters I think we need to find a way to put that info in the NetBeans sources?  It should be possible to build from source and get the right artefacts.  That might apply to other things in that JSON too?

> Aside that, on the voting candidate if we want to releaseAAALTS we can put it on a distinctive folder and not renaming per Vote the sources zip

Voting candidates need to be manually signed and renamed at present - that doesn't seem much of an issue.  File renaming is the easy part.  Ideally all the betas for NB 11.1 would have been labelled in the UI.

> For NetBeans bundle file,ant will modify a Bundle file on source repo, git will say uncommited change. We need a gentle way to do this modification.
(No sure a good idea, alter the build of concerned module to preprocess bundle file before compilation).

Why?  Doing find and replace on the file as it's copied across in the build seems the right approach here?  It involves adding a few lines to the build scripts of a couple of modules inside `/nb`.

ebarboni:
@neilcsmith-net:

ReleaseJsonProperties ant do the job of getting json info for a branch (default master) apidoc 'can be generalised to get more infos'. It transform json into ant properties + xml file. if no json available during build it's a bit like not having a dependencies. (Break build).
The current netbeans-TLP has to download json itself because it's old build.

a bit of rationale :
We would like to have no "branding" commit on branch cut.
With @vieiro we would like to use the json info for our INFRA, building the apidoc folder, the netbeans website meta dl page. Should not only be in the netbeans source because on every release we would have to recommit the new information on all previous branches. 
I personaly want the apidoc to know the "future version" at anytime. If user bookmark 9.0 version it will then show that there are more recent version.

To sum up:
the workflow would be like that
RELEASEA-betan with betan branding (no vote)
then once acceptance ok
RELEASEA no more beta branding but artifacts rename -vcn (voting cantidate)
(more or less like we have now except the branding)

Your right mail would be better.

neilcsmith-net:
> We would like to have no "branding" commit on branch cut.

+1 to reducing the amount of branding changes required on the release branch (or master tag).  But -1 from me to not having _any_ version  / branding commits in the source tree.  IMO someone should be able to checkout the sources at any particular tag / commit and build the "correct" version of the IDE.  And that build shouldn't change over time.

What you're aiming at I agree with, just wary of moving further away from reproducible builds than towards it. :smile:

ebarboni:
I think, we are not good yet for reproducible build, buildnumber is an issue, jdk version, ant version, maven version should be taken into account.
jar ordering, jar building, (removing timestamp) much more :D

Have maybe a different plan but may statisfy better.

trying an example:
```
git branch master with hash 000
git branch releaseA with following hash;
git hash 123 for beta1 -- this is the good beta version
git hash 234 [fix apidoc]
git hash 345 for vc1 -- and this is the good release version
git hash 456 [fix apidoc]
```

```
json (simplified)
releaseA {name:releaseA
          statut:released
          milestone[orabettername] {
          123:{datas:beta1,t:beta,p:1},456:{datas:vc1,t:vc,p:2}
          }}
```


I'm not very convinced yet on keeping branding into codebase. Because for me it's only the git hash that make the version. 
Dream but not too far from reality:
I imagine a future build system that use git within ant task (there are still mercurial related task in the antsrc folder, maybe having for git is not big deal) and builnum to be git hash (not jenkins buildnumber).
I imagine a json with only branding for "milestone". 
I imagine the source generating artefacts task to add at root a file release.properties containing (scmhash=,scmbranch=).

User with git
Cloning releaseA
1. try to build,  
2.   git say hash 456 branch releaseA, 
3.   get json info for releaseA hash456 -> no hash info, use releaseA-dev branding

Cloning hash 345
1. try to build,  
2.   git say hash 345 branch releaseA, 
3.   get json info for releaseA hash456 ->info found, 
4.   use releaseA branding (as t==vc and statut==released and last p:position in the list of t==vc)

Cloning hash 123
1. try to build,  
2.   git say hash 123 branch releaseA, 
3.   get json info for releaseA hash456 ->info found, 
4.   use releaseA-beta1(datas) branding (as t!=vc)

Cloning on the 234, 456 hash
1.  same as cloning releaseA, user is between milestone, is on releaseA-dev

Cloning master
1.   git say hash 000 branch master,
2.   get json info for master hash 000 -> no info, use master-dev branding or dev directly 

User downloading sources, not cloning (
1.   git not found,fall back to release.properties informations
2.   no release properties fall back to master-dev or dev directly 

User downloading sources zip from apache distribution (with release.properties)
1. try to build
2.   git not found,fall back to release.properties informations
3.   get json info and build accordingly

neilcsmith-net:
> I think, we are not good yet for reproducible build, buildnumber is an issue, jdk version, ant version, maven version should be taken into account.

Yes, we're a long way from it!  But it seems to be the right direction of travel.  The build number would be top of the priorities in my opinion because of the issues that caused the re-roll of the Maven vote.  Although part of me thinks making impl. and spec. versions the same would be a good change as well.  Where else is build number used?

> I'm not very convinced yet on keeping branding into codebase. Because for me it's only the git hash that make the version.

The second sentence I totally agree with.  You could achieve same with tagged commits of the version information.  Although really that will be suffix (eg. `-beta1`) only - we'll need all other builds off the release branch to be eg. `11.1-dev+buildnumber` or templating things like the update centres won't work.  "Branding" here includes things that will have functional ramifications.  

> I imagine a future build system that use git within ant task ... and builnum to be git hash

Could even parse that directly from `.git/HEAD` ?

ebarboni:
.git/HEAD contains informations that at first I thoug it was enough. But if you git checkout a particular hash it cannot track the branchs, it's unrelated (detached - french LOCALE :/) head and it only get hase. 
only that is working in this case
`git` branch --remote --verbose --no-abbrev --contains | sed -rne 's/^[^\/]*\/([^\ ]+).*$/\1/p'`

I do not understand the update center part sorry.

neilcsmith-net:
> .git/HEAD contains informations that at first I thoug it was enough.

Fair enough - didn't look enough in depth.

> I do not understand the update center part sorry.

One of the things set in branding that needs to pick up the version number (if we want to stop committing it directly) is the URL's of the two update centres, as part of the PR you asked about earlier - eg. https://github.com/apache/netbeans/pull/1304/commits/d7b4dada46349798a1869081fe2327dd0c35d6d6

We should make that eg. `https://netbeans.apache.org/nb/plugins/dev/catalog.xml.gz` in master and handle the actual redirect elsewhere, then replace dev with release eg. `https://netbeans.apache.org/nb/plugins/11.1/catalog.xml.gz`

This needs to be set for every build of the release branch and work locally to be useful though - it can't be just on particular hashes.

ebarboni:
to make it clear to me
on master -> https://netbeans.apache.org/nb/plugins/dev/catalog.xml.gz
on release111 (any hash inside) ->https://netbeans.apache.org/nb/plugins/11.1/catalog.xml.gz

Release Manager has to put catalog manually ?

neilcsmith-net:
@ebarboni yes, currently a manual commit, and what's in master is (was?) broken.  Having just a standard `/dev/...` link was my thought on moving the existing master links somewhere we can just handle via redirects same as releases.

Full links works, and possibly better, although I was initially thinking we could replace `/dev/` with `/{version}/` ?

neilcsmith-net:
@ebarboni the commit history on this looks like it's going a bit awry now?!  Maybe need to rebase, or close and redo this PR?

PR #1271, opened 2019-05-30
[NETBEANS-2577] Prevent Maven libraries from being exluded

sdedic:
I have essentially swapped two pieces of code: the *explicit* information from NB installation (`update_tracking/*.xml`) takes precedence.

Original behaviour:
- if a file that *just happens* to be an OSGi bundle (or NB module) is present anywhere, it's filtered according to excluded modules
- filters out also `mave/lib` contents, which is not managed by NB module system, but contains libraries supporting embedded maven

New behaviour:
- the Platform defines what module the file belongs to. Maven libraries, although OSGi bundles themselves, will be attributed to maven embedder module
- if not defined, and the file contains NB module or OSGi bundle manifest, it will be recognized as a module itself

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/ModuleSelector.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/ModuleSelector.java
@@ -78,24 +78,6 @@ public final class ModuleSelector extends BaseExtendSelector {
         }
         
         String module = null;
-        if (file.getName().endsWith(".jar")) {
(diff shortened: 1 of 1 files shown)


sdedic:
@JaroslavTulach it seemed to work; I built a distro from master and branch, and the filelist was just the same. Also note that the bug affects mainly NB platform applications, not NetBeans itself.

sdedic:
just rebased this on the latest master.

PR #1273, opened 2019-05-30
[NETBEANS-2614,NETBEANS-1586] Improve icon scaling on HiDPI displays, and prepare ImageUtilities for HiDPI icons

eirikbakke:
This PR contains most of the changes that are needed to make ImageUtilities work with HiDPI icons. No publicly visible API changes are needed; the old javax.swing.Icon interface is actually well-suited for providing scalable graphics. The main challenge is to ensure that conversions between javax.swing.Icon and java.awt.Image, which are frequent throughout the NetBeans codebase, always retain the underlying scalable Icon instance.

The main motivation for this PR is to enable SVG icon support, which I will contribute in a separate PR. There's an added bonus, though: With all icon painting now centralized through ImageUtilities.ToolTipImage.paintIcon, we can add some rendering tweaks that make existing low-resolution icons look slightly better on HiDPI displays. See the attached before/after screenshots.

This issue is tracked under https://issues.apache.org/jira/browse/NETBEANS-2614 . SVG icon support is tracked under https://issues.apache.org/jira/browse/NETBEANS-2604 .

Credit: Emilian Bold did a lot of initial work on this before, at https://github.com/emilianbold/nextbeans/commit/0f99dba0c1b3e8e0bc4e7cec407b53d30e85ead1 . The main improvement in this PR is that arbitrary HiDPI scaling factors are supported. This is important on Windows, where non-integral scaling factors such as 150% can occur (MacOS, in contrast, always uses 200% for Retina screens). This PR also avoids the use of MultiResolutionImage, which does not work properly on Windows (see https://bugs.openjdk.java.net/browse/JDK-8212226 ).

To make reviewing easier, I have split up this PR into several commits, which can be reviewed one-at-a-time.

(To compare before/after screenshots, be sure to view the images at 100% scaling.)
Windows 150pct HiDPI Scaling, Before Patch
Windows 150pct HiDPI Scaling, After Patch
Windows 200pct HiDPI Scaling, Before Patch
Windows 200pct HiDPI Scaling, After Patch

### DIFF
--- a/java/jshell.support/src/org/netbeans/modules/jshell/editor/CommandCompletionProvider.java
+++ b/java/jshell.support/src/org/netbeans/modules/jshell/editor/CommandCompletionProvider.java
@@ -337,7 +337,7 @@ public class CommandCompletionProvider implements CompletionProvider{
                             n = f.getLookup().lookup(Node.class);
                         }
                         if (n != null) {
-                            return (ImageIcon)ImageUtilities.image2Icon(n.getIcon(BeanInfo.ICON_COLOR_16x16));
+                            return new ImageIcon(n.getIcon(BeanInfo.ICON_COLOR_16x16));
                         }
                     }
                 }
--- a/ide/editor.breadcrumbs/src/org/netbeans/modules/editor/breadcrumbs/BreadCrumbComponent.java
+++ b/ide/editor.breadcrumbs/src/org/netbeans/modules/editor/breadcrumbs/BreadCrumbComponent.java
@@ -64,7 +64,8 @@ import org.openide.util.WeakListeners;
  */
 public class BreadCrumbComponent<T extends JLabel&Renderer> extends JComponent implements PropertyChangeListener {
 
-    private final Image SEPARATOR = ImageUtilities.loadImage("org/netbeans/modules/editor/breadcrumbs/resources/separator.png");
+    private final Icon SEPARATOR =
+        ImageUtilities.image2Icon(ImageUtilities.loadImage("org/netbeans/modules/editor/breadcrumbs/resources/separator.png"));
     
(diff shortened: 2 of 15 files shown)


java/jshell.support/src/org/netbeans/modules/jshell/editor/CommandCompletionProvider.java
```diff
@@ -337,7 +337,7 @@ protected ImageIcon getIcon() {
                             n = f.getLookup().lookup(Node.class);
                         }
                         if (n != null) {
-                            return (ImageIcon)ImageUtilities.image2Icon(n.getIcon(BeanInfo.ICON_COLOR_16x16));
+                            return new ImageIcon(n.getIcon(BeanInfo.ICON_COLOR_16x16));
```
JaroslavTulach: Suspicious. Nobody shall instantiate icons. Code should call into `ImageUtilities` all the time, I believe.

eirikbakke: It's a case where additional work, including a public API change in CompletionUtilities.renderHtml, would be required to change method signatures from ImageIcon to Icon.

The change made here here is solely for the purpose of avoiding a ClassCastException. Once we actually start adding SVG icons we can go and migrate individual sites from ImageIcon to Icon, as necessary.

eirikbakke:
Fixed a few test failures, and added a bunch more tests. I have had this patch running in my IDE for a week; will soon merge this one (and then continue work on the separate SVG support PR).

eirikbakke:
Just tested this patch (+the SVG patch) on MacOS as well. Seem not to be causing problems. Will merge this one.

PR #1274, opened 2019-05-31
[NETBEANS-2519] Fixing automatic module name generation when the modu…

jlahoda:
…le name has trailing digits.

See:
https://issues.apache.org/jira/browse/NETBEANS-2519


geertjanw:
I see no reason not to merge, and we need this, so I am merging.

PR #1275, opened 2019-05-31
Release110

b582q9:
Every time, very slow javadoc index search when there are several sources declared. Javadoc index search can take about 1m+
Screenshot_20190531_163948


junichi11:
@b582q9 Here is not a proper place to report your issues. So, if you have issues, requests, and more, please report them to the JIRA: https://issues.apache.org/jira/projects/NETBEANS/issues

If you have questions, please join the mailing lists, then ask there: https://netbeans.apache.org/community/mailing-lists.html

Please don't use "Pull requests" as an issue tracker.

PR #1280, opened 2019-06-02
Java with Maven -> Java Frontend App + MOE is broken since Gradle support introduced

dukescript:
Try NetBeans 11 and do: New project -> Java with Maven -> Java Frontend Application -> ... -> Run on iOS devices + Use Multi OS Engine (needs Mac). Then create the project...

Nested `client-moe` project is recognized as Gradle, while it is a nested Maven project. The IDE gets totally confused by that.

This fix mitigates the situation by preferring Maven over Gradle, if parent folder is also Maven, but not Gradle. Is it OK this way, @lkishalmi?

### DIFF
--- a/groovy/gradle/src/org/netbeans/modules/gradle/NbGradleProjectFactory.java
+++ b/groovy/gradle/src/org/netbeans/modules/gradle/NbGradleProjectFactory.java
@@ -47,24 +47,31 @@ public final class NbGradleProjectFactory implements ProjectFactory2 {
 
     @Override
     public boolean isProject(FileObject dir) {
-        boolean ret = FileUtil.toFile(dir) != null;
-        if (ret) {
-            FileObject pom = dir.getFileObject("pom.xml"); //NOI18N
-            if ((pom != null) && pom.isData() && GradleSettings.getDefault().isPreferMaven()) {
-                ret = false;
-            } else {
-                File suspect = FileUtil.toFile(dir);
-                GradleFiles files = new GradleFiles(suspect);
-                if (!files.isRootProject()) {
-                    Boolean inSubDirCache = GradleProjectCache.isKnownSubProject(files.getRootDir(), suspect);
-                    ret = inSubDirCache != null ? inSubDirCache : files.isProject();
-                } else {
-                    ret = true;
-                }
+        return isProjectCheck(dir, GradleSettings.getDefault().isPreferMaven());
+    }
+
+    static boolean isProjectCheck(FileObject dir, final boolean preferMaven) {
+        if (dir == null || FileUtil.toFile(dir) == null) {
+            return false;
+        }
+        FileObject pom = dir.getFileObject("pom.xml"); //NOI18N
+        if (pom != null && pom.isData()) {
+            if (preferMaven) {
+                return false;
+            }
+            final FileObject parent = dir.getParent();
+            if (parent != null && parent.getFileObject("pom.xml") != null) { // NOI18N
+                return isProjectCheck(parent, preferMaven);
             }
         }
-
-        return ret;
+        File suspect = FileUtil.toFile(dir);
+        GradleFiles files = new GradleFiles(suspect);
+        if (!files.isRootProject()) {
+            Boolean inSubDirCache = GradleProjectCache.isKnownSubProject(files.getRootDir(), suspect);
+            return inSubDirCache != null ? inSubDirCache : files.isProject();
+        } else {
+            return true;
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


groovy/gradle/src/org/netbeans/modules/gradle/NbGradleProjectFactory.java
```diff
@@ -47,24 +47,31 @@
...
-            } else {
-                File suspect = FileUtil.toFile(dir);
-                GradleFiles files = new GradleFiles(suspect);
-                if (!files.isRootProject()) {
-                    Boolean inSubDirCache = GradleProjectCache.isKnownSubProject(files.getRootDir(), suspect);
-                    ret = inSubDirCache != null ? inSubDirCache : files.isProject();
-                } else {
-                    ret = true;
-                }
+        return isProjectCheck(dir, GradleSettings.getDefault().isPreferMaven());
```
lkishalmi: Well in this case, I'd rather like an 
if (preferMaven && mavenProjectCheck(...)) return false;
<Gradle Project check>
kind of implementation.

dukescript: What exactly do you want? I don't understand.

groovy/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectFactoryTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertFalse(NbGradleProjectFactory.isProjectCheck(null, true));
+    }
+
+    public void testPomAndGradle() throws Exception {
+        FileObject prj = root;
+        FileObject pom = FileUtil.createData(prj, "pom.xml");
+        FileObject gradle = FileUtil.createData(prj, "build.gradle");
+
+        assertTrue("Gradle wins", NbGradleProjectFactory.isProjectCheck(prj, false));
+        assertFalse("Pom wins", NbGradleProjectFactory.isProjectCheck(prj, true));
```
dukescript: This is the test for NetBeans 11 behavior. It is unchanged.

groovy/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectFactoryTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public void testPomAndGradleNested() throws Exception {
+        FileObject parentPrj = root;
+        FileObject parentPom = FileUtil.createData(parentPrj, "pom.xml");
+        FileObject prj = FileUtil.createFolder(parentPrj, "child");
+        FileObject pom = FileUtil.createData(prj, "pom.xml");
+        FileObject gradle = FileUtil.createData(prj, "build.gradle");
+
+        assertFalse("Pom wins on settings", NbGradleProjectFactory.isProjectCheck(prj, true));
+        assertFalse("Pom wins on parent pom", NbGradleProjectFactory.isProjectCheck(prj, false));
```
dukescript: This is a test for the new, changed behavior. If there is:
```
pom.xml
subdir/pom.xml
subdir/build.gradle
```
then the `subdir` is treated as Maven project.

groovy/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectFactoryTest.java
```diff
@@ -0,0 +1,90 @@
...
+        FileObject parentGradle = FileUtil.createData(parentPrj, "build.gradle");
+        FileObject prj = FileUtil.createFolder(parentPrj, "child");
+        FileObject pom = FileUtil.createData(prj, "pom.xml");
+        FileObject gradle = FileUtil.createData(prj, "build.gradle");
+
+        assertFalse("Parent Pom wins on settings", NbGradleProjectFactory.isProjectCheck(parentPrj, true));
+        assertTrue("Parent Gradle wins", NbGradleProjectFactory.isProjectCheck(parentPrj, false));
+
+        assertFalse("Pom wins on settings", NbGradleProjectFactory.isProjectCheck(prj, true));
+        assertTrue("Gradle wins on parent build.gradle", NbGradleProjectFactory.isProjectCheck(prj, false));
```
dukescript: Yet another test showing the old behavior is kept. If there is:
```
pom.xml
build.gradle
subdir/pom.xml
subdir/build.gradle
```
then Gradle still takes precedence at it used to in NetBeans 11.

PR #1281, opened 2019-06-04
[NETBEANS-2639] Compile mavensrc to Java 7, not 8

jglick:
[NETBEANS-2639](https://issues.apache.org/jira/browse/NETBEANS-2639)

The bundled Maven is [built to Java 7](https://github.com/apache/maven/blob/maven-3.3.9/pom.xml#L46-L47) so we cannot build the spy module against Java 8 as https://github.com/apache/netbeans/pull/1144 did—it will not link. Same for the [current release](https://github.com/apache/maven/blob/maven-3.6.1/pom.xml#L50-L51). Java 5 is too old for JDK 11 javac, which only supports 6 and later, so we can choose 6 or 7.

### DIFF
--- a/java/maven/build.xml
+++ b/java/maven/build.xml
@@ -30,8 +30,8 @@
         <javac srcdir="mavensrc" destdir="build/mavenclasses" 
             debug="${build.compiler.debug}" 
             deprecation="${build.compiler.deprecation}" 
-            source="${javac.source}" 
-            target="${javac.target}" 
+            source="1.7"
+            target="1.7"
             includeantruntime="false">
             <classpath>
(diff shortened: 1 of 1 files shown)


jglick:
Forgot to mention that I did test this on Ubuntu via

```sh
JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 ant -f java/maven clean run
```

verifying that

* the build succeeds with or without patch
* the Output Window for Maven is broken without patch and works with patch

matthiasblaesing:
Change is sane ant reasoning sound (I think I wrote something along the same line in the referenced PR).

PR #1283, opened 2019-06-06
[NETBEANS-2629]  Fix Error SourceGroup node on relative outside the project paths.

lkishalmi:
Removed project ownership check which can resolve
this issue, however this could cause further problems as it potentially
makes the sources "shareable" which can result wrong results in the
FileOwnerQuery.


lkishalmi:
Probably it is better not to allow such a situation. Closing it now.

PR #1285, opened 2019-06-07
[NETBEANS-2457]: Provided Error Hint to turn on Preview Features in ant/maven Java projects

arusinha:
Provided  error fix for below kind of error in Java ant/maven projects. The fix will word on any preview features added in the jdk.

switch expressions are a preview feature and are disabled by default.
(use --enable-preview to enable switch expressions)

Issue Link:
https://issues.apache.org/jira/browse/NETBEANS-2457

Test Cases were not added as currently couldn't find support to create ant/maven project through code in  Test Base classes of java.hints module


java/java.hints/nbproject/project.xml
```diff
@@ -278,6 +278,24 @@
                         <specification-version>1.25</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
```
jlahoda: Having java.hints depend on all these project modules does not seem right - if e.g. project.ant is disabled, should java.hints be disabled as well? To me, does not seem so. I'd suggest to:
-place maven handling to maven.hints (duplicate the ErrorRule there)
-place the ant handling either to a separate module; or avoid any hard dependency on project.ant, etc. here (after all, the nbproject/project.properties should be readable without the dependency).

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
```diff
@@ -0,0 +1,415 @@
...
+
+    private static final Set<String> ERROR_CODES = new HashSet<String>(Arrays.asList(
+            "compiler.err.preview.feature.disabled.plural")); // NOI18N
+    private static final String ENABLE_PREVIEW_FLAG = "--enable-preview";   // NOI18N
+    private static final String JAVAC_COMPILER_ARGS = "javac.compilerargs"; // NOI18N
+    private static final String RUN_JVMARGS = "run.jvmargs"; // NOI18N
+
+    @Override
+    public Set<String> getCodes() {
+        float jvmVersion = Float.parseFloat(System.getProperty("java.specification.version")); // NOI18N
```
jlahoda: Not clear what is the reason to check the java spec. version here? If the error is reported, then we can suggest adding the option, and ignore it othewise?

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
```diff
@@ -0,0 +1,415 @@
...
+import java.util.Map.Entry;
+import java.util.Set;
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.api.java.source.CompilationInfo;
+import org.netbeans.api.project.FileOwnerQuery;
+import org.netbeans.modules.java.hints.spi.ErrorRule;
+import org.netbeans.spi.editor.hints.ChangeInfo;
+import org.netbeans.spi.editor.hints.Fix;
+import org.netbeans.spi.project.support.ant.AntProjectHelper;
+import org.netbeans.spi.project.support.ant.EditableProperties;
```
jlahoda: FWIW, there is also org.openide.util.EditableProperties, in case it would be useful.

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
```diff
@@ -0,0 +1,415 @@
...
+    private boolean isMavenProject(Project prj) {
+
+        FileObject pom = prj.getProjectDirectory().getFileObject("pom.xml");
+        return (pom != null) && pom.isValid();
+
+    }
+
+    private boolean isAntProject(Project prj) {
+
+        FileObject buildFile = prj.getProjectDirectory().getFileObject("build.xml");
```
jlahoda: This check feels way too weak: I'd suggest to at least check for nbproject/project.xml and nbproject/project.properties in addition to build.xml. Or, (if this is in a module that depends on project.ant), it might be possible to look for AntProjectHelper in the project's lookup.

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
```diff
@@ -0,0 +1,415 @@
...
+                    if (MAVEN_COMPILER_GROUP_ID.equals(plugin.getGroupId())
+                            && MAVEN_COMPILER_ARTIFACT_ID.equals(plugin.getArtifactId())) {
+                        return plugin;
+                    }
+                }
+            }
+            return null;
+        }
+
+        private Plugin createMavenEclipseCompilerPlugin() {
```
jlahoda: Why "eclipse"?

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
arusinha: if you have nb-javac plugin installed, error will show even if you are running in jdk-8

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
arusinha: My mistake i should had corrected the function name.
The code was copied from https://github.com/apache/netbeans/blob/master/groovy/maven.groovy/src/org/netbeans/modules/maven/groovy/extender/AddMavenCompilerPlugin.java

java/java.hints/src/org/netbeans/modules/java/hints/errors/EnablePreviewFeature.java
jlahoda: Yes, but why the fix shouldn't be shown in that case? The project should have JDK 12+ configured as its platform, so the option will work? (Maybe check if CompilationInfo.getSourceVersion() == SourceVersion.lastest(), to avoid some clashes between --enable-preview and -source, but I don't see how the IDE's runtime platform could impact the decisions here.)

neilcsmith-net:
Theoretically we're past feature freeze for NB 11.1.  I'll let @jlahoda comment on whether we should allow through or retarget for NB 11.2.

geertjanw:
It would be nice to have this one in nb11.1, because it is a good supporting feature for the various preview features we have in 11.1 (and also in 11.0), i.e., it simplifies their enablement.

arusinha:
@jlahoda  I had addressed all the review comment. Request you to review

neilcsmith-net:
@jlahoda merging this for NB11.1-beta3. I believe all changes requested have been made? Please raise concerns if required before the voting candidate is built so can be addressed or reverted. Thanks.

PR #1288, opened 2019-06-09
[NETBEANS-2653] Generate toString() with StringBuilder

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-2653

As the same as current behavior, "+" operator is used by default. 
nb-gen-toString-1
nb-gen-toString-3

### DIFF
--- a/java/java.editor/src/org/netbeans/modules/java/editor/codegen/Bundle.properties
+++ b/java/java.editor/src/org/netbeans/modules/java/editor/codegen/Bundle.properties
@@ -42,6 +42,7 @@ LBL_hashcode_select=Select fields to be included in hashCode():
 LBL_tostring=toString()...
 LBL_generate_tostring=Generate toString()
 LBL_tostring_select=Select fields to be included in toString():
+LBL_tostring_use_stringbuilder=&Use StringBuilder
 LBL_implement_method=Implement Method...
 LBL_generate_implement=Generate Implement Methods
 LBL_implement_method_select=Select methods to implement:
--- a/java/java.editor/src/org/netbeans/modules/java/editor/codegen/ToStringGenerator.java
+++ b/java/java.editor/src/org/netbeans/modules/java/editor/codegen/ToStringGenerator.java
@@ -22,8 +22,11 @@ import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.BlockTree;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.ModifiersTree;
+import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.StatementTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.TypeParameterTree;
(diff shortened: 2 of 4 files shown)


kishida:
Is this better to reduce the invocation of `append` for the performance?
```
StringBuilder sb = new StringBuilder();
sb.append("NewClass{test=").append(test);
sb.append(", test2=").append(test2);
sb.append(", test3=").append(test3);
sb.append('}');
return sb.toString();
```
or
```
StringBuilder sb = new StringBuilder();
sb.append("NewClass{" +
  "test=").append(test);
sb.append(", " +
  "test2=").append(test2);
sb.append(", " +
  "test3=").append(test3);
sb.append('}');
return sb.toString();
```

We can include the first string in the constructor.
```
StringBuilder sb = new StringBuilder("NewClass{" +
  "test=").append(test);
```

junichi11:
> Is this better to reduce the invocation of `append` for the performance?

Fixed. Thank you!

junichi11:
Yes, I don't also think people use old versions(e.g. 1.4) in NB. We should be able to remove the unnecessary code (for checking version) if NB would provide support for old Java versions no longer. Just in case, I left them as it is :)

neilcsmith-net:
> But we also could have replaced the old code with StringBuilder and call it a day.

Given JEP 280 wouldn't that be the wrong approach?  Actually, I'm dubious about this PR.

emilianbold:
Hm, interesting remark. So in this case leaving the + is best as the
compiler takes care of everything.

There is on case where using stringbuilder might be nice: as a first
version that gets modified by hand, for example to conditionally append
some text. I suspect in that case the compiler will create multiple
stringbuilder while in our case the user would remove a line a replace it
with an if.

In this situation we need that checkbox.

--emi

dum., 9 iun. 2019, 18:45 Neil C Smith <notifications@github.com> a scris:

AlexFalappa:
If I am not mistaken from Java 8 upwards the compiler is able to use `StringBuilder` under the hood for simple string concatenation like that generated by NB in toString implementations out of a handful of class member variables.

Explicit use of `StringBuilder` is recommended when concatenating string literals in a loop.

See [here](http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html) for a deepest analysis and some benchmarks.

To sum up this modification will be beneficial for those on Java 6 and 7 from a performance point of view

neilcsmith-net:
@AlexFalappa and would be potentially detrimental for anyone on Java 9+ - see [here].(https://dzone.com/articles/jdk-9jep-280-string-concatenations-will-never-be-t)  I'm pretty sure the StringBuilder optimization in javac predates Java 8 too.

So, this might make sense as an option, but probably not the default or something to encourage.

jlahoda:
I wonder what is the advantage of someone generating the toString with explicit StringBuilders? Isn't that what (for 5 <= JDK <=8) javac will produce itself?

If someone wants to customize the toString in such a way that the StringBuilder would help, would including a possibility to expand string concatenation to StringBuilder in the Alt-Enter popup help more?

Anyway, as this is default off, and tested, I don't have strong objections, but I don't see a strong reason for this either. Sorry.

junichi11:
I've added this because it seems that other IDEs has a feature like this. This doesn't mean making people use StringBuilder forcibly. Up to users whether it is used. (The checkbox is unchecked by default.)

Ideas:
- Never add this feature
or
- Add the checkbox when 5 <= version <= 8
or
- Add the checkbox always

I don't have a strong opinion we really need this, but in any case, we should add unit tests because currently, there is no unit test for this generator :)

AlexFalappa:
@neilcsmith-net I was pointing out that this feature is useful to a limited user audience but unharmful if off by default.

I personally prefer to use Benno Marckiewitz _Additional Java Hints_ plugin that offer a group of hints aimed at transforming between concatenation, `StringBuilder`, `StringBuffer` and other neat things. I am able to consciously use `StringBuilder` when I really need.

neilcsmith-net:
@AlexFalappa and I was only pointing out that your comment about concatenation in a loop and benchmark links may be inaccurate for Java 9+, and the performance comment for 6 and 7 is wrong - hence interest in @jlahoda opinion.  Personally, I think that fully generated methods should be legible and performant defaults, and this would complicate the UI and fail at both.  Perhaps you, I and Jan actually agree that hints would be the correct way to approach this use case when the end user does want to customize the generated code?!

AlexFalappa:
@neilcsmith-net I understood your reasoning.

Agreed on the use of hints.

junichi11:
I'm just going to merge this on the weekend if there are no strong objections.

- Other IDEs have this feature
- This feature is unchecked by default

So, I don't think that this is too bad. (If this is too bad, other IDEs should delete this feature.) If someone reports as a problem, we can disable this easily(just disable checkbox).

PR #1289, opened 2019-06-09
[NETBEANS-2654] Add debugger support for Gradle Web Projects

lkishalmi:


groovy/gradle.java/src/org/netbeans/modules/gradle/java/spi/debug/GradleJavaDebugger.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package org.netbeans.modules.gradle.java.spi.debug;
+
+/**
+ * Instance would be found in the project's lookup allow to attach debugger
+ * to the project. This is essentially the same as MavenDebugger for Maven
+ * projects.
+ */
+public interface GradleJavaDebugger {
```
JaroslavTulach: Increment module version in `manifest.mf`. Add `@since` tag. Put an entry to `apichanges.xml`.

lkishalmi: Thank you for that! Just a question on the spec version, after release it bumped from 1.0 to 1.1 on master. So spec version 1.1 is not yet released, do I still need to increase that number?

JaroslavTulach: Depends on how precise you want to be? Users of our official releases will be OK with both choices. However if you had users updating NBMs from our daily builder, then they would already have version 1.1 without your API changes - if only the module using the new API was updated, then they wouldn't link properly. I am usually increasing the specification version for every API change.

PS: Dependencies of modules that require the new API should of course be updated as well.

geertjanw:
Ready to merge this for 11.1?

lkishalmi:
I'm going to add the API increase and some docs, but will be merged today to be ready for 11.1

lkishalmi:
If travis checks out green then this shall go into master then to 11.1

PR #1292, opened 2019-06-10
NETBEANS-2656 remove old JDK migrators under Inspect and Transform

rtaneja1:
Removed Configurations for JDK5, 7. Renamed JDK 10 config as JDK11. No new hints were added for JDK 11 hence no config was added but since it's LTS release, it's better to support config for JDK 11.


geertjanw:
Makes sense. +1 from me.

geertjanw:
Great, merging this.

PR #1293, opened 2019-06-10
[NETBEANS-2657] - Removing obsolete update centers from daily builds

jkovalsky:
NetBeans daily builds contain obsolete and no longer working update centres. This pull request attempts to resolve the problems by removing the useless UCs and leaving just two: one for modules from the latest standard distribution and one for modules from the latest plugin portal version.
More details in #[NETBEANS-2657](https://issues.apache.org/jira/browse/NETBEANS-2657)

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -23,19 +23,9 @@ OpenIDE-Module-Display-Category=Infrastructure
 OpenIDE-Module-Long-Description=It's module which contains the declaration of NetBeans update centers for Autoupdate module.
 OpenIDE-Module-Short-Description=Declares NetBeans autoupdate centers.
 
-Services/AutoupdateType/dev-update-provider.instance=Additional Development Plugins
-Services/AutoupdateType/standard-update-provider.instance=Latest Development Build
-Services/AutoupdateType/pluginportal-update-provider.instance=Plugin Portal
-Services/AutoupdateType/82pluginportal-update-provider.instance=NetBeans 8.2 Plugin Portal
+Services/AutoupdateType/distribution-update-provider.instance=NetBeans Distribution
+Services/AutoupdateType/pluginportal-update-provider.instance=NetBeans Plugin Portal
 #NOI18N
-URL_LatestBuild=http://bits.netbeans.org/dev/nbms-and-javadoc/lastSuccessfulBuild/artifact/nbbuild/nbms/updates.xml.gz
+URL_Distribution=https://builds.apache.org/job/netbeans-linux/lastSuccessfulBuild/artifact/nbbuild/nbms/updates.xml.gz?{$netbeans.hash.code}
 #NOI18N
-URL_Default_N=http://updates.netbeans.org/netbeans/updates/dev/uc/final/main/catalog.xml.gz?{$netbeans.hash.code}
-#NOI18N
-URL_PluginPortal=http://plugins.netbeans.org/nbpluginportal/updates/11.0/catalog.xml.gz
-#NOI18N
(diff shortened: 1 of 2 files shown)


geertjanw:
Awesome, I approve. :-)

geertjanw:
Merging, no reason not to do so.

geertjanw:
Maybe we should include the 8.2 update center (disabled) so that C/C++ can be installed?

neilcsmith-net:
@jkovalsky this appears to have broken the daily build ability to download nb-javac, etc.  The linked update center doesn't exist (yet)?!

jkovalsky:
And the update center has been already created yesterday. Now we need to get the nb-javac plugin to the Maven Central repository. @jlahoda can you please take care of that?

geertjanw:
Sorry, there's some info missing on my end -- what is the update center URL, how does it work together with Maven Central now, and where exactly should the co-ordinates for nb-javac be registered, assuming we can upload nb-javac to Maven Central?

geertjanw:
This also appears to be a problem for Apache NetBeans 11.0, somehow, i.e., not just for the daily build -- can we do what is needed to get the current release, i.e., 11.0, to correctly access nb-javac and the Oracle JS Parser? Or should that not be impacted by this?

jkovalsky:
No, this does not have any influence on Apache NetBeans 11.0. I verified myself that I could install both Oracle JS Parser and nb-javac plugins smoothly.

GeertjanWielenga:
But why must nb-javac be uploaded to Maven Central? Why not the same as always before? And why is ‘experimental’ in the name of the update center?

neilcsmith-net:
Having nb-javac in Maven Central rather than on Oracle infra sounds a positive move.  Would agree with removing "experimental" from the UC name though.

What are we doing about UC for nb-javac, etc. in 11.1 - going back to same as 11.0 or using this?

jkovalsky:
> But why must nb-javac be uploaded to Maven Central? Why not the same as always before?

Because we want to get rid of the Plugin Portal 2.0 hosted on Oracle infra and rather use Plugin Portal 3.0 hosted on Apache infra.

> And why is ‘experimental’ in the name of the update center?

Because the plugins on that update center are not verified.

jkovalsky:
> What are we doing about UC for nb-javac, etc. in 11.1 - going back to same as 11.0 or using this?

For 11.1 I thought we would rely on the same source as for 11.0 i.e. Plugin Portal 2.0

geertjanw:
We need nb-javac and Oracle JS Parser to be in an update center dedicated to 11.1 (or maybe not, i.e., not if they are not different to what we're using in 11.0), we do not want to host them in Maven Central (we don't know how to do that yet), i.e., it should all be as similar to 11.0 as possible for 11.1, I think.

neilcsmith-net:
Surely that's an orthogonal issue?  The UC catalog location doesn't affect whether we use Maven Central to serve the plugins from.  There would probably be benefits to using the intended future location in 11.1.  But if I need to switch this back in the release branch, please let me know.

geertjanw:
> > And why is ‘experimental’ in the name of the update center?
> 
> Because the plugins on that update center are not verified.

Such as what, for example, i.e., which plugins will be in this update center?

jkovalsky:
> Such as what, for example, i.e., which plugins will be in this update center?

http://plugins.netbeans.org/nbpluginportal/updates/11.0/catalog-experimental.xml

jkovalsky:
> But if I need to switch this back in the release branch, please let me know.
If I understand https://cwiki.apache.org/confluence/display/NETBEANS/Release+Schedule correctly, the 11.1 release branch will be created from master, won't it? In such case you will have to update the mf-layer.xml and Bundle.properties file to contain the same values as their 11.0 versions. If 11.1 is branched from 11.0 no such change is obviously needed.

PR #1295, opened 2019-06-11
Using SourceFile attribute when indexing

JaroslavTulach:
Often multiple source roots share a single target directory. With `preferBinaries()` being `true` the index (used by Go to Type dialog) incorrectly reports classes being in all source roots.

This PR fixes that by checking the `SourceFile` attribute of each `.class` and indexes only those that come from the correct source root.

The PR also fixes the `BinaryAnalyserTest` by adding missing annotation sources.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
@@ -20,6 +20,7 @@
 package org.netbeans.modules.java.source.indexing;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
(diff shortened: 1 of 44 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
```diff
@@ -189,7 +193,17 @@ protected void index(final Iterable<? extends Indexable> files, final Context co
                     at = 0;
                     for (URL singleBinaryRoot : binaryRoots) {
                         FileObject singleBinaryRootFo = binaryRootsFo[at++];
-                        JavaBinaryIndexer.doIndex(context, singleBinaryRoot);
+                       JavaBinaryIndexer.doIndex(context, singleBinaryRoot, (cf) -> {
+                            String src = cf.getSourceFileName();
+                            if (src != null) {
+                                String srcName = src.substring(src.lastIndexOf(File.separatorChar) + 1);
+                                if (!shortNames.contains(srcName)) {
```
sdedic: Just `return shortNames.contains(srcName)` ?

PR #1303, opened 2019-06-17
Listening on random port made possible with ListeningDICookie.create(-1)

JaroslavTulach:
It's possible to create a JPDA connection listener on a random port via standard API.


JaroslavTulach:
CCing @entlicher

JaroslavTulach:
Neil, would you be so kind and approve merging of this change into 11.1? It is not a bugfix in Apache NetBeans per se, but it is a bugfix for [IGV](https://www.graalvm.org/docs/graalvm-as-a-platform/implement-language/#igv) to remove a nasty reflection that workaround the missing API.

I am not convinced the change is compatible and should have no negative impact on stability of 11.1.

neilcsmith-net:
Fine with me, on your head be it.  I assume you didn't mean "I am not convinced"?! :smile:

PR #1305, opened 2019-06-17
Cleanup junit assert

BradWalker:
There are some deprecation warnings sent by using JUnit..

[repeat] /home/bwalker/netbeans-11.0/java/java.hints.test/src/org/netbeans/modules/java/hints/test/api/HintTest.java:57: warning: [deprecation] Assert in junit.framework has been deprecated

```
[repeat] import junit.framework.Assert;
[repeat] ^
[repeat] /home/bwalker/netbeans-11.0/java/java.hints.test/src/org/netbeans/modules/java/hints/test/api/HintTest.java:58: warning: [deprecation] Assert in junit.framework has been deprecated
[repeat] import static junit.framework.Assert.assertEquals;
[repeat] ^
[repeat] /home/bwalker/netbeans-11.0/java/java.hints.test/src/org/netbeans/modules/java/hints/test/api/HintTest.java:59: warning: [deprecation] Assert in junit.framework has been deprecated
```

These warning are because the JUnit class has changed. The method has stayed the same..

So changing from
import junit.framework.Assert;

to
import org.junit.Assert;

will fix all of these..


matthiasblaesing:
These changes break the unittests:
- TinyTest#testAssertEqualsForArrays
- TinyTest#testAssertEqualsForArraysWithMessages
The issue is, that you are changing test code that is supplied as a string. For these it is not enough, that they still compile.

BradWalker:
Since Matthias has found a problem, I'll close this and pick it up at a later date..

PR #1309, opened 2019-06-18
Finish cleanup to url

BradWalker:
There are a few remaining File.toURL() deprecations remaining, remove them..

```
   [repeat] /home/bwalker/netbeans-11.0/ide/spi.palette/src/org/netbeans/modules/palette/ui/TextImporterUI.java:257: warning: [deprecation] toURL() in File has been deprecated
   [repeat]                 largeIconPath = iconFile.toURL().toExternalForm();
```

Also remove this deprecation that was exposed from the previous cleanup..

```
   [repeat] /home/bwalker/netbeans-11.0/enterprise/websvc.manager/src/org/netbeans/modules/websvc/manager/codegen/Wsdl2Java.java:211: warning: [deprecation] getURL() in WebServiceData has been deprecated
   [repeat]             String wsdlUrlName = new File(webServiceData.getURL()).toURI().toURL().toString();
   [repeat]                
```                                         ^

### CODE REVIEW

File: enterprise/websvc.manager/src/org/netbeans/modules/websvc/manager/model/WebServiceData.java [hunk 1]
```diff
@@ -216,16 +216,6 @@ public void setWsdlFile(String fileName) {
         this.wsdlFile = fileName;
     }
 
-    /**
-     * 
-     * @return the WSDL file (absolute path)
-     * @deprecated use {@link #getWsdlFile()} instead
-     */
-    @Deprecated
-    public String getURL() {
```
matthiasblaesing: This changes the API of the module. If you want to remove this, you'll have to
- check all current users of the module if the method is used (it is friend-only) and fix these
- you'll have to increment major version and update specification version (see: http://wiki.netbeans.org/VersioningPolicy)

matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/3498ed8164c3d82c390dd4dad477630b43c2b827 without the API change.

PR #1311, opened 2019-06-20
[NETBEANS-2417] Gradle Logo Updates

jlstrater:
This is a follow-up to https://github.com/apache/netbeans/pull/1269#issue-283183154

The design team recommended the use of this light grey icon for both the dark and light backgrounds. It's the same icon used by IntelliJ and that I'll submit to Eclipse as well. 

As for the licenses, I checked and these are Apache 2 licensed so no change to the license file is necessary.


louhy:
There is also this semi-related issue with the "red coffee cups" obscuring red error indicators: 
https://issues.apache.org/jira/browse/NETBEANS-2608

Demonstrated here on the left (with old ones from other PR on right), with NB default and Darcula light/dark themes: https://imgur.com/yZYqTY7

I have to say I'm disappointed we aren't getting the blue versions, IMO they looked much better alongside NB's other color schemes. IntelliJ has a different design where they don't show the icon in their main workspace (it's only one "project" at a time anyway), and IIRC Eclipse doesn't use icons at all in their main workspace. NetBeans design is different in that it uses colored icons for _each_ project. 

The grey isn't as bad on the dark background, but on the default NetBeans white background the grey icons are almost invisible to me. Grey kinda makes me think "disabled/inactive" in this context. 

(Also I think NetBeans already uses greyed-out icons to indicate "de-activated module" project types when creating a new project, but that's a minor issue.)

jlstrater:
Please refer to #1269

PR #1312, opened 2019-06-20
[NETBEANS-2400] Fix for AWS Cloud missing Regions

peedeeboy:
What this PR does
=================
* Updates AWS regions in NB to include all current AWS regions
* Updates AWS Java SDK to 1.11.569 - needed as newer regions use v4 signing protocol
* Enables creation of JavaEE 7 applications when AWS container is running Tomcat 8.x
* Filters available AWS Containers to only those running Tomcat

Controversies!
============
* AWS Java SDK 1.11.569 moves/removeds some classes/methods.  This causes the NB sigtest to abort the build.  I don't think the AWS SDK will ever be really stable in the way NB's internal APIs usually are (especially when we jump to AWS SDK v2), so I have added the sigtest.gen.fail.on.error=false project property to ignore the sigtest errors...
* I have not updated the org.apache.commons.codec and org.apache.commons.logging modules with new .jar files.  From my testing, the AWS SDK worked fine with the current libraries.  If the AWS modules are expanded, these may need to be updated if any new methods/classes are consumed..

Files Changed
=============
Amazon Elastic Beanstalk Module
-------------------------------
* AmazonDeploymentFactory.java - Passes regionCode into new AmazonDeploymentManager instances
* AmazonInstance.java - Uses the new Client Builder classes (which require the Region Code, rather than Endpoint) rather than the now deprecated constructors.  Creates a default instance profile with sensible (?) values, as the Java SDK doesn't create one by default!
* AmazonInstanceTest.java - Now uses system temp directory rather than hardcoded path and is altered for modified AmazonInstance constructor
* AmazonJ2EEServerInstanceProvider.java - Now also stores Region Code in properties
* AmazonWizardComponent.java - Contains full list of current AWS instances and also includes their Region Code
* AmazonDeploymentManager.java - Passes in Region Code to AmazonInstance when deploying an application
* AmazonJ2eePlatformImpl2.java - Now checks if Container is running Tomcat 8.x and if so allows creation of Java EE 7 application
* Project.xml - Module now depends on Amazon SDK module 1.16
* AmazonInstanceManager.java - Now stores Region Code in instance properties
* AmazonWizardPanel.java - Now stores Region Code in properties
* AmazonWizardIterator.java - Passes Region Code into AmazonInstanceManager 
* ViewAdminConsoleAction.java - Uses Region Code to generate URL, rather than parsing Endpoint
* AmazonJ2EEServerWizardComponent - Filters AWS Containers to only those running Tomcat

Amazon SDK Module
-----------------
* binaries-list - Removed references to old dependencies, and added Maven central references for new dependecies
* project.xml - Dependencies updated
* manifest.mf - API version updated to 1.16
* project.properties - Added sigtest.gen.fail.on.error=false to prevent sigtest stopping the build due to deprecated/removed methods in the SDK.  Updated dependency locations
* Several license files removed & added

Testing
=======
* If running from NB, build Amazon SDK module, then Amazon Elastic Beanstalk module.  Otherwise, build NB.
* In NB, Services tab -> Cloud -> Right-click -> Add Cloud...
* Cloud: Amazon Beastalk -> Next -> Pick a region (check all AWS regions now in dropdown) -> Enter Key ID and Secret Access Key -> Next -> Finish
* File menu -> New Project -> Java with Maven -> Web Application -> Next >
* Complete Name & Location Details -> Next >
* Click Add next to Server
* Choose Server = Amazon Beanstalk -> Next >
* Complete Application Name, Environment Name & Environment URL fields
* Pick the Container running Amazon Linux 3.1.6 & Tomcat 8.5
* Click Finish
* Make sure new Server is selected in dropdown and Java EE version = Java EE 7 Web -> Finish

* Services tab -> AWS Beanstalk node -> Right-click -> Open AWS Console
* Log in to AWS.  Go and make a cup of tea whilst you wait for the new Environment to start (status = Green), takes ~ 5 minutes
* AWS Beanstalk node -> Right-click -> Refresh.  Under the Servers node, the AWS Tomcat node show now show running icon

* Projects -> Right-click on project -> Run
* Go and make another cup of tea whilst AWS reconfigures (~ 3 minutes)
* Check output and make sure deployment successful.  
* Browser will open.  Most likely a blank page.  Refresh and application should be displayed...

Don't forget to Terminate the environment and delete the application in AWS control panel!!


peedeeboy:
@matthiasblaesing - if you get chance, I would appreciate your critical appraisal, particularly around license files in AWS SDK module...

peedeeboy:
Updated -license.txt headers as per https://cwiki.apache.org/confluence/display/NETBEANS/Legal+FAQ.  Hopefully, this will resolve the Travis build failure...

peedeeboy:
Updated -license.txt files (again!) so that they *exactly* match the content of the `nbbuild/licenses` folder.  Added binary dependencies that are the same as Gradle module to `ignored-overlaps`, as two module are probably never likely to be updated at the same time....

3rd time lucky?

matthiasblaesing:
I'll have a look at it, will most probably take till next week though.

peedeeboy:
@matthiasblaesing No rush!  Appreciate your time and patience.

enterprise/libs.amazon/external/ion-java-1.0.2-license.txt
```diff
@@ -0,0 +1,208 @@
+Name: Amazon Ion Java
+Origin: Amazon Labs
+Version: 1.0.2
+License: Apache-2.0
+Description: A Java implementation of the Amazon Ion data notation
+URL: https://github.com/amznlabs/ion-java/
```
matthiasblaesing: there is a NOTICE file in the git referenced ion-java repository, I would add it (as you already did for aws-java-sdk-core)

enterprise/libs.amazon/external/jackson-annotations-2.6.0-license.txt
```diff
@@ -0,0 +1,209 @@
+Name: Jackson-annotations
+Origin: FasterXML
+Version: 2.6.0
+License: Apache-2.0
+Description: Core annotations used for value types, used by Jackson data binding package
+URL: http://github.com/FasterXML/jackson
```
matthiasblaesing: Wouldn't https://github.com/FasterXML/jackson-annotations be a better url?

enterprise/libs.amazon/external/joda-time-2.8.1-notice.txt
```diff
@@ -0,0 +1,5 @@
+=============================================================================
+= NOTICE file corresponding to section 4d of the Apache License Version 2.0 =
+=============================================================================
```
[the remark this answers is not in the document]
peedeeboy: Removed, as requested.

enterprise/libs.amazon/external/ion-java-1.0.2-license.txt
peedeeboy: NOTICE file added as `enterprise/libs.amazon/external/ion-java-1.0.2-notice.txt`

enterprise/libs.amazon/external/jackson-annotations-2.6.0-license.txt
peedeeboy: I think I pulled that URL from the pom file for the library, but agree that your suggestion makes more sense.  I have updated accordingly.

I have also merged Amazon SDK licenses into one file, as requested.

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstance.java
```diff
@@ -83,18 +85,25 @@
     
     public static final String DEFAULT_EMPTY_APPLICATION = "empty-nb-app-1.0.war";
     
+    private final String INSTANCEPROFILE_NAMESPACE = "aws:autoscaling:launchconfiguration"; // NOI18N
+    private final String INSTANCEPROFILE_OPTIONNAME = "IamInstanceProfile"; // NOI18N
+    private final String INSTANCEPROFILE_VALUE = "aws-elasticbeanstalk-ec2-role"; // NOI18N
+    
+
```
matthiasblaesing: I would change this to a single line of separation.

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstance.java
```diff
@@ -131,10 +144,10 @@ private static AWSCredentials getCredentials(String keyId, String key) {
...
     public void testConnection() {
         assert !SwingUtilities.isEventDispatchThread();
-        AWSElasticBeanstalk client = new AWSElasticBeanstalkClient(getCredentials());
-        if (regionUrl != null) {
-            client.setEndpoint(regionUrl);
-        }
+        AWSElasticBeanstalk client = AWSElasticBeanstalkClientBuilder.standard()
+                                        .withCredentials(new AWSStaticCredentialsProvider(getCredentials()))
+                                        .withRegion(regionCode)
+                                        .build();
```
matthiasblaesing: I got the expression, that this sequence is repeated in this file. Would it make sense to introduce a private method "createAWSElasticBeanstalkClient" and unify the initialization code for the client there?

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstanceManager.java
```diff
@@ -113,7 +116,7 @@ private void store(AmazonInstance ai) {
                 continue;
             }
             String key = new String(ch);
-            result.add(new AmazonInstance(name, keyId, key, region));
+            result.add(new AmazonInstance(name, keyId, key, region, code));
```
matthiasblaesing: Just to make sure - does this work if code is null? That situation would be reached when an old userdir is read with a new version of netbeans.

enterprise/cloud.amazon/test/unit/src/org/netbeans/modules/cloud/amazon/AmazonInstanceTest.java
```diff
@@ -36,9 +36,10 @@ public AmazonInstanceTest() {
 
     @Test
     public void testCreateEmptyWar() throws IOException {
-        AmazonInstance ai = new AmazonInstance("dommy", "dummy", "dummy", "dummy");
+        AmazonInstance ai = new AmazonInstance("dommy", "dummy", "dummy", "dummy", "dummy");
         byte[] warData = ai.createEmptyWar();
-        Files.write(Paths.get("/home/matthias/test.zip"), warData, StandardOpenOption.CREATE, StandardOpenOption.WRITE);
+        String filePath = System.getProperty("java.io.tmpdir") + "/test.zip";
```
matthiasblaesing: Yepp, this test was not sane ... (and I introduced the hardcoded path), thanks for fixing.

enterprise/libs.amazon/nbproject/project.xml
```diff
@@ -39,23 +39,66 @@
             <friend-packages>
                 <friend>org.netbeans.modules.cloud.amazon</friend>
                 <package>com.amazonaws</package>
+                <package>com.amazonaws.auth</package>
+                <package>com.amazonaws.client</package>
+                <package>com.amazonaws.client.builder</package>
                 <package>com.amazonaws.services.elasticbeanstalk</package>
                 <package>com.amazonaws.services.elasticbeanstalk.model</package>
                 <package>com.amazonaws.services.s3</package>
+                <package>com.amazonaws.services.s3.internal</package>
```
matthiasblaesing: I packages are already called `internal` I would not expose them (if they are needed, then ignore this :-))

peedeeboy:
Thanks for the feedback @matthiasblaesing.  Unfortunately, I cannot make the requested changes as it seems like it is no longer possible to create Amazon Beanstalk server instances in NB11.1 or builds from master? (Only Tomcat, Glassfish and Payara servers can be created)

I'm not sure if this is as accidental bug that has been introduced, or if it is a deliberate decision to remove AWS support?  If the later, this PR could be Closed.

geertjanw:
Why would there be a deliberate decision to remove AWS support? :-) If there's a bug, just file an issue, and since you're doing great with your enhancements and fixes, please also consider fixing it.

matthiasblaesing:
As I already said: I don't own an AWS account, so you need to tell me a bit more, where you expect what. There was some work in the JavaEE area to support JavaEE8 and this might have disrupted things, but with a recent local build, I have the "Cloud" entry in the services tab. Please check if the JavaEE plugins are all activated.

peedeeboy:
With a fresh install of NB11.1 and enabling all Java + JavaEE features, where the Amazon Beanstalk entry is missing is:

New Project -> Java with Maven -> Java Web -> Server & Settings -> JavaEE7 -> Add

Amazon Beanstalk should be in the list of Server types to add to the Project, but isn't anymore (Only Tomcat, Glassfish & Payara).  

I remember seeing something on the NB Mailing list about duplicate entries in this dialog after Payara was merged in, so suspect it may be related to this (rather than JavaEE8), but without tracing the related commits/PRs and going through the code I can't be certain at this point.

I will have a look myself if I get time.

matthiasblaesing:
I think you mean PR #1347, indeed that PR looks as if it touched relevant code. @jGauravGupta could you please have a look at this PR. Long story short: AWS entries are expected to be present in the list of Java EE servers but are not anymore.

jGauravGupta:
Sure @matthiasblaesing, soon I will have look on this issue.

peedeeboy:
Thank you, @jGauravGupta !!

geertjanw:
Ready to be merged or not?

peedeeboy:
@geertjanw - not quite! Once #1412 is merged, I can rebase off master, make the the couple of changes @matthiasblaesing requested and test everything plays nicely together :)

geertjanw:
Excellent.

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstance.java
peedeeboy: Done :)

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstance.java
peedeeboy: Done :)  There are a couple of methods where the credentials are being passed in from outside, for those two methods, the Client is still being constructed in those methods...

enterprise/libs.amazon/nbproject/project.xml
peedeeboy: com.amazonaws.services.s3.internal needs to be exposed or building the Amazon EB module will fail due to not being able to access com.amazonaws.services.s3.internal.S3DirectSpi

enterprise/cloud.amazon/src/org/netbeans/modules/cloud/amazon/AmazonInstanceManager.java
peedeeboy: Great catch.  I didn't even think of this :(

We need to use the region code as setting a region endpoint by URL is deprecated in newer AWS SDK, and won't work for newer regions (e.g. London).

So I've broken the details of the Amzon regions out from the Wizard component into a seperate AmazonRegion enum.  Then I'm checking in AmazonInstanceManager if code is null, and if it is, I can retrieve it from the enum based on the URL.  I beleive that will mean old userdirs containing just the URL won't break with the new versions of the modules...

peedeeboy:
Requested changes made.  Hope all ok now :)

enterprise/libs.amazon/nbproject/project.xml
matthiasblaesing: Why? they should be in the same module and thus being able to directly access all classes.

peedeeboy: All I can say right now is that the `AmazonS3` interface (in `com.amazonaws.services.s3` package) extends the `com.amazonaws.services.s3.internal.S3DirectSpi` interface, and if the s3.internal pacakge is not exposed building the EB module will fail with:

```
 error: cannot access S3DirectSpi
            s3.getObjectMetadata(bucket, DEFAULT_EMPTY_APPLICATION);
  class file for com.amazonaws.services.s3.internal.S3DirectSpi not found
```

I might do some testing tomorrow to see if this always happens in NB modules when an Interface extends another Interface in an inacessible package... but it does seem a bit odd.

matthiasblaesing: I reproduced it - yeah the resolver needs access to the internal class. Thanks for checking.

matthiasblaesing:
Looks good to me.

PR #1313, opened 2019-06-21
Download binaries task is broken

JaroslavTulach:
Turns out that my f2c5275 causes a regression. `DownloadBinaries` task from `harness/tasks.jar` is unusable as the `harness/tasks.jar` misses now necessary `ConfigureProxy` class.

To prevent such unfortunate error I wanted to write a test. My attempts to unit test the task failed and at the end I decided to do 022de83 

The fix then was simple: 02b8d54


JaroslavTulach:
I discovered while working on https://github.com/oracle/visualvm/pull/137/ CCing @thurka

PR #1317, opened 2019-06-23
[NETBEANS-1696] PHP 7.4 Support

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1696

## Numeric Literal Separator
See https://wiki.php.net/rfc/numeric_literal_separator

- Fix ASTPHP5Scanner.flex and PhpColoringScanner.flex
- Add unit tests
- Fix PHP74UnhandledError

nb-php7 4-numeric-literal-separator
nb-php7 4-numeric-literal-separator-error

## New custom object serialization mechanism
See https://wiki.php.net/rfc/custom_object_serialization

- Add `__serialize()` and `__unserialize(array $data)` magic methods
- Add/Fix unit tests
- Didn't check the version but we should be able to do that easily if needed


tmysik:
@junichi11 

> Didn't check the version but we should be able to do that easily if needed

What you mean exactly?

tmysik:
@junichi11 

Maybe try to add test for e.g. `11_22` (`Use cases to avoid` in [docs](https://wiki.php.net/rfc/numeric_literal_separator))? If I understand it correctly, it is valid (it should be avoided but still, it is not a syntax error). Am I right? Thanks.

junichi11:
> What you mean exactly?

It means that `__serialize()` and `__unserialize()` are always available although they are available since PHP 7.4. i.e. The two methods are added without checking the PhpVersion.

https://github.com/apache/netbeans/blob/71b02dbfeabb0db923400cab33d5ef1d3c740dba/php/php.editor/src/org/netbeans/modules/php/editor/elements/MethodElementImpl.java#L95-L98


> Maybe try to add test for e.g. `11_22` (`Use cases to avoid` in [docs](https://wiki.php.net/rfc/numeric_literal_separator))? If I understand it correctly, it is valid (it should be avoided but still, it is not a syntax error). Am I right? 

It means this, right?
```php
// don't do this:
$phoneNumber = 345_6789;
$creditCard = 231_6547_9081_2543;
$socialSecurity = 111_11_1111;
```
Yes, you are right. These numbers are not syntax errors.
I'll add numbers added the separator(_) randomly to the test file later.
e.g. 
```php
12345_6_7_89_0;
1_23_456_7_89.8_8_888_88;
0x1_234_ABCD;
0b0_1_0111_11;
012_3_456;
```
Thanks for the heads up :) Maybe, I'll push tomorrow.

tmysik:
@junichi11 

> It means that `__serialize()` and `__unserialize()` are always available although they are available since PHP 7.4. i.e. The two methods are added without checking the PhpVersion.

Totally OK, in the code base, we always support the latest PHP runtime (and provide detailed info in PhpDoc and/or error hints).

> It means this, right?

Yes, exactly.
(In fact, I would expect that only "grouping numbers by 3" will be possible but apparently it is not.)

Thanks!

junichi11:
@tmysik Fixed. Thanks!

tmysik:
@junichi11 Great, thank you!

PR #1318, opened 2019-06-23
Travis, run ide modules tests

blackleg:
Change Travis config to run ide modules tests that are working currently.


JaroslavTulach:
I am always glad to see more tests! This change looks amazing. Moreover the new job with tests runs just 13 min!

PR #1319, opened 2019-06-23
Partial Reparse without nb-javac

jlahoda:
This patch attempts to add partial reparse (i.e. when the user types into the editor, inside a single method only, reparse and re-attribute only that given method, not the whole file), which (should) speed up many features significantly. The partial reparse is used when nb-javac is installed, but the goal here is to add it also when nb-javac is not use/available.


jlahoda:
I've changed the partial reparser to be off by default, and I think it would be good to merge it into the master - it would make it much easier to find pending problems.

geertjanw:
Can you provide some scenarios we can try to see that this is working correctly?

arvindaprameya:
Awesome thank you Hanza !!

lkishalmi:
Info: Add "-J-Djava.enable.partial.reparse=true" to the command line options in netbeans.conf to enable it. As I'm living on the weekly updated NetBeans master builds, I've turned it on for myself, in order to be able to report if something goes south.

PR #1323, opened 2019-06-25
[NETBEANS-2745]:Fixed Convert Switch to Switch-Expression hint for default case without break

arusinha:
…fault case without break

The PR is bug fix for PR#1193 (https://issues.apache.org/jira/browse/NETBEANS-2349) which is part of NB-11.1 release

Convert old switch statement to switch expression in following case:

Old code:  (Assign same variable in every case with one statement)

int x = 5;
String str = "foo";
switch (str)

{ case 1: str = "bar";break; case 2: str = "baz";break; default: str = "zuq" }

 

expected fix:

int x = 5;
String s = switch (str)

{ case 1 -> "bar"; case 2 -> "baz"; default -> "zuq"; }

But the hint is disabled in above scenario.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
@@ -3114,7 +3114,11 @@ public class Utilities {
     public static boolean isCompatibleWithSwitchExpression(SwitchTree st) {
         boolean firstCase = true;
         Name leftTreeName = null;
-        for (CaseTree ct : st.getCases()) {
+        int caseCount = 0;
+        List<? extends CaseTree> cases = st.getCases();
+
+        for (CaseTree ct : cases) {
+            caseCount++;
             List<StatementTree> statements = new ArrayList<>(ct.getStatements());
             switch (statements.size()) {
                 case 0:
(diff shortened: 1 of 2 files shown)


java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
```diff
@@ -525,7 +525,65 @@ public void testSwitch2SwitchExpressionMultiCase() throws Exception {
                               "     }\n" +
                               "}\n");
     }
-
+     public void testSwitch2SwitchExpressionMultiCase2() throws Exception {
```
junichi11: 5 whitespaces

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
```diff
@@ -525,7 +525,65 @@ public void testSwitch2SwitchExpressionMultiCase() throws Exception {
...
+                              "         String result;\n" +
+                              "         result = switch (p) {\n" +
+                              "             case 1, 2 -> \"2\";\n" +
+                              "             case 3 -> \"3\";\n" +
+                              "             default -> \"default\";\n" +
+                              "         };\n" +
+                              "     }\n" +
+                              "}\n");
+    }
+public void testSwitch2SwitchExpressionOnlyDefault() throws Exception {
```
junichi11: Missing indentation?

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
```diff
@@ -3114,16 +3114,29 @@ public Void visitClass(ClassTree node, Void p) {
...
+                        StatementTree statementTree = statements.get(0);
+                        JCTree.JCExpressionStatement jceTree = (JCTree.JCExpressionStatement) statementTree;
+                        if (!(jceTree.expr instanceof JCTree.JCAssign)) {
+                            return false;
+                        }
+                        JCTree.JCAssign assignTree = (JCTree.JCAssign) jceTree.expr;
+                       if (firstCase || leftTreeName != null && leftTreeName.contentEquals(((JCTree.JCIdent) assignTree.lhs).name)) {
+                            break;
+                        } 
+                       else return false;
```
junichi11: Maybe, we should use `{}` and should format the `if` statement.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
arusinha: had corrected the same in recent commit.

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
arusinha: had corrected the same in recent commit.

ApolloLV:
You appear to have used NETBEANS-2475, which is an unrelated bug report in JIRA.

arusinha:
Thanks @ApolloLV , I had corrected the commit message now

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
```diff
@@ -3114,16 +3114,30 @@ public Void visitClass(ClassTree node, Void p) {
     public static boolean isCompatibleWithSwitchExpression(SwitchTree st) {
         boolean firstCase = true;
         Name leftTreeName = null;
-        for (CaseTree ct : st.getCases()) {
+        int CaseCount =0;
```
rtaneja1: Please correct source format and preferably var name also, start with lowercase letter.

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
```diff
@@ -3114,16 +3114,30 @@ public Void visitClass(ClassTree node, Void p) {
...
             List<StatementTree> statements = new ArrayList<>(ct.getStatements());
             switch (statements.size()) {
                 case 0:
                     break;
                 case 1:
                     if (firstCase && leftTreeName == null && statements.get(0).getKind() == Tree.Kind.RETURN) {
                         break;
-                    } else {
-                        return false;
+                    } else if (CaseCount == cases.size() && statements.get(0).getKind() == Tree.Kind.EXPRESSION_STATEMENT) {
```
rtaneja1: Majority of the code in case 1 is same as case 2, can reuse?

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
arusinha: corrected in last commit

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
arusinha: had  reused the common code in last commit

PR #1324, opened 2019-06-25
[NETBEANS-2730] try to update to nbm-maven-plugin from Apache + backwar…

ebarboni:
…d compatibility on mojo

Try to keep old mojo working.

This is an initial trial but it also need archetypes release before merge.


neilcsmith-net:
OK, re-triggered tests pass so merging.

PR #1326, opened 2019-06-25
[NETBEANS-2747] Apache NetBeans 11 shortcut card sources

geertjanw:
Making the title generic so we don't have to change each release, since the shortcuts are stable and not changing for some time already.


neilcsmith-net:
@geertjanw thanks, saves me a job on my list for today. There was a commit for NB11.0 for this but unfortunately had extra things we don't need.

Travis seems to failing all over the shop at the moment!  May just merge these two as they don't touch anything that can break anything?!

geertjanw:
Agree, merge these two, when/if you’re ready for them.

PR #1328, opened 2019-06-25
HashMap is not thread safe. Field map is typically synchronized by this. However, in two places, field map is not protected with synchronized.

paulward24:
The current field

```map```

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L112

is typically protected by synchronization on ```this```, e.g., 

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L386

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L390

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L426

For a total of 16 locations.

However, in these two places, ```map``` is not protected by synchronization:

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L207

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L210

Note that in the same method, only 4 lines above:

https://github.com/apache/netbeans/blob/master/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java#L202

```map``` is again protected by synchronization:

```
synchronized (this) {
    attr = (Attr) map.get(attrName);
}
```

This CR protected the two instances motioned above (line 207 and
line 210) with synchronization like in line 202.

### DIFF
--- a/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java
+++ b/platform/openide.filesystems/src/org/openide/filesystems/XMLMapAttr.java
@@ -204,10 +204,14 @@ final class XMLMapAttr implements Map {
 
         Object retVal = null;
         if (attr == null && origAttrName.startsWith("class:")) { // NOI18N
-            attr = (Attr) map.get(origAttrName.substring(6));
+            synchronized (this) {
+                attr = (Attr) map.get(origAttrName.substring(6));
+            }
             retVal = attr != null ? attr.getType(params) : null;
         } else if (attr == null && origAttrName.startsWith("raw:")) { // NOI18N
-            attr = (Attr) map.get(origAttrName.substring(4));
+            synchronized (this) {
+                attr = (Attr) map.get(origAttrName.substring(4));
+            }
             if (attr != null && attr.keyIndex == 9) {
                 return attr.methodValue(attr.value, params).getMethod();
             }


paulward24:
I see the CI fail, but unclear to me why this small change would break anything.  Maybe somebody more familiar with the project can take a look.

The CI fail is something about:

The command "ant $OPTS clean" exited with 1.

vieiro:
Wow, all those "synchronized" things look terrific to me as well :-)
While we're at it, what about a cleanup and using a "Collections.synchronizedMap" instead? Wouldn't that be good enough?

matthiasblaesing:
Switching to `Collections#synchronizedMap` would alter the semantics -- the synchronisation covers more than just the access to the map, but also the serialization of it (see `write`). So this is the minimal invasive change, but I agree, the class might need some love.

matthiasblaesing:
Merged - thank you.

vieiro:
Thank you, @matthiasblaesing  and @paulward24

PR #1329, opened 2019-06-25
[NETBEANS-2729] Make netbeans and netbeans-dev snap descriptor coexists.

lkishalmi:


neilcsmith-net:
Given changes already in release111 can this be merged post-release?  If so, should tag with NB11.2

PR #1331, opened 2019-06-26
Add back Third Party Libraries plugin center.

neilcsmith-net:
Add back Third Party Libraries plugin center.

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -33,3 +33,7 @@ URL_Distribution=https://netbeans.apache.org/nb/updates/11.1/updates.xml.gz?{$ne
 URL_PluginPortal=https://netbeans.apache.org/nb/plugins/11.1/catalog.xml.gz
 #NOI18N
 URL_82PluginPortal=http://updates.netbeans.org/netbeans/updates/8.2/uc/final/distribution/catalog.xml.gz
+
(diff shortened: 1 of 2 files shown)


nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
```diff
@@ -33,3 +33,7 @@ URL_Distribution=https://netbeans.apache.org/nb/updates/11.1/updates.xml.gz?{$ne
 URL_PluginPortal=https://netbeans.apache.org/nb/plugins/11.1/catalog.xml.gz
```
jkovalsky:
```suggestion
URL_PluginPortal=http://plugins.netbeans.org/nbpluginportal/updates/11.0/catalog.xml.gz
```

neilcsmith-net: That is already where it goes to at the moment - see https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess#L9  There was a discussion a while back about trying to keep all IDE inbound links under one root where possible - eg. https://netbeans.apache.org/nb/

neilcsmith-net:
@jkovalsky can we keep this to just re-adding the Third-Party Libraries back in?  As mentioned off-list, I'm not sure we should add the version number in the descriptions, particularly not with 11.0 in it, and the plugin link should be functionally correct already.  Happy to reconsider between beta2 and beta3 next week, but I need to sort this build now or it won't happen this week.

PR #1333, opened 2019-06-26
NETBEANS-544: Merge conflicts resolver doesn't recognize "|||||||" conflict markers

hodawg04:
Updated ResolveConflictsExecutor in Subversion using Git as a reference.


matthiasblaesing:
Change looks reasonable and I assume, that it was manually tested. Merging. Thank you.

PR #1335, opened 2019-06-27
[NETBEANS-2667] Manage JOptionPane border

YannLeCorse:
In the NB Dialog API, the JOptionPane is wrapped on a JPanel and the buttons moved on another panel. This leads to UI issues when the Look & Feel defines a special border for JOptionPane. The goal of this change is to retrieve the border defined by the L&F and apply it at the content pane level. Doing that we reduce inconsistencies between dialogs opened through the Java APIs or through the Netbeans APIs.

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
@@ -15,9 +15,6 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
- *
- * Contains a bug fix contributed by Experian. © Experian, 2019. Licensed under Apache 2.0.
- *
  */
 
 package org.netbeans.core.windows.services;
@@ -108,14 +105,14 @@ import org.openide.util.Utilities;
  */
 class NbPresenter extends JDialog
 implements PropertyChangeListener, WindowListener, Mutex.Action<Void>, Comparator<Object> {
-    
+
     /** variable holding current modal dialog in the system */
     public static NbPresenter currentModalDialog;
     private static final ChangeSupport cs = new ChangeSupport(NbPresenter.class);
     private static Boolean isJava17 = null;
-    
+
     protected NotifyDescriptor descriptor;
-    
+
     private final JButton stdYesButton = new JButton(NbBundle.getBundle(NbPresenter.class).getString("YES_OPTION_CAPTION")); // NOI18N
     private final JButton stdNoButton = new JButton(NbBundle.getBundle(NbPresenter.class).getString("NO_OPTION_CAPTION")); // NOI18N
     private final JButton stdOKButton = new JButton(NbBundle.getBundle(NbPresenter.class).getString("OK_OPTION_CAPTION")); // NOI18N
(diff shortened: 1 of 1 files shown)


platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
```diff
@@ -15,6 +15,9 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
+ *
+ * Contains a bug fix contributed by Experian. © Experian, 2019. Licensed under Apache 2.0.
```
junichi11: I don't think that we should add this here.

@matthiasblaesing What do you think?

YannLeCorse: Build failed because of that but I've asked to add it to comply the license. Happy to remove it if it's not needed.

**Request I received**

Notice changes needed
Clauses 4(b) and (c) of the licence are the relevant clauses. 4(b) says:
 
_“You must cause any modified files to carry prominent notices stating that You changed the files”_
 
How we comply with this depends how the file is structured. If the notice for a piece of code is found at the start of that particular code, then it can be a simple notice:
 
_Contains a bug fix contributed by Experian. © Experian, 2019. Licensed under Apache 2.0._
 
If all the notices are together at the front of the file, then our notice will also need to describe the particular code we’ve fixed. This notice format would work:
 
_NAME OF THE PARTICULAR CODE contains a bug fix contributed by Experian. © Experian, 2019. Licensed under Apache 2.0._
 
4(c) says that we must retain all notices already present in the software. Hopefully more straightforward! Please let me know if you would like to discuss.

junichi11: I'm not an expert about legal problems, sorry. So, I could be wrong. The license header would be an irregular pattern. If we have to change the license header in the future, we have to consider irregular patterns. It is the reason why I wrote "here". 

I think maybe Matthias(@matthiasblaesing ) could advise something. Let's wait :)

BTW, probably, that is not the cause of build failure because other PRs also fail.

geertjanw: Yes. This causes the build to fail. Is that special license notice really necessary, @YannLeCorse?

matthiasblaesing: To my understanding: This code is owned by the ASF. The ASF either requires contributions to be to small to create a separate copyright (I'm not sure that I like the term "trivial" for bugfixes) or to be explicitly donated to the ASF. So if you don't add the above line and propose a PR, I would consider it donated to the ASF. It is also small enough to be called a trivial bugfix (it is work, it is appreciated, but it is small). This of course assumes, that you are in a position to donate the code, if you are not, please indicate so.
TL;DR: For code we don't import from a third party (and thus don't own), ensure that you are the author, you donate the code to the ASF, so please stay with the normal ASF headers.

YannLeCorse: It's a stupid bug fix (trivial is too kindly), of course it's not necessary but I've been asked to evaluate our FOSS contribution process. I'll check again with my legal to get rid of it. 

Any comments for the rest ?

YannLeCorse:
Apologies for the removal of trailing spaces which occurs during the removal of the license header modification.
If needed I'll cancel this PR and create a new one. Just let me know.

geertjanw:
Great to have this PR in any way that works best for you and also works in every other sense.

YannLeCorse:
@junichi11 https://netbeans.apache.org/participate/submit-pr.html is saying that we should not touch the PR's after the submission. Should I squash nevertheless or should I cancel it and create a new one (with a single commit) ?

neilcsmith-net:
@YannLeCorse I think we should review that text, as it's in direct opposition to point 10 for a start! :smile: Changes to meet reviews and other requirements should go in the same PR.  I think that point is intended to stop people changing what has already been reviewed or +1'd?

Unless someone has a burning requirement for me to merge for NB11.1 beta today, this will be held back to merge for NB11.2 when that window opens in a couple of weeks anyway.

junichi11:
@YannLeCorse 
> Should I squash nevertheless or should I cancel it and create a new one (with a single commit) ?

No, don't worry :) You can leave this as it is.

That label(Need squashing) is for the person who merges this. 
If a PR has some commits and each commit is meaningful and the review was done, the author should squash properly before it is merged, I suppose. In this PR, we need just one commit, so mergers can squash and merge via the Squash and merge button :)

matthiasblaesing:
No one raised points against this, the change is pretty self contained and looks sane.  Will merge.

PR #1336, opened 2019-06-27
[NETBEANS-2756] Added root project relative license file search.

lkishalmi:


lkishalmi:
@neilcsmith-net Well, not a major improvement, but pretty trivial one. Could be considered for 11.1.

PR #1092, opened 2019-01-21
Using Graal.js 19.0.0 via Scripting in platform/core.network

JaroslavTulach:
Since #1011 there is a generic way to request a `ScriptEngine` via `Scripting` API. This PR brings the benefits of that API abstractions to `platform/core.network` module while making sure the actual engine used for PAC resolution can be selected by the vendor of a NetBeans Platform based application and properly tested. This balances the plugability with the necessary safety needed while resolving PAC configurations.

The security is achieved by the `ALLOWED_PAC_ENGINES` branding API. Each vendor of a NetBeans Platform based application is encouraged to enumerate the one(s) engine that is considered trusted. By default we enlist `Nashorn`, `Graal.js` and `GraalVM:js` - all of them tested and to my best knowledge secure. I trust the `GraalVM:js` engine the most. Why list others? Because the `GraalVM:js` engine  is only provided as an optional module - many NetBeans Platform applications won't have it - then it is natural to fallback to `Nashorn` (also considered secure at the moment).

Should NetBeans IDE re-brand the `ALLOW_PAC_ENGINES` value? Possible. NetBeans IDE is known to carry `GraalVM:js` implementation, so it could allow only its usage while resolving PAC. I haven't done so, but I have written a test in `nb/ide.branding` module to verify that `GraalVM:js` is really selected in the NetBeans IDE.

Why use `Scripting.newBuilder().build()` to create `ScriptEngineManager` at all? Because various vendors may have different opinion on which engine to use. @matthiasblaesing had a preference for Rhino. OracleLabs require [GraalVM](http://graalvm.org) based implementation in [VisualVM](http://visualvm.io) and co. The pluggable nature of the `Scripting` API allows anyone to hook the preferred engine of choice into own NetBeans Platform based application and then select it for PAC resolution using `ALLOWED_PAC_ENGINES` branding API.

To achieve the needed safety levels, this PR also upgrades from RC versions of GraalVM libraries to supported 19.0.0 release. Unlike RCs, this version is considered stable. I'd rather see next version of NetBeans using stable version of GraalVM libraries.

PR #1337, opened 2019-06-28
[NETBEANS-2764] NoClassDefFoundError: com/oracle/truffle/api/library/Library

JaroslavTulach:
@lbruun reported that the PAC proxy configuration is broken after my #1092. Alas, it is. The problem isn't in the code, but in the packaging. One NetBeans wrapper module forget to export new API package `com.oracle.truffle.api.library`.

The fix exports the package and enhances the tests to verify that the `GraalVM:js` is capable to execute JavaScript even inside of the NetBeans Runtime Container with all the ClassLoader checks being on. 

Sorry for the trouble.

### DIFF
--- a/webcommon/libs.truffleapi/nbproject/project.xml
+++ b/webcommon/libs.truffleapi/nbproject/project.xml
@@ -41,6 +41,7 @@
                 <package>com.oracle.truffle.api.frame</package>
                 <package>com.oracle.truffle.api.instrumentation</package>
                 <package>com.oracle.truffle.api.interop</package>
+                <package>com.oracle.truffle.api.library</package>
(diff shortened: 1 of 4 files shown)


webcommon/libs.truffleapi/nbproject/project.xml
```diff
@@ -41,6 +41,7 @@
                 <package>com.oracle.truffle.api.frame</package>
                 <package>com.oracle.truffle.api.instrumentation</package>
                 <package>com.oracle.truffle.api.interop</package>
+                <package>com.oracle.truffle.api.library</package>
```
JaroslavTulach: This was the missing package.

PR #1339, opened 2019-06-29
[NETBEANS-2762]: Ensure that if any Java code template parameters nee…

jlahoda:
…ds local vars, the locals are computed.

Without this patch, it may happen that a code template parameter which needs parsing but not local variables is found first, parsing is inited without the locals, and the locals are then missing for the future parameters which need them. See:
https://issues.apache.org/jira/browse/NETBEANS-2762

e.g. the "fore" code template.


geertjanw:
Wow, great you worked on this. But why did this work in 8.2 without this fix? We should definitely include this in 11.1.

jlahoda:
There was this optimalization done after 8.2:
http://hg.netbeans.org/main/rev/2d82f8c0a290

But, sadly, it goes too far, and disables locals computation even in cases where it shouldn't, IMO, be disabled.

PR #1349, opened 2019-07-02
[NETBEANS-2785] Move NIO2 File Watcher implementation into the front.

lkishalmi:
Well, this change makes the existing NIO2 based file notifier implementation in fornt of the platform dependent ones (Positions: Windows: 100, MacOS: 300, Linux: 500)
However this is trivial change, I feel a bit late for 11.1, so I propose to merge it into master after the 11.1 is out.
If it works we can get rid of the platform specific implementations in NB 12.0

### DIFF
--- a/platform/masterfs.macosx/src/org/netbeans/modules/masterfs/watcher/macosx/OSXNotifier.java
+++ b/platform/masterfs.macosx/src/org/netbeans/modules/masterfs/watcher/macosx/OSXNotifier.java
@@ -41,7 +41,7 @@ import org.openide.util.lookup.ServiceProvider;
  *
  * @author Tomas Zezula
  */
-@ServiceProvider(service=Notifier.class, position=300)
+@ServiceProvider(service=Notifier.class, position=100)
 public final class OSXNotifier extends Notifier<Void> {
     private static final Level DEBUG_LOG_LEVEL = Level.FINE;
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
@lkishalmi has anyone tested this on macOS?  AFAIK macOS still doesn't have a real implementation of NIO2 watchers.  I ship my own stuff without the native handlers and not noticed a problem, but there tend to be a lot less files in play at any one time there.

lkishalmi:
Well, indeed MacOS still uses the Poller implementation: https://hg.openjdk.java.net/jdk/jdk/file/19d0b382f086/src/java.base/macosx/classes/sun/nio/fs/BsdFileSystem.java
Thanks for pointing this out. I'm going to realign the positions according to that.

lkishalmi:
Changed the order to the following:
- 100: MacOSX
- 200: NIO2
- 400: Windows
- 500: Linux

lkishalmi:
@tzezula Any opinion on this? I'd like to introduce this one early in 11.2 so if it would cause problems we can detect/revert the change before the release process would really spin up for 11.2.

lkishalmi:
@geertjanw @JaroslavTulach shall we move this further. It is pretty easy to revert in case we need that. I'd like to see this in the Thursday's netbeans-dev snap drop.

PR #1350, opened 2019-07-02
[NETBEANS-2727] Update the image of the "Pause I/O Checks" toolbar bu…

geertjanw:
…tton

In the main toolbar, the old NetBeans blue icon is very visible, by means of this PR, the image for pausing the profiling of Apache NetBeans itself is updated to make use of a new image that makes use of the new Apache NetBeans logo.


geertjanw:
Since 11.1 is the first release of NetBeans outside the Apache incubator, would be cool if the old blue NetBeans icon were to be replaced in the toolbar with the new Apache NetBeans icon.

PR #1351, opened 2019-07-02
[NETBEANS-2727] New images for Start/Stop Self Profiling of IDE

geertjanw:
Use new Apache NetBeans icon instead of old blue NetBeans icon in toolbar for start/stop self-profiling of IDE.


geertjanw:
Since 11.1 is the first release of NetBeans outside the Apache incubator, would be cool if the old blue NetBeans icon were to be replaced in the toolbar with the new Apache NetBeans icon.

PR #1352, opened 2019-07-02
[NETBEANS-2240] (partial) Switch DeferredComplationFailureHandler to …

jlahoda:
…javac move before running JavacTaskImpl.enter(), to ensure CompletionFailures are handled properly.

If the errors are not handled properly during enter, they then cause attr to be skipped, leading to (among other things) the NPE from PostFlowAnalysis->Type.erasure. When javac is doing something, the DeferredComplationFailureHandler should be in the "javac" mode. Normally javac's API methods will switch the mode, but JavacTaskImpl.enter is not an API method, so it is not doing the switch.


neilcsmith-net:
@jlahoda @geertjanw Should we aim to get this in NB11.1?

PR #1355, opened 2019-07-04
NP-20 Disable Payara Micro module

jGauravGupta:


neilcsmith-net:
See #1356 - hopefully good to go for release, so closing this.  Note that a similar change will be required in master once we've opened the merge window for NB 11.2.

PR #1356, opened 2019-07-05
Add Payara Micro dependency in j2ee.kit

neilcsmith-net:
It seems that ergonomics is not triggering enabling of the Payara Micro module and ends up in some sort of loop in the wizard iterator.  This adds the Payara Micro module as a dependency in j2ee.kit.  It seems to fix the issue, but I'm not sure it's the right approach to this problem?  PR against release branch for now in case we want to use this in NB 11.1 and fix differently in master.  Can re-roll against master if required.

### DIFF
--- a/enterprise/j2ee.kit/nbproject/project.xml
+++ b/enterprise/j2ee.kit/nbproject/project.xml
@@ -95,6 +95,12 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.modules.payara.micro</code-name-base>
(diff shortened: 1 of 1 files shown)


GeertjanWielenga:
Connecting this to j2ee.kit makes sense, especially if it works.

neilcsmith-net:
Considered re-rolling against master but until we fix the plugin center changes in master post-release it doesn't work there to test anyway.  Once we decide what we do in master for dev plugin centers we should consider adding this change there.  So, keeping to release branch only for now.

PR #1358, opened 2019-07-06
Get OpenJFX samples working in release branch.

neilcsmith-net:
Add OpenJFX samples module to JavaFX cluster, and add as dependency to javafx2.kit to enable via ergonomics.  Again against release branch - review and add to master post-release when plugin centers for dev are fixed.

### DIFF
--- a/javafx/javafx2.kit/nbproject/project.xml
+++ b/javafx/javafx2.kit/nbproject/project.xml
@@ -80,6 +80,12 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.modules.openjfx.samples</code-name-base>
+                    <run-dependency>
+                        <specification-version>1.0</specification-version>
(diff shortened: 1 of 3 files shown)


geertjanw:
This has not been put into master, can you do this @neilcsmith-net, or tell me how?

neilcsmith-net:
@geertjanw probably easiest to make a clean PR with the same changes for master, although ideally we need to check this really does all required.  I can do it, but not for a few days.  I didn't merge through master at the time because the update centres were broken and therefore more difficult to test.

geertjanw:
Looking at it some more, and advised by @jtulach, ultimately would be best to delete the new openjfx.samples module completely and replace with tw @TemplateRegistrations here:

https://github.com/apache/netbeans/blob/master/java/maven/src/org/netbeans/modules/maven/newproject/MavenWizardIterator.java

PR #1359, opened 2019-07-06
[NETBEANS-2792] Close IO only when all registered Gradle Action Hook finished.

lkishalmi:
A safe thing to be included in 11.1.


lkishalmi:
Though it is a minor and might be cosmetic only.
https://issues.apache.org/jira/browse/NETBEANS-2792

PR #1363, opened 2019-07-08
 [NETBEANS-1696] PHP 7.4 Support Arrow Functions 2.0

junichi11:
## Arrow Functions 2.0

See https://wiki.php.net/rfc/arrow_functions_v2

nb-php74-arrow-functions

### 1st commit

- Fix ASTPHP5Parser.cup, ASTPHP5Scanner.flex, and Php5ColoringScanner.flex i.e. parser and lexers
- Add ArrowFunctionDeclaration as an ASTNode
- Add ArrowFunctionScope
- Add code templates for arrow functions
- Fix PHP74UnhandledError and UnusableTypesUnhandledError
- Fix GotoDeclaration and MarkOccurrences features
- Add ASTErrorExpression node to avoid that an ArrowFunctionDeclaration node is handled as an ASTError
- Add unit tests

### 2nd commit

 - Fix the formatting feature

### 3rd commit

- Fix the code completion feature

### the others

Missing changes for `__serialize` and `__unserialize`

### About ASTErrorExpression node

e.g. these are syntax errors
```php
fn($x) => return $x;
fn(int $number) =>;
```

The parser result(the first case) is like the following:
<details>
<summary>parser result: before</summary>

```
    <Program start='0' end='836'>
        <Statements>
            <ASTError start='815' end='824'/>
            <ReturnStatement start='825' end='835'>
                <Variable start='832' end='834' isDollared='true'>
                    <Identifier start='833' end='834' name='x'/>
                </Variable>
            </ReturnStatement>
        </Statements>
    </Program>
```
</details>

The first case is no problem because it is a complete statement and it is a syntax error. But the second case has a problem. It is an incomplete statement(i.e. a user is typing...). That part is handled as an ASTError. So, the parameter is not scanned in the `ModelVisitor`. i.e. `$number` is not added to code completion items.

I would like to avoid that. So I've added `ASTErrorExpression` as an `ASTNode`(see `ASTPHP5Parser.cup`). That way, we can scan parameters even if there is a syntax error in an expression part of `ArrowFunctionDeclaration`.

Parser result after it is added:
<details>
<summary>parser result: after</summary>

```
    <Program start='0' end='836'>
        <Statements>
            <ExpressionStatement start='815' end='835'>
                <ArrowFunctionDeclaration start='815' end='824' isReference='false' isStatic='false'>
                    <FormalParameter start='818' end='820' isMandatory='true' isVariadic='false'>
                        <ParametrType>
                        </ParametrType>
                        <ParametrName>
                            <Variable start='818' end='820' isDollared='true'>
                                <Identifier start='819' end='820' name='x'/>
                            </Variable>
                        </ParametrName>
                        <DefaultValue>
                        </DefaultValue>
                    </FormalParameter>
                    <ASTErrorExpression start='822' end='824'/>
                </ArrowFunctionDeclaration>
            </ExpressionStatement>
        </Statements>
    </Program>
```
</details>

## Not Yet

- Hint features e.g. UnusedVariablesHint


php/php.editor/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletion.java
```diff
@@ -1548,8 +1553,23 @@ private void autoCompleteConstants(final PHPCompletionResult completionResult, P
...
                 }
             }
-            Collection<? extends VariableName> declaredVariables = ModelUtils.filter(variableScope.getDeclaredVariables(),
-                    nameKind, request.prefix);
+
+            List<VariableName> allDeclaredVariables = new ArrayList<>(variableScope.getDeclaredVariables());
+            // for nested arrow functions
+            if (variableScope instanceof ArrowFunctionScope) {
+                Scope inScope = variableScope.getInScope();
+                while (inScope != null
```
tmysik: The `null` check is not needed, `instanceof` always returns `false` for `null`.

junichi11: You are right. I'll fix it with the next changes. Thanks!

junichi11:
Yes, it became a huge patch since I added many unit tests (There are about 220 files related to tests)... Thank you!

PR #1364, opened 2019-07-08
@deprecation in the comment section , without an @Deprecation annotat…

BradWalker:
There are over 100 of the following errors..

{quote}   [repeat] /home/bwalker/netbeans/enterprise/payara.common/src/org/netbeans/modules/payara/common/ui/IpComboBox.java:180: warning: [dep-ann] deprecated item is not annotated with @Deprecated
   [repeat]     public IpComboBox(final Object items[])
   [repeat]               ^ {quote}  

Added "@Deprecated" annotation to remove the warnings.. This  does not effect the deprecation that will need to be addressed.


BradWalker:
Thanks very much Matthias.. The only reason that I reordered the "deprecated" comment was to make it more "consistent" with industry practice. Usually the deprecated comments are the last thing.. But, if you prefer, in the future I can refrain from doing so.

matthiasblaesing:
Merged - thank you.

PR #1368, opened 2019-07-11
Let Gradle Frontend Application run on JDK11

dukescript:
While testing the NetBeans 11.1 vc1 it turned out that Gradle Frontend Application project cannot be executed on JDK11, only on JDK8. The workaround is to run the IDE on JDK8, or change the project's JDK to JDK8, but given that many users use newer JDKs, having the support only on JDK8 would cause a lot of confusion when trying this new great NetBeans feature!

If possible, let's backport this to 11.1 next voting candidate (if any). CCing @neilcsmith-net , @lkishalmi 

The steps to reproduce the error are:
* start NetBeans 11.1 vc1 with an empty user directory and on JDK11: `/netbeans-11.1-vc1/bin/netbeans --userdir /tmp/ud11 --jdkhome /jdk-11`
* New Project/Java with Gradle/Java Frontend Application
* Download and Activate - installs NbJavac and JavaFX wrapper modules
* Generate the project somewhere
* Build the project - fails on `Task: compileRetrolambdaMain FAILED` - d46c11b fixes that
* Debug the project - fails with `NoClassDefFoundError: javafx/application/Application` - 1b5db1e fixes that
* Run the project - works but prints exception `IllegalArgumentException at net.orfjackal.retrolambda.asm.ClassReader` -  f3095c1 fixes that.

Sorry for being so late with this bugfix. At the time of creation of the skeleton project the retrolamda plugin 3.7.1 wasn't yet available and  support for JDK11 wasn't possible at all. Now it would be unfortunate to not have it in NetBeans 11.1. Please consider backporting it. Thanks for your consideration.

### DIFF
--- a/groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/desktop_build.gradle.fmk
+++ b/groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/desktop_build.gradle.fmk
@@ -18,10 +18,14 @@
     under the License.
 
 -->
-apply plugin: 'java'
-apply plugin: 'application'
+plugins {
+    id 'java'
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.7' apply false
+}
 
 def commonProject = project.parent
+def jdk8 = System.getProperty("java.version").startsWith("1.8")
 
 mainClassName = '${packageBase}.DesktopMain'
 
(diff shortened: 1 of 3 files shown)


geertjanw:
Unless there is a new vc, I suggest we remember that 11.2 is not far away and that this is not an LTS, so the fact only JDK 8 is supported for this is not a blocker for 11.1.

neilcsmith-net:
There are also various issues around the Maven Frontend Application wizard (eg. https://issues.apache.org/jira/browse/NETBEANS-2439 and others) All could do with looking at during NB 11.2 window, and could consider pushing via AUC to NB 11.1 if feasible.

geertjanw:
Yes, providing the various enhancements Toni is working on could be done via the update center, like Laszlo did for Gradle support during 11.0. Would that work, Toni? Trying to find a balance here between our various needs and wants. :-)

dukescript:
Now, when there is a chance for 11.1 VC2 due to NETBEANS-2824: Can you @neilcsmith-net consider including this fix as well, please?

neilcsmith-net:
@dukescript you're in luck! :smile: I'll merge for beta4.  Please make sure all is working correctly in it when ready before I roll vc2. Thanks!

PR #1369, opened 2019-07-11
[NETBEANS-2343]: Only use the LSP HyperlinkProvider for files that ar…

jlahoda:
…e covered by some LSPBindings.

(Avoid highlighting everything as a hyperlink.)


neilcsmith-net:
@jlahoda @geertjanw considering we're doing another NB 11.1 beta and this looks a minimal but useful fix, shall we merge?

GeertjanWielenga:
Good plan, I agree.

PR #1370, opened 2019-07-11
[NETBEANS-2831] Removed accidentally left in CDDL_GPL License headers.

lkishalmi:


neilcsmith-net:
Thanks!  A sanity check on beta4 sources when ready to make sure we've caught them all would be appreciated.

PR #1376, opened 2019-07-12
[NETBEANS-2700] Let `ant gen-sigtests-release` pass!

JaroslavTulach:
As reported by @neilcsmith-net, the `ant gen-sigtests-release` is currently broken. Here are my changes that shall fix it. Plus I've included the check into Travis job, so next time we hopefully avoid problems like this.


JaroslavTulach:
It would be better to get this into 11.1 as then the steps described in [NETBEANS-2700](https://issues.apache.org/jira/browse/NETBEANS-2700) will be "followable".

neilcsmith-net:
It will be.  Should we also commit the sigtest files when they're generated into the release branch as well as master?

JaroslavTulach:
> Should we also commit the sigtest files when they're generated into the release branch as well as master?

We haven't done that previously as far as I remember, but possible and it makes sense.

PR #1377, opened 2019-07-12
[NETBEANS-2839] - cleanup Class raw type..

BradWalker:
Clean up a bunch of warnings related to use of a raw Class type..

Basically I just changed "Class" to "Class<?>"..

### DIFF
--- a/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassLoaderSupport.java
+++ b/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassLoaderSupport.java
@@ -90,7 +90,7 @@ final class ClassLoaderSupport extends URLClassLoader implements FileChangeListe
      * @throws ClassNotFoundException
      */
     @Override
-    protected Class findClass (String name) throws ClassNotFoundException {
+    protected Class<?> findClass (String name) throws ClassNotFoundException {
         Class c = super.findClass (name);
         if (c != null) {
             org.openide.filesystems.FileObject fo;
--- a/java/performance/actionsframework/src/org/netbeans/actions/simple/SimpleInvoker.java
+++ b/java/performance/actionsframework/src/org/netbeans/actions/simple/SimpleInvoker.java
@@ -49,7 +49,7 @@ class SimpleInvoker {
         return name;
     }
 
-    private Class clazz = null;
+    privateClass<?> clazz = null;
     private Class getTargetClass() throws ClassNotFoundException {
         if (clazz == null) {
             clazz = Class.forName(targetClass);
@@ -72,7 +72,7 @@ class SimpleInvoker {
             if (isDirect) {
                 getTargetMethod().invoke(null, null);
             } else {
-                Class clazz = getTargetClass();
+               Class<?> clazz = getTargetClass();
                 Object o = context.get(clazz);
                 if (o == null) {
                     throw new NullPointerException ("No instance of " + clazz + " in context");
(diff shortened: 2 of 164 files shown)


ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassLoaderSupport.java
```diff
@@ -90,7 +90,7 @@ private ClassLoaderSupport (final ClassPath cp, final ClassLoader parentClassLoa
      * @throws ClassNotFoundException
      */
     @Override
-    protected Class findClass (String name) throws ClassNotFoundException {
```
matthiasblaesing: Please don't change signature from raw type to generic type without a clear plan. This can introduce compatibility issues (I suffered from it in another project).

profiler/lib.profiler/src/org/netbeans/lib/profiler/server/ClassLoaderManager.java
```diff
@@ -431,7 +431,7 @@ private Class getLoadedClass(String className) {
     }
 
     private Class getLoadedClassInThisLoaderOnly(String className) {
-        Class clazz = null;
+       Class<?> clazz = null;
```
matthiasblaesing: I think this is a general problem: you changed white-space (see removed space).

java/performance/actionsframework/src/org/netbeans/actions/simple/SimpleInvoker.java
```diff
@@ -49,7 +49,7 @@ public String getName() {
         return name;
     }
 
-    private Class clazz = null;
+    privateClass<?> clazz = null;
```
neilcsmith-net: Huh?! Surely no longer valid code.

ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassLoaderSupport.java
neilcsmith-net: @matthiasblaesing ah, yes, thanks for that reminder! :smile:  From me, -1 to this PR as is.  Maybe split into two?  A PR changing internal method variables should be fairly safe?  Then another for considering method arguments and return values where appropriate / with a plan?

BradWalker: So it sounds like I should just split this into 2 changes.. 1 - method variables and 2 - method signature.. Agreed that #2 needs some planning and thought but #1 should be pretty strait forward..

Correct?

java/performance/actionsframework/src/org/netbeans/actions/simple/SimpleInvoker.java
BradWalker: I always do an "ant clean ; ant" build before I submit a pull-request. FYI. This file was not getting compiled so this error was never seen.

What's happening is this file is only built as part of the qa-functional tests. So it was not seen..

ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassLoaderSupport.java
matthiasblaesing: Yes, splitting this and maybe just doing the non-API parts right now sound like a good plan.

BradWalker:
This will be fixed in a later git commit..

PR #1380, opened 2019-07-14
Update htmlparser/lexer to support html5 attribute syntax

matthiasblaesing:
With HTML5 almost any character is allowed in an attribute name and angular is actively using this. With angular attribute names like this are common:

- <input [value]='test' /> (property binding)
- <input (change)='test' /> (event binding)
- <input [(value)]='test' /> (twoway binding)
- <input #ref /> (template reference)
- <input *ngIf='test' /> (structural directive)

This changeset updates the html.parser and html.lexer modules to support the syntax.

It is not the intention of this changeset to change the html validator. The validator for example will still (rightly) complain, that the angular attribute names are not serializeable as XML.


matthiasblaesing:
The upload of the new htmlparser binary succeeded and travis is happy.

@Chris2011 could you please have a look at the changeset. I get the feeling, that you might be interested in it anyway.

Chris2011:
Had a look and it seems good to me. Thx for implementing it :). I saw your comment about the validator, so the HTML file will still underline all attributes that are not "correct"?

matthiasblaesing:
Yes, if you don't like these hints, open the options dialog, the editor category, hints tab and uncheck "HTML Validator -> Attributes".

PR #1382, opened 2019-07-15
Fix NPE if PAC cannot be resolved

akuhtz:
Fix NPE if PAC cannot be resolved, e.g. UnknownHostException if the proxy pac url is configured but not reachable.


matthiasblaesing:
I'm not sure if it would be better to retry initializing the proxy would also be a good idea, but I see, that this makes more sense then exploding with an NPE. Thank you.

PR #1383, opened 2019-07-16
Reduce default combo box model warnings

BradWalker:
I've tried to reduce the number of warnings related to the use of DefaultComboBox()..

I did this by providing a proper type to the variable declaration..


junichi11:
We have to change types of guarded sections with a GUI editor (Properties pain)
nb-pr-1383

BradWalker:
This pull request needs a little bit more work by me.. So rather than leave it open, I'm going to close it. I will re-open in the future..

PR #1385, opened 2019-07-17
Cleanup raw class type

BradWalker:
I've tried to only cleanup some of the  variables that use a raw Class type.


BradWalker:
bwalker@brads-T460s:~/netbeans$ find ./ -type f -name "*java" -exec egrep -l " Class clazz =" {} \; | xargs sed -i 's/ Class clazz =/ Class<?> clazz =/g'
bwalker@brads-T460s:~/netbeans$

This should only pick up the variables that use a Class raw type. And not methods that might be doing the same thing. So no signature changes needed..

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/ide/commands/WildflyManagementAPI.java
```diff
@@ -119,7 +119,7 @@ static Object createOperation(WildflyDeploymentFactory.WildFlyClassLoader cl, Ob
     // ModelNode
     static Object createReadResourceOperation(WildflyDeploymentFactory.WildFlyClassLoader cl, Object modelNode, boolean recursive)
             throws ClassNotFoundException, NoSuchMethodException, IllegalAccessException, InvocationTargetException {
-        Class clazz = cl.loadClass("org.jboss.as.controller.client.helpers.Operations"); // NOI18N
+        Class<?> clazz = cl.loadClass("org.jboss.as.controller.client.helpers.Operations"); // NOI18N
         Class modelClazz = cl.loadClass("org.jboss.dmr.ModelNode"); // NOI18N
```
matthiasblaesing: This looks strange - why was line 122 changed, but 123 not? This seems to be the case at multiple places.

BradWalker: This is a result of the egrep cmd only picking up instances of "Class clazz".. This seemed like the easiest thing to do to reduce the warning message and do no harm at the same time.

Hopefully, I can reduce more of these messages and make the code review easy by doing more of this.. I really don't like creating a big code review because it makes more work for you and others.

matthiasblaesing:
I'm not sure if I'm happy with this changeset. The warnings are there for a reason, mechanical replacement with the most open generic, plasters over these warnings. This raises the question why they exist in the first place.

BradWalker:
I understand and mostly agree with your assessment about this change. Let's take this piece of code for discussion before my change:

`Class clazz = cl.loadClass("org.jboss.as.controller.client.helpers.Operations");`

So what Netbeans is doing is loading a Class and removing the type information by setting it to a raw type. 

My change simply maintains that type information. I know it's a pretty generic statement. But, the Java Language Standard, sec. 4.8 says the following:

> The use of raw types is allowed only as a concession to compatibility of legacy code. The use of raw types in code written after the introduction of generics into the Java programming language is strongly discouraged. It is possible that future versions of the Java programming language will disallow the use of raw types. 

Why does this code exist? My rookie opinion is this code is "old" and therefore was never designed to work with generics in the first place. 

Also, given the limited resources I'm not sure things like this will ever be fixed properly. There are simply more pressing needs.

As we fix more of these raw type errors, it should make us more diligent in terms of creating better code.

matthiasblaesing:
Ok - as these `Class` declarations are not in the signatures, tighter bounds can be added later. Will merge. Thank you.

PR #1387, opened 2019-07-22
a plus sign too much

ramiswailem:
was removed


matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/984954778cc724cfe37e4c095034d3a45b100f19 - I updated the commit message, so that it is clearer why the change is necessary.

PR #173, opened 2017-10-21
[NETBEANS-97] Don't add braces around try/synchronized blocks when reformatting

bondolo:
Extraneous braces will also be removed according to user preference

PR #1392, opened 2019-07-22
[NETBEANS-97] Don't add braces around try/synchronized blocks when reformatting

bondolo:
Extraneous braces will also be removed according to user preference

See previous PR which I was unable to update for context:

https://github.com/apache/netbeans/pull/173

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/Braces.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/Braces.java
@@ -54,19 +54,19 @@ import org.openide.util.NbBundle;
 public class Braces {
 
     static final EnumSet<JavaTokenId> nonRelevant = EnumSet.<JavaTokenId>of(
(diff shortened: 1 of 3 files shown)


bondolo:
ping. Now includes requested unit tests

matthiasblaesing:
@lbruun @sdedic could you please have a look at this and see if the concerns raised on the original PR are fixed?

geertjanw:
There doesn't seem to be a reason not to merge this, though maybe you can provide some screenshots here to show everything that this adds on the user interface level and what this looks like in the editor?

geertjanw:
The reason I ask this is that I don't know right now exactly what to put in the New & Noteworthy for this one. My understanding is that if a certain new option is set in netbeans.conf, then a specific kind of formatting will take place.

geertjanw:
OK, we really need documentation here, but clearly everything that has been asked for has been provided and it's time to merge this now.

PR #1396, opened 2019-07-23
[NETBEANS-2878]: Compile the probe for J2SE platforms on the target p…

jlahoda:
…latform itself, to avoid issues when compiling on the java.j2seplatform module on JDKs that do not support -target 1.1.

This is a little bit slower, of course, but hopefully more reliable (assuming the ability to build NetBeans on JDK 11 is worth it) - using the javac from the target should ensure the outcome will be runnable on the target.


lkishalmi:
Really nice solution! +1

PR #1399, opened 2019-07-24
[NETBEANS-2880] Add a L&F property to define the ellipsis icon

YannLeCorse:


YannLeCorse:
Thanks for the review, I'll add them asap.

Sorry for the trailing spaces, this comes from the settings we use at office... Is there a NB settings that we can apply on the IDE to ensure that code matches standards ? (i.e. keep trailing spaces, don't organize imports, etc.)

matthiasblaesing:
Removing trailing spaces is a sensible decision. It requires however, that the projects are already clean, else you introduces changes all over the place, that you did not intent. Two things to prevent that:

1.) I assume you have "remove trailing space from" option activated in netbeans, I have that too, but I use "Modified lines only". That way, code you edited gets corrected, everything else is left untouched.
2.) Review commits prior to push

Sorry I don't think there is a silver bullet to tackle the whitespace problem easily.

matthiasblaesing:
Merged via: https://github.com/apache/netbeans/commit/f0d26e4294bf2090a4a705a3317120504d932572

PR #1403, opened 2019-07-28
[NETBEANS-1802] Basic infrastructure to load a project with a Project Type

lkishalmi:
Well this would be a possible start of a long journey. To make folders which has multiple build flavors be loaded as a specific project. Please check the code it should be backward compatible.

### DIFF
--- a/ide/projectapi.nb/nbproject/project.properties
+++ b/ide/projectapi.nb/nbproject/project.properties
@@ -15,5 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 is.eager=true
-javac.source=1.6
+javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial
--- a/ide/projectapi.nb/src/org/netbeans/modules/projectapi/nb/NbProjectManager.java
+++ b/ide/projectapi.nb/src/org/netbeans/modules/projectapi/nb/NbProjectManager.java
@@ -23,10 +23,12 @@ import java.awt.EventQueue;
 import java.io.IOException;
 import java.lang.ref.Reference;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
@@ -42,6 +44,7 @@ import org.netbeans.api.project.ProjectManager.Result;
 import org.netbeans.spi.project.ProjectFactory;
 import org.netbeans.spi.project.ProjectFactory2;
 import org.netbeans.spi.project.ProjectManagerImplementation;
+import org.netbeans.spi.project.ProjectManagerImplementation2;
 import org.netbeans.spi.project.ProjectState;
 import org.openide.filesystems.FileChangeAdapter;
 import org.openide.filesystems.FileChangeListener;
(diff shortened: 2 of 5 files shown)


JaroslavTulach:
Yes, with the growing number of cases where people mix various build technologies into the same directories, it is becoming more and more clear that the NetBeans system of one project type per directory doesn't fit many real world projects anymore.

My friend told me two months ago about his struggle with a project combining `package.json` and `build.gradle`. I wanted to help him with a "Multi Project" module. The idea was to create a decorator `Project` implementation that would delegate to all the other `Project` implementations that can physically recognize and handle the same directory. When user invokes build - it would invoke build via NPM and also via Gradle. The logical view would be merged. Queries like classpath would be delegated to both projects and their result merged as well.

Of course, this is not always what the user wants, so I was thinking of a UI that would allow the user to select the (set of) project(s) to delegate to. E.g. user could choose that only Gradle is preferred and the IDE should only invoke Gradle actions. Later, the user could switch to NPM view and work with the project only from the node.js perspective, completely ignoring Gradle.

Such "Multi Project" would implement "natures" in a reasonably natural way. "natures" - e.g. the current `Project` implementations would completely care about their physical representation on disk and physical commands to execute. The "Multi Project" would be just the UI to control & delegate to the "natures".

As far as I can tell @lkishalmi attempt is heading in the same direction. Just it starts with API change, I wanted to prototype the solution from outside. Solving the problem may turn into significant effort and having the ability to offer it as an extension module could help us get some user feedback before we change such a core functionality of the IDE. Ideally, once we succeed, all `Project` instances obtained from `ProjectManager` would be instances of the "multi project".

Btw. let's CC @jglick - he introduced the `ProjectManager` and fought hard against "natures".

jlahoda:
I am afraid I am not sure how exactly this patch should help? Projects are often through the FileOwnerQuery, which does not know currently which type of project it is trying to load.

When I was thinking of this, I was imagining two possible approaches:
1. a Multi Project similar to what @jtulach was proposing. Could work when the two projects merged into the Multi Project are sufficiently different (like possibly Java Gradle and NPM). Unlikely to work for two Java projects, but that is presumably fine. Maybe we would have a Multi Project that would allow only such combinations of projects that are known to work. This could be prototyped as an extension module.
2. have an action that would tell the project infra to recognize the given directory as a project of a particular type, and ignore all rest. Probably not reasonably doable as an extension module, but possibly also simpler to do.

I think that the code here might be to some degree useful for either of these, but details on what is needed will likely be more clear only when the actual UI so this is implemented.

lkishalmi:
Well, this patch is just a very first step. I'm planning to add some others down on the road, I just would like it to be visible from the start, to be able to talk this through as a process not just one big change.
@JaroslavTulach Unfortunately my direction is not that far reaching. Having general project with natures seems to be a compelling idea, though that would be a big step to jump. I'm not sure if I can make that jump happen. Though if anyone could deliver that, I would happily ditch this set of patches in favor of that. So I'm not trying to reaching that high.
So my plan is to try something built on the current ProjectFactory2 and ProjectManager.Result. My steps would be:
1. Enhance the PM API to be able to query/load Project "natures" recongnized by the IDE.
1. Improve the isProject2 usage to actually return project type
1. Add an Open Project As action if more than one project "nature" is detected on a folder.
1. Fix: There could be some glitches around re-opening the projects.
1. Probably add more UI entry for the project "nature" selection

I have something half working on 3., but need to do some real work on 2.

Chris2011:
What is the progress of this issue here and how can I help here?

lkishalmi:
Well, it is hard to say. It requires some dedication and coordination. Right now we have a ProjectFactory2 interface to implemented by project providers, but most projects do not even implement ProjectFactory correctly. This is easily fixable for Maven and Gradle projects, but the others needs some serious attention.

PR #1404, opened 2019-08-03
[NETBEANS-2925] Possible solution to not to filter out sources from the buildDir

lkishalmi:
[NETBEANS-2925](https://issues.apache.org/jira/browse/NETBEANS-2925)

Well, I need some advice here. Returning mixed on the contents of build dir is probably not the best thing to do. Shall I initiate a different package view with different filter for generated sources instead?

### DIFF
--- a/groovy/gradle.java/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImpl.java
+++ b/groovy/gradle.java/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImpl.java
@@ -38,15 +38,18 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import javax.swing.Icon;
 import javax.swing.event.ChangeListener;
 import org.netbeans.api.java.project.JavaProjectConstants;
+import org.netbeans.api.project.FileOwnerQuery;
 import org.netbeans.api.project.Project;
+import org.netbeans.api.project.ProjectManager;
 import org.netbeans.api.project.ProjectUtils;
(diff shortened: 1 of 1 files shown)


jglick:
I am not sure what you are trying to do but I am pretty sure this is not the right way to do it.

The `maven` module displays package roots for any `build/generated-{,test-}sources/*/` it finds. The fact that these are `NOT_SHARABLE` is irrelevant.

lkishalmi:
Thanks, Jesse!
Checked the Maven implementation and made something similar in Gradle as well.

groovy/gradle.java/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImpl.java
```diff
@@ -162,13 +164,7 @@ public GradleSourcesImpl(Project proj) {
...
                 }
             }
-            Collections.sort(ret, new Comparator<SourceGroup>() {
-
-                @Override
-                public int compare(SourceGroup o1, SourceGroup o2) {
-                    return o1.getName().compareTo(o2.getName());
-                }
-            });
+            Collections.sort(ret, (SourceGroup o1, SourceGroup o2) -> o1.getName().compareTo(o2.getName()));
```
jglick:
```suggestion
            Collections.sort(ret, Comparator.comparing(SourceGroup::getName));
```

or something like that (look it up).

lkishalmi: Thanks! Learn a new thing every day!
The test will be a new PR as it is not that trivial to create one for this.

PR #1406, opened 2019-08-06
[NETBEANS-2962] Make pinned snap startable in KDE's Plasma

trixon:
Fix https://issues.apache.org/jira/projects/NETBEANS/issues/NETBEANS-2962?filter=allissues

### DIFF
--- a/nbbuild/packaging/netbeans_snap/snap/gui/netbeans.desktop
+++ b/nbbuild/packaging/netbeans_snap/snap/gui/netbeans.desktop
@@ -23,4 +23,4 @@ Exec=netbeans %F
 Categories=Development;IDE
 Icon=${SNAP}/meta/gui/icon.png
 Terminal=false
-
+StartupWMClass=Apache NetBeans 11.1


lkishalmi:
Also you can test your changes, just clone the repo, go into the netbeans_snap or netbeans-dev_snap directory and execute snapcraft to build that snap. when it is ready you can test it with:
snap install netbeans....snap --dangerous --classic
Also the dev version is built and uploaded on every Thursday so your changes would be there for test for "broader" audiance: snap install netbeans-dev --channel edge --classic

neilcsmith-net:
> I'd remove the version number from the end.

@lkishalmi but it won't work in that case! Looks like we might need to reconsider how the following code works, or find a way to template along with other uses of the version numbers in the build?

https://github.com/apache/netbeans/blob/master/platform/core.windows/src/org/netbeans/core/windows/view/ui/MainWindow.java#L123
https://netbeans.org/bugzilla/show_bug.cgi?id=183739

trixon:
> Also you can test your changes, just clone the repo, go into the netbeans_snap or netbeans-dev_snap directory and execute snapcraft to build that snap. when it is ready you can test it with:
> snap install netbeans....snap --dangerous --classic
> Also the dev version is built and uploaded on every Thursday so your changes would be there for test for "broader" audiance: snap install netbeans-dev --channel edge --classic

Thanks @lkishalmi , I'll give it a go and test the snaps locally.

As @neilcsmith-net said, it's really picky about the string content.
I'm using 'xprop' to get the proper value, and as a side note, there used to be 'IDE' in there too.

> WM_CLASS(STRING) = "sun-awt-X11-XFramePeer", "Apache NetBeans IDE 11.0"
> WM_CLASS(STRING) = "sun-awt-X11-XFramePeer", "Apache NetBeans 11.1"

I'm not sure but I think that renaming the desktop file(s) would fix it too, if the snap environment allows that that is.

If this is part of a wider version management, how do I proceed with this PR?

lkishalmi:
Well I've put the StartupVMClass generation in the Snap build process. You can check if it works as designed on revision 14 available on the Snap store latest/edge channel (just a few minutes ago, so you have to manually trigger a refresh or wait a few hours to get in sync.)

PR #1408, opened 2019-08-06
Fix update centre links in master

neilcsmith-net:
- Change master branch UC links to match standard format for server-side redirecting / release templating
- Re-add special nbres based 3rd-party UC (OpenJFX only?)
- Re-add 8.2 centre while still required in releases

As discussed in a wider PR that @ebarboni is working on, the idea is that these links will follow a standard form where `/dev/` is replaced by version (eg. `/11.1/`) automatically in the build process.

Actual targets are currently managed in the `.htaccess` at https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess#L10

Redirects are set to match 11.1  for testing at present, although the redirect for `/nb/updates/dev/` should perhaps be `https://builds.apache.org/job/netbeans-linux/lastSuccessfulBuild/artifact/nbbuild/nbms/`?

We might want to consider adding `?{$netbeans.hash.code}` to the plugin centre link as well in future?

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -25,7 +25,15 @@ OpenIDE-Module-Short-Description=Declares NetBeans autoupdate centers.
 
 Services/AutoupdateType/distribution-update-provider.instance=NetBeans Distribution
 Services/AutoupdateType/pluginportal-update-provider.instance=NetBeans Plugin Portal
+Services/AutoupdateType/82pluginportal-update-provider.instance=NetBeans 8.2 Plugin Portal
+
 #NOI18N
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@ebarboni thanks! Yes, same reason I insisted on having a separate link for plugins in 11.1 despite it directing to same location as 11.0 currently - more flexibility to switch things around as we need.

I think 8.2 is mainly there for C/C++ support now.  But as and until we're sure we're dropping from the next release I think we should keep in master too.  Particularly with the release work you're looking at, the less changes needed between master and release112 the better!

I manually retriggered Travis tests which passed second time around - Travis seems to be having issues with us again today!

jkovalsky:
Looks good to me. I don't think that adding ?{$netbeans.hash.code} brings any benefit, because most users do not check content of the Plugin Portal UC periodically whereas Standard Distro UC is checked for updates automatically by default.

neilcsmith-net:
Thanks!  OK, merging.

PR #1409, opened 2019-08-06
[NETBEANS-2917]: Support for yield in Switch-Expression for JDK-13

arusinha:
Support for Yield with value for Switch Expression in JDK-13.

Broadly below issues will be  addresses as part of yield support
1.  Syntax Highlighting of Yield keyword
2.  Formatting of Yield Tree
3. Autocomplete of Yield Keyword.

This PR aims to cater Yield support needed in JDK 13 as per the below JEP
https://openjdk.java.net/jeps/354

Please refer below issue link for further details.
https://issues.apache.org/jira/browse/NETBEANS-2917


java/java.source.base/src/org/netbeans/modules/java/source/TreeShims.java
```diff
@@ -115,6 +117,21 @@ public static ExpressionTree getValue(BreakTree node) {
         }
     }
 
+    public static ExpressionTree getYieldValue(Tree node) {
+        if (!node.getKind().toString().equals("YIELD")) {
+            return null;
+        }
+        try {
+            Class yieldTreeClass = Class.forName("com.sun.source.tree.YieldTree");
```
rtaneja1: Please add //NOI18N

arusinha:
Had addressed all review comments in last commit

PR #1347, opened 2019-07-02
[NETBEANS-2775] New Java Web Application Project Duplicate Servers

jGauravGupta:
Java EE Server can be registered in 2 ways :

1. Service tab > Servers > Add Server (source code in **Common Server[org.netbeans.modules.server]** module)
2. New Project > Java with Maven > Create Web Application > Add server  (source code in **Java EE Server Registry[org.netbeans.modules.j2eeserver]** module)

In this PR, I have removed the duplicate panel and wizard source code from **Java EE Server Registry** module and reused the wizard source from **Common Server** module.

The removed wizard lists server multiple times based on URI fragment e.g :
- 2 GlassFish entries for URI fragment gfv3ee6 & gfv5ee8 
- 3 Payara entries  for URI fragment pfv3ee6, pfv4ee7, and pfv5ee8

PR #1410, opened 2019-08-06
NP-47 Multiple ServerRegistry support in AddServerInstanceWizard

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta.jc@gmail.com>

The issue caused by https://github.com/apache/netbeans/pull/1347
And relates to https://github.com/apache/netbeans/pull/1312

### DIFF
--- a/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/devmodules/api/ServerManager.java
+++ b/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/devmodules/api/ServerManager.java
@@ -88,8 +88,8 @@ public final class ServerManager {
      */
     public static String showAddServerInstanceWizard(Map<String, String> props) {
         checkDispatchThread();
-        ServerInstance instance = AddServerInstanceWizard.showAddServerInstanceWizard(props);
-        return instance.getProperty(InstanceProperties.URL_ATTR);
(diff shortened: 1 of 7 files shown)


peedeeboy:
Hi @jGauravGupta !  Thanks for looking into this.

Whilst there is now a "Amazon Beanstalk" option in the Add Server wizard when creating a new project, it seems to be opening `org.netbeans.modules.cloud.amazon.ui.AmazonWizardComponent` which creates a new node under Services -> Cloud.

This doesn't create a new instance under Services -> Server, so it is still not possible to create the Project.

I believe that the New Server wizard used to open 'org.netbeans.modules.cloud.amazon.ui.serverplugin.AmazonJ2EEServerWizardComponent' when "Amazon Beanstalk" was selected, which does create a node under Services -> Server.

So the workflow to set up AWS used to be to go to Services -> Cloud -> Add Instance and create the AWS instance including Secret Key etc.

Then Project ->New Project -> Java Web and in the Add Server dialog choose "Amazon Beanstalk" and create a Server that referenced that cloud instance

(It wasn't a great workflow to start with!).  Hope that makes sense?  I can provide some screenshots if needed :)

jGauravGupta:
Thanks, @peedeeboy  your review helped, Closing the PR in favour of https://github.com/apache/netbeans/pull/1412

PR #1413, opened 2019-08-08
[NETBEANS-2910] Suggest using brackets for accessing arrays instead of curly braces

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-2910

See: https://wiki.php.net/rfc/deprecate_curly_braces_array_access

This feature is **disabled** by default because I don't think that many people are using curly braces. 

netbeans-2910-suggest-using-bracket-syntax

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/parser/astnodes/ArrayAccess.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/parser/astnodes/ArrayAccess.java
@@ -19,12 +19,15 @@
 package org.netbeans.modules.php.editor.parser.astnodes;
 
 /**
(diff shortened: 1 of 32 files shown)


junichi11:
Travis CI failed. I'll restart build later.

```
BUILD FAILED
/home/travis/build/apache/netbeans/nbbuild/build.xml:1985: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nbbuild/templates/common.xml:709: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nbbuild/templates/common.xml:702: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/enterprise/web.core/build.xml:25: java.io.IOException: Cannot connect to http://netbeans.apache.org
```

PR #1415, opened 2019-08-08
Post-install appearance of NetBeans on Linux with high dpi display is awful

timboudreau:
After downloading and installing NetBeans 11.1 on Gentoo Linux on an HP Spectre laptop with a 1920x1080 display, I get an IDE using unusably small fonts with bad antialiasing:
https://timboudreau.com/files/screen/08-05-2019_07-03-04.png

Adding just `--fontsize 16` and `-J-Dawt.useSystemAAFontSettings=lcd` makes it usable:
https://timboudreau.com/files/screen/08-05-2019_07-30-32.png

This is a serious problem in that the only way to change the font size globally is in a configuration file that needs to be manually edited - new users are unlikely to even find the documentation of how to set it.  Much more likely they simply decide NetBeans is junk and use a different IDE.

This pull request addresses font size on non-Mac and non-Windows operating systems by using the graphics device's DPI and current resolution to find a font size which best matches a target number of characters per physical device inch.  Running with `--fontsize` bypasses this code.

Additionally, it adds the LCD antialiasing line switch above as the default (it's 2019 - you can't even give CRT monitors away - assuming an LCD monitor is a safe assumption and at worst will not look nearly as bad as the current defaults on a CRT monitor).  For completeness, here is the IDE running on a new userdir with o.n.swing.plaf from this pull request:
https://timboudreau.com/files/screen/08-08-2019_13-22-28.png


matthiasblaesing:
Please don't merge this - it turns my setup _much_ worse. With this PR netbeans looks as if I turned blind and netbeans tries to compensate by inflating font-size (that is netbeans on my primary screen (1920x1080), not visible is my secondary 1280x1024):

Bildschirmfoto_2019-08-08_19-51-45

timboudreau:
Is that GTKLookAndFeel?

timboudreau:
I'm curious what you get for output [running this Gist](https://gist.github.com/timboudreau/5fe88764db08627948a87d8a4cb31a99) which does the same computation as the patch, but outputting a bunch of information - and how the monitor size computed from the graphics subsystem information compares with the real-world physical size of your monitor.

matthiasblaesing:
The LAF in the screenshot is the default LAF (Metal) - I removed the testuser dir and called `ant tryme`.

The referenced git outputs:

```
matthias@athena:~/tmp$ java -cp . ComputedFontSizeTester 
CURRENT DISPLAY MODE: 1920x1080
LARGEST DISPLAY MODE: 1920x1080
TOOLKIT REPORTS DPI 96
SCREEN DIAGONAL IN INCHES: 15
SCREEN SIZE IN INCHES: 20" x 11,2"
find best size with AffineTransform[[1.0, 0.0, 0.0], [0.0, 1.0, 0.0]] and heuristic font size 18 @ 96 DPI
10 chars @ 18pt = 130px - iteration 0
10 chars @ 17pt = 130px - iteration 1
10 chars @ 16pt = 110px - iteration 2
10 chars @ 15pt = 90px - iteration 3
Computed target font size 15
TARGET: 15
matthias@athena:~/tmp$ 
```

This was run on a dual-head setup - primary screen (and the only one reported above) is a notebook display. The resolution is correct, the physical size:
diagonal: 39,5cm (15.55 inch)
width: 34,5cm (13,58 inch)
height: 19,5cm (7,68inch)

timboudreau:
Okay, that's interesting.

So, a big part of the problem is we're detecting your notebook screen, not the larger attached monitor.

Not to mention that Toolkit reports the DPI from *some* monitor, but it's not clear which (I first attempted to do that computation using the "normalizing transform" from the graphics device, but it seemed to be pretty unreliable).  Hmm.

nb/ide.launcher/netbeans.conf
```diff
@@ -62,7 +62,7 @@ netbeans_default_cachedir="${DEFAULT_CACHEDIR_ROOT}/dev"
...
...es -J--add-opens=java.base/java.net=ALL-UNNAMED -J--add-opens=java.base/java.lang.ref=ALL-UNNAMED -J--add-opens=java.base/java.lang=ALL-UNNAMED -J--add-opens=java.base/java.security=ALL-UNNAMED -J--add-opens=java.base/java.util=ALL-UNNAMED -J--add-opens=java.desktop/javax.swing.plaf.basic=ALL-UNNAMED -J--add-opens=java.desktop/javax.swing.text=ALL-UNNAMED -J--add-opens=java.desktop/javax.swing=ALL-UNNAMED -J--add-opens=java.desktop/java.awt=ALL-UNNAMED -J--add-opens=java.desktop/java.awt.event=ALL-UNNAMED -J--add-opens=java.prefs/java.util.prefs=ALL-UNNAMED -J--add-opens=jdk.jshell/jdk.jshell=ALL-UNNAMED -J--add-modules=jdk.jshell -J--add-exports=java.desktop/sun.awt=ALL-UNNAMED -J--add-exports=java.desktop/java.awt.peer=ALL-UNNAMED -J--add-exports=java.desktop/com.sun.beans.editors=ALL-UNNAMED -J--add-exports=java.desktop/sun.swing=ALL-UNNAMED -J--add-exports=java.desktop/sun.awt.im=ALL-UNNAMED -J--add-exports=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED -J--add-exports=java.management/sun.management=ALL-UNNAMED -J--add-exports=java.base/sun.reflect.annotation=ALL-UNNAMED -J--add-exports=jdk.javadoc/com.sun.tools.javadoc.main=ALL-UNNAMED -J-XX:+IgnoreUnrecognizedVMOptions"
```
lkishalmi: Well, check this out:https://cwiki.apache.org/confluence/display/NETBEANS/Font+Rendering+Issues
Actually setting it to lcd will switch the default one (gasp) on Windows to lcd resulting somewhat worse AA rendering. Also it seems red hat based distros do not suffer from this one, so the best solution would be leave as it is. The Snap package however set the AA as on at the moment. Though using Java 12 without this setting has the best result of font AA on linux.

matthiasblaesing:
I tried overriding the DPI by forcing Xorg to identifiy the "right" physical size. After that I tried to adjust gtk/xfce/Xft settings to get a workable desktop and failed. So I switched back again to 96DPI and with that most fonts looks sane. The same holds true for the default netbeans sizes (on both JDK8 and 11).

timboudreau:
A big chunk of the problem here is how many fingers there are in the resolution pie:

 * Monitor firmware may or may not be reporting a broken value to the kernel
 * User may or may not [have overridden it](https://www.96boards.org/documentation/consumer/dragonboard/dragonboard410c/guides/force-display-res.md.html)
 * X configuration may or may not try to adjust it further
 * Different JDKs do different things to try and fix it

The alternative is, on first start, run a UI that modifies the netbeans.conf - I've actually got that written, before I said to myself "This is insane, you should be able to figure it out on startup" - one that simply shows a bunch of examples for font size, and then for LCD settings, and says "which one looks better?" and lets the user pick.

I'll throw that up on Github and link it here.  It's not as nice as solving the problem automagically, but it's better than the IDE looking like complete crap for a new user.

naitsirch:
As already mentioned in #1446 I always have to add `-J-Dswing.aatext=true -J-Dawt.useSystemAAFontSettings=lcd` to `netbeans_default_options` for every new installation or update on Debian with KDE.

It would be nice if this issue could be solved.

timboudreau:
This pull request is superseded by [pull request 1446](https://github.com/apache/netbeans/pull/1446) which addresses the issues here, approaching the problem from a completely different direction.

PR #1416, opened 2019-08-09
add read jvmargs from ~/.gradle/gradle.properties. 

one-leaf:
You can limit memory by set org.gradle.jvmargs=-Xmx1024m -XX:MaxPermSize=256m


groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -803,7 +812,34 @@ public void setStackTrace(StackTrace st) {
     }
 
     public void configure(ConfigurableLauncher launcher) {
-        launcher.setJvmArguments(getArgs(EnumSet.of(SYSTEM)));
+        List<String> jvmargs = getArgs(EnumSet.of(SYSTEM));
+        String userHomeDir = System.getProperty("user.home"); 
+        File gradlePropertiesFile = Paths.get(userHomeDir, ".gradle", "gradle.properties").toFile();
+        if (gradlePropertiesFile.exists()){
+            Properties pps = new Properties();
```
lkishalmi: Probably you would put this as a private method, as actually the priority of the jvm arg resolve is project gradle.properties, root project gradle.properties, then gradle user home gradle. propertties.
Also for the grade user home, please use the GradleSettings.getGradleUserHome()

groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -803,7 +812,34 @@ public void setStackTrace(StackTrace st) {
...
+            InputStream in = null;
+            try {
```
lkishalmi: try with resource?

groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -803,7 +812,34 @@ public void setStackTrace(StackTrace st) {
...
+                in = new BufferedInputStream(new FileInputStream(gradlePropertiesFile));
+                pps.load(in);
+                if (pps.containsKey("org.gradle.jvmargs")){
+                    String jvmargs_value = pps.getProperty("org.gradle.jvmargs");
```
lkishalmi: Constant declaration would be nice, tough a //NOI18N shall be added.

one-leaf:
I'm sure that this is actually required, Gradle Tooling API with Gradle Daemon can't auto add jvmargs from gradle.properties file. only run gradle command can add jvmargs from gradle.properties. If you have a better way to limit memory of gradle daemon in netbeans, please let me know.

lkishalmi:
Ok. Go ahead and do the requested changes. Thank you for your contribution!

PR #1417, opened 2019-08-10
[NETBEANS-1872] Prevent backtracking loops in CSS Parser

matthiasblaesing:
The references issue found cases, where the CSS grammar lead to
an inifite backtracking loop. This patch works around that limitation
by requireing that the parsing progresses.

The implementation counts the read calls issued without progressing
through the stream. If the read count withouth progress exceeds
10.000.000 reads, parsing is stopped.

The number was found experimentally - it needs to be large enough to
handle real world cases and low enough that progress happens early
enough.

### DIFF
--- a/ide/css.lib/src/org/netbeans/modules/css/lib/nbparser/CssParser.java
+++ b/ide/css.lib/src/org/netbeans/modules/css/lib/nbparser/CssParser.java
@@ -19,6 +19,7 @@
 package org.netbeans.modules.css.lib.nbparser;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
@junichi11 regarding issue https://issues.apache.org/jira/browse/NETBEANS-1872 I think it is worth adding the time limited parser I proposed in the issue. This is not a solution, but can mitigate the problem and limit the impact. Even if the grammar of the CSS parser is fixed for the concrete case, it could still spin out of control.

I think this is simple enough to permit merging. What do you think?

junichi11:
@matthiasblaesing I agree with you :)

matthiasblaesing:
Thanks - will merge.

PR #1418, opened 2019-08-11
[NETBEANS-1047] Don't add whitespace to a wrong place when textual operators are used

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1047

Related to https://netbeans.org/bugzilla/show_bug.cgi?id=240274

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
@@ -1326,10 +1326,12 @@ public class FormatVisitor extends DefaultVisitor {
         }
 
         while (ts.moveNext() && ts.offset() < node.getRight().getStartOffset()
-                && ts.token().id() != PHPTokenId.PHP_TOKEN && ts.token().id() != PHPTokenId.PHP_OPERATOR
+                && ts.token().id() != PHPTokenId.PHP_TOKEN && !LexUtilities.isPHPOperator(ts.token().id())
                 && lastIndex < ts.index()) {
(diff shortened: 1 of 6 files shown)


Baneeishaque:
Is the issue exist?

I used php in 8.2 version for my production works, i didn't encounte this issue.

K, i will try these works on nb 11.

junichi11:
Yes, we can reproduce it in NetBeans 8.2.
netbeans1047

Baneeishaque:
Let me check.

php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
```diff
@@ -2558,4 +2560,9 @@ protected static boolean isWhitespace(final CharSequence text) {
     private static boolean isAnonymousClass(ASTNode astNode) {
         return astNode instanceof ClassInstanceCreation && ((ClassInstanceCreation) astNode).isAnonymous();
     }
+
+    private static boolean isPHPOperator(PHPTokenId id) {
```
tmysik: Could be maybe moved to some `util` class as it seems to me to be quite useful in general. Not sure about that so just a hint.

tmysik:
Please, let me know about my hint above, I am ready to merge.

Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
junichi11: Will do it. Thanks!

junichi11:
@tmysik Fixed. Thank you!

PR #1422, opened 2019-08-12
[NETBEANS-2940] Cache ArchiveRootProviders

lkishalmi:
My intention was to improve the 40s search for ArchiveRootProviders shown at [NETBEANS-2940](https://issues.apache.org/jira/browse/NETBEANS-2940?focusedCommentId=16905420&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16905420)

I'm not sure if this change good/safe enough. I'd ask the reporter to try this out first.
As far as I've checked it worked.

### DIFF
--- a/platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
+++ b/platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
@@ -41,6 +41,7 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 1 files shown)


lkishalmi:
Well, if no one has comment against this, I'm going to merge this one tomorrow morning (PDT). As the change is pretty small this can be removed later if it would cause problems.

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -2323,17 +2324,24 @@ private static boolean isArchiveFileImpl(final URL url, final boolean strict) {
...
+                    res = archiveRootProviders.get();
+                    res.addLookupListener((ev) -> {
+                        archiveRootProviderCache = null;
+                    });
+                }
             }
+            archiveRootProviderCache = new LinkedList<>(res.allInstances());
         }
-        return res.allInstances();
+        return archiveRootProviderCache;
```
matthiasblaesing: Why is this safe? If I'm reading it correctly after the first query no new ArchiveRootProvider will be found. So If at an unfortunate moment this is called before all ArchiveRootProviders are found (or they are loaded at runtime) this will return invalid results.

lkishalmi: In Line 2335 we add a listener to the lookup result, so if new Archive providers would be registered in the system the cache would be invalidated and it's content would be re-read. Also so far there are 2 ArchiveRootProvider implementations in the system one for jars and one for jrt-s.
Profile snapshots attached to NETBEANS-2940 shows that NetBeans can spend considerable amount of time searching the non-jar provider almost every time when an archive root has to be evaluated.

matthiasblaesing: Thank you for the explanation - indeed I overlooked the lookup listener and its invalidation implementation. I toyed a bit with the ProxyLookup and it looks to react correctly.
Looks sane to me now.

lkishalmi: Thank you for the review! Let's see how does it work in the real life, then.

arusinha:
hi @lkishalmi , I noticed an issue with this PR.
Once you install nb-javac plugin, you will not be able to  run any java ant project in JDK-10/12/13.
There is no issue if the project is run using JDK-8.
Compilation of Sample project in JDK-10/12/13 have no issue, only it doesn't do anything on 'Run File' action

PR #1426, opened 2019-08-14
[NETBEANS-2980] Fix potential NPE on ImportAnalysis2.getPackage()

lkishalmi:
Trivial patch preventing NPE. The result of this method is already checked for null, so returning that would be fine.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/pretty/ImportAnalysis2.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/pretty/ImportAnalysis2.java
@@ -258,7 +258,7 @@ public class ImportAnalysis2 {
 
     //Note: this method should return either "orig" or a IdentifierTree or MemberSelectTree
     //no other tree type is not allowed - see ImmutableTreeTranslator.translateStable(Tree)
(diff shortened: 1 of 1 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/pretty/ImportAnalysis2.java
```diff
@@ -390,7 +390,7 @@ private boolean checkPackagesForStarImport(String pkgName, CodeStyle cs) {
     }
 
     private PackageElement getPackageOf(Element el) {
-        while (el.getKind() != ElementKind.PACKAGE) el = el.getEnclosingElement();
+        while ((el != null) && (el.getKind() != ElementKind.PACKAGE)) el = el.getEnclosingElement();
```
ehsavoie: More style than anything else but shouldn't a new variable and enclosing bracket makes the code more readable ?

lkishalmi: Not really. This is just a "one liner" private method

PR #1427, opened 2019-08-14
[NETBEANS-1836] Applied proposed patch, not adding self to the remove selection.

lkishalmi:


jlahoda:
I think @jtulach would be a better reviewer for this.

A few comments:
-not clear to me what exactly runs out of memory (added a comment in that sense to the Jira bug)
-not clear if the nodes that are currently being deleted are actually removed from the selection model? (Since with this patch, they are no longed added to remSel.)
-related to that - do tests pass (and, ideally, is there a test verifying the removed nodes are properly removed from the selection)?
-is there a chance to create a test for this?

PR #1429, opened 2019-08-15
Fix SaasUtilTest test in websvc.saas.api module

blackleg:
Fix SaasUtilTest test in websvc.saas.api module.

Command to run this test:
```bash
ant -f websvccommon/websvc.saas.api -Dtest.type=unit -Dtest.includes=org/netbeans/modules/websvc/saas/util/SaasUtilTest.java test-single
```


matthiasblaesing:
Looks sane - thank you.

PR #1430, opened 2019-08-16
[NETBEANS-2661] Trying to avoid excessive computations by using Binar…

jlahoda:
…yForSource query on the current source root, rather than iterating through all binary roots and using SourceForBinaryQuery for them.


lkishalmi:
On the first sight it looks sane. What is the "work-in-progress" label for? Are we expecting additional tests?

jlahoda:
I didn't have any particular/specific issues in mind, just wasn't sure if this is safe. But I guess the only way to really find out is to try it in the real world (the SourceForBinary and BinaryForSource queries should agree on the source<->binary mapping, and if they don't, we probably don't have a better way to find out than simply try - I briefly discussed this today with Tomas Zezula).

lkishalmi:
Just tested it. The performance boost is awesome. It can be felt even for relative small set of projects. I have not find any problems with it so far.

geertjanw:
> Just tested it. The performance boost is awesome. It can be felt even for relative small set of projects. I have not find any problems with it so far.

That's a great review of this PR. :-)

PR #1437, opened 2019-08-17
Fix tests in ide/db module

blackleg:
Fix tests in ide/db module.


matthiasblaesing:
I get nervous if people catch `Throwable`, but I think in this case its ok. Merging. Thank you.

PR #1439, opened 2019-08-18
Fix tests in ide/db.mysql module

blackleg:
Fix tests in ide/db.mysql module, tests are disable if class `com.mysql.jdbc.Driver` isn't available in the classpath.


matthiasblaesing:
I can understand why you check for the mysql driver, but without the driver none of the tests is run - so why pretend different? I would not add this module to the test list.

blackleg:
@matthiasblaesing I removed the module from the test list. 
My idea is prepare this module to permit execute the `ant test` command for all modules and this doesn't fail if the mysql driver isn't present.

matthiasblaesing:
Understood. Thank you.

PR #1441, opened 2019-08-18
Fix tests in ide/versioning.core module

blackleg:
Fix tests in ide/versioning.core module


blackleg:
@matthiasblaesing I cancelled this pull request in favour of this other: #1445

PR #1442, opened 2019-08-18
Fix tests in ide/versioning module

blackleg:
Fix tests in ide/versioning module


matthiasblaesing:
Could you please describe what you are fixing and how it is archived? I see that before your change I see failures and afterwards they are gone, but the why is missing. The same is true for the `ide/versioning.core` module. Thank you.

blackleg:
@matthiasblaesing I cancelled this pull request in favour of this other: https://github.com/apache/netbeans/pull/1445

PR #1446, opened 2019-08-20
An alternate solution to horrendous default font sizes and antialiasing on Linux

timboudreau:
Given that the previous pull request ran aground on the unreliability of OS-reported screen resolution and frequently erroneous monitor EDID information, a simple alternative is to simply ask the user _what looks good?_

This patch adds:

 * A standalone, *user-friendly* Java application for initial configuration of font size,
    antialiasing and a few other things, in `nb/conf-file-configurer`
 * Patches `nbbuild/build.xml` to build it and include it in `$NB_ROOT/nb/configurer`
 * Patches the launch script to run it under the following (conservative) conditions:
   * If running on a Unix whose `uname` is not "Darwin" (Mac OS)
   * If no variable is defined in `netbeans.conf` named `hw_sig` (in the future this might be used for
detecting if the screen configuration has changed) - `hw_sig` will be written by the configurer
   * The configurer has not been run and cancelled by the user in the past (in which case the
config file will contain `hw_sig=aborted` - in the case someone just doesn't want to do this,
we should get out of their way, not bother them on every restart)

See nb/conf-file-configurer/README.md for details.

Tested on Gentoo Linux w/ jdks 8/9/10/11/12, OpenIndiana (Solaris derivative) w/ jdk-8u221, Mac OS JDK 9.0.4.


naitsirch:
Will this PR fix https://askubuntu.com/questions/32624/ugly-fonts-in-netbeans-how-can-i-make-it-use-the-system-font/410001#410001?
I always have to add `-J-Dswing.aatext=true -J-Dawt.useSystemAAFontSettings=lcd` to `netbeans_default_options` for every new installation or update which is quite annoying.

naitsirch:
Couldn't this configuration be integrated into the installer?

naitsirch:
Thanks for the hint. I'll try it.

neilcsmith-net:
@naitsirch lots of people will not be using the installer - better it's handled elsewhere.

timboudreau:
> Couldn't this configuration be integrated into the installer?

Do not assume everyone uses an installer.  It could be where there *is* an installer, but is completely unhelpful if you build it yourself, or install it with the OS package manager, or use a distro like Gentoo where the "package manager" compiles everything from source with optimizations for your machine.

timboudreau:
For the folks mentioning `-J-Dswing.aatext=true` and `-J-Dawt.useSystemAAFontSettings=lcd`:

 * `-J-Dswing.aatext=true` - Has no effect on the JDK's swing classes since JDK 6, but is used by NetBeans in a few places[1].  In most (all?) cases, with another patch I recently submitted, `useSystemAAFontSettings` will be preferentially used if present
 * `-J-Dawt.useSystemAAFontSettings=lcd` - There are actually several options for this [documented here](https://docs.oracle.com/javase/7/docs/technotes/guides/2d/flags.html) - the UI this patch adds will show the user each one and let them choose which looks best (`lcd` is usually decent but frequently there are better choices for a specific monitor, although if you have multiple monitors it forces you to choose one that may be right for one and wrong for the other)

> [1] Places swing.aatext is used:
> ide/editor.completion/src/org/netbeans/modules/editor/completion/PatchedHtmlRenderer.java - this is a fork of org.openide.awt.HtmlRenderer which should get merged with it and deleted
> ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/CompositeFCS.java - tries to get hints right within the font color settings options window and perhaps elsewhere - does not look as if it also checks LCD hints
> ide/editor/lib/src/org/netbeans/modules/editor/lib/EditorRenderingHints.java - looks like a copy/paste of the code from CompositeFCS
> ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/output/OutputView.java - has my name as co-author but I don't recall it. Looks like a fork of something from core.output2, maybe ExtPlainView
> ide/team.commons/src/org/netbeans/modules/bugtracking/commons/LinkButton.java
> ide/versioning.util/src/org/netbeans/modules/versioning/history/LinkButton.java - these two both look like copy/pastes of the same class. Handles lcd correctly
> nb/welcome/src/org/netbeans/modules/welcome/content/Utils.java - ensures antialiasing the welcome screen 
> platform/core.output2/src/org/netbeans/core/output2/ExtPlainView.java - output window antialiasing, handles lcd correctly
> platform/core.output2/src/org/netbeans/core/output2/WrappedTextView.java - wrapped text view for the output window, handles lcd correctly
> platform/o.n.core/src/org/netbeans/beaninfo/editors/ColorEditor.java - color editor for beaninfo, probably completely unused these days
> platform/o.n.core/src/org/netbeans/beaninfo/editors/FontEditor.java - ditto - old beaninfo editor
> platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/ColorUtil.java - handles rendering hints for editor tabs
> platform/openide.awt/src/org/openide/awt/HtmlLabelUI.java - wraps HtmlRenderer (used by explorer views and much more - a much higher performance html renderer than Swing's

neilcsmith-net:
Where are we at with this?  Attempting to merge for NB 11.2 or leaving aside for NB 11.3 or later?  While we're past merge window, if it's desirable to get in NB 11.2 let's do it while it can go in the first beta!

neilcsmith-net:
Also, Travis failed again after retriggering so may be a sign of an issue .. or not. Few spurious failures at the moment.

eirikbakke:
Some comments regarding font size on HiDPI screens. I looked through https://github.com/apache/netbeans/pull/1415 as well.

Recentish JDK versions (9 and above) handle HiDPI screens very well on both Windows and MacOS, including in multi-monitor setups. Each monitor gets assigned a DPI scaling factor (e.g. 1x, 2x, 1.5x), and GraphicsConfiguration.getDefaultTransform() will be scaled accordingly. No need to change font sizes at the Swing level. Dragging a window from a regular screen (1x) to a HiDPI screen (e.g. 2x) will cause fonts to be rendered with a higher device resolution, but does not change the logical size of the font. Are you sure this does not work on Linux as well?

For example, could you configure your Linux multi-monitor setup with one screen at 2x scaling and one at 1x scaling, and then run the following? What output do you get?

```
import java.awt.GraphicsDevice;
import java.awt.GraphicsEnvironment;

public class ShowDPIScaling {
  public static final void main(String args[]) {
    for (GraphicsDevice gd : GraphicsEnvironment.getLocalGraphicsEnvironment().getScreenDevices())
      System.out.println(gd.getDefaultConfiguration().getDefaultTransform().getScaleX());
  }
}
```

This is not really related to what the monitor reports at a hardware level--it's just an OS-level setting (though on Windows and MacOS, the OS may take its _default_ from monitor hardware). Is there not an equivalent setting somewhere on Linux? If there is, then perhaps NetBeans should show a message explaining to the user how to configure this OS-level setting, for each individual monitor, rather than changing a single setting that applies to all monitors?

As you have probably discovered, Toolkit.getScreenResolution() should not be relied on. This is true on Windows and MacOS as well. On Windows, it behaves differently depending on whether the user has logged out and back in again since the last time the monitor setup was changed and so on. The correct way to detect HiDPI scaling factors is through GraphicsConfiguration.getDefaultTransform()--though this is normally handled transparently by Swing/Java2D.

eirikbakke:
Here's the JDK patch that introduced HiDPI support on Linux, by the way:

https://bugs.openjdk.java.net/browse/JDK-8137571

The per-monitor DPI scaling factor is retrieved in the native method X11GraphicsDevice.getNativeScaleFactor(int):

https://github.com/openjdk/jdk/blob/d7a0fb9ebc898e76207c27166b81630e837a064a/src/java.desktop/unix/classes/sun/awt/X11GraphicsDevice.java
https://github.com/openjdk/jdk/blob/7a052c979c9ef8a01a4a52195a4ab786ebbce543/src/java.desktop/unix/native/libawt_xawt/awt/awt_GraphicsEnv.c

timboudreau:
It wouldn't surprise me if the continuous build fails - this patch changes the workflow of first-start of the IDE, and for running tests without popping up the UI, it would need to alter the netbeans.conf file or similar to cause the launch script to skip that step and launch the IDE.

eirikbakke:
I've added a related PR at https://github.com/apache/netbeans/pull/3113 . It should fix the antialiasing problem, and the font sizing problem at least in the 2x HiDPI scaling case.

mbien:
haven't heard complaints about font settings for a while and this is catching dust, should we close this?

eirikbakke:
@mbien I think the anti-aliasing settings are now reliably auto-detected on Linux by https://github.com/apache/netbeans/pull/3113 and https://github.com/apache/netbeans/pull/6016 .

The special case of 2x HiDPI scaling is also auto-detected (in the first patch); other scaling factors are not supported in OpenJDK on Linux. Users who prefer larger fonts without full 2x scaling everywhere would currently have to use the "--fontsize" setting like before.

A lot of other visual issues are also likely improved by the switch to FlatLAF as the default LAF a while back. If new issues are discovered it would be useful to get new, updated reports/screenshots.

PR #1447, opened 2019-08-20
HtmlRenderer.createLabel() has been broken for years

timboudreau:
I wrote it sometime around 2002; later, to improve performance in its primary use-case as a cell-renderer, no-op overrides of listener adding and layout methods were added - without considering `HtmlRenderer.createLabel()` which hands the caller a `JLabel` that uses the high-performance limited-subset HTML renderer of `org.openide.awt.HtmlRenderer`.

Recently I've had several situations in the Antlr plugins where an HTML label without the overhead of creating a DOM under the hood was exactly what I needed, and I wound up forking this class.

Since the current API does not work as advertised, a better long term solution is simply to fix it.  The attached patch does that, without breaking the performance optimizations of the cell renderer use-case.

### DIFF
--- a/platform/openide.awt/manifest.mf
+++ b/platform/openide.awt/manifest.mf
@@ -2,5 +2,5 @@ Manifest-Version: 1.0
 OpenIDE-Module: org.openide.awt
 OpenIDE-Module-Localizing-Bundle: org/openide/awt/Bundle.properties
 AutoUpdate-Essential-Module: true
-OpenIDE-Module-Specification-Version: 7.74
+OpenIDE-Module-Specification-Version: 7.75
 
--- a/platform/openide.awt/src/org/openide/awt/HtmlLabelUI.java
+++ b/platform/openide.awt/src/org/openide/awt/HtmlLabelUI.java
@@ -98,31 +98,51 @@ class HtmlLabelUI extends LabelUI {
     private static int textWidth(String text, Graphics g, Font f, boolean html) {
         if (text != null) {
             if (html) {
-                return Math.round(
-                    Math.round(
-                        Math.ceil(
-                            HtmlRenderer.renderHTML(
+                return (int) Math.ceil(
+                        HtmlRenderer.renderHTML(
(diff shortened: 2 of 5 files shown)


geertjanw:
Thanks for this work. Seems to me like there's no reason not to merge this.

junichi11:
Should add `NOI18N`s ?

geertjanw:
Should this be merged, anything blocking merging?

matthiasblaesing:
@geertjanw Junichi hinted, that the new files should get `NOI18N` where sensible. I admit, that I don't know how these comments are used, but I suspect, they are used by sourcecode scanners looking for translateable strings.

junichi11:
> they are used by sourcecode scanners looking for translateable strings.

e.g. Tools > Internationalization

matthiasblaesing:
@junichi11 thanks - didn't now that this existed :-)

birgersp:
@timboudreau with all due respect, I think you should give this a read: https://chris.beams.io/posts/git-commit/

neilcsmith-net:
@timboudreau can we address the concerns?  Would like to merge in next couple of days if targetting 11.2 so it can be in next beta.

neilcsmith-net:
OK, moving to 11.3 milestone pending changes or review by @junichi11

junichi11:
I don't think it is a big problem(`NOI18N`), but I'll just wait for an answer from Tim.

lkishalmi:
@timboudreau how we are going with this one?

junichi11:
@lkishalmi @ebarboni Let's merge this since there is no reply.

matthiasblaesing:
We sat a bad sample - if someone else had authored this and would not bother to react to comments, I would drop the patch. @timboudreau could you please comment on this and the request for an update to the patch?

junichi11:
@matthiasblaesing You are right. I'll remove my comment.

ebarboni:
well not for 11.3 I guess.

PR #1448, opened 2019-08-20
Fix a few cases where rendering hints were handled incorrectly

timboudreau:
Having spent a bunch of time on antialiasing settings lately, I ran across a couple of cases where the code to set them up can result in grayscale antialiasing settings clobbering the ones set correctly either by the JDK from the OS, or via, e.g. `-J-Dawt.useSystemAAFontSettings=lcd_hrgb`


matthiasblaesing:
Looks sane to me.

PR #1450, opened 2019-08-20
[TRAVIS] Run tests with mysql database

blackleg:
Run tests that depends with a mysql database.


.travis.yml
```diff
@@ -179,3 +178,20 @@ matrix:
...
+          jdk: openjdk8
+          services:
+            - mysql
+          env:
+            - OPTS="-quiet -Dcluster.config=minimal -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false"
+            - OPTS_TEST="-Dtest-unit-sys-prop.ignore.random.failures=true -Dvanilla.javac.exists=true -Dtest-unit-sys-prop.mysql.user=root -Dtest-unit-sys-prop.mysql.password=password"
+          before_script:
+            - echo "ALTER USER root@'localhost' IDENTIFIED BY 'password';\nFLUSH PRIVILEGES;\n" | mysql -u root
+            - wget https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.23/mysql-connector-java-5.1.23.jar -O $HOME/mysql-connector-java-5.1.23.jar
+            - wget https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.17/mysql-connector-java-8.0.17.jar -O $HOME/mysql-connector-java-8.0.17.jar
```
matthiasblaesing: Please use the netbeans mechanism for downloading external dependencies (derby for example is loaded by it). While the most common case for the external dependencies mechanism is download for redistribution, it can also handle downloads just for compile time. In ide/derby is a sample.

blackleg: I don't know if I'm right but the Mysql driver was removed from source code due to incompatibility with Apache licence and the driver can't be a dependency or downloaded from the source code.

It's the reason that I use a wget command to download the driver independently of the source code.

matthiasblaesing: No the mysql driver was removed from distribution - if that caused unittests to fail that should be corrected. The ASF rules allow usage at compile time (you can fetch gcc and compile a project with it, you just can't redistribute it for example).

The download mechanism is distinct from the distribution mechanism, so you can download files with the existing infrastructure and use it at compile time, if you don't add the files to a project.properties files, they will not be part of a distribution.

blackleg: @matthiasblaesing I have some doubts about the netbeans download mechanism. 

- Do I need to add a license.txt file and a notice.txt file for the Mysql driver?
- Can I load the Mysql driver in the test classpath using the `project.properties` file like in this example?

```java
test.unit.cp.extra=\
    ${nb_all}/db/external/derby-10.2.2.0.jar
```

matthiasblaesing: @blackleg to your questions:

Yes you need a `-license.txt` file, it is required for all dependencies. For GPLed dependencies there are two constraints:
1) there needs to be a "Type" header, which marks it as "compile-time": https://github.com/apache/netbeans/blob/master/ide/derby/external/derby-10.14.1.0-license.txt#L5
2) there needs to be a "Comment" header, that explains why the dependency is there

For 2) stating, that the driver is required for the unittests of the DB support would be a valid explaination.


Adding the file to the `binaries-list` file just downloads it, it is not automatically packed for distribution or added to a classpath. So yes, you can reference it as shown in your example.

blackleg: @matthiasblaesing I changed the pull request to use the netbeans download mechanism. Could you review it? I'm not sure if the licence files are correct.

matthiasblaesing:
Looks good to me. Is it ok with you, if I squash the changes together into one changeset?

blackleg:
@matthiasblaesing There isn't any problem for me if you prefer squash the changes.

matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/aec700e82092af3851fd9d8883eadfcd942ee035. Thank you.

neilcsmith-net:
@blackleg the Travis "Test with MySQL database" introduced here is failing almost half the time at the moment for spurious reasons and I keep having to retrigger.  If you get a chance could you check what the issue is?  Thanks!

blackleg:
@neilcsmith-net I'm trying to find a solution for that.

blackleg:
@neilcsmith-netI didn't found the error, I think it is a problem with the travis environment. I can't reproduce the error in my computer but I created this pull request to try to prevent the randomly errors. https://github.com/apache/netbeans/pull/1550

PR #1451, opened 2019-08-20
Improve/fix unittests in webcommon cluster

matthiasblaesing:
With this PR, all unittests in the webcommon cluster work apart from:

- cordova
- lib.v8debug
- javascript2.requirejs
- javascript2.nodejs

The unittests are added to the travis build to prevent regressions.


matthiasblaesing:
@blackleg could you please have a look at this commit:

https://github.com/apache/netbeans/pull/1451/commits/3737d6bb6cc3c8addd74b6a767f7bc7be6399189

The  definition of `TEST_MODULES` is not picked up by the build. The log is here: 

https://travis-ci.org/apache/netbeans/jobs/574516404

Neither is the variable reported in the "environment variables" block, nor is it picked up in the test execution. The strange thing is, that the same happens for the block of modules, that already existed. I don't see the problem.

blackleg:
@matthiasblaesing Try to define the TEST_MODULES variable in one line instead split in one line per module.

matthiasblaesing:
@blackleg thank you for your input - yes switching to a single line indeed works (with the drawback, that it makes it unreadable). For now I removed the commit that enables automatic tests through travis again, as the test seem to be unstable (ok for manual use, not so ok for automatic use)

@sdedic @JaroslavTulach  could you please have a look at this? I think most of the changes are straight forward fixes, but this might raise questions:

https://github.com/apache/netbeans/pull/1451/commits/0ae2e4639ae1d842e2a288dacab99491979bf4db

I decided to remove these tests, as I could not find a trace for the license of the source files. While the files for the *doc tests are also not straight copies of the jsdoc files, they are close enough to the files I could find to be believably derived from the GPLed sources. For the removed tests, only files in the netbeans bugzilla could be found, which is IMHO not enough.

matthiasblaesing:
If noone objects, I'll fix the merge conflicts by Wednesday and merge then.

PR #1452, opened 2019-08-20
[NETBEANS-3011] - reduce the amount of redundant cast warnings..

BradWalker:
This work simply reduces the redundant cast warnings in the code base..

```
[repeat] /home/bwalker/netbeans/ide/xml.catalog.ui/src/org/netbeans/modules/xml/catalog/CatalogPanel.java:302: warning: [cast] redundant cast to DataObject
 [repeat] DataObject dobj = (DataObject) n.getLookup().lookup(DataObject.class);
 [repeat] ^
```


matthiasblaesing:
Thank you. Merged via https://github.com/apache/netbeans/commit/677a618943aa4b88f905840a8812b6b94e10caab. It would be good if you create a PR per Changeset. In this PR you can see, that you pulled in your PRs into this one, this should not be done.

PR #1453, opened 2019-08-21
Fix tests for ide/editor.structure module

blackleg:
Fix tests for ide/editor.structure module
- Remove assert in `DocumentModel.java` that brokes the tests.
- Add insane module is a test dependency.
- Change deprecated code.

### DIFF
--- a/ide/editor.structure/test/unit/src/org/netbeans/modules/editor/structure/api/DocumentModelTest.java
+++ b/ide/editor.structure/test/unit/src/org/netbeans/modules/editor/structure/api/DocumentModelTest.java
@@ -262,7 +262,7 @@ public class DocumentModelTest extends NbTestCase {
      *
      * @author Marek Fukala
      */
-    private class FakeDocumentModelProvider implements DocumentModelProvider {
+    private static class FakeDocumentModelProvider implements DocumentModelProvider {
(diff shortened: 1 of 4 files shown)


ide/editor.structure/test/unit/src/org/netbeans/modules/editor/structure/api/DocumentModelTest.java
```diff
@@ -294,48 +305,70 @@ public void updateModel(DocumentModel.DocumentModelModificationTransaction dtm,
         
     }
     
-    private static class DocumentModelListenerAdapter implements DocumentModelListener {
+    private class DocumentModelListenerAdapter implements DocumentModelListener {
```
matthiasblaesing: Please make these classes static again - non static inner classes hold a hard reference to their enclosing object. which can cause suptle memory issues. Yes I know, that this is just test code, but you also added @Override annotations (which is ok), so I'd like this to be reverted.

That is also true for FakeDocumentModelProvider.

blackleg: @matthiasblaesing I reverted the changes.

PR #1457, opened 2019-08-23
NETBEANS-3020 Cleanup Welcome screen

geertjanw:
Remove the left panel in the 'What's New' tab and change 'Blogs' to 'News' in the right panel, so that the Apache NetBeans blogs provides the news in the Welcome Screen and so that the dependency on netbeans.org is removed.


nb/welcome/src/org/netbeans/modules/welcome/resources/Bundle.properties
```diff
@@ -64,19 +60,19 @@ ACD_AllNews=Open the list of all news articles in the default web browser
 
 LBL_AllArticles=All Tutorials >>
 #NOI18N
-URL_AllArticles=http://netbeans.org/kb/kb.html?utm_source=netbeans&utm_campaign=welcomepage
+URL_AllArticles=https://netbeans.apache.org/help/index.html
```
jkovalsky: Maybe [https://netbeans.apache.org/kb/docs/index.html](https://netbeans.apache.org/kb/docs/index.html) here would be better?

YTN01:
I saw few lines that got commented out. Any reason for not removing them?

geertjanw:
> I saw few lines that got commented out. Any reason for not removing them?

True, fixing that.

nb/welcome/src/org/netbeans/modules/welcome/resources/Bundle.properties
geertjanw: Problem is that the https://netbeans.apache.org/kb/docs/index.html page is incomplete at the moment and better not to expose it too much right now.

geertjanw: Problem is that https://netbeans.apache.org/kb/docs/index.html is incomplete right now and better not to highlight it at this point yet.

nb/welcome/src/org/netbeans/modules/welcome/resources/tutorial4.url
```diff
@@ -1 +1 @@
-http://netbeans.org/kb/trails/php.html?utm_source=netbeans&utm_campaign=welcomepage
\ No newline at end of file
+https://netbeans.apache.org/kb/docs/php/index.html
```
vieiro: Hi @geertjanw , @neilcsmith-net , what about keeping the the "?utm_source=netbeans&utm_campaign=welcomepage" suffix even in the new URL? Who knows, maybe we can detect requests in the future...

neilcsmith-net:
@vieiro aside from flexibility to redirect, the other reason for keeping links inbound from the IDE under `/nb/` was to allow for the option of tracking requests separately.

vieiro:
@neilcsmith-net I know. Whatever is simpler for @geertjanw would do. Not sure we will be able to track anything, though (that's Infra business, IIRC).

geertjanw:
I'm going to keep this PR as it is, if we need to do other things in this area, we can handle it via other PRs and issues. Thanks for the comments, squashing and merging now.

nb/welcome/src/org/netbeans/modules/welcome/resources/Bundle.properties
```diff
@@ -97,7 +89,7 @@ URL_IncubatorLogo=http://incubator.apache.org
 #NOI18N
 URL_ApacheLogo=http://www.apache.org
```
thomas-mc-work: Why not `https://…`?

PR #1458, opened 2019-08-23
[NETBEANS-3022] - reduce the number of unchecked call warnings..

BradWalker:
Reduce the number of unchecked call warnings. like this..

```
   [repeat] /home/bwalker/netbeans/webcommon/javascript2.nodejs/src/org/netbeans/modules/javascript2/nodejs/editor/model/NodeJsObjectInterceptor.java:101: warning: [unchecked] unchecked call to ArrayList(Collection<? extends E>) as a member of the raw type ArrayList
   [repeat]             List<? extends DeclarationScope> childrenScopesCopy = new ArrayList(globalScope.getChildrenScopes());
   [repeat]                                                                   ^
```

This is done by  simply adding the diamond operator <>..

In addition, I also had to update the project.properties file at a few places to allow the use of the diamond operator..

```
compile:
   [depend] Deleted 14 out of date files in 0 seconds
 [nb-javac] Compiling 10 source files to /home/bwalker/netbeans/platform/masterfs/build/classes
   [repeat] warning: [options] bootstrap class path not set in conjunction with -source 1.6
   [repeat] /home/bwalker/netbeans/platform/masterfs/src/org/netbeans/modules/masterfs/filebasedfs/naming/NamingFactory.java:50: error: diamond operator is not supported in -source 1.6
   [repeat]         final List<FileInfo> list = new ArrayList<>(queue);
   [repeat]                                                   ^
   [repeat]   (use -source 7 or higher to enable diamond operator)
   [repeat] 1 error
   [repeat] 1 warning
  [nbmerge] Failed to build target: all-masterfs
```

In these few places I changed the source to 1.8..


matthiasblaesing:
Merged via 71495732dc6f38f884bd0c25616eb489e23ef7c0. I rebased the changeset onto master.

PR #1461, opened 2019-08-25
[NETBEANS-2832] Added input reading to Gradle projects.

lkishalmi:


groovy/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
```diff
@@ -152,6 +156,12 @@ public void run() {
 
             outStream = new EscapeProcessingOutputStream(new GradlePlainEscapeProcessor(io, config, false));
             errStream = new EscapeProcessingOutputStream(new GradlePlainEscapeProcessor(io, config, true));
+            try {
+                inStream = new ReaderInputStream(io.getIn(), "UTF-8"); //NOI18N
```
matthiasblaesing: Is the JVM running the daemon configured to used UTF-8 as default encoding? I know that the java ecosystem is moving to UTF-8 as the default encoding, but is this a safe assumption here?

matthiasblaesing: Actually the IDE should not be the problem. The input comes directly from Swing, so it is (module bugs) clean, then it is passed through a UTF-8 writer, that converts the string to a byte sequence. So if the target is clean, this should be stable (yeah famous last words). I would also let it stay as is.

PR #1462, opened 2019-08-27
[NETBEANS-3019] Using an actual/existing entered CompilationUnit to c…

jlahoda:
…heck accessibility, instead of possibly a scratch one, which may not have an owning module filled.


neilcsmith-net:
@jlahoda retriggered failing test to check.  Is this intended as a fix for 11.2?  If so, please add the milestone and/or add me to review and I'll merge in.

jlahoda:
Would be nice for 11.2, but OK if too late.

PR #1468, opened 2019-08-30
NETBEANS-1097 "Copy Dependent Libraries" does not work if "Source/Bin…

geertjanw:
…ary Format" is greater than JDK 8

Removing the below in build-impl.xml which seems to cause more problems than it helps:

&lt;not&gt;
    &lt;istrue value="${{modules.supported.internal}}"&gt;
&lt;/not&gt;

See https://issues.apache.org/jira/browse/NETBEANS-1097.

### DIFF
--- a/java/java.j2seproject/src/org/netbeans/modules/java/j2seproject/resources/build-impl.xsl
+++ b/java/java.j2seproject/src/org/netbeans/modules/java/j2seproject/resources/build-impl.xsl
@@ -294,15 +294,6 @@ is divided into following sections:
                         <istrue value="${{not.archive.disabled}}" />
                     </or>
                 </condition>
-                <condition property="do.mkdist">
-                    <and>
-                        <isset property="do.archive"/>
-                        <isset property="libs.CopyLibs.classpath"/>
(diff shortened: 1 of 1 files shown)


jlahoda:
I believe the reason why this was introduced is that for modular J2SE projects, the copying does not work (specifically, copying of module path is not supported, AFAIK). It might still be OK, but possibly a better solution would be to base the decision on module-info:
https://github.com/apache/netbeans/pull/1469

It is not completely correct, as a project without module-info can still depend on module on the module path (which is then not copied, and the simple java -jar would not work anyway), but I think that is an acceptable trouble, as such projects are likely to be very rare.

wharnisch:
Perhaps this additional info found at [https://stackoverflow.com/questions/55994732/netbeans-11-does-not-copy-libs-into-dist-folder](url) is important:
Setting do.mkdist=true in the nbproject/private/private.properties file also solves the problem
Regards

geertjanw:
OK, I will close this one and merge https://github.com/apache/netbeans/pull/1469.

PR #1469, opened 2019-09-01
[NETBEANS-1097] Copy the libs when module-info is not present and do …

jlahoda:
…not copy them if it is present.


geertjanw:
OK, merging, makes sense.

PR #1470, opened 2019-09-04
NP-49 Payara Platform 5.193 support

jGauravGupta:


geertjanw:
Hard to argue with this, it comes from @jGauravGupta from Payara, and it only affects Payara-related code, so going to simply merge it.

PR #1471, opened 2019-09-04
NP-50 Upgrade Payara Micro Maven plugin to 1.0.6 and custom contextRoot support

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta.jc@gmail.com>


geertjanw:
Makes sense, is by @jGauravGupta from Payara, and only impacts Payara, so merging.

PR #1472, opened 2019-09-04
Fixed typo in gitattributes file

lbruun:
`text` was misspelled as `test`. The impact has been that git hasn't enforced any particular line-ending style for files of type .xsd. It will do so now, similar to other text type files.


matthiasblaesing:
Good eye :-)

PR #1473, opened 2019-09-05
[NETBEANS-2751] Register OpenJFX sample artifacts as NetBeans project…

geertjanw:
… templates

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/newproject/MavenWizardIterator.java
+++ b/java/maven/src/org/netbeans/modules/maven/newproject/MavenWizardIterator.java
@@ -84,6 +84,18 @@ public class MavenWizardIterator implements WizardDescriptor.BackgroundInstantia
     public static WizardDescriptor.InstantiatingIterator<?> javafxSample() {
         return ArchetypeWizards.definedArchetype("org.codehaus.mojo.archetypes", "sample-javafx", "0.5", null, LBL_Maven_JavaFx_Sample_Archetype());
     }
+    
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
For the code itself: The javafx-archetype-simple archetype is invalid, it is only present in version 0.0.1, not 0.0.2, the same is true for javafx-archetype-fxml.

The list in the "Java with Maven" category becomes pretty long. Maybe we should use a subfolder "From Archetype", where the generic "Project from Archetype" is the first option and selected archetypes come after it. This is my try to illustrate it:

Before:

now

After:

after

geertjanw:
0.0.2 was definitely available recently, will check.

Agree completely with your folder restructuring -- or maybe those from archetype should be in the Samples category instead?

tiainen:
We've just released version 0.0.3 which updates to the latest release of OpenJFX 13.

geertjanw:
Thanks for the info!

geertjanw:
But I don't see it here: https://mvnrepository.com/artifact/org.openjfx/javafx-archetype-simple

That's maybe where @matthiasblaesing was looking for it too?

matthiasblaesing:
I'm looking on maven central (the default maven repository): 
https://search.maven.org/search?q=javafx-archetype-simple

This indicates, that 2.0.0 was published yesterday, which would explain why it was not present when I tested.

tiainen:
It was available from maven central when I posted my message. Synchronization with search.maven.org can take up to a few hours (it's listed there now btw). Synchronization with mvnrepository.com probably takes even longer.

geertjanw:
Closing this one. Pushing a new one with a different and more complete solution.

PR #1474, opened 2019-09-08
Improving compatibility with the records-and-sealed branch of OpenJDK…

jlahoda:
… Amber - if JCClassDecl does not have the expected constructor, let javac create the tree.


neilcsmith-net:
For NB 11.2?  If ready, please merge, thanks.

PR #1475, opened 2019-09-08
Ability to use the NetBeans' LSP server instead of the embedded Java support for selected projects

jlahoda:


Chris2011:
This can be handled by the user right? I mean they can choose what they want, right?

jlahoda:
I think it would be good to have this at some point, but it does not necessarily need to be 12.6. So given the branch is likely to happen this week, we can aim to merge this soon after the branch, and see.

neilcsmith-net:
@jlahoda we're getting close to that week away from branching point again!  Are we looking to get this in NB13 still?

jlahoda:
@neilcsmith-net  - lets defer again - we have enough related conflicting changes in 13.

PR #1480, opened 2019-09-09
[NETBEANS-3075] - reduce number of warning w/ raw type iterator

BradWalker:
I've tried to reduce the amount of iterator raw type warnings that are emitted..

Here is an example of what I'm trying to fix.

```
   [repeat] /home/bwalker/netbeans/enterprise/websvc.manager/src/org/netbeans/modules/websvc/manager/model/WebServiceGroup.java:102: warning: [rawtypes] found raw type: Iterator
   [repeat]             Iterator iter = listeners.iterator();
   [repeat]             ^
```


matthiasblaesing:
Please check this PR again Brad - it contains only merges from master (which should not be there).

BradWalker:
Will fix this in another pull request..

PR #1405, opened 2019-08-04
JEDDICT-322 Export public-packages & friend-packages to Jeddict modules

jGauravGupta:
Github issue: https://github.com/jeddict/jeddict/issues/322

Internal API of the Apache NetBeans Platform used by Jeddict modules:

| Netbeans Module name | Netbeans Module location | APIs | Used by Jeddict modules | Requirement |  Why access needed? |
| --- | --- | --- | --- | --- | --------- |
| org.netbeans.modules:org-netbeans-modules-j2ee-persistenceapi | netbeans\java\j2ee.persistenceapi |  org.netbeans.modules.j2ee.persistence.api, org.netbeans.modules.j2ee.persistence.spi | io.github.jeddict.jcode.util, io.github.jeddict.jpa.modeler, io.github.jeddict.reveng | friend-packages | Java persistence is core feature of Jeddict modeler so persistence.api module is needed to find persistence.xml location in project, to identify persistence descriptor properties, to list available Java Entity classes in project during Java Classes reverse engineering, etc. |
| org.netbeans.modules:org-netbeans-modules-j2ee-persistence | netbeans\java\j2ee.persistence | org.netbeans.modules.j2ee.persistence.dd, org.netbeans.modules.j2ee.persistence.dd.common, org.netbeans.modules.j2ee.persistence.unit, org.netbeans.modules.j2ee.persistence.entitygenerator, org.netbeans.modules.j2ee.persistence.jpqleditor, org.netbeans.modules.j2ee.persistence.jpqleditor.ui, org.netbeans.modules.j2ee.persistence.editor, org.netbeans.modules.j2ee.persistence.provider, org.netbeans.modules.j2ee.persistence.wizard, org.netbeans.modules.j2ee.persistence.wizard.jpacontroller, org.netbeans.modules.j2ee.persistence.wizard.fromdb, org.netbeans.modules.j2ee.persistence.wizard.library, org.netbeans.modules.j2ee.persistence.wizard.unit | io.github.jeddict.jcode.util, io.github.jeddict.jpa.modeler, io.github.jeddict.reveng, io.github.jeddict.orm.generator,  io.github.jeddict.runtime, io.github.jeddict.rest.generator, io.github.jeddict.mvc.generator, io.github.jeddict.infra | friend-packages | JPQL Editor integration in JPA Modeler, persistence.xml updated during source generation |
[...]

PR #1482, opened 2019-09-09
JEDDICT-322 org.netbeans.modules.db.metadata.model convert friend to public packages

jGauravGupta:
Github issue: https://github.com/jeddict/jeddict/issues/322

Friend API of the Apache NetBeans Platform used by Jeddict modules:

| NetBeans Module name | NetBeans Module location | APIs | Used by Jeddict modules | Requirement |  Why access needed? |
| --- | --- | --- | --- | --- | --------- |
| org.netbeans.modules:org-netbeans-modules-db-metadata-model | netbeans\ide\db.metadata.model | org.netbeans.modules.db.metadata.model.api |  io.github.jeddict.jpa.modeler, io.github.jeddict.db.modeler | public-packages | To visualize Tree-Structured tables from Services > Database in Jeddict modeler |

### DIFF
--- a/ide/db.metadata.model/apichanges.xml
+++ b/ide/db.metadata.model/apichanges.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<?xml-stylesheet type="text/xml" href="../../nbbuild/javadoctools/apichanges.xsl"?>
+<!DOCTYPE apichanges PUBLIC "-//NetBeans//DTD API changes list 1.0//EN" "../../nbbuild/javadoctools/apichanges.dtd">
+
+<apichanges>
+    <apidefs>
+        <apidef name="api">DB Metadata Model API</apidef>
+    </apidefs>
+
+    <changes>
+        <change id="public-api">
+            <api name="db.metadata.model.api"/>
+            <summary>Making o.n.m.db.metadata.model.api and o.n.m.db.metadata.model.spi public API</summary>
+            <version major="1" minor="17"/>
+            <date day="02" month="11" year="2019"/>
+            <author login="jgauravgupta"/>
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
What was done to verify, that the exported API is sane?

jGauravGupta:
PR https://github.com/apache/netbeans/pull/1405 is now separated in multiple PR per modules. Updated the description for the usage of API in Jeddict plugin.

matthiasblaesing:
Don't take this wrong, but I repeat my question: What was done to verify, that the API is sane? I'm not too happy with the DB modules api.

jGauravGupta:
Hi @matthiasblaesing,

What's the process to check the sanity of the module e.g more unit tests, examples needed etc?

junichi11:
Can't we add your codenamebase to api friend? Maybe, it is better than changing to public, I suppose.
My Plugins do it:
- https://github.com/apache/netbeans/blob/master/ide/bugtracking.commons/nbproject/project.xml
- https://github.com/junichi11/netbeans-github-issues-plugin/blob/master/pom.xml

jGauravGupta:
Hi @junichi11,

In the PR https://github.com/apache/netbeans/pull/1405, I have added the codenamebase as a friend but @neilcsmith-net suggested to remove the friend-only restrictions. 
So I am creating PRs for all the modules required by Jeddict which have API/SPI packages and exported to friends marked as public.

junichi11:
@jGauravGupta I see. I'm not sure whether it is good. Maybe, @JaroslavTulach can advise the proper way?

neilcsmith-net:
@junichi11 we had a lot of discussion on public vs friend APIs about a year ago. There was little consensus aside from a recognition that we need to move towards stabilising and exposing necessary APIs.  These APIs may not be perfect, and perhaps we need a better in-development and deprecation mechanism, but IMO we should not be adding in friend access for third-party projects - that is just adding a stability issue by the back door.

I suggested @jGauravGupta propose a PR to expose what is required, and (I think) also start a mailing list thread on this.  With a move to quarterly releases we need to move away from plugin authors requiring either use of implementation versions or friend access.

Adding in a review request from @lkishalmi who started that earlier conversation on public vs friend.

matthiasblaesing:
> What's the process to check the sanity of the module e.g more unit tests, examples needed etc?

That is a good question, but I think outside the scope of a PR discussion: this is the reason I asked what was done. I would like to discuss this on the mailinglist to get some more people to voice their opinion.

As we are nearing feature freeze, I would like to postpone this till release of 11.2 is done. As a work around, could you use yenta (https://github.com/jglick/yenta)?

jGauravGupta:
Cool, I was not aware of **yenta**, I will check it out. Thanks @matthiasblaesing.

neilcsmith-net:
> As we are nearing feature freeze, I would like to postpone this till release of 11.2 is done. As a work around, could you use yenta (https://github.com/jglick/yenta)?

I'd somehow missed that was an option for IDE plugins as well as RCP applications?!  @jGauravGupta if that can handle requirements for 11.2, great!  If not, another workaround that could be considered is a PR for the friend dependencies against the 11.2 release branch (when it exists)?  With a view to a better fix in master by 11.3 in line with @matthiasblaesing comment on postponing?

JaroslavTulach:
It would be better to see the generated Javadoc before approving the integration. Anyway I think it is relatively good. Let's go on and I hope you'll be able to fix problems later.

PR #1483, opened 2019-09-09
[NETBEANS-3076] Try to create unit-test with Gradle Projects

lkishalmi:
This is actually a unittest for NETBEANS-2925. Separated as it affects other modules with javac.source upgrade 1.6 -> 1.8
Also it needs to interact with Gradle which might fail, and is so requires more plumbing.

### DIFF
--- a/groovy/gradle.java/nbproject/project.properties
+++ b/groovy/gradle.java/nbproject/project.properties
@@ -20,6 +20,9 @@ javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial
 nbm.module.author=Laszlo Kishalmi
 
+test-unit-sys-prop.test.netbeans.dest.dir=${netbeans.dest.dir}
+test-unit-sys-prop.java.awt.headless=true
+
 release.external/org.jacoco.core-0.8.3.jar=modules/gradle/org.jacoco.core.jar
 release.external/asm-7.0.jar=modules/gradle/asm.jar
 release.external/asm-commons-7.0.jar=modules/gradle/asm-commons.jar
--- a/groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
+++ b/groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.gradle.java.classpath;
+
+import java.io.File;
+import static junit.framework.TestCase.assertNotNull;
+import org.netbeans.api.java.project.JavaProjectConstants;
+import org.netbeans.api.project.Project;
+import org.netbeans.api.project.ProjectManager;
+import org.netbeans.api.project.ProjectUtils;
+import org.netbeans.api.project.SourceGroup;
+import org.netbeans.api.project.Sources;
+import org.netbeans.junit.NbTestCase;
+import org.netbeans.modules.apisupport.project.InstalledFileLocatorImpl;
+import org.netbeans.modules.project.uiapi.ProjectOpenedTrampoline;
+import org.netbeans.spi.project.ui.ProjectOpenedHook;
+import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
+import org.openide.filesystems.test.TestFileUtils;
+
+/**
+ *
+ * @author lkishalmi
+ */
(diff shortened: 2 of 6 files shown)


groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+        Sources srcs = ProjectUtils.getSources(prj);
+        SourceGroup[] groups = srcs.getSourceGroups(JavaProjectConstants.SOURCES_TYPE_JAVA);
+        assertEquals(2, groups.length);
+        assertTrue(groups[0].contains(source));
+        assertFalse(groups[0].contains(generated));
+        assertTrue(groups[1].contains(generated));
+        assertFalse(groups[1].contains(source));
+    }
+
+    private static File getTestNBDestDir() {
```
matthiasblaesing: Why is the test project not created inside the workdir? I don't understand the value to use a different directory.

groovy/gradle.java/nbproject/project.properties
```diff
@@ -20,7 +20,10 @@ javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial
 nbm.module.author=Laszlo Kishalmi
 
+test-unit-sys-prop.test.netbeans.dest.dir=${netbeans.dest.dir}
+test-unit-sys-prop.java.awt.headless=true
```
matthiasblaesing: Any reason to force this to headless? If not required, I would try to stay with defaults.

groovy/gradle.java/nbproject/project.properties
```diff
@@ -20,7 +20,10 @@ javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial
 nbm.module.author=Laszlo Kishalmi
 
+test-unit-sys-prop.test.netbeans.dest.dir=${netbeans.dest.dir}
```
matthiasblaesing: See comment in the testcase.

groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        clearWorkDir();
+        d = FileUtil.toFileObject(getWorkDir());
+        destDirF = getTestNBDestDir();
+    }
+
+    public void testGeneratedSources() throws Exception { // #187595
```
matthiasblaesing: Is my interpretation correct, that this should test if a folder holding a gradle project is recognized and basic java project data is available?

groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+        Project prj = ProjectManager.getDefault().findProject(d);
+        assertNotNull(prj);
+        ProjectOpenedTrampoline.DEFAULT.projectOpened(prj.getLookup().lookup(ProjectOpenedHook.class));
+        Sources srcs = ProjectUtils.getSources(prj);
+        SourceGroup[] groups = srcs.getSourceGroups(JavaProjectConstants.SOURCES_TYPE_JAVA);
+        assertEquals(2, groups.length);
+        assertTrue(groups[0].contains(source));
+        assertFalse(groups[0].contains(generated));
+        assertTrue(groups[1].contains(generated));
+        assertFalse(groups[1].contains(source));
```
matthiasblaesing: Is the order of the returned `SourceGroups` guaranteed? If not, this may randomly fail if order is reversed.

groovy/gradle.java/nbproject/project.properties
timboudreau: @matthiasblaesing - if you can run a test headless, you should.  In particular, I know there are tests in exception handling code and the dialogs api for whether the toolkit is headless.  While there are tests that must open a UI a drive it, this tends to hijack focus on developer machines unless you run tests in a virtual X framebuffer.  Since thrown exceptions can result in dialogs unexpectedly popping up, any test that *can* run headless should prefer that, and only tests that simply must interact with windows and will fail otherwise should not be run headless.

matthiasblaesing: What made you think I disagree with the assessment, that unittests should be run without GUI interaction? I just don't like setting properties all through the codebase, because they have the tendency not to be maintained. We learned  in the discussion about the `javac.source`, that people forget why a property was set. We also learned, that system default can be better, then your own implementation, as seen in the discussion about the default heap use calculation, which was switch to use the JDK default.

groovy/gradle.java/nbproject/project.properties
lkishalmi: This one is needed to get the installed file locator working properly. Without that the Gradle side plugin cannot be loaded.

groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
lkishalmi: Yes sourcegroups are strictly ordered.

groovy/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImplTest.java
lkishalmi: It is actually a test for #1404  when you have a source directory inside the build dir. The elements of that directory are filtered out by the default sourcegroup implementation.

groovy/gradle.java/nbproject/project.properties
lkishalmi: The headless true was just copied from another module which dealt with project ui module. As I need to actually open the project to get the information from Gradle instead of the heuristics.

PR #1484, opened 2019-09-10
JEDDICT-322 org.netbeans.modules.maven convert friend to public packages

jGauravGupta:
Github issue: https://github.com/jeddict/jeddict/issues/322

Friend API of the Apache NetBeans Platform used by Jeddict modules:

| NetBeans Module name | NetBeans Module location | APIs | Used by Jeddict modules | Requirement |  Why access needed? |
| --- | --- | --- | --- | --- | --------- |
| org.netbeans.modules:org-netbeans-modules-maven | netbeans\java\maven |org.netbeans.modules.maven.api, org.netbeans.modules.maven.api.customizer, org.netbeans.modules.maven.configurations, org.netbeans.modules.maven.execute.model | io.github.jeddict.jcode.util | public-packages | To parse or update the project pom during source code generation |

### DIFF
--- a/java/maven/nbproject/project.xml
+++ b/java/maven/nbproject/project.xml
@@ -592,6 +592,7 @@
                 <package>org.netbeans.modules.maven.api.execute</package>
                 <package>org.netbeans.modules.maven.api.output</package>
                 <package>org.netbeans.modules.maven.api.problem</package>
+                <package>org.netbeans.modules.maven.configurations</package>
                 <package>org.netbeans.modules.maven.execute</package>
                 <package>org.netbeans.modules.maven.execute.model</package>
                 <package>org.netbeans.modules.maven.execute.model.io.jdom</package>
--- a/java/maven/nbproject/project.xml
+++ b/java/maven/nbproject/project.xml
@@ -583,51 +583,7 @@
                     </test-dependency>
                 </test-type>
             </test-dependencies>
-            <friend-packages>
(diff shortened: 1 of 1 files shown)


java/maven/nbproject/project.xml
```diff
@@ -636,6 +592,7 @@
                 <package>org.netbeans.modules.maven.api.execute</package>
                 <package>org.netbeans.modules.maven.api.output</package>
                 <package>org.netbeans.modules.maven.api.problem</package>
+                <package>org.netbeans.modules.maven.configurations</package>
                 <package>org.netbeans.modules.maven.execute</package>
```
jtulach: This is not good idea. While publishing `org.netbeans.modules.maven.api.*` is OK, the `org.netbeans.modules.maven` shall probably stay as friend.

jGauravGupta: AFAIK both friend and public APIs can't be defined in the same module.

JaroslavTulach: That is true. The module would have to be split into two. The API and the implementation.

JaroslavTulach: Try to refactor the functionality you need into `maven.api` module. A bit more work, but it will pay off for the whole Apache NetBeans project.

jGauravGupta:
Ok, closing this PR and will create another one with Split Maven modules.

PR #1487, opened 2019-09-10
[NETBEANS-3078]:Provided Error Hint to turn on Preview Features in An…

arusinha:
…t Java projects

Jira Link: https://issues.apache.org/jira/browse/NETBEANS-3078

Suppose we had used switch-expression in Java application (JDK-12)
e.g.,
int a = switch (num)
{      case 1 ->1;     default->2; }
;
It will show the below error:

switch expressions are a preview feature and are disabled by default.
(use --enable-preview to enable switch expressions)

The proposed hint will add the needed JVM parameter in run/compiler properties

### DIFF
--- a/java/ant.hints/build.xml
+++ b/java/ant.hints/build.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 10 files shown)


geertjanw:
Great, very useful, merging,

PR #1488, opened 2019-09-10
Update netbeans bundled JNA library to 5.4.0

matthiasblaesing:
The JNA release cycles not only added featured, but also improved stability
and deadlock resilience. This PR updates the JNA version bundled with JNA
and updates it to the current version.

### DIFF
--- a/contrib/libs.svnClientAdapter.svnkit/manifest.mf
+++ b/contrib/libs.svnClientAdapter.svnkit/manifest.mf
@@ -1,5 +1,5 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.libs.svnClientAdapter.svnkit/1
 OpenIDE-Module-Localizing-Bundle: org/netbeans/libs/svnclientadapter/svnkit/Bundle.properties
-OpenIDE-Module-Specification-Version: 1.25
+OpenIDE-Module-Specification-Version: 1.26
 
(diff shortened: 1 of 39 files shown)


matthiasblaesing:
@neilcsmith-net with Praxis LIVE you are directly affected by this change. Could you please have a look at this?

- I updated in-tree users where they were either incompatible or used deprecated calls.
- Usage as a dependency is present in the `External Execution Process` and `JSch Agent Proxy library`. For the jsch agent proxy I checked out the original source code and updated the dependency to jna 5.4.0 (it compiled cleanly). For the `External Execution Process` JNA is only used from the `ProcessTreeKiller`. Missing the source code for the latter I decompiled it tried building it with a recent version of JNA. I feel confident, that this works.

neilcsmith-net:
@matthiasblaesing thanks for the offer to review.  This won't affect PraxisLIVE, as [for a long time](https://praxisintermedia.wordpress.com/2012/01/04/jna-3-4-0-on-the-netbeans-platform/) it has shipped with its own choice of version.

I can't test this until next week, but +1 in principle to this change.  It would also be good to remove the friend API restriction here?  JNA is top of my annoying use of friend API with third-party dependencies!  If you read the blog post link above, you'll notice me moaning about this in 2012. :smile:

matthiasblaesing:
@neilcsmith-net I removed the friend-only restriction. I think JNA is mature enough to indicate, when it breaks compatibility and then netbeans has a mechanism to indicate the break as I utilized in this PR.

matthiasblaesing:
I rereviewed my changes and think they are safe. If really problems creep in, I'm around to fix.

PR #1489, opened 2019-09-10
[NETBEANS-3075] - reduce number of warning w/ raw type iterator

BradWalker:
I've tried to reduce the amount of iterator raw type warnings that are emitted from
code that looks like this..

```
   [repeat] /home/bwalker/netbeans/enterprise/websvc.manager/src/org/netbeans/modules/websvc/manager/model/WebServiceGroup.java:102: warning: [rawtypes] found raw type: Iterator
   [repeat]             Iterator iter = listeners.iterator();
   [repeat]             ^
```

### DIFF
--- a/ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
+++ b/ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
@@ -149,8 +149,7 @@ public class DataObjectSyncSupport extends SyncSupport implements Synchronizator
             // check whether tree representation is loaded
             
             synchronized (reps) {
-                for (Iterator it = reps.iterator(); it.hasNext();) {
-                    Representation next = (Representation) it.next();
+                for (Representation next : reps) {
                     if (next.level() > 1) {
                         modelLoaded = true;
                     }                               
--- a/ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
+++ b/ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
@@ -115,10 +115,9 @@ public class DataObjectSyncSupport extends SyncSupport implements Synchronizator
 
         if (rep.represents(Document.class)) {
             synchronized (reps) {
-                for (Iterator it = reps.iterator(); it.hasNext();) {
-                    Representation next = (Representation) it.next();
+                for (Representation next : reps) {
                     if (next.represents(FileObject.class)) {
-                        it.remove();
+                        reps.remove(next);
                     }                               
                 }
             }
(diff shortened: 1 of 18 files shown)


ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
```diff
@@ -115,10 +115,9 @@ public void addRepresentation(Representation rep) {
 
         if (rep.represents(Document.class)) {
             synchronized (reps) {
-                for (Iterator it = reps.iterator(); it.hasNext();) {
-                    Representation next = (Representation) it.next();
+                for (Representation next : reps) {
                     if (next.represents(FileObject.class)) {
-                        it.remove();
+                        reps.remove(next);
```
matthiasblaesing: This is dangerous and will most probably blow at runtime. You are modifying the collection you are iterating, which will result in a `ConcurrentModificationException`.

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -2324,24 +2320,17 @@ private static boolean isArchiveFileImpl(final URL url, final boolean strict) {
...
+            res = new ProxyLookup(
+                Lookups.singleton(new JarArchiveRootProvider()),
+                Lookup.getDefault()).lookupResult(ArchiveRootProvider.class);
+            if (!archiveRootProviders.compareAndSet(null, res)) {
+                res = archiveRootProviders.get();
             }
-            archiveRootProviderCache = new LinkedList<>(res.allInstances());
         }
-        return archiveRootProviderCache;
+        return res.allInstances();
```
matthiasblaesing: This introduces a regression compared to https://github.com/apache/netbeans/commit/290420cfff1af3f72dd5eb86f864edd74ff78fe8

BradWalker: I think I properly reverted it.. Thanks for catching this..

ide/xml/src/org/netbeans/modules/xml/sync/DataObjectSyncSupport.java
BradWalker: I wasn't sure if that was correct or not.. It's in a synchronized code block, though I think the problem you described still exists because the iterator is unspecified if the underlying collection is modified while the iteration is in progress..

Anyways, I think I properly reverted it..

matthiasblaesing:
Sorry, but you'll need to review your git commits. Have a look at your commits with `gitk`, `git log -n 10 --graph --oneline`  or some other history visualizing tool. In a single PR I would expect a single string of commits (merging in a PR is _very_ rare). In this case it looks as if there are two very similar commits, that show small differences and are then merged.
It may help to remember how a github PR works: You create a branch with the code you want a project to merge. In this branch you add commits, that (in the best case) form logical groups of changes. If you want to update the PR, you can add commits to that branch. If multiple commits are done in the branch, that form one logical changeset, we can merge these commits at merge time.
If important changes from master are needed, you can merge or rebase. A rebase is more difficult, but creates a cleaner commit history and makes it easier to cleanup a PR at the end and spot the changes done to it.

matthiasblaesing:
Please also check your git settings. You switched identity from Brad to Charles (email).

PR #1490, opened 2019-09-10
Report all failed binaries for DownloadBinaries task

matthiasblaesing:
When a binary fails the hash check a BuildException is raised and the
build process is directly terminated. With this change, all binaries are
checked and all failures are reported. If one download fails, a
BuildException is raised after trying to download all binaries.


matthiasblaesing:
@jlahoda thank you for the review

PR #1491, opened 2019-09-10
[NETBEANS-2953] Fix for NPE after pressing 'DTDs and XML Schemas' men…

peedeeboy:
*What this pull request does*
Prevents Netbeans from crashing with a NPE when selecting Tools -> DTDs and XML Schemas

*Files Changed*
enterprise/payara.jakartaee/src/org/netbeans/modules/payara/jakartaee/layer.xml 
Changed so that the method used when registering is getDefaultRuntimeDDCatalog() instead of getRuntimeDDCatalog().  This now matches the equivilent method called when registering GlassFish DD catalog - getEE6RunTimeDDCatalog()

*Testing*
Navigate to Tools menu -> DTD and XML Schemas
Before this PR: NB will crash with a NPE
After this PR: The Enterprise Catalog dialog will open successfully


peedeeboy:
@jGauravGupta as this is Payara related, would you be kind enough to check you are happy with this very minor change? :)  cheers!

geertjanw:
Thanks, merging.

PR #1497, opened 2019-09-13
[NETBEANS-3072] Use actual ImageIcons as code completion actually like.

lkishalmi:
Most probably the regression was caused by [NETBEANS-2614](https://issues.apache.org/jira/browse/NETBEANS-2614).
This is a potentially harmless way to avoid class cast exception as for some reason Groovy editor requests ImageIcon-s

### DIFF
--- a/java/java.source/src/org/netbeans/modules/java/ui/Icons.java
+++ b/java/java.source/src/org/netbeans/modules/java/ui/Icons.java
@@ -19,15 +19,12 @@
 
 package org.netbeans.modules.java.ui;
 
-import java.awt.Image;
 import java.util.Collection;
 import java.util.Collections;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Modifier;
-import javax.swing.Icon;
 import javax.swing.ImageIcon;
 import org.openide.util.ImageUtilities;
-import org.openide.util.Utilities;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
I think @eirikbakke is in a better position to answer this, but my understand is, that using an ImageIcon forces you to use an Image as basis for your icon, you can't back it with other implementations. Using a plain Icon allows you to switch implementations under the hood without affecting the users.

Using an image for an icon is trivial, but an Icon could be drawn by java code or could be backed by an SVG that is scaled and rendered at the right size.

I had a quick look at this and it seems, that the problem can be tracked down into `org.netbeans.spi.editor.completion.support.CompletionUtilities#renderHtml` that method expects an `ImageIcon`. So to remove the requirement, the whole chain from `editor.completion` to the groovy editor would need to be fixed.

eirikbakke:
The patch looks fine. As long as you use ImageUtilities.loadImageIcon, the resulting icon will actually still support HiDPI scaling. (It's a special subclass implementation of ImageIcon.)

The ClassCastException was due to assuming that ImageUtilities.image2icon returns an ImageIcon, which is no longer the case.

There's a lot of places in the NetBeans codebase that require an ImageIcon specifically, and the new ImageUtilities implementation allows these to remain mostly untouched.

java/java.source/src/org/netbeans/modules/java/ui/Icons.java
```diff
@@ -44,64 +41,59 @@
...
-            return ImageUtilities.image2Icon (img);
-        }
+    public static ImageIcon getBusyIcon () {
+        ImageIcon icon = ImageUtilities.loadImageIcon(WAIT, false);
+        return icon;
     }
             
     
-    public static Icon getElementIcon( ElementKind elementKind, Collection<Modifier> modifiers ) {
+    public static ImageIcon getElementIcon( ElementKind elementKind, Collection<Modifier> modifiers ) {
```
eirikbakke: You can also remove the cast and "// todo: what happens, if i get a CCE here?" comment at https://github.com/apache/netbeans/blob/master/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionItem.java#L1006 .

lkishalmi: That change would require an API change and sigtest fails on that one. I'd left that out right now.

PR #1499, opened 2019-09-13
[NETBEANS-3082]: Fixed formatting issue happening on use of Switch-Ex…

arusinha:
…pression

Jira Link: https://issues.apache.org/jira/browse/NETBEANS-3082
Suppose We have below  Code Snippet

int a = switch(1)
{ default : yield 1 }
;
System.out.println(a);

Expressions/Statements used after Switch-Expression are getting wrongly indented.


geertjanw:
Thanks, merging.

PR #1503, opened 2019-09-14
[NETBEANS-2751] OpenJFX Samples

geertjanw:
This does the following:

- Integrates the OpenJFX Gluon samples correctly, using their new 0.0.3 release.
- Restructures the Samples category in the New Project dialog in "Java with Maven" and "Java with Ant", just like the project types were organized in that way in 11.1.
- Tries to hide the Ant-based JavaFX samples a bit, since we want to encourage the usage of Maven rather than Ant, i.e., in the same way as the project types at the top of the dialog prioritize Maven, now the samples do as well.
- Removes the module that was added in 11.1 that did not optimally integrate the OpenJFX Gluon samples. (Only the module needed to be removed, not the references to it in cluster.properties and elsewhere, since these were not in main to begin with, only in release111 and never moved to main.)


geertjanw:
Screen Shot 2019-09-14 at 20 39 33

That's the new structure in the Samples category from this pull request.

geertjanw:
This should go into 11.2, hope my adding the nb11.2 label helps to achieve that.

javafx/javafx2.samples/src/org/netbeans/modules/javafx2/samples/resources/layer.xml
```diff
@@ -24,351 +24,353 @@
...
-                        <attr name="template" boolvalue="true"/>
-                        <attr name="instantiatingIterator" newvalue="org.netbeans.modules.javafx2.samples.JavaFXSampleProjectIterator"/>
-                        <attr name="instantiatingWizardURL" urlvalue="nbresloc:/org/netbeans/modules/javafx2/samples/resources/WebViewBrowser.html"/>
-                        <attr name="prjType" stringvalue="javafx2"/>
-                    </file>
-
+                <folder name="Standard">
+                    <folder name="JavaFX">
+                        <attr name="position" intvalue="110"/>
+                        <attr name="displayName" bundlevalue="org.netbeans.modules.javafx2.samples.Bundle#Templates/Project/Samples/JavaFX"/>
```
matthiasblaesing: This bundle value does not exist anymore. See the change in Bundle.properties, line 25: https://github.com/apache/netbeans/blob/df832297fc4fac4d6f3549b8202564e56958f4e4/javafx/javafx2.samples/src/org/netbeans/modules/javafx2/samples/Bundle.properties#L25

geertjanw:
Thanks, fixed.

geertjanw:
Also thanks for liking and approving it.

geertjanw:
I don’t think this is a fundamental change, just a little bit of reorganization, i.e., a lot of debate and discussion is not needed for this one (correct me if I am wrong and reopen if applicable) and so I am merging this now.

JaroslavTulach:
Thanks for removing the ` javafx/openjfx.samples` module.

PR #1504, opened 2019-09-14
 [NETBEANS-3094] NBI cleaner process must attempt to remove read-only files too

lbruun:
Currently the Windows NBI cleaner process fails to delete files/folders if they have the read-only attribute set. With this change the cleaner will - in such case - attempt to remove the read-only attribute before proceeding to delete the file/folder.

I've deliberately not squashed as each commit is a story of its own.


geertjanw:
Looks good to me.

geertjanw:
Following request to not squash.

PR #1506, opened 2019-09-15
[NETBEANS-3096]:Updation for external nb-javac jar in libs.javacapi a…

arusinha:
…nd libs.javaimpl modules with nb-javac jar for jdk-13
Changes:
1. Updation of external jars of libs.javacapi & libs.javacimp modules.
2. Correction of Test cases of JavaCompetion for JDK-13 
Find below the list of test-cases which are failing with nb-javac (for jdk-13) but the same scenario is working in editor when used with nb-javac plugin for jdk-13
https://issues.apache.org/jira/browse/NETBEANS-3099

### DIFF
--- a/java/java.source.nbjavac/src/org/netbeans/modules/java/source/nbjavac/parsing/TreeLoader.java
+++ b/java/java.source.nbjavac/src/org/netbeans/modules/java/source/nbjavac/parsing/TreeLoader.java
@@ -101,7 +101,9 @@ public class TreeLoader extends LazyTreeLoader {
     private static final ThreadLocal<Boolean> isTreeLoading = new ThreadLocal<Boolean>();
 
     public static void preRegister(final Context context, final ClasspathInfo cpInfo, final boolean detached) {
-        context.put(lazyTreeLoaderKey, new TreeLoader(context, cpInfo, detached));
+        LazyTreeLoader instance = context.get(lazyTreeLoaderKey);
+        if (instance == null)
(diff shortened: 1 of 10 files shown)


java/java.source.nbjavac/src/org/netbeans/modules/java/source/nbjavac/parsing/TreeLoader.java
```diff
@@ -101,7 +101,9 @@
     private static final ThreadLocal<Boolean> isTreeLoading = new ThreadLocal<Boolean>();
 
     public static void preRegister(final Context context, final ClasspathInfo cpInfo, final boolean detached) {
-        context.put(lazyTreeLoaderKey, new TreeLoader(context, cpInfo, detached));
+        LazyTreeLoader instance = context.get(lazyTreeLoaderKey);
```
arusinha: This check is required otherwise we will get below error

	java.lang.AssertionError: duplicate context value
	at com.sun.tools.javac.util.Context.put(Context.java:141)
	at org.netbeans.modules.java.source.nbjavac.parsing.TreeLoader.preRegister(TreeLoader.java:106)
	at org.netbeans.modules.java.source.nbjavac.parsing.TreeLoader$TreeLoaderRegistryImpl.enhance(TreeLoader.java:579)
	at org.netbeans.modules.java.source.parsing.JavacParser.lambda$createJavacTask$10(JavacParser.java:796)

PR #1507, opened 2019-09-15
[NETBEANS-3085] Improving text block handling.

jlahoda:
Avoiding SourceUtils.isTextBlockSupported API. Some features seem to make sense even if text blocks are not enabled - e.g. having a text block already in the code, then lexing it as a text block seems fine. Type-completing code in such a way that the text block would be created is (probably) not fine when text blocks are created.

Also, fixing a problem in PHP with indenting the typed text, by introducing a special API method that preforms the indent, and leaving the existing methods work as before.


junichi11:
@jlahoda all php.editor unit tests passed with this changes. Thank you, Jan!

PR #1508, opened 2019-09-16
Improving code completion appearance for both LSP client and Java LSP…

jlahoda:
… server.

Some more to come after 11.2, but this could improve appearance for 11.2.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
@@ -37,8 +37,11 @@ import javax.swing.text.StyledDocument;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionItemKind;
 import org.eclipse.lsp4j.CompletionList;
+import org.eclipse.lsp4j.CompletionOptions;
 import org.eclipse.lsp4j.CompletionParams;
+import org.eclipse.lsp4j.MarkupContent;
 import org.eclipse.lsp4j.ParameterInformation;
+import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.SignatureHelp;
 import org.eclipse.lsp4j.SignatureInformation;
 import org.eclipse.lsp4j.TextDocumentIdentifier;
@@ -62,6 +65,7 @@ import org.openide.filesystems.FileObject;
 import org.openide.text.NbDocument;
 import org.openide.util.Exceptions;
 import org.openide.util.ImageUtilities;
+import org.openide.xml.XMLUtil;
 
 /**
  *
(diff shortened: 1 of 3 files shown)


ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
```diff
@@ -226,36 +223,66 @@ public void render(Graphics grphcs, Font font, Color color, Color color1, int i,
...
+                                                            content = new MarkupContent();
+                                                            content.setKind("plaintext");
+                                                            content.setValue(resolved.getDocumentation().getLeft());
+                                                        } else {
+                                                            content = resolved.getDocumentation().getRight();
+                                                        }
+                                                        switch (content.getKind()) {
+                                                            case "markdown":
+                                                            default:
+                                                            case "plaintext": documentation.append("<pre>\n").append(content.getValue()).append("\n</pre>"); break;
```
junichi11: all cases pass here?

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
```diff
@@ -226,36 +223,66 @@ public void render(Graphics grphcs, Font font, Color color, Color color1, int i,
...
+                                            resolved = temp;
+                                        } else {
+                                            resolved = i;
+                                        }
+                                        if (resolved.getDocumentation() != null || resolved.getDetail() != null) {
+                                            resultSet.setDocumentation(new CompletionDocumentation() {
+                                                @Override
+                                                public String getText() {
+                                                    StringBuilder documentation = new StringBuilder();
+                                                    documentation.append("<html>\n");
```
junichi11: should add `NOI18N`s?

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
junichi11: @jlahoda Is this intentional?

jlahoda: (I apologize for a late reply.) Yes, intentional for now. I have a an update for this, which actually converts the markdown to HTML (so that it can be shown nicely in the HTML pane), and also that converts the HTML to markdown for the server, but this involves a new external library, and it felt a little late in the release cycle to do that, so I opted for this simpler thing. The markdown should be still readable, and for the next release, it is really a few lines of changes here to do the conversion.

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
jlahoda: Probably yes, I'll try to remember that and add that on next update.

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CompletionProviderImpl.java
junichi11: No problem :) Thank you for your explanation.

PR #1509, opened 2019-09-16
[NETBEANS-3097]:Corrected font-color type name for text-block(JDK-13)

arusinha:


neilcsmith-net:
Out of interest, why is this capitalized with a space when the others are lower case and hyphenated?

geertjanw:
Agreed. Needs to be registered the same way as for all the others.

arusinha:
Hi @neilcsmith-net / @geertjanw  should I retain the old name text-block ?

neilcsmith-net:
@arusinha that depends what your reason for changing it is?  What does "corrected" mean?  To make it work or to make it look better?

arusinha:
Name was changed to make it look better

geertjanw:
The display name should be Text Block, the programmatic name text-block. See how it is done for the others.

jlahoda:
Should be possible to put text-block=Text Block to the Bundle.properties in the same directory.

neilcsmith-net:
Looks good to me.  No need for the cherry picking label - we no longer cherry pick!  When merged it will be in 11.2.

java/java.editor/src/org/netbeans/modules/java/editor/resources/fontsColors.xml
```diff
@@ -87,7 +87,7 @@
     <fontcolor name="mod-protected" />
     <fontcolor name="mod-public" />
     <fontcolor name="mod-keyword" default="keyword" />
-    <fontcolor name="text-block" bgColor="EEDDDD" />
+    <fontcolor name="mod-unindented-text-block" bgColor="EEDDDD" />
```
arusinha: had reverted back to old name "mod-unindented-text-block"

PR #1314, opened 2019-06-21
 [NETBEANS-1696] PHP 7.4 Support

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1696

## Typed Properties 2.0

See https://wiki.php.net/rfc/typed_properties_v2

- Fix ASTPHP5Parser.cup
- Fix MarkOccurrences, GotoDeclaration, and CodeCompletion features
- Handle unusable types(void and callable) as errors in UnusableTypesUnhandledError class to avoid complicating the grammar(cup file)
- Add and fix unit tests
- Fix FieldRedeclarationHintError (Add the isAnnotation() method to FieldElements)

## Spread Operator in Array Expression

See https://wiki.php.net/rfc/spread_operator_for_array

- Fix ASTPHP5Parser.cup
- Add UnpackableArrayElement as a new AST node (ArrayCreation holds ArrayElements as a list, so UnpackableArrayElement extends not Expression but ArrayElement)
- Fix PHP74UnhandledError
- Improve code completion for const statements (There was no proper context for them, so far, so add GLOBAL_CONST_EXPRESSION and CLASS_CONST_EXPRESSION as context)
- Add unit tests for CC, GotoDeclaration, MarkOccurrences, formatting, and parser

**NOTE** The following cases are not handled in the parser as errors:

- `[$a, ...$b] = [1, 2, 3];`
- `[...42];` `[..."string"];`
- `[...['a' => 'b']];`

We can handle these in a HintError class as errors if needed.

PR #1510, opened 2019-09-17
[NETBEANS-3103] Formatting is broken with class constant code completion(self::)

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-3103

This issue is related to #1314 .

- The parser ignores "::" because example code is incorrect syntax
- Check whether the token behind self or parent is "::" in the FormatVisitor

Example code:
```php
<?php
class ClassName {

    // ^ is the caret position
    // select "self::" or "parent::" in CC list
    // i.e. private const CONSTANT = self::;
    private const CONSTANT = ^;

}
```

Before:
```php
<?php
class ClassName {

    private const CONSTANT = self

    ::;

}
```

After:
```php
<?php
class ClassName {

    private const CONSTANT = self::;

}
```

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
@@ -914,6 +914,19 @@ public class FormatVisitor extends DefaultVisitor {
 
             }
             scan(node.getInitializers());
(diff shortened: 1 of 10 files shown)


junichi11:
@neilcsmith-net This is related to #1314. So we should fix this in NB11.2, I think. Is it OK?
@tmysik If it's OK with Neil and you approve it, please merge this.

BTW, https://issues.apache.org/jira/browse/NETBEANS-3085 is fixed. So, all unit tests pass.

Thanks!

tmysik:
Merging, thank you.

PR #1512, opened 2019-09-17
Update OpenJFX dependency to 13

matthiasblaesing:
OpenJFX 13 contains several bugfixes and at least one fix for a crasher bug:
https://bugs.openjdk.java.net/browse/JDK-8223377

In addition to bumping the version, this commit also exposes
the OpenJFX modules as user installable modules. This way
the end-user can easily install them, although Apache Netbeans
can't be bundled with OpenJFX (license: GPLv2-CPE).


matthiasblaesing:
@eppleton could you please have a look at this PR? You introduced the base idea of the module and so are best qualified to review. Thank you!

neilcsmith-net:
> OpenJFX 13 contains at least one known crasher issue:

I assume that means _fixes_ the issue?

> this commit also exposes the OpenJFX modules as user installable modules

I wonder if this is a fix, or at least a workaround, for an ergonomics bug where the front-end (html4j) project wizard remains unusable.  Can't find the JIRA issue for that now though!

matthiasblaesing:
Thank you for the heads up - yes the description was misleading (fixed). 

This will not fix installation problems, but I think I saw a similar problem. I think the problem is, that the modules depend on `org.netbeans.libs.javafx`, but that module is present, the modules though delegates either to the JDK internal JavaFX module or to the maven depedencies, if they are installed.

My change lessens the pain, because with this change, the JavaFX modules are visible in the plugin manager (You'll need to disable the update center with the name "Netbeans Distribution". The modules in the existing update center are marked to be hidden in the update manager and I assume, that this causes the modules to be hidden regardless of sources).

neilcsmith-net:
@matthiasblaesing yes, I think it was do with testing across different JDK versions, and the lack of visibility meant the module couldn't be enabled later, so this sounds like it should help with that.  I've definitely read a JIRA report from someone else that brings this up, but still can't find it now (I hate JIRA!)

Once the build branding stuff is fixed up, I'll merge this before creating the release branch for first beta.

PR #1513, opened 2019-09-17
[NETBEANS-3106] Add javahome property to groovyc task

graben:


neilcsmith-net:
@lkishalmi OK, so for 11.2?  Add milestone if you're happy with this.

graben:
@neilcsmith-net: Does that mean approval okay from your side?

lkishalmi:
I've added the milestone 11.2. I assume there are relative small audience of groovy projects, as the code completion bug for groovy arrived quite late after we broke that on 11.1. So let's have this in 11.2 beta 2.

graben:
@lkishalmi: 👍

PR #1514, opened 2019-09-17
[NETBEANS-3107] - cleanup use of generics

BradWalker:
I'm trying to clean up the use of generics when there is a private method or variable.

For example, we see a lot of the follow code in the source tree:

`private Map/*<Connection, Driver>*/ conn2Driver = new WeakHashMap();`

Change this to

`private Map<Connection, Driver> conn2Driver = new WeakHashMap<>();`

Java really wants us to use Generics and this looks like it was a work in progress. So I'm simply
trying to complete some of it.

### DIFF
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/results/BaseCallGraphBuilder.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/results/BaseCallGraphBuilder.java
@@ -42,7 +42,7 @@ public abstract class BaseCallGraphBuilder implements ProfilingResultListener, C
 
     //~ Instance fields ----------------------------------------------------------------------------------------------------------
 
-    protected List /*<Runnable>*/ afterBatchCommands = new ArrayList /*<Runnable>*/();
+    protected List <Runnable> afterBatchCommands = new ArrayList<>();
     protected ProfilingSessionStatus status;
     protected final Set cctListeners = new CopyOnWriteArraySet();
     protected WeakReference clientRef;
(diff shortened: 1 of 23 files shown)


profiler/lib.profiler/src/org/netbeans/lib/profiler/results/cpu/cct/CCTResultsFilter.java
```diff
@@ -46,7 +46,7 @@
     }
     
     public static interface EvaluatorProvider {
-        Set/*<Evaluator>*/ getEvaluators();
+        Set<Evaluator> getEvaluators();
```
matthiasblaesing: This touches the interface of the module, there are more places in the same code, that could get some generic love, but that is not "cleanup", so could you please revert this change?

profiler/lib.profiler/src/org/netbeans/lib/profiler/results/BaseCallGraphBuilder.java
```diff
@@ -42,7 +42,7 @@
 
     //~ Instance fields ----------------------------------------------------------------------------------------------------------
 
-    protected List /*<Runnable>*/ afterBatchCommands = new ArrayList /*<Runnable>*/();
+    protected List <Runnable> afterBatchCommands = new ArrayList<>();
```
matthiasblaesing: See comment in `CCTResultsFilter.java`.

BradWalker: I should have thought about that.. But, when I start making these mass changes, I kinda get into a rhythm.

And that's why we do code reviews..

Thanks for the code review!

sdedic:
Any objections to merge ?

neilcsmith-net:
@sdedic yes, big objection right now! We're in merge freeze for 11.2. It will be announced on dev@ when master is reopened.

matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/f2611e358c181935500ea4d9d9142fb850504a72. Thank you.

PR #1518, opened 2019-09-19
Fix Ant errors caused by ParseGitBranch task (Travis issue, etc.)

neilcsmith-net:
Update ParseGitBranch to assume master if more than one line of info returned from git branch. This happens in case of Travis CI tests, and is currently breaking them.  Commented out old code for now - should review again later.


neilcsmith-net:
@ebarboni Thanks!  Will do for now.  

Later we could also consider environment variables?  Travis at least has `TRAVIS_BRANCH` https://docs.travis-ci.com/user/environment-variables/#default-environment-variables  Does Jenkins give similar?

PR #1520, opened 2019-09-19
[NETBEANS-2940] Invalidate cacheed ArchiveRootProviders on change

matthiasblaesing:
This is a followup to commit:

290420cfff1a "[NETBEANS-2940] Cache ArchiveRootProviders"

The issue is:

archiveRootProviders.compareAndSet(null, res)

returns true, when the set was successful. In this case, the 
LookupListener needs to registered, that invalidates the cache list of
ArchiveRootProviders. In the original commit, this logic was inverted,
so the LookupListener was not installed and the cache never invalidated.

In addition the archiveRootProviderCache was made volatile to prevent
wrong caching across threads.

### DIFF
--- a/platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
+++ b/platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
@@ -2326,7 +2326,7 @@ public final class FileUtil extends Object {
                 res = new ProxyLookup(
                     Lookups.singleton(new JarArchiveRootProvider()),
                     Lookup.getDefault()).lookupResult(ArchiveRootProvider.class);
-                if (!archiveRootProviders.compareAndSet(null, res)) {
+                if (archiveRootProviders.compareAndSet(null, res)) {
                     res = archiveRootProviders.get();
                     res.addLookupListener((ev) -> {
                         archiveRootProviderCache = null;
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
The issue was reported here:

https://github.com/apache/netbeans/pull/1422#issuecomment-533055970

Debugging showed, that before the change in the steady state (after the IDE is full started),  getArchiveRootProviders()  holds two ArchiveRootProviders:
- org.openide.filesystems.JarArchiveRootProvider
- org.netbeans.modules.java.j2seplatform.platformdefinition.jrtfs.NBJRTArchiveRootProvider

while after the change from PR #1422 only the JarArchiveRootProvider is reported.

@arusinha could you please see if this PR fixes the issue for you?

lkishalmi:
Seems to be right. I'm just building the PR to test it out. The negation before compareAndSet was in the original implementation, that's probably why it was not that effective. I'll come back with the results soon.

lkishalmi:
@neilcsmith-net This is a critical fix, shall be in the first beta release of 11.2.

jlahoda:
+1

neilcsmith-net:
@lkishalmi "luckily" beta 1 is delayed due to a few issues with the new Jenkins release pipeline.  So, I'll merge this in.  (heads-up the Snap stuff with probably need reworking to use the new Jenkins build?)

PR #1521, opened 2019-09-19
Cleanup redundant casts

BradWalker:
The compiler is warning us we have many redundant casts.

They look like this.
```
[repeat] /home/bwalker/src/netbeans/platform/masterfs/src/org/netbeans/modules/masterfs/filebasedfs/naming/NamingFactory.java:253: warning: [cast] redundant cast to FileNaming
 [repeat] FileNaming cachedElement = (ref != null) ? (FileNaming) ref.get() : null;
 [repeat] 
```

So just simply clean it up..

### DIFF
--- a/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/ui/IpComboBox.java
+++ b/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/ui/IpComboBox.java
@@ -307,8 +307,7 @@ public class IpComboBox extends JComboBox<IpComboBox.InetAddr> {
         for (i = 0; i < count; i++) {
             InetAddr element = dataModel.getElementAt(i);
             // Passed IP address has highest priority.
-            if (((InetAddress) ip).equals(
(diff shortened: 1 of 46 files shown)


matthiasblaesing:
@BradWalker I'm currently working through this - one request: Don't merge branches after you created a PR (if not for a very good reason). I tried to rebase your patch and the merge of Erics branch makes it unnessary difficult.

matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/f1592df218c1add0f4088c0cb2f58eed02f649d4. Thank you.

I'm not sure if I asked about it in the past - you are doing valuable work and I understood, that you plan to continue. It would be helpful if you could get a ICLA on file. See here: https://www.apache.org/licenses/contributor-agreements.html

BradWalker:
Hey @matthiasblaesing thanks for the feedback! Also, I filed an ICLA this weekend, FYI.

matthiasblaesing:
@BradWalker great! Thank you.

PR #1522, opened 2019-09-20
Use json from nbbuild not from ide

ebarboni:
quick fix to use json lib from nbbuild and not from ide cluster.


neilcsmith-net:
@ebarboni as per Slack message, think correct fix is adding files to https://github.com/apache/netbeans/blob/master/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ignored-overlaps

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ignored-overlaps
```diff
@@ -34,6 +34,9 @@ ide/o.apache.commons.lang/external/commons-lang-2.6.jar java/maven.embedder/exte
 # upstream problem
 nbbuild/external/langtools-9.zip nbbuild/external/langtools-9.zip
 
+# Used to parse data during build, but 
+nbbuild/external/json-simple-1.1.1.jar
```
neilcsmith-net:
```
# List of patterns (one pair per line) to skip over in VerifyLibsAndLicenses.testBinaryUniqueness.
# Both sides of the pair must match for the pair to be skipped.
```
Think you need the other file location here too?!  Or is it working locally anyway?

PR #1524, opened 2019-09-20
 use URL information direct from JSON + verbose

ebarboni:
Fix for using URL directly + typo + information display


neilcsmith-net:
:+1: and thanks for the typo fix!

PR #1525, opened 2019-09-20
fix path for xml validation

ebarboni:
This should fix build for NetBeansTLP-master
Path were using flat structure instead cluster structure

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -322,13 +322,13 @@
     <target name="-validate-layers" depends="init">
         <parsemanifest manifest="${manifest.mf}" attribute="OpenIDE-Module-Layer" property="layer.file"/>
         <!-- DTD from sources -->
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
I don't know the full context, but the change looks sane to me.

ebarboni:
@matthiasblaesing it's for master build on jenkins that fail on a particular target.

neilcsmith-net:
@ebarboni while the fix looks good, I'm interested to know why this isn't causing a problem elsewhere (old Jenkins build / Travis / local)?!

ebarboni:
I'm not sure we use build-nbms build-source-zips build-javadoc often. But locally it fails using that parameter. Not sure what happens since release 111 that was building ok. But the location in the build file for the dtd are wrong since cluster refactor.

neilcsmith-net:
Yes, that's what bothers me! :smile: Still, let's merge this fix.

PR #1527, opened 2019-09-23
update path of gitinfo to be aside user.build.properties

ebarboni:
do not allow to commit gitginfo


neilcsmith-net:
@ebarboni I'm not sure this is exactly the right approach.  If someone checks out the sources with git, they want that file to be generated in `nbbuild/build` only, surely?  However, with the source zip, we want to provide the contents for that file for them.

So, perhaps we should write that file as part of the source zip at `nbbuild/gitinfo.properties`.

But the build should check if `nbbuild/gitinfo.properties` exists, and copy to `nbbuild/build/gitinfo.properties`. Then proceed as before - if `nbbuild/build/gitinfo.properties` doesn't exist, generate it.

That way, the source zip works, and a git checkout remains in sync.

Or am I missing something?

PR #1528, opened 2019-09-23
[NETBEANS-519] Update jgit

matthiasblaesing:
- Update/Add dependencies
- Fix API differences
- Fix Unittests


matthiasblaesing:
@ocafebabe I picked up your start of upgrading jgit and got all dependencies integrated, fixed all compile errors and parts of the unittest failures. I'm a bit stuck now, as it looks as if the StatusCommand shows broken behaviour. If your interested I would be grlade for a second pair of eyes.

ide/c.jcraft.jzlib/nbproject/project.xml
```diff
@@ -28,7 +28,7 @@
             <public-packages/>
             <class-path-extension>
                 <runtime-relative-path>com-jcraft-jzlib.jar</runtime-relative-path>
-		<binary-origin>external/jzlib-1.0.7.jar</binary-origin>
+		<binary-origin>external/jzlib-1.1.3.jar</binary-origin>
```
junichi11: There are TAB spaces.

ide/libs.git/src/org/netbeans/libs/git/jgit/commands/CherryPickCommand.java
```diff
@@ -426,14 +427,17 @@ protected boolean mergeTreeWalk (TreeWalk treeWalk, boolean ignoreConflicts)
                         throws IOException {
                     boolean ok = true;
                     boolean hasWorkingTreeIterator = tw.getTreeCount() > T_FILE;
+		    boolean hasAttributeNodeProvider = treeWalk.getAttributesNodeProvider() != null;
```
junichi11: There are TAB spaces.

nbbuild/cluster.properties
```diff
@@ -462,6 +463,8 @@ nb.cluster.ide=\
         o.eclipse.mylyn.wikitext.textile.core,\
         o.n.swing.dirchooser,\
         o.openidex.util,\
+	o.slf4j.api,\
+	o.slf4j.jdk14,\
```
junichi11: There are TAB spaces [L466, 467].

ide/libs.git/src/org/netbeans/libs/git/jgit/commands/GetCommonAncestorCommand.java
```diff
@@ -68,7 +68,7 @@ protected void run () throws GitException {
             throw new GitException(ex);
         } finally {
             if (walk != null) {
-                walk.release();
+                walk.close();
```
Saljack: You can use try-with-resources.
```java
   Repository repository = getRepository();
    if (revisions.length == 0) {
      revision = null;
    }
    else {
      try (RevWalk walk = new RevWalk(repository)) {
        List<RevCommit> commits = new ArrayList<>(revisions.length);
        for (String rev : revisions) {
          commits.add(Utils.findCommit(repository, rev, walk));
        }
        revision = getSingleBaseCommit(walk, commits);

      }
      catch (MissingObjectException ex) {
        throw new GitException.MissingObjectException(ex.getObjectId().toString(), GitObjectType.COMMIT);
      }
      catch (IOException ex) {
        throw new GitException(ex);
      }
    }
```

ide/libs.git/src/org/netbeans/libs/git/jgit/commands/GetPreviousCommitCommand.java
```diff
@@ -81,7 +81,7 @@ protected void run () throws GitException {
                 throw new GitException(ex);
             } finally {
                 if (walk != null) {
-                    walk.release();
+                    walk.close();
```
Saljack: You can use try-with-resources.

ocafebabe:
> @ocafebabe I picked up your start of upgrading jgit and got all dependencies integrated, fixed all compile errors and parts of the unittest failures. I'm a bit stuck now, as it looks as if the StatusCommand shows broken behaviour. If your interested I would be grlade for a second pair of eyes.

Hi @matthiasblaesing, do you still need help?

matthiasblaesing:
@ocafebabe yes help would still be appreciated - there are two known problems:
- the `testMoveFromSymlinkFolder` test still fails and I don't see why
- the versioning tests fails (https://www.travis-ci.org/matthiasblaesing/netbeans/jobs/589626983) as the git module can't be activated

If you have ideas how to fix it, it would be great.

matthiasblaesing:
@ocafebabe The problem of the versioning tests is fixed (dependencies were placed on the wrong module). That leaves the unittest `testMoveFromSymlinkFolder`.

matthiasblaesing:
Some more info: the problem seems to be, that the behavior of jgit regarding symlinks (and their normalization) changed. The Filter that is applied in `org.netbeans.libs.git.jgit.commands.MoveTreeCommand.run()` (line 88) does not exist in the git repository and thus the file can't be removed from the index. We would need to parse the path manually and resolve all symlinks.

The CLI git client has a very simple solution, it just fails ....

```
matthias@athena:~/src/netbeans/ide/libs.git/build/test/unit/work/o.n.l.g.j.c.R/testMoveFromSymlinkFolder/testMoveFromSymlinkFolder_wc$ git mv folder/folder/file tgt/file
fatal: not under version control, source=folder/folder/file, destination=tgt/file
matthias@athena:~/src/netbeans/ide/libs.git/build/test/unit/work/o.n.l.g.j.c.R/testMoveFromSymlinkFolder/testMoveFromSymlinkFolder_wc$ 
```

matthiasblaesing:
Ok - having written my previous comment, I decided to try a potential fix: resolve the symlinks prior to passing them to jgit (89e8ea8). It seems to work. I have some serious doubts, that it is a good idea to support this, but at least it works again now. Unittests are locally green now. I'd appreciate a critical look/further input/testing.

matthiasblaesing:
Rebased onto master and did the squashing. I plan to merge this mid next week if no-one objects till then.

ide/libs.git/src/org/netbeans/libs/git/jgit/commands/MoveTreeCommand.java
```diff
@@ -72,12 +70,14 @@ protected MoveTreeCommand (Repository repository, GitClassFactory gitFactory, Fi
 
     @Override
     protected void run() throws GitException {
+        File sourceFile = tryNormalizeSymlink(this.source);
+        File targetFile = tryNormalizeSymlink(this.target);
         if (!keepSourceTree && !after) {
             rename();
         }
+        sourceFile = tryNormalizeSymlink(sourceFile);
```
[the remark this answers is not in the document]
matthiasblaesing: I think I wanted to make sure, that is called in any case. The calls after the rename might suffice though. Would you mind keeping it as is?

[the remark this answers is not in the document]
Saljack: Yes. It just look weird for me on first sight.

ebresie:
Not sure if this is an issue or not but in my environment (i.e. relatively slow machine, slow network, etc) when using the 11.3-beta3 to pull the netbeans git code within Netbeans (assume with the jgit updates), it seems to make progress but after pulls, checks deltas, and then checking of files it seems to stall.  

I'm not sure if this helps but attached is my profile file that I had while running this.
[selfsampler.zip](https://github.com/apache/netbeans/files/4208736/selfsampler.zip)

matthiasblaesing:
I only had a quick look at the sample, but it looks as if the whole time is spend reading objects. The calls that really take the time are:

java.io.RandomAccessFile.readBytes[native] ()
java.io.FileDescriptor.close0[native] ()

I'm surprised especially about the second one. I would expect a close operation to be a very quick operation. This seems to be a windows system. Could you see if disabling the virus scanner helps? Maybe the taskmanager can show if there are other processes using the C: drive.

I did not observe problems on my older notebook. However this is on linux and with an SSD.

PR #1529, opened 2019-09-24
[NETBEANS-2914]: Elements.getPackageElement works better/safer when a…

jlahoda:
… ModuleSymbol is given.


neilcsmith-net:
@jlahoda intended for 11.2?

jlahoda:
If we can catch 11.2, yes, would be nice (it is a bugfix).

neilcsmith-net:
@jlahoda that's what I assumed.  Will be merged for beta2.  Any other PR intended for 11.2, please add me as reviewer and also the milestone.

PR #1530, opened 2019-09-24
Branding for upgrader and autoupdate

ebarboni:
not intended for beta1; just to show one option for the branding.

### DIFF
--- a/nb/autoupdate.pluginimporter/build.xml
+++ b/nb/autoupdate.pluginimporter/build.xml
@@ -22,4 +22,10 @@
 <project basedir="." default="build" name="nb/autoupdate.pluginimporter">
     <description>Builds, tests, and runs the project org.netbeans.modules.autoupdate.pluginsimporter</description>
     <import file="../../nbbuild/templates/projectized.xml"/>
+    <target name="compile" depends="projectized-common.compile,-branding"/>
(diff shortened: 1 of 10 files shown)


neilcsmith-net:
@ebarboni thanks for looking at this. But I think it might be better to leave to 11.3 timeframe.  We already have the required code for the release hardcoded in.

It's also worth having a conversation about whether we should automate the check and provide people a list of old installs to choose from, given people might decide to revert to LTS for some reason, or want to upgrade from beta to release. Did have brief conversation with @geertjanw about that during last release phase.

ebarboni:
@neilcsmith-net  do you think this is mergable to mimic the old way. 
Of course, need a new works to  for choosing from old install but that's another PR.

neilcsmith-net:
@ebarboni haven't tested, but looks a good plan for 11.3+! :+1:  Until final bug fixing in 11.2 I hadn't really considered how much is importable via `Tools / Options / Import` anyway.  Does that cover enough?  Not plugins at least.  Maybe longer term we enhance and open / point at that dialog instead?

ebarboni:
Well, I'm not the good user for that I'm not importing previous project. Maybe someone will have a nice workflow to propose later

PR #1534, opened 2019-09-27
[NETBEANS-3168] New Plugin Portal 11.2 Update Center for Apache NetBeans IDE 11.2

jkovalsky:
Adding new Plugin Portal Update Center specific for Apache NetBeans IDE 11.2 because of nb-javac plugin 1.72 which is incompatible with Apache NetBeans IDE 11.1.

Fixes #NETBEANS-3168

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -32,6 +32,8 @@ URL_Distribution=@@metabuild.DistributionURL@@
 #NOI18N
 URL_PluginPortal=@@metabuild.PluginPortalURL@@
 #NOI18N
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
This is no longer the way to make this change and merging it will break the way the build templating now works.  There should be no need to do anything other than add a website PR to change the redirect at https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess#L11

geertjanw:
Wow, interesting. :-)

geertjanw:
But we need two update centers to be registered for 11.2, i.e., how would we register both of these:

http://plugins.netbeans.org/nbpluginportal/updates/11.0/
http://plugins.netbeans.org/nbpluginportal/updates/11.2/

neilcsmith-net:
Two options here

- merge the catalogs into a single XML at http://plugins.netbeans.org/nbpluginportal/updates/11.2/ 
- we could add a PR (into master first) to add a separate UC for nb-javac from now on (@jlahoda might have a thought on that as we were talking about something similar this morning)

PR #1536, opened 2019-09-28
[NETBEANS-3040] Diff to feature does not work for Remote files in C/C++

SiddheshRane:
Diff file chooser converts editor FileObjects to java.io.File for selection.
In this process files with remote sources (rfs:/) become null.
This fix checks if selected file is not a disk file, providing alternate
means to retrieve the FileObject

This bug and fix is not related to C++ module; except that it manifests there.

### DIFF
--- a/ide/diff/src/org/netbeans/modules/diff/DiffAction.java
+++ b/ide/diff/src/org/netbeans/modules/diff/DiffAction.java
@@ -48,7 +48,6 @@ import org.netbeans.modules.diff.builtin.SingleDiffPanel;
 import org.netbeans.modules.diff.options.AccessibleJFileChooser;
 import org.openide.DialogDisplayer;
 import org.openide.ErrorManager;
-import org.openide.util.lookup.ProxyLookup;
 
 /**
  * Diff Action. It gets the default diff visualizer and diff provider if needed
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
Looks sensible.  As sane as they are(!), try not to make whitespace changes - makes seeing what's actually changed harder. @geertjanw pull in and test in next beta?

SiddheshRane:
White space changes occurred due to default Netbeans formatting. I missed them as my diff settings are set to ignore them. We need some standard netbeans based formatting for the whole project.

Also I have found a bug in this fix. If you choose a file from the editor buffer selector and then choose a different file from the file chooser, the new file does not get selected for diff.

Should I add the updated code to this pull request?

matthiasblaesing:
> White space changes occurred due to default Netbeans formatting. I missed them as my diff settings are set to ignore them. We need some standard netbeans based formatting for the whole project.

No this is not the netbeans default formatting. The whitespace changes are only introduced, if you set "Remove Trailing Whitespace From" to "All Lines" in the Options Dialog, Pane "Editor", Tab "On Save". The default is "None". I suggest to switch to "Modified Lines only", that way you don't get white-space changes all over the place, but only in locations you actually modified.

For the diff settings: Looks like netbeans indeed by default ingores leading/trailing white-space.

neilcsmith-net:
@SiddheshRane yes, please update your existing branch and keep in this PR.  We can squash and merge here when ready.  Possibly revert the non-relevant whitespace changes for now too?

SiddheshRane:
Squashed commits and reverted whitespace changes.

neilcsmith-net:
@SiddheshRane great, thanks.

PR #1537, opened 2019-09-29
Pack200 is deprecated for removal, disabling it by default when creat…

jlahoda:
…ing an NBM.

As pack200 is deprecated for removal, I don't think we can count on it being present. So I propose to change the default, and produce non-pack200 NBMs.

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -435,7 +435,7 @@
         <mkdir dir="${build.dir}"/>
         <property name="nbm.target.cluster" value=""/> <!-- fallback -->
         <property name="license.file.override" value="${license.file}"/>
-        <property name="use.pack200" value="true"/>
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
Sounds sane to me.

neilcsmith-net:
@jlahoda want me to pull in for this week's beta? Agreed sounds a sane move (and has the additional minor benefit of making it a little easier to peek inside them).

jlahoda:
Regarding the beta - for me, it would make sense to include this. But I'd like to give Jaroslav a chance to comment.

neilcsmith-net:
@jlahoda OK.  I generally wait for all reviewers anyway, so I won't merge yet and let @JaroslavTulach comment.  Would like to get it in beta2 if targetting 11.2 and not 11.3 though.

PR #1538, opened 2019-09-29
Putting nb-javac NBM directly into the NetBeans distribution.

jlahoda:
This is to ensure each build of NetBeans gets the version of nb-javac it needs, rather than something "unknown" from the plugin portal. (We could still distribute updates of nb-javac on an update center, although I suspect we will not.) This is similar to what is done for JavaFX.

One sharp edge is that nb-javac is not on maven central, so the download happens from netbeans.osuosl.org or hg.netbeans.org, which are probably less stable.

Comments?


geertjanw:
Excellent, I like it.

neilcsmith-net:
Thanks! Definitely a better way forward.  Some questions though -

1. Are we aiming for 11.2 with this?  If so, could we merge for this week's beta so we have a chance to fix or revert if any issues arise?
2. Can we then keep the existing plugin catalog from 11.0?  I assume this will override the nb-javac in that OK?
3. Are there any implications for shipping an update to nb-javac if it proves necessary?
4. @matthiasblaesing recently changed the OpenJFX setting for `AutoUpdate-Show-In-Client` to `true`. Do we want to make that change here too?  There are other reasons why it makes more sense there though, but it might be good to have it visible and controllable?
5. Am I reading this right that it will continue to always install in the userdir even if the installation directory is writeable?  (I think it should)
6. What are the feasibilities for actually moving this to Maven Central anyway?  That should be preferred if possible, and would be good to have the sources easily accessible alongside.  Or is it at least possible to have a source bundle adjacent to the binaries that we can link to for anyone who requests them?
7. (with some selfish reasons :wink:, although I might not be the only one - possibly CoolBeans?) What, if any, implications are there for this change on anyone who wants to pre-include nb-javac in their build?

jlahoda:
Answers to Neil's questions:
1. Not sure about 11.2, I think I'd welcome input on that (feels like a possibly too big change to do closely to release, so the existing approach, if it works fine, might be better/safer).
2. Yes, I believe we can keep the catalogs. These new modules have a newer version, so should be preferred.
3. I think we can produce an update in general, but there will be a problem with such updates if we only have a single update center (like plugin portal) for multiple releases. Not much different from the current state.
4. The nb-javac.api and nb-javac.impl modules are not shown in the UI, but the nb-javac module, which depends on them, is, and I believe that is correct (the user does not necessarily care about api/impl - they simply want "nb-javac" to be present).
5. Yes, I believe that's the current behavior.
6. I don't think I can comment on the Maven Central. We probably could add the source bundle alongside the binary.
7. Not sure if this will help much - some more changes might be necessary to help with this.

neilcsmith-net:
Thanks @jlahoda On 1, I'm happy with either - so let's see what other people think.  However, we do need _a_ fix for the nbjavac distribution by beta2 (Oct 2nd).  There will be at least one more beta (Oct 9th), so if we do decide to go with this approach, we should merge now and emphasise we want people to help test it in the announcement.  That mean there's still time to revert to the old approach.  If we don't hit beta2 with this then I'd definitely be in favour of holding it until 11.3.

On 3, that's true.  But we do have separate links in the IDE for each release, even though they currently redirect to the same place.

On 4, me being stupid, sorry!  Missed the other manifest.

On 7, yes, though I'll have to rethink my [current hack for this](https://github.com/praxis-live/praxis-live/blob/master/nbproject/platform.xml#L39)! After this PR we might want to think how downstream projects/distributions, and people building from source, can add these things in at build time ? No need for them to be restricted by ASF rules.

geertjanw:
Talked to Jan Lahoda this morning, he may have some thoughts before we merge this.

neilcsmith-net:
@geertjanw OK, will hold merge for now, but will need to know whether to proceed or take another approach for beta2 by the morning unless we delay it.  Testing locally good so far.

geertjanw:
It works for you, where are the JARs coming from?

neilcsmith-net:
What do you mean exactly?  Didn't check where it was downloading from.  Did check it _was_ downloading them.

geertjanw:
I want to be sure we’re aware where we’re getting them from, whether they’re the right version, and that we’re aware how to upload updates to them.

neilcsmith-net:
Presumably from the links at https://github.com/apache/netbeans/blob/a038116518f5dc94a4ad643d9d5122d8d1ebb4c6/extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external presumably in that order of preference?!  It is that file being downloaded.

See answer to my Q3 above - updates will still need to be done via an update centre, which does leave us potentially with the same issue we currently have unless we start having a plugin centre for each release.  That's one reason I pushed for separate links in each release even if pointing to same place, we can at least retrospectively apply that.

The other question might be if it's possible to update the code here via distribution update center to provoke an updated download?  Or can we add the external links to the distribution catalog file instead of the plugin catalog file?

geertjanw:
Well, having them on https://netbeans.osuosl.org is fine, I think.

PR #1540, opened 2019-09-30
Set buildnumber from metabuild info (branch and hash)

neilcsmith-net:
Set buildnumber to "${metabuild.RawVersion}-${metabuild.hash}" if both values have been computed to ensure consistency across builds of the same sources.

### DIFF
--- a/nbbuild/default.xml
+++ b/nbbuild/default.xml
@@ -162,7 +162,12 @@
     <tstamp>
     	<format property="buildday" pattern="d MMM yyyy" locale="en" />
     </tstamp>
-
+    <condition property="buildnumber" value="${metabuild.RawVersion}-${metabuild.hash}">
+        <and>
+            <isset property="metabuild.hash" />
+            <isset property="metabuild.RawVersion" />
+        </and>
+    </condition>
     <property environment="hudson"/>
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
@jlahoda could you check this meets any concerns you had in the thread at https://lists.apache.org/thread.html/c07d384c06beea76a9b2d0e36fbdef87bf5cc7d44eb7b0e18fbf0954@%3Cdev.netbeans.apache.org%3E ? If we're good, I'd really like to merge this today to meet beta2 build.  Thanks!

This will (should!), on master or any non-release branch give (except when detached head)

```
dev-<GIT_HASH>
```

On release branches eg.

```
11.2-beta2-<GIT_HASH>
```

ebarboni:
Seems to build yyyyMMdd-githash instead of expected dev-githash

neilcsmith-net:
@ebarboni that's the old default - date and shortened git hash. Will still fall back to that if the metabuild properties aren't available.  Can you check that's the case?

ebarboni:
metabuild.hash should be forced via -D
works in this case.

neilcsmith-net:
@ebarboni yes, both branch and hash should be set via properties if not being picked up from environment variables on the build system.  Hopefully #1541 will then get us into a position of parsing that from the .git folder, but I'll check and merge after beta2.  Also need to check this is correctly working from beta2 source bundle when ready.

PR #1541, opened 2019-09-30
Parse metabuild branch and hash from .git folder as fallback

neilcsmith-net:
I noticed while doing the buildnumber PR that the HgId task almost does what we need as a fallback to get branch and hash from .git folder.  I think this covers the cases where head isn't detached, and it's only needed for local checkout builds now anyway.


neilcsmith-net:
@ebarboni thanks!  I did try with setting environment variables locally to the same as Jenkins and seems to correctly prefer those.  Will wait until after beta2 to merge this anyway so it doesn't delay that if there is any issue on the build server.

neilcsmith-net:
@ebarboni will merge this now - time to check moving the branch/hash checks around is still picking up the right thing on Jenkins before next beta.

PR #1543, opened 2019-10-02
[NETBEANS-3193] Support remote files in Open Recent File menu

SiddheshRane:
This patch replaces system dependent file paths with URLs
and FileUtil.toFileObject with URLMapper.findFileObject to enable
opening remote files from Open Recent File menu

### DIFF
--- a/ide/utilities/src/org/netbeans/modules/openfile/RecentFileAction.java
+++ b/ide/utilities/src/org/netbeans/modules/openfile/RecentFileAction.java
@@ -27,14 +27,16 @@ import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
-import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.List;
 import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 import org.netbeans.modules.openfile.RecentFiles.HistoryItem;
 import org.openide.awt.*;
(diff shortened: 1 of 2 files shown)


SiddheshRane:
This commit changes the format of file paths in recent files list stored on disk. If a user updates this module, then **the first time none of the recent files will be able to open** as they are in old format. The recent list will get cleared. After that recent list will populate with newly closed files and work normally.

A smoother transition can be provided in which URLs are constructed with context `new URL(<url of C: or />, <url or file path in recent list>)`, although I am not sure whether it would be necessary

SiddheshRane:
File access was already present in the original code and also in the current releases of netbeans. The only change I have made is replace file paths with URL schemes.

I do think the overall history handling could benefit from code overhaul. But what alternative remains to support remote files? This may become more important, now that talks of a cloud based netbeans are ongoing.
Please suggest me if you have any workarounds.

matthiasblaesing:
Ok - for me local file access and network access fall into to different categories:

For local file access you can normally expect quick responses and thus get away with running it on the EDT. To be clear: It should not be done on the EDT (virus scanners, mounted network drives), but it can work.

For the remote part all bets are off. If you are on a high latency link round trip times in the range of seconds are possible and this is a no-go on the EDT.

The suggestion I have: Move things that may be caught doing high latency work into a background thread (RequestProcessor).

ebarboni:
@matthiasblaesing can we move to 12.0 ?

matthiasblaesing:
@ebarboni yes - I have removed the milestone, as said in the review for me having this on the EDT sounds dangerous, so this is work in progress.

Chris2011:
Hey @SiddheshRane are you still want to change the things as requested?

PR #1547, opened 2019-10-04
[NETBEANS-3164]:Fixed issue in Text-Block

Akshay-Gupta-Oracle:
[NETBEANS-3164]:Fixed issue in Text-Block

### DIFF
--- a/extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
+++ b/extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
@@ -1,4 +1,4 @@
 CRC:96561489
 SIZE:3411241
-URL:http://netbeans.osuosl.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
+URL:https://netbeans.osuosl.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
 URL:https://hg.netbeans.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
--- a/extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
+++ b/extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
@@ -1,4 +1,4 @@
-CRC:2001595618
-SIZE:3405169
-URL:https://netbeans.osuosl.org/binaries/58ECEA2987072A9AD96F2ECF82C18AB90CEF8CFD-nb-javac-13-impl.jar
-URL:https://hg.netbeans.org/binaries/58ECEA2987072A9AD96F2ECF82C18AB90CEF8CFD-nb-javac-13-impl.jar
+CRC:96561489
+SIZE:3411241
+URL:http://netbeans.osuosl.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
+URL:https://hg.netbeans.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
--- a/extra/nbjavac.api/release/modules/ext/nb-javac-13-api.jar.external
+++ b/extra/nbjavac.api/release/modules/ext/nb-javac-13-api.jar.external
@@ -1,4 +1,4 @@
-CRC:205892258
+CRC:4231515899
 SIZE:195677
-URL:https://netbeans.osuosl.org/binaries/E6F49BAA176938C82F4EBAA148F8A8AB13556233-nb-javac-13-api.jar
-URL:https://hg.netbeans.org/binaries/E6F49BAA176938C82F4EBAA148F8A8AB13556233-nb-javac-13-api.jar
+URL:https://netbeans.osuosl.org/binaries/8250FAEF60A423DA3D5EFA0EE699FB69D906D86C-nb-javac-13-api.jar
+URL:https://hg.netbeans.org/binaries/8250FAEF60A423DA3D5EFA0EE699FB69D906D86C-nb-javac-13-api.jar
--- a/java/libs.javacapi/external/binaries-list
+++ b/java/libs.javacapi/external/binaries-list
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-E6F49BAA176938C82F4EBAA148F8A8AB13556233 nb-javac-13-api.jar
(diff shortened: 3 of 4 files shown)


Akshay-Gupta-Oracle:
CRC generated using windows cygwin64.

Akshay-Gupta-Oracle:
JDK12 preview feature:
nb-javac jars are modified to fix compilation issue of text block in this particular scenario
https://issues.apache.org/jira/browse/NETBEANS-3164

graben:
What issue does this fix address? Still having errors with IDE running different JDK version than 13.

geertjanw:
Yes, you'll get an updated nb-javac in the next days, we're very aware of this issue, and are working on it.

graben:
That's why I'm asking. I manually switched the nb-javac impl in current beta2 to the one that this patch does link to. But it does not work better than before.

geertjanw:
You're going to need to sit tight a bit, we'll announce it when it's there.

extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
```diff
@@ -1,4 +1,4 @@
-CRC:2001595618
-SIZE:3405169
-URL:https://netbeans.osuosl.org/binaries/58ECEA2987072A9AD96F2ECF82C18AB90CEF8CFD-nb-javac-13-impl.jar
-URL:https://hg.netbeans.org/binaries/58ECEA2987072A9AD96F2ECF82C18AB90CEF8CFD-nb-javac-13-impl.jar
+CRC:96561489
+SIZE:3411241
+URL:http://netbeans.osuosl.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar
```
neilcsmith-net: Is the change from https to http intended here?!

geertjanw:
Most likely not intended.

extra/nbjavac.impl/release/modules/ext/nb-javac-13-impl.jar.external
Akshay-Gupta-Oracle: It does not matter

geertjanw: Yes, it does, we need to use https please.

Also, does this PR solve the problem of text block support not working if the JDK of NetBeans is not set to JDK 13?

Akshay-Gupta-Oracle: Yes after using new jar files, We can use nbjavac with JDK8 as base JDK of netbeans and JDK13 as platform for project. Textblock(which is preview feature of 13) will work fine.
nbfix 3164

geertjanw: Well, see above, @graben still has the same problem as before.

What is the version that one should see when installing the nb-javac that fixes this problem?

neilcsmith-net:
So, I'm assuming intended for NB11.2-beta3? Adding milestone - please remember to add for future PR intended for releases. Thanks!

With revised nb-javac support ( cc @jlahoda ) would we be able to push updates to these files via distribution update centre to trigger download of new external binaries if we need to update the library after release rather than falling back to plugin centre?

Akshay-Gupta-Oracle:
> That's why I'm asking. I manually switched the nb-javac impl in current beta2 to the one that this patch does link to. But it does not work better than before.

How did you switch it to new impl
In order to change the nb-javac-impl and nb-javac-api jars you have to follow following steps

First install nb-javac plugin and then restart NB
After that stop NB and then you have to replace the jar files (whose address can be found written in netbeans About->help->User directory, for windows it is C:\Users\<username>\AppData\Roaming\NetBeans\11.2-beta1 ) in that directory further go inside modules->ext there you can find the jars. replace the jar files with new downloaded jars( rename the downloaded jars by removing hash and removing -13- to match the name with old jars).

Then go back to location - C:\Users\<username>\AppData\Roaming\NetBeans\11.2-beta1
you can see .lastModified file, use following command there(from bash)
touch .lasModified
This will update the jars of nb-javac for netbeans

I did the same and after that text blocks are working for me with base jdk of net beans as JDK8.

graben:
I did it exactly this way but with JDK 11 as runtime JDK without success.

geertjanw:
Can you put here the exact URLs you used to get the two JARs?

graben:
Sure:
https://netbeans.osuosl.org/binaries/8250FAEF60A423DA3D5EFA0EE699FB69D906D86C-nb-javac-13-api.jar
https://netbeans.osuosl.org/binaries/344C8C2A8B421A52ABE725A677BF75659C17FEB6-nb-javac-13-impl.jar

graben:
Well, I did another clean install of beta2 with fresh profile and manual "patch" of nb-javac. Now it seems to work. Maybe my latest profile got buggy by manipulation. ;-)

geertjanw:
Excellent, thanks for the help in verifying this fix, much appreciated.

graben:
Welcome! But there are still funny behaviours in the editor for textblocks:
`public static String x = """test""";`
trying to correct with line break end in
`public static String x = """"`
`+ "test""";`
trying to remove the crappy forth " in the first line doesn't make it better :)

geertjanw:
Can you create a new issue for this?

graben:
Sure. https://issues.apache.org/jira/browse/NETBEANS-3202

neilcsmith-net:
@jlahoda what is the current spec version?  I currently see same version in NB 11.2-beta2 as NB 11.1.  Just tried with clean userdir and disabling plugin centre first - I don't see nb-javac at all, although I do see JavaFX modules available ???  Enabling Java cluster fails when plugin centre disabled.  That doesn't sound right?!

jlahoda:
Ok, I know now what is going on - the nb-javac modules are not part of the "3rdparty" cluster, and hence are not part of the source zip built by build-source-config (for cluster.config=release). And so they are not in the beta :-(. I didn't realize the cluster.properties need to change as well, and all is working for a build from the sources in git (as the modules are in the checkout). I apologize for that. 

I've submitted:
https://github.com/apache/netbeans/pull/1556

to add the modules to the 3rdparty cluster, which should ensure they are part of the source distro.

For the record, the spec versions of these modules is "2.0".

Akshay-Gupta-Oracle:
@jlahoda can this pull request be merge before #1556

neilcsmith-net:
@Akshay-Gupta-Oracle #1556 has been closed in favour of #1560 Perhaps coordinate there with @jlahoda to make sure that fix includes these changes?

jlahoda:
@neilcsmith-net are you OK with me merging this? (I'll then merge master into #1560.)

neilcsmith-net:
@jlahoda yes, said same on 1560. If you can merge them both, then I'll look at beta3 tomorrow. Thanks!

PR #1550, opened 2019-10-06
Retry mysql tests in Travis config

blackleg:
Change Travis config to retry mysql tests that fails randomly when run in Travis environment.
Use [travis_retry](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) command to retry the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -213,13 +213,13 @@ matrix:
           services:
             - mysql
           env:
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
@blackleg thanks!  Let's see if this improves things.  In fact, there are a couple of other tests that occasionally fail on network errors that might benefit from this.  Possibly review and improve further after NB 11.2.

PR #1551, opened 2019-10-06
Module list to log missing code base

eduardkareletc:
addresses issue NETBEANS-3206


junichi11:
It seems that this PR includes incorrect merges.
(not `apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ModuleList.java` but `apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ModuleList.java`)
You should sync your master branch with the master branch of apache/netbeans repo before you create a new branch for this pr.
Please add the issue number to the commit title.

Please read https://netbeans.apache.org/participate/submit-pr.html

BTW, 
```java
    public static String findText(Node parent) {
    	if (parent == null)
    		return null;
```
- Please add curly braces to the if statement
- Please add `@CheckForNull` if `null` is returned
- Please use 4 spaces instead of a TAB space
e.g. 
```java
    @CheckForNull
    public static String findText(Node parent) {
        if (parent == null) {
    	    return null;
        }
```
Maybe, you should create a new PR, then should close this.
Thanks.

eduardkareletc:
Rework of previous pull request to address NB3206, this time with correct files.

eduardkareletc:
I did find time to redo the changes in the way requested. Only the last two commits are relevant the earlier ones should be ignored. I have no idea how to get that done. I'll try to close this one and redo the PR.

jlahoda:
FWIW, regarding the null check in findText, I don't think the method was intended to handle nulls - rather, the caller should probably make sure it is not passing nulls there.

eduardkareletc:
The semantic definition of findElement() is to return null if the element is missing, findText does the same, returns null if there is no text.  There is no difference for the user when text is missing because the element is empty or because it does not exist. The proposed change brings consistency to that API. With that consistency a nested call to findElement can provide the parameter to  findText. That is what is used in the ModuleList class.

jlahoda:
Yes, I understand findElement() returns null if it cannot find the element - but most here methods don't silently accept nulls as parameters, I believe.

In any case, sorry I don't see this bringing consistency - with this patch, if I do "findText(findElement(...))", it won't fail with NPE, but if I do "findElement(findElement(...), ...)" it will, I think. dtto for findSubElements, although I could see an argument about the inconsistency being fine for findSubElements.

PR #1552, opened 2019-10-06
Use Material CSS to make the example look nicer

dukescript:
In order to make the Java Frontend Application Gradle project type more attractive, it would be better to make it less HTML-like, and more component like. Using Material CSS is a simple way to do it. 

Hopefully @geertjanw finds the new UI more attractive to shoot a video about the excellent support NetBeans provide when it comes to *JavaFX light* developement. @neilcsmith-net, can we get it into 11.2? It is just a simple HTML change, no code is affected. Thank you.

CCing @lkishalmi

### DIFF
--- a/groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/src_main_webapp_pages_index.html.fmk
+++ b/groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/src_main_webapp_pages_index.html.fmk
@@ -22,24 +22,59 @@
 <!DOCTYPE html>
 <html>
     <head>
-        <title></title>
+        <title>Java Frontend Application</title>
         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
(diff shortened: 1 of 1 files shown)


dukescript:
FYI: The application generated from the new project template is [available here](https://dukescript.com/javafxlight/) online.

groovy/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/src_main_webapp_pages_index.html.fmk
```diff
@@ -22,24 +22,59 @@
...
-            </li>
-        </ul>
+        <div class="navbar-fixed">
+            <nav>
+                <div class="naí-wrapper" >
+                    <a href="#" class="brand-logo center">Java Frontend Application</a>
+                    <ul id="nav-mobile" class="right hide-on-med-and-down">
+                        <li><a target="_blank" href="http://netbeans.org">NetBeans</a></li>
+                        <li><a target="_blank" href="https://github.com/dukescript/kt-mvvm-demo/tree/FXBeanInfo#readme">JavaFX</a></li>
+                        <li><a target="_blank" href="https://materializecss.com/">Material CSS</a></li>
```
neilcsmith-net: Shouldn't that say Materialize CSS ?  Otherwise happy to merge for 11.2.

PR #1555, opened 2019-10-07
NETBEANS-3185 use last maven-nbm-plugin

ebarboni:
This is last part of fix for NETBEANS-3185.  It's to use the 4.3 version of plugin
As vote is in progress it's not yet available.


Chris2011:
I think the vote was successful and on base of this vote, he created the PR.

PR #1556, opened 2019-10-07
Must include the extra/nbjavac* modules in the 3rdparty cluster, so t…

jlahoda:
…hey are part of the standard distro.

Otherwise, the nb-javac modules are not available for builds done from the actual source release.


jlahoda:
We could put these into rat exclusions, but the license info seemed better to me - these description files are under the Apache license, just cannot easily hold the license.

neilcsmith-net:
The JavaFX ones are excluded though. We should be consistent one way or another IMO.

matthiasblaesing:
Then please lets move the JavaFX ones to `licenseinfo.xml`. There is no real good reason not to state the license of the files. That way we don't need to discuss which license applies.

jlahoda:
Please wait a little - Travis is still failing validly on this.

neilcsmith-net:
@jlahoda I noticed - just retriggered to see, although looked like it might be a valid issue - complaining about one of the nb-javac jars.  Don't worry, I merge nothing until it's green - just gives me headaches!

neilcsmith-net:
@jlahoda can you clarify what being part of source build but not binary build means?  Concerned about us replicating a blocking issue that arose with JavaFX.

Out of interest, having looked at JavaFX files being in `external` rather than `release/modules/ext` is there a reason this one has to be done differently? Also, eg. https://github.com/apache/netbeans/blob/master/extra/libs.javafx.linux/nbproject/project.properties#L23

jlahoda:
Frankly, I am getting concerned as well. Maybe it is time to use the backup plan?

Overall, what I don't like about the way JavaFX is handled is that if I do a build from source, the JavaFX modules are installed already. (These are in the extra cluster, which is I hope not packed in the binary distro, but still.) Not sure if that's technically OK for JavaFX, but I can understand why Jaroslav prefers that for JavaFX - some stuff simply does not work without it. But for javac, I'd *really* prefer if it wasn't installed "automatically". I think users should be given a realistic chance to use the javac from JDK, even if they build from source.

Part of this is that build of e.g. extra/nbjavac.api will not download the nb-javac binary - the intent is to only create an NBM that will download the binary when installed.

So, we have a source distro, which includes the 3rdparty cluster, and the nbjavac modules must be part of that in order for them to be part of the distro. But building the source distro will build the 3rdparty cluster as well, and so would include the nbjavac modules, and I think it is preferred to not do that (because we would need to install the external nb-javac library for this to work correctly). So I tried to disable that build, although I agree that's ugly. I would prefer some cleaner solution, but I suspect that will take time (esp. if we try to include JavaFX as well). One possibility that comes to mind (and probably cleaner) would be to not put the nbjavac modules into the extra cluster, but rather embed them directly in some module (like nb/updatecenters itself). But I'd need to thing about that and discuss further.

Regarding external+mapping in properties vs. release - there is no reason why it would have to be done this way, but using release is cleaner (and works for all ant NetBeans projects, while the mapping in properties only works inside the main repo, AFAIK )

What was the blocking issue with JavaFX?

jlahoda:
Seems that trying to layer up workarounds to make things for sanely for modules in extra/3rdparty cluster will not work. https://github.com/apache/netbeans/pull/1560 strives to solve the issues in a different way.

neilcsmith-net:
@jlahoda totally understand your problem over the JavaFX approach.  That they get included when doing a straight build of the source is an annoyance here too.  We need some options to disable that - causes issues in testing releases as well.

However, the binary builds as done on the build server do not include the extra cluster at all.  The binary build is Apache compliant AFAIK.  And JavaFX must be explicitly enabled.  What we need to also do is document better how an end user can build that compliant binary locally.

However, for NB 11.2 I'd prefer we copy the JavaFX approach fully and then seek to address the above issues as soon as master reopens for 11.3?  Or revert and switch back to plan B and a temporary update centre?

I can't remember the exact issues with the JavaFX support.  @matthiasblaesing probably knows better.  I think we initially thought the binaries were being included in places they shouldn't, but was actually more to do with licenses showing up where they shouldn't?

matthiasblaesing:
> I can't remember the exact issues with the JavaFX support. @matthiasblaesing probably knows better. I think we initially thought the binaries were being included in places they shouldn't, but was actually more to do with licenses showing up where they shouldn't?

You are right. The binaries from the extra cluster were picked up and their license (GPLv2-CPE) was included in the LICENSE file although they were not included in the final zip. That is the benefit of #1560, as the binaries are never present at build time, we don't need a hack to exclude them.

neilcsmith-net:
@jlahoda thanks!  I see two test failures.  First is with RAT - think license-info files have wrong links for the jar.external files now?!  Second is ergonomics - will retrigger to make sure it's related - have seen that fail without cause a few times.

jlahoda:
I hope the rat/verify-libs-and-licenses problem might be gone in the last commit (and if not, that can be fixed, I think). But I am afraid I have no idea what's about the ergonomics failure. Likely another interaction with modules that are there, but should not?

neilcsmith-net:
Closing in favour of #1560

PR #1559, opened 2019-10-08
[NETBEANS-3218] Add inline hint support for literals used in constructors

graben:
@jlahoda : Plz have a look.

### DIFF
--- a/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
+++ b/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
@@ -1100,10 +1100,15 @@ public abstract class SemanticHighlighterBase extends JavaParserResultTask {
             }
 
             TreePath pp = getCurrentPath().getParentPath();
-            if (pp.getLeaf() != null &&
-                pp.getLeaf().getKind() == Kind.METHOD_INVOCATION) {
-                MethodInvocationTree inv = (MethodInvocationTree) pp.getLeaf();
-                int pos = inv.getArguments().indexOf(node);
+            Tree leaf = pp.getLeaf();
+            if (leaf != null &&
+                (leaf.getKind() == Kind.METHOD_INVOCATION || leaf.getKind() == Kind.NEW_CLASS)) {
+                int pos = -1;
+                if (leaf.getKind() == Kind.METHOD_INVOCATION) {
+                    pos = MethodInvocationTree.class.cast(leaf).getArguments().indexOf(node);
+                } else if (leaf.getKind() == Kind.NEW_CLASS) {
+                    pos = NewClassTree.class.cast(leaf).getArguments().indexOf(node);
+                }
                 if (pos != (-1)) {
                     Element invoked = info.getTrees().getElement(pp);
                     if (invoked != null && (invoked.getKind() == ElementKind.METHOD || invoked.getKind() == ElementKind.CONSTRUCTOR)) {
--- a/java/java.editor.base/test/unit/src/org/netbeans/modules/java/editor/base/semantic/DetectorTest.java
+++ b/java/java.editor.base/test/unit/src/org/netbeans/modules/java/editor/base/semantic/DetectorTest.java
@@ -258,46 +258,54 @@ public class DetectorTest extends TestBase {
         performTest("Test.java",
                     "package test;" +
                     "public class Test {" +
+                    "    public Test(String param) {" +
+                    "    }" +
                     "    public void api(String param1, int param2, int param3, float param4, Object... param5) {" +
                     "    }" +
                     "    private int getValue() {" +
                     "        return -1;" +
                     "    }" +
                     "    private void test() {" +
+                    "        new Test(\"\")" +
                     "        api(\"\", 2, getValue(), 1.0f);" +
                     "        api(\"\", 2, getValue(), 1.0f, null);" +
                     "        api(\"\", 2, getValue(), 1.0f, null, null);" +
                     "    }" +
                     "}",
                     "[PUBLIC, CLASS, DECLARATION], 0:26-0:30\n" +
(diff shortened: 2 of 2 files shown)


graben:
If okay I can extend this PR by adding inline hint feature for variables, too.

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
```diff
@@ -1103,9 +1103,14 @@ public Void visitLiteral(LiteralTree node, Void p) {
 
             TreePath pp = getCurrentPath().getParentPath();
             if (pp.getLeaf() != null &&
-                pp.getLeaf().getKind() == Kind.METHOD_INVOCATION) {
-                MethodInvocationTree inv = (MethodInvocationTree) pp.getLeaf();
-                int pos = inv.getArguments().indexOf(node);
+               (pp.getLeaf().getKind() == Kind.METHOD_INVOCATION || pp.getLeaf().getKind() == Kind.NEW_CLASS)) {
```
junichi11: Maybe, should fix the indent. Now, there are 3 spaces, but 4 or 8 spaces are better.

graben: Sorry, my fault. Corrected!

graben:
> -it is better to check for Tree (and Element and TypeMirror) types using getKind() == Kind., as is done in the existing test. The API does not guarantee that an AST node that is instanceof MethodInvocationTree will actually model a method invocation - a single implementation class may implement any number of the AST interfaces, and select its nature using getKind().

Well, I have to cast to specific class so testing with instanceof seams necessary to me. I can also check the kind but I don't think it will worth it?!

jlahoda:
Doing e.g. "tree instanceof MethodInvocationTree" is not required in the trees API. If "tree.getKind() == Kind.METHOD_INVOCATION", then the tree can be cast to MethodInvocationTree, otherwise, it should not be cast, even if it the cast itself might pass.

The instanceof ensures the instance in the "tree" variable can be cast to MethodInvocationTree, but it does not ensure the actual instance models the method invocation node, because the actual instance may be of e.g. type:
class X implements MethodInvocationTree, NewClassTree, ... {}

I.e. the actual implementation classes may implement any number of the API interfaces, so instanceof is unreliable. Tree.getKind() is the documented way to detect tree types.

graben:
IMHO it's unnecessary to check the kind at this moment. It's already checked in the if before. The only thing done is casting to a specific class (MethodInvocationTree or NewClassTree) and calling a method declared in those two interfaces. The check and cast is only needed to ensure runtime and compile time safety. Even if the current implementation of the tree will implement something crude like class X implements MethodInvocationTree, NewClassTree the code ensures that only the right kind of tree will reach this line of code and it can syntactically call getArguments() method.

jlahoda:
Yes, strictly speaking, in this particular case, this would work. But I am not sure what's the problem in simply using getKind() instead of instanceof? To the best of my knowledge, that's what the API expects (and this is not a NetBeans API, so we cannot change the approach here, just adhere or not adhere to it here). I.e. the idiom in this API is to check the Kind, and then cast, no instanceof - why is it necessary to deviate from it here?

graben:
Well, as you already said it's not save how the tree is really implemented. So the kind could return another type th an the cast would. In such a case an exception would be the result by unsafe cast. That's why I would prefer the instanceof check to save the cast would not throw an exception.

graben:
But I can combine to ensure both. :blush:

jlahoda:
Checking getKind() and then casting is generally a safe idiom for Trees, Elements and TypeMirrors. Using instanceof is generally not a safe idiom to use on these. There is some code in the IDE that was written before the API was redesigned this way, and these still use instanceof, but I was always hoping the use of instanceof on Tree would fade away over time, not increase.

(The instanceof works now only because both MethodInvocationTree and NewClassTree have getArguments() method, which is called in both braches. But if we ever need to call a different method in any of the branches, things may break. NewClassTree.getClassBody() looks like a possible candidate for a method that may need to be called at some point in one of the branches.)

graben:
@jlahoda : Did switch to check getKind instead of instanceof and enhance unit test. IMHO next should be to move code to its own method and also add to visitidentifier and visitmethodinvokation to make feature available for variables and method returns, too.

graben:
Once merged next PR https://github.com/graben/netbeans/commit/971888fd7412845996dab0e466cfe6799fcdf2dd

graben:
@jlahoda : Ready to merge? :-)

graben:
@jlahoda : Just another reminder.

GeertjanWielenga:
Hi @graben, thanks for your great work, have you provided an Apache iCLA for your contributions or would you have an issue if you were to be asked to provide one?

graben:
Hi @geertjanw, I have not provided an iCLA, yet. Not needed one in the past. :-) If you liked me to provide one, no problem.

graben:
@jlahoda, @GeertjanWielenga : Is there still something that might prevent merging?

matthiasblaesing:
To me this change looks sane - I don't quite get where the `prprts` comes from:

Bildschirmfoto_2019-12-14_13-03-07

According to the src.zip associated with the java platform of that project:

```
    /**
     * Creates an empty property list with the specified defaults.
     *
     * @param   defaults   the defaults.
     */
    public Properties(Properties defaults) {
        this.defaults = defaults;
    }
```

I would have expected `defaults`.

For the merging: Did you consider the ICLA? The ASF requires all contributions to be donated (see the header of the file you edited). For small contributions we can treat the PR as the acknowledgement of that requirment, but the ICLA makes it clear and easier in the future.

graben:
ICLA send to secretary. Example works for me:
image

jlahoda:
The "prprts" is surely a problem somewhere else - the source was apparently not consulted for the name, and so a default was used. Why that happened is not clear to me, but it does not have much to do with this patch.

Thanks for the work - I'll merge.

PR #1560, opened 2019-10-09
Making sure the internal nb-javac modules are part of the source distro

jlahoda:
This is another attempt to fix the missing nb-javac NBMs in the source distro, and replaces:
https://github.com/apache/netbeans/pull/1556

This time, the nb-javac modules are simply part of nb/updatecenters, i.e. not part of the main build at all. Not sure if this is viable for 11.2. Should pass Travis, unlike 1556.


geertjanw:
If this works and Neil approves, I approve. Thanks for the hard work on this issue.

neilcsmith-net:
I have to say I'd prefer the previous approach, and copying JavaFX approach for NB 11.2 given we're meant to be finalising last beta, with view to improving that in master after release?  I also have a concern that this has failed all Travis tests at the moment - will retrigger them now and see if it's a glitch.

neilcsmith-net:
Travis failures seem consistent.

matthiasblaesing:
For the build failure - this is the relevant part of the log:

```
BUILD FAILED
/home/travis/build/apache/netbeans/nbbuild/build.xml:1278: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nb/updatecenters/build.xml:105: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nb/updatecenters/extras/build.xml:24: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nb/updatecenters/extras/nbproject/build-impl.xml:62: Cannot find NetBeans build harness. 
Check that nbplatform.default.netbeans.dest.dir and nbplatform.default.harness.dir are defined. 
On a developer machine these are normally defined in ${user.properties.file}=${netbeans.user}/build.properties 
but for automated builds you should pass these properties to Ant explicitly. 
You may instead download the harness and platform: -Dbootstrap.url=.../tasks.jar -Dautoupdate.catalog.url=.../updates.xml
```

Indeed this is locally reproducible, by running `ant clean` after a successful prior build.

neilcsmith-net:
I would prefer using the approach with less changes from a release point of view, even if we revise afterwards, but will leave decision up to @jlahoda what you'd prefer in 11.2?

However, @matthiasblaesing not sure what you mean by more reproducible for end users?  Or zipping the netbeans file?  `ant build` will give a NetBeans zip that does not include the extras cluster.  We should probably rewrite the instructions at http://netbeans.apache.org/download/dev/index.html to specify `ant build` and not to use the `nbbuild/netbeans` folder.  There are other differences from the released binaries.

We could probably fix up `ant tryme` to not use extras by default - remove from the clusters file and pass as an optional extra cluster to the launcher via ant?

jlahoda:
I was actually working on replicating what JavaFX is doing, but I currently ended up with a license file that includes GPL, and didn't have time to try to solve that.

matthiasblaesing:
@jlahoda the code that excludes the extra cluster from license can be seen here: https://github.com/apache/netbeans/pull/1143/files

@neilcsmith-net What I meant to say above is, that if the "netbeans" folder of a local build is closer to the zip we distribute, the final distributed structure is more thoroughly tested. The local build of netbeans currently includes the OpenJFX libraries and does not install them. This does not replicated the experience of an end user.

neilcsmith-net:
@matthiasblaesing I'm sure we can change even your habits! :wink:

jlahoda:
I've pushed an update to https://github.com/apache/netbeans/pull/1556, which (if it passes Travis) uses the same approach as JavaFX for nb-javac. Had to do changes to the filtering, because otherwise CreateLicenseSummary was picking up the javac jars from libs.javacapi/libs.javacimpl (which are not part of the distribution, and are not part of the netbeans directory), and mixing them with the jars in the extra cluster.

But, to me, that is at most for 11.2 - to me, it feels like we are adding workaround for something that should not need workarounds. The approach in this PR is much saner, IMO.

As for excluding the extra cluster from ant tryme, etc. - traditionally, many (experimental, prototype, unstable, etc.) modules were compiled in the extra cluster. We don't have any like that right now, but closing the door to have them does not sound necessary. Not having a "3rdparty" cluster (mapped to "extra" folder) that is half in the distro and half isn't sounds like a much better direction.

And I think I agree with Matthias that the closer the outcome of running "ant"  is to running "ant build-source-config && <unpack> && ant build && <unzip>", the better. After all, the issue here is that I have tested the result of "ant", not the result of the build after the full loop.

neilcsmith-net:
> And I think I agree with Matthias that the closer the outcome of running "ant" is to running "ant build-source-config && && ant build && ", the better. After all, the issue here is that I have tested the result of "ant", not the result of the build after the full loop.

We're all in agreement there, but something for post 11.2.  I'm kind of neutral on approach - pros and cons to both, and possible to make either work, but lot of sense to what you're saying here.

I understand the history of the extra cluster, and it does make sense to separate.  But, whatever happens with this it would still be good to make it opt-in so anything in it isn't run (or built?) by default.  I'm not saying close the door to it - I'm suggesting making `ant tryme` and running the built zip more equivalent.

And from an end-user perspective, I'd prefer `ant` was at least equivalent to `ant build` and we stopped telling people that the main build output is `nbbuild/netbeans` as opposed to the `NetBeans-***-zip` - even if content except for extra isn't that different.

jlahoda:
I am still not sure what's the problem in https://github.com/apache/netbeans/pull/1556 - but this patch IMO leads to less workarounds on various places, so while bigger, feels safer to me.

neilcsmith-net:
Thanks @jlahoda So, let's go with this and close #1556.  However, this will also break #1547 I think?  Do you want to integrate the changes from that into this before I merge?

Later, once NB 11.3 window opens be good to bring JavaFX to match this as @matthiasblaesing suggested.  It would also be good if we can find a way to make this optional so that someone can instead choose to build from source and get the plugins pre-installed - no reason a downstream user, distribution or project need be hampered by the same restrictions we are with binaries (see recent user@ education question).

jlahoda:
Would it make sense to merge 1547, I would then merge master here and then we could merge this? Should be only a few hours for me for the rebase.

neilcsmith-net:
Up to you - your changes here move those files I think, so wasn't sure of consequences of that.  Happy for you to handle merging both to master as best works for you.

PR #1561, opened 2019-10-09
Always specify javac.target

JaroslavTulach:
Always specify javac.target to allow compilation on JDK11+ and execution on JDK8. Once I compiled NetBeans with JDK11:
```
nbbuild$ ant build -Dnbjdk.home=/jdk11
```
but then I wanted to compile again with JDK8
```
nbbuild$ ant build -Dnbjdk.home=/jdk8
```
it failed with "Incompatible class file error". In general we have to specify `target="${javac.target}"` to prevent JDK11+ to use too modern bytecode formats. This PR contains two places where the missing `target` was causing problems in my case.

### DIFF
--- a/ide/db/build.xml
+++ b/ide/db/build.xml
@@ -32,7 +32,7 @@
 
     <target name="compile-lib" depends="init,fake-jdbc-40">
         <mkdir dir="${build.dir}/lib-classes" />
-        <javac srcdir="libsrc" destdir="${build.dir}/lib-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6">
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
@JaroslavTulach we are in feature freeze, merges to master should have be cleared with @neilcsmith-net before merging.

JaroslavTulach:
Opps. I haven't realized that. Sorry, feel free to backout and I'll be more careful next time.

neilcsmith-net:
@JaroslavTulach no problem!  This looks like it's probably safe enough to leave in.

Out of interest, should we be considering using the `release` attribute as well in some of these?

PR #1562, opened 2019-10-10
[NETBEANS-3515] NP-38 Payara Server Hot reload support

jGauravGupta:
Payara Server `redeploy` asadmin command supports Hot Deploy feature via new `--hotDeploy` flag which is intended for machine-to-machine communication (IDE, build tools etc).

PR integrates with `redeploy` asadmin command which can be enabled in the properties panel of server:
HotReload

**Commits related to Core enterprises modules:**
- Prevent duplicate recording: https://github.com/apache/netbeans/pull/1562/commits/93dd854a0f81410c604208762570cbb6ded6085c
- Record deleted files: https://github.com/apache/netbeans/pull/1562/commits/b0f8c9e424da1735ecc7d51b17e7fdfe3a43359a
- Improve deploy on save handling: https://github.com/apache/netbeans/pull/1562/commits/97408cd86b9bd7faf97a49b12488f4ddc0f5272b
- Deleted classes not removed from maven web project target/${artifactId-version}/WEB-INF/classes directory:
https://github.com/apache/netbeans/pull/1562/commits/ed25c5bb60b86ced06d435b5a399fec5b88302cd


junichi11:
I have a question. What does `NP-38` mean? Issue number for Payara support? Payara support was already donated to Apache NetBeans, wasn't it? If so, I think we should use the issue number of JIRA for NetBeans. i.e. https://issues.apache.org/jira/projects/NETBEANS/issues

jGauravGupta:
Hi @junichi11,

Yes NP-38 is an internal Jira ticket id that is used internally in Payara Ecosystem projects & helps in Sprint planning.

junichi11:
@jGauravGupta Thank you for your answer :)
I'm not sure whether that's OK. Could you please ask via the mailing list about using an issue id of another JIRA without issue id of NetBeans JIRA on commit messages?

I'm worried about the following:
- Payara modules are already in NetBeans repo.
- Other developers can't understand what that number means.
- Payara team can look at the issue, but other than those people can't do it.
- After another module(e.g. MyModule) is donated in the future, if the author uses the issue number(e.g. [MYMODULE-55]) of another bug tracker, maybe it would be the cause of confusing.

enterprise/payara.common/src/org/netbeans/modules/payara/common/CommonServerSupport.java
```diff
@@ -710,7 +712,9 @@ public boolean isRemote() {
             return ServerAdmin.<ResultString>exec(instance, new CommandRedeploy(
                     name, Util.computeTarget(instance.getProperties()),
                     contextRoot, properties, libraries,
-                    url != null && url.contains("ee6wc")), stateListener);
```
matthiasblaesing: What is `ee6wc` and `deployer:pfv`? Some comment would help do understand this.

enterprise/payara.common/nbproject/org-netbeans-modules-payara-common.sig
```diff
@@ -1803,7 +1805,7 @@ meth protected boolean getJdbcDriverDeploymentProperty()
 meth protected boolean getLoopbackProperty()
 meth protected boolean getPreserveSessionsProperty()
 meth protected boolean getStartDerbyProperty()
-meth protected void store(boolean,boolean,boolean,boolean,boolean,boolean,org.netbeans.modules.payara.common.PayaraInstance)
+meth protected void store(boolean,boolean,boolean,boolean,boolean,boolean,boolean,org.netbeans.modules.payara.common.PayaraInstance)
```
matthiasblaesing: This is exported API (fried-only) was it checked, that all friends are compatible?

enterprise/payara.tooling/nbproject/org-netbeans-modules-payara-tooling.sig
```diff
@@ -431,7 +431,7 @@ supr org.netbeans.modules.payara.tooling.admin.Command
 hfds COMMAND
 
 CLSS public org.netbeans.modules.payara.tooling.admin.CommandRedeploy
-cons public init(java.lang.String,java.lang.String,java.lang.String,java.util.Map<java.lang.String,java.lang.String>,java.io.File[],boolean)
+cons public init(java.lang.String,java.lang.String,java.lang.String,java.util.Map<java.lang.String,java.lang.String>,java.io.File[],boolean,boolean)
```
matthiasblaesing: Same comment as in `payara.common` do friends still work?

enterprise/payara.common/nbproject/org-netbeans-modules-payara-common.sig
jGauravGupta: `org.netbeans.modules.payara.common.ui.InstancePanel#store` is part of `payara.common` module and not used by any other modules.

enterprise/payara.tooling/nbproject/org-netbeans-modules-payara-tooling.sig
jGauravGupta: `org.netbeans.modules.payara.tooling.admin.CommandRedeploy#init` is part of `payara.tooling` module and only used by `payara.common` module.

enterprise/payara.common/src/org/netbeans/modules/payara/common/CommonServerSupport.java
jGauravGupta: That's URI fragment, see here: https://github.com/apache/netbeans/blob/master/enterprise/payara.common/src/org/netbeans/modules/payara/common/ServerDetails.java#L44-L250
URI fragment is assigned to each version of Payara Server e.g: deployer:pfv4ee7, deployer:pfv5ee8 etc, where pfv4 stands for Payara Fish version 4 and ee7 stands for supported Java EE version. Hence here checking for Payara uri fragment instead of a specific version.

PR #1563, opened 2019-10-10
NETBEANS-3185 use last maven-nbm-plugin

ebarboni:
Archetypes need to reflect the 4.3 changes too.

Need to wait for vote. 

Note sure on how to simplify the workflow yet.


ebarboni:
vote is ok.

PR #1564, opened 2019-10-10
[NETBEAN-3230] - remove hg-clean target from build.xml

BradWalker:
The Netbeans project no longer uses Mercurial as it's primary source code control system. It's now all managed in a Git repo. Therefore, we no longer need to carry around the technical debt in build.xml..

For example, the clean target  has the following

`     <try-else first="-hg-clean" second="-real-clean"/>`

Yet, "-hg-clean" has the following fail and never will get executed..

`    <fail message="Not running in a Mercurial checkout."> `

So it can safely be removed..

Before

```
....
[releasejson] Branding computed info: metabuild.logcli=-J-Dnetbeans.logger.console=true -J-ea

clean:

-hg-clean:
 [try-else] /tmp/netbeans/nbbuild/build.xml:1343: Not running in a Mercurial checkout.
 [try-else] Target -hg-clean failed, falling back to -real-clean

-jdk-pre-preinit:

-jdk-preinit:
...
```

After

```
[releasejson] Branding computed info: metabuild.logcli=-J-Dnetbeans.logger.console=true -J-ea

clean:

-jdk-pre-preinit:

-jdk-preinit:

```

### DIFF
--- a/nbbuild/build.xml
+++ b/nbbuild/build.xml
@@ -1279,9 +1279,7 @@ It is possible to use -Ddebug.port=3234 -Ddebug.pause=y to start the system in d
   </target>
 
   <target name="clean" depends="-init-clean,bootstrap" description="Clean everything possible.">
-      <taskdef name="try-else" classname="org.netbeans.nbbuild.TryElse" classpath="${nbantext.jar}"/>
-      <property name="vanilla.javac.exists" value="true"/>
-      <try-else first="-hg-clean" second="-real-clean"/>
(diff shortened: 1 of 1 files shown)


JaroslavTulach:
Isn't there a `git clean` equivalent of `hg clean`? If so, it might make sense to use it. But until it is found, removing the outdates Hg related code is a good step anyway. CCing @jglick

BradWalker:
I'll work on looking into `git clean`..

PR #1566, opened 2019-10-13
Default JDK for Maven projects

jtulach:
I am running NetBeans on JDK13 (to benefit from the great work Oracle guys put into `javac` in JDK13 and its co-operation with NetBeans).

However I don't want to develop my own code on such bleeding edge platform. I usually want my Maven projects to compile on JDK8, sometimes on JDK11. Never on JDK13.

There is a way to select non-default JDK per Maven project, but I'd like to do that for all projects at once. This PR solves it - it allows one to specify "Default JDK" in Maven settings and that one is then used for each Maven project without specifically selected JDK.


neilcsmith-net:
@JaroslavTulach added 11.3 milestone - let's keep this one out of master until 11.2 is complete please! :wink: 

Great addition though!  Would find this useful at the moment.

(@geertjanw I think this is solving the reverse issue)

PR #1570, opened 2019-10-14
[NETBEANS-3202] Fix issue in Text Block

Akshay-Gupta-Oracle:
[NETBEANS-3202] Fix issue in Text Block

### DIFF
--- a/java/java.editor/src/org/netbeans/modules/editor/java/TypingCompletion.java
+++ b/java/java.editor/src/org/netbeans/modules/editor/java/TypingCompletion.java
@@ -547,6 +547,11 @@ class TypingCompletion {
         return posWithinQuotes(doc, caretOffset, JavaTokenId.STRING_LITERAL);
     }
     
+ static boolean posWithinTextBlock(Document doc, int caretOffset) {
+        TokenSequence<JavaTokenId> javaTS=javaTokenSequence(doc,caretOffset, false);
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
If intended for 11.2(?), please add milestone and add me as reviewer when ready.  That added if statement probably needs formatting and braces too?

junichi11:
@Akshay-Gupta-Oracle Please use `[NETBEANS-XXXX]`(e.g. `[NETBEANS-3202]`) instead of `[Netbeans:3202]` because the link of this PR is added to the JIRA.

java/java.editor/src/org/netbeans/modules/editor/java/TypingCompletion.java
```diff
@@ -547,6 +547,11 @@ static boolean posWithinString(Document doc, int caretOffset) {
         return posWithinQuotes(doc, caretOffset, JavaTokenId.STRING_LITERAL);
     }
     
+ static boolean posWithinTextBlock(Document doc, int caretOffset) {
+        TokenSequence<JavaTokenId> javaTS=javaTokenSequence(doc,caretOffset, false);
+        javaTS.movePrevious();
```
sdedic: check if `javaTS` is not null and that `javaTS.movePrevious` actually didn't return `false` (there would be no token for `javaTS.token()`

PR #1571, opened 2019-10-14
[NETBEANS-3198] fallback to hard-coded build date in Options API

neilcsmith-net:
Hard-code the date (201910010000) when we switched from build number with date to build number with git hash as quick fix for Options Import / Export


neilcsmith-net:
Seems to work locally, although there is a slightly odd difference in module implementation versions locally.  I'll need to test again once beta3 is built on Jenkins.

PR #1572, opened 2019-10-15
[NETBEANS-3240] - more cleanup of raw type warnings..

BradWalker:
Clean up a bunch of raw type warnings..

Things like this:

```
   [repeat] /home/bwalker/src/netbeans/ide/spi.debugger.ui/src/org/netbeans/modules/debugger/ui/actions/BreakpointCustomizeAction.java:124: warning: [rawtypes] found raw type: Class
   [repeat]             Class biClass;
   [repeat]             ^
   [repeat]   missing type arguments for generic class Class<T>
   [repeat]     T extends Object declared in class Class
```


matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/33f8ee20c31f67ff0e449f6d21575d1d29a28f38. Thank you.

PR #1574, opened 2019-10-17
[NETBEANS-3243] Prioritize "--standard" parameter of the path of Code Sniffer

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-3243

#### Before:
If `/path/to/phpcs --standard=/path/to/ruleset.xml -s` is set to the path of Code Sniffer, `--standard` parameter is added twice.
```
"/usr/bin/php" "/path/to/phpcs" "--standard=/path/to/ruleset.xml" "-s" "--standard=MySource" "--report=xml" "--extensions=phtml,php,php4,php5,php3,phpt,inc" "--encoding=UTF-8" "/path/to/target.php"
```

#### After:
If the path have that parameter, just use only it.
```
"/usr/bin/php" "/path/to/phpcs" "--standard=/path/to/ruleset.xml" "-s" "--report=xml" "--extensions=phtml,php,php4,php5,php3,phpt,inc" "--encoding=UTF-8" "/path/to/target.php"
```

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodeSniffer.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodeSniffer.java
@@ -64,7 +64,8 @@ public final class CodeSniffer {
     static final File XML_LOG = new File(System.getProperty("java.io.tmpdir"), "nb-php-phpcs-log.xml"); // NOI18N
 
     // #270987 use --standard instead of --runtime-set default_standard
-    private static final String STANDARD_PARAM = "--standard=%s"; // NOI18N
+    private static final String STANDARD_PARAM = "--standard"; // NOI18N
+    private static final String STANDARD_PARAM_FORMAT = STANDARD_PARAM + "=%s"; // NOI18N
     private static final String RUNTIME_SET_PARAM = "--runtime-set"; // NOI18N
     private static final String DEFAULT_STANDARD_PARAM = "default_standard"; // NOI18N
     private static final String LIST_STANDARDS_PARAM = "-i"; // NOI18N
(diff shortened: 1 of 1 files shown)


junichi11:
@tmysik Could you please review this although we can't merge this into the master branch yet?

neilcsmith-net:
@junichi11 I'll reopen master in a couple of days. Need to merge #1594 and then increment spec versions first.

junichi11:
@neilcsmith-net I see. Thank you for the information :)

php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodeSniffer.java
```diff
@@ -237,8 +238,11 @@ public void run() {
     private List<String> getParameters(String standard, FileObject file, boolean noRecursion) {
         Charset encoding = FileEncodingQuery.getEncoding(file);
         List<String> params = new ArrayList<>();
-        // #270987 use --standard
-        params.add(String.format(STANDARD_PARAM, standard));
+        // NETBEANS-3243 the path of Code Sniffer may have --standard parameter
+        if (!codeSnifferPath.contains(STANDARD_PARAM)) {
```
tmysik: Nitpick: A bit more precise/safe would be to check for `--standard=` to avoid confusion with (possible future?) parameters like `--standard-log-file` etc.

junichi11: Yes, I'll fix it.

junichi11: Fixed: https://github.com/apache/netbeans/compare/7fdb5a05aa6b55d6f2f65731b3e192bdcd24897f..2f983b95ea6878c8d027a65caf75ce431c9b012e

tmysik: Nice, thanks. BTW I am not sure whether the command cannot be `[...] --standard <something> [...]` (I mean using ` ` instead of `=`).

junichi11: Maybe, you are right. We should check both cases, just in case. Will fix it.

junichi11: Fixed.

junichi11:
@neilcsmith-net Can we merge PRs for 11.3? Or, do we have to wait for #1593?

neilcsmith-net:
@junichi11 as announced, master is open. The PR I would have liked to prioritize was #1589 but that's been closed. NETBEANS-3257 is the one urgent fix we need.  #1593 is just a possible addition if we have a patch update.  Either will need careful cherry picking to release branch anyway as the spec version change has been made on master which means we can't sync any more - so other commits on master shouldn't cause issues if you want to start merging.

junichi11:
@neilcsmith-net I see. I'll merge them later. Thank you!

PR #1577, opened 2019-10-18
[NETBEANS-3249] Chrome connector connection with NetBeans

geertjanw:
When NetBeans gets the handshake from Chrome, it attempts to find the session it stored, but because the name has had '/private' inserted, the lookup fails and NetBeans refuses to debug. This fixes that.


matthiasblaesing:
I tried to reproduce this, but somewhere between Chromium, the chrome connector missing from the apache netbeans build and some strange chromium behavior I got lost. Can you give steps how to reproduce the problem?

geertjanw:
Yes, this should be closed, https://github.com/apache/netbeans/pull/1803 is the right approach.

PR #1579, opened 2019-10-18
[NETBEANS-1985] Honour initial values of stateful delegate actions.

sdedic:
I have found this item in stash queue :(
The fix actually attempts NOT to overwrite the delegate's initialized values. Usual stateless delegates were fine, but stateful delegates initialize their selected state during construction. Without this fix, each new *context* action would assume initial selected state from the last global state, sometimes incorrect. The user-perceived behaviour seems erratic, as the visual state is "selected", while the model is "unselected", so the next click will select the already selected action.

Next, I've noticed that I've committed some API (!) classes into *incorrect folder* and that bad state lasted for several releases. The package declaration is correct (but does not match the folder), so they compile fine.


sdedic:
I have added some test cases, the `testContextDelegate` actually manifests the bug fixed - without the change in https://github.com/apache/netbeans/pull/1579/files#diff-48f7abf390587864f69a958530934b70 the test will fail.

During testing, other bug(s) surfaced - all `StatefulAction`s for all types and all properties were `equal`, which could make them NOT receive enable change notification from `ContextManager`.

Please review.

PR #1580, opened 2019-10-18
[NETBEANS-3252] NodeListModel must clear VisualizerNode count cache b…

sdedic:
…efore dispatching event.

Simply reorder the cache clear and event dispatch.


svatoun:
eh, I still need to provide the test...

PR #1584, opened 2019-10-22
[NETBEANS-3277] - cleanup the external binary cache

BradWalker:
Historically,  NetBeans has cached external binaries in a directory called ~/.hgexternalcache.

This change changes the name to nb_external_cache.

It also moves it out of the users home directory and into the .netbeans directory. This puts it into a place where it should be since things like this need to live in the .netbeans directory.

### DIFF
--- a/nbbuild/default-properties.xml
+++ b/nbbuild/default-properties.xml
@@ -66,7 +66,7 @@
 
   <property name="test.dist.dir" location="${nb.build.dir}/testdist"/>
   
-  <property name="binaries.cache" location="${user.home}/.hgexternalcache"/>
+  <property name="binaries.cache" location="${user.home}/.netbeans/nb_external_cache"/>
   <property name="binaries.server" value="https://netbeans.osuosl.org/binaries/"/>
 
   <target name="-load-build-properties">
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Caching dependencies
         uses: actions/cache@v2
         with:
-          path: ~/.hgexternalcache
+          path: ~/.netbeans/nb_external_cache
           key: ${{ runner.os }}-${{ hashFiles('**/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
 
@@ -86,7 +86,7 @@ jobs:
       - name: Caching dependencies
         uses: actions/cache@v2
         with:
-          path: ~/.hgexternalcache
+          path: ~/.netbeans/nb_external_cache
           key: ${{ runner.os }}-${{ hashFiles('**/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
 
@@ -108,7 +108,7 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: |
-            ~/.hgexternalcache
+            ~/.netbeans/nb_external_cache
             ~/Library/Caches/Homebrew
           key: ${{ runner.os }}-${{ hashFiles('**/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
 
 cache:
   directories:
-  - $HOME/.hgexternalcache
+  - $HOME/.netbeans/nb_external_cache
 
 matrix:
     include:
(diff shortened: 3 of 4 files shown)


lkishalmi:
I'm sorry Brad, I played with this on Github UI. Could you force update your branch from master so this PR would be the same one liner as before. I think this one is worth to be merged.

BradWalker:
> I'm sorry Brad, I played with this on Github UI. Could you force update your branch from master so this PR would be the same one liner as before. I think this one is worth to be merged.

Hey Laszlo, thanks very much for working on this! I've rebased to the latest master..

nbbuild/default-properties.xml
```diff
@@ -66,8 +66,8 @@
 
   <property name="test.dist.dir" location="${nb.build.dir}/testdist"/>
   
-  <property name="binaries.cache" location="${user.home}/.hgexternalcache"/>
-  <property name="binaries.server" value="https://netbeans.osuosl.org/binaries/"/>
+  <property name="binaries.cache" location="${user.home}/.netbeans/nb_external_cache"/>
+  <property name="binaries.server" value="http://netbeans.osuosl.org/binaries/"/>
```
matthiasblaesing, line 70: Is the switch from `https` to `http` intentional? If so, why?

BradWalker: It is not intentional.. It was very late last night after getting the kids to bed, I could have done that by accident.

It is now fixed..

matthiasblaesing:
@hectorespert could you please have a look at this? This will impact the unittests as caching depends on the location of the binary cache.
@BradWalker it would be great if you could have a look at the unittests as well:

https://github.com/apache/netbeans/blob/master/.travis.yml#L35
https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml#L35
https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml#L89

and incoming:

https://github.com/apache/netbeans/blob/3dccfed705d19d1ed1ca0eb2550da911d9cc210d/.github/workflows/main.yml#L111

hectorespert:
@BradWalker You should update Travis config and GItHub Actions config to allow these use the new cache folder.

BradWalker:
> @BradWalker You should update Travis config and GItHub Actions config to allow these use the new cache folder.

Hey @hectorespert , I've updated the places that I missed.. Did a build and ran some tests.. Should be good to go now.. Thanks for the feedback..

BradWalker:
Thanks @hectorespert for the review & comments.

timboudreau:
OS distros usually define a place for cache files, which the IDE respects (e.g. ${user.home}/.cache

It would be preferable to figure that out (env var?) and use that.

AlexFalappa:
As already pointed out by @timboudreau it looks that _NetBeans cache dir_ rather than _NetBeans user dir_ is better suited to hold cached data.

For example NetBeans cache dir on Linux is $HOME/.cache/netbeans/<version>

lkishalmi:
What about this PR?

matthiasblaesing:
I had a look at the code, the netbeans launcher uses. There are two code paths:

Linux/mac OS (from `nb/ide.launcher/unix/netbeans`):

```bash
case "`uname`" in
    Darwin*)
        # set default userdir and cachedir on Mac OS X
        DEFAULT_USERDIR_ROOT="${HOME}/Library/Application Support/NetBeans"
        DEFAULT_CACHEDIR_ROOT=${HOME}/Library/Caches/NetBeans
        ;;
    *) 
        # set default userdir and cachedir on unix systems
        DEFAULT_USERDIR_ROOT=${HOME}/.netbeans
        DEFAULT_CACHEDIR_ROOT=${HOME}/.cache/netbeans
        ;;
esac
```

Windows (from `nb/ide.launcher/windows/nblauncher.cpp`):

```c++
string NbLauncher::getDefaultCacheDirRoot() {
    TCHAR defCacheDirRootChar[MAX_PATH];
    if (FAILED(SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, 0, defCacheDirRootChar))) {
        return std::string();
    }
    defCacheDirRoot = constructApplicationDir((string) defCacheDirRootChar, true);
    defCacheDirRoot.erase(defCacheDirRoot.rfind('\\'));
    logMsg("Default Cachedir Root: %s", defCacheDirRoot.c_str());
    return defCacheDirRoot;
}
```

The shell variant is trivial to replicate, but the windows code is harder - if we are willing to pull `jna-platform` into the build, the function is already build. Alternative 1: We copy the function definition from jna-platform (JNA has no notice, so only the copyright will remain with JNA), Alternatve 2: Someone say how to access the special folders from the CLI.

BradWalker:
> we can probably close this?

Yes.

I will readdress this in the future. But, at this point it's the proper thing to do.

PR #1586, opened 2019-10-23
[TRAVIS] Move check-line-endings script and RAT report

blackleg:
Move the check-line-endings script and the verification of the RAT report to run in a separate build job.


jlahoda:
Looks good to me.

PR #1588, opened 2019-10-24
A simple UI to register new languages using TextMate grammars and Language Servers

jlahoda:


Chris2011:
Uhhh, that sounds nice :). I'm exited.

Chris2011:
Will test it until the weekend, I know it is work in progress, just testing and will tell you what I think :)

Chris2011:
Hey @jlahoda is the option hidden atm? I pulled your PR, built it, but I can't see your options.

jlahoda:
There must be a project opened (so that ergonomics enable the editor). Then I see Tools/Options/Editor/Language Servers.

Chris2011:
Thx, will check it. Didn't check the code deep yet, just saw, that you have an option created :)

jlahoda:
Good enough to be considered for inclusion now, I think.

Chris2011:
I have a list with some couple of things to let you know. Unfortunately I'm sick and can't handle it today. I will do it until the weekend. I think those are not complicated but I need to verify them again. So if you can wait, that would be very nice :)

Chris2011:
Hey Jan,

here is my list:

- Adding XML was not possible, please have a look into this tmLanguage file: https://github.com/Chris2011/TextmateBeans/blob/master/src/main/resources/org/chrizzly/textmatebeans/cucumber/feature.tmLanguage.xml
- Don't make the name optional. When I add a new language and I let the name blank I can't see anything in the list:

image

As you can see, I added 2 languages, but no one knows which one.

- Add the icon, the name, the mimetype into the table
- Show JSON and XML files inside of the JFileChooser

Couple of questions:
- Is it correct, that I don't need to create a mimetype anymore? Would be nice to see how the mimetype looks like. I mean it would be good to change the mimetype too. e.g. I want that my mimetype is text/x-cs for C# files.

After this changes, everything is fine and works well :)

jlahoda:
What is the point in changing the mime type? What if the user chooses a mime type that clashes with something. I guess some strong usecase is missing here?

Chris2011:
True. Ignore this point.

One more thing. What are the priorities?

I mean what happens, when I add a textmate grammar to a language that already exists? What will win? For the usecase, maybe my grammar is more up to date than the one, coded inside the IDE? For example no one updates the kotlin grammar for years and after 1 or 2 years I decide to add a new one via textmate.

jlahoda:
I'd like to merge tomorrow, unless there are serious objections.

Chris2011:
Thx for the fixes for empty name and adding the icon. Finally we have a very nice functionality to add textmate grammars easily. Very great work @jlahoda.

Just want to ping my question up:

> What are the priorities?
> 
> I mean what happens, when I add a textmate grammar to a language that already exists? What will win? For the usecase, maybe my grammar is more up to date than the one, coded inside the IDE? For example no one updates the kotlin grammar for years and after 1 or 2 years I decide to add a new one via textmate.
> 
> Are there any conflicts or will the textmate then just ignored?

Just a question, no objection.

jlahoda:
@Chris2011 - right now, I don't think there are any priorities specified, which will probably mean the elements registered in the GUI will either get a random handling, or be ignored (more likely the latter). It may be possible to do improvements there, but as part of that we need to match the mime types to the existing ones, and specify ordering/priorities. So a future task, I guess.

Chris2011:
Ok, thx for the info :)

PR #1590, opened 2019-10-24
[NETBEANS-1091] changed colors for links, rss text, titles, bottom li…

kriffer:
This is the simplest solution for https://issues.apache.org/jira/browse/NETBEANS-1091 without unnesessary code change.
Making the colors lighter make them  look better with the dark themes.
All colors are present in one style on the Start Page tabs.
Tested with WIndows 7, 10 (looks nice with hight contrast dark as well)


junichi11:
Thank you for your contribution.

As Matthias wrote, I'm not sure whether these changes are needed except for `nb/welcome/src/org/netbeans/modules/welcome/resources/Bundle.properties`.
At least `php/*` and `webcommon/*` files must not be removed. They should be needed as data of sample projects.

It would be nice if you attach screenshots of normal and dark themes (before and after).

kriffer:
Thanks for the review.
Correct, there is only  b440496  (for nb/welcome/src/org/netbeans/modules/welcome/resources/Bundle.properties) is valid.
Any other changes made after merging appeared here by mistake, so please don't count them. 
Would be great if you gave me a hint how to make a correction in this PR to get rid of unnecessary changes (rebase?)...
Here are screenshots:
dark before changes
dark_before

dark after change
dark_after

light after change
light_after

kriffer:
This is not DarkLaF itself. It is about using NB with Windows dark theme- the issue is reproducible on any of dark theme on Win7 (changing via 'personalize settings' )

PR #1591, opened 2019-10-24
Rework docker connection validation, so we can find out why docker connection failed.

Ihsahn:


Ihsahn:
Build took to long so Travis decided to kill it, unfortunately I have no permission to restart it.

neilcsmith-net:
@Ihsahn retriggered the failing test.

Ihsahn:
> @Ihsahn retriggered the failing test.

thx!

Ihsahn:
@matthiasblaesing can you take a look?

matthiasblaesing:
If I remember correctly, log messages on level SEVERE raise an "exception" bubble in the IDE. While this might be helpful, I as a user would prefer to have the message in the place where I can do something about it. I think it would be good to see if the either the value `WizardDescriptor.PROP_ERROR_MESSAGE` can be enhanced or if another output channel is needed.

Ihsahn:
@matthiasblaesing reworked. You were right about exception bubble and I was also mislead by logged exceptions (I thought that they come from my code, but apparently that was different logger...) - which I also fixed now.

ide/docker.api/src/org/netbeans/modules/docker/api/DockerAction.java
```diff
@@ -1004,9 +1004,8 @@ public StreamItem fetch() {
         }
     }
 
-    public boolean ping() {
+    public boolean pingWithExceptions() throws IOException, MalformedURLException {
         assert !SwingUtilities.isEventDispatchThread() : "Remote access invoked from EDT";
-        try {
             Endpoint s = createEndpoint();
```
matthiasblaesing: The idention is off - please move it to the correct level (one level to the left).

ide/docker.api/src/org/netbeans/modules/docker/api/DockerAction.java
```diff
@@ -1004,9 +1004,8 @@ public StreamItem fetch() {
         }
     }
 
-    public boolean ping() {
+    public boolean pingWithExceptions() throws IOException, MalformedURLException {
```
matthiasblaesing: I would not use such a precise `throws` clause. I would use `Exception` that way the code can evolve without being tied to the fixed API.

ide/docker.ui/src/org/netbeans/modules/docker/ui/wizard/DockerConnectionPanel.java
```diff
@@ -246,14 +258,24 @@ public void validate() throws WizardValidationException {
             }
 
             DockerAction action = new DockerAction(instance);
-            if (!action.ping()) {
+            boolean pingResult = false;
+            
+            try {
+                pingResult = action.pingWithExceptions();
+            } catch (Exception ex) {
+                LOGGER.log(Level.FINE, "docker connection ping failed", ex); // NOI18N
```
matthiasblaesing: I would move this to `Level.INFO`. It is worth logging and INFO will not reach the UI, but in the log it is still useful.

Ihsahn:
@matthiasblaesing Comments addressed

matthiasblaesing:
Merged via https://github.com/apache/netbeans/commit/d5a1a036f6771916a326ec84025d0ad64a51b4a7 (sorry for some reason the author information was lost on rebase - I had to fix a wrong exception catch)

PR #1592, opened 2019-10-24
[NETBEANS-3002] NullPointerExceptions during PHPStan inspection

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-3002

- The file name value can be null

e.g.
```xml
<file>
  <error severity="error" message="Ignored error pattern #TEST# was not matched in reported errors." />
</file>
```
So, in this case, just ignore the error.

- There is a difference in file name value between the old version and the new version

e.g. Inspect the following app directory (myproject is the working directory)
- myproject/app/Test1.php
- myproject/app/Test2.php

old version: the relative path from target directory
```xml
<file name="Test1.php">
  <error line="13" column="1" severity="error" message="Binary operation &quot;+&quot; between string and 2 results in an error." />
</file>
<file name="Test2.php">
  <error line="9" column="1" severity="error" message="Anonymous function should return string but returns void." />
  <error line="9" column="1" severity="error" message="Result of closure (void) is used." />
</file>
```

new verison: the relative path from working directory
```xml
<file name="app/Test1.php">
  <error line="13" column="1" severity="error" message="Binary operation &quot;+&quot; between string and 2 results in an error." />
</file>
<file name="app/Test2.php">
  <error line="9" column="1" severity="error" message="Anonymous function should return string but returns void." />
  <error line="9" column="1" severity="error" message="Result of closure (void) is used." />
</file>
```

So, check also the relative path from the working directory.

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParser.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParser.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.netbeans.api.annotations.common.CheckForNull;
+import org.netbeans.api.annotations.common.NullAllowed;
 import org.netbeans.modules.php.analysis.results.Result;
 import org.netbeans.modules.php.api.util.FileUtils;
 import org.openide.filesystems.FileObject;
@@ -56,15 +57,16 @@ public class PHPStanReportParser extends DefaultHandler {
     private Result currentResult = null;
     private String currentFile = null;
     private final FileObject root;
-    private final File workDir;
+    @NullAllowed
+    private final FileObject workDir;
 
-    private PHPStanReportParser(FileObject root, File workDir) throws SAXException {
+    private PHPStanReportParser(FileObject root, @NullAllowed FileObject workDir) throws SAXException {
         this.xmlReader = FileUtils.createXmlReader();
         this.root = root;
         this.workDir = workDir;
     }
 
-    private static PHPStanReportParser create(Reader reader, FileObject root, File workDir) throws SAXException, IOException {
+    private static PHPStanReportParser create(Reader reader, FileObject root, @NullAllowed FileObject workDir) throws SAXException, IOException {
         PHPStanReportParser parser = new PHPStanReportParser(root, workDir);
         parser.xmlReader.setContentHandler(parser);
         parser.xmlReader.parse(new InputSource(reader));
@@ -72,7 +74,7 @@ public class PHPStanReportParser extends DefaultHandler {
     }
 
     @CheckForNull
-    public static List<Result> parse(File file, FileObject root, File workDir) {
+    public static List<Result> parse(File file, FileObject root, @NullAllowed FileObject workDir) {
         try {
             sanitizeFile(file);
             try (Reader reader = new BufferedReader(new InputStreamReader(new FileInputStream(file), "UTF-8"))) { // NOI18N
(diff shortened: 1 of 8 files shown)


junichi11:
@tmysik Could you please review this although we can't merge this into the master branch yet?

php/php.code.analysis/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParser.java
```diff
@@ -184,6 +192,15 @@ private String getCurrentFile(String fileName) {
         }
         if (rootDirectory.isFolder()) {
             FileObject current = rootDirectory.getFileObject(sanitizedFileName);
+            if (current == null) {
+                // NETBANS-3022 try checking relative file path from working directory
+                if (workDir != null) {
+                    rootDirectory = FileUtil.toFileObject(workDir);
```
tmysik: Nitpick: Although cached, this is quite a time-consuming operation. Consider adding a class field

```java
@Nullable
private final FileObject workDir;
```

But this is really just a note, not a problem at all, of course.

php/php.code.analysis/test/unit/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParserTest.java
```diff
@@ -36,7 +36,8 @@ public PHPStanReportParserTest(String name) {
 
     public void testParse() throws Exception {
         FileObject root = getRoot("phpstan/PHPStanSupport");
-        List<Result> results = PHPStanReportParser.parse(getLogFile("phpstan-log.xml"), root);
+        File workDir = FileUtil.toFile(root);
```
tmysik: Note: `FileObject` to `File` operation is fast and without any extra costs.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParser.java
junichi11: Does it mean that we should change the following to it?
https://github.com/apache/netbeans/blob/1ee4fdb02036bd3d626a1484181a9b9a7bc785e1/php/php.code.analysis/src/org/netbeans/modules/php/analysis/parsers/PHPStanReportParser.java#L59

e.g.
```java
    @Nullable
    private final FileObject workDir;
    private PHPStanReportParser(FileObject root, File workDir) throws SAXException {
        this.xmlReader = FileUtils.createXmlReader();
        this.root = root;
        this.workDir = FileUtil.toFileObject(workDir);
    }
```

junichi11: Or should return `FileObject` instead of `File` in findWorkDir() method?
https://github.com/apache/netbeans/blob/1ee4fdb02036bd3d626a1484181a9b9a7bc785e1/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/PHPStan.java#L138-L149

tmysik: It is up to you to decide. I remember that personally, I was trying to use `FileObject` as much as possible because that is how NetBeans works with files. However, `FileObject` has many disadvantages too, e.g. if it is `null` you have no idea what the original file path was (typically problem in an API, here, `File` is typically a better choice).

Back to this situation - I would perhaps try to use `@Nullable FileObject ` for `findWorkDir` method (and log `ERROR` if `projectDir` is not found because that should never happen, right?).

Thanks!

junichi11: I'll use `FileObject` and leave the log if the project directory is not found. Thank you!

junichi11:
Fixed. Use `FileObject` instead of `File`.
I realized the file name is absolute path if the working directory is null (i.e. the target file or directory is not in the PHP project). So, I've also fixed it. Thanks.

tmysik:
@junichi11 

> Fixed. Use `FileObject` instead of `File`.
> I realized the file name is absolute path if the working directory is null (i.e. the target file or directory is not in the PHP project). So, I've also fixed it. Thanks.

Perfect!

PR #1593, opened 2019-10-25
NETBEANS-3290 fix common xml

ebarboni:
assumption of being in root were false. fixing delta in path

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -328,7 +328,7 @@
         <pathconvert property="dtd.cp">
             <pathfileset>
                 <path refid="cluster.path.id"/>
-                <filename name="platform/core/org-openide-filesystems.jar"/>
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
@ebarboni added Cherry Pick label. If we shipped an NB 11.2 update to `org-netbeans-modules-apisupport-harness.nbm` I'm assuming this would get picked up OK?

ebarboni:
ok,this is to flag for in case of update

oyarzun:
I've tested this as well and it fixes the issue.

neilcsmith-net:
@ebarboni can you merge this, then cherry pick locally on top of release112 (probably with spec version increment to apisupport.harness) and see if installing that updated NBM in 11.2 release fixes the issue?  If so, can you then open a PR against release112 and I'll look at shipping this with other updates.  Thanks!

PR #1595, opened 2019-10-25
[TRAVIS] Fix Travis config when run tests for ide modules

blackleg:
Current Travis configuration doesn't run the tests for the ide modules only does a build of the source code.

It's the same problem that @matthiasblaesing found when he try to add tests in webcommon cluster: https://github.com/apache/netbeans/pull/1451


matthiasblaesing:
Looks sane to me. Thank you.

PR #1599, opened 2019-10-27
[NETBEANS-1657] PHP visibility keywords via drop-down list

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1657
- Add ConvertVisibilitySuggestion
- Note: Just convert the visibility keyword, i.e. don't check whether there is a problem (e.g. "public" -> "private")
- In the case of interface, it is available if the visibility is implicit public ("" -> "public")

e.g.
Before:
```php
interface MyInterface {
    const IMPLICIT_CONST = "implicit";
    function implicitPublicMethod($param);
}
```
After:
```php
interface MyInterface {
    public const IMPLICIT_CONST = "implicit";
    public function implicitPublicMethod($param);
}
```

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/api/PhpModifiers.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/api/PhpModifiers.java
@@ -32,6 +32,12 @@ public final class PhpModifiers extends Modifier {
 
     private static final PhpModifiers EMPTY = new PhpModifiers(NO_FLAGS);
 
+    // visibility
+    public static final String VISIBILITY_VAR = "var"; // NOI18N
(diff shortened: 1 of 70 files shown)


tmysik:
@junichi11 Is this ready for review? It is WIP...

What is the use-case for `public mymethod()` in an `interface`?

Thanks.

junichi11:
@tmysik 

> Is this ready for review? It is WIP...

Yes, I did it to disable the "Merge pull request" button, but I'll click the "Ready for review" button later(after I fixed the below problem) because you know that we can't merge this yet :)

> What is the use-case for public mymethod() in an interface?

Do nothing because an interface can't have `private` and `protected` methods.

I realized that abstract class cases are missing. So, I'll fix it later (maybe, tomorrow).

Thanks!

PR #1603, opened 2019-10-30
[NETBEANS-3276] Let GradleJavaEEProjectSettings be always registered.

lkishalmi:
Previously the GradleJavaEEProjectSettings has only been registered in the project lookup if that had the war plugin loaded. Loosing that plugin from the project could result that some listeners on the JavaEE project preferences would not be removed.
Adding the settings to all Gradle project probably won't really hurt as it is only used by the JavaEE functionality when installed.


JaroslavTulach:
> Loosing that plugin from the project could result that some listeners on the JavaEE project preferences would not be removed.

Wouldn't using `WeakListeners` be a more suitable choice?

lkishalmi:
@JaroslavTulach Thank you for the insight! Though I think weak listeners are great, probably not here. I made more debugging on this. It seems the JavaEE support holds a reference to the WebBrowserProvider even after it got removed from the project just in order to gracefully remove that listener. The problem is my sloppy handling of the preference field of that object.

PR #1589, opened 2019-10-24
[NETBEANS-3257]: Fixed issue on refactoring rename of java method

Akshay-Gupta-Oracle:
[NETBEANS-3257]: Fixed issue on refactoring rename of java method

PR #1604, opened 2019-10-31
[NETBEANS-3257]: Fixed issue on refactoring rename of java method

Akshay-Gupta-Oracle:
Changes fix for refactoring rename of java methods.
https://issues.apache.org/jira/browse/NETBEANS-3257

nb-javac jars are oracle signed now.

https://github.com/apache/netbeans/pull/1589
this PR got accidentally closed while renaming branch and there is no option to reopen it.

### DIFF
--- a/java/libs.javacapi/external/binaries-list
+++ b/java/libs.javacapi/external/binaries-list
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-8250FAEF60A423DA3D5EFA0EE699FB69D906D86C nb-javac-13-api.jar
+FC7D04B1F1AE92A87F113096862112BE7E6970D4 nb-javac-13-api.jar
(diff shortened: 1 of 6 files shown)


neilcsmith-net:
@jlahoda can you review this so we can merge to master, and then either you or @Akshay-Gupta-Oracle open a cherry-picked PR against release112?  Be good to know if applying this update via NBM in 11.2 release both installs the updated nb-javac and fixes the issue before we go ahead with release build and patch voting.  Thanks!

jlahoda:
I tested, seems to fix the issue. Normally, if one install the NBM with the fix, then one needs another round of update to get the updated nb-javac. But we could manually put the nb-javac modules on the distro AUC.

One thing that is missing is an update to the spec version of the updatecenters module, but we can do that on the cherry pick, master has updated spec version anyway.

jlahoda:
FYI, cherry-pick:
https://github.com/apache/netbeans/pull/1659

PR #1605, opened 2019-11-01
[NETBEANS-1855] Autocomplete for property and method without $this->

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-1855

e.g.

```php
class MyClass {
    public test() {
        code// CC here : $this->codeCompletion() is added
    }

    private codeCompletion() {}

}

```
netbeans-1855-prototype

### DIFF
--- a/php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion
+++ b/php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion
@@ -4,7 +4,11 @@ echo "class property: |$this->name\n";
 PACKAGE    ANS                             [PUBLIC]   null
 PACKAGE    CNS                             [PUBLIC]   null
 CLASS      Book                            [PUBLIC]   issue153707.php
+METHOD     getName()                       [PUBLIC]   Book
+METHOD     setName($name)                  [PUBLIC]   Book
 VARIABLE   ? $name                         [PUBLIC]   issue153707.php
+VARIABLE   ? name                          [PUBLIC]   Book
+CONSTANT   TEST 'Simon'                    [PUBLIC]   Book
 KEYWORD    Book $this->                               null
 KEYWORD    parent::                                   null
 KEYWORD    self::                                     null
--- a/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletionTest.java
+++ b/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletionTest.java
@@ -781,6 +781,7 @@ public class PHPCodeCompletionTest extends PHPCodeCompletionTestBase {
     }
 
     public void testIssue153707_01() throws Exception {
+        // FIXME: maybe, should add only variables and $this->
         checkCompletion("testfiles/completion/lib/_base/issue153707.php", "class property: ^", false);
     }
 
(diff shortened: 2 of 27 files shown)


php/php.editor/test/unit/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletionTest.java
```diff
@@ -781,6 +781,7 @@ public void testClassConstructorOptionalParam_01() throws Exception {
     }
 
     public void testIssue153707_01() throws Exception {
+        // FIXME: maybe, should add only variables and $this->
         checkCompletion("testfiles/completion/lib/_base/issue153707.php", "class property: ^", false);
```
junichi11: There are unnecessary CC items. I don't think this is a big problem. I just marked it as FIXME.
#### Code
https://github.com/apache/netbeans/blob/f2611e358c181935500ea4d9d9142fb850504a72/php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php#L15-L19

#### Result
https://github.com/apache/netbeans/blob/f2611e358c181935500ea4d9d9142fb850504a72/php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion#L1-L218

tmysik: Agreed.

php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion
```diff
@@ -4,7 +4,11 @@ echo "class property: |$this->name\n";
 PACKAGE    ANS                             [PUBLIC]   null
 PACKAGE    CNS                             [PUBLIC]   null
 CLASS      Book                            [PUBLIC]   issue153707.php
+METHOD     getName()                       [PUBLIC]   Book
+METHOD     setName($name)                  [PUBLIC]   Book
 VARIABLE   ? $name                         [PUBLIC]   issue153707.php
+VARIABLE   ? name                          [PUBLIC]   Book
```
tmysik: Is this expected?

tmysik:
@junichi11 

Just one minor question otherwise looks good. Let me know, please, I will merge it (can I, right?).

Thanks.

php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion
junichi11: Yes, this `name` variable means `$this->name`. `$name` is the parameter. 
https://github.com/apache/netbeans/blob/fb38c4ec8b578efd70fc84d81fcff8a16b495deb/php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php#L3-L19

(But we should improve this case because there are unnecessary CC items.)

junichi11:
@tmysik 
> I will merge it (can I, right?)

Sure :) Thank you!

php/php.editor/test/unit/data/testfiles/completion/lib/_base/issue153707.php.testIssue153707_01.completion
tmysik: OK, thanks.

PR #1612, opened 2019-11-04
[NETBEANS-2439] Java Frontend Application cannot be activated on JDK11

dukescript:
add upper limit for recommended modules

changing configuration panel ui to show recommended and required installs, because otherwise nbjavac will be installed automatically.


JaroslavTulach:
Clearly the NETBEANS-2439 problem is real and this is a fix. It may not be the nicest one, but it is a fix. Let's integrate it.

PR #1613, opened 2019-11-05
[NETBEANS-3341] - Upgrade asm-all.jar 5.0.1 to 5.1

pepness:
1.-Upgrade asm-all-5.0.1 version to asm-all-5.1including it's license 

2.-Add constructor parameter to use ASMv5 API:
`ClassNode clazz = new ClassNode(Opcodes.ASM5);`

3.-Add documentation to one class that extends a class from asm.jar


pepness:
I will make another PR becuase the ticket https://issues.apache.org/jira/browse/NETBEANS-3346 will be implemented instead(Upgrade asm-all jar to asm 7.2).

PR #1618, opened 2019-11-05
    [NETBEANS-3345] - cleanup fallthrough warnings..

BradWalker:
[NETBEANS-3345] - cleanup fallthrough warnings..
    
    There are numerous places in the code where the following warning is issued.
    
```
       [repeat] /home/bwalker/src/netbeans/ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java:718: warning: [fallthrough] possible fall-through into case
       [repeat]                 case WS: //@import |
       [repeat]                 ^
       [repeat] /home/bwalker/src/netbeans/ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java:824: warning: [fallthrough] possible fall-through into case
       [repeat]                         default:
       [repeat]                         ^
       [repeat] /home/bwalker/src/netbeans/ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java:1015: warning: [fallthrough] possible fall-through into case
       [repeat]             case mediaBody:
       [repeat]             ^
    
```
I've looked over each and every place where this warning is emitted. The vast majority of these
warnings are due to lexer spaghetti. Which happens often when dealing w/ look-ahead parsers..
    
The simple fix is to add a "@SuppressWarnings("fallthrough")" to the method. as the code looks
like the designer intended this behavior.
    
There is one place where I did find a bug..
    
_platform/core.network/src/org/netbeans/core/network/proxy/NetworkProxyReloader.java_
    
It's missing a break statement in this file.

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/java/hints/errors/SearchModuleDependency.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/java/hints/errors/SearchModuleDependency.java
@@ -113,6 +113,7 @@ public class SearchModuleDependency implements org.netbeans.modules.java.hints.s
     }
 
     @Override
+    @SuppressWarnings("fallthrough")
     public List<Fix> run(final CompilationInfo info, String diagnosticKey,
             final int offset, TreePath treePath, Data<Void> data) {
         cancel.set(false);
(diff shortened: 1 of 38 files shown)


matthiasblaesing:
I revisited this and reread the change in `NetworkProxyReloader.java` I don't think it is right. The message in default reads:

> System network proxy: fell to default (correct if direct mode went before)

So the fallthrough here is intentional. This `break` must not be inserted.

The whole changeset needs to be rebased onto master to be mergeable.

BradWalker:
I've fixed this as requested.. I was always concerned about the break in `NetworkProxyReloader.java`.. I just wasn't sure where to look..
Thanks!

matthiasblaesing:
Looks sane. Thank you.

PR #1619, opened 2019-11-07
Configure proxy before trying to download release.json from its URL

JaroslavTulach:
OracleLabs engineers have problems building NetBeans due to the new download from `metabuild.jsonurl` - it helps if we set the proxy up first. Here is my change for your consideration.

Let's see whether the Travis builds are OK and then we could integrate.

### DIFF
--- a/nbbuild/build.xml
+++ b/nbbuild/build.xml
@@ -67,6 +67,7 @@
     <taskdef name="deregisterexternalhook" classname="org.netbeans.nbbuild.extlibs.DeregisterExternalHook" classpath="${build.ant.classes.dir}"/>
     <deregisterexternalhook root=".."/>
     <taskdef name="downloadbinaries" classname="org.netbeans.nbbuild.extlibs.DownloadBinaries" classpath="${build.ant.classes.dir}"/>
+    <taskdef name="configureproxy" classname="org.netbeans.nbbuild.extlibs.ConfigureProxy" classpath="${nbantext.jar}"/>
     <property name="have-downloadbinaries-task" value="true" />
     <downloadbinaries cache="${binaries.cache}" server="${binaries.server}">
         <manifest dir="${nb_all}">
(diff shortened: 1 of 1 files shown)


nbbuild/build.xml
```diff
@@ -67,6 +67,7 @@
     <taskdef name="deregisterexternalhook" classname="org.netbeans.nbbuild.extlibs.DeregisterExternalHook" classpath="${build.ant.classes.dir}"/>
     <deregisterexternalhook root=".."/>
     <taskdef name="downloadbinaries" classname="org.netbeans.nbbuild.extlibs.DownloadBinaries" classpath="${build.ant.classes.dir}"/>
+    <taskdef name="configureproxy" classname="org.netbeans.nbbuild.extlibs.ConfigureProxy" classpath="${nbantext.jar}"/>
```
matthiasblaesing: Update the classpath to `${build.ant.classes.dir}` and it should work.

sdedic:
Seems good; `build.xml` contains yet another `<get>` for sigtests, shouldn't be the proxy configured there as well ?

JaroslavTulach:
Anyone to approve or merge?

neilcsmith-net:
I added @ebarboni to the review list as he wrote this. But if it's working everywhere good from my perspective. Longer term refactoring the json stuff into its own target might make sense, and still not sure we want skip existing set to false either.

jlahoda:
I wonder if all this (the get of the json and also the proxy config) shouldn't be made optional. Possibly on by default, but allowing to easily disable it e.g. by setting a property. I am sometimes without an internet connection, and editing build.xml just to get the build working (when all the real dependencies are available in the .hgexternalcache) seems unnecessary. Especially if what one does is not a release build, i.e. the json config won't really help much anyway.

neilcsmith-net:
@jlahoda it should already be possible to set `metabuild.jsonurl` to a local file. But, yes, agreed.

PR #12, opened 2017-09-26
properties for by module report

ebarboni:
Hi this is a little modification to allow rat target to report only one folder at a time. I was puzzled by the size of full rat report.

Usage example:
ant rat -Dmodule=uihandler 

default ant rat is working like before

Regards

PR #1620, opened 2019-11-08
[NETBEANS-3346] - Upgrade asm-all-5.0.1 version to asm-7.2

pepness:
-Upgrade asm-all-5.0.1 version to asm-7.2

-Add constructor parameter to use ASMv7 API

`ClassNode theClass = new ClassNode(Opcodes.ASM7);`
If you don't set the ASM version API it will use the default version from the jar:

- asm-all-5.0.1.jar = Opcodes.ASM5
- asm-6.1.jar = Opcodes.ASM6
- asm-7.2.jar = Opcodes.ASM7

-Visit @ConstructorDelegate parameters as array. (by sdedic)

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -78,10 +78,10 @@
         <tstamp>
             <format property="module.build.started.time" pattern="yyyy-MM-dd'T'HH:mm:ss.SSSZ"/>
         </tstamp>
-        <condition property="asm.jar" value="${nbplatform.active.dir}/platform/core/asm-all-5.0.1.jar">
-            <available file="${nbplatform.active.dir}/platform/core/asm-all-5.0.1.jar"/>
+        <condition property="asm.jar" value="${nbplatform.active.dir}/platform/core/asm-7.2.jar">
+            <available file="${nbplatform.active.dir}/platform/core/asm-7.2.jar"/>
         </condition>
-        <property name="asm.jar" location="${platform/libs.asm.dir}/core/asm-all-5.0.1.jar"/>
+        <property name="asm.jar" location="${platform/libs.asm.dir}/core/asm-7.2.jar"/>
     </target>
 
     <target name="-release.dir">
--- a/platform/libs.asm/nbproject/project.xml
+++ b/platform/libs.asm/nbproject/project.xml
@@ -29,21 +29,13 @@
                 <subpackages>org</subpackages>
             </public-packages>
             <class-path-extension>
-                <runtime-relative-path>asm-all-5.0.1.jar</runtime-relative-path>
-                <binary-origin>external/asm-all-5.0.1.jar</binary-origin>
-            </class-path-extension>
-<!--            <class-path-extension>
-                <runtime-relative-path>ext/asm-5.0.1/asm-5.0.1.jar</runtime-relative-path>
-                <binary-origin>external/asm-5.0.1.jar</binary-origin>
+                <runtime-relative-path>asm-7.2.jar</runtime-relative-path>
+                <binary-origin>external/asm-7.2.jar</binary-origin>
             </class-path-extension>
             <class-path-extension>
-                <runtime-relative-path>ext/asm-5.0.1/asm-commons-5.0.1.jar</runtime-relative-path>
-                <binary-origin>external/asm-commons-5.0.1.jar</binary-origin>
+                <runtime-relative-path>asm-tree-7.2.jar</runtime-relative-path>
+                <binary-origin>external/asm-tree-7.2.jar</binary-origin>
             </class-path-extension>
-            <class-path-extension>
-                <runtime-relative-path>ext/asm-5.0.1/asm-util-5.0.1.jar</runtime-relative-path>
-                <binary-origin>external/asm-util-5.0.1.jar</binary-origin>
-            </class-path-extension>-->
         </data>
     </configuration>
 </project>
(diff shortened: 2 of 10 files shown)


lkishalmi:
/home/travis/build/apache/netbeans/nbbuild/build.xml:1669: Failed VerifyLibsAndLicenses test(s):
Some license files were badly formatted
platform/libs.asm/external/asm-7.2-license.txt has a trailing space on line #12
platform/libs.asm/external/asm-tree-7.2-license.txt has a trailing space on line #12

lkishalmi:
Also See: https://github.com/apache/netbeans/pull/1108#issuecomment-458698431 you do not have to have two licence file.

pepness:
Thanks, I'll check it out.

platform/libs.asm/nbproject/project.xml
```diff
@@ -29,8 +29,12 @@
...
-                <runtime-relative-path>asm-all-5.0.1.jar</runtime-relative-path>
-                <binary-origin>external/asm-all-5.0.1.jar</binary-origin>
+                <runtime-relative-path>asm-7.2.jar</runtime-relative-path>
+                <binary-origin>external/asm-7.2.jar</binary-origin>
+            </class-path-extension>
+            <class-path-extension>
+                <runtime-relative-path>asm-tree-7.2.jar</runtime-relative-path>
+                <binary-origin>external/asm-tree-7.2.jar</binary-origin>
             </class-path-extension>
 <!--            <class-path-extension>
```
sdedic: Could be this obsolete part removed ?

nbbuild/templates/common.xml
```diff
@@ -78,10 +78,10 @@
         <tstamp>
             <format property="module.build.started.time" pattern="yyyy-MM-dd'T'HH:mm:ss.SSSZ"/>
         </tstamp>
-        <condition property="asm.jar" value="${nbplatform.active.dir}/platform/core/asm-all-5.0.1.jar">
-            <available file="${nbplatform.active.dir}/platform/core/asm-all-5.0.1.jar"/>
+        <condition property="asm.jar" value="${nbplatform.active.dir}/platform/core/asm-7.2.jar">
```
sdedic: This `${asm}` is passed as asm location to `processjsannotations` task, but its property only accepts a single file (used to be `asm-all*.jar`) so the `-tree` part can be now missing from task's runtime.
@JaroslavTulach wrote the task

pepness: Both the asm-7.2. & asm-tree-7.2 jars appears in the Startup Classpath. Or is this a different topic?
[asm-72v7-dedic-patch.log](https://github.com/apache/netbeans/files/3846985/asm-72v7-dedic-patch.log)

platform/libs.asm/nbproject/project.xml
pepness: I will remove that part

nbbuild/templates/common.xml
sdedic: See [line 209](https://github.com/apache/netbeans/blob/5da8a26a2435256f969d7ae3f49a2964bfc6b361/nbbuild/templates/common.xml#L209): 
```
        <processjsannotation classes="${build.classes.dir}" asm="${asm.jar}">
            <classpath refid="cp"/>
        </processjsannotation>
```
The task first loads asm from `cp`, then from the `asm` property; I don't know if it is guaranteed that both ASM jars appear on `cp` (maybe not, because that's why the task has an alternative property). But that property only allows to add one jar, we have two of them now. @jtulach  should know better what that task requires.

JaroslavTulach: I don't remember clearly, but I think that inside of `nbbuild` one classpath is valid and inside of `harness` build another classpath is valid.

platform/libs.asm/nbproject/project.xml
pedro-w: Which bit is obsolete, do we not need both asm & asm-tree?

pepness: I think sdedic refers to the commented code in that file. The `asm-tree.jar` library have `org.objectweb.asm.tree` that it is used in `Asm.java` and `NbJShellAgent.java` files.

sdedic: @pepness correct

nbbuild/templates/common.xml
sdedic: I am not able to find an example to test with. The `FnUtils` class that is reflective-loaded by `processjsannotations` is at https://github.com/apache/netbeans-html4j/blob/master/boot/src/main/java/org/netbeans/html/boot/impl/FnUtils.java ; looking at the source, it should be satisfied with just the basic `asm`, seems it does not need types from `asm-tree`.

sdedic:
Oops... sorry for the review re-request. Mis-clicked in the page :(

nbbuild/templates/common.xml
pepness: Doing a "Find in Project" of `org.objectweb.asm.tree` just three files use `asm-tree`

- `ide/ide.kit/test/qa-functional/data/whitelist_1.txt`
- `java/lib.jshell.agent/agentsrc/org/netbeans/lib/jshell/agent/NbJShellAgent.java`
- `platform/core.startup/src/org/netbeans/core/startup/Asm.java`

pedro-w: Do you think the 'asm' parameter to the `processjsannotations` task is still needed? Could the relevant jars be included as an element in the 'classpath' parameter? In other words, do the asm jars need to be treated specially? As sdedic says, the logic now is that the contents of the asm parameter are not used if anything in the classpath parameter contains the letters "asm", and I don't know enough of the history to understand why that might be.

junichi11:
@sdedic You can also see incremental changes if you click`force-pushed` although you may already know that :)

PR #1625, opened 2019-11-13
[NETBEANS-3256] Upgrade Gradle Tooling API to 6.0

lkishalmi:
Well probably this won't be the final Gradle Tooling we deliver for 11.3, as this one is really fresh, but we need to update to 6.x early enough.

@matthiasblaesing I left the wrapper and the binary at 4.10.2 as that is good enough for now, however I've noticed that we have separate license and notice file for those. Theoretically we only bundle the tooling API. So we shall not include those into the license and NOTICE file, as they are just build time dependencies. Is there some ignore lists for them?


lkishalmi:
BTW accidentally fixed the DownloadBinaries Ant task, and with that I've updated the maven and OSUL URL-s to HTTPS

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
```diff
@@ -80,6 +80,19 @@ public void setServer(String server) {
         this.server = server;
     }
 
+    private String repos = "https://repo1.maven.org/maven2/";
```
matthiasblaesing: This feels a bit sneaked in (support for non-central repositories), but I see why it might be a good idea.

lkishalmi: Yes, Gradle do not publish to Maven Central and I realized just after I've implemented this feature that we did have an issue using non-maven central repositories. I've ended up uploading the Gradle Tooling API to OSUL, but I kept the feature in.

lkishalmi:
@matthiasblaesing, @JaroslavTulach, or @neilcsmith-net can I have a quick review on this one? This would be the base of some other changes, template updates etc. So I would like to merge it early.

lkishalmi:
I've started with that dependency. Unfortunately in that case the IDE does not recognize org.slf4j.api and org.slf4j.jdk14 as modules, the whole set of the modules become uninstallable by the module manager. I've searched all the project.xml how they are referencing that module. Only jgit has a build time reference for that. Strangely there is no other external classpath reference as well. I've tried without the external classpath reference, hoping that it would be picked up "somehow", but got NoClassDefFound exception as soon as I turn to the Gradle Tooling.
So this was the best thing I could come up with.

matthiasblaesing:
There is one big problem: if there is no dependency on the slf4j modules, they will not be pulled in, if gradle is installed independently of git. Maybe just bundling the library is the easiest solution at this point (I'm experimenting, but I think the difference is, that jgit pulls in slf4j via the OSGI runtime and gradle expects them on the classpath).

lkishalmi:
Well what is being done now with the runtime classpath extension is the same as before. It is the location of the jars different, taken from the ide cluster. It is really just a runtime dependency slf4j is never used in the module source, just from the Gradle Tooling API.
It is working actually. I saw some relative references mostly in the maven embedder module.

Vitexus:
There is problem during build:

`[downloadbinaries] Downloading: https://netbeans.osuosl.org/binaries/12C7F938F06B39C5689D5C48F1418AF00A4696E9-gradle-tooling-api-6.0.jar`

Ok but see on https://netbeans.osuosl.org/binaries/ there is olny older 70A74F970B22AC1226F6A2983E38A77FA90E3BE9-gradle-tooling-api-4.10.2.jar so i get:

`[downloadbinaries] Could not download 12C7F938F06B39C5689D5C48F1418AF00A4696E9-gradle-tooling-api-6.0.jar to /home/vitex/.hgexternalcache/12C7F938F06B39C5689D5C48F1418AF00A4696E9-gradle-tooling-api-6.0.jar: /home/vitex/Projects/VitexSoftware/Packaging/netbeans.deb/netbeans/nbbuild/build.xml:217: Could not download 12C7F938F06B39C5689D5C48F1418AF00A4696E9-gradle-tooling-api-6.0.jar from https://netbeans.osuosl.org/binaries/: null
`

`

matthiasblaesing:
I checked - the file listing does not show the files, but the file is there. Please try again.

PR #1626, opened 2019-11-13
Use .exe suffix when looking for jmod on Windows

pedro-w:
I was seeing "Warning: nbjdk.active=... or nbjdk.home=... is an invalid Java platform" warnings when building NB on Windows. The reason was that the build script was not finding the `jmod` executable in the JDK. This simple patch makes sure to look for it using the .exe suffix, using the same mechanism as it currently uses for `javac` et al.

### DIFF
--- a/nbbuild/jdk.xml
+++ b/nbbuild/jdk.xml
@@ -98,7 +98,7 @@
                 <available file="${nbjdk.javadoc}" type="file"/>
                 <or>
                     <available classname="java.lang.Object" classpath="${nbjdk.bootclasspath}" ignoresystemclasses="true"/>
-                    <available file="${nbjdk.home}/bin/jmod"/>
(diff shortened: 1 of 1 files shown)


pedro-w:
For info, the clause at lines 100-101 fails
https://github.com/apache/netbeans/blob/3f0095729e90bb458642f9aa09c954bdcf2fb877/nbbuild/jdk.xml#L93-L104
and the warning is here:
https://github.com/apache/netbeans/blob/3f0095729e90bb458642f9aa09c954bdcf2fb877/nbbuild/jdk.xml#L116
This was introduced by 4b857cc95c5cb236427d7f488b61ef10c64a52de in January 2019

matthiasblaesing:
Looks sane - thank you.

PR #1630, opened 2019-11-15
Align OSGI symbolic bundle name and module name for SLF4J and Bouncy Castle

matthiasblaesing:


matthiasblaesing:
@lkishalmi could you please have a look at this? The result from #1625 looked problematic and thus I did some digging. I realized that there are two module systems at play: NetBeans and OSGI. The two need to align and in this case the code name base and the symbolic bundle name were different. I moved the SLF4J and Bouncy Castle libraries to their symbolic bundle names, adjusted the dependency of gradle to point to the new modules.

I activated gradle successfully and git also worked.

matthiasblaesing:
Thanks for the quick check - yeah the git diff is suboptimal... I can open gradle projects and start builds, the git functions also still work, so I'm merging this.

PR #1638, opened 2019-11-17
Always prefer showing of errors in notifications window

dukescript:
There is a nice Notifications UI that can collect and help users process exceptions. Use it always, even in development builds. E.g. don't consider the assertion status when setting the minimal report level.


geertjanw:
Looks good, any reason not to merge?

eirikbakke:
Seems like a good idea to not depend on the assertion flag.

mbien:
This would require a rebase if we still want to integrate it. Seems like everyone was happy but nobody pressed merge :)

PR #1642, opened 2019-11-20
[NETBEANS-3380] Fix default of checkOnlyImmutables option in Constant…

errael:
…Name hint. Note: this is my first pull request. It is not my intention to put a git branch into the history, but I see my local name here; convenient, but I'm not sure if that's a problem.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/suggestions/ConstantNameHint.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/suggestions/ConstantNameHint.java
@@ -127,7 +127,7 @@ public class ConstantNameHint {
         }
         Preferences prefs = ctx.getPreferences();
         
-        boolean onlyImmClasses = prefs.getBoolean(PREF_CHECK_ONLY_IMMUTABLES, false);
+        boolean onlyImmClasses = prefs.getBoolean(PREF_CHECK_ONLY_IMMUTABLES, DEFAULT_CHECK_ONLY_IMMUTABLES);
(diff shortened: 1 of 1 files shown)


emilianbold:
+1

Oddly enought that default seems to be ignored since the initial commit https://github.com/emilianbold/netbeans-releases/commit/6931108eecf53d12b0dc2f098b2faf7dd9b8ba0d#diff-e2e4f443582ccbb80235972412cd9a7a 

Although there's a commit https://github.com/emilianbold/netbeans-releases/commit/32859074a9248edc00a9e9aff29b05c7ab16510b#diff-e2e4f443582ccbb80235972412cd9a7a that references two bugzilla issues #253096, #252965 and immutable checking.

errael:
>
> ...that default seems to be ignored...
>
Could it be that in your situation the preference had been set and so the default would no longer matter?

I ran into this just after a clean install with no preferences (options) imported, and the problem went away after I toggled the option with the dialog forcing a value into the file. I was able to recreate the bug without totally starting from scratch by closing NB, deleting the preference line "onlyCheckImmutables=true" from the file

userdir/config/Preferences/org/netbeans/modules/java/hints/default/org.netbeans.modules.java.hints.suggestions.ConstantNameHint.properties,

restart NB and observe the bug: the options dialog had "only check immutable" checked, but in the editor something like

    _private static final Map<Date, Object> tcChecked = new WeakHashMap<Date, Object>();_

had hint even though _tcChecked_ was mutable.

errael:
(I'm out of practice :-) Should have included some images in the bug report)

emilianbold:
Yes, I know why it happened: nobody really tested that with a fresh install and the flag never needed the default value.

PR #1643, opened 2019-11-20
[NETBEANS-3268] Add hints for variables and method return parameters

graben:
Waiting for #1559 getting merged.


graben:
@jlahoda : Could you plz have a look on this.

jlahoda:
Two questions:
-why not member selects? (Like Math.PI.)
-I wonder if we are getting to a place where we need a customization for this, so that users can choose if they want to see the names only in some specific cases.

graben:
Never thought about member selects but let me have a look it should be easy with the code separated in its own method. I think customization will be a thing to think about once this feature is confirmed to be activated by default maybe for 12.0.

graben:
@jlahoda : Add member select to highlighter and test.

graben:
@jlahoda : Ready to merge?

graben:
@jlahoda : Can we get this into 11.3?

PR #1609, opened 2019-11-02
[NETBEANS-3311]  Providing support convert text block to string

Akshay-Gupta-Oracle:
Gives a warning for every Text block and provide a fix for converting Text blocks to String.
https://issues.apache.org/jira/browse/NETBEANS-3311

PR #1646, opened 2019-11-21
[NETBEANS-3311] Providing support convert text block to string

Akshay-Gupta-Oracle:
Gives a warning for every Text block and provide a fix for converting Text blocks to String.
https://issues.apache.org/jira/browse/NETBEANS-3311

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
@@ -44,9 +44,9 @@ public class ConvertTextBlockToString {
     @Messages("ERR_ConvertTextBlockToString=Text block may not be supported")//NOI18N
     public static ErrorDescription computeWarning(HintContext ctx) {
         TokenSequence<?> ts = ctx.getInfo().getTokenHierarchy().tokenSequence();
+        if(ts==null)return null;
         ts.move((int) ctx.getInfo().getTrees().getSourcePositions().getStartPosition(ctx.getPath().getCompilationUnit(), ctx.getPath().getLeaf()));
-        boolean moveNext = ts.moveNext();
-        if (!moveNext || ts.token().id() != JavaTokenId.MULTILINE_STRING_LITERAL) {
+        if (!ts.moveNext() || ts.token().id() != JavaTokenId.MULTILINE_STRING_LITERAL) {
             return null;
         }
 
@@ -83,24 +83,24 @@ public class ConvertTextBlockToString {
 
         }
 
-        static private ExpressionTree buildTree(String arr[], int li, TransformationContext ctx) {
-            for (int i = 0; i < arr.length; i++) {
-                arr[i] = arr[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
-                arr[i] = arr[i].replaceAll("\\Q\\\'\\E", "\'"); // NOI18N
+        static private ExpressionTree buildTree(String textBlockLines[], int currentLine, TransformationContext ctx) {
+            for (int i = 0; i < textBlockLines.length; i++) {
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\'\\E", "\'"); // NOI18N
             }
-            if (arr.length == 1) {
-                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(arr[0]), ctx.getWorkingCopy().getTreeMaker().Literal(""));
+            if (textBlockLines.length == 1) {
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[0]), ctx.getWorkingCopy().getTreeMaker().Literal(""));// NOI18N
             }
-            if (li == 1) {
-                if (li == arr.length - 1) {
-                    return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(arr[li - 1] + "\n"), ctx.getWorkingCopy().getTreeMaker().Literal(arr[li]));
+            if (currentLine == 1) {
+                if (currentLine == textBlockLines.length - 1) {
+                    return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine - 1] + "\n"), ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine]));// NOI18N
                 }
-                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(arr[li - 1] + "\n"), ctx.getWorkingCopy().getTreeMaker().Literal(arr[li] + "\n"));
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine - 1] + "\n"), ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine] + "\n"));// NOI18N
             }
-            if (li == arr.length - 1) {
-                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, buildTree(arr, li - 1, ctx), ctx.getWorkingCopy().getTreeMaker().Literal(arr[li]));
+            if (currentLine == textBlockLines.length - 1) {
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, buildTree(textBlockLines, currentLine - 1, ctx), ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine]));
             }
-            return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, buildTree(arr, li - 1, ctx), ctx.getWorkingCopy().getTreeMaker().Literal(arr[li] + "\n"));
+            return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, buildTree(textBlockLines, currentLine - 1, ctx), ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[currentLine] + "\n"));// NOI18N
         }
     }
 }
--- a/java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.java.hints.jdk;
+
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.util.TreePath;
+import org.netbeans.api.java.lexer.JavaTokenId;
+import org.netbeans.api.java.source.CompilationInfo;
+import org.netbeans.api.lexer.TokenSequence;
+import org.netbeans.spi.editor.hints.ErrorDescription;
+import org.netbeans.spi.editor.hints.Fix;
+import org.netbeans.spi.java.hints.ErrorDescriptionFactory;
+import org.netbeans.spi.java.hints.Hint;
+import org.netbeans.spi.java.hints.HintContext;
+import org.netbeans.spi.java.hints.JavaFix;
+import org.netbeans.spi.java.hints.TriggerTreeKind;
+import org.openide.util.NbBundle.Messages;
+
+@Hint(displayName = "#DN_ConvertTextBlockToString", description = "#DESC_ConvertTextBlockToString", category = "general", minSourceVersion = "13")
+@Messages({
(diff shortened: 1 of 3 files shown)


Akshay-Gupta-Oracle:
Following PR got closed accidentally while rebasing it and a new PR get created while reopening it
https://github.com/apache/netbeans/pull/1609

Apologies for any inconvenience

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+@Messages({
+    "DN_ConvertTextBlockToString=Convert Text block to String",
+    "DESC_ConvertTextBlockToString=Text Block may not be supported in older versions of java older then 13"
+})
+public class ConvertTextBlockToString {
+
+    @TriggerTreeKind(Tree.Kind.STRING_LITERAL)
+    @Messages("ERR_ConvertTextBlockToString=Text block may not be supported")//NOI18N
+    public static ErrorDescription computeWarning(HintContext ctx) {
+        TokenSequence<?> ts = ctx.getInfo().getTokenHierarchy().tokenSequence();
```
junichi11: Maybe, should rock the `Document` when `TokenSequence` is got.
(The following log should be shown: !!WARNING!! Missing WRITE-LOCK when accessing TokenHierarchy)
@jlahoda Need not do that? 

Then should check whether `ts` is `null`.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+        ts.move((int) ctx.getInfo().getTrees().getSourcePositions().getStartPosition(ctx.getPath().getCompilationUnit(), ctx.getPath().getLeaf()));
+        boolean moveNext = ts.moveNext();
+        if (!moveNext || ts.token().id() != JavaTokenId.MULTILINE_STRING_LITERAL) {
```
junichi11: Nitpick: You can also write like the following
```java
        if (!ts.moveNext() || ts.token().id() != JavaTokenId.MULTILINE_STRING_LITERAL) {
```

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        @Override
+        protected void performRewrite(TransformationContext ctx) {
+            ExpressionTree ext = buildTree(orignalStringArr, orignalStringArr.length - 1, ctx);
+            ctx.getWorkingCopy().rewrite(ctx.getPath().getLeaf(), ext);
+
+        }
+
+        static private ExpressionTree buildTree(String arr[], int li, TransformationContext ctx) {
```
junichi11: Should use proper parameter names because it is not easy to understand what `arr` and `li` mean.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+            for (int i = 0; i < arr.length; i++) {
+                arr[i] = arr[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
+                arr[i] = arr[i].replaceAll("\\Q\\\'\\E", "\'"); // NOI18N
+            }
+            if (arr.length == 1) {
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(arr[0]), ctx.getWorkingCopy().getTreeMaker().Literal(""));
```
junichi11: Should add `// NOI18N`
`"\n"`s as well

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Can you please also tell that how to lock the document??

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Done, thanks

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Thank you for suggestion.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
junichi11: e.g. see https://bits.netbeans.org/dev/javadoc/org-netbeans-modules-lexer/architecture-summary.html

@sdedic @jlahoda Could you tell @Akshay-Gupta-Oracle to handle it if you know the proper way in Java editor? or need not do that? Thanks.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+    "DN_ConvertTextBlockToString=Convert Text block to String",
+    "DESC_ConvertTextBlockToString=Text Block may not be supported in older versions of java older then 13"
+})
+public class ConvertTextBlockToString {
+
+    @TriggerTreeKind(Tree.Kind.STRING_LITERAL)
+    @Messages("ERR_ConvertTextBlockToString=Text block may not be supported")//NOI18N
+    public static ErrorDescription computeWarning(HintContext ctx) {
+        TokenSequence<?> ts = ctx.getInfo().getTokenHierarchy().tokenSequence();
+        if(ts==null)return null;
```
junichi11: Please use braces `{}` then please add spaces around `==`, before `(`, after `)`.
i.e.
```java
if (ts == null) {
    return null;
}
```

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        String orignalString = ctx.getPath().getLeaf().toString();
+        String orignalStringArr[] = textBlockToStringArr(orignalString.substring(1, orignalString.length() - 1));
+        Fix fix = new FixImpl(ctx.getInfo(), ctx.getPath(), orignalStringArr).toEditorFix();
+        return ErrorDescriptionFactory.forName(ctx, ctx.getPath(), Bundle.ERR_ConvertTextBlockToString(), fix);
+    }
+
+    private static String[] textBlockToStringArr(String textBlock) {
+        String oneLine[] = textBlock.split("\\\\n", -1);    // NOI18N
```
sdedic: When processing 
```
        String s = """   
                   
                   
        """;
```
The hint will produce code like
```
String s = "\n" + "\n" + "";
```

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+
+        }
+
+        static private ExpressionTree buildTree(String textBlockLines[], int currentLine, TransformationContext ctx) {
+            for (int i = 0; i < textBlockLines.length; i++) {
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\'\\E", "\'"); // NOI18N
+            }
+            if (textBlockLines.length == 1) {
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[0]), ctx.getWorkingCopy().getTreeMaker().Literal(""));// NOI18N
```
sdedic: I don't understand this one -- if there's just line, no newlines included (ending `"""` delimiter follows the content string on the same line), the produced code is still
```
        String s = "s" + "";
```

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+        @Override
+        protected void performRewrite(TransformationContext ctx) {
+            ExpressionTree ext = buildTree(orignalStringArr, orignalStringArr.length - 1, ctx);
+            ctx.getWorkingCopy().rewrite(ctx.getPath().getLeaf(), ext);
+
+        }
+
+        static private ExpressionTree buildTree(String textBlockLines[], int currentLine, TransformationContext ctx) {
+            for (int i = 0; i < textBlockLines.length; i++) {
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
```
sdedic: This works for escaped (double)quotes, but not for other escape characters. So for example
```
        String s = """   
            \\
            """;
```
yields
```
String s = "\\\\\n" + "";
```

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+    @Messages("ERR_ConvertTextBlockToString=Text block may not be supported")//NOI18N
+    public static ErrorDescription computeWarning(HintContext ctx) {
+        TokenSequence<?> ts = ctx.getInfo().getTokenHierarchy().tokenSequence();
+        if(ts==null)return null;
+        ts.move((int) ctx.getInfo().getTrees().getSourcePositions().getStartPosition(ctx.getPath().getCompilationUnit(), ctx.getPath().getLeaf()));
+        if (!ts.moveNext() || ts.token().id() != JavaTokenId.MULTILINE_STRING_LITERAL) {
+            return null;
+        }
+
+        String orignalString = ctx.getPath().getLeaf().toString();
```
sdedic: I think that some errors described later could be fixed, if here `((LiteralTree)ctx.getPath().getLeaf()).getValue()` is used instead of `toString`, which implicitly quotes.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+
+        static private ExpressionTree buildTree(String textBlockLines[], int currentLine, TransformationContext ctx) {
+            for (int i = 0; i < textBlockLines.length; i++) {
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\"\\E", "\""); // NOI18N
+                textBlockLines[i] = textBlockLines[i].replaceAll("\\Q\\\'\\E", "\'"); // NOI18N
+            }
+            if (textBlockLines.length == 1) {
+                return ctx.getWorkingCopy().getTreeMaker().Binary(Tree.Kind.PLUS, ctx.getWorkingCopy().getTreeMaker().Literal(textBlockLines[0]), ctx.getWorkingCopy().getTreeMaker().Literal(""));// NOI18N
+            }
+            if (currentLine == 1) {
```
sdedic: Really so many cases need separate handling ?

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
junichi11: @sdedic Thank you for making it clear!

Akshay-Gupta-Oracle: Thank you @sdedic and @junichi11 for helping :)

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Ok, did this, thanks

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: this is to support going back and forth for textblock to string and string to textblock.
does extra "" cause any issue??
for example if I dont put extra +"" then
          String s="""
                         akshay
                        """;
will turn into String s="akshay\n"; and this does not give hint to convert to TextBlock
However "akshay\n" + ""; does

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Do you think it should be 
String s = "\n" + "\n";??

sdedic: According to the specification, the literal content starts AFTER the first newline. So there are two newlines ... without any content after the 2nd one. Yes, functionally it is a no-op, but having a *refactoring* to produce redundant code like *+ ""*  is ... well, not a production quality.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
sdedic: It does not any issue. It's just redundant so it should not be produced by a refactoring.

The question is why `String s="akshay\n";` does not allow conversion to a text block ? You can call it a feature (it IMHO actually spoils the formatting if converted, in this case) - a user how hit convert to String and does not like the outcome can Undo the operation. 

An option could be made, to provide the hint on all strings with **at least** one newline within them, or at least N newlines. The user can then tune up to his desired style.

But IMHO there's no use-case for generating redudnant code that would cause code reviewer to demand a fix.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
```diff
@@ -0,0 +1,106 @@
...
+    "DESC_ConvertTextBlockToString=Text Block may not be supported in older versions of java older then 13"
+})
+public class ConvertTextBlockToString {
+
+    @TriggerTreeKind(Tree.Kind.STRING_LITERAL)
+    @Messages("ERR_ConvertTextBlockToString=Text block may not be supported")//NOI18N
+    public static ErrorDescription computeWarning(HintContext ctx) {
+        TokenSequence<?> ts = ctx.getInfo().getTokenHierarchy().tokenSequence();
+        if(ts==null)return null;
+        ts.move((int) ctx.getInfo().getTrees().getSourcePositions().getStartPosition(ctx.getPath().getCompilationUnit(), ctx.getPath().getLeaf()));
```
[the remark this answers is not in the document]
Akshay-Gupta-Oracle: Done thank you

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertTextBlockToString.java
Akshay-Gupta-Oracle: Done

Akshay-Gupta-Oracle:
@sdedic 
Done with the changes.

Akshay-Gupta-Oracle:
@ebarboni please review it. can it be accommodated in 11.3??

PR #1648, opened 2019-11-21
Fixing end position in the document change event in case multiple lin…

jlahoda:
…es are deleted.


GeertjanWielenga:
Great, thank you.

PR #1651, opened 2019-11-23
[NETBEANS-3427] Import Dark Metal and Dark Nimbus LaF from Oracle Donation 4

lkishalmi:
Just did the necessary modifications for the import. Changing license headers, and include the module into the build chain.

Right now the module automatically switches to Dark Metal on install, also the java level can be moved from 6 -> 8. I think that could be dealt with later when this "historical" import PR got merged.


lkishalmi:
image
image

PR #1652, opened 2019-11-23
[NETBEANS-3428] Initial support of FlatLaf Look and Feels

lkishalmi:
Well, this is just a demo PR. If the community agrees and someone would iron this out and if Travis agrees we can even merge it.
Almost no work has been done on my side, just a small kick-off

image

As you see there are a lot of work to do.


platform/o.n.swing.flatlaf/src/org/netbeans/swing/flatlaf/Bundle.properties
```diff
@@ -0,0 +1,9 @@
+# To change this license header, choose License Headers in Project Properties.
```
eirikbakke: Travis complained about this one...

lkishalmi: Fixed that one.

lkishalmi:
#1651 Shows a bit more what could/needs to be done around the LaF. Also a matching editor color theme would be good. Checked it with Norway Today and Citi Lights, not good.

platform/o.n.swing.flatlaf/src/org/netbeans/swing/flatlaf/Installer.java
```diff
@@ -0,0 +1,34 @@
...
+package org.netbeans.swing.flatlaf;
+
+import javax.swing.UIManager;
+import org.openide.modules.ModuleInstall;
+
+public class Installer extends ModuleInstall {
+
+    @Override
+    public void validate() throws IllegalStateException {
+        UIManager.installLookAndFeel(new UIManager.LookAndFeelInfo("Flat Dark", NbFlatDarkLaf.class.getName()));
```
eirikbakke: If you're planning on merging this before it's fully done, maybe rename this "FlatLaf Dark (experimental)" so people don't assume it's meant to be used when they see it as an option in the Preferences pane.

lkishalmi:
I'd just merge as it is. Let the community polish it out. These issues are usually easy fixes and quite visible, so I guess would attract contributors. Though it would be good that someone more artistic than me, could guide the style (color palette, font, spacing, etc.) guidance to this.

geertjanw:
Agree with @lkishalmi and this is great work.

neilcsmith-net:
+1 with caveats - would prefer to see this hidden by a system property flag until it's ready, or do the work in a separate branch rather than master?  Or failing that, agree with @eirikbakke mark as experimental. Current release process is meant to mean anything in master is release ready!

A font colour scheme similar to how Darcula would probably work well.  I have a Darcula fork in use at the moment - will check the font colors in that.

Judging from that screenshot, it needs my Options hack. :wink: https://github.com/Revivius/nb-darcula/blob/master/src/main/java/com/revivius/nb/darcula/ui/OptionsAwareLabelUI.java

lkishalmi:
> Judging from that screenshot, it needs my Options hack. wink https://github.com/Revivius/nb-darcula/blob/master/src/main/java/com/revivius/nb/darcula/ui/OptionsAwareLabelUI.java

Well it needs a lot of hacks, of course yours are welcome as well!

lkishalmi:
Let's the play begin.

ramiswailem:
Nice work!

Error when building with cmd

ant -DthreadCount=16 -quiet -silent -nice 10 -Dcluster.config=full -Ddo.build.windows.launchers=true build build-nbms generate-uc-catalog build-source-zips -Dpermit.jdk9.builds=true -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false;


/root/NetBeansProjects/netbeans-src/nbbuild/build.xml:660: Target "all-o.n.swing.laf.dark" does not exist in the project "main". It is used from target "nbmerge-build-one-cluster".

lkishalmi:
o.n.swing.laf.dark is another PR for Dark Nimbus and Dark Metal

PR #1656, opened 2019-11-24
[NETBEANS-2533] Update CSS browser compatibility help - include versi…

peedeeboy:
**What this PR does**
  * Updates the CSS help to include the version number that a property was supported by a given browser.  Data is based on [Mozilla browser-compat-data](https://github.com/mdn/browser-compat-data)
  * Adds in Edge, Chome Android, Opera Android, Firefox Android & Safari iOS browsers

css_help

**Licensing**
browser-compat-data is licensed under [Create Commons Zero (CCO) Public Domain](https://github.com/mdn/browser-compat-data/blob/master/LICENSE)

If I am interpreting things correctly:
  * CC0 allows the original data to be manipulated (I have some shell scripts to parse the JSON to create Java properties files)
  * Apache's guidance seems to suggest these should be treated as Category A (?), and that attribution is required.  I have added links into the files based on brower-compat-data, should I do more?


peedeeboy:
Unit tests in css.lib module are failing as I'd missed that the Zoom CSS property is only currently defined in the webkit.properties file (as <match-anything>).

Could somebody with the appropriate permissions please label this PR as Work In Progress, whilst I think/work on a good solution for dealing with browser specific properties where we don't have a regular CSS property to proxy to?

Chris2011:
I think it is good to go but wouldn't it be better to load he data from the server instead of hardcoded it into it? Everytime they change smth and I think it changes a lot, we need to update the stuff. Or am I wrong?

PR #1617, opened 2019-11-05
[NETBEANS-3335] Enhance robustness of HTML Lexer

matthiasblaesing:
After the update of the HTML lexer to support HTML5 style attributes,
new states in the lexer can be reached. Instead of bailing out with a
NullPointerException in HtmlLexer#equals, report comparisons with NULL
as false.

PR #1658, opened 2019-11-24
Backport of PR #1617

matthiasblaesing:
As requested in #1617 this is a straight backport of the patches
to the release112 branch. For details about the changeset please
see the original PR, the only additional change is the bump of
the spec version.


neilcsmith-net:
Thanks @matthiasblaesing Will merge and test with view to vote and pushing update this week.

To double check, looking at just shipping org-netbeans-modules-html-lexer.nbm ?

matthiasblaesing:
Yes shipping `org-netbeans-modules-html-lexer.nbm` will be enough. The changes in `php.smarty` are the unittests, that Junichi provided and are not needed at runtime.

PR #1661, opened 2019-11-24
[NETBEANS-3440] Allow to define a palette for fontcolors XML Schema.

lkishalmi:
This one contains a new DTD which probably have to be installed on netbeans.apache.org/dtds/ as well.
It might help for #1652 in the future.

### DIFF
--- a/ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
+++ b/ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
@@ -608,7 +608,8 @@ public final class ColoringStorage implements StorageDescription<String, Attribu
         try {
             return new Color ((int) Long.parseLong (color, 16), (color.length() != 6)); // formats: #AARRGGBB or #RRGGBB (Alpha is defaulted to 255)
         } catch (NumberFormatException ex) {
-            throw new Exception ();
+            LOG.log(Level.WARNING, "Can''t parse ''{0}'' to a Color. Using RED as a replacement.", color);
+            return Color.red;
         }
     }
 }
--- a/ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
+++ b/ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
@@ -237,15 +237,15 @@ public final class ColoringStorage implements StorageDescription<String, Attribu
                 FileObject baseFolder = FileUtil.getConfigFile("Editors"); //NOI18N
                 FileObject f = FileUtil.createData(baseFolder, settingFileName);
                 f.setAttribute(FA_TYPE, type);
-                
-                Map<String, AttributeSet> added = new HashMap<String, AttributeSet>();
-                Map<String, AttributeSet> removed = new HashMap<String, AttributeSet>();
+
+                    Map<String, AttributeSet> added = new HashMap<String, AttributeSet>();
+                    Map<String, AttributeSet> removed = new HashMap<String, AttributeSet>();
                 Utils.diff(defaultFontColors, fontColors, added, removed);
-                
+
                 ColoringsWriter writer = new ColoringsWriter();
                 writer.setAdded(added);
                 writer.setRemoved(removed.keySet());
-                
+
                 Utils.save(f, writer);
             }
         });
@@ -258,7 +258,7 @@ public final class ColoringStorage implements StorageDescription<String, Attribu
         assert profile != null : "The parameter profile must not be null"; //NOI18N
         
         FileObject baseFolder = FileUtil.getConfigFile("Editors"); //NOI18N
-        Map<String, List<Object []>> files = new HashMap<String, List<Object []>>();
+        Map<String, List<Object []>> files = new HashMap<>();
         SettingsType.getLocator(this).scan(baseFolder, mimePath.getPath(), profile, true, defaults, !defaults, false, files);
         
         assert files.size() <= 1 : "Too many results in the scan"; //NOI18N
@@ -289,9 +289,11 @@ public final class ColoringStorage implements StorageDescription<String, Attribu
     private static final String HIGHLIGHTING_FILE_NAME = "editorColoring.xml"; // NOI18N
     
     private static final String E_ROOT = "fontscolors"; //NOI18N
+    private static final String E_COLORDEF = "colordef"; //NOI18N
     private static final String E_FONTCOLOR = "fontcolor"; //NOI18N
     private static final String E_FONT = "font"; //NOI18N
     private static final String A_NAME = "name"; //NOI18N
+    private static final String A_COLOR = "color"; //NOI18N
     private static final String A_FOREGROUND = "foreColor"; //NOI18N
     private static final String A_BACKGROUND = "bgColor"; //NOI18N
     private static final String A_STRIKETHROUGH = "strikeThrough"; //NOI18N
@@ -305,8 +307,8 @@ public final class ColoringStorage implements StorageDescription<String, Attribu
     private static final String V_ITALIC = "italic"; //NOI18N
     private static final String V_PLAIN = "plain"; //NOI18N
     
-    private static final String PUBLIC_ID = "-//NetBeans//DTD Editor Fonts and Colors settings 1.1//EN"; //NOI18N
-    private static final String SYSTEM_ID = "http://www.netbeans.org/dtds/EditorFontsColors-1_1.dtd"; //NOI18N
+    private static final String PUBLIC_ID = "-//NetBeans//DTD Editor Fonts and Colors settings 1.2//EN"; //NOI18N
+    private static final String SYSTEM_ID = "http://netbeans.apache.org/dtds/EditorFontsColors-1_2.dtd"; //NOI18N
 
     private static final String FA_TYPE = "nbeditor-settings-ColoringType"; //NOI18N
     public static final String FAV_TOKEN = "token"; //NOI18N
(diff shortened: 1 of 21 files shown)


ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -258,25 +264,23 @@ public void delete(MimePath mimePath, String profile, boolean defaults) throws I
...
+                for(Object [] info : profileInfos) {
+                    FileObject settingFile = (FileObject) info[1];
 
-                        settingFile.delete();
+                    // Skip files with wrong type of colorings
+                    if (!type.equals(getColoringFileType(settingFile))) {
+                        continue;
                     }
+
+                    settingFile.delete();
```
eirikbakke: Doing a separate commit with just cosmetic changes (lambdafication, diamond operator, override annotations etc.) would encourage quick reviews in the future :-)

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -305,8 +311,8 @@ public void run() throws IOException {
     private static final String V_ITALIC = "italic"; //NOI18N
     private static final String V_PLAIN = "plain"; //NOI18N
     
-    private static final String PUBLIC_ID = "-//NetBeans//DTD Editor Fonts and Colors settings 1.1//EN"; //NOI18N
-    private static final String SYSTEM_ID = "http://www.netbeans.org/dtds/EditorFontsColors-1_1.dtd"; //NOI18N
+    private static final String PUBLIC_ID = "-//NetBeans//DTD Editor Fonts and Colors settings 1.2//EN"; //NOI18N
+    private static final String SYSTEM_ID = "http://netbeans.apache.org/dtds/EditorFontsColors-1_2.dtd"; //NOI18N
```
eirikbakke: Can we still correctly read settings of the previous format?

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -365,7 +372,14 @@ public ColoringsReader(FileObject f, String mimePath) {
             try {
                 if (name.equals(E_ROOT)) {
                     // We don't read anythhing from the root element
-                    
+                    colorrefs.clear();;
```
eirikbakke: double semicolon

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -365,7 +372,14 @@ public ColoringsReader(FileObject f, String mimePath) {
...
+                } else if (name.equals(E_COLORREF)) {
+                    String nameAttr = attributes.getValue(A_NAME);
+                    String colorAttr = attributes.getValue(A_COLOR);
+                    String oldRef = colorrefs.put(nameAttr, colorAttr);
```
eirikbakke: Should null-check and check validity of nameAttr/coloAttr before putting them in the map

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -365,7 +372,14 @@ public ColoringsReader(FileObject f, String mimePath) {
...
+                    if (oldRef != null) {
+                        LOG.log(Level.WARNING, "Color reference: '" + nameAttr + "' previosly defined as '" + oldRef + "' is now '" + colorAttr +"'");
```
eirikbakke: previosly should be previously

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -289,9 +293,11 @@ public void run() throws IOException {
     private static final String HIGHLIGHTING_FILE_NAME = "editorColoring.xml"; // NOI18N
     
     private static final String E_ROOT = "fontscolors"; //NOI18N
+    private static final String E_COLORREF = "colorref"; //NOI18N
```
eirikbakke: It seems that this is the element that _defines_ the color reference, rather than uses it. I would call that "colordef" rather than "colorref", or something similar. The "reference" is the place that refers to the previously defined color.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
eirikbakke: Add a check that nameAttr is not, in fact, a valid color string. This could lead to very confusing results otherwise, e.g. redefining "ff0000" by accident. Perhaps require color references to use some special syntax, e.g. a color defined as "editorBackground" must be referenced "$editorBackground" (do colorrefs.put("$" + nameAttr, colorAttr) to achieve this). Perhaps also require the names to follow Java identifier requirements, e.g. start with Character.isJavaIdentifierStart and every subsequent character must be Character.isJavaIdentifierPart.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
```diff
@@ -447,6 +456,12 @@ public ColoringsReader(FileObject f, String mimePath) {
                 attribs = null;
             }
         }
+
+        Color stringToColor(String name) throws Exception {
+            String resolve = colorrefs.getOrDefault(name, name);
+            return ColoringStorage.stringToColor(resolve);
```
eirikbakke: See previous comment about requiring references to follow a reserved syntax that can't be confused with a real color.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/EditorFontsColors-1_2.dtd
```diff
@@ -0,0 +1,62 @@
...
+        The vocabulary for an editor fonts and colors settings file.
+
+    PUBLIC ID  : -//NetBeans//DTD Editor Fonts and Colors settings 1.2//EN
+    SYSTEM ID  : http://netbeans.apache.org/dtds/EditorFontsColors-1_2.dtd
+-->
+
+<!ELEMENT fontscolors ( colorref*, fontcolor* ) >
+<!ATTLIST fontscolors >
+
+<!ELEMENT colorref EMPTY >
```
eirikbakke: As before, maybe call this colordef rather than colorref.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: Going to do that.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: Yes. Tests can approve that.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/EditorFontsColors-1_2.dtd
lkishalmi: colordef is better, idea accepted! Thank you!

neilcsmith-net:
Bit tied up with other things to review properly, but +1 to the principle - great idea!

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: The cosmetic changes ended up in a separate commit.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: fixed

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: Null check is actually done by the parser as both name and color attributes are requested. If they are missing we got a parser exception first before we could check for null.
As of masking predefined colors, I do not see any problem with that. These color definitions is local to the file where they are defined. The whole idea is coming from my attempt to port https://github.com/morhetz/gruvbox/ to NetBeans. In order to support both light and dark mode, I'd just flip the palette definitions.

As of add a distinctive syntax when referring the color definitions. I was thinking about it as well, though I'm natural on that one. The only reason for opting that out is only that less typing makes me happy.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: fixed

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: Changed.

ide/editor.settings.storage/src/org/netbeans/modules/editor/settings/storage/fontscolors/ColoringStorage.java
lkishalmi: Answered there.

PR #1662, opened 2019-11-24
[NETBEANS-3439] - Upgrade JavaCC from 3.2 to 7.0.5

pepness:
Notes:
-Add support for Java7 language features
-Requires Java 7
-Added support for C++
-Internal refactoring
-Several bug fixes and enhancements
-Update license

[JavaCC License](https://github.com/javacc/javacc/blob/master/LICENSE)
[JavaCC Web Page](https://github.com/javacc/javacc)
[JavaCC Release Notes](https://github.com/javacc/javacc/blob/master/docs/release-notes.md)


wumpz:
Be aware that version 7.0.5 has some incompatibilities with 7.0.4 that still has to be adressed:

https://github.com/javacc/javacc/pull/73

pepness:
Hi Tobias,

Thanks for pointing this out, I will review it.

lkishalmi:
BUILD FAILED
/home/travis/build/apache/netbeans/nbbuild/build.xml:1614: Failed VerifyLibsAndLicenses test(s):
ide/db.sql.visualeditor/external/javacc-7.0.5-license.txt refers to nonexistent nbbuild/licenses/BSD-javacc-7.0.5
java/performance/external/javacc-7.0.5-license.txt refers to nonexistent nbbuild/licenses/BSD-javacc-7.0.5

java/performance/external/binaries-list
```diff
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-EE416CB5D6AA88473EFE487F42DC1410360F7948 javacc-3.2.jar
+3314CFC8BBD82725013EA997506EE5C525E8153A javacc-7.0.5.jar
```
ebarboni: Could it be possible to use the javacc-7.05 like this 
3314CFC8BBD82725013EA997506EE5C525E8153A net.java.dev.javacc:javacc:7.0.5

matthiasblaesing: Yes and that would be the right way. We try to drop usage of the binary server to the cases where it is actually necessary, a standard artifact from maven central should be preferred.

pepness: I fix it, thanks for the observation.

ebarboni:
in this folder there is a tracking of license. BSD seems to be per libraries
https://github.com/apache/netbeans/tree/master/nbbuild/licenses 

You should move the file 3.2 to 7.0.5, and needs to be changed to actual content in 7.0.5 of javacc.

pepness:
Fixed it and I updated the PR description. Thanks.

ebarboni:
I think the performance/benchmark is not active yet. But I didn't see a wrong generation. This is not very testable IMHO.

ebarboni:
if we drop we should ask dev ML.
maybe we should store to another "git" the removed to make visible what's gone. To be able to redo again.

matthiasblaesing:
The regression noticed before is fixed in 7.0.10 (according to changelog), travis and github actions are happy, this has minimal impact on the IDE, so lets get this in.

PR #1668, opened 2019-11-25
[NETBEANS-3450] Fixed compiler warnings concerning rawtypes Stream

mklaehn:
There are compiler warnings about rawtype usage with a Stream like the following
```
   [repeat] /Users/martin/Repositories/git/netbeans/ide/extexecution.process.jdk9/src/org/netbeans/modules/extexecution/process/jdk9/ProcessesImpl.java:77: warning: [rawtypes] found raw type: Stream
   [repeat]             try (Stream s = (Stream) PROCESS_HANDLE_DESCENDANTS.invoke(handle, (Object[]) null)) {
   [repeat]                  ^
   [repeat]   missing type arguments for generic class Stream<T>
   [repeat]     T extends Object declared in interface Stream
```


mklaehn:
Merging (1 approval without comment in the last nine days)

PR #1669, opened 2019-11-25
[NETBEANS-3451] Fixed compiler warnings concerning rawtypes Iterable

mklaehn:
There are compiler warnings about rawtype usage with a Iterable like the following
```
   [repeat] .../java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java:197: warning: [rawtypes] found raw type: Iterable
   [repeat]     public static <E> Iterable<E> checkedIterableByFilter(final Iterable raw, final Class<E> type, final boolean strict) {
   [repeat]                                                                 ^
   [repeat]   missing type arguments for generic class Iterable<T>
   [repeat]     T extends Object declared in interface Iterable
```


mklaehn:

PR #1670, opened 2019-11-25
[NETBEANS-3453] - Upgrade mysql-connector-java from 5.1.23 to 5.1.48

pepness:
Changes since version 5.1.23:
- 207 Bugs fixes
- 51 Improvements

[MySQL Connector Java Web Page](https://dev.mysql.com/downloads/connector/j/)
[MySQL Connector Java Release Notes](https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1.html)


emilianbold:
Note NetBeans does not bundle the connector due to licensing reasons. So, incrementing that version seems like a good idea, but it's only used for one test...

If you are working in this area, take a look at MySQLProcedure.java and `.createValue` which mentions http://bugs.mysql.com/bug.php?id=41269 which has been closed for a decade.

PR #1671, opened 2019-11-25
[NETBEANS-3454] Fixed compiler warnings concerning rawtypes Reference

mklaehn:
There are compiler warnings about rawtype usage with a Reference like the following

 [nb-javac] .../platform/openide.util.ui/test/unit/src/org/openide/util/actions/CallbackSystemActionTest.java:102: warning: [rawtypes] found raw type: Reference
 [nb-javac]         Reference actionref = new WeakReference(systemaction);
 [nb-javac]         ^
 [nb-javac]   missing type arguments for generic class Reference<T>
 [nb-javac]   where T is a type-variable:
 [nb-javac]     T extends Object declared in class Reference

### DIFF
--- a/platform/openide.nodes/src/org/openide/nodes/BeanChildren.java
+++ b/platform/openide.nodes/src/org/openide/nodes/BeanChildren.java
@@ -48,8 +48,8 @@ public class BeanChildren extends Children.Keys {
      * and child, resp.
      * See #7925.
      */
-    private static final java.util.Map<Node,Reference[]> nodes2Beans = 
-            new WeakHashMap<Node,Reference[]>(); // Map<Node,Reference<Object>[2]>
+    private static final java.util.Map<Node, Reference<?>[]> nodes2Beans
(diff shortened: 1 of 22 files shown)


platform/openide.nodes/src/org/openide/nodes/BeanChildren.java
```diff
@@ -48,8 +48,8 @@
      * and child, resp.
      * See #7925.
      */
-    private static final java.util.Map<Node,Reference[]> nodes2Beans = 
-            new WeakHashMap<Node,Reference[]>(); // Map<Node,Reference<Object>[2]>
+    private static final java.util.Map<Node, Reference<?>[]> nodes2Beans
```
svenreimers: Formatting style change - is this intentional?

mklaehn: I've had NetBeans format the changed lines.
So yes, Intentional.

mklaehn:

PR #1672, opened 2019-11-25
[NETBEANS-3455] Fixed compiler warnings concerning rawtypes SoftRefer…

mklaehn:
…ence

There are compiler warnings about rawtype usage with a SoftReference like the following
```
    [repeat] .../webcommon/javascript2.model/src/org/netbeans/modules/javascript2/model/api/Index.java:215: warning: [rawtypes] found raw type: SoftReference
   [repeat]                         CACHE_INDEX_RESULT_LARGE.put(key, new SoftReference(value));
   [repeat]                                                               ^
   [repeat]   missing type arguments for generic class SoftReference<T>
   [repeat]     T extends Object declared in class SoftReference
```


mklaehn:

PR #1674, opened 2019-11-26
Backport NETBEANS-3290 fix to 11.2 release branch.

neilcsmith-net:
Any comments @jlahoda / @geertjanw on shipping updates to the platform harness via NBM update?


neilcsmith-net:
@geertjanw yes, seems to apply OK locally, but a bit concerned about the structure of this one compared to other updates.  Still, if it does cause any issues in updating we can always revert the update catalog reference for that one NBM.

neilcsmith-net:
OK, tests passed so merging.

PR #1676, opened 2019-11-26
[NETBEANS-3458] Fixed compiler warnings concerning rawtypes Performer

mklaehn:
There are compiler warnings about rawtype usage with a Performer like the following
```
   [repeat] .../platform/openide.awt/src/org/openide/awt/ContextAction.java:461: warning: [rawtypes] found raw type: Performer
   [repeat]                 Performer l = (Performer)obj;
   [repeat]                 ^
   [repeat]   missing type arguments for generic class Performer<Data>
   [repeat]   where Data is a type-variable:
   [repeat]     Data extends Object declared in class Performer
```


mklaehn:
Merging (1 approval without comment requesting a change in the last eight days)

PR #1677, opened 2019-11-26
[NETBEANS-3459] Fixed compiler warnings concerning rawtypes TargetPan…

mklaehn:
…elProvider

There are compiler warnings about rawtype usage with a TargetPanelProvider like the following
```
   [repeat] .../ide/target.iterator/src/org/netbeans/modules/target/iterator/api/TargetChooserPanel.java:218: warning: [rawtypes] found raw type: TargetPanelProvider
   [repeat]         for (TargetPanelProvider targetPanelProvider : providers) {
   [repeat]              ^
   [repeat]   missing type arguments for generic class TargetPanelProvider<T>
   [repeat]     T extends Object declared in interface TargetPanelProvider
```


mklaehn:
Merging (1 approval without comment requesting a change in the last eight days)

PR #1680, opened 2019-11-26
[NETBEANS-3462] Fixed compiler warnings concerning rawtypes SimpleFuture

mklaehn:
There are compiler warnings about rawtype usage with a SimpleFuture like the following
```
 [nb-javac] .../enterprise/maven.j2ee/src/org/netbeans/modules/maven/j2ee/ear/model/ApplicationMetadataModelImpl.java:110: warning: [rawtypes] found raw type: SimpleFuture
 [nb-javac]         return new SimpleFuture(runReadAction(action));
 [nb-javac]                    ^
 [nb-javac]   missing type arguments for generic class SimpleFuture<R>
 [nb-javac]   where R is a type-variable:
 [nb-javac]     R extends Object declared in class SimpleFuture
```


lkishalmi:
Seems trivial. Let's move it to the master.

PR #1681, opened 2019-11-26
[NETBEANS-3463] Fixed compiler warnings concerning rawtypes Observabl…

mklaehn:
…eActionListener

There are compiler warnings about rawtype usage with a ObservableActionListener like the following
```
   [repeat] .../ide/dlight.nativeexecution/src/org/netbeans/modules/nativeexecution/support/ObservableAction.java:150: warning: [rawtypes] found raw type: ObservableActionListener
   [repeat]         for (ObservableActionListener l : ll) {
   [repeat]              ^
   [repeat]   missing type arguments for generic class ObservableActionListener<T>
   [repeat]     T extends Object declared in interface ObservableActionListener
```


lkishalmi:
Seems trivial. Let's move it to the master.

PR #1682, opened 2019-11-26
[NETBEANS-3464] Fixed compiler warnings concerning rawtypes WeakSet

mklaehn:
There are compiler warnings about rawtype usage with a WeakSet like the following
```
   [repeat] .../webcommon/web.javascript.debugger/src/org/netbeans/modules/web/javascript/debugger/locals/VariablesTreeExpansionModel.java:47: warning: [rawtypes] found raw type: WeakSet
   [repeat]     private Set expandedNodes = new WeakSet();
   [repeat]                                     ^
   [repeat]   missing type arguments for generic class WeakSet<E>
   [repeat]     E extends Object declared in class WeakSet
```


lkishalmi:
Trivial. Merged.

PR #1685, opened 2019-11-27
[NETBEANS-3474] Added o.n.swing.laf.dark and o.n.swing.flatlaf to the core.kit

lkishalmi:


lkishalmi:
Merging this as it makes the new LaF-s available in platform also fixing the the dark LaF is accidentally ended up in the extra cluster.
Travis is only failing the missing hibernate.

PR #1686, opened 2019-11-27
[NETBEANS-3476] - cleanup redundant casts..

BradWalker:
There are a lot of places where the compiler tells us we are using redundant casts..

   [repeat] /home/bwalker/src/netbeans/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/parser/GroovyVirtualSourceProvider.java:436: warning: [cast] redundant cast to Statement
   [repeat]             Statement stat = (Statement) stats.get(0);
   [repeat]                              ^

This change removes all the redundant casts..


matthiasblaesing:
Agreed.

PR #1689, opened 2019-11-27
[NETBEANS-3479] Fixed compiler warnings concerning rawtypes LinkedHas…

mklaehn:
…hMap

There are compiler warnings about rawtype usage with a LinkedHashMap like the following
```
 [nb-javac] .../ide/spi.debugger.ui/test/unit/src/org/netbeans/api/debugger/PropertiesTest.java:460: warning: [rawtypes] found raw type: LinkedHashMap
 [nb-javac]             DEFAULT_VALUES.put(PROPERTY_NAMES[12], new LinkedHashMap());
 [nb-javac]                                                        ^
 [nb-javac]   missing type arguments for generic class LinkedHashMap<K,V>
 [nb-javac]   where K,V are type-variables:
 [nb-javac]     K extends Object declared in class LinkedHashMap
 [nb-javac]     V extends Object declared in class LinkedHashMap
```

### DIFF
--- a/webcommon/web.clientproject.api/src/org/netbeans/modules/web/clientproject/api/json/JsonFile.java
+++ b/webcommon/web.clientproject.api/src/org/netbeans/modules/web/clientproject/api/json/JsonFile.java
@@ -70,7 +70,7 @@ public final class JsonFile {
     private static final ContainerFactory CONTAINER_FACTORY = new ContainerFactory() {
 
         @Override
-        public Map<Object, Object> createObjectContainer() {
+        public Map<String, Object> createObjectContainer() {
             return new LinkedHashMap<>();
         }
 
--- a/webcommon/web.clientproject.api/src/org/netbeans/modules/web/clientproject/api/json/JsonFile.java
+++ b/webcommon/web.clientproject.api/src/org/netbeans/modules/web/clientproject/api/json/JsonFile.java
@@ -70,13 +70,13 @@ public final class JsonFile {
     private static final ContainerFactory CONTAINER_FACTORY = new ContainerFactory() {
 
         @Override
-        public Map createObjectContainer() {
-            return new LinkedHashMap();
+        public Map<Object, Object> createObjectContainer() {
+            return new LinkedHashMap<>();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


webcommon/web.clientproject.api/src/org/netbeans/modules/web/clientproject/api/json/JsonFile.java
```diff
@@ -70,13 +70,13 @@
     private static final ContainerFactory CONTAINER_FACTORY = new ContainerFactory() {
 
         @Override
-        public Map createObjectContainer() {
-            return new LinkedHashMap();
+        public Map<Object, Object> createObjectContainer() {
```
matthiasblaesing: This is the implementation of the `org.json.simple.parser.ContainerFactory` interface. If the simple json library introduces generics, this might blow. I had a similar problem a few weeks ago. Someone implemented an interface from JNA and decided to stick generics onto it (at that time JNA used a rawtype). This broke as recent JNA versions introduced generics at that point and were incompatible with the generic signature. At this point for example `<String,Object>` could be a valid alternative.

mklaehn: Done.

matthiasblaesing: Sorry - misunderstanding - that was _one_ option how the API of `org.json.simple.parser.ContainerFactory` could be modified by the authors. `CONTAINER_FACTORY` implements this interface:

```java
package org.json.simple.parser;

import java.util.List;
import java.util.Map;

public interface ContainerFactory {
    public Map createObjectContainer();
    public List creatArrayContainer();
}
```

You can't predict how the interface will evolve, so don't speculate about it. The authors of simple json can add generics any time and when they do, that might clash with the signature this change introduces. I think this is a very valid place to keep rawtypes (maybe suppress them).

mbien:
Same here. Closing due to no activity in the last two years. Simply reopen and rebase if you want to continue.

PR #1693, opened 2019-11-28
[NETBEANS-3485] Fixed compiler warnings concerning rawtypes Enumerati…

mklaehn:
…on in projects of cluster config platform

There are compiler warnings about rawtype usage with a Enumeration like the following
```
   [repeat] .../platform/openide.util.ui/src/org/openide/ServiceType.java:359: warning: [rawtypes] found raw type: Enumeration
   [repeat]                 Enumeration en = r.services(clazz);
   [repeat]                 ^
   [repeat]   missing type arguments for generic class Enumeration<E>
   [repeat]     E extends Object declared in interface Enumeration
```

### DIFF
--- a/platform/core.netigso/test/unit/src/org/netbeans/core/netigso/NetigsoOSGiActivationVisibleTest.java
+++ b/platform/core.netigso/test/unit/src/org/netbeans/core/netigso/NetigsoOSGiActivationVisibleTest.java
@@ -192,7 +192,7 @@ public class NetigsoOSGiActivationVisibleTest extends SetupHid {
     }
     
     public void testResourcesFromBundle() throws Exception {
-        Enumeration res = toEnable.getResources("org/foo/Something.txt");
+        Enumeration<URL> res = toEnable.getResources("org/foo/Something.txt");
         assertEnumeration("Bundle can get its own resources", res);
     }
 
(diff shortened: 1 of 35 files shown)


platform/core.netigso/test/unit/src/org/netbeans/core/netigso/NetigsoOSGiActivationVisibleTest.java
```diff
@@ -203,19 +203,19 @@ public void testResourcesFromModule() throws Exception {
 
     public void testResourcesDirectFromBundle() throws Exception {
         Method loadResource = directBundle.getMethod("loadResources", String.class, ClassLoader.class);
-        Enumeration res = (Enumeration) loadResource.invoke(null, "org/foo/Something.txt", null);
+        Enumeration<URL> res = (Enumeration<URL>) loadResource.invoke(null, "org/foo/Something.txt", null);
```
JaroslavTulach: This results in unchecked cast warnng, right?

mklaehn: True.
The aim of this PR was too get the number of rawtypes warnings concerning Enumeration reduced. In this project this resulted in fixing the rawtypes warning by adding an unchecked cast warning.
This is also one of the reasons why I switched to fix as many (hopefully all) warnings in a single project.

PR #1697, opened 2019-11-28
[NETBEANS-3489] Fixed compiler warnings concerning rawtypes Callable

mklaehn:
There are compiler warnings about rawtype usage with a Callable like the following
```
   [repeat] .../platform/core.network/src/org/netbeans/core/network/utils/LocalAddressUtils.java:103: warning: [rawtypes] found raw type: Callable
   [repeat]     private static final Callable<InetAddress> C1 = new Callable(){
   [repeat]                                                         ^
   [repeat]   missing type arguments for generic class Callable<V>
   [repeat]   where V is a type-variable:
   [repeat]     V extends Object declared in interface Callable
```


lkishalmi:
Consider using lambda instead.

mklaehn:
Done

PR #1700, opened 2019-11-29
[TRAVIS] Remove timeouts for platform/api.htmlui tests

blackleg:


matthiasblaesing:
I also had a look at this and with increased verbosity of the TestNG tests, I saw timeouts in the test preparation methods. Maybe we should just bump the timeouts from 9s to 30s?

blackleg:
It can be a good fix, but can we remove the timeouts directly instead change these?

matthiasblaesing:
Valid question - I admit, that I'm neither to deep in TestNG nor in the htmlui. But considering that stable unittest should not run into timeouts, that would also be a valid approach (IMHO).

blackleg:
I'm going to edit this pull request to remove timeouts and test if tests are stable before merge.

blackleg:
@matthiasblaesing  I removed all timeouts and Través passes the build.

PR #1703, opened 2019-11-30
[NETBEANS-3428] FlatLaf: Fixed colors of categories buttons in Options dialog

DevCharly:
Simple fix for Options dialog categories button colors.

Uses existing colors from UI defaults.
So this should also work fine with other FlatLaf themes.

The magnifier left to the quick search field is hard to see, but this should be fixed with a SVG icon.

image

### DIFF
--- a/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
+++ b/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
@@ -124,11 +124,12 @@ public class OptionsPanel extends JPanel {
     private final boolean isMac = UIManager.getLookAndFeel ().getID ().equals ("Aqua");
     private static final boolean isNimbus = UIManager.getLookAndFeel ().getID ().equals ("Nimbus");
     private static final boolean isMetal = UIManager.getLookAndFeel() instanceof MetalLookAndFeel;
+    private static final boolean isFlatLaf = UIManager.getLookAndFeel().getID().startsWith("FlatLaf");
     private final boolean isGTK = UIManager.getLookAndFeel ().getID ().equals ("GTK");
     private final Color selected = isMac ? new Color(221, 221, 221) : getSelectionBackground();
-    private final Color selectedB = isMac ? new Color(183, 183, 183) : new Color (149, 106, 197);
+    private final Color selectedB = isMac ? new Color(183, 183, 183) : (isFlatLaf ? selected : new Color (149, 106, 197));
     private final Color highlighted = isMac ? new Color(221, 221, 221) : getHighlightBackground();
-    private final Color highlightedB = new Color (152, 180, 226);
+    private final Color highlightedB = isFlatLaf ? highlighted : new Color (152, 180, 226);
     //private final Color iconViewBorder = new Color (127, 157, 185);
     private final ControllerListener controllerListener = new ControllerListener ();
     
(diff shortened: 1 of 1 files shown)


platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
```diff
@@ -964,7 +965,7 @@ private boolean checkSize(Dimension componentSize) {
     }
 
     private static Color getTabPanelBackground() {
-        if( isMetal || isNimbus ) {
+        if( isMetal || isNimbus | isFlatLaf ) {
```
lkishalmi: isMetal || isNimbus | isFlatLaf  ->   isMetal || isNimbus || isFlatLaf at all places maybe this trio could be replaced an additional isNative() method.

DevCharly: Good idea.

neilcsmith-net: Why `|` rather than `||`?

neilcsmith-net:
Thanks for looking at this, but I'm -1 to it as is.  The reason I was going to port the PraxisLIVE hack that is also used with Darcula as an _interim_ solution is that the proliferation of UI specific code across modules like this is a problem.  This will not help the other dark LaFs either.  We should move to removing hard-coded colours throughout and use colours from the UI properties.  And ideally would remove all LaF-specific code.

platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
neilcsmith-net: And surely `isNative` is the opposite meaning?!  Incidentally, would `LookAndFeel::isNativeLookAndFeel` be a better thing to look for rather than using using LaF specific flags?

DevCharly: > Why `|` rather than `||`?

Ups, this is a typo.

DevCharly:
Using colors from UI defaults would be of course the much better solution.
I just made the minimum changes to avoid any impact on other LaFs.

platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
DevCharly: Have fixed the `|` and renamed `isNative` to `useUIDefaultsColors`

neilcsmith-net:
Surely the way to do that is to replace the `isMetal` flag with a `useUIDefaultsColors` flag equivalent to 

```java
private final boolean useUIDefaultsColors = !UIManager.getLookAndFeel().isNativeLookAndFeel();
```
?
I also have a feeling those fields should not be `static`.

lkishalmi:
The fields shall be not static as eventually we might support on the fly LaF changes.
isNativeLookAndFeel() is a kind of problematic it only returns true for Windows and probably Mac. On Linux the GtkLookAndFeel is based on Synth, so it is a kind of non-native native LaF.

I agree with Nick that there shall be no LaF specific hack in the code, especially no had coded colors. Though I would not vote against this change as this is the least intrusive to support Flatlaf based Look and Feels. On the other hand I'd create a JIRA issue to mark it to be sort out later.
We have several places where we patch the LaF all around the code base we shall mark them, and probably came up with a plan/guideline how to fix them. But that's another story.

DevCharly:
Have changed the fields and related methods to non-static.

neilcsmith-net:
OK, retriggered the failing test, and withdrawing my -1, assuming we don't forget to look at a better fix later! :smile:

I still think longer term we should move to special-casing the LaFs that *shouldn't* use values from the UIDefaults (here and elsewhere).  Which as far as I can tell is usually the native ones - and to be a pedant - https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/share/classes/com/sun/java/swing/plaf/gtk/GTKLookAndFeel.java#L1559

lkishalmi:
Created https://issues.apache.org/jira/browse/NETBEANS-3528 for the cleanup.

As of being pedant, lesson learned. Do not trust on 15 year old memory...

PR #1706, opened 2019-11-30
[NETBEANS-3501] Fixed compiler warnings concerning rawtypes Privilege…

mklaehn:
…dAction

There are compiler warnings about rawtype usage with a PrivilegedAction like the following
```
   [repeat] .../groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/compiler/ClassNodeCache.java:310: warning: [rawtypes] found raw type: PrivilegedAction
   [repeat]                 URL file = (URL) AccessController.doPrivileged(new PrivilegedAction() {
   [repeat]                                                                    ^
   [repeat]   missing type arguments for generic class PrivilegedAction<T>
   [repeat]     T extends Object declared in interface PrivilegedAction
```

### DIFF
--- a/groovy/groovy.editor/nbproject/project.properties
+++ b/groovy/groovy.editor/nbproject/project.properties
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 javac.compilerargs=-Xlint -Xlint:-serial
-javac.source=1.7
+javac.source=1.8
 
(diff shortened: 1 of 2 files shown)


lkishalmi:
Well as this is an solo occurrence, it would be probably better to write the thing with lambda instead.

mklaehn:
I've tried to keep the change small but using a lambda works for me too.

mklaehn:
Done

PR #1709, opened 2019-11-30
[TRAVIS] Compile all modules with OpenJDK11

blackleg:
Add Travis job to test if source compiles the full project with OpenJDK11.


matthiasblaesing:
Looks sane.

PR #1721, opened 2019-12-01
JEDDICT-322 Add Jeddict modules to friend-packages 

jGauravGupta:
Replacement of PR https://github.com/apache/netbeans/pull/1405 with only friend request commits.
And creating the separate PR for Public API/SPI request.


geertjanw:
Makes sense to me, this is the standard way of doing this.

neilcsmith-net:
I still think it's a bad idea to add more friend dependencies for external projects - that's just public API by the back door.  But if this is the pragmatic (and still required?) way to get the required support in for 11.3 then I guess it's the right stopgap approach?

ebarboni:
merging

PR #1722, opened 2019-12-02
[NETBEANS-3518] - fix bug in AbstractSummaryView.java

BradWalker:
I was working on my usual code cleanup when I encountered a bug in the method AbstractSummaryView() in AbstractSummaryView.java..

Notice the following code was not type-casting correctly..

```
-                if (selection.length == 1) {
-                    if (selection[0] instanceof ShowAllEventsItem) {
-                        showRemainingFiles(((ShowAllEventsItem) selection[0]).getParent(), true);
-                    } else if (selection[0] instanceof ShowLessEventsItem) {
-                        showRemainingFiles(((ShowAllEventsItem) selection[0]).getParent(), false);
-                    } else if (selection[0] instanceof MoreRevisionsItem) {
```

Specifically, an instance of ShowLessEventsItem should be properly cast..

`+                        showRemainingFiles(((ShowLessEventsItem) selection.get(0)).getParent(), false);`

While I had the hood opened, I fixed a few other Java warnings..


matthiasblaesing:
Change looks good. Thank you.

PR #1724, opened 2019-12-02
More stabilization of the vanilla javac indexer.

jlahoda:
This increases stability of indexing without nb-javac when there are errors in the source code - in particular when some classes are missing. Before, this lead to more exceptions/broken caches.


lkishalmi:
Well Travis is only failing on hibernate, so I guess this one is good to go.

PR #1737, opened 2019-12-04
[NETBEANS-3411] Added support for reading Kotlin based Gradle projects.

lkishalmi:
Changed the order of mime resolvers: <xml based>, <service registered> -> <service registered>, <xml based> 
Right now I have 93 xml based registrations and only 2 service regs: Docker and Gradle+Kotlin
I think it might be better to comb them together in a single ordered list, but as most probably you register a mime resolver as service when the xml based is not enough or mask something, I think putting the service registered ones in the front is not a bad thing. Also it might be better just to add a registered service based mime resolver for the xml based stuff and only use the service based ones with their specified ordering.


groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
```diff
@@ -174,6 +184,20 @@ static String getFileOrProjectName(FileObject primaryFile) {
         return ret;
     }
 
+    @ServiceProvider(service = MIMEResolver.class, position = -10)
+    public static class GradleKotlinMIMEResolver extends MIMEResolver {
```
JaroslavTulach: When I was working on the [performance team](http://wiki.netbeans.org/Performance) one of the goals we had was to replace all MIME resolvers with declarative ones. Introducing new programmatic ones goes against such effort.

platform/openide.filesystems/src/org/openide/filesystems/MIMESupport.java
```diff
@@ -194,8 +194,11 @@ public void resultChanged(LookupEvent evt) {
 
             ERR.fine("Computing resolvers"); // NOI18N
 
-            List<MIMEResolver> all = new ArrayList<MIMEResolver>(declarativeResolvers());
+            List<MIMEResolver> declarative = declarativeResolvers();
```
JaroslavTulach: No test failures? That is bad. I would have hoped that the order of resolvers is tested by some test.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
```diff
@@ -57,16 +59,24 @@
...
+@DataObject.Registrations({
+        @DataObject.Registration(
         mimeType = GradleDataObject.MIME_TYPE,
         iconBase = "org/netbeans/modules/gradle/resources/gradle.png",
         displayName = "#LBL_GradleFile_LOADER",
         position = 300
-)
+    ),
+    @DataObject.Registration(
+        mimeType = GradleDataObject.KOTLIN_MIME_TYPE,
```
JaroslavTulach: Possibly you could handle this without special "gradle+kotlin" MIME type. Just register your own `DataObject.Factory` into "kotlin-only" MIME type and position it before the regular `KotlinDataObject`.

lkishalmi: Thank you for the tip, I'm going to try that, so I won't have to modify the FS code.

platform/openide.filesystems/src/org/openide/filesystems/MIMESupport.java
lkishalmi: Well before Docker mime resolver, there were no service provider based resolvers, so I guess that's one of the reasons of missing tests. But I'm not a historian. Let's see if I can avoid this change.

java/kotlin.editor/src/org/netbeans/modules/kotlin/editor/layer.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jlahoda: (Assuming this is still needed:) It would probably be better to enhance the existing annotation-based declarations, rather than using an explicit layer.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
jlahoda: If this is still needed, it should be possible to tweak the default mime resolver to check the provided extensions and look at getNameExt() instead of just getExt() if the provided extension contains a dot.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
lkishalmi: Well it seems the default declarative MIME resolver is more powerful than documented at first sight. So it is possible to do the recognition without this change.
Moreover it is also possible to convert the Dockerfile MIME resolver to the declarative one. I'm going to do that in another PR along with some improvement on the documentation.

java/kotlin.editor/src/org/netbeans/modules/kotlin/editor/layer.xml
lkishalmi: I've finally find out how to do that. Thank you!

groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
lkishalmi: Could come up with a not perfect but good enough declarative mime type registration.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDataObject.java
lkishalmi: Finally could get rid of this extra declaration matching on kts extension and .gradle.kts in the filename. Not perfect, but good enough.
The custom mime-type remained as it we registering a custom "Run Gradle" action on that.

lkishalmi:
Thank you for the reviews! I think they helped make this thing round. Please check again if it is good enough to be merged!

PR #1738, opened 2019-12-04
Updating to HTML/Java API version 1.7

JaroslavTulach:
Time to update to new version of HTML/Java API which has been released in November.

OracleLab's IGV is currently using 1.6.1 (bundled together with NetBeans 11.2) and we run into a GC related problem of premature release of the model objects: [fix by this change](https://github.com/apache/netbeans-html4j/commit/cd3229252279ee02de13230909393f88cb5de18e).

The fix is part of NetBeans HTML/Java API version 1.7 - hence upgrading to it (in NetBeans as well as IGV) shall fix our problems.


nbbuild/templates/projectized.xml
```diff
@@ -835,15 +836,15 @@ If you are sure you want to build with JDK 9+ anyway, use: -Dpermit.jdk9.builds=
           <os family="windows"/>
         </condition>
         <exec executable="${maven.embedder.dir}/maven/bin/mvn${.cmd}" failonerror="true">
-            <arg value="-DforcedVersion=SNAPSHOT"/>
```
jtulach: @jglick, please take a look.

nbbuild/templates/projectized.xml
```diff
@@ -810,7 +810,8 @@ If you are sure you want to build with JDK 9+ anyway, use: -Dpermit.jdk9.builds=
         <delete file="${report.file}" failonerror="false"/>
     </target>
 
-    <target name="patch-for-maven" depends="init" description="Patch local Maven repo's -DmavenVersion of this module with one built from sources.">
+    <target name="patch-for-maven" depends="maven-install"/> <!-- deprecated -->
```
jtulach: @jglick, I was always wondering what `patch-for-maven` target stands for. After many years I finally took a look and found out that it is the task I was always looking for. Probably a sign the target's name isn't best. I propose to name it `maven-install`.

jtulach:
With the a7b0a4d and 1a414ea  changes I am able to invoke
```
netbeans$ for i in platform/net.java.html*; do ant -f $i maven-install || break; done
```
and successfully install the HTML/Java API OSGi bundles into the local Maven repository.

nbbuild/templates/projectized.xml
jglick: So just replacing `SNAPSHOT` with `dev-SNAPSHOT`? Seems fine.

PR #1740, opened 2019-12-04
[NETBEANS-3542] - cleanup unchecked conversion warnings..

BradWalker:
There are a LOT of the following warnings in the build..

   [repeat] /home/bwalker/src/netbeans/platform/openide.filesystems/src/org/netbeans/modules/openide/filesystems/declmime/MIMEResolverProcessor.java:190: warning: [unchecked] unchecked conversion
   [repeat]         Set<String> s = new TreeSet();
   [repeat]                         ^
   [repeat]   required: Set<String>
   [repeat]   found:    TreeSet

This change cleans up many of these.. But, it's not complete as there are so many of them..


matthiasblaesing:
Looks sane.

BradWalker:
Thanks very much for your help with this..

PR #1741, opened 2019-12-04
[NETBEANS-3543] Corrected compiler warnings in extide/o.apache.tools.…

mklaehn:
…ant.module project

Fixed all compiler warnings concerning linter warnings types of

* deprecation,
* overrides,
* rawtypes and
* unchecked.

### DIFF
--- a/extide/o.apache.tools.ant.module/nbproject/project.properties
+++ b/extide/o.apache.tools.ant.module/nbproject/project.properties
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-javac.compilerargs=-Xlint:unchecked
+javac.compilerargs=-Xlint -Xlint:-serial
 javac.source=1.8
 spec.version.base=3.91.0
 compile.ant.jar=${ant.core.lib}
--- a/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/BridgeImpl.java
+++ b/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/BridgeImpl.java
@@ -31,7 +31,6 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.net.URL;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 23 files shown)


JaroslavTulach:
Hello Martin, thanks for the cleanup.

If you removed all compilation warnings from the module, should you also make sure that once a new warning is introduced, the compilation fails?

Because without such automatic check your work may not last. Traditionally the NetBeans community never paid much attention to warnings - as such I wouldn't even noticed if I re-introduced a warning.

There is a `-Werror` javac flag. You should enable it when setting the `javac.compilerargs` property. Or (better) change `common.xml` to support a new property `javac.failOnWarnings=false` (`false` by default) that each cleaned up module could set to `true`.

mklaehn:
> Hello Martin, thanks for the cleanup.
> 
> If you removed all compilation warnings from the module, should you also make sure that once a new warning is introduced, the compilation fails?
> 
> Because without such automatic check your work may not last. Traditionally the NetBeans community never paid much attention to warnings - as such I wouldn't even noticed if I re-introduced a warning.

I fully agree with you.

However, the only thing I've accomplished is correcting the linter warnings for specific types - namely the one listed in the description of this PR. With the current value of the `javac.compilerargs` property no more linter warnings are produced. This however does not mean that the compiler does not produce no warnings. The following warnings are an exerpt of the warning still produced during the build process:

```
.../nbbuild/build/public-package-jars/org-netbeans-api-annotations-common.jar(/org/netbeans/api/annotations/common/NonNull.class): warning: Cannot find annotation method 'when()' in type 'Nonnull': class file for javax.annotation.Nonnull not found
warning: unknown enum constant When.ALWAYS
  reason: class file for javax.annotation.meta.When not found
.../nbbuild/build/public-package-jars/org-netbeans-api-annotations-common.jar(/org/netbeans/api/annotations/common/NullAllowed.class): warning: Cannot find annotation method 'when()' in type 'Nonnull'
warning: unknown enum constant When.MAYBE
warning: No processor claimed any of these annotations: org.netbeans.api.annotations.common.NullAllowed,org.netbeans.api.annotations.common.NonNull
```

Setting the `-Werror` javac flag then causes the build to fail rather fast.

```
error: warnings found and -Werror specified
1 error
5 warnings
```

I have not investigated what causes these messages nor how to fix them. That would propably be a job for someone that is more knowledgeable with the NetBeans buildsystem.

> There is a `-Werror` javac flag. You should enable it when setting the `javac.compilerargs` property. Or (better) change `common.xml` to support a new property `javac.failOnWarnings=false` (`false` by default) that each cleaned up module could set to `true`.

My - at this stage rather pragmatic - opinion would be to add the `-Werror` to the `javac.compilerargs` property. Anyone acustomed to javac and its commandline argument knows what that does.
Admittably a new property `javac.failOnWarnings=false` has its charm as well but who knows how it is used in the build.

Just a gut feelling...

matthiasblaesing:
There are unresolved comments older than a year. Closing.

PR #1744, opened 2019-12-05
[NETBEANS-3550] - cleanup warnings related to deprecated JUnit Asserts

BradWalker:
There are quite a few errors msgs that are easy to fix..

    [repeat] /home/bwalker/src/netbeans/harness/nbjunit/src/org/netbeans/junit/NbModuleSuite.java:56: warning: [deprecation] Assert in junit.framework has been deprecated
   [repeat] import junit.framework.Assert;
   [repeat]                       ^

The simple fix is to use the org.junit.Assert class..

-import junit.framework.Assert;
+import org.junit.Assert;

Everything stays the same..


BradWalker:
Not as easy to fix as expected.. Will need some test harness changes so I'm closing this..

PR #1746, opened 2019-12-05
[NETBEANS-3552] - Upgrade GraalVM from 19.0.0 to 19.3.0

pepness:
Notes:
- Added support for date, time, timezone and duration values in polyglot
- Better sub-process control
- Improved FileSystem management
- Updated Node.js to version 12.10.0
- Added support for reading environment variables
- Improved debugger
- Added support for temporary files and directories

Notes for ICU4J:
- Java 7 as minimum runtime
- Support Unicode 12.1
- Support CLDR 36 (Unicode Common Locale Data Repository)

[GraalVM Web Page](https://www.graalvm.org/)
[GraalVM SDK Release Notes](https://github.com/oracle/graal/blob/master/sdk/CHANGELOG.md)
[GraalJS Release Notes](https://github.com/graalvm/graaljs/blob/master/CHANGELOG.md)
[GraalVM Truffle Release Notes](https://github.com/oracle/graal/blob/master/truffle/CHANGELOG.md)
[ICU4J Release Notes](https://github.com/unicode-org/icu/releases/)

PR #1748, opened 2019-12-05
[TRAVIS] Refactor GRAALVM tests

blackleg:
Refactor GRAALVM tests

### DIFF
--- a/nbbuild/travis/scripting.sh
+++ b/nbbuild/travis/scripting.sh
@@ -28,19 +28,10 @@ if [ -z "$GRAALVM" ]; then
   GRAALVM=`pwd`/$BASE
 fi
 
-# test on regular VM
-
-ant -f platform/api.scripting/build.xml test
-ant -f ide/libs.graalsdk/build.xml test
-ant -f webcommon/libs.graaljs/build.xml test
-ant -f platform/core.network/build.xml test
-ant -f profiler/profiler.oql/build.xml test
-ant -f platform/api.htmlui/build.xml test
-
 $GRAALVM/bin/gu install python
 $GRAALVM/bin/gu install R
 
-# test on GraalVM
+# Test on GraalVM
 
 JAVA_HOME=$GRAALVM ant -f platform/api.scripting/build.xml test
 JAVA_HOME=$GRAALVM ant -f ide/libs.graalsdk/build.xml test
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,9 +54,6 @@ matrix:
         - env: SIGTEST=true
           jdk: openjdk8
 
-#        - env: TARGET="build-platform" SCRIPT=nbbuild/travis/scripting.sh
-#          jdk: openjdk8
-
         - env: TARGET="build-platform" SCRIPT=nbbuild/travis/gensigtest.sh
           jdk: openjdk8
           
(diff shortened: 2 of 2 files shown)


JaroslavTulach:
Work in #1746 is also touching the same `scripting.sh` file.

blackleg:
@JaroslavTulach This pull request only changes the Travis config and disables tests for `platform/core.network` module that fails when runs in Travis VM.

nbbuild/travis/scripting.sh
```diff
@@ -28,24 +28,15 @@ if [ -z "$GRAALVM" ]; then
   GRAALVM=`pwd`/$BASE
 fi
 
-# test on regular VM
-
-ant -f platform/api.scripting/build.xml test
```
JaroslavTulach: Should this really be removed?

blackleg: Tests on regular VM are executed in other Travis jobs like this: https://github.com/apache/netbeans/blob/master/.travis.yml#L362

blackleg:
> Are the `libs.graalsdk`, `libs.graaljs`, `profiler.oql` & co. tests executed anywhere else?

Yes, they are executed in other Travis jobs.
- [platform/api.htmlui](https://github.com/apache/netbeans/blob/master/.travis.yml#L105)
- [platform/api.scripting](https://github.com/apache/netbeans/blob/master/.travis.yml#L110)
- [ide/libs.graalsdk](https://github.com/apache/netbeans/blob/master/.travis.yml#L251)
- [webcommon/libs.graaljs](https://github.com/apache/netbeans/blob/master/.travis.yml#L362)
- [profiler.oql](https://github.com/apache/netbeans/blob/master/.travis.yml#L348)

The tests for [platform/core.network](https://github.com/apache/netbeans/blob/master/.travis.yml#L123) doesn't work when they are executed runs in a Travis job and are disabled in config.

nbbuild/travis/scripting.sh
```diff
@@ -28,24 +28,15 @@ if [ -z "$GRAALVM" ]; then
...
 $GRAALVM/bin/gu install R
 
-# test on GraalVM
+# Test on GraalVM
 
 JAVA_HOME=$GRAALVM ant -f platform/api.scripting/build.xml test
 JAVA_HOME=$GRAALVM ant -f ide/libs.graalsdk/build.xml test
 
-JAVA_HOME=$GRAALVM ant -f platform/core.network/build.xml test
+#JAVA_HOME=$GRAALVM ant -f platform/core.network/build.xml test
```
JaroslavTulach: I cannot integrate any change that doesn't work for `core.network`. There is a code that deals with proxy setup and it has to work on all supported JVMs.

blackleg: There is a problem with core.network module tests and Travis environment: https://github.com/apache/netbeans/pull/1773 and https://github.com/apache/netbeans/pull/1763.
 You could accept and merge this pull request to allow Travis run tests with GRAALVM for the rest of modules.

blackleg: Problems with `core.network` module tests and Travis env were fixed in this pull request: https://github.com/apache/netbeans/pull/1835

matthiasblaesing:
Thank you.

matthiasblaesing:
@backleg please don't use the github features, if your user account is not configured properly (properly means with a public visible email address and your real name). In this case the master merge has broken author information. Considering it is only a merge I don't consider it critical, but the github GUI is a mess regarding sane author information.

jtulach:
Thanks a lot @blackleg for fixing the NetBeans+GraalVM testing!

PR #1750, opened 2019-12-06
Updating BugZilla URLs from old Oracle infra to the new ASF one

jkovalsky:
Now that we have a static snapshot of former NetBeans BugZilla hosted on Apache infrastructure it looks like a good idea to update links everywhere. For example in JavaDoc templates.


neilcsmith-net:
What about other references to that link throughout the code base.  Might be good as well / instead to get a redirect setup for that URL?

PR #1754, opened 2019-12-06
Redefine jellytools stable tests

blackleg:
Redefine jellytools stable tests to enable in Travis config


matthiasblaesing:
Looks sane to me.

PR #1756, opened 2019-12-07
[NETBEANS-3558] Spring Framework 5.2.2

blackleg:
Update java/libs.springframework module to add Spring Framework 5.2.2 library dependency for ant projects. [NETBEANS-3558](https://issues.apache.org/jira/browse/NETBEANS-3558)

Screenshot from 2019-12-22 13-24-24


java/libs.springframework/external/spring-aop-5.2.2.RELEASE-license.txt
```diff
@@ -0,0 +1,282 @@
+Name: Spring Framework
+Description: Java/J2EE application framework
+Version: 5.2.2.RELEASE
+Origin: Spring
+License: Apache-2.0-spring5.2.2
+URL: https://spring.io/projects/spring-framework
```
matthiasblaesing: This is a group of dependencies with the same license + notice (my assumption). If so, please use the "Files" header to reference all files the license+notice apply to and remove the additional license + notice files. That way it is better readable.

blackleg: Changed

PR #1757, opened 2019-12-07
[NETBEANS-3559] Corrected compiler warnings in Bootstrap project.

mklaehn:
Fixed all compiler warnings concerning linter warnings types of

* deprecation,
* rawtypes and
* unchecked.

As all compiler warning in this project could be fixed / suppressed when appropriate the flag `-Werror` was added to build property `javac.compilerargs`.

### DIFF
--- a/platform/o.n.bootstrap/src/org/netbeans/ModuleInstaller.java
+++ b/platform/o.n.bootstrap/src/org/netbeans/ModuleInstaller.java
@@ -191,7 +191,7 @@ public abstract class ModuleInstaller {
      * @since org.netbeans.core/1 > 1.6
      * @see "#27853"
      */
-    public void refineClassLoader(Module m, List parents) {
+    public void refineClassLoader(Module m, List<? extends ClassLoader> parents) {
         // do nothing
     }
 
--- a/platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
+++ b/platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
@@ -299,7 +299,7 @@ public final class ModuleManager extends Modules {
         return completeLookup;
     }
     // Access from ChangeFirer:
-    final void fireModulesCreatedDeleted(Set created, Set deleted) {
+    final void fireModulesCreatedDeleted(Set<Module> created, Set<Module> deleted) {
         if (Util.err.isLoggable(Level.FINE)) {
             Util.err.fine("lookup created: " + created + " deleted: " + deleted);
         }
@@ -311,7 +311,7 @@ public final class ModuleManager extends Modules {
      * @see #PROP_MODULES
      */
     public Set<Module> getModules() {
-        return new HashSet<Module>(modules);
+        return new HashSet<>(modules);
     }
 
     final int getModuleCount() {
@@ -731,7 +731,7 @@ public final class ModuleManager extends Modules {
         }
 
         @Override
-        protected synchronized Class loadClass(String name, boolean resolve) throws ClassNotFoundException {
+        protected synchronized Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
             ProxyClassLoader priviledged = null;
             NetigsoLoader osgi = null;
             if (!name.startsWith("java.")) { // NOI18N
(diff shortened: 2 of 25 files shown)


platform/o.n.bootstrap/src/org/netbeans/TopSecurityManager.java
```diff
@@ -165,6 +165,7 @@ private void notifyWrite(String file) {
     }
     
     static boolean officialExit = false;
+    @SuppressWarnings("rawtypes")
```
matthiasblaesing: Why not switch signature to `static Class<?>[] getStack()`? This is package private, so no API concern there.

platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
```diff
@@ -896,7 +896,7 @@ void refineDependencies(Module m, Set<Dependency> dependencies) {
         return installer.refineProvides (m);
     }
     /** Used by Module to communicate with the ModuleInstaller re. classloader. */
-    public ClassLoader refineClassLoader(Module m, List parents) {
+    public ClassLoader refineClassLoader(Module m, List<ClassLoader> parents) {
```
matthiasblaesing: Would `public ClassLoader refineClassLoader(Module m, List<? extends ClassLoader> parents)` make more sense? I have really no idea what this does (it is not used in the NB codebase) so it is difficult to judge.

platform/o.n.bootstrap/src/org/netbeans/ModuleInstaller.java
```diff
@@ -191,7 +191,7 @@ public Manifest loadManifest(File jar) throws IOException {
      * @since org.netbeans.core/1 > 1.6
      * @see "#27853"
      */
-    public void refineClassLoader(Module m, List parents) {
+    public void refineClassLoader(Module m, List<ClassLoader> parents) {
```
matthiasblaesing: See comment in `ModuleManager`

platform/o.n.bootstrap/src/org/netbeans/TopSecurityManager.java
mklaehn: Because SecurityManager.getClassContext() (https://github.com/apache/netbeans/pull/1757/files/8a18d35d32b2fcf2c88123f6d63f6428b176e911#diff-d7c955f1bda54e7555b9b9f280635a50L176) returns Class[] only.
However while thinking about my response I have to admit you're right.

platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
mklaehn: From what I understand it enabled a ModuleInstaller to "refine" the classpath of a module. Default Implementation is a no-op.
Searching the codebase however does not find any usage of this method. However the code give a bugzille reference for https://bz.apache.org/netbeans/show_bug.cgi?id=27853.

I foresee no problem changing the type of the list to `List<? extends ClassLoader>`

platform/o.n.bootstrap/src/org/netbeans/ModuleInstaller.java
mklaehn: See above

platform/o.n.bootstrap/src/org/netbeans/TopSecurityManager.java
mklaehn: Because SecurityManager.getClassContext() (https://github.com/apache/netbeans/pull/1757/files/8a18d35d32b2fcf2c88123f6d63f6428b176e911#diff-d7c955f1bda54e7555b9b9f280635a50L176) returns Class[] only.
However while thinking about my response I have to admit you're right and will change the signature.

platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
mklaehn: Done

platform/o.n.bootstrap/src/org/netbeans/ModuleInstaller.java
mklaehn: Done

platform/o.n.bootstrap/src/org/netbeans/TopSecurityManager.java
mklaehn: Done

PR #1760, opened 2019-12-08
[NETBEANS-3428] Update FlatLaf from 0.20 to 0.21

DevCharly:
This PR updates FlatLaf to 0.21, which includes following fixes/improvements for NetBeans:

- Fixed "text" color in Dark theme. This fixes the text color in NbExplorer and NbList (was black, now very light gray).

Here is a full list of changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/0.21


DevCharly:
BTW is there a reason that the FlatLaf jar is renamed from `flatlaf-0.21.jar` to `com-formdev-flatlaf.jar` in `nbbuild/netbeans/platform/modules/ext`, while all other jars in `ext` keep their original names?

I would prefer to keep the original name (including version number).

lkishalmi:
Well the name change is not necessary. It happened  due to one my early attempt so get the jar included into the module system. It is usually a good practice to remove the version from the jar, so when you update the module it gets overwritten and does not accumulate up in the user dir. However with built-in modules it is not that important either (updates are usually delivered with new releases).
So do what you wish for renaming.

DevCharly:
Ok, thanks, I'll rename it.

Another question: is there any documentation on howto update external libraries?
I managed it somehow, but maybe there is a better way.

What I did:
1. open `binaries-list` in editor and change version
2. build project, which tells me the hash is wrong and shows the correct hash
3. copy correct hash to `binaries-list`
4. rename license file
5. open `nbproject/project.properties`, `nbproject/project.xml` and license file in editor and change versions
6. build

Is this the right way? Thanks.

lkishalmi:
4.5 Change the versions in the license file (you did that as well)

But pretty much that covers the procedure. When there is some change in the exposed API-s it is wise to increase the module version number in the manifest.mf as well.

PR #1767, opened 2019-12-10
[NETBEANS-3260] : Corrected output for modular project while building…

singh-akhilesh:
… using ant script

Bug link : https://issues.apache.org/jira/browse/NETBEANS-3260


Mohan-Sarilla:
👍 I have tested the fix , Its working fine.

matthiasblaesing:
@singh-akhilesh please adjust your author information. The realname part should read as your real name it currently looks some kind of "username" (singh-akilesh instead of "Akhilesh Singh"). If your current git client does not show it, git CLI can be used, run `git log` to get detail info for the current branch.
@arusinha the author is not a committer, so if you are satisfied with the change feel free to merge.

PR #1769, opened 2019-12-10
Flatlaf fontscolors

lkishalmi:
Add Darcula like font coloring to FlatLaf Dark Look and Feel.
Thanks for the hard work @AlexFalappa !


lkishalmi:
image

This is how DarkLaf looks today.

bondolo:
Is there a daytime view? Dark themes are fine for night-time but I wouldn't want to use one all day.

DevCharly:
Light theme is coming soon.
I'm working on the editor/view tabs for FlatLaf, which also adds the Light theme.

image

And in dark:
image

Should be ready tomorrow...

lkishalmi:
Starting to look really cool.

I'm just wandering for the toolbar icons to have a monochrome filter... 
(and the spacing a bit cramped) but that's for another day...

AlexFalappa:
@DevCharly gorgeous!

eirikbakke:
@lkishalmi Monochrome icons might be confused with disabled icons. Another idea is to invert the brightness of each pixel while keeping the hue (e.g. in LAB color model). (As you maybe already know, ImageUtilities has support for these kinds of image filters.)

lkishalmi:
Oh, I've expressed myself incorrectly. I'd not just go monochrome, but 
black and white, and gray for disabled. (mor properly a foregroudn 
color, a background and a disabled one)

  Something like Sifr icons in LibreOffice. This is just an idea, I 
won't have time to test that, it is just the toolbar on flatlaf with big 
icons and that spacing is visually disturbing.

eirikbakke:
@lkishalmi Have you tried the "Small Toolbar Icons" option? (Available by right-clicking the main NetBeans toolbar.) What do you think about enabling this option by default, in all LAFs?

Enabling "Small Toolbar Icons" by default will have one advantages in the future, when we slowly start replacing bitmap toolbar icons with SVG versions. Currently, the toolbar requires two icon versions of each action--both 16x16 and 24x24. This is a hassle--in the future I imagine there will be just a single SVG image for each action which can be scaled to either size. It would be best if this could be a 16x16 version, which is then automatically scaled up to 24x24 if "Small Toolbar Icons" is disabled. Since this will scale up border widths as well, though, this means toolbar icons will look slightly more consistent in "Small Toolbar Icons" mode.

PR #1771, opened 2019-12-11
[NETBEANS-3428] FlatLaf: UI delegates for editor and view tabs in main window

DevCharly:
Editor and view tabs in main window implemented in flat style.

image

Also added "FlatLaf Light" to LaF combobox in Options.

image

The tabs and split dividers scale also on Java 8 (and on Linux) in the same was as FlatLaf scales.

Note: When using non-default document tab options (e.g. left/right/bottom tab placement, multi-row tabs, etc), the editor tabs are shown differently because a different renderer is used, which is not implemented in this PR. These options can be found in Options dialog at "Appearance > Document Tabs".

All tab colors can be configured in [FlatDarkLaf.properties](https://github.com/DevCharly/netbeans/blob/flatlaf-tabs/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatDarkLaf.properties#L24-L51) and [FlatLightLaf.properties](https://github.com/DevCharly/netbeans/blob/flatlaf-tabs/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLightLaf.properties#L18-L47).
Colors, insets, etc, which are common to Dark and Light themes are in [FlatLaf.properties](https://github.com/DevCharly/netbeans/blob/flatlaf-tabs/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLaf.properties#L25-L53).

These three `.properties` files are detected by FlatLaf using a service loader.

Values starting with `@@` (in the `.properties` files) are references to other values. They may come from FlatLaf core `.properties` files: https://github.com/JFormDesigner/FlatLaf/tree/master/flatlaf-core/src/main/resources/com/formdev/flatlaf


geertjanw:
So beautiful!

DevCharly:
Thx @geertjanw 😄

lkishalmi:
Build fails on sigtest public API signature test. One possible solution is to increase the Major Release version.
image

It needs to be added in the FlatLaf Look and Feel Project metadata as well:
image

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
```diff
@@ -0,0 +1,50 @@
...
+public class FlatLFCustoms extends LFCustoms {
+
+    @Override
+    public Object[] createApplicationSpecificKeysAndValues() {
+        return new Object[] {
+            EDITOR_TAB_DISPLAYER_UI, "org.netbeans.swing.laf.flatlaf.ui.FlatEditorTabDisplayerUI",
+            VIEW_TAB_DISPLAYER_UI, "org.netbeans.swing.laf.flatlaf.ui.FlatViewTabDisplayerUI",
+
+            EDITOR_TAB_CONTENT_BORDER, DPISafeBorder.matte(0, 1, 1, 1, UIManager.getColor("TabbedContainer.editor.contentBorderColor")),
+            VIEW_TAB_CONTENT_BORDER, DPISafeBorder.matte(0, 1, 1, 1, UIManager.getColor("TabbedContainer.view.contentBorderColor")),
```
junichi11: Should add `// NOI18N`s

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/HiDPIUtils.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Calculates the width of a border in device pixels.
+     * The difference to com.formdev.flatlaf.util.UIScale.scale() is that this method
+     * always rounds down the result, which gives nice small 1px borders at 150% and 175%.
+     * UIScale.scale() rounds up and would return 2px at 150% and 175%.
+     */
+    public static int deviceBorderWidth(double scale, int logical) {
+        if (logical <= 0)
+            return 0;
```
junichi11: Nitpick: Should use curly braces(`{}`)

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatEditorTabCellRenderer.java
```diff
@@ -0,0 +1,244 @@
...
+
+    public FlatEditorTabCellRenderer() {
+        super(painter, new Dimension(tabInsets.left + tabInsets.right, tabInsets.top + tabInsets.bottom));
+    }
+
+    @Override
+    public Dimension getPadding() {
+        Dimension d = super.getPadding();
+        if (isShowCloseButton() && !Boolean.getBoolean("nb.tabs.suppressCloseButton"))
+            d.width += findCloseIcon().getIconWidth() + UIScale.scale(CLOSE_ICON_RIGHT_PAD);
```
junichi11: Should use curly braces({}) and add `// NOI18N`

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,270 @@
...
+        buttonIDs.put("scroll_left", TabControlButton.ID_SCROLL_LEFT_BUTTON);
+        buttonIDs.put("scroll_right", TabControlButton.ID_SCROLL_RIGHT_BUTTON);
+        buttonIDs.put("drop_down", TabControlButton.ID_DROP_DOWN_BUTTON);
+        buttonIDs.put("maximize", TabControlButton.ID_MAXIMIZE_BUTTON);
+        buttonIDs.put("restore", TabControlButton.ID_RESTORE_BUTTON);
+        Map<String, Integer> buttonStates = new LinkedHashMap<String, Integer>();
+        buttonStates.put("default", TabControlButton.STATE_DEFAULT);
+        buttonStates.put("pressed", TabControlButton.STATE_PRESSED);
+        buttonStates.put("disabled", TabControlButton.STATE_DISABLED);
+        buttonStates.put("rollover", TabControlButton.STATE_ROLLOVER);
```
junichi11: Should add `// NOI18N`s

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,270 @@
...
+        Color fgColor = foreground;
+        {
+            Color closeColor = (buttonId == TabControlButton.ID_CLOSE_BUTTON)
+                    ? closeRolloverBackground : null;
+            if (buttonState == TabControlButton.STATE_DISABLED) {
+                fgColor = disabledForeground;
+            } else if (buttonState == TabControlButton.STATE_PRESSED) {
+                bgColor = closeColor != null ? closeColor : pressedBackground;
+                if (closeColor != null && closeRolloverForeground != null)
+                    fgColor = closeRolloverForeground;
```
junichi11: Should use curly braces(`{}`)

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,270 @@
...
+            } else if (buttonState == TabControlButton.STATE_ROLLOVER) {
+                bgColor = closeColor != null ? closeColor : rolloverBackground;
+                if (closeColor != null && closeRolloverForeground != null)
+                    fgColor = closeRolloverForeground;
```
junichi11: Should use curly braces(`{}`)

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
matthiasblaesing: Just curious - this comment clutters the codebase - shouldn't it only be used at places where translation is expected?

DevCharly:
@lkishalmi thanks for the info and for the solution.

While I understand what sigtest is good for, I actually don't understand that it fails in this case.
I've just exported more packages, which adds API, but does not change or remove API.

> One possible solution is to increase the Major Release version.

Just curious, what other solutions are possible?

matthiasblaesing:
This is the signature check result (the raw lock is better analysable than the "nice" GUI one:

https://api.travis-ci.org/v3/job/623688897/log.txt

```
Class path: "/home/travis/build/apache/netbeans/platform/libs.flatlaf/external/flatlaf-0.21.jar:/home/travis/build/apache/netbeans/nbbuild/netbeans/platform/modules/ext/flatlaf-0.21.jar:/home/travis/build/apache/netbeans/nbbuild/netbeans/platform/modules/org-netbeans-libs-flatlaf.jar"
Constant checking: on
Found in total: 140 classes
RETURN TYPE FOR FlatFileChooserUI::createDirectoryComboBoxRenderer CHANGED FROM javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxRenderer TO javax.swing.DefaultListCellRenderer
Hidden return type (can't convert) : com.formdev.flatlaf.ui.FlatScrollPaneUI$Handler in method public com.formdev.flatlaf.ui.FlatScrollPaneUI$Handler com.formdev.flatlaf.ui.FlatScrollPaneUI.getHandler()
Hidden return type (can't convert) : com.formdev.flatlaf.ui.FlatSpinnerUI$Handler in method public com.formdev.flatlaf.ui.FlatSpinnerUI$Handler com.formdev.flatlaf.ui.FlatSpinnerUI.getHandler()
Selected by -Package: 133 classes
Written to sigfile: 196 classes(and 8 inner classes)
```

So it was not the added API, but a changed one.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
junichi11: I think that we should add it if translation is not expected, but I could be wrong.

matthiasblaesing: The interesting part would be: why should we add this throughout the codebase. From my perspective it is distracting, causes friction and has a low return factor. I can see the use case when a class is translated, then it helps to keep untranslateable strings. But why do the work before it is clear, that translation is expected?

junichi11: I'm not sure... Maybe, oracle's people(@jlahoda @sdedic ) know something?

svenreimers: I think if you use the Internationalization tooling in NetBeans across multiple files it will skip strings found on line with //NOI18N. So it may indeed be useful.

junichi11: @svenreimers I think so, too. Thank you.

lkishalmi:
Other solutions to the sig-check is probably make the package to be exposed to friends only. I'd avoid that. @matthiasblaesing  and/or @JaroslavTulach may have other options as well.

DevCharly:
@matthiasblaesing well, sigtest does not tell the truth in this case because nothing has changed at the three reported methods. But they have in common that the methods are `public` or `protected` and return a type that is (package) `private`. And they are all in package `com.formdev.flatlaf.ui`, which was not exported before.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/DPISafeBorder.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.swing.border.Border;
+import javax.swing.border.MatteBorder;
+
+/**
+ * A border similar to {@link MatteBorder}, but which avoids visual artifacts from rounding errors
+ * under non-integral HiDPI scaling factors (e.g. 150%).
+ *
+ * @author Eirik Bakke (ebakke@ultorg.com)
+ */
+final class DPISafeBorder implements Border {
```
eirikbakke: At some point we might expose this as a common utility class, but I think it's fine to make a copy of it for now.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/HiDPIUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Paint at scale factor 1x to avoid rounding issues at 125%, 150% and 175% scaling.
+     * <p>
+     * Scales the given Graphics2D down to 100% and invokes the
+     * given painter passing scaled deviceWidth and deviceHeight.
+     * <p>
+     * Transformation to device pixels borrowed from class
+     * {@link org.netbeans.swing.plaf.windows8.DPISafeBorder}.
+     */
+    public static void paintAtScale1x(Graphics g0, int x, int y, int width, int height, HiDPIPainter painter) {
```
eirikbakke: Again, it might be nice to factor out this code eventually, but we can wait until we have seen a few different uses of it.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,270 @@
...
+ * Javadoc).
+ *
+ * This is a copy of Windows8VectorTabControlIcon with modifications:
+ * - use "close" icon from AquaVectorTabControlIcon
+ * - arrow icons changed to match arrows in FlatLaf (chevron or triangle)
+ * - minimize, maximize and restore icons changed
+ * - increased size from 14x14 to 16x16
+ * - scales in Java 8 and Linux
+ *
+ * @author Eirik Bakke
```
eirikbakke: Feel free to drop the author tag here, or add your name...

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/Startup.java
```diff
@@ -421,29 +421,39 @@ private LFCustoms findDefaultCustoms() {
...
+            if (UIUtils.isXPLF()) {
+                if (isWindows8() || isWindows10()) {
+                    buf.append("Windows8LFCustoms"); //NOI18N
+                } else if (isWindowsVista() || isWindows7()) {
+                    buf.append("VistaLFCustoms"); //NOI18N
+                } else {
+                    buf.append("XPLFCustoms"); //NOI18N
+                }
+            } else {
+                buf.append("LFCustoms"); //NOI18N
```
eirikbakke: With the new changes, didn't we already check the "Nb.LFCustoms" key?

What are the changes here supposed to do?

DevCharly: No, "Nb.LFCustoms" key was not checked before.
It was (and still is) "Nb.{look-and-feel-id}LFCustoms".

Now "Nb.LFCustoms" is checked first, but it is read from the "Look and feel defaults", which are created by `LookAndFeel.getDefaults()`. Each LaF has its own independent "Look and feel defaults". Other existing LaFs do not set "Nb.LFCustoms", so the LAF ID is used as before.

`UIManager` uses two independent sets of UI defaults: "developer defaults" and "Look and feel defaults". `UIManager.put()` writes into the "developer defaults". `UIManager.get()` reads first from the "developer defaults" and if the value does not exist there, it reads from the "Look and feel defaults". For a more detailed description see section "Defaults" in the [UIManager Javadoc](https://docs.oracle.com/javase/8/docs/api/javax/swing/UIManager.html).

The problem with FlatLaf is that there are multiple themes, which all have different IDs. Well, currently only two themes are used, but maybe more in the future.

An I think it is possible to put one or more of the JSON files from [here](https://github.com/JFormDesigner/FlatLaf/tree/master/flatlaf-demo/src/main/resources/com/formdev/flatlaf/demo/intellijthemes) into a new NetBeans plugin, invoke `UIManager.installLookAndFeel()` from the plugin installer, and ready is a NetBeans theme plugin that can be uploaded to plugin portal.
In this case it is good to have `FlatLFCustoms` installed, independent from the LAF ID.

> What are the changes here supposed to do?

New are only following lines: https://github.com/apache/netbeans/blob/d11c6bf5d20dcd79f0bb4010061391eb50004c4c/platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/Startup.java#L426-L434

The other differences are removed ending whitespace and indentation:

image

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/HiDPIUtils.java
DevCharly: 👍  

I've just avoided to add new API to other modules...

DevCharly:
Hmm, I've increased the Major Release version on the FlatLaf library, but sigtest still fails with the same errors as before...

~~~
RETURN TYPE FOR FlatFileChooserUI::createDirectoryComboBoxRenderer
   CHANGED FROM javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxRenderer
   TO javax.swing.DefaultListCellRenderer

Hidden return type (can't convert) : com.formdev.flatlaf.ui.FlatScrollPaneUI$Handler
   in method public com.formdev.flatlaf.ui.FlatScrollPaneUI$Handler
   com.formdev.flatlaf.ui.FlatScrollPaneUI.getHandler()

Hidden return type (can't convert) : com.formdev.flatlaf.ui.FlatSpinnerUI$Handler
   in method public com.formdev.flatlaf.ui.FlatSpinnerUI$Handler
   com.formdev.flatlaf.ui.FlatSpinnerUI.getHandler()
~~~

I can fix the two issues with `getHandler()` in a new FlatLaf release, because these methods should be private.

The first error is tricky, because `FlatFileChooserUI` does not have a method `createDirectoryComboBoxRenderer`. This method is in the super class `MetalFileChooserUI` and returns private type `DirectoryComboBoxRenderer` (which extends `DefaultListCellRenderer`).

Seems that sigtest has problems with public or protected methods that return (package) private types.

Is this a known issue?
@jtulach @emilianbold

JaroslavTulach:
> Seems that sigtest has problems with public or protected methods that return (package) private types.

It is not sigtest that has a problem. The library exposing such methods has a problem!

Why should an API expose a method that returns package private type? What the callers are going to do with it? How will the Javadoc look like? Having such methods just doesn't make any sense when designing Java APIs.

Of course, the problem is when a 3rd party library which hasn't really paid much attention to its API is then re-exported by NetBeans module and becomes subject to signature testing. There are three possible solutions I can think of:
- disable the signature testing
- fix the library
- improve [sigtest](https://github.com/jtulach/netbeans-apitest) to have an option that would allow the above practice

As an author of [Practical API Design](http://apidesign.org) I am in favor of fixing the library, but I understand that it may be more practical to choose one of the other solutions.

DevCharly:
@JaroslavTulach
Full agree that that it is a problem in the library to expose such methods and I will fix the two `getHandler()` methods in my library, which are public through an oversight.

But the third problem is in Java 8 class `MetalFileChooserUI`, which I can not fix.
My library simply contains a public class `FlatFileChooserUI` that extends `MetalFileChooserUI`.
`FlatFileChooserUI` must be public because it is loaded by Swing via reflection.

The issue in `MetalFileChooserUI` has been fixed in Java 9: [JDK-8167176](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8167176)
https://github.com/openjdk/jdk/commit/bc442826a9d2ee64ef237c89696b616a0ea7dea0#diff-51ee8b27454e4164799775a61250edee

So, I think I'll rework this PR in a way that it is not necessary to export the package `com.formdev.flatlaf.ui`, which contains the class that extends `MetalFileChooserUI`.

DevCharly:
@eirikbakke 
> 1. In the bottom right-hand corner of the editor in the screenshots, there is one more "X" icon that has not been updated.

Yes, this "X" is part of the bread crumb component, which is not part of this PR. I'd like to fix this (and more) in another PR.

E.g. there should be a separator between the editor toolbar and the editor pane. Also "Action Items" sliding button at the right side needs an update. Color of separator between editor and bread crumb is wrong in dark theme.

DevCharly:
@eirikbakke 
> 2. The "minimize" icon ("_" symbol) in the "Projects" and "Navigator" areas could be made slightly thinner, to match the revised "X" icons.

Yes, I see what you mean.
The line is currently 2px at 100%, 3px at 150% and 4px at 200%

Initially I tried 1px at 100%, 2px at 150% and 2px at 200%, which is the same line width as in the maximize and restore icons, but thought that then the minimize button is hard to discover. But maybe I'm wrong?

image
image
image

This could be an alternative (at 100%):

image

at 150%:

image

DevCharly:
@eirikbakke 
> 3. The editor tabs are a bit tall--perhaps take 2 or 3 pixels off the top padding?

Funny because I considered to make them even taller 😉 

The FlatLaf JTabbedPane tabs are 32px tall, the editor tabs "only" 28px (at 100%).

image

I'll play a little bit...
At least the width of the bottom separator line should be removed from the editor tab top padding...

DevCharly:
@eirikbakke 
> 4. The non-selected tab should probably have a border so it can be distinguished from the rest of the tab area that doesn't contain any tabs.

Hmm, but the tabs are highlighted when the mouse is moved over a tab. So the user gets feedback when he moves the mouse over tabs or over the tab area that doesn't contain any tabs...

image

lkishalmi:
BTW now that all checks passed, code style issues have been addressed, shouldn't we merge this one to the master?

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,272 @@
...
+     * @return null if the requested icon is not available in vector format
+     */
+    public static Icon get(int buttonId, int buttonState) {
+        return INSTANCES.get(new SimpleEntry<Integer,Integer>(buttonId, buttonState));
+    }
+
+    @Override
+    protected void paintIcon(Component c, Graphics2D g, int width, int height, double scaling) {
+        // scale on Java 8 and Linux
+        scaling *= userScaleFactor;
```
eirikbakke: Have you tested setting a userScaleFactor != 1? Shouldn't the width/height parameters be changed as well in this case?

DevCharly: Yes, width and height are scaled in the constructor with `UIScale.scale()`

eirikbakke:
@lkishalmi @DevCharly I'm happy with merging--future work can be added in fresh commits.

@DevCharly
On the "minimize" icon discussion: I think the first three screenshots in your comment were OK. Perhaps you can compensate for thinness by making sure the line fills the entire device pixel width available to the icon. Compare for instance with the "minimize" button shown on Windows 10 on the top-right corner of any OS-painted window.

It still looks vertically misaligned next to the X--it should be like an underscore, not a hyphen (so "X _", not "X -").

On the editor tab height discussion: Most screens are widescreen these days, which means vertical space is at a real premium. A Retina MacBook Pro with a 2560x1600 device pixel resolution effectively ends up having only 800 logical pixels available in the vertical direction--30% _less_ than than the laptop I had back in 2004! For this reason I would encourage keeping things tight in the vertical direction, at least in the editor area.

On the tab border discussion: The hover effect is nice, but how does the user know where to aim the mouse when moving it towards a tab? It's not clear where the boundary between one tab and the next is. And in the editor tabs, when there are 3-4 adjacent non-selected tabs open, it will not be immediately visually clear whether the "X" belongs to the tab to the left or the right.

In the screenshots on https://www.formdev.com/flatlaf, the borderless tab works well, because each tab is only a text component. I think the presence of an X button and an icon for each tab makes the borderless approach less feasible. You could add more horizontal space between the tabs, but that would decrease the number of tabs the user can see at once (and I, for one, frequently have 15-20 tabs open).

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -0,0 +1,272 @@
...
+            int xy = round(3 * scaling);
+            int wh = round(10 * scaling);
+            /* Draw one larger window. The getWindowSymbol method ensures we are using the same
+            window border thickness as for ID_RESTORE_BUTTON. */
+            g.fill(getWindowSymbol(scaling, xy, xy, wh, wh));
+        } else if (buttonId == TabControlButton.ID_SLIDE_GROUP_BUTTON) {
+            // Draw a simple bar towards the bottom of the icon.
+            int barX = round(3 * scaling);
+            int barY = round(7 * scaling);
+            int barWidth = round(10 * scaling);
```
eirikbakke: With the VectorIcon interface, it's best to use the supplied device pixel dimensions rather than assuming the logical width you set in the constructor, i.e. do "barWidth = width - 2 * barX" instead. Otherwise you might technically draw outside the available area (due to rounding errors at non-integral HiDPI scaling factors).

(Not terribly likely to cause problems, though.)

DevCharly:
@lkishalmi +1 on merging

matthiasblaesing:
Please don't merge like this. The author information is broken:

> Karl T &lt;DevCharly@users.noreply.github.com&gt;

This is neither a fullname, nor a valid email address. Compared to other Apache projects, we are _very_ relaxed with merge policy. We need to know who was the author, so that if it there is a legal we have solid ground.

If possible I would back-out (e978ba9b235ca1eaeb6f6fddf7dca46c20c337fc) with a revert comment and create a new commit with the proper author information, preferred by @DevCharly personally.

Please don't understand this as an attack - I like the work and value it, so it should also be attributeable.

mweisshaupt1988:
I have tried this with a recent build but it does not look right.
Is this theme supported on Linux? I'm running Ubuntu 19.10 with OpenJDK-13.

Bildschirmfoto von 2019-12-12 23-05-38

lkishalmi:
That's from last week. Try to install it via Snap: snap install netbeans-dev --classic
It looks like this at the moment:
image

ggam:
To @DevCharly and all everyone involed in this effort: THANKS. Impressive appearance enhancement, specially on Linux, which at least on Fedora looked way worse than on Windows.

I'm already using the nightly build just because of this (with light theme).

PR #1773, opened 2019-12-12
[TRAVIS] Fix core network tests

blackleg:
Fix core network tests when run in a Travis worker. 

Edits `/etc/hosts` to remove hostname in the localhost entry.

Before editing: 
```
## Managed by Chef on packer-5df1faf8-b61a-05e1-6e00-7c129c076cc3.c.eco-emissary-99515.internal :heart_eyes_cat:
## cookbook:: travis_build_environment
##     file:: templates/default/etc/cloud/templates/hosts.tmpl.erb
127.0.1.1 travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 ip4-loopback xenial64
127.0.0.1 localhost nettuno travis vagrant travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 
```

After editing:
```
## File modified by nbbuild/travis/fix_hosts_file.groovy
## Managed by Chef on packer-5df1faf8-b61a-05e1-6e00-7c129c076cc3.c.eco-emissary-99515.internal :heart_eyes_cat:
## cookbook:: travis_build_environment
##     file:: templates/default/etc/cloud/templates/hosts.tmpl.erb
127.0.1.1 travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 ip4-loopback xenial64
# 127.0.0.1 localhost nettuno travis vagrant travis-job-adbfb820-92fd-40ed-9452-91ae7fb08218 
127.0.0.1 localhost nettuno travis vagrant 
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,15 @@ addons:
       - ant-optional
       - xvfb
       - openjfx
+      - groovy
+
 install:
   - export PATH="$PATH:$TRAVIS_BUILD_DIR/nbbuild/travis"
   - export DISPLAY=:99.0
   - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
   - sleep 3
+  - sudo $TRAVIS_BUILD_DIR/nbbuild/travis/fix_hosts_file.groovy
+
 script:
   - if [ "x$OPTS" == "x" ]; then OPTS="-quiet -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false"; fi
   - if [ "x$TARGET" == "x" ]; then TARGET="build"; fi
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
I don't think this is a good idea to modify the environment till the test passes. @lbruun could you please have a look at this? Travis resolved `myIpAddress` to 127.0.0.1, which is rejected by the testscript:

https://api.travis-ci.org/v3/job/625323616/log.txt

lbruun:
It goes like this: The method which tries to figure out the host's own IP address is rather elaborate.  If it can't determine anything then it returns the IPv4 loopback address, `127.0.0.1`.

Therefore the test of that method is constructed so that if it returns `127.0.0.1` it is seen a failure in the method. However, along comes Travis and Docker which by design (I guess) doesn't have a network interface.

I would need some time to dig into Travis/Docker to see if the returned `127.0.0.1`  is actually a formally correct response. If so, the test on `myIpAddress` would need to be removed.

I agrees with Matthias that changing the Travis environment is the wrong way to go about for now.

lbruun:
This issue is fixed by [PR 1835](https://github.com/apache/netbeans/pull/1835) which means the fix proposed in this present PR is redundant  - IMHO.

Yes, the Travis environment is somewhat odd in the way it does `/etc/hosts` but I guess no more odd than other virtualized environments. Bottom line: The `core.network` code should be able to handle such situation too .. and with PR 1835 it does.

blackleg:
Closed in favor of https://github.com/apache/netbeans/pull/1835

PR #1284, opened 2019-06-06
[NETBEANS-2646] Improve tabcontrol border appearance (HiDPI)

eirikbakke:
This patch improves the appearance of various main window borders under fractional HiDPI scaling factors (e.g. 150%). This applies only on Windows for now. See https://issues.apache.org/jira/browse/NETBEANS-2646 .

This PR does not depend on any of the other HiDPI-related PRs.

See the attached screenshots (view at 100% scaling).
BordersBeforePatch
BordersAfterPatch

PR #1777, opened 2019-12-13
[NETBEANS-3592] Workarounds for JDK Windows LAF bugs on HiDPI screens

eirikbakke:
Provide workarounds for various JDK bugs that caused the Windows LAF to be unusable on many HiDPI configurations, typically when an external monitor (or projector) is involved, or is disconnected or connected. Summary:

1) Fix tiny or huge GUI font size on various configurations.
2) Fix tiny or huge UI control icons when main monitor was not the HiDPI monitor at logon. (Workaround for https://bugs.openjdk.java.net/browse/JDK-8211715.)
3) Fix uneven borders on text components on non-integral HiDPI scaling factors (e.g. 150%). (A similar improvement was previously done for the NetBeans tabcontrol borders, see https://github.com/apache/netbeans/pull/1284 .)

See full details on https://issues.apache.org/jira/browse/NETBEANS-3592

Problem 1, incorrect font size
Problem 2, OS controls wrong size (before patch)
Problem 2, OS controls wrong size (after patch)
Problem 3, uneven text component borders
WindowsSwingMultiMonitorTest


lkishalmi:
Well for me it looks insane. I a good way of course. Thank you!

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/AllLFCustoms.java
```diff
@@ -154,6 +154,7 @@ public static void initCustomFontSize (int uiFontSize) {
         switchFont("windowTitleFont", fontTranslation, uiFontSize, nbDialogBold); // NOI18N
         switchFont("Spinner.font", fontTranslation, uiFontSize, nbDialogPlain); // NOI18N
         switchFont("FormattedTextField.font", fontTranslation, uiFontSize, nbDialogPlain); // NOI18N
+        switchFont("Slider.font", fontTranslation, uiFontSize, nbDialogPlain); // NOI18N
```
junichi11: :+1: Should add `// NOI18N`s to strings of the other files that are not expected translation as well.

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/windows8/DPISafeBorder.java
```diff
@@ -44,16 +44,26 @@
...
      */
     public static Border matte(int top, int left, int bottom, int right, Color color) {
-        return new DPISafeBorder(new Insets(top, left, bottom, right), color);
+        return fromDelegate(new MatteBorder(new Insets(top, left, bottom, right), color));
     }
 
-    private DPISafeBorder(Insets insets, Color color) {
-        if (insets == null)
+    public static Border fromDelegate(Border delegate) {
+        if (delegate == null)
```
junichi11: Should use curly braces(`{}`) although existing code doesn't use them.

lkishalmi: Well, my humble opinion is that we shall avoid throwing NullPointerExceptions directly from the code. It would be better to do an ```assert <whatever> != null;``` instead.

eirikbakke: Will add curly braces... I keep forgetting the code style. (The previous code was mine, too.)

Assertions can be disabled, which make them more or less useless for enforcing invariants--no way to know down the line whether they were enabled or not (e.g. when reading a log file while trying to debug an intermittent problem down the road).

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/AllLFCustoms.java
eirikbakke: Will do.

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/windows8/DPISafeBorder.java
matthiasblaesing: I agree - I tried to run netbeans with `substance` laf (or one of its siblings) and the author decided to enforce the "things interacting with GUI have to run on the EDT" rule. I ended badly (netbeans tries to do to much work off the EDT) and if Swing would have enforced the rule, netbeans could not have gone down the path in the first place. Now it is difficult to get rid of this.

The Code can be made better readable by using `java.util.Objects#requireNonNull`, then the missing braces will also not be a problem :-)

junichi11: Maybe, we can also use `@NonNull` to make "non-null" clear ? e.g. `public static Border fromDelegate(@NonNull Border delegate)`

matthiasblaesing: For the annotations we'd need a dependency on `Common Annotations`. That modules causes headaches: https://github.com/apache/netbeans/pull/1764

eirikbakke:
I've left a link to this patch at the original JDK bug for the icon scaling issue, at https://bugs.openjdk.java.net/browse/JDK-8211715 .

sopgreg:
@eirikbakke We just saw your pull request. Do you plan on opening OpenJDK tickets for the border thickness and icon scaling bugs and provide this patch to the OpenJDK? We would like to use these bugfixes in a standard L&F for our applications.

eirikbakke:
I already opened a ticket for issue 2 (icon scaling; JDK-8211715) and was working on a report for issue 3 (border widths). If you want, feel free to submit a ticket for issue 1 (font scaling).

I don't have a dev environment set up to build the JDK itself, so I'm unlikely to submit JDK patches myself. Note that the workarounds here are not patches that can be applied against the JDK itself--though you can certainly copy them into your client applications.

problemzebra2:
There's now an issue in the OpenJDK bug tracker concerning the "line thickness": https://bugs.openjdk.java.net/browse/JDK-8241561

eirikbakke:
@problemzebra2 Ah, great!

PR #1778, opened 2019-12-13
[NETBEANS-3581] Improvements to DropDownButton/DropDownToggleButton

eirikbakke:
DropDownButton/DropDownToggleButton are most commonly used in the various NetBeans toolbars--there is one pair of DropDownButton buttons in the editor toolbar, for instance.

(I use both DropDownButton and DropDownToggleButton in my NetBeans Platform application, and have tested both there. I have tested DropDownButton in the IDE as well. There are no current uses of DropDownToggleButton in the IDE.)

Details:
* Copy an old bugfix from DropDownButton to DropDownToggleButton (an added
  updateRollover call), to fix a problem where the latter would show the wrong
  icon after an icon change.
* Fix a bug where DropDownButton would occasionally get in a state where two
  mouse clicks were required to open it.
* Prevent dropdown button menus from showing up when the button is disabled.
* If DropDownButton has no action listeners registered, allow the user to click
  anywhere in the button area (not just the arrow area on the side) to open the
  dropdown menu.
* Review all differences between DropDownButton and DropDownToggleButton, and
  make sure they are legitimate. (These two classes are nearly identical, save
  for a few changes.)

dropdown button examples

### DIFF
--- a/platform/openide.awt/src/org/openide/awt/DropDownButton.java
+++ b/platform/openide.awt/src/org/openide/awt/DropDownButton.java
@@ -24,7 +24,6 @@ import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionAdapter;
 import java.beans.PropertyChangeEvent;
-import java.beans.PropertyChangeListener;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
@@ -78,11 +77,8 @@ class DropDownButton extends JButton {
 
         resetIcons();
 
-        addPropertyChangeListener(  DropDownButtonFactory.PROP_DROP_DOWN_MENU,new PropertyChangeListener() {
-            @Override
-            public void propertyChange( PropertyChangeEvent e ) {
-                resetIcons();
-            }
+        addPropertyChangeListener(DropDownButtonFactory.PROP_DROP_DOWN_MENU, (PropertyChangeEvent e) -> {
+            resetIcons();
         });
 
         addMouseMotionListener( new MouseMotionAdapter() {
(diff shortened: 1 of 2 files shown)


platform/openide.awt/src/org/openide/awt/DropDownButton.java
```diff
@@ -78,6 +79,7 @@ public DropDownButton( Icon icon, JPopupMenu popup ) {
         resetIcons();
 
         addPropertyChangeListener(  DropDownButtonFactory.PROP_DROP_DOWN_MENU,new PropertyChangeListener() {
```
lkishalmi: Lambda?

platform/openide.awt/src/org/openide/awt/DropDownButton.java
```diff
@@ -152,22 +159,33 @@ public void mouseExited( MouseEvent e ) {
...
                     JPopupMenu menu = getPopupMenu();
                     if( null != menu ) {
                         menu.removePopupMenuListener( this );
                     }
+                    /* If the popup was closed by a mouse click inside the button area, the button
+                    may also receive a mousePressed event, although this seems not to be guaranteed.
+                    Ignore any such button press while the popup is closing, to avoid interpreting
+                    the press as a click to open the menu again. */
+                    popupClosingInProgress = true;
+                    SwingUtilities.invokeLater(new Runnable() {
```
lkishalmi: Lambda?

eirikbakke: Ah, sure. Didn't realize they existed in Java 8.

eirikbakke:
I'm going to test this on MacOS too, before merging it.

eirikbakke:
Latest version now tested on both Windows and MacOS. Merging.

PR #1779, opened 2019-12-13
[NETBEANS-3196] Fix lazy Node loading when switching from Children.LEAF to another Children instance

eirikbakke:
There's a bug in org.openide.nodes.Node.setChildren(Children). Basically, if Children.LEAF.getNodesCount() is _ever_ called, from any module, then Node implementations that switch from Children.LEAF to another Children implementation will no longer get lazy loading behavior.

(This bug was found in a NetBeans Platform application that relies heavily on lazy loading of nodes. It calls setChildren(Children.LEAF) in order to prevent an an expansion handle from appearing in OutlineView for nodes that are known to have no children, and then calls setChildren again if children later do appear.)

See https://issues.apache.org/jira/browse/NETBEANS-3196 for more details.

### DIFF
--- a/platform/openide.nodes/src/org/openide/nodes/Node.java
+++ b/platform/openide.nodes/src/org/openide/nodes/Node.java
@@ -500,7 +500,7 @@ public abstract class Node extends FeatureDescriptor implements Lookup.Provider,
             hierarchy = ch;
             hierarchy.attachTo(Node.this);
 
-            if (wasInited && hierarchy != Children.LEAF) {
+            if (wasInited && !wasLeaf && hierarchy != Children.LEAF) {
                 // init new children if old was inited
                 hierarchy.getNodesCount();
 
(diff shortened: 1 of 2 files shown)


sdedic:
What about a test :) ?

eirikbakke:
Yes, should add a test here, though I haven't had time to return to this yet. Though this patch has been running without problem in my working IDE and on my platform application for 4 months now.

eirikbakke:
Finally got around to adding a test here...

eirikbakke:
I'll leave this PR here for comments for another 3 months, and then merge if there are no objections :-)

matthiasblaesing:
@eirikbakke if you still intent to get this in (given this has a test and was extensively tested by you I think it is very sensible), I would merge early after 12.5 is released or even now (master is already open for changes for 12.6).

eirikbakke:
OK, will merge now, then! Thanks!

PR #1785, opened 2019-12-16
[NETBEANS-3586] : Same module name when created, breaks things. Showi…

singh-akhilesh:
…ng null pointer exception on screen

https://issues.apache.org/jira/browse/NETBEANS-3586


sdedic:
What was the NPE, could you please add the stacktrace to the JIRA issue ?

singh-akhilesh:
Added stack trace. Ref : https://issues.apache.org/jira/browse/NETBEANS-3586

PR #1789, opened 2019-12-18
[NETBEANS-3428] Update FlatLaf from 0.21 to 0.22

DevCharly:
This PR updates FlatLaf to 0.22, which includes following fixes/improvements for NetBeans:

- Changed ProgressBar styling to reduce height so that it fits better into the NB status bar.

Here is a full list of changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/0.22

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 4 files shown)


lkishalmi:
Travis is fine with the change. Just rebase and merge.

DevCharly:
Don't think that I have access rights to do so...?

PR #1793, opened 2019-12-18
[NETBEANS-3204] Add space before arrow in switch case

graben:


jlahoda:
I think this is a move in a good direction, but I wonder if:
-the spaces around the arrow should be customizable (although could be done separately as well)
-if there is no block after the case <label>->, the follow-up statement should be put at the same line as the case (which is presumably a separate thing)

graben:
@jlahoda : Ready to merge?

graben:
@jlahoda : Can we get this into 11.3?

PR #1794, opened 2019-12-18
[TRAVIS] Enable tests for ide/xml.wsdl.model module

blackleg:
Enable tests for ide/xml.wsdl.model module


matthiasblaesing:
Looks sane. Thank you.

PR #1799, opened 2019-12-19
[NETBEANS-3615] - enum is now a keyword and not an identifier

BradWalker:
As of Java 1.5, the "enum" is a keyword and can not longer be used as an identifier.

Change the use of "package org.openide.util.enum" to "package org.openide.util.enumeration"..

### DIFF
--- a/harness/nbjunit/test/unit/src/org/netbeans/testjunit/AskForOrgOpenideUtilEnumClass.java
+++ b/harness/nbjunit/test/unit/src/org/netbeans/testjunit/AskForOrgOpenideUtilEnumClass.java
@@ -41,7 +41,7 @@ public class AskForOrgOpenideUtilEnumClass extends TestCase {
             if (l == NbTestCase.class.getClassLoader()) {
                 fail("This test shall not be loaded by the same classloader!");
             }
-            Class<?> access = Class.forName("org.openide.util.enum.ArrayEnumeration");
+            Class<?> access = Class.forName("org.openide.util.enumeration.ArrayEnumeration");
             System.setProperty("en.one", "OK");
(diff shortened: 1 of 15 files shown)


matthiasblaesing:
_Every_ class in that module is deprecated since 06/2004. That are 15 years. Lets drop the module and be done with it. We gain nothing from doing an incompatible change (you'd need to document it, update the apichanges.xml) to reinstate these dead classes.

BradWalker:
> 
> 
> _Every_ class in that module is deprecated since 06/2004. That are 15 years.
> Lets drop the module and be done with it. 

Just so I'm clear.. Let remove the org.openide.util.enum from the build..
Correct?

matthiasblaesing:
I'm inclined to leave the module as is - someone was there before us - from the `build.xml`:

```
    <!-- ... -->
    <available property="is.jdk9" classname="java.lang.Module"/>

    <target name="compile" depends="-copy" unless="is.jdk9">
        <antcall target="projectized-common.compile"/>
    </target>
    <!-- ... -->
```

This translates to: "If the build is run on JDK 9 and newer, just skip this module." The moment we drop build support for JDK8, the module will be gone. We could drop it now, but do we gain anything?

jlahoda:
I agree with Matthias - I don't think there's a point in renaming the package. The sole purpose of the module now is compatibility. So we could drop it in some way (like move the binary to some update centre, where anyone who would want to use it could download it).

BradWalker:
Sorry to be delayed in responding, holidays here in the US.. I'm in agreement with the previous comments..  I'll just leave it for now and continue on with other code cleanup issues.. Thanks for looking at this..

PR #1803, opened 2019-12-21
[NETBEANS-3249] Chrome connector connection with NetBeans

matthiasblaesing:
The call to toRealPath ensures, that symlinks are resolved. It was
observed, that chrome on macOS reports the file url with symlinks
resolved, so align with that.

Closes: #1577


geertjanw:
Tried this, and it works, thanks, from my point of view this can be merged.

matthiasblaesing:
Great - thanks for testing @geertjanw.

PR #883, opened 2018-09-18
[NETBEANS-1227] Mark Windows launchers as DPI-aware

eirikbakke:
Add manifest information to the netbeans64.exe, app64.exe, and nbexec64.exe executables (and their 32-bit equivalents) to tell Windows that Java supports HiDPI scaling, like javaw.exe already does. This enables HiDPI support on Windows, without needing to fuzz around with special settings. This works on both Java 9 and Java 10. See https://issues.apache.org/jira/browse/NETBEANS-1227 .

For those who are still using Java 8, NetBeans will now appear small (but sharp) on HiDPI screens. The old behavior, of NetBeans appearing normal size but blurry, can be reverted by right-clicking the NetBeans shortcut, and clicking Properties->Compatibility->Change HiDPI Settings->Override high DPI scaling behavior->Scaling performed by System. If we expect this to be a problem, an alternative would be have the launcher detect the use of Java 8 (and Windows 8.1 minimum) and turn off DPI awareness programmatically using the Windows [SetProcessDpiAwareness](https://docs.microsoft.com/en-us/windows/desktop/api/shellscalingapi/nf-shellscalingapi-setprocessdpiawareness) API call. But I'd rather recommend that people with HiDPI screens should run NetBeans with Java 9 or above, where support for such screens is actually available.

This pull request also fixes some relative paths that broke when directories were moved around, and makes two other minor fixes.

Note that the new Windows launchers (app64.exe and netbeans64.exe) seem not to suppress the console properly; this is an unrelated bug that was present before this patch.

PR #1804, opened 2019-12-22
[NETBEANS-2954] Fix broken tab dragging on HiDPI displays (Windows)

eirikbakke:
On Windows, on displays that have a HiDPI scaling set (e.g. 150% or 200%), drag-and-drop to reposition editors and modes in the NetBeans window system does not work properly. See https://issues.apache.org/jira/browse/NETBEANS-2954 .

This appears to be due to a JDK bug, where java.awt.dnd.DragSourceEvent.getLocation() returns the wrong screen coordinate when HiDPI displays are involved. The workaround is to use MouseInfo.getPointerInfo() instead--that one works correctly.

Both MouseInfo and DragSourceEvent get their coordinates from a call to the "GetCursorPos" Windows API function. The latter returns device coordinates rather than logical coordinates for DPI-aware applications (including java.exe and the netbeans launcher since https://github.com/apache/netbeans/pull/883 ).

But in the implementation of MouseInfo.getPointerInfo(), extra code was added to convert the device coordinates to logical coordinates, as part of the original JDK patch which introduced HiDPI support on Windows. See Java_sun_awt_windows_WMouseInfoPeer_fillPointWithCoords in https://github.com/openjdk/jdk/blame/6bab0f539fba8fb441697846347597b4a0ade428/src/java.desktop/windows/native/libawt/windows/MouseInfo.cpp . This was forgotten in the code that initializes the coordinates for DragSourceEvent (see GetCursorPos call for call_dSCenter/call_dSCmotion in https://github.com/openjdk/jdk/blob/1440dc39400f89b7df8af2249f30eaf5092c1574/src/java.desktop/windows/native/libawt/windows/awt_DnDDS.cpp ).

The workaround was tested on both single and multi-monitor setups.

In a few remaining cases, involving multi-monitor setups with different DPI scaling factors, the semi-transparent preview thumbnail that appears while dragging an editor may end up on the wrong screen. This is due to a different JDK bug ( https://bugs.openjdk.java.net/browse/JDK-8211999 ). In these cases, however, the orange tab outline, which appears during dragging to illustrate where the tab would end up if dropped, does appear in the correct place, which makes this less of a usability issue.

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/view/dnd/DragAndDropFeedbackVisualizer.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/view/dnd/DragAndDropFeedbackVisualizer.java
@@ -128,8 +128,10 @@ public class DragAndDropFeedbackVisualizer {
     }
 
     public void update(DragSourceDragEvent e) {
-        if( null != dragWindow )
-            dragWindow.setLocation( e.getLocation().x-dragOffset.x, e.getLocation().y-dragOffset.y );
+        if( null != dragWindow ) {
+            Point location = WindowDnDManager.getLocationWorkaround(e);
+            dragWindow.setLocation( location.x-dragOffset.x, location.y-dragOffset.y );
+        }
     }
 
     public void dispose( boolean dropSuccessful ) {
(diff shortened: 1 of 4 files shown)


lkishalmi:
Is there any reason not to merge this one?

eirikbakke:
I'll go through it once more myself (checking for all usages of the relevant buggy methods) and then merge it.

eirikbakke:
I reviewed all usages of DragSourceEvent, DragSourceDragEvent, and DragSourceDropEvent in the Netbeans repo, and ensured the workaround was applied whenever required. Will merge once Travis gives it a green check.

eirikbakke:
Also checked to see if the bug is present in DragGestureEvent.getDragOrigin. It's not; that one works correctly on HiDPI displays. Merging.

eirikbakke:
The underlying JDK bug now has ticket https://bugs.openjdk.java.net/browse/JDK-8238575

PR #1806, opened 2019-12-23
 If an annotation processor crashes, just log the error and continue, rather than crash the whole processing.

jlahoda:


lkishalmi:
Cool!

PR #1813, opened 2019-12-24
[NETBEANS-3633] Use HTTPS to download Gradle Version information

lkishalmi:
Also removed some rawtype warnings from the Settings panel.

### DIFF
--- a/groovy/gradle/src/org/netbeans/modules/gradle/GradleDistributionManager.java
+++ b/groovy/gradle/src/org/netbeans/modules/gradle/GradleDistributionManager.java
@@ -33,6 +33,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


groovy/gradle/src/org/netbeans/modules/gradle/GradleDistributionManager.java
```diff
@@ -188,8 +188,10 @@ public NbGradleVersion evaluateGradleWrapperDistribution(File rootDir) {
         List<NbGradleVersion> ret = new ArrayList<>();
         JSONParser parser = new JSONParser();
         try {
-            URL allVersions = new URL("http://services.gradle.org/versions/all"); //NOI18N
-            try (InputStreamReader is = new InputStreamReader(allVersions.openStream())) {
+            URL allVersions = new URL("https://services.gradle.org/versions/all"); //NOI18N
+            HttpsURLConnection conn = (HttpsURLConnection) allVersions.openConnection();
+            conn.setInstanceFollowRedirects(true);
+            try (InputStreamReader is = new InputStreamReader(conn.getInputStream())) {
```
matthiasblaesing: I would add the charset to the InputStreamReader. There are still JDKs out there, that don't use UTF-8 by default and in the future the default might change. `java.nio.charset.StandardCharsets.UTF-8` would be the value I'd use.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDistributionManager.java
```diff
@@ -188,8 +188,10 @@ public NbGradleVersion evaluateGradleWrapperDistribution(File rootDir) {
         List<NbGradleVersion> ret = new ArrayList<>();
         JSONParser parser = new JSONParser();
         try {
-            URL allVersions = new URL("http://services.gradle.org/versions/all"); //NOI18N
-            try (InputStreamReader is = new InputStreamReader(allVersions.openStream())) {
+            URL allVersions = new URL("https://services.gradle.org/versions/all"); //NOI18N
+            HttpsURLConnection conn = (HttpsURLConnection) allVersions.openConnection();
+            conn.setInstanceFollowRedirects(true);
```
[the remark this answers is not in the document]
lkishalmi: Removed. The one thing it cannot do is the HTTP -> HTTPS redirect, but that does not needed either.

groovy/gradle/src/org/netbeans/modules/gradle/GradleDistributionManager.java
lkishalmi: Added that. Thanks!

PR #1815, opened 2019-12-25
Ubuntu installs Chromium as a snap and the config directory is moved

matthiasblaesing:
The problem results in netbeans not being able to find the installed plugin and
thus not loading the debugging support for HTML5/JS projects.


lkishalmi:
Looks good. Thanks!

lkishalmi:
I might miss a JIRA issue, though this one is trivial enough.

matthiasblaesing:
I don't understand the religion about JIRA. If the jira issue is just the same as the commit message, I totally don't see point.

PR #1816, opened 2019-12-25
[NETBEANS-3636] Update Spring Framework 4.3.25

blackleg:
[NETBEANS-3636](https://issues.apache.org/jira/browse/NETBEANS-3636) Update java/libs.springframework module to update Spring Framework 4.3.25 library dependency for ant projects.


java/libs.springframework/external/spring-framework-4.3.25.RELEASE-notice.txt
```diff
@@ -0,0 +1,11 @@
+Spring Framework 4.3.25.RELEASE
+Copyright (c) 2002-2019 Pivotal, Inc.
```
matthiasblaesing: You can remove everything from this file from line 3 onwards. We are required to carry the _relevant_ information from the NOTICE file of the upstream libraries. In this case lines 4-6 belong into the license file and line 8-11 are two broad to carry anything meaningful.

It would be great if you could update the notice file for 5.X also  (I missed that in that review).

Thank you.

blackleg: Removed

matthiasblaesing:
Thank you.

PR #1818, opened 2019-12-26
[NETBEANS-3638] Add Indenting options for XML

lkishalmi:
Same trick, used for Yaml.


lkishalmi:
Noticed that. Tried to add a different preview file which is well formed, however I cannot set that up in the file attribute, I'm debugging it. In the worst case I just fix the original preview file. With that we would lose showcasing the font color settings on erorred element.

lkishalmi:
XMLIndents

PR #1826, opened 2019-12-28
[NETBEANS-3644] Adjusted Welcome Screent to FlatLAF

lkishalmi:
image
image

There would be still some work to do, like the colors of the hyperlinks and the logo for the light theme shall be adjusted, that is for another PR. I think this one is still an improvement.


lkishalmi:
StarScreenDark2
StartScreenLight2

lkishalmi:
Also changed the output colors matching goes better with the FlatLaf Dark editor theme:
FlatLafDarkOutput

matthiasblaesing:
I'm not sure, but I think I see negative fallout from this. I noticed today, that the start screen of the master build looked strange:

Bildschirmfoto_2020-01-18_16-05-03

The screenshot was taken on Ubuntu 19.10 with Xfce4 as desktop environment. This was observed with the Ubuntu provided Open JDK 11.0.5 and Oracle JDK 1.8.0_201. The LAF is Metall (automatically chosen).

Could you please have a look at this?

PR #1827, opened 2019-12-28
[NETBEANS-3041] Honor gradle.properties processing order for JVM args.

lkishalmi:


groovy/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/GradleCommandLineTest.java
```diff
@@ -277,4 +284,26 @@ public void testCombine3() {
...
+
+    @Test
+    public void testJVMArgs2() throws IOException {
+        TemporaryFolder root = new TemporaryFolder();
+        root.create();
+        File props = root.newFile("gradle.properties");
+        Files.write(props.toPath(), Arrays.asList("org.gradle.jvmargs=\"-Dfile.encoding=UTF-8\" -Dsomething=\"space value\""));
+        List<String> jvmargs = new ArrayList<>();
+        GradleCommandLine.addGradleSettingJvmargs(root.getRoot(), jvmargs);
+        assertEquals(Arrays.asList("-Dfile.encoding=UTF-8", "-Dsomething=space value"), jvmargs);
```
matthiasblaesing: This looks suspicious. I would have expected:

```java
assertEquals(Arrays.asList("-Dfile.encoding=UTF-8", "-Dsomething=\"space value\""), jvmargs);
```

groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -825,45 +826,56 @@ public void setStackTrace(StackTrace st) {
...
+                } catch (IOException ex) {
+                    LOGGER.log(Level.INFO, "Cannot read property file: '" + f.getAbsolutePath() + "' as: " + ex.getMessage());
                 }
-            } catch (IOException ex) {
-                LOGGER.log(Level.INFO, "Cannot read property file: '" + f.getAbsolutePath() + "' as: " + ex.getMessage());
             }
         }
     }
 
+    final static String removeSingleQuotes(String s) {
```
matthiasblaesing: What is the reason for the special handling of single quotes after they are already parsed?

groovy/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/GradleCommandLineTest.java
lkishalmi: It should be really "-Dsomething=space value" as that would be passed to JVM and the JVM does parse the quotes.

groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
lkishalmi: Unfortunately the command line parser in NetBeans does not parse the single quotes. It's maybe a bug there. It was easier to add a workaround to the most common case here than fix it there. Also the impact of fixing it in the utilities might have unwanted consequences.

groovy/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/GradleCommandLineTest.java
matthiasblaesing: Ok - verified with a small java programm.

groovy/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
matthiasblaesing: But this only works for the simplest of cases doesn't it? This:

```java
import java.util.Arrays;

class test {
	public static void main(String[] argv) {
		System.out.println(Arrays.toString(argv));
	}
}
```

Run as:

```
matthias@athena:~$ java -cp . test -Dsomething="space value"
[-Dsomething=space value]
matthias@athena:~$ java -cp . test -Dsomething='space value'
[-Dsomething=space value]
matthias@athena:~$ 
```

shows, that also embedded single quotes need to be removed. Is this really needed or just a theoretical issue?

lkishalmi: Well thought it over and fixed the BaseUtilities.parseParameters instead in #1829 after all it was fun.

PR #1829, opened 2019-12-29
[NETBEANS-3648] Support single quotes in parameter parsing

lkishalmi:


platform/openide.util/src/org/openide/util/BaseUtilities.java
```diff
@@ -1036,12 +1036,11 @@ public static Object toPrimitiveArray(Object[] array) {
...
     */
     public static String[] parseParameters(String s) {
-        int NULL = 0x0; // STICK + whitespace or NULL + non_"
-        int INPARAM = 0x1; // NULL + " or STICK + " or INPARAMPENDING + "\ // NOI18N
-        int INPARAMPENDING = 0x2; // INPARAM + \
-        int STICK = 0x4; // INPARAM + " or STICK + non_" // NOI18N
-        int STICKPENDING = 0x8; // STICK + \
-        ArrayList<String> params = new ArrayList<String>(5);
+        final int NULL = 0x0; // STICK + whitespace or NULL + non_"
+        final int IN_PARAM = 0x1; // NULL + " or STICK + " or INPARAMPENDING + "\ // NOI18N
```
matthiasblaesing: I would remove the comments here - I don't find them helpful and they distract me from the code.

lkishalmi: That's true, I've spent some time to figure out what would the STICK and others mean in the original parser. Removed it for good.

PR #1835, opened 2020-01-01
[NETBEANS-3652] PAC: myIpAddress() returns "127.0.0.1" when executed in Travis env

lbruun:
Currently, tests fails because the `LocalAddressUtils.getMostLikelyLocalInetAddresses()` - when executed by Travis - returns "127.0.0.1" which means the code was not able to determine the local host's IP.

This PR fixes this problem by improving local IP address discovery:

An additional method is added to the multiple ways of finding own host's IP: **UDP Socket**. In some environments (e.g. Travis) this addition will be able to find the IP where the other methods fail. However, discovering the IP of the local host remains an imprecise discipline.

Also, the `getMostLikelyLocalInetAddresses()` method is now more precise about filtering out loopback addresses potentially returned by one of the underlying methods. Returning the loopback address from this method should really, really be a last resort.

Test of the `core.network` module is re-instated in `.travis.yml` because with this change the test no longer fails when executed in Travis env.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,7 +120,7 @@ matrix:
             - hide-logs.sh ant $OPTS -f platform/core.kit test
             - hide-logs.sh ant $OPTS -f platform/core.multiview test
             - hide-logs.sh ant $OPTS -f platform/core.netigso test
-            #- hide-logs.sh ant $OPTS -f platform/core.network test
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
@blackleg @lbruun thank you both.

PR #1839, opened 2020-01-03
[NETBEANS-3428] FlatLaf: new close, left, right, drop-down and maximi…

DevCharly:
…ze buttons for multi-tabs, breadcrumb bar, search bar, etc

Old:

image

New:

image

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/ButtonFactory.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/ButtonFactory.java
@@ -26,6 +26,7 @@ import javax.swing.Action;
 import javax.swing.Icon;
 import javax.swing.JButton;
 import javax.swing.Timer;
+import javax.swing.UIManager;
 import org.netbeans.core.multitabs.impl.TabListPopupAction;
 import org.netbeans.swing.tabcontrol.TabbedContainer;
 import org.netbeans.swing.tabcontrol.event.TabActionEvent;
@@ -47,7 +48,11 @@ public final class ButtonFactory {
      */
     public static JButton createDropDownButton( Controller controller ) {
         final JButton btn = new JButton( new TabListPopupAction(controller) );
-        btn.setIcon( ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/down.png", true) ); //NOI18N
+        Icon icon = UIManager.getIcon("nb.multitabs.button.dropdown.icon");
+        if (icon == null) {
+            icon = ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/down.png", true ); //NOI18N
+        }
+        btn.setIcon( icon );
         btn.setActionCommand( "pressed"); //NOI18N
         btn.setFocusable( false );
         btn.setToolTipText( NbBundle.getMessage(ButtonFactory.class, "Hint_DocumentList") );
@@ -61,7 +66,11 @@ public final class ButtonFactory {
      */
     public static JButton createMaximizeButton( final Controller controller ) {
         final JButton btn = new JButton();
-        btn.setIcon( ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/maximize.png", true) ); //NOI18N
+        Icon icon = UIManager.getIcon("nb.multitabs.button.maximize.icon");
+        if (icon == null) {
+            icon = ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/maximize.png", true ); //NOI18N
+        }
+        btn.setIcon( icon );
         btn.setToolTipText( NbBundle.getMessage(ButtonFactory.class, "Hint_MaximizeRestore") );
         btn.addActionListener( new ActionListener() {
 
@@ -83,7 +92,10 @@ public final class ButtonFactory {
      */
     public static JButton createScrollLeftButton( Action scrollAction ) {
         JButton btn = new TimerButton( scrollAction );
-        Icon icon = ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/left.png", true); //NOI18N
+        Icon icon = UIManager.getIcon("nb.multitabs.button.left.icon");
+        if (icon == null) {
+            icon = ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/left.png", true ); //NOI18N
+        }
         btn.setIcon( icon );
         btn.setToolTipText( NbBundle.getMessage(ButtonFactory.class, "Hint_ScrollLeft") );
         return btn;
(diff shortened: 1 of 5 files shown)


platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -130,6 +131,21 @@ protected void paintIcon(Component c, Graphics2D g, int width, int height, doubl
         Color bgColor = new Color(0, 0, 0, 0); // Alpha zero means no background.
         Color fgColor = foreground;
         {
+            int buttonState = this.buttonState;
+            if (buttonState == -1) {
+                // get button state from component
+                if (!c.isEnabled()) {
+                    buttonState = TabControlButton.STATE_DISABLED;
```
eirikbakke: Have you tested that this state ends up being used? Are you sure the button does not end up trying to make its own disabled version using an ImageFilter? (In the latter case the disabled version would not look good on HiDPI screens.)

platform/openide.awt/src/org/openide/awt/CloseButtonFactory.java
```diff
@@ -141,6 +141,9 @@ private static boolean isGTKLaF () {
     }
 
     private static Icon getCloseTabImage() {
+        if( null == closeTabImage ) {
+            closeTabImage = UIManager.getIcon("nb.close.tab.icon.enabled.icon"); //NOI18N
```
eirikbakke: Adding new keys to UIManager.getDefaults() constitutes an API change; should probably update apichanges.xml. And perhaps go over the names of the new keys one more time to make sure they follow the convention of the existing keys.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
```diff
@@ -45,6 +47,21 @@
             // scale on Java 8 and Linux
             SPLIT_PANE_DIVIDER_SIZE_VERTICAL, UIScale.scale(4),
             SPLIT_PANE_DIVIDER_SIZE_HORIZONTAL, UIScale.scale(4),
+
+            // for org.openide.awt.CloseButtonFactory
```
eirikbakke: I remember from https://github.com/apache/netbeans/pull/859/files that the close-button drawing code had to be duplicated between and org.openide.awt.AquaVectorCloseButton and o.n.swing.tabcontrol.plaf.AquaVectorTabControlIcon. Are both of the corresponding locations for FlatLAF updated in this PR? Do the new UIManager keys prevent the need for this duplication?

platform/core.multitabs/src/org/netbeans/core/multitabs/ButtonFactory.java
```diff
@@ -47,7 +48,11 @@
      */
     public static JButton createDropDownButton( Controller controller ) {
         final JButton btn = new JButton( new TabListPopupAction(controller) );
-        btn.setIcon( ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/down.png", true) ); //NOI18N
+        Icon icon = UIManager.getIcon("nb.multitabs.button.dropdown.icon");
+        if (icon == null) {
+            icon = ImageUtilities.loadImageIcon( "org/netbeans/core/multitabs/resources/down.png", true ); //NOI18N
```
eirikbakke: This is just a fallback, right? I assume all of these icons are vector-drawn to work with HiDPI screens?

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
DevCharly: Yes, this works. The reason is that class `FlatTabControlIcon` is not an instance of `javax.swing.ImageIcon`. So `LookAndFeel.getDisabledIcon()` returns `null` and the normal icon is used.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
```diff
@@ -130,6 +131,21 @@ protected void paintIcon(Component c, Graphics2D g, int width, int height, doubl
         Color bgColor = new Color(0, 0, 0, 0); // Alpha zero means no background.
         Color fgColor = foreground;
         {
+            int buttonState = this.buttonState;
+            if (buttonState == -1) {
```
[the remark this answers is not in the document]
DevCharly: done

platform/core.multitabs/src/org/netbeans/core/multitabs/ButtonFactory.java
DevCharly: No, not just a fallback.
Vector icons are only used for FlatLaf.

Other LaFs still use the bitmap images.

DevCharly: But should be not very difficult to extend `Windows8VectorTabControlIcon` and `AquaVectorTabControlIcon` in the same way I've extended `FlatTabControlIcon`...

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
DevCharly: Yes, I thinks so. Also FlatLaf is a single module, but Aqua/Windows LAF support is scattered to many modules.
Anyway, there is only one close button in FlatLaf in `FlatTabControlIcon`.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatTabControlIcon.java
eirikbakke: Great!

platform/openide.awt/src/org/openide/awt/CloseButtonFactory.java
lkishalmi: While introducing new keys is an API change, the keys are not documented in ```apichange.xml```. That could be an additional task, or we could introduce a LAF guide on wiki. I would not hold this one off due to that.

eirikbakke: Sure--it's not worth any extra effort now.

PR #1843, opened 2020-01-04
Allow ".external" files to define message digest values for download verification

matthiasblaesing:
This formalises the format of the ".external" files and introduces a new
content line, that allows specification of a message digest algorithm
and the digest value, that results from applying the message digest to
the referenced URL.

At download time the message digest value is calculated from the
downloaded data and compared. If it does not match, an exception is
raised.

Multiple message digest can be provided and message digests with
unsupported algorithms are ignored. This allows for future evolution,
when new algorithms are added and older ones are phased out.


emilianbold:
+1

emilianbold:
PS: Now you might as well revisit PR #1260 and see how to add the same (SHA 256/512) message digest algorithm and the digest value to external binaries.

jlahoda:
Sounds great! (Eventually, we could hopefully get rid of actually having the CRC there - that is very inconvenient to get.)

emilianbold:
Yes, ideally CRC would be optional if another digest is available (since any of them would be better).

matthiasblaesing:
Thank you for the review - I'll merge as is and will revisit later. Removing CRC will be more work, as it is used internally and not only as checksum.

jlahoda:
@matthiasblaesing - sure I was just talking about a possible future improvements. I can even look at that myself, if you prefer.

PR #1844, opened 2020-01-04
[NETBEANS-2523] Installer creates invalid Win Registry value

lbruun:
The `WindowsRegistry_set32BitValue0()` function was incorrectly using the length of the Registry name (e.g. 'NoModify') as the the length of value. Since the Registry value we are setting is of type DWORD the length is indeed always exactly 4 bytes, regardless of OS bitness. 

Solution: `sizeof(dword)` will always return 4, but we might as well have hard-coded the 4.

Impact: The impact of bug has been seen when NBI sets the `NoModify` registry value in the `Uninstall` section of the Registry. The bug has existed for years but has only minor impact, for example with system analyzers flagging the 'NoModify' value as suspect or not being able to read the value with tools such as Powershell.

### DIFF
--- a/nbi/engine/native/jnilib/windows/src/jni_WindowsRegistry.c
+++ b/nbi/engine/native/jnilib/windows/src/jni_WindowsRegistry.c
@@ -430,7 +430,7 @@ JNIEXPORT void JNICALL Java_org_netbeans_installer_utils_system_windows_WindowsR
     DWORD  dword     = (DWORD) jValue;
     LPBYTE byteValue = (LPBYTE) &dword;
     
-    if (!setValue(getMode(jMode),getHKEY(jSection), key, name, REG_DWORD, byteValue, sizeof(name), 0)) {
+    if (!setValue(getMode(jMode),getHKEY(jSection), key, name, REG_DWORD, byteValue, sizeof(dword), 0)) {
         throwException(jEnv, "Cannot set value");
     }
     


matthiasblaesing:
@rtaneja1 could you please also have a look at this?

rtaneja1:
Change looks fine to me.

matthiasblaesing:
Thank you both.

rtaneja1:
Ibruun, Have you tested with this change, I have not worked on this code, do you know how this change gets consumed by installer. If this goes into some .dll, then it needs to be rebuilt and uploaded to external binaries server. I had updated nlw.exe code, please refer https://github.com/apache/netbeans/pull/1196.

lbruun:
@rtaneja1 
The actual build process for these NBI artifacts is a blur to me.  That was the case even in the pre-Apache days.

I can tell you how it is meant to be:

The build result of the project we've just changed is a (quite small) `.dll` file. This lib is a helper library for the NBI Engine and thus gets physically embedded inside `nbi-engine.jar`. (if you open up a copy of `nbi-engine.jar` you'll find a path called `native/jnilib/windows/` where it lives). The `.dll` library is only used from Java code (it is JNI stuff) and is as such unrelated to `nlw.exe` which is simply the launcher for that Java code..

As I said my knowledge of the build process is near zero. What I wonder myself are:

1. Have we actually triggered a new build of the `.dll`. If so, where did the build result go?
2. How does the build result, the .dll file, get embedded into `nbi-engine.jar` ?. Do we need to trigger that build too?

I believe that even in the pre-Apache days there were manual steps involved but that is just a hunch.

In the end the test of the change is simple:  Do an install of the IDE itself using installer (.exe) and when finished open `regedit.exe`. Have a look at the screenshot in [NETBEANS-2523](https://issues.apache.org/jira/browse/NETBEANS-2523). The yellow highlight is what goes wrong. It should display the value "1" instead of displaying "(invalid DWORD (32-bit) value)".

rtaneja1:
I think this is part of http://netbeans.osuosl.org/binaries/2B892D9E648A792EDB0F705378E06F84C322E208-nbi.engine-external-binaries-11.zip
and needs to be built and uploaded manually.

PR #1692, opened 2019-11-28
Add TypeScript editor

jlahoda:
An attempt to provide TypeScript editor features using the LSP client/server.

PR #1845, opened 2020-01-04
Adding LICENSE/NOTICE for the VisualStudio Code extension for Java

jlahoda:
We have a sever for the language server protocol that supports Java editing, and there is also a VisualStudio Code extension wrapping the server so that it can be installed into the VS Code. This patch tries to add LICENSE/NOTICE for the binary of the extension, so that is a (hopefully) a valid convenience binary. I'd like to do a (preview) release of the extension sometime soon.

Some enhancements are done to the CreateLicenseSummary task, so that the files can be generated conveniently. The "Prepare Bundles" tool (that packs npm packages and autodetects licenses, etc.) from https://github.com/apache/netbeans/pull/1692 is also enhanced and generalized, and is used to generate the LICENSE/NOTICE entries for the included npm packages.


java/java.lsp.server/build.xml
```diff
@@ -51,4 +51,65 @@
...
+            <arg value="vscode/" />
+            <arg value="${build.dir}/" />
+        </exec>
+
+        <delete dir="${build.dir}/bundles" />
+        <mkdir dir="${build.dir}/bundles/package" />
+        <copy file="vscode/package.json" todir="${build.dir}/bundles/package" />
+
+        <exec executable="mvn" failonerror="true" dir="${nb_all}/nbbuild/misc/prepare-bundles">
+            <arg value="install" />
```
matthiasblaesing: Shouldn't package be enough? `exec:java` should use the build package from target, not the one from the local repository.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+
+/**
+ * Prepare bundles and license files for a group of node modules.
+ */
+public class PrepareBundles {
+
+    private static final String[] LICENSE_FILE_NAMES = {
```
matthiasblaesing: Just a thought: Would it be easier to use a (list of) case insenstive regular expression(s) to match this? I would bet there are "LICENSE-ALv2.txt", "LICENSE-EPL.txt" and so on out there and the special case for MIT will only catch the most probable one.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+        "License",
+        "LICENSE",
+        "LICENSE.txt",
+        "LICENSE-MIT.txt",
+        "license.txt",
+        "License.txt",
+        "LICENSE.md",
+        "license.md"
+    };
+    private static final String nl = System.getProperty("line.separator");
```
matthiasblaesing: I would hardcode this to `\n`. That way it is not platform dependent and most editor deal perfectly fine with unix line endings.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+
+                for (String l : LICENSE_FILE_NAMES) {
+                    if (Files.isReadable(module.resolve(l))) {
+                        licenseText = readString(module.resolve(l));
+                        break;
+                    }
+                }
+
+                if (licenseText == null) {
+                    String hardcodedLicenseName = name + "-" + version + "-license";
```
matthiasblaesing: I understand that this allows supplying license information for packages, that can't be correctly parsed. Looks ok, though I see potential problems in the future, if we get into the habit of bundling "many" JS libraries.

I would consider removing the null check in 132. That way invalid license info could be overriden.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+                    licenseText = "Parts of this work are licensed:\n" +
+                                  licenseText +
+                                  "\n\n" +
+                                  "Parts of this work are licensed:\n" +
+                                  readString(thirdpartynoticestxt);
+                } else {
+                    licenseText = licenseText;
+                }
+
+                Path noticestxt = module.resolve("CopyrightNotice.txt");
```
matthiasblaesing: This seems to be problematic. One sample I found from a quick look at the generated NOTICE file was this:

```
/*! *****************************************************************************
Copyright (c) Microsoft Corporation. All rights reserved. 
Licensed under the Apache License, Version 2.0 (the "License"); you may not use
this file except in compliance with the License. You may obtain a copy of the
License at http://www.apache.org/licenses/LICENSE-2.0  
 
THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE, 
MERCHANTABLITY OR NON-INFRINGEMENT. 
 
See the Apache Version 2.0 License for specific language governing permissions
and limitations under the License.
***************************************************************************** */
```

The author did not understand the purpose of the NOTICE file and if we'd integrate this file by hand, I'd request it to be completely removed, as nothing of the text has a place in the NOTICE file.

I admit though, that it is difficult to detect and override. Maybe notices should only done by hand?

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+
+                if (Files.isReadable(thirdpartynoticestxt)) {
+                    licenseText = "Parts of this work are licensed:\n" +
+                                  licenseText +
+                                  "\n\n" +
+                                  "Parts of this work are licensed:\n" +
+                                  readString(thirdpartynoticestxt);
+                } else {
+                    licenseText = licenseText;
+                }
```
matthiasblaesing: The `else` block is unnessary.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+                    files.forEach(p -> {
+                        if (p == module) return ;
+                        try {
+                            String relative = module.getParent().relativize(p).toString();
+                            boolean isDir = Files.isDirectory(p);
+                            ZipEntry ze = new ZipEntry(relative + (isDir ? "/" : ""));
+                            out.putNextEntry(ze);
+                            if (!isDir) {
+                                if (relative.equals("package.json")) {
+                                    out.write(packageJsonText.replace(targetDir.toString(), "").getBytes());
```
matthiasblaesing: This should be `getBytes("UTF-8")`.

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+        @Override
+        public String toString() {
+            return "LicenseDescription{" + "version=" + version + ", description=" + description + ", homepage=" + homepage + ", licenseKey=" + licenseKey + ", licenseText=" + licenseText + ", bundles=" + bundles + '}';
+        }
+
+    }
+    private static class LicenseUses {
+        private final String key;
+        private final String licenseText;
+        private final List<String> projects = new ArrayList<>();
```
matthiasblaesing: Accessing a private field from outside the class looks strange to me. Would you consider moving the declarations here to `public final`?

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
```diff
@@ -0,0 +1,295 @@
...
+        return Arrays.asList(licenseText.replaceAll("[ \n\r\t]+", " ").split(" "));
+    }
+    private static class LicenseDescription {
+        private final String name;
+        private final String version;
+        private final String description;
+        private final String homepage;
+        private final String licenseKey;
+        private final String licenseText;
+        private final List<String> bundles = new ArrayList<>();
```
matthiasblaesing: Same comment as 287.

jlahoda:
Thanks for the comments, Matthias! I'll work on them

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
jlahoda: I thought this should be there for the "Copyright (c) Microsoft Corporation. All rights reserved." line. But I could either drop this (if not needed at all), or use a hardcoded set of notices if needed (e.g. with just the single line or so.)

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
jlahoda: Yes, the more libraries we handle, the more difficult this gets - but still hopefully easier than doing all this by hand. (Esp. the VS Code extensions uses a lot of libraries - much more than I'd expect.)

nbbuild/misc/prepare-bundles/src/main/java/org/netbeans/prepare/bundles/PrepareBundles.java
matthiasblaesing: I read this as the normal license header for the application of licenses - you state the Copyright, the years you touched the file in, the copyright holder and the license. Nothing more - but it was only one of the entries, others were more suble but also held more information than necessary. I have not yet a good idea how to fix this...

junichi11:
@jlahoda It seems that Travis CI failed with the master branch after this was merged. Could you please take a look at it?

PR #1846, opened 2020-01-05
[NETBEANS-3428] FlatLaf: fixes for form designer

DevCharly:
Before:

image

After:

image

Before:

image

After:

image

The reason for the too small icons in the above images is that I took the screenshots running NB in Java 8 on a150% HiDPI screen.

### DIFF
--- a/java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
+++ b/java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
@@ -472,12 +472,13 @@ public class FormLoaderSettings implements HelpCtx.Provider   {
      * @return background color of the designer.
      */
     public java.awt.Color getFormDesignerBackgroundColor() {
-        Color defaultBackground = UIManager.getColor( "Tree.background" ); //NOI18N
-        if( null == defaultBackground )
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.background"); //NOI18N
+        if (defaultBackground == null)
+            defaultBackground = UIManager.getColor("Tree.background"); //NOI18N
+        if (defaultBackground == null)
             defaultBackground = Color.white;
         int rgb = getPreferences().getInt(PROP_FORMDESIGNER_BACKGROUND_COLOR , defaultBackground.getRGB());
         return new Color(rgb);        
-        
     }
 
     /**
@@ -497,9 +498,11 @@ public class FormLoaderSettings implements HelpCtx.Provider   {
      * @return color of the border of the designer.
      */
     public java.awt.Color getFormDesignerBorderColor() {
-        int rgb = getPreferences().getInt(PROP_FORMDESIGNER_BORDER_COLOR , new Color(224, 224, 255).getRGB());
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.borderColor"); //NOI18N
+        if (defaultBackground == null)
+            defaultBackground = new Color(224, 224, 255);
+        int rgb = getPreferences().getInt(PROP_FORMDESIGNER_BORDER_COLOR , defaultBackground.getRGB());
         return new Color(rgb);        
-        
     }
 
     /**
--- a/java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
+++ b/java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
@@ -472,7 +472,7 @@ public class FormLoaderSettings implements HelpCtx.Provider   {
      * @return background color of the designer.
      */
     public java.awt.Color getFormDesignerBackgroundColor() {
-        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.background"); //NOI18N
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.background"); //NOI18N
         if (defaultBackground == null)
             defaultBackground = UIManager.getColor("Tree.background"); //NOI18N
         if (defaultBackground == null)
@@ -498,7 +498,7 @@ public class FormLoaderSettings implements HelpCtx.Provider   {
      * @return color of the border of the designer.
      */
     public java.awt.Color getFormDesignerBorderColor() {
-        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.borderColor"); //NOI18N
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.borderColor"); //NOI18N
         if (defaultBackground == null)
             defaultBackground = new Color(224, 224, 255);
         int rgb = getPreferences().getInt(PROP_FORMDESIGNER_BORDER_COLOR , defaultBackground.getRGB());
--- a/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
+++ b/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
@@ -20,6 +20,7 @@
 package org.netbeans.swing.laf.flatlaf;
 
 import com.formdev.flatlaf.util.UIScale;
+import java.awt.Color;
 import javax.swing.UIManager;
 import org.netbeans.swing.laf.flatlaf.ui.FlatTabControlIcon;
 import org.netbeans.swing.plaf.LFCustoms;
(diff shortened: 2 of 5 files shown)


lkishalmi:
For me this looks good.

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
```diff
@@ -472,12 +472,13 @@ public void setPaletteInToolBar(boolean value) {
      * @return background color of the designer.
      */
     public java.awt.Color getFormDesignerBackgroundColor() {
-        Color defaultBackground = UIManager.getColor( "Tree.background" ); //NOI18N
-        if( null == defaultBackground )
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.background"); //NOI18N
+        if (defaultBackground == null)
+            defaultBackground = UIManager.getColor("Tree.background"); //NOI18N
+        if (defaultBackground == null)
```
eirikbakke: Which LAF is used in the form designer? Is it the same as the IDE? In that case, should the background color not be taken from the LAF rather than from a special constant? (Also, I wonder why "Tree.background" was used previously instead of the default JPanel background.)

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
```diff
@@ -497,9 +498,11 @@ public void setFormDesignerBackgroundColor(java.awt.Color value) {
      * @return color of the border of the designer.
      */
     public java.awt.Color getFormDesignerBorderColor() {
-        int rgb = getPreferences().getInt(PROP_FORMDESIGNER_BORDER_COLOR , new Color(224, 224, 255).getRGB());
+        Color defaultBackground = UIManager.getColor("nb.formdesigner.designer.borderColor"); //NOI18N
```
eirikbakke: Isn't this what the PROP_FORMDESIGNER_BORDER_COLOR property is for? Or does that have some other purpose?

Also, shouldn't this come from whichever LAF is being used to display form designer components? (As in my previous comment.)

Also, why is the new property using the word "designer" twice?

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
```diff
@@ -63,6 +67,12 @@
             "nb.multitabs.button.left.icon", FlatTabControlIcon.get(TabControlButton.ID_SCROLL_LEFT_BUTTON), // NOI18N
             "nb.multitabs.button.right.icon", FlatTabControlIcon.get(TabControlButton.ID_SCROLL_RIGHT_BUTTON), // NOI18N
             "nb.multitabs.button.rollover", true, // NOI18N
+
+            // Change some colors from ColorUIResource to Color because they are used as
+            // background colors for checkboxes (e.g. in org.netbeans.modules.palette.ui.CategoryButton),
+            // which in FlatLaf paint background only if background color is not a UIResource.
+            "PropSheet.setBackground", new Color(UIManager.getColor("PropSheet.setBackground").getRGB()),
+            "PropSheet.selectedSetBackground", new Color(UIManager.getColor("PropSheet.selectedSetBackground").getRGB()),
```
eirikbakke: Am I right that these changes (in FlatLFCustoms) are related to the LAF's appearance as a whole, i.e. not specifically related to the form designer?

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
DevCharly: As far as I know the form designer always uses the IDE LAF.
But it is possible to preview the form with other LAFs in new window. (right-click on panel: "Preview Design > LAF")

> In that case, should the background color not be taken from the LAF rather than from a special constant?

No sure what you mean. `UIManager.getColor()` gets the color from the LAF (if set).

`Tree.background` was probably used because it is white, while `Panel.background` is gray.
BTW `nb.formdesigner.designer.background` is currently not used. I've added it for completeness.

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
DevCharly: The class `FormLoaderSettings` manages the options that you can edit in "Options > Java > GUI Builder", but not all options in this class are implemented in the Options dialog. So it is not possible to change the background or border color in the Options dialog.

DevCharly: > Also, why is the new property using the word "designer" twice?

The idea was that `nb.formdesigner` is the module prefix and `designer` the component name within that module. But you're right. This is unnecessary. I'll remove it.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
DevCharly: Yes.

The controlFont is also used in other places. E.g. as fallback font in `HtmlRenderer`. The "Terminal" uses the size of controlFont to create its monospaced font for "Terminal" view and "Options > Misc. > Terminal > Preview".

The `PropertySheet` component is also used outside of the form designer.

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
eirikbakke: Oh, right, the constant is from the LAF. My main concern was that it would seem logical to take it from the JPanel background, rather than invent a new constant. But since the existing code doesn't do that, I suppose it makes sense to add a constant to preserve backwards-compatibility.

How come Tree.background is not dark for FlatLAF, though?

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
eirikbakke: Makes sense.

FlatLAF is adding a lot of new UIDefaults keys. Is there a reason why these were not needed to implement the other LAFs (e.g. Darcula, GTK, Aqua, Windows)? Is it related to the fact that FlatLAF is being implemented in a separate module, as opposed to in the o.n.swing.plaf module?

(Adding more UIDefaults keys seems fine... it just requires the usual extra care associated with API changes.)

eirikbakke: Also, is there a convention of putting constants at the end of org.netbeans.swing.plaf.LFCustoms for every key in used in UIDefaults?

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
eirikbakke: I don't see a change in the property sheet checkbox rendering in the before/after screenshots. Am I missing something?

DevCharly: The checkbox is in the designer palette, but it does not look like a checkbox because the check-icons are changed:

image

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
DevCharly: > Is there a reason why these were not needed to implement the other LAFs (e.g. Darcula, GTK, Aqua, Windows)?

Because the default values work for light themes. But look not so good for dark themes. Here are screenshots of Darcula and Dark Metal. The design border (red arrow) is way too light:

image

> Also, is there a convention of putting constants at the end of org.netbeans.swing.plaf.LFCustoms for every key in used in UIDefaults?

Don't know. Most of them can only be used in subclasses because they are protected. In case of FlatLaf, the values are in the properties files and constants are not used.

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
DevCharly: > How come Tree.background is not dark for FlatLAF, though?

It is dark and it is used.

I've first added UI value for the designer border color (because it was too bright) and thought that it is good idea to do the same for the designer background in case need to change it.

eirikbakke: Sounds reasonable.

java/form/src/org/netbeans/modules/form/FormLoaderSettings.java
eirikbakke: OK, I don't have strong opinions here...

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
eirikbakke: Ah! Tricky use of JCheckBox there...

eirikbakke:
Looks fine to me. Thanks for answering my questions!

Last, I noticed your comment: "The reason for the too small icons in the above images is that I took the screenshots running NB in Java 8 on a 150% HiDPI screen". Did that get fixed in this patch? (It seems so from the screenshots? Was that the CONTROLFONT property?)

PR #1848, opened 2020-01-05
[NETBEANS-3666] improve performance of multitabs on startup if many files are open

DevCharly:
This is an attempt to fix/workaround the startup performance problem if many files are open and multitabs are used (which is not the default).

There are actually two problems:

1. `getPreferredWidth()` is invoked very often from the multitabs table listener (~60000 times for 200 open files)
2. the text (usually a filename) is always HTML (but does not contain other HTML tags), which involves slow HTML parsing

This PR addresses only 2.

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
@@ -149,7 +149,7 @@ public class TabDataRenderer implements TableCellRenderer {
             // The HTML rendering makes startup slow (and CPU load high).
             // To workaround this, remove the leading "<html>" if text does not
             // contain HTML tags or entities.
-            String prefix = "<html>";
+            String prefix = "<html>"; // NOI18N
             if (text.startsWith(prefix)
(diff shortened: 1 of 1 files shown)


platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
```diff
@@ -141,6 +141,21 @@ int getPreferredWidth( Object value ) {
...
             }
+
+            // On startup, this method is invoked very often if many files are open
+            // (~60000 times for 200 open files).
+            // The text (usually a filename) always starts with "<html>",
+            // but does not contain other HTML tags.
+            // The HTML rendering makes startup slow (and CPU load high).
+            // To workaround this, remove the leading "<html>" if text does not
+            // contain HTML tags or entities.
+            String prefix = "<html>";
```
lkishalmi: //NOI18N at the end of the line please.

lkishalmi:
Seems to be pretty harmless.

lkishalmi:
Next time you can really do an amended forced push for such small changes, so I can just rebase and merge it.

eirikbakke:
Nice!

PR #1849, opened 2020-01-05
NETBEANS-3254 Remove allowAmbiguousCommands flag

lhochet:
Removes the jdk.lang.Process.allowAmbiguousCommands flag from netbeans.conf.

Instead defines a NETBEANS_MAVEN_COMMAND_LINE environment variable
local to the ProcessBuilder used to launch Maven, and pass a reference to that
variable on the cmd.exe command line.

With that change, the quote prefixing and suffixing of the Maven command
line prevented the command line from being properly launched so it was removed.
(Tested both with the embedded Maven 3.3.9 and an external Maven 3.6.3)


ebarboni:
Hi @lhochet, how to test that ?

lhochet:
Hello @ebarboni, 
I think these steps would demonstrate the fix:
1. create a test branch off master
1. clean build NetBeans
1. remove -J-Djdk.lang.Process.allowAmbiguousCommands=true  from etc/netbeans.conf
1. run NetBeans with a JDK post that was release in October or earlier this week (8 or above 11 -- my own test, 15+6)
1. create a new Maven Java Application, add a Main that prints hello world
1. try to build it
1. expected error: Cannot run program "cmd" (in directory "D:\dev\tests\mavenproject1"): Malformed argument has embedded quote: "C:\softs\apache-maven-3.6.3\bin\mvn.cmd" -Dmaven.ext.class.path=E:\netbeans\apache-netbeans\nbbuild\netbeans\java\maven-nblib\netbeans-eventspy.jar -Dfile.encoding=UTF-8 install
1. merge the branch for this PR (I think adding my fork of NetBeans as a remote will expose the branch)
1. rebuild NetBeans
1. run NetBeans again
1. retry to build the Maven project
1. double check by then trying to run the project

ebarboni:
works, allowAmbiguousCommands is only for this maven execution ?

neilcsmith-net:
@ebarboni it was certainly only added late in 11.2 because of reports of this bug with Maven on the JDK updates that had just been released. I don't know whether other things might be affected though - Maven was very obvious - but IDE wasn't tested without that option for very long!

Nice fix btw!

ebarboni:
@neilcsmith-net  thanks for comment.
@lhochet can you consider reverting the nb/ide.launcher/netbeans.conf  changes.( to avoid potential regression for others)
Fix for maven itself is ok to merge

lhochet:
@ebarboni sure, done.

It could still be of use to external processes (.bat/.cmd particularly) being launched from NetBeans and having double quote(s) within double quoted arguments.
eg. 
cmd /c ""path to maven with spaces/mvn.cmd" install"
or
cmd /c C:\softs\apache-maven-3.6.3\bin\mvn.cmd  install -Dfoo="bar team"
(though this later one will fail because it will be passed by Java as cmd.exe" /c "C:\softs\apache-maven-3.6.3\bin\mvn.cmd   -Dfoo=\"bar team\" install)

ebarboni:
maybe we should remove just after 11.3 release to see on longer period

neilcsmith-net:
@ebarboni that would be good and NetCAT for 12.0 should hopefully catch any other issues.

PR #1853, opened 2020-01-08
[NETBEANS-3673] Fix int spinner sizes in options dialog format:imports.

errael:
Simple fix using gui editor. Used IndentationPanel as example. Details in bug report.


lkishalmi:
Looks good to me.

PR #1855, opened 2020-01-08
InputStream.read(<array>) may read less bytes than requested - using …

jlahoda:
…readFully to read all requested bytes.

Unfortunately, I don't know how to test this. Basically, InputStream.read(<array>) is specified in such a way that if may read less bytes than requested. But the CLIHandler is not checking how many bytes where actually read, and so if the method reads a shorter chunk of bytes than expected, the method will continue and will interpret the next byte as a command/reply - but that byte is "random", and the server and client desynchronize and the connection effectively stops to work. Luckily, the instance of the InputStream here is a DataInputStream and it always had the readFully method, so we can use it.

### DIFF
--- a/platform/o.n.bootstrap/src/org/netbeans/CLIHandler.java
+++ b/platform/o.n.bootstrap/src/org/netbeans/CLIHandler.java
@@ -818,7 +818,7 @@ public abstract class CLIHandler extends Object {
                                     if (howMuch > outputArr.length) {
                                         outputArr = new byte[howMuch];
                                     }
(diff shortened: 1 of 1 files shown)


lkishalmi:
Do we have an actual issue/bug which relates to this? I'm just corious that this kind of fixes do not happen on somebody browsing the code.
Also maybe this has similar origins: https://issues.apache.org/jira/browse/NETBEANS-2510

JaroslavTulach:
Silly error. Sorry.

PR #1856, opened 2020-01-09
[NETBEANS-3674] change JTextComponent Keymap from EventDispatchThread

errael:
With this change, my plugin's JTextComponent's keymap listener is now invoked from the EDT. This removes the Swing threading violation.


errael:
I realize now that this change has a minor behavioral difference from the original code. The "if(arr.length>0)..." should not have been added. I don't actually think it matters, but...

How should I proceed? Another pull request?

matthiasblaesing:
@errael it might work to push into the branch backing this PR, but I would advise against that and would create a new PR and reuse the jira issue id.

PR #1857, opened 2020-01-10
[NETBEANS-3616]:Provided support of enable-preview hint and debug opt…

arusinha:
…ion in Java Single Source File

Jira Link : https://issues.apache.org/jira/browse/NETBEANS-3616


jlahoda:
I guess two questions for now:
-why parse the target address from the output, and not let the IDE listen and pass the port on which the IDE is listening to the debugee? Many if not all (Java) projects are doing that - see the JPDAStart classes.
-the handling of Processes feels suspicious - it is not clear why there are subclasses of Process. Moreover - if the compilation fails, will the user see the compilation error messages from javac? I see External Execution only supports a single Process, so this is obviously tricky.

arusinha:
Q: -why parse the target address from the output, and not let the IDE listen and pass the port on 
which the IDE is listening to the debugee
Ans: My initial approach was adding a specific port for debug, but later changed it to dynamic port as I wanted to avoid this extra configuration in file properties.

Q:  Moreover - if the compilation fails, will the user see the compilation error messages from javac? I see External Execution only supports a single Process, so this is obviously tricky.

Ans:  In output window compilation/run errors are getting captured.

e.g.,

 Error on Compilation. Please verify Source code and Compiler VM Options:
/scratch/arusinha/JEP-330-trial/NewCla.java:26: error: not a statement
        aaa
        ^
/scratch/arusinha/JEP-330-trial/NewCla.java:26: error: ';' expected
        aaa
           ^
Note: /scratch/arusinha/JEP-330-trial/NewCla.java uses preview language features.
Note: Recompile with -Xlint:preview for details.
2 errors

ebarboni:
unclear status, I guess for 12.0

arusinha:
hi @ebarboni , I am working on the review comments. Once done will it be possible to accommodate the PR in 11.3?

arusinha:
@jlahoda , I had incorporated all the review comments in last commit

ebarboni:
@arusinha it's a bit complicated for me. if @jlahoda can review fast and is ok I will merge.

java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/CompileProcess.java
```diff
@@ -0,0 +1,77 @@
...
+        List<String> compileCommandList = new ArrayList<>();
+        if (Utilities.isUnix()) {
+            compileCommandList.add("bash"); //NOI18N
+            compileCommandList.add("-c"); //NOI18N
+        }
+
+        Object compilerVmOptionsObj = fileObject.getAttribute(SingleSourceFileUtil.FILE_VM_OPTIONS);
+        String vmOptions = compilerVmOptionsObj != null ? (String) compilerVmOptionsObj : "";
+
+        compileCommandList.add(javacPath + " -g" + " " + vmOptions + " " + fileObject.getPath());
```
lahodaj: FWIW - I suspect this will not work very well for paths with spaces. Doing things like:
compileCommandList.add(javacPath);
compileCommandList.add("-g");
...

might work better in multiple ways.

java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/CompileProcess.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public Process setupProcess() {
+        File javaBinPath = new File(new File(System.getProperty("java.home")), "bin");  //NOI18N
+        String javaPath = javaBinPath.getAbsolutePath() + "//java";  //NOI18N
+
+        String javacPath = javaBinPath.getAbsolutePath() + "//javac";  //NOI18N
+        List<String> compileCommandList = new ArrayList<>();
+        if (Utilities.isUnix()) {
+            compileCommandList.add("bash"); //NOI18N
```
lahodaj: I wonder why use bash here - is that to split the whitespace-deliminated command line constructed below to parts? Better would be to avoid bash and have a single entry in the list per command/argument.

java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/DebugProcess.java
```diff
@@ -0,0 +1,84 @@
...
+class DebugProcess {
+
+    private static final Logger LOG = Logger.getLogger(DebugProcess.class.getName());
+    private final Pattern JVM_ARGS_PATTERN = Pattern.compile("(.*) (--source[ ]* [0-9]*)(.*)");  //NOI18N
+
+    public Process setupProcess(FileObject fileObject, String port) {
+        try {
+
+            List<String> commandsList = new ArrayList<>();
+            if (Utilities.isUnix()) {
```
lahodaj: (Similar comments as above.)

java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/JPDAStart.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+    }
+
+    public String getTransport() {
+        return TRANSPORT;
+    }
+
+    private static final class CopyReaderWriter implements Runnable {
```
lahodaj: This is unused, correct? Could be removed.

ebarboni:
Thks for review @lahodaj. @arusinha can you consider a new pr to adress the remarks.

arusinha:
@ebarboni ,As suggested i will create a separate PR addressing the reviews

PR #1858, opened 2020-01-10
[NETBEANS-3428] Update FlatLaf from 0.23 to 0.24 and other fixes

DevCharly:
Improvements and fixes thru FlatLaf 0.24:
- smooth scrolling (in the same way as already in NB for Windows LaF)
- fixed missing borders in table header (e.g. in "Watches" view)
- removed border from "Change Visible Columns" icon button in right-top corner of tables (e.g. in "Watches" view)
- progress bar in "Flat Dark" theme is now blue

Other fixes in NB code:
- "Flat Dark" theme: lighter gray color for editor/view borders and separators in main window
- ascending/descending icons in table header replaced with FlatLaf icons
- get rid of the focus indicator (black rectangle) in trees (HtmlLabelUI.java)
- fixed unreadable info text in "Find in Project" dialog (in dark theme)
- fixed too bright separators in "Find in Project" dialog (in dark theme) (BasicSearchForm.java)
- fixed wrong selection colors in `OutlineView` (e.g. in "Watches" view)

Before:
image

After:
image

Before:
image

After:
image

Before:
image
The orange arrows indicate issues that are not yet fixed. There are JButton's with EmptyBorder between the parenthesis and JButton's have a minimum width in FlatLaf, whats the reason for the empty space between the parenthesis.

After:
image

### DIFF
--- a/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
+++ b/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
@@ -26,7 +26,6 @@ import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.ItemSelectable;
-import java.awt.SystemColor;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.ItemEvent;
@@ -877,7 +876,6 @@ final class BasicSearchForm extends JPanel implements ChangeListener,
             c.insets = new Insets(5, 5, 5, 5);
             c.fill = GridBagConstraints.HORIZONTAL;
             JSeparator separator = new JSeparator(JSeparator.HORIZONTAL);
-            separator.setForeground(SystemColor.controlShadow);
             add(separator, c);
             row++;
         }
--- a/platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
+++ b/platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
@@ -269,24 +269,28 @@ public class OutlineView extends JScrollPane {
             defaultTreeActionListener, KeyStroke.getKeyStroke(KeyEvent.VK_ENTER, 0, false), JComponent.WHEN_FOCUSED
         );
 
-        final Color focusSelectionBackground = outline.getSelectionBackground();
-        final Color focusSelectionForeground = outline.getSelectionForeground();
-        outline.addFocusListener(new java.awt.event.FocusListener(){
-            @Override
-            public void focusGained(java.awt.event.FocusEvent ev) {
-                outline.setSelectionBackground(focusSelectionBackground);
-                outline.setSelectionForeground(focusSelectionForeground);
-            }
+        // toggle selection colors depending on whether table is focused or not
+        if (!UIManager.getLookAndFeel().getID().startsWith("FlatLaf")) {
+            final Color focusSelectionBackground = outline.getSelectionBackground();
+            final Color focusSelectionForeground = outline.getSelectionForeground();
+            outline.addFocusListener(new java.awt.event.FocusListener(){
+                @Override
+                public void focusGained(java.awt.event.FocusEvent ev) {
+                    outline.setSelectionBackground(focusSelectionBackground);
+                    outline.setSelectionForeground(focusSelectionForeground);
+                }
 
-            @Override
-            public void focusLost(java.awt.event.FocusEvent ev) {
-                outline.setSelectionBackground(SheetCell.getNoFocusSelectionBackground());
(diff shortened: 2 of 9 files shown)


lkishalmi:
Both the code and the UI look good to me, though @eirikbakke has beeter eyes for that.

platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
```diff
@@ -877,7 +876,6 @@ public void addSeparator() {
             c.insets = new Insets(5, 5, 5, 5);
             c.fill = GridBagConstraints.HORIZONTAL;
             JSeparator separator = new JSeparator(JSeparator.HORIZONTAL);
-            separator.setForeground(SystemColor.controlShadow);
```
eirikbakke: Not sure which separator this is. Does removing this make sense for other LAFs as well?

platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
```diff
@@ -269,24 +269,28 @@ public void keyReleased(KeyEvent e) {
...
-        final Color focusSelectionBackground = outline.getSelectionBackground();
-        final Color focusSelectionForeground = outline.getSelectionForeground();
-        outline.addFocusListener(new java.awt.event.FocusListener(){
-            @Override
-            public void focusGained(java.awt.event.FocusEvent ev) {
-                outline.setSelectionBackground(focusSelectionBackground);
-                outline.setSelectionForeground(focusSelectionForeground);
-            }
+        // toggle selection colors depending on whether table is focused or not
+        if (!UIManager.getLookAndFeel().getID().startsWith("FlatLaf")) {
```
eirikbakke: The diff is a bit confusing here. Is this just skipping a bunch of code for the FlatLaf case?

Why is the code not needed for FlatLaf? Shouldn't something rather be done so that focusSelectionBackground/focusSelectionForeground ends up being the right color?

(If I remember correctly, there might very well be some bugs in OutlineView around this code; I implemented workarounds in my NetBeans Platform application many years ago. If the changes suggested here are really workarounds for bugs, then maybe the bugs should be fixed instead.)

Maybe add a copy of the isFlatLaf() method for consistency with HtmlLabelUI. That will make it easier to grep for duplicated logic in the future.

eirikbakke:
This is really great work... thanks for doing this.

By the way, I have a made an improved SVG version of the "choose columns" icon that exists in the upper right-hand corner of some table components in these screenshots. I can commit that one later.

platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
DevCharly: This are the two horizontal separators in the "Find in Projects" dialog (see screenshots above). In Windows LAF there is no difference because default separator foreground is already controlShadow. Did not check other LAFs, but even if there is difference, this should be no problem. Its just a simple separator in a dialog. Default separator colors should work fine in dialogs.

platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
DevCharly: GitHub diff is not that good. In SmartGit it looks like this:
image

> Why is the code not needed for FlatLaf?

Because FlatLaf does this itself. With the NB code enabled it did not work anymore...

But in Windows LAF it works fine. Do not think that there is a problem in the NB code.

platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
eirikbakke: Makes sense.

platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
eirikbakke: > Because FlatLaf does this itself. With the NB code enabled it did not work anymore...

That would be a good comment to add in the code.

eirikbakke:
By the way, I saw you implemented smooth scrolling in the FlatLAF codebase, based on the IntelliJ implementation at https://github.com/JetBrains/intellij-community/blob/31e1b5a8e43219b9571951bab6457cfb3012e3ef/platform/platform-api/src/com/intellij/ui/components/SmoothScrollPane.java#L141-L185 . The Windows LAF smooth scrolling implementation is also based on work from the IntelliJ folks, but was based on an earlier IntelliJ commit ( https://github.com/JetBrains/intellij-community/commit/34b9dfd0585937c3731e06a89554d1dc86f7f235 ).

Do you think your newer implementation might be better in any way? If so, I might consider using it for the Windows LAF as well.

PR #1865, opened 2020-01-12
[NETBEANS-3428] FlatLaf: multi-tabs in FlatLaf style

DevCharly:
This PR implements FlatLaf style tabs for Multi-Tabs.

Multi-Tabs are used in main window for editors when at least one of the marked options in the below screenshot are enabled. Otherwise another tab control is used, which was implemented in PR #1771

image

Multi-Tabs internally use a `JTable` for tab rendering and therefore do not support UI delegates.
So it was necessary to extend Multi-Tabs for basic customizing. 

Everything changed by this PR is controlled thru new values in UI defaults (`UIManager.get(...)`. Those UI defaults are not set in other LAFs. So other LAFs work as before.

Each tab is painted as a simple filled rectangle and a "underline" bar. The "underline" bar placement depends on the tab placement. See screenshots below. There are separator lines between tabs if tab placement is "top" or "bottom". Hover effect is not supported by multi-tabs and can be implemented later.

Thru UI values it is possible to change all colors, tab insets, and height of "underline" bar. By setting underline height to zero it can be hidden.

Tab placement "top" and "multi-row tabs" enabled:
image

Tab placement "top" and "show parent folder name" enabled:
image

Tab placement "bottom":
image

Tab placement "left":
image

Tab placement "right":
image

The "slight blue tint" of the tabs area indicates that this tab group is active. If a view is active/focused, then the editor tabs became gray and the view group tabs get a "slight blue tint". This makes it easier to see the active/focused editor/view.
image

Same happens when editor area is split:
image

Light version with "same background color for files from same project" option enabled:
image

Some "before" screenshots for comparison:

image

image

image

This screenshot shows a problem of basic Multi-Tabs: Which editor is the active one?
image


eirikbakke:
Great, more flatLAF!

> The "slight blue tint" of the tabs area indicates that this tab group is active. If a view is active/focused, then the editor tabs became gray and the view group tabs get a "slight blue tint". This makes it easier to see the active/focused editor/view.
> This screenshot shows a problem of basic Multi-Tabs: Which editor is the active one?

Why does the entire tab group need to get a different color, and not just the focused tab? In the non-multi-tab case, both the Aqua and Windows 8 LAFs simply give the focused tab a different color. See screenshot:

which_tab_group_is_selected_indication

Is there something about the multi-tabs setting that makes this infeasible?

I also prefer the more compact tabs shown in https://user-images.githubusercontent.com/5604048/72223615-2239a080-3571-11ea-90fd-f185af00575b.png .

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ControlsToolbar.java
```diff
@@ -58,6 +58,7 @@ public void updateUI() {
     private void configure() {
         setFloatable( false );
         setFocusable( false );
+        setOpaque( false );
```
eirikbakke: Did you check to make sure this did not cause problems in other LAFs?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
```diff
@@ -63,15 +62,27 @@ public void stateChanged(ChangeEvent e) {
...
-        backGroundColors.add( new Color( 241, 255, 209 ) );
-        backGroundColors.add( new Color( 255, 225, 209 ) );
-        backGroundColors.add( new Color( 228, 255, 216 ) );
-        backGroundColors.add( new Color( 227, 255, 158 ) );
-        backGroundColors.add( new Color( 238, 209, 255 ) );
+
+        // load background colors from UI defaults if available
+        if (UIManager.getColor("nb.multitabs.project.1.background") != null) {
+            for (int i = 1; i <= 100; i++) {
+                Color color = UIManager.getColor("nb.multitabs.project." + i + ".background");
```
eirikbakke: Is this a gradient of some sort? Better to specify start and end colors, and interpolate, than have up to 100 properties. That will look better on HiDPI displays too (since you can actually get a smooth gradient rather than N hard steps).

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
```diff
@@ -48,6 +53,12 @@
  */
 public class TabDataRenderer implements TableCellRenderer {
 
+    private static final Insets tabInsets = UIManager.getInsets("nb.multitabs.tabInsets"); // NOI18N
+    private static final int underlineHeight = UIManager.getInt("nb.multitabs.underlineHeight"); // NOI18N
+    private static final Color underlineColor = UIManager.getColor("nb.multitabs.underlineColor"); // NOI18N
+    private static final Color inactiveUnderlineColor = UIManager.getColor("nb.multitabs.inactiveUnderlineColor"); // NOI18N
+    private static final Color activeBackground = UIManager.getColor("nb.multitabs.activeBackground"); // NOI18N
```
eirikbakke: As before, when adding UIManager keys, I assume that you feel confident that they new keys are needed, and that you've figured out what the naming convention is for them...

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
eirikbakke: And you probably don't need to do the interpolation yourself... there's Graphics2D.setPaint(new GradientPaint(...))

lkishalmi:
I like the blue tint of the entire tab group. With that, I find it easier to locate the focused component. That might be really subjective, however, when I discovered, I had that feeling. Maybe I'm getting older and my eye is not as pixel sharp as it used to be...

eirikbakke:
If the entire focused tab was made blue, it would be easier to spot. This screenshot looks great to me: https://user-images.githubusercontent.com/5604048/72223615-2239a080-3571-11ea-90fd-f185af00575b.png (except for the lack of tab borders in the "Projects"/"Files" and "Navigator" groups)

If you can see which _tab_ is active, then you obviously know which tab group is active.

Also, I suspect most people have just one editor tab group open--so there will just be a constant blueish rectangle on the screen all the time.

But of course, the people who do the work get to make the final call :-)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ControlsToolbar.java
DevCharly: Yes, tested with Window, Mac, Nimbus and Metal LaFs. `ControlsToolbar` is a package private class and only used in in `AbstractTabDisplayer`.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
DevCharly: Yes, I've searched in the multitabs modules for other keys, which mostly use "nb.multitabs." as prefix. So I used the same prefix.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
DevCharly: > Is this a gradient of some sort?

No, NetBeans has an option "same background color for files from same project" that paints tabs in different colors. Each open project gets its own color. The used colors are hardcoded into `ProjectColorTabDecorator` and work good with light LAFs, but not with dark LAFs. So I've added a way to get those colors from [UI defaults](https://github.com/DevCharly/netbeans/blob/d3c9d24efe5b8e48b19c91795c19e99354d8c9bd/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatDarkLaf.properties#L68-L76).

eirikbakke: Ah, makes sense.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
eirikbakke: Good.

PR #1867, opened 2020-01-12
NETBEANS-3686 Add spacing to the number appended to the session name …

lhochet:
…when debugging a program twice

Adds ' - '  between the project name and the number added when a project is debugged more than once


java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/LaunchingSessionProvider.java
```diff
@@ -99,7 +99,7 @@ static String findUnique (String sessionName) {
         for (i = 0; i < k; i++)
            if (!m.contains (Integer.valueOf(i)))
                break;
-        if (i > 0) sessionName = sessionName + i;
+        if (i > 0) sessionName = sessionName + " - " + i;
```
lkishalmi: //NOI18N

lhochet: added, thanks

PR #1868, opened 2020-01-13
[NETBEANS-3428] Fix the border of QuickSearch

lkishalmi:
image
image

I thought it would better fit a 1 sized border with background color than a one with the usual lighter gray component color.


lkishalmi:
@DevCharly @eirikbakke What do you think?

DevCharly:
LGTM 👍

eirikbakke:
Looks good.

PR #1870, opened 2020-01-13
Support for pattern matching in instanceof.

jlahoda:


lkishalmi:
Is this for 11.3 or 12.0?

jlahoda:
Preferably 11.3 if still possible.

neilcsmith-net:
Not that he's announced freeze yet, but adding @ebarboni to reviewers!  Agree that landing in 11.3 preferable to 12.0 though, for various reasons.

PR #1871, opened 2020-01-14
[NETBEANS-3115] Resolve third-party source against open Gradle projects.

lkishalmi:
I've started a poll on the dev maillist if I need to implement a switch for this or not.

### DIFF
--- a/groovy/gradle.java/src/org/netbeans/modules/gradle/java/queries/OpenGradleProjectForBinary.java
+++ b/groovy/gradle.java/src/org/netbeans/modules/gradle/java/queries/OpenGradleProjectForBinary.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.gradle.java.queries;
+
+import java.io.File;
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
The poll was a bit short. So are maven and gradle implementations aligned/behaving similar?

neilcsmith-net:
@matthiasblaesing don't Maven and Gradle themselves behave differently in this respect though?

matthiasblaesing:
How do they behave different? They define binary dependencies and these can have sources attached. My scenario I tested:

- I create a project `library` (just a `DemoLib` class with a `dummyMethod`)
- I `maven -Prelease-profile install`ed version `1.0`, `2.0` and `3.0-SNAPSHOT`
- I created an application project that depends on library version `3.0-SNAPSHOT`, both projects are open in netbeans
- I invoke "Navigate -> Go to Source" on `demoLib.dummyMethod()`

At this point netbeans takes me to the code in the project that I have open.

- I change the dependency to `2.0`
- I invoke "Navigate -> Go to Source" on `demoLib.dummyMethod()`

At this point the source code of the source jar is opened and shown.

- I checkout version `2.0` of `library`
- I invoke "Navigate -> Go to Source" on `demoLib.dummyMethod()`

At this point I'm taken to the source of the open project.

From my perspective the behavior of the Integration of Maven in NetBeans in completely sane: If you have the source for a maven project open _and_ the version of the project matches the version of the dependency then the project source code is used, else the source jar is used to show the source.

If you don't do this, you will get misleading results. I once accidentally chased a bug that I misdiagnosed, because the lines in the source code did not match the one in the stack trace because I had the wrong version opened.

IMHO this would be a sane implementation for gradle as well.

neilcsmith-net:
> From my perspective the behavior of the Integration of Maven in NetBeans in completely sane: If you have the source for a maven project open and the version of the project matches the version of the dependency then the project source code is used, else the source jar is used to show the source.

Wholeheartedly agree!  Lots of Maven in NetBeans is completely sane. :smile:  I just also had in mind the comments at https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local  This might reflect a need for a somewhat different approach between the two?  But I don't use Gradle often enough to not need a refresher read each time so may be wrong - was actually reading that link yesterday.  The important thing is, as you say, the sources match exactly what you're compiling against.

lkishalmi:
With this code in Gradle works the same as Maven in NetBeans. The poll is still open, if anyone would really debate on adding this behind a switch, I'd implement that in another PR. Though there are no response so far.

matthiasblaesing:
Thank you for the confirmation. I think having maven and gradle behave identical or at least very similar make it more consistent to work with NetBeans.

PR #1872, opened 2020-01-14
[NETBEANS-3695] - not proper to use a static variable through a refer…

BradWalker:
…ence

It's really not proper to use a static variable through a reference because they can't be derived. The proper thing to do is use them with a Class name..

   [repeat] /home/bwalker/src/netbeans/java/java.hints.ui/src/org/netbeans/modules/java/hints/spiimpl/refactoring/ConfigurationsComboModel.java:175: warning: [static] static variable should be qualified by type name, KeyEvent, instead of by an expression
   [repeat]             if (ke.getKeyCode() == ke.VK_ENTER || ke.getKeyCode() == ke.VK_ESCAPE) {
   [repeat]                                      ^
   [repeat] /home/bwalker/src/netbeans/java/java.hints.ui/src/org/netbeans/modules/java/hints/spiimpl/refactoring/ConfigurationsComboModel.java:175: warning: [static] static variable should be qualified by type name, KeyEvent, instead of by an expression
   [repeat]             if (ke.getKeyCode() == ke.VK_ENTER || ke.getKeyCode() == ke.VK_ESCAPE) {
   [repeat]

This change fixes that..

### DIFF
--- a/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/LinkButton.java
+++ b/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/LinkButton.java
@@ -119,10 +119,10 @@ public class LinkButton extends JButton implements MouseListener, FocusListener 
     }
     
     public void setColors(Color linkColor, Color linkInFocusColor, Color mouseOverLinkColor, Color visitedLinkColor) {
-        this.linkInFocusColor = linkInFocusColor;
-        this.linkColor = linkColor;
-        this.mouseOverLinkColor = mouseOverLinkColor;
-        this.visitedLinkColor = visitedLinkColor;
+        LinkButton.linkInFocusColor = linkInFocusColor;
+        LinkButton.linkColor = linkColor;
+        LinkButton.mouseOverLinkColor = mouseOverLinkColor;
+        LinkButton.visitedLinkColor   = visitedLinkColor;
     }
 
     private void init() {
--- a/enterprise/glassfish.javaee/src/org/netbeans/modules/glassfish/javaee/RunTimeDDCatalog.java
+++ b/enterprise/glassfish.javaee/src/org/netbeans/modules/glassfish/javaee/RunTimeDDCatalog.java
@@ -614,9 +614,9 @@ public class RunTimeDDCatalog extends GrammarQueryManager implements CatalogRead
         Enumeration en = ctx.getDocumentChildren();
         while (en.hasMoreElements()) {
             Node next = (Node) en.nextElement();
-            if (next.getNodeType() == next.DOCUMENT_TYPE_NODE) {
+            if (next.getNodeType() == Node.DOCUMENT_TYPE_NODE) {
                 return null; // null for web.xml specified by DTD
-            } else if (next.getNodeType() == next.ELEMENT_NODE) {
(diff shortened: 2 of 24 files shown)


ide/team.commons/src/org/netbeans/modules/bugtracking/commons/LinkButton.java
```diff
@@ -119,10 +119,10 @@ public LinkButton() {
...
     
     public void setColors(Color linkColor, Color linkInFocusColor, Color mouseOverLinkColor, Color visitedLinkColor) {
-        this.linkInFocusColor = linkInFocusColor;
-        this.linkColor = linkColor;
-        this.mouseOverLinkColor = mouseOverLinkColor;
-        this.visitedLinkColor = visitedLinkColor;
+        linkInFocusColor = linkInFocusColor;
+        linkColor = linkColor;
+        mouseOverLinkColor = mouseOverLinkColor;
+        visitedLinkColor   = visitedLinkColor;
```
mklaehn: Seems wrong and unrelated to the described issues

mklaehn: Well, maybe not unrelated but wrong.

mklaehn: Short answer something like `LinkButton.linkInFocusColor = linkInFocusColor` for these four.

Long answer: If you simply remove the `this.` you essentially assign the method parameter to itself. Highlighting in NetBeans Editor will show that as well.

BradWalker: No problem. It's now changed.. Thanks for the feedback..

eirikbakke:
This change touches a lot of different modules--have you tested each affected feature to make sure no problems are introduced?

BradWalker:
I have used the changes as part of my day-to-day use of Netbeans. It works correctly and the C/C++ module continued to run with no problem.  Most of the changes are pretty benign: instead of referencing a class static field through a reference we use the properly qualified type..

For example,

```
            if (next.getNodeType() == **next**.DOCUMENT_TYPE_NODE) {
            if (next.getNodeType() == **Node**.DOCUMENT_TYPE_NODE) {
```

Notice how I'm only referencing the class field..

eirikbakke:
OK, never mind, the changes seem safe. Though in general, it's best to do these kinds changes only once you're actually working on, and testing, a given piece of code.

It's very easy to introduce errors (as with LinkButton), and code should assumed to be "broken unless tested".

BradWalker:
I mostly agree with your comments..  My focus is on trying to get a "clean" build with a very minimal amount of warnings..

eirikbakke:
Yes, I think cleaning up the build output is an honorable cause. It's just more work than it seems, because of the need to test every piece of code that is changed. For trivial changes you can sometimes just stare at the diff and think "what could possibly go wrong?"--and yet, somtimes it does.

BradWalker:
Agreed.. That's the risk we take otherwise this code will never get cleaned up.. Also, I've noticed a LOT of places where there are "bugs" as a result of me just looking at the code when I make the changes.

Thanks for merging this!

PR #1876, opened 2020-01-16
Fixing the VS Code extension build by using new TypeScript to build it.

jlahoda:


JaroslavTulach:
The PR seems to fix the vscode build problem. Merging.

PR #1878, opened 2020-01-17
Adding necessary add-exports to the Java LSP server; removing the add…

jlahoda:
… exports for javadoc, which is no longer needed.

(Basically a small cleanup to avoid unnecessary text in the standard error output.)


ebarboni:
looks good to me merging

PR #1879, opened 2020-01-17
[NETBEANS-3647] Update FtpClient.java - compatibility with IIS

oli0044:
Allow non-strict interpretation of RFC 959 to enhance compatibility with some FTP servers (e.g. Microsoft IIS FTP).

cf. https://issues.apache.org/jira/browse/NETBEANS-3647

### DIFF
--- a/php/php.project/src/org/netbeans/modules/php/project/connections/ftp/FtpClient.java
+++ b/php/php.project/src/org/netbeans/modules/php/project/connections/ftp/FtpClient.java
@@ -158,7 +158,7 @@ public class FtpClient implements RemoteClient {
             if (LOGGER.isLoggable(Level.FINE)) {
                 LOGGER.log(Level.FINE, "Connecting to {0} [timeout: {1} ms]", new Object[] {configuration.getHost(), timeout});
             }
-            ftpClient.setStrictReplyParsing(false);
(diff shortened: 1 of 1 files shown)


oli0044:
Hi Matthias,

thanks for the quick reply. 

I tried to change all your requests:
 1. Added my full name and public email address to the GitHub profile
 2. Commited the changes once more by putting the JIRA issue id and title in the commit message as well as the description.

Hope that solves it.

Greetings from Germany

Oli4

matthiasblaesing:
Thank you Oliver, the problem is now, that there are two commit (the second with valid author information). If you don't want to use a "real" git client, I suggest to close this PR and do it again.

For the commit message I missed to describe the "best" format. The jira identifier should be placed in brackets, the first line should be the summary, details in followup lines. In this case I suggest:

```
[NETBEANS-3647] Update FtpClient.java - compatibility with IIS

FTP upload stops when uploading to IIS FTP reporting a Truncated server reply: '550 '.
The broken behavior can be worked around by configuring the ftp client to be
lenitent when parsing server replies.
```

oli0044:
ok, thanks Matthias for the tips. sorry, it's my first time, so I'm a still bit clumsy.

Will close it and reopen it like you suggested.

PR #1883, opened 2020-01-18
[NETBEANS-3428] FlatLaf: improving editor and view tabs in main window

DevCharly:
This PR improves PR #1771

Editor and View tabs:
- tab separators
- tab text moved up 1px
- smaller tab insets
- minimize icon improved

View tabs only:
- minimized space between tab text and close button
- center tab text if close button is hidden
- make sure that as much tab text as possible is shown by dynamically reducing left/right margins
- avoid empty tabs (truncate text if necessary)

New:

image

Old (from PR #1771):

image

The tab text in views is now centered (was left aligned) because left aligned text looks horrible with tab separators.
image

When horizontal space becomes rare, the left/right tab margins are reduced.
Empty tabs (no tab text painted) does no longer occur.
image


lkishalmi:
Jut a note the close button on tab panels is touching the text, really visible when the hover icon is on a 2-3px distance would be better.
image
The different text-icon distance can be seen above.

PR #1884, opened 2020-01-18
[NETBEANS-3713] FlatLaf: progress bar improvements

DevCharly:
- progress bar in status bar now always has large height (has changed height before depending on whether text is painted in progress bar)
- cancel button replaced with vector icon
- fixed color of separator in "Processes" popup
- remove separator of last row in "Processes" popup if process was removed
- use preferred height for progress bars in "Processes" popup

Before ("Processes" popup shown):
image

Before, separator was not removed after background scanning finished:
image

After:
image

"Processes" popup:
image

### DIFF
--- a/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
+++ b/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
@@ -72,6 +72,11 @@ public class FlatLFCustoms extends LFCustoms {
             "nb.multitabs.button.right.icon", FlatTabControlIcon.get(TabControlButton.ID_SCROLL_RIGHT_BUTTON), // NOI18N
             "nb.multitabs.button.rollover", true, // NOI18N
 
+            // for module progress.ui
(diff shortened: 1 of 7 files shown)


humphreygao:
hello,  I use NetBeans 11.3-beta3 in Windows 10 OS and find that progress bar leaves vertical stripes during scrolling. But I am not sure whether it is common phenomenon.
Here are pictures(All in FlatLaf Look and Feel): 
blue stripe
blue stripe  light
big progress bar light

DevCharly:
@humphreygao this occurs only on HiDPI screens at 125%, 150%, 175% or 225% scaling when running NetBeans in a Java 9 or later VM.

It is fixed in FlatLaf 0.27 in commit https://github.com/JFormDesigner/FlatLaf/commit/0910bd23c43e26361339d541437ed303e21dde71, but NetBeans uses 0.26.

Should be fixed in NB 12.
For 11.3 it is probably too late and FlatLaf 0.27 contains too many changes to be used in NB 11.3.

@ebarboni @lkishalmi if there is a need to fix this in NB 11.3, I could build a FlatLaf 0.26.1 release that contains only the single [commit](https://github.com/JFormDesigner/FlatLaf/commit/0910bd23c43e26361339d541437ed303e21dde71) that fixes the issue?

humphreygao:
@DevCharly Ok, thank you for your reply.

ebarboni:
Hi would be better to wait for 12. this is not critical or blocker. and we are in voting phase

PR #1886, opened 2020-01-19
Update to Groovy 2.5.9 and enable groovy editor tests

svenreimers:
Enables tests for groovy.editor on travis
Fixed NETBEANS-1150
Fixed NETBEANS-3668


svenreimers:
There is one small problem with one completion test. Seems to be caused by the upgrade to 2.5.9. Looking into the root cause right now - I am not sure what the correct assumption for the completion is so one possibility to get this PR in is to disable the test because it is related to Grails, which we do not support at the moment.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/inference/TypeInferenceVisitor.java
```diff
@@ -194,4 +193,29 @@ private static boolean sameVariableName(ASTNode node1, FieldNode node2) {
                 && ((VariableExpression) node1).getName().equals(node2.getName());
     }
 
+    private ClassNode deriveExpressonType(Expression expression) {
```
mario-s: What abbout a single return statement here? IMHO those are too many. It's also a bit inconsistent with the previous method (visitDeclarationExpression).

ebarboni:
hi, sorry will not merge because of the test and pending discussion

svenreimers:
Discussion closed. Fix applied. Test removed. Hoping all will green...

ebarboni:
hi @svenreimers, is the https://issues.apache.org/jira/browse/NETBEANS-1150 fixed now ?

svenreimers:
Yes...Totally forgot about this...Thanks

PR #1887, opened 2020-01-19
Use correct constant to fetch welcome screen topbar foreground

matthiasblaesing:
Commit 5fb6012a7d2bfd233b4b9ba8b9b3fbc6d98e1171 moved the definition
of the foreground color into the resource bundle. Instead of the correct
constant COLOR_TAB_FOREGROUND COLOR_TAB_BACKGROUND is referenced.

The consequence of the wrong constant is an unreadable welcome screen
on LAFs, that don't override the colors.


matthiasblaesing:
The problem can be observed in this screenshot: https://github.com/apache/netbeans/pull/1826#issuecomment-575907403

lkishalmi:
This one is trivial. Thank you!

PR #1888, opened 2020-01-19
[NETBEANS-3428] FlatLaf: optimized editor and view tabs in main window

DevCharly:
Fix for this: https://github.com/apache/netbeans/pull/1883#issuecomment-576024263

Editor and View tabs:
- 2px larger space between tab text and close icon
- close icon moved up 1px so that it is not on the same baseline as the tab text

Multi-tabs:
- tab text moved up 1px so that it is at the same baseline as View tabs


ebarboni:
looks good merging.

PR #1889, opened 2020-01-20
[NETBEANS-3717] Allow LaF to change the color of the selected item in Code Completion

lkishalmi:
Some Look and Feels use different colors for displaying selected and focused list cells. FlatLaf is one of them. This little change allows the LaF to control the selected and by that focused item in the code completion list. It could be confusing to decide which item is selected, especially on two item lists.

Before:

CompletionFlatLafBefore

After:
CompletionFlatLafAfter


ebarboni:
@lkishalmi ok for me merge at will

PR #1890, opened 2020-01-21
[NETBEANS-3693] Create only one instance of javac

Akshay-Gupta-Oracle:
[NETBEANS-3693] Create only one instance of javac for parsing multiple files when refactored


java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -415,9 +422,28 @@ private void parseImpl(
                     break;
                 default:
                     init (snapshot, task, false);
+                    DiagnosticListener<JavaFileObject> diagnosticListener;
```
lahodaj: I was hoping this could work in such a way that if the processing of the files does not fit into memory, it would fall back to the slow "one javac per file". It is not as good as current nb-javac, but I hope it would work reasonably OK. Would also be good to have some tests for that.

java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -181,8 +182,8 @@
     private FileObject root;
     //ClassPaths used by the parser
     private ClasspathInfo cpInfo;
-    //Count of files the parser was created for
-    private final int sourceCount;
+    //XXX: Count of files the parser was created for
```
lahodaj: Would probably be good to cleanup the comment here (i.e. adjust it to the current state).

java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -582,24 +608,37 @@ Phase moveToPhase (final Phase phase, final CompilationInfoImpl currentInfo,
...
+                        CompilationUnitTree put = parsedTrees.put( oneFileTree.getSourceFile().getName(), oneFileTree);
+                    }
+                    unit = trees.iterator().next();
+                    }
                 }
-                CompilationUnitTree unit = it.next();
+               
                 currentInfo.setCompilationUnit(unit);
-                assert !it.hasNext();
+                //assert !it.hasNext();
```
lahodaj: Probably just delete the assert.

Akshay-Gupta-Oracle:
> Left some initial comments. Please note that at least the "org.netbeans.modules.java.source.parsing.JavacParserTest" appears to be failing.

There is one test case failing - testMultiSourceVanilla
Which has the statement - "assertFalse(Objects.equals(storedJLObject, jlObject));" which I think will not pass when only one instance of javac is created.

lahodaj:
> > Left some initial comments. Please note that at least the "org.netbeans.modules.java.source.parsing.JavacParserTest" appears to be failing.
> 
> There is one test case failing - testMultiSourceVanilla
> Which has the statement - "assertFalse(Objects.equals(storedJLObject, jlObject));" which I think will not pass when only one instance of javac is created.

Right - but that means the test needs to be fixed (or removed), correct? I.e. we cannot have failing tests, I assume?

Akshay-Gupta-Oracle:
> > > Left some initial comments. Please note that at least the "org.netbeans.modules.java.source.parsing.JavacParserTest" appears to be failing.
> > 
> > 
> > There is one test case failing - testMultiSourceVanilla
> > Which has the statement - "assertFalse(Objects.equals(storedJLObject, jlObject));" which I think will not pass when only one instance of javac is created.
> 
> Right - but that means the test needs to be fixed (or removed), correct? I.e. we cannot have failing tests, I assume?

Yes, will do that.

lkishalmi:
@jlahoda or @lahodaj as the 12.0 merge window is open, can you review the changes?

jlahoda:
> @jlahoda or @lahodaj as the 12.0 merge window is open, can you review the changes?

Currently, without nb-javac, refactoring run file-by-file (one javac instance per file), which increases chances it fits into the memory. This patch changes that to one javac instance for all files in the refactoring - that significantly improves the speed, but lowers the chances the processing will fit into the memory. I still think a middle ground - try to fit all into the memory, and run fast if it fits, but fallback to file-by-file processing if it doesn't - is the best combination. And, unless I am mistaken, this patch does not provide that.

Akshay-Gupta-Oracle:
@jlahoda 
Does the issue in this PR(Making multiple instances of javac for parsing multiple files) is addressed in this other PR #2013  ??
Or should I keep it open??

jlahoda:
> @jlahoda
> Does the issue in this PR(Making multiple instances of javac for parsing multiple files) is addressed in this other PR #2013 ??
> Or should I keep this one open??

I don't think there is a significant overlap between these PRs - this PR is (as far as I can tell) speeding up refactoring. #2013 is fixing the refactoring.java tests+adding them into the Travis configuration.

neilcsmith-net:
@ebarboni probably better to stick with what we've been testing with and push this back to 12.1?

lahodaj:
I guess I was hoping the processing would adjust dynamically (i.e. the processing would switch to file-by-file if it would run out of memory).

Anyway - my proposal is to add an option to enable this and push (after master opens), so that we can test in practice.

lahodaj:
By an option, I mean a command line option like -J-Djava.enable.single.javac=true; which can be tested using Boolean.getBoolean("java.enable.single.javac").

neilcsmith-net:
What's the status on this one?  @lahodaj safe to merge for 12.1?

lahodaj:
Sorry, but the builds say there is some problem with the patch - could you please investigate? Thanks!

neilcsmith-net:
At this point in time, think we need to bump this to 12.2.

PR #1892, opened 2020-01-21
[TRAVIS] check platform build

ebarboni:


ebarboni:
reproducing the way cluster plaftorm is produced on release phase.

PR #1893, opened 2020-01-21
[NETBEANS-2604] move o.apache.commons.codec to platform cluster

ebarboni:
Hi,
 this is a follow up on NETBEANS-2604 because of 
platform/o.apache.commons.io requiering codec on ide/o.apache.commons.codec.

On a pure platform build it will break. So moving to platform.


eirikbakke:
Yes, had I noticed this dependency I would have included this into the SVG support patch. So given that we concluded it was OK to move commons.io into the platform cluster, this change should be uncontroversial as well.

JaroslavTulach:
Should this process continue, we'll end up delivering all Apache libraries as part of Apache NetBeans Platform ;-)

PR #1897, opened 2020-01-24
[NETBEANS-3745] Clean up abandoned user dirs

lkishalmi:
This one is pretty trivial.


lkishalmi:
Janitor1

ebarboni:
@lkishalmi I put in beta 2.

neilcsmith-net:
@lkishalmi any way to only run this on first launch, or offer the option to mute rather than remove?  This is rapidly becoming my least favourite feature of beta2.  cc/ @geertjanw

GeertjanWielenga:
Would indeed be an idea to have it disabled by default and require a setting in netbeans.conf and/or Options window to enable it. The feature itself is very useful but also potentially very noisy in its current default state.

neilcsmith-net:
@geertjanw I agree with the idea of having it enabled by default, because for many users that's right (at least once), but there really needs to be a way to tell it to shut up - I get what you were complaining about at FOSDEM now! :smile:

@lkishalmi or maybe we use similar logic as https://github.com/apache/netbeans/blob/master/platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java#L331 to only run this on _second_ run of IDE?  I assume there's no chance of overlap with import upgrade task with this - ie. removing a userdir in process of being imported?

ebarboni:
I did not notice the annoyance but any PR to solve welcome.

lkishalmi:
I'm going to work something out for that.

lkishalmi:
See #1925

PR #1898, opened 2020-01-25
Fix startup warnings from netbeans 11.3

matthiasblaesing:
Analysing the startup warnings from netbeans 11.3 revealed some problems:

- we have colliding positions/missing positions in the config file system (mimetype registrations)
- missing code base passing to InstalledFileLocator
- missed updates for library definitions, that don't anymore match the included files

I think this is a candidate for 11.3 as it fixes real problems (broken library definitions) and in general I would expect a release to be as warning free as possible.

@lkishalmi @jlahoda @Chris2011 I'd like you to have a look at this as you worked in the area/authored the code
@ebarboni please see see if you aggree to consider this for 11.3

### DIFF
--- a/enterprise/servletjspapi/nbproject/project.properties
+++ b/enterprise/servletjspapi/nbproject/project.properties
@@ -16,7 +16,9 @@
 # under the License.
 
 is.autoload=true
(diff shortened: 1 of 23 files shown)


ebarboni:
merging for beta 2

PR #1901, opened 2020-01-26
Make groovy-all.xml parseable.

lkishalmi:
This shall go into 11.3


lkishalmi:
Well, the catch is from @JaroslavTulach

PR #1902, opened 2020-01-28
[NETBEANS-3749][NETBEANS-3757]: fixing behavior of LSP-based code com…

jlahoda:
…pletion w.r.t. trigger characters and printing text

Affects, for example, TypeScript editing. Pressing Enter in the code completion may lead to a garbled output; and the code completion is not following the trigger characters sent from the server.


ebarboni:
refreshing travis build. will integrate if ok

PR #1904, opened 2020-01-29
TextMate lexer should listen on SystemFS, so newly added grammars wor…

jlahoda:
…k almost immediatelly.

This is especially useful when ergonomics enables new modules, like the typescript.editor.

Might be nice for 11.3, but not absolutely critical, I think.


ebarboni:
merging for beta2

PR #1908, opened 2020-01-31
NETBEANS-3777 Use pom type instead of jar for bom artefacts

ebarboni:
fix for missing dependencies  in springframework

### DIFF
--- a/java/libs.springframework/src/org/netbeans/libs/springframework/spring-framework300.xml
+++ b/java/libs.springframework/src/org/netbeans/libs/springframework/spring-framework300.xml
@@ -68,7 +68,7 @@
                 org.springframework:spring-context-support:3.2.7.RELEASE:jar
                 org.springframework:spring-core:3.2.7.RELEASE:jar
                 org.springframework:spring-expression:3.2.7.RELEASE:jar
-                org.springframework:spring-framework-bom:3.2.7.RELEASE:jar
+                org.springframework:spring-framework-bom:3.2.7.RELEASE:pom
                 org.springframework:spring-instrument:3.2.7.RELEASE:jar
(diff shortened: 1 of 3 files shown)


blackleg:
@ebarboni Could you add the spring-framework-bom dependency to the 5.2.2.RELEASE version? https://github.com/apache/netbeans/blob/master/java/libs.springframework/src/org/netbeans/libs/springframework/spring-framework500.xml

ebarboni:
@blackleg, it was not part of the dependencies for 5.2.2, is it not working without bom ?

blackleg:
>  is it not working without bom ?
How can I test if it is working?

ebarboni:
You need to compile NetBeans. (I'm not javaee,spring expert)

But 5.x support is broken on ant and maven in NetBeans.

Some removed class are used in project setup (dispatcher-servlet).
Cannot find class [org.springframework.web.servlet.mvc.support.ControllerClassNameHandlerMapping

ebarboni:
make it for all version

lkishalmi:
Does it really work? I'm assuming this is for Ant projects. AFAIK the list of jar-s is provided instead of the bom, however if the bom is working, then probably we shall not set the list of jars.

ebarboni:
This only fix the non downloadble dependencies for maven. Not sure the bom take all the other dependencies. 

maven and ant NetBeans created web application cannot be run with 5.x. (bad conf file)

blackleg:
> maven and ant NetBeans created web application cannot be run with 5.x. (bad conf file)

Is it a blocking for the 11.3 release?

ebarboni:
Will merge.
@blackleg would be good for 5.X to have a workaround saying how to modify config file to get it work
NETBEANS-3558

PR #1909, opened 2020-01-31
[NETBEANS-3286] Naive fix for CME in Call Hierarchy

lkishalmi:
I've accidentally bumped into this one and found a reproducible test case, so I've tried to fix it.
 
Well it is a real naive fix, cacthing and handlinng ConcurrentModificationException. Maybe the affected hash map could be a concurrent one, then that could be better, or we could dig up the root cause why the concurrency modifications happen at first place. I leave that to @jlahoda 

However this patch seems to work and harmless. We have 5 issues in JIRA related to this one, so if it is not too late let put it into 11.3 (beta2 or rc1)


ebarboni:
merging for 11.3 next milestone

PR #1910, opened 2020-02-01
Only initialize the LSP server once; prevent NPE in LSP code completion.

jlahoda:
Two small (in size)/big (in effect) fixes for the LSP client - avoid duplicate initialization (while waiting for the server to come up), and prevent NPE in code completion when no trigger characters are specified. These were unfortunate oversights/mistakes.

If there's any chance on putting this into 11.3, that would be absolutely awesome.


ebarboni:
making it for 11.3 next milestone

PR #1915, opened 2020-02-03
[NETBEANS-3790] - eliminate all StringBufferInputStream deprecations

BradWalker:
eliminate all references to the following deprecation:

   [repeat] /home/bwalker/src/netbeans/platform/openide.explorer/src/org/openide/explorer/propertysheet/SheetTable.java:50: warning: [deprecation] StringBufferInputStream in java.io has been deprecated
   [repeat] import java.io.StringBufferInputStream;
   [repeat]


jlahoda:
I am afraid the replacement of StringBufferInputStream with StringReader does not look quite right - the API client asked for an InputStream, so we should not return them a Reader (they could ask for a Reader if that is what they need). Not completely sure what is the best replacement, but new ByteArrayInputStream(html.getBytes()) might be sensible (will use the platform's encoding, which feels like a sensible/defendable thing to do). There might be some compatibility impact, but probably very small/negligible.

PR #1919, opened 2020-02-04
[NETBEANS-3739] FlatLaf: fix missing versioning colors for modified/ignored files in FlatLaf Dark

DevCharly:
See https://issues.apache.org/jira/browse/NETBEANS-3739 for background information about the fix.

Before:
image

After:
image

### DIFF
--- a/platform/openide.awt/src/org/openide/awt/HtmlLabelUI.java
+++ b/platform/openide.awt/src/org/openide/awt/HtmlLabelUI.java
@@ -408,7 +408,7 @@ class HtmlLabelUI extends LabelUI {
 
         if (!replace) {
             dif = difference(fg, bg);
-            replace = dif < 80;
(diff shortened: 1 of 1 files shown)


ebarboni:
Hi as per @AlexFalappa  suggestion on #1918, it would be good to have the rendering on Nimbus, FlatLaf light, Metal. 
Because changes are in openide.awt

lkishalmi:
A good catch! Thank you!
Though I do not know why we are doing such overrides at all, in the first place...
Fixing that is for another time.

DevCharly:
FlatLaf Light:
image

Windows LAF:
image

Nimbus:
image

Metal:
image

ebarboni:
Thanks. Just to be sure NB112testCopy1.java is ignored in version control ?

DevCharly:
Yes

PR #1921, opened 2020-02-04
[NETBEANS-3735] FlatLaf: repaint desktop after closing the last editor tab

DevCharly:
Now always do an explicit repaint of the editor area to fix issue NETBEANS-3735.
It is maybe pure chance that the effect does not occur in other LAFs because they change the border of the editor area, which does the repaint.


lkishalmi:
Really Cool! Thanks!

PR #998, opened 2018-10-31
[NETBEANS-1586] Make ImageUtilities.createDisabledIcon work with HiDPI icons

eirikbakke:
After the introduction of the org.openide.util.VectorIcon class in https://github.com/apache/incubator-netbeans/pull/859 , ImageUtilities.createDisabledIcon should be updated to work with such resolution-independent icons, as well as ImageIcon instances that delegate to a MultiResolutionImage. This patch achieves that, without making assumptions about the specific Icon implementation. Tracked as https://issues.apache.org/jira/browse/NETBEANS-1586 .

The new ImageUtilities.createDisabledIcon implementation was tested with both a VectorIcon and an ImageIcon, including with the latter delegating to a MultiResolutionImage, on OpenJDK 10.0.2 on Windows 10 with a HiDPI/non-HiDPI dual monitor setup. Also tested the VectorIcon case on MacOS with Retina and non-Retina configurations.

(This is the only change to ImageUtilities I'm doing for now. The other methods in ImageUtilities will eventually need to be updated to support HiDPI icons as well. That's left for future work, tracked in https://issues.apache.org/jira/browse/NETBEANS-1583 )

PR #1922, opened 2020-02-05
[NETBEANS-3671] Fix two ImageUtilities issues

eirikbakke:
Fix two problems caused by the recent (HiDPI-related) ImageUtilities revamp ( https://github.com/apache/netbeans/pull/998 and https://github.com/apache/netbeans/pull/1273 ).

* Make the Icon instance returned by ImageUtilities.loadImageIcon properly serializable.
* Make FilteredIcon (used by ImageUtilities.createDisabledIcon) robust against icons with zero or negative dimensions. Add a log message for the negative dimension case (once observed in NETBEANS-3671).

See:
https://issues.apache.org/jira/browse/NETBEANS-3769 (though in this case the HiDPI-related exception turned out not to be the cause of the bug)
https://issues.apache.org/jira/browse/NETBEANS-3671

### DIFF
--- a/platform/openide.util.ui/src/org/openide/util/CachedHiDPIIcon.java
+++ b/platform/openide.util.ui/src/org/openide/util/CachedHiDPIIcon.java
@@ -78,6 +78,10 @@ public abstract class CachedHiDPIIcon implements Icon {
         this.height = height;
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


ebarboni:
ok to squash and commit ? "Added some tests for the previous bugfixes." will be out of context

eirikbakke:
Yes, a squash was anticipated here.

PR #1925, opened 2020-02-05
Add UI to be able to disable Janitor on the Startup

lkishalmi:
Well, I've added some UI Support:

The Options Dialog:

JanitorOptions

The cleanup action:
JanitorOptions2


AlexFalappa:
Hi @lkishalmi , just looking at the screenshots I find a bit odd that label ending with **"... #days:"**. As far as I can tell there are no such abbreviations nowhere else in NetBeans.

I suggest having either a "days" label after the combo box or changing the label into something along the lines of "Removal age (number of days):".

There is also a typo in the last label: "**Junitor** run." instead of "**Janitor** run."

platform/janitor/src/org/netbeans/modules/janitor/JanitorOptionsPanel.java
```diff
@@ -0,0 +1,147 @@
...
+                    .addComponent(spUnusedDays, javax.swing.GroupLayout.PREFERRED_SIZE, javax.swing.GroupLayout.DEFAULT_SIZE, javax.swing.GroupLayout.PREFERRED_SIZE))
+                .addGap(18, 18, 18)
+                .addComponent(btRunNow)
+                .addPreferredGap(javax.swing.LayoutStyle.ComponentPlacement.UNRELATED)
+                .addComponent(lbRunNowInfo)
+                .addContainerGap(176, Short.MAX_VALUE))
+        );
+    }// </editor-fold>//GEN-END:initComponents
+
+    @Messages("lbRunNowInfo_Info=Junitor run. Check your Notifications.")
```
AlexFalappa: Typo here. Should be "Janitor"

lkishalmi:
@neilcsmith-net @AlexFalappa Thank you for your valuable comments! I'm working on these changes. If you find anything else that makes the wording more human please point that out!

AlexFalappa:
Labels are better now

lkishalmi:
Then let's have this merged.

PR #1875, opened 2020-01-16
[NETBEANS-3700] No more harcoding for Code Coverage Highlights.

lkishalmi:
The Options Panel:
CodeCoverahe-Highlights

The default colors:
CodeCoverage-NetBeans

Norway Today:
CodeCoverage-NorwayToday

FlatLaf Dark:
CodeCoverage-FlatlafDark

Well these are the colors I could came up with, I guess they could be improved by someone more artistic than me.

PR #1926, opened 2020-02-05
[NETBEANS-3792] No more hardcoded colors in Remove Surrounding Code action

AlexFalappa:
Exposed remove surrounding code color codes in editor color profile (Highlighting section):
options

Added color definitions to _NetBeans_, _Norway Today_, _CityLights_, _NetBeans55_, _NetBeans Earth_ and _FlatLaf Dark_ color profiles.

FlatLaf Dark before:
before-dark

FlatLafDark after:
after-dark

### DIFF
--- a/ide/defaults/src/org/netbeans/modules/defaults/BlueTheme-editor.xml
+++ b/ide/defaults/src/org/netbeans/modules/defaults/BlueTheme-editor.xml
@@ -67,4 +67,6 @@
     <fontcolor name="coverage-inferred" bgColor="1F4F40"/>
     <fontcolor name="coverage-partial" bgColor="909040"/>
 
+    <fontcolor name="remove-surround-code-delete" foreColor="ff32538C" bgColor="ff16253F"/>
+    <fontcolor name="remove-surround-code-remain" bgColor="ff0C3F33"/>
  </fontscolors>
--- a/ide/defaults/src/org/netbeans/modules/defaults/CityLights-editor.xml
+++ b/ide/defaults/src/org/netbeans/modules/defaults/CityLights-editor.xml
@@ -48,4 +48,6 @@
     <fontcolor name="trailing-whitespace"/>
     <fontcolor name="nbeditor-bracesMatching-sidebar"/>
     <fontcolor name="indent-guide-lines" foreColor="white"/>
+    <fontcolor name="remove-surround-code-delete" foreColor="lightGray" bgColor="ff333333"/>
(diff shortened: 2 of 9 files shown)


AlexFalappa:
I was wondering whether it would be more correct to move the bundle strings and default font color definitions from `ide/editor` module to `java/java.editor` module.

Something like PR #1875 by @lkishalmi .

AlexFalappa:
I have force pushed a new version where the bundle messages and fontcolors definitions are now in the `java/java.editor` module, closer to the _Remove Surrounding Code_ action.

AlexFalappa:
Dark Nimbus (with Norway Today color profile) after:
norway-today

AlexFalappa:
Colors in light color profiles (_NetBeans, NetBeans55, NetBeans Earth_) are those previously hardcoded in the action.

FlatLafLight (with NetBeans color profile) after:
flatlaflight-netbeans

eirikbakke:
Myself, I'm not very familiar with the conventions of where to put which constants etc.

AlexFalappa:
@lkishalmi some advanced functionalities are a bit hidden and hard to find, a _Tips and Tricks_ section in the wiki could be a good idea to raise user awareness.

neilcsmith-net:
@lkishalmi I didn't either - `ALT-BACKSPACE` or in Source menu - hmm, learn something new every day! :smile:

+1 fix looks like a good idea, and near to action makes sense - although begs the question why the action (if not the implementation) is Java specific.

Now, off to find out why `Insert Code` stopped working ... :confused:

matthiasblaesing:
The author information for this commit is broken and should not have been merged. Neither the author name nor the email is valid. @AlexFalappa please fix your git settings to prevent this in the future. For all committers, please ensure, that commits like these are not merged.

AlexFalappa:
@matthiasblaesing just realised the initial commit has a wrong user. This may be due to the fact that the initial commit was performed at work and not at home where my git credentials are correct.

Apologise for the mess. Will be more careful in the future. My other PR should be correct in this regard.

eirikbakke:
Perhaps a Travis test for correct git user/email would be good. E.g. check it the git info against an "Authors.txt" file in the root directory of the repo, where new committers can add their names and email addresses.

ebarboni:
@eirikbakke @matthiasblaesing is @eirikbakke proposition something we can do ? storing mail + name is something ok ?

PR #1927, opened 2020-02-05
[NETBEANS-3621] Guard agains NPE in Find usages

lkishalmi:
Preventing NPE when the element is null which is lately quite possible. I do not know if that is the right behavior I leave it for our compiler expert, however this one does not do any harm.


ebarboni:
looks ok to me, feel free to merge after more competent review than me

rtaneja1:
Change looks safe. Could not reproduce the NPE on constructor-Call Hierarchy in 11.2.

PR #1929, opened 2020-02-06
[NETBEANS-3342] Netbeans Installer on Java 11 and Mac OS X functional again.

MeFisto94:
So the issue is simple: The API of `com.apple.eawt.Application` has changed, it has been deprecated since Java 6 and as of Java 11 the methods have become unavailable causing a runtime failure when trying to run the Installer on Mac OS X with Java 11.

The API has changed (no big Application Listener but distinct handlers) and the code has been moved to java.awt.desktop.

Additonally I removed other deprecated methods (e.g. getMouseLocationOnScreen) because it's either not available any more or will be in a very soon future.

This PR has only one drawback: It requires a compilation with `-Dpermit.jdk9.builds=true`, though, there is no way around it unless preventing to run on Java 11+ JVMs


matthiasblaesing:
Quick first impression: This will not fly. NetBeans targets JDK 8, so neither source 11, nor using methods introduced after 8 will work.
BUT looking at the code it should be possible to use reflection and version checking to use the correct invocations based on the runtime JDK.

matthiasblaesing:
Second comment: Please ensure, that your author information is valid. In this case neither the author email, nor the author name look like valid information.

MeFisto94:
Well, it's my github nick and the github no-reply e-mail.
Do you require a clearname and the associated E-Mail or are there any special requirements? (i.e. having a @apache.org e-mail or something?).

I will give the reflection based approach a thought and see what I can do. I assume you require a squashed PR (i.e. force-pushing my branch?)

matthiasblaesing:
The Apache Foundation requires the donation of the changes to it. As the Apache NetBeans PMC we are bound by that requirement. For committers of the project this requires a signed ICLA. For people starting work on Apache NetBeans and/or only wanting to contribute small fixes we need the trackable intention that they want to contribute their code. In this case it is recorded in JIRA, that you created the PR and thus it can be expected, that you read the file headers and thus are aware, that you are donating your change.

Given the first paragraph this implies, that we require valid names and emails in commits.

For the update of the PR it depends on the state. If the history is irrelevant with a new changeset, a forced push is ok, as a complete re-review will be needed anyway. That would be the case, if you take a completly different approach (switch to reflection would be such a case). If a new commit is incremental it should be added as a new commit, so that observers of this PR can see the differences and concentrate on that. Depending on the complexity of the PR, full or selective squashing can be applied before merge.

oyarzun:
@matthiasblaesing @MeFisto94 I took a stab at this since we had run into the same issue. 

https://github.com/oyarzun/netbeans/commit/8c8e2ec0a7f92dcfc2472bbd40a90e09571efa27

This works and compiles with Java 8 and 11. I modeled it after platform/applemenu. Should I create a separate PR?

PR #1930, opened 2020-02-06
[NETBEANS-3428] FlatLaf Dark: fix text color when using project colored tab

DevCharly:
If option "same background color for files from the same project" (Options > Appearance > Document Tabs) is enabled, editor tabs have different background colors, but the text color was always black. This PR changes the text color for FlatLaf Dark to (nearly) white to make it consistent with other text and maybe easier to read. Other LAFs are not affected.

Before (3 files from 3 projects):
image

After:
image

The change in `FlatLaf.properties` scales the "underline" height of selected editor tab (in multi-tabs) in Java 8 on HiDPI screens.

Before (Java 8 at 150%):
image

After:
image


ebarboni:
looks ok merging

PR #1931, opened 2020-02-06
[NETBEANS-3806] - cleanup hashcode in DBElementsCollection

BradWalker:
I came across this issue in Netbeans with the DBElementsCollection class.

Basically the hascode is being returned as a String but the code removes checking for type by treating the hashcode as an object.

This change keeps it as a String and removes the folloiwng warning..

[repeat] /home/bwalker/src/netbeans/java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/DBElementsCollection.java:41: warning: [rawtypes] found raw type: HashSet
 [repeat] protected static transient HashSet instances = new HashSet();
 [repeat] ^
 [repeat] missing type arguments for generic class HashSet<E>
 [repeat] where E is a type-variable:
 [repeat] E extends Object declared in class HashSet

### DIFF
--- a/java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/DBElementsCollection.java
+++ b/java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/DBElementsCollection.java
@@ -28,19 +28,23 @@ import org.netbeans.modules.dbschema.*;
  */
 public class DBElementsCollection implements DBElementProperties {
 
-    /** Object to fire info about changes to */
+    /**
+     * Object to fire info about changes to
+     */
     DBElementImpl owner;
-
     DBElement[] _elms;
 
-    /** Array template for typed returns */
+    /**
+     * Array template for typed returns
+     */
     private Object[] _template;
 
-    //workaround for bug #4396371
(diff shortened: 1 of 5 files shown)


eirikbakke:
In this one there's a lot of formatting changes unrelated to the subject of the bug/description.

It's best to avoid making formatting changes for their own sake, except when the code is being changed in any case. Formatting make merges more difficult and clutters git history, e.g. when doing "git blame".

BradWalker:
I agree with your comment about the formatting changes. But, I just couldn't read the code before my changes.

Since I do my changes in Netbeans, I just decided to reformat it so that I could simply read it..

But, I agree and in the future will keep the format changes to a very minimum..

eirikbakke:
You can add more commits to this Pull Request by pushing to the same branch. Alternatively, you can amend the commit and force-push to the same branch. Then the diff will update here.

(This is how these Pull Requests usually go--people request changes, the original submitter pushes additional commits or force-pushed amendments, etc., until everybody is happy. In the end, the chain of commits in the PR is usually "squashed" by GitHub into a single, clean commit that goes onto the NetBeans master branch.)

PR #1935, opened 2020-02-08
[NETBEANS-3769] Use WindowsNotifier not NIO2. Avoids dangling directo…

errael:
Revert to using the native Notifier,  WindowsNotifier.java, rather than NIO2. After open/close a project, use SysInternals to see active handles, owned by NetBeans, for the directories in a project. On windows, this prevents remove/rename so the associated NB actions fail. See [NETBEANS-2785], [NETBEANS-3769].

I looked through close and didn't see where the watch keys for the directories are removed, but it would have been very easy to miss; for that matter, I'm not sure how they are added.

This change may have performance implications; the change was made to  improve performance on linux. I didn't see specific comments about windows. But if Notifier.removeWatch is being skipped, this seems like a memory leak on all systems. Since the windows notifier has been used for 11.2, it can probably be re-enabled after fixing this watch key leak.

### DIFF
--- a/platform/masterfs.windows/src/org/netbeans/modules/masterfs/watcher/windows/WindowsNotifier.java
+++ b/platform/masterfs.windows/src/org/netbeans/modules/masterfs/watcher/windows/WindowsNotifier.java
@@ -50,7 +50,9 @@ import org.openide.util.lookup.ServiceProvider;
  *
  * @author nenik
  */
-@ServiceProvider(service=Notifier.class, position=400)
+// This must come before NioNotifier to avoid dangling directory references.
+// See [NETBEANS-2785], [NETBEANS-3769]
+@ServiceProvider(service=Notifier.class, position=150)
 public final class WindowsNotifier extends Notifier<Void> {
     static final Logger LOG = Logger.getLogger(WindowsNotifier.class.getName());
 


lkishalmi:
I guess @errael  is experiencing this one: [JDK-6972833](https://bugs.openjdk.java.net/browse/JDK-6972833)
I'm fine with reverting the change back.

errael:
The issue that started the performance review, [NETBEANS-2291], speaks about huge numbers of things on reference queues. I'm wondering if NIO2 has a major object leak if it's keeping watchers on things that are closed. I recall in the rename case I saw closing before rename. Shouldn't NB be removing unused watchers? I don't know the implications or how difficult this would be. Should there be an issue on it?

matthiasblaesing:
I had a quick test today (not to scientific). The last few week I used a nightly build @work and noticed, that my angular builds semi randomly failed because the dist folder could not be deleted. With this PR applied multiple builds ran clean without the failures I observed. It should be noted, that this is not a workday, so not totally compareable to my day job.

errael:
@matthiasblaesing is there any indication from your test on performance impact of this PR?

matthiasblaesing:
Today I tried to get a feeling and I did not notice a difference in performance, but could work better with node builds. I have no numbers to underpin this, just a feeling.

lkishalmi:
Let's just be merged this before the new beta/rc/vc we were using the Windows implementation for years so that should be safe.

errael:
Yeah, the Windows notifier hasn't had a functional change for over 6 years. It was built for NB using native windows notifiers; just guessing but I think there's a good chance there won't be much, if any, performance degradation.

PR #1936, opened 2020-02-08
Heapview FlatLaF alternative colors

lkishalmi:
Well, this is just an alternative coloring of the HeapView on FlatLaF. Personally I think the yellow draws more attention to the eye than it should be. I also changed the view background color match the toolbar so it is more flat that way. If agreed it is trivial to merge for 11.3.
See the samples:
HeapViewDark

HeapView2

@DevCharly, @pedro-w  ?

### DIFF
--- a/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatDarkLaf.properties
+++ b/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatDarkLaf.properties
@@ -97,19 +97,6 @@ nb.formdesigner.gap.min.color=darken(@background,5%)
 nb.formdesigner.gap.border.color=lighten(@background,10%)
 nb.formdesigner.saw.color=lighten(@background,15%)
 
-
-#---- HeapView ----
-
-# text
-nb.heapview.foreground=#ddd
-
-# grid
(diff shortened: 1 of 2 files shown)


eirikbakke:
Looks nice! Perhaps the "flat" (panel-colored) background should just be used for the regular LAF as well.

lkishalmi:
Oh, just a before illustration:
MemoryGauge1

GeertjanWielenga:
I like it.

graben:
Might be a bit too "flat" without the background difference of the chart panel. This way the possible maximum is not visible but blue is definitely much better than yellow.   ;-)

lkishalmi:
Well, the possible maximum is vary especially on G1. I agree that having a different color as background would be more informative, but the question is that do we need to be that informative or just looking cool is enough. In this case I'd vote for the second one.

eirikbakke:
I don't have a strong opinion on this one... the function seems to be preserved well enough without an explicit max point, since the top and bottom of the plot area is symmetrical around the overlaid text (which shows the max in any case). I think the main function of the plot is to show changes in memory consumption over time, while the text shows the exact current ratio.

AlexFalappa:
I roughly agree with @graben, blue better than yellow and if the background would be a bit different would help to understand the chart extension.

However I have no strong opinions on this.

PR #1940, opened 2020-02-10
Fix member references in Jackpot scripts

jlahoda:
Fixing three problems related to the declarative Java hints:
-the declarative hints use :: as a separator between snippets and conditions - but :: is also used in meber references. So this patch is trying to disambiguate the two uses.
-implicit lambdas in the target snippets get erroneously explicit parameter handling (i.e. get types for the parameter names, often even wrong like "(ERROR)")
-for implicit lambdas with a single parameters, the parameter is still wrapped with parentheses
--added a configuration option for this; and this may be the most doubtful thing in the patch


AlexFalappa:
I like the code style change for having no parenthesis around single parameters lambdas.

I always found myself removing them after invoking code completion.

Another thing I would like to see is no explicit body in code completion of lambdas, i.e. having this:
```java
aStream.filter(s -> | );
```
instead of this:
```java
aStream.filter((s) -> {
    return false; //To change body of generated lambdas, choose Tools | Templates.
});|
```
where `|` denotes the final cursor position.

But I agree this should be covered in a different PR.

PR #1942, opened 2020-02-12
[NETBEANS-3317]: field TypeVar.bound has been encapsulated and rename…

jlahoda:
…d to _bound - should use getUpperBound() instead.

Might be reasonable to do for 11.3, if there's time. If not, then it will need to wait for 12.0.


ebarboni:
Does this help solving https://issues.apache.org/jira/browse/NETBEANS-3860 ?

ebarboni:
merging and prepare beta3

jlahoda:
Correct, it should fix that one too - I've just closed a few duplicates, including NETBEANS-3860. Thanks for merging!

PR #1948, opened 2020-02-14
Simple linux notifications integration module

blackleg:
Simple integration to use native linux notification system ([libnotify library](https://developer.gnome.org/libnotify/)).

Based on @emilianbold's [OpenBeans notifications module](https://github.com/OpenBeans/OpenBeans/tree/master/pkgsrc-coolbeans/ide/files/platform/ro.emilianbold.notifications) and [Intellij libnotify wrapper](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ui/LibNotifyWrapper.java).

Screenshot_2020-02-15_10-26-12
Screenshot_2020-02-15_10-26-31


blackleg:
@emilianbold  I updated my first comment in this pull request to add two screenshots to show how this integration looks.

matthiasblaesing:
Should this be discussed first? I.e. not just a PR, but a discussion on the mailing list? GUI changes have the tendency to introduce bikeshedding and in this case the question why implementations for windows/mac OS are not done will come up quickly. I assume this is a base for discussion as the finer parts of the api (action listener, balloon detail, priority) are currently uncovered.

For the libnotify bindings: Is the native encoding of the java platform the right encoding? Or does it need to be UTF-8 or something like that? By default JNA maps strings to (C) char by using the default platform encoding.

Personal comment: For me I don't like the Optional style programming here. Maybe I'm getting old, but simple null checks look much more straight forward and are IMHO more readable.

emilianbold:
I will not be the one merging anyhow, I was planning on adding @eirikbakke
for the review.

Of course, all you said can be discussed, I just gave my input.

I think it is hard to do all OSes at once because people have one OS where
they need this. Having to require at least 2 involved people means it never
gets done.

I also don't write code like that and find that sometimes an if and a null
check is more readable but I did find the style of using Optional so much
intriguing. Reminded me of nodejs for some reason.

--emi

sâm., 15 feb. 2020, 17:35 Matthias Bläsing <notifications@github.com> a
scris:

eirikbakke:
I don't have a running Linux system to test on, unfortunately.

OS-level notifications should probably not be the default behavior--though it could certainly be an option. And yes, if such an option is added, it would be lovely for it to work on for instance MacOS as well. I think that could be added later, though.

On default behavior: Users frequently get annoyed when apps reach "outside their sandbox"--e.g. Acrobat Reader begging to be updated, web pages asking permission to send notifications, Chrome tabs playing audio, or Spotify showing an album cover overlay when you try to adjust volume for a different app. For this reason I think NetBeans notifications should be shown within NetBeans only, at least by default.

.travis.yml
```diff
@@ -261,6 +261,8 @@ matrix:
             #- ant $OPTS -f ide/localhistory test
             - ant $OPTS -f ide/lsp.client test
             - ant $OPTS -f ide/notifications test
+            #- ant $OPTS -f ide/notifications.linux test
+            - ant -Dcluster.config=minimal -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false -Dtest-unit-sys-prop.ignore.random.failures=true -f ide/notifications.linux test
```
eirikbakke: Was there already some notifications.linux module in the NetBeans repo at some point? Where did the commented-out reference to this module come from?

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/FallbackNotification.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.netbeans.modules.notifications.linux;
+
+import org.openide.awt.Notification;
+
+/**
+ *
+ * @author Hector Espert
+ */
+public class FallbackNotification extends Notification {
```
eirikbakke: The fallback should probably be the existing notification system, rather than a null implementation.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/jna/Libnotify.java
```diff
@@ -0,0 +1,40 @@
...
+    
+    boolean notify_is_initted();
+    
+    boolean notify_init(String app_name);
+    
+    void notify_uninit();
+    
+    Pointer notify_notification_new(String summary, String body, String icon);
+    
+    boolean notify_notification_show(Pointer notification, Pointer error);
```
eirikbakke: These method names do not follow the Java naming conventions--I take it this is because they need to match names in native libraries?

(As long as the class does not get exposed outside of this module, that's fine. I assume it's just a ServiceProvider module.)

matthiasblaesing: These are JNA library bindings. If you don't want to provide a function name mapper JNA requires the java method name to match the native method name.
The whole jna-platform (bindings for several default libraries on various osses) is bound that way. While other styles might be more in line with the java naming conventions, keeping the native name helps finding documentation, so I strongly suggest to keep the binding like this.

eirikbakke: Yep, that's fine.

emilianbold:
> On default behavior: Users frequently get annoyed when apps reach "outside their sandbox"--e.g. Acrobat Reader begging to be updated, web pages asking permission to send notifications, Chrome tabs playing audio, or Spotify showing an album cover overlay when you try to adjust volume for a different app. For this reason I think NetBeans notifications should be shown within NetBeans only, at least by default.

Many users like it when an app behaves like the other apps and uses what the OS offers: notification center, keychains, configuration settings, etc.

And I think such OS-level notifications are a good, visual, bullet point to put in some release notes.

> Should this be discussed first? I.e. not just a PR, but a discussion on the mailing list?

I think passing this through dev@ and extensive API review severely decreases the chances of this PR going anywhere.

I will remove myself from this PR and leave you to it.

Good job @blackleg, the module is quite well done!

matthiasblaesing:
@emilianbold I did not reject it, I raised potential problems. As it is only an incomplete implementation of the `NotificationDisplayer` api we need to address what happens when features outside the supported range are requested, how adapters could be implemented and how the feature should be exposed to the user. I don't see as unreasonable to point that out.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/LinuxNotificationDisplayer.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private static final String LIBNOTIFY = "libnotify.so.4";
+
+    private static final String APP_NAME = "netbeans";
+
+    private Optional<Libnotify> optionalLibnotify;
+
+    public LinuxNotificationDisplayer() {
+        try {
+            this.optionalLibnotify = Optional.of(Native.load(LIBNOTIFY, Libnotify.class));
```
svenreimers: I know we typically do not use module installer these days, but I think putting the loading of the native library in a module install to let the install fail if the library is not available may be worth a try - with this I can imagine all other if / else / optional  stuff may not be needed anymore.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/LinuxNotificationDisplayer.java
```diff
@@ -0,0 +1,119 @@
...
+        return getDefaultNotificationDisplayer()
+                .map(diplayer -> diplayer.notify(title, icon, detailsText, detailsAction, priority))
+                .orElseGet(FallbackNotification::new);
+    }
+
+    @Override
+    public Notification notify(String title, Icon icon, JComponent balloonDetails, JComponent popupDetails, Priority priority) {
+        return getDefaultNotificationDisplayer()
+                .map(diplayer -> diplayer.notify(title, icon, balloonDetails, popupDetails, priority))
+                .orElseGet(FallbackNotification::new);
```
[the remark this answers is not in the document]
svenreimers: Maybe using module install (see above)

svenreimers:
Due to the fact that interaction of the notification with the IDE (think click on link) is not possible from the outside, i.e. the os native part of the world, I think the native notifications may be nice but can only be an add on to the in application notification (read Swing based in IDE notification). So what is probably needed from my point of view is the possibility to have some (all?) notifications be sent to the OS notification as well, if desired by the user (with the possibility to link back to the real notification in the IDE if supported by the OS). Sounds like a long shot to get this done right??? I think the hook provided so far is especially useful for notifying inside the IDE (in whatever way), but a notification outside may just be another level e.g. (if native notifying module available forward to this as well, if configured by user).

Comments?

Thanks to @blackleg for kicking this off.

blackleg:
> Should this be discussed first? I.e. not just a PR, but a discussion on the mailing list? GUI changes have the tendency to introduce bikeshedding and in this case the question why implementations for windows/mac OS are not done will come up quickly. I assume this is a base for discussion as the finer parts of the api (action listener, balloon detail, priority) are currently uncovered.
> 
@matthiasblaesing @svenreimers @emilianbold  @eirikbakke I have no problem if you prefer to discuss this on the mailing list and wait to continue developing this or not when the mailing list get a decision.

blackleg:
> For the libnotify bindings: Is the native encoding of the java platform the right encoding? Or does it need to be UTF-8 or something like that? By default JNA maps strings to (C) char by using the default platform encoding.
> 
@matthiasblaesing I'm not sure at this point but IntelliJ implementation of libnotify binding use the native encoding of the java platform.

neilcsmith-net:
Not sure how many people here know GObject system that well, and the libnotify docs are a little unforthcoming on ownership rules, but I think the lack of `g_object_unref` mapping might be a sign of a memory leak in this.

 > Due to the fact that interaction of the notification with the IDE (think click on link) is not possible from the outside

@svenreimers why not?  eg. libnotify allows for registering a callback function, although I have a feeling there may be a little more integration with GLib required than just passing that to get it working.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/LinuxNotificationDisplayer.java
blackleg: @svenreimers Is there an example of ModuleInstall that I can use?

.travis.yml
blackleg: > Was there already some notifications.linux module in the NetBeans repo at some point? 
> Where did the commented-out reference to this module come from?
No, it is only commented for debug propurses.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/LinuxNotificationDisplayer.java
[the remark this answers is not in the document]
blackleg: I tried to do it using a ModuleInstall, but I couldn't.  
Instead I'm using a [OnStart](http://bits.netbeans.org/dev/javadoc/org-openide-modules/org/openide/modules/OnStart.html)/[OnStop](http://bits.netbeans.org/dev/javadoc/org-openide-modules/org/openide/modules/OnStop.html) classes to do a similar behavior, but module can't disable itself.

ide/notifications.linux/src/org/netbeans/modules/notifications/linux/FallbackNotification.java
blackleg: Fallback is the existing notification system, `FallbackNotification` was the fallback of the fallback, but I removed this because it creates confusion.

AlexFalappa:
From a user point of view it would be great if native notifications would be automatically used when the main window is minimized or not in focus. I don't know if this is feasible or not.

Think of this: launch a lenghty compilation, switch to the browser to look at something, native notification popping up to warn build complete.

I wouldn't mind if a first implementation does not have link clicking support, those notifications usually happen with the main window in focus.

eirikbakke:
> I have no problem if you prefer to discuss this on the mailing list and wait to continue developing this or not when the mailing list get a decision.

I think the need for an extended discussion occurs primarily when it comes to changing the _default_ behavior. If it's just something that interested users can enable in the Options dialog, then the stakes are much lower, and it also allows improvements to be made gradually without implementing everything at once (MacOS support etc.).

neilcsmith-net:
Well, -1 from me until the memory handling question above is answered.  Test files definitely show expected use of `g_object_unref` - eg. https://github.com/GNOME/libnotify/blob/mainline/tests/test-basic.c

blackleg:
I'm going to close this pull request. I will change the approach because IMHO implement this require a lot of modifications in the current implementation of notifications module.

To compensate this, there are two pull requests to improve testing for the notifications system.

[Add tests in notifications module](https://github.com/apache/netbeans/pull/1979)
[Test SimpleNotificationDisplayer implementation](https://github.com/apache/netbeans/pull/1953)

PR #1949, opened 2020-02-15
Typo fixes in doco for api.annotations.common

apjanke:
Just a couple suggested typo fixes for the API documentation in api.annotations.common.

### DIFF
--- a/platform/api.annotations.common/src/org/netbeans/api/annotations/common/package-info.java
+++ b/platform/api.annotations.common/src/org/netbeans/api/annotations/common/package-info.java
@@ -18,11 +18,10 @@
  */
 
 /**
- * The API containingg common annotations (namely for defect detection)
+ * The API containing common annotations (namely for defect detection)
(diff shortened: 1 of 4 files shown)


platform/api.annotations.common/src/org/netbeans/api/annotations/common/package-info.java
```diff
@@ -18,11 +18,11 @@
  */
 
 /**
- * The API containg common annotations (namely for defect detection)
+ * The API containingg common annotations (namely for defect detection)
```
mklaehn:
```suggestion
 * The API containing common annotations (namely for defect detection)
```

apjanke: [Muphry's Law](https://en.wikipedia.org/wiki/Muphry%27s_law) strikes again! Fixed. Thanks.

ebarboni:
thanks for the typeo merging

PR #1950, opened 2020-02-15
[NETBEANS-3873] Fix project from maven archetype wizard

AlexFalappa:
This fixes the description text area in the second step of the "Project from Maven Archetype" wizard having a light background on dark LAFs.

See bug on Jira for screenshot of problem. After fix:
after-archetype-project

The hardcoded color remains as a fallback but should be rarely used now.


AlexFalappa:
This could go into NetBeans 11.3 as it is a one liner.

Priority to the voting process. It could be merged if there will be a voting candidate 2.

ebarboni:
merging it seeems not hurting change

PR #1953, opened 2020-02-17
Test SimpleNotificationDisplayer implementation

hectorespert:
Add test for SimpleNotificationDisplayer implementation and resolve some warnings.

### DIFF
--- a/platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
+++ b/platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.openide.awt;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.swing.JComponent;
+import org.junit.Test;
+import org.netbeans.junit.NbTestCase;
+import org.openide.awt.NotificationDisplayer.Category;
+import org.openide.awt.NotificationDisplayer.Priority;
+
+/**
+ *
+ * @author Hector Espert
+ */
+public class NotificationDisplayerTest extends NbTestCase {
+    
+    private NotificationDisplayer notificationDisplayer;
+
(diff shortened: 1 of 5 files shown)


platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.Test;
+import org.netbeans.junit.NbTestCase;
+import org.openide.awt.NotificationDisplayer.Category;
+import org.openide.awt.NotificationDisplayer.Priority;
+
+/**
+ *
+ * @author Hector Espert
+ */
+public class NotificationDisplayerTest extends NbTestCase {
```
eirikbakke: Maybe rename to SimpleNotificationDisplayerTest.

platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public NotificationDisplayerTest(String name) {
+        super(name);
+    }
+    
+    @Override
+    public void setUp() {
+        notificationDisplayer = NotificationDisplayer.getDefault();
+        assertNotNull(notificationDisplayer);
+        assertEquals("SimpleNotificationDisplayer", notificationDisplayer.getClass().getSimpleName());
```
eirikbakke: Should we make assumptions about which NotificationDisplayer is the default? That's not specified by the API.

Are these tests that should pass for _any_ implementation of NotificationDisplayer?

platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    public void testNotify_4args() {
+        Notification notification = notificationDisplayer.notify("title", null, "details", null);
+        assertNotNull(notification);
+        assertEquals("NotificationImpl", notification.getClass().getSimpleName());
```
eirikbakke: That's an implementation detail--ideally the tests should not break if you change the implementation while following the API specs.

platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
```diff
@@ -0,0 +1,119 @@
...
+        assertEquals("title", StatusDisplayer.getDefault().getStatusText());
+    }
+
+    @Test
+    public void testNotify_6args_4() {
+        Notification notification = notificationDisplayer.notify("title", null, (JComponent)null, (JComponent)null, Priority.LOW, "default_category_error");
+        assertNotNull(notification);
+        assertEquals("NotificationImpl", notification.getClass().getSimpleName());
+        assertEquals("title", StatusDisplayer.getDefault().getStatusText());
```
eirikbakke: These tests are repetitive--maybe factor out into a single testNotify(String expectedTitle, Notification notification) method.

platform/openide.awt/test/unit/src/org/openide/awt/NotificationDisplayerTest.java
blackleg: > Should we make assumptions about which NotificationDisplayer is the default? That's not specified by the API.

It doesn't assume that `SimpleNotificationDisplayer` is the default NotificationDisplayer, only is for testing that if any other is not present, fallback implementation is loaded.

PR #1955, opened 2020-02-19
Hook formating/rangeFormating into LSP Client

matthiasblaesing:
I noticed that the typescript editor is lacking formatting. I investigated and
learned, that the LSP protocol in principle supports formatting and the
typescript language server also contains the necessary support.

This changeset hooks the _format_ and _rangeFormat_  functions into the
IDE.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.lsp.client.bindings;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.swing.text.BadLocationException;
+import javax.swing.text.StyledDocument;
+import org.eclipse.lsp4j.DocumentFormattingParams;
+import org.eclipse.lsp4j.DocumentRangeFormattingParams;
+import org.eclipse.lsp4j.FormattingOptions;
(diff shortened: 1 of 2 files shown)


ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private final Context ctx;
+
+    public Formatter(Context ctx) {
+        this.ctx = ctx;
+    }
+
+    @Override
+    public void reformat() throws BadLocationException {
+        Object stream = ctx.document().getProperty(Document.StreamDescriptionProperty);
```
jlahoda: Better that getting the StreamDescriptionProperty is to use (org.netbeans.modules.editor.)NbEditorUtilities.getFileObject (less code and less error prone). We could also add LSPBindings.getBindings(Document), to simplify the code, but surely not urgent.

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
```diff
@@ -0,0 +1,153 @@
...
+    private void rangeFormat(FileObject fo, LSPBindings bindings) throws BadLocationException {
+        DocumentRangeFormattingParams drfp = new DocumentRangeFormattingParams();
+        drfp.setTextDocument(new TextDocumentIdentifier(Utils.toURI(fo)));
+        drfp.setOptions(new FormattingOptions(
+            IndentUtils.indentLevelSize(ctx.document()),
+            IndentUtils.isExpandTabs(ctx.document())));
+        drfp.setRange(new Range(
+            Utils.createPosition(ctx.document(), ctx.startOffset()),
+            Utils.createPosition(ctx.document(), ctx.endOffset())));
+        List<? extends TextEdit> edits = Collections.EMPTY_LIST;
```
jlahoda: Nit: might be better to use Collections.emptyList(), as using Collections.EMPTY_LIST leads to an unchecked cast (at least I believe so).

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
```diff
@@ -0,0 +1,153 @@
...
+        } catch (InterruptedException | ExecutionException ex) {
+            LOG.log(Level.INFO,
+                String.format("LSP document format failed for {0}", fo),
+                ex);
+        }
+        applyTextEdits(edits);
+    }
+
+    private void applyTextEdits(List<? extends TextEdit> edits) throws BadLocationException {
+        Collections.sort(edits, rangeReverseSort);
```
jlahoda: Would it make sense to call Utils.applyEditsNoLock here? Could be a new overload taking "start" and "end" offsets; or a predicate like "BiPredicate<Integer, Integer> acceptor", which would then be used to filter out the interesting edits (although the latter feels fairly ugly). Would help in case we find a bug in the way the edits are applied.

(the rangeReverseSort is a nicer variant of what is in Utils.applyEditsNoLock, so could be moved into Utils and be used instead of the current code.)

matthiasblaesing:
@jlahoda thank you for your input. I pushed an update, that tries to implement the suggested changes.  For applyTestEdits I decided to use the simple variant (start and end index instead of a BiPredicate) - I have not a clear vision where a more complex predicate than the start and end filter could be needed. As the function is purely used internally we can still evolve it it if necessary.

PR #1957, opened 2020-02-20
Ensure a clean license stub is used for source config builds

matthiasblaesing:
Before this changeset the LICENSE file from the root of the
repository/source directory was used to create the LICENSE file of the
binary build. This worked correctly for the repository because there
only the ALv2 is present, but a source config build in the LICENSE file
also the references to the license files for the embedded source files
with foreign licenses are included.

A binary build from a source config thus contains the license info of
the source config _and_ the binary build. This is not correct, as the
binary config can contains less source only files because they are only
required at build time.

To fix this, a new license-stub.txt is introduced, that only contains
the ALv2 regardless whether it comes from the repository or from a
source config build.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -203,6 +203,6 @@
 
 ********************************************************************************
 Apache NetBeans repository includes a number of source files that are not
-covered by the apache license. These files are refered to by licenseinfo.xml
+covered by the Apache license. These files are refered to by licenseinfo.xml
 files which hold the license information.
 ********************************************************************************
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -201,3 +201,8 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 
+********************************************************************************
+Apache NetBeans repository includes a number of source files that are not
+covered by the apache license. These files are refered to by licenseinfo.xml
```
BradWalker: A very, very minor nit.. I believe that apache should be capitalized to be Apache.

jlahoda: +1

ebarboni:
Hi @matthiasblaesing , I'm in the signing process (on my box it's a bit long)
Is the issue solved on the VC2, the artefacts are here https://builds.apache.org/view/M-R/view/NetBeans/job/netbeans-TLP/job/netbeans/job/release113/20/artifact/dist/netbeans/11.3/vc2/

Would be to know if ok for you.

matthiasblaesing:
@ebarboni both source and binary look sane to me. Thank you.

PR #1962, opened 2020-02-22
[NETBEANS-3909] SQL Execution should reuse Output Window if possible

matthiasblaesing:
When executing an SQL statement, each time a new exection is started by
the user a new output windows is created. This quickly clutters the
output window pane. Instead of this it would be better to reuse the
existings output window.


AlexFalappa:
Faced the described behaviour, very annoying.

Please get this in.

PR #1963, opened 2020-02-23
[NETBEANS-1603] I can't see MySQL stored procedures in NetBeans

sbrbot:
There is a bug in NetBeans for long time. I noticed it in NetBeans v8.2 and reported it then, but it hasn't been resolved yet. The bug is that NetBeans does not show MySQL routines in database node tree because obsolete (and now non existent) table MYSQL.PROC is used. Currently for this purpose INFORMATION_SCHEMA's tables like ROUTINES, TRIGGERS and PARAMETERS are used exclusively. So old NetBeans queries were changed accordingly in order to show routines. Additionally, "Procedures" node name was changed into more suitable term "Routines" because this node contains both routine types: Procedures and Functions. (Currently this node in NetBeans contains triggers as well, although they should be shown under Table's node, I'll try to implement it in next change). Additional data about routines (like routine parameters data types, return type, is routine deterministic, etc) are fetched now and used in NetBeans. Routine view (info) dialog was enlarged little bit to be able to better show routine's parameters and body.


ide/db/src/org/netbeans/modules/db/explorer/dlg/Bundle.properties
```diff
@@ -300,7 +300,7 @@ ConnectionProgress_Established=Connection established.
 ConnectionProgress_Failed=Connection failed.
 ViewProcedureDialog.lName.text=&Name:
 ViewProcedureDialog.lParams.text=&Parameters:
-ViewProcedureDialog.lBody.text=&Source:
+ViewProcedureDialog.lBody.text=&Body:
```
matthiasblaesing: Any reason for changing this? I think Source matches as well as Body - do we need the bikeshedding?

ide/db/src/org/netbeans/modules/db/explorer/dlg/Bundle.properties
```diff
@@ -322,7 +322,7 @@ mysql.from=http://dev.mysql.com/downloads/connector/j/
 oracle.driver.name.prefix=ojdbc7.jar, ojdbc7_g.jar, ojdbc7dms.jar
 mysql.driver.name.prefix=mysql-connector-java
 oracle.driver.name=ojdbc7.jar
-mysql.driver.name=mysql-connector-java-8.0.17.jar
+mysql.driver.name=mysql-connector-java-8.0.19.jar
```
matthiasblaesing: Not strictly necessary for the current changeset, but ok.

ide/db.metadata.model/src/org/netbeans/modules/db/metadata/model/jdbc/mysql/MySQLSchema.java
```diff
@@ -36,7 +36,7 @@
 
 /**
  *
- * @author David, Jiri Rechtacek
+ * @author David, Jiri Rechtacek, Stjepan Brbot
```
matthiasblaesing: The @author entries are mostly legacy entries - but ok.

ide/db/src/org/netbeans/modules/db/explorer/node/Bundle.properties
```diff
@@ -33,7 +33,7 @@ ND_ForeignKey=Foreign Key
 ND_PrimaryKey=Primary Key
 ND_Index=Index
 
-ND_ProcedureList=List of Stored Procedures
+ND_ProcedureList=List of Stored Routines
```
matthiasblaesing: Please keep the original name if there not evidence, that other DBMS apart from mysql use Routine instead of procedure. If there is evidence, please elaborate.

ide/db/src/org/netbeans/modules/db/explorer/node/Bundle.properties
```diff
@@ -54,7 +54,7 @@ ND_IndexList=List of Indexes
 DriverListNode_DISPLAYNAME=Drivers
 ForeignKeyListNode_DISPLAYNAME=Foreign Keys
 IndexListNode_DISPLAYNAME=Indexes
-ProcedureListNode_DISPLAYNAME=Procedures
+ProcedureListNode_DISPLAYNAME=Routines
```
matthiasblaesing: See comment for line 36

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureListNode.java
```diff
@@ -30,7 +30,7 @@
  * @author Rob Englander
  */
 public class ProcedureListNode extends BaseNode {
-    private static final String NAME = "Procedures"; // NOI18N
+    private static final String NAME = "Routines"; // NOI18N
```
matthiasblaesing: See comment for `Bundle.properties` line 36.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNodeProvider.java
```diff
@@ -220,23 +215,20 @@ public void run(Metadata metaData) {
...
                                 object2type = new HashMap<>();
                                 String schemaEscaped = schemaName.replace("'", "''");
-                                String query = "SELECT OBJECT_NAME, STATUS, OBJECT_TYPE" // NOI18N
-                                        + " FROM SYS.ALL_OBJECTS " // NOI18N
-                                        + " WHERE OWNER='" + schemaEscaped + "' "// NOI18N
-                                        + " AND ( OBJECT_TYPE = 'PROCEDURE' OR OBJECT_TYPE = 'TRIGGER' OR OBJECT_TYPE = 'FUNCTION' )";  // NOI18N
+                                String query = "SELECT object_name,status,object_type" // NOI18N
+                                             + " FROM sys.all_objects" // NOI18N
+                                             + " WHERE owner='" + schemaEscaped + "'"// NOI18N
+                                             + " AND object_type IN ('PROCEDURE','FUNCTION','TRIGGER')";  // NOI18N
```
matthiasblaesing: This is code for oracle why was it changed for mysql?

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
```diff
@@ -711,22 +723,21 @@ private static void updateProcedureProperties(ProcedureNode node,
         node.addProperty(ps);
         Type type = provider.getType(node.getName());
         if (type == null) {
-            LOG.log(Level.INFO, "Unknown type of object {0}", node.getName()); //NOI18N
+            LOG.log(Level.INFO, "Unknown type of object {0}", node.getName());
```
matthiasblaesing: //NOI18N was dropped

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
```diff
@@ -657,29 +672,26 @@ public boolean isEditSourceSupported() {
...
             if (nameIdx != -1) {
                 // don't duplicate owner
-                if (upperSource.substring(nameIdx + toFind.length()).trim().startsWith(owner.toUpperCase() + '.')) { // NOI18N
+                if (upperSource.substring(nameIdx + toFind.length()).trim().startsWith(owner.toUpperCase() + '.')) {
                     return source;
                 }
                 res = source.substring(0, nameIdx + toFind.length()) +
                         owner +
-                        '.' + // NOI18N
+                        '.' +
```
matthiasblaesing: //NOI18N was dropped

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
```diff
@@ -657,29 +672,26 @@ public boolean isEditSourceSupported() {
...
         private String fqn(String source, String owner) {
             String upperSource = source.toUpperCase();
-            String toFind = getTypeName(getType()) + " "; // NOI18N
+            String toFind = getTypeName(getType()) + " ";
             String res = source;
             int nameIdx = upperSource.indexOf(toFind);
             if (nameIdx != -1) {
                 // don't duplicate owner
-                if (upperSource.substring(nameIdx + toFind.length()).trim().startsWith(owner.toUpperCase() + '.')) { // NOI18N
+                if (upperSource.substring(nameIdx + toFind.length()).trim().startsWith(owner.toUpperCase() + '.')) {
```
matthiasblaesing: //NOI18N was dropped

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
```diff
@@ -657,29 +672,26 @@ public boolean isEditSourceSupported() {
...
             expression.append(SPACE).append(DELIMITER).append(NEW_LINE); // NOI18N
-            // unset delimiter
             expression.append("DELIMITER ; ").append(NEW_LINE); // NOI18N
             return expression.toString();
         }
 
         private String fqn(String source, String owner) {
             String upperSource = source.toUpperCase();
-            String toFind = getTypeName(getType()) + " "; // NOI18N
+            String toFind = getTypeName(getType()) + " ";
```
matthiasblaesing: //NOI18N was dropped

ide/db/src/org/netbeans/modules/db/explorer/dlg/Bundle.properties
sbrbot: The reason for this is that source code of database routines consist of header (definition of parameters and attributes) and routine body. Analogy like in html structure. NetBeans in this dialog box field shows only body part of routine while header (parameters) part is shown in another field. So it would be more suitable to call this source part - "body". The same terminology for this part of routine is used in db documentation. Finally, not big deal with this but now with this commit it was good opportunity to change this and I did it.

ide/db.metadata.model/src/org/netbeans/modules/db/metadata/model/jdbc/mysql/MySQLSchema.java
sbrbot: Generally I think that it's not bad idea to have some contributor names here because if someone else wants to involve himself in code change he/she can contact others if help is needed.

ide/db/src/org/netbeans/modules/db/explorer/node/Bundle.properties
sbrbot: Oracle and other databases separate functions and procedures into separate logical entities. NetBeans in its specific implementation for supporting these entities combining all them under the same node. Moreover, NetBeans under node whic is called "Procedures" includes triggers as well. Triggers generally are related to their table they are triggering upon. The reason for this is specific implementation in NetBeans is that triggers are not fully covered with MetaData functionality in JDBC implementation and one has to run SQL queries to fetch data about triggers inside DB. So in NetBeans this specific implementation combining all entities; Procedures, Functions, and Triggers IMHO is better named "Routines" than "Procedure" (Procedure which inside have Procedures, Functions, and Triggers).

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNodeProvider.java
sbrbot: Just wanted to keep the same SQL formatting for all them.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: My mistake here, this dot does not require I18N. It should be dot in all languages (not comma) because it concatenates owner and entity.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: The same as previous.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: Here omitted comment is Ok, blank should not be I18N-ed.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: Should log message be ever I18N-ed?

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
junichi11: No. Strings might be recognized by tools even if it is blank. Please don't remove it. Basically, please leave it as it is.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
junichi11: If it is translated, it's hard for a reader to understand.
e.g. A reporter writes "オブジェクトの型は未知です。" in Japanese to an issue. If the reader is not a Japanese, maybe, the reader can't understand what it means. 
Basically, please leave it as it is.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
junichi11: Perhaps, I suppose we need not add it for char, but basically, you should leave "existing" `// NOI18N`s as it is because it is not related to the issue.

matthiasblaesing:
@sbrbot my request boils down to this: It is easier to merge a PR if it does not raise discussions. @junichi11 and I most probably don't agree on everything, but if your PR does not touch areas it does not need to touch it will get approval easier. In this case changing the choise of words or removing NO18N markers does nothing to help the PR.

The same goes for the change to the Oracle statements. It is well possible, that the DB treats the statement case sensitive, then your change introduces a regression. The point is: I did not test on Oracle DB and my reading is, that you did neither, so it would be better to keep the case as-is.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: Agree, this //NOI18N comment should be kept.

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNode.java
sbrbot: Agree, comment // NOI180N should stay.

matthiasblaesing:
@sbrbot  there is no update - did you forget to push?

ide/db/src/org/netbeans/modules/db/explorer/dlg/Bundle.properties
sbrbot: Sorry I don't want to make big phylosophy of this. My simple elaboration is; cars are cars, trucks are trucks (not cars), if one wants to use one term for both of them, one cannot call them "cars" but "vehicles". The same analogy applies here: Oracle has stored procedures and stored functions and manages them differently (although similary) Triggers even more. (e.g. https://docs.oracle.com/cd/B25329_01/doc/appdev.102/b25108/xedev_programs.htm) and functions does not call "procedures", the term used for both of them would be "routines" (e.g. https://docs.oracle.com/cd/E17952_01/mysql-5.0-en/stored-routines.html). Even more, NetBeans combines under the same node Procedures, Functions and Triggers entities and I think "routines" is more propriate upper term for the combining navigator tree node than to call Procedures, Functionad and Triggers as "Procedures" only. (BTW here we have recursion) :-) I hope you would accept my explanation and reasoning (although I know how it goes with applying "new terminology" in reality).

matthiasblaesing: @sbrbot you are bikeshedding. Wether you call it Procedures or Routines does not matter. It does not help your cause. So please remove it. You still have not updated the PR (not even the things you aggreed to change).

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureNodeProvider.java
sbrbot: By default, Oracle identifiers (table names, column names, etc.) are case-insensitive. You can make them case-sensitive by using quotes around them (eg: SELECT * FROM "My_Table" WHERE "my_field" = 1). SQL keywords (SELECT, WHERE, JOIN, etc.) are always case-insensitive. I tested this on Oracle 11 database and can confirm 100% it does work fine.

ide/db/src/org/netbeans/modules/db/explorer/dlg/Bundle.properties
sbrbot: Returned &Source

ide/db/src/org/netbeans/modules/db/explorer/node/Bundle.properties
sbrbot: Returned Procedures

ide/db/src/org/netbeans/modules/db/explorer/node/Bundle.properties
sbrbot: Returned Procedures

ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureListNode.java
sbrbot: Returned Procedures

matthiasblaesing:
I had another look at this today (github does not notify when you push updates!), what I noticed:
- history is still not linear. At this point I suggest to either redo the changes cleanly or do a full rebase with squashing of this PR, so that it becomes reviewable again. (if you need advise, please ask on the dev  mailinglist)
- `ide/db.metadata.model/src/org/netbeans/modules/db/metadata/model/resources/Bundle.properties` changes line endings. Please revert.
- `ide/db/libsrc/org/netbeans/lib/ddl/impl/ProcedureArgument.java` contains unnecessary changes (only formatting). Please revert.
- ` ide/db/src/org/netbeans/modules/db/explorer/node/ProcedureListNode.java` contains only whitepace changes. Please revert.
- The parameters subquery in `ProcedureNode.java` that returns `routine_params` looks like it would need an order by (else the server is free to reorder)
- There are still changes in the oracle code in `ProcedureNode.java`, the same is true for `ProcedureNodeProvider.java`

matthiasblaesing:
Cosed in favor of #2010

PR #1974, opened 2020-02-24
[NETBEANS-3903] - Netbeans 11.2 fails to detect successful startup of…

nam-van-nguyen:
… Tomcat 9.0.31

  + Added new line to the HTTP head request to the Tomcat so it would return successfully (according to HTTP specs)


matthiasblaesing:
Please check your settings for the name - it currently reads (namnull). You can check it with the commandline client. Run `git log -n 1` while the NETBEANS-3903 branch is checked out. 

The commit message in general is ok, however  it is customary to leave an empty line between the summary (firstline) and the description (the other lines). Messages should also be wrapped to a maximum of 72 characters per line.

PR #1979, opened 2020-02-25
Add tests in notifications module

hectorespert:
Add tests in notifications module

### DIFF
--- a/ide/notifications/test/unit/src/org/netbeans/modules/notifications/NotificationSettingsTest.java
+++ b/ide/notifications/test/unit/src/org/netbeans/modules/notifications/NotificationSettingsTest.java
@@ -16,31 +16,28 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+package org.netbeans.modules.notifications;
 
-package org.netbeans.modules.notifications.filter;
-
-import org.openide.util.NbBundle;
+import org.junit.Test;
+import org.netbeans.junit.NbTestCase;
 
 /**
- * Utilities
+ *
+ * @author Hector Espert
  */
-final class Util {
+public class NotificationSettingsTest extends NbTestCase {
 
-    public static String getString(String key) {
-        return NbBundle.getMessage(Util.class, key);
+    public NotificationSettingsTest(String name) {
+        super(name);
     }
 
-    public static char getChar(String key) {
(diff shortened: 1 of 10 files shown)


ide/notifications/test/unit/src/org/netbeans/modules/notifications/StatusLineElementTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testGetStatusLineElement() {
+        StatusLineElementProvider statusLineElementProvider = Lookup.getDefault().lookup(StatusLineElementProvider.class);
+        assertNotNull(statusLineElementProvider);
+        assertEquals(StatusLineElement.class, statusLineElementProvider.getClass());
+        
+        Component component = statusLineElementProvider.getStatusLineElement();
+        assertNotNull(component);
+        assertEquals(FlashingIcon.class, component.getClass());
```
eirikbakke: Maybe adding a comment here why we must make an assumption about which implementation class is used for the default? This is testing something that is not really guaranteed from the API.

ide/notifications/test/unit/src/org/netbeans/modules/notifications/NotificationSettingsTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class NotificationSettingsTest extends NbTestCase {
+
+    public NotificationSettingsTest(String name) {
+        super(name);
+    }
+
+    @Test
+    public void testIsSearchVisible() {
+        NotificationSettings.setSearchVisible(false);
+        assertFalse(NotificationSettings.isSearchVisible());
```
eirikbakke: Did you ever see these fail? If so, it would be good to add a comment about it.

blackleg: I didn't see these fail, but I preferred to test these because they aren't a simple setter/getter.

ide/notifications/test/unit/src/org/netbeans/modules/notifications/StatusLineElementTest.java
blackleg: I added an error message in assert method indicating that FlashingIcon is the default implementation.

PR #1980, opened 2020-02-25
[NETBEANS-2446] Use *parent* of saved projectFolder as projectLocation

errael:
NOTE: this PR only addresses the problem caused by the first changeset listed. I'm filing a 2nd issue for the spurious directory creation which is the 2nd changeset. But I'm including it here for reference/completeness since it is in the issue report.

NETBEANS-2446 problem comes from two commits (seen in main-silver)
```
    changeset:   244321:021157efa98e
    parent:      244318:5036dc46020f
    user:        Milos Kleint <mkleint@netbeans.org>
    date:        Thu Jan 24 15:12:37 2013 +0100
    summary:     #217087 new wizard able to pick up the parent directory from Action's properties as initial value
```
and
```
    changeset:   301468:76d9cafcee35
    branch:      ArchetypesUI268677
    parent:      299970:0c5aa5cdb86a
    user:        Jaroslav Tulach <jtulach@netbeans.org>
    date:        Tue Oct 25 21:52:49 2016 +0200
    summary:     #268677: Recognizing .archetype template files and using them to instantiate projects via mvn archetype
```
The first change took out these two lines near the start of java/maven/.../BasicPanelVisual.read(WizardDescriptor settings)
```
        } else {
            projectLocation = projectLocation.getParentFile();
```
This is really unrelated to the overall changeset, except that PROJECT_PARENT_FOLDER is nearby.

Without these two lines, in NewProject > JavaWithMaven > JavaApplication, if you alternately click "next"/"back" you'll see
>    F:\tmp\proj1\proj1\proj1\proj1\

adding another "proj1" with each "next". Putting back these two lines fixes it.

The 2nd changeset problem (not addressed by this PR), "Recognizing .archetype template ...", causes spurious directory creation and makes the overall problem highly visible. This problem creates this deep directory tree even though finish is not pushed. Pushing cancel still creates directories.

The directory creation comes from java/maven/.../BasicPanelVisual.store(WizardDescriptor d)
```
    void store(WizardDescriptor d) {
        ...
        d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, parentFolder);
        if (d instanceof TemplateWizard) {
            parentFolder.mkdirs();
            ((TemplateWizard) d).setTargetFolder(DataFolder.findFolder(FileUtil.toFileObject(parentFolder)));
        }
```
The store method is called to save panel settings; it could be for cancel, finish, back, next.

I don't know if this code is for convenience, to avoid later calculations, or if it's required. I'll file a 2nd issue for this.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/newproject/BasicPanelVisual.java
+++ b/java/maven/src/org/netbeans/modules/maven/newproject/BasicPanelVisual.java
@@ -509,13 +509,14 @@ public class BasicPanelVisual extends JPanel implements DocumentListener, Window
         }
         String name = projectNameTextField.getText().trim();
         String folder = createdFolderTextField.getText().trim();
-        final File parentFolder = new File(folder);
+        final File projectFolder = new File(folder);
         
-        d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, parentFolder);
+        // PROJECT_PARENT_FOLDER confusing, better name is PROJECT_BASE_FOLDER
+        d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, projectFolder);
         if (d instanceof TemplateWizard) {
             ((TemplateWizard) d).setTargetFolderLazy(() -> {
-                parentFolder.mkdirs();
-                return DataFolder.findFolder(FileUtil.toFileObject(parentFolder));
+                projectFolder.mkdirs();
+                return DataFolder.findFolder(FileUtil.toFileObject(projectFolder));
             });
         }
         d.putProperty("name", name); //NOI18N
@@ -561,10 +562,14 @@ public class BasicPanelVisual extends JPanel implements DocumentListener, Window
                 handle = null;
             }
         }        
-        File projectLocation = (File) settings.getProperty(CommonProjectActions.PROJECT_PARENT_FOLDER); //NOI18N
-        if (projectLocation == null || projectLocation.getParentFile() == null || !projectLocation.getParentFile().isDirectory()) {
+        // PROJECT_PARENT_FOLDER confusing, better name is PROJECT_BASE_FOLDER
+        File projectFolder = (File) settings.getProperty(CommonProjectActions.PROJECT_PARENT_FOLDER); //NOI18N
+        File projectLocation;
+        if (projectFolder == null || projectFolder.getParentFile() == null || !projectFolder.getParentFile().isDirectory()) {
             projectLocation = ProjectChooser.getProjectsFolder();
-        } 
+        } else {
+            projectLocation = projectFolder.getParentFile();
+        }
         this.projectLocationTextField.setText(projectLocation.getAbsolutePath());
         
         String projectName = (String) settings.getProperty("name"); //NOI18N


errael:
Issue [NETBEANS-3918] for 2nd part about spurious directory creation.

AlexFalappa:
Could this also fix [NETBEANS-3875](https://issues.apache.org/jira/browse/NETBEANS-3875) ?

The issue is related to the New Project from Maven Archetype wizard that creates the folder before invoking maven archetype plugin and the specific archetype failing because of the existing empty folder.

errael:
I filed a second issue, [NETBEANS-3918], which is about spurious directory creation. I'll check to see which, if any, of those issues might include the problem in [NETBEANS-3875]

java/maven/src/org/netbeans/modules/maven/newproject/BasicPanelVisual.java
```diff
@@ -562,7 +562,9 @@ void read(WizardDescriptor settings) {
         File projectLocation = (File) settings.getProperty(CommonProjectActions.PROJECT_PARENT_FOLDER); //NOI18N
         if (projectLocation == null || projectLocation.getParentFile() == null || !projectLocation.getParentFile().isDirectory()) {
             projectLocation = ProjectChooser.getProjectsFolder();
-        } 
+        } else {
+            projectLocation = projectLocation.getParentFile();
```
matthiasblaesing: This looks like a contradiction to the name of the property in line 562. Could it be, that the property is set to the wrong value?

matthiasblaesing: Answering myself, yes this is broken (in `BasicPanelVisual#store`):

```java
        String name = projectNameTextField.getText().trim();
        String folder = createdFolderTextField.getText().trim();
        final File parentFolder = new File(folder);
        
        d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, parentFolder);
```

I would rename `parentFolder` to `targetFolder` and change the last line to `d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, targetFolder.getParentFile());`. But that is untested.

errael: The naming caused me considerable confusion. 

Note that your suggestion `d.putProperty("projdir", folder.getParentFile())` produces different and incorrect results. The String value of the property name ...PROJECT_PARENT_FOLDER is "projdir"; and in older areas of code is a literal string rather than a define. Your suggestion, save folder.getParentFile() as property value was actually my first attempt to fix this. Things got weird and changes started radiating; reverted that attempt. The property "projdir" is globally used by the wizard infrastructure and it means the base directory of the new project and not the parent directory of the newly created project directory.

After a circuitous route, I decided to find the the screen for a plain old java project. It is 
`java/java.j2seproject/src/org/netbeans/modules/java/j2seproject/ui/wizards/PanelProjectLocationVisual.java`
and it uses the same names and has similar code to its maven counterpart. (Undoubtedly the prototype for this screen)

So, while it's tempting to rename things so they make sense. I am hesitant to diverge from the original code. BTW, in the code you quote from the store() method, renaming the variable `parentFolder` to `projectDirectory` would be accurate. Then when the two lines are restored it would be like saying `projectLocation = projectDirectory.getParentFile()`. It could have read like that, but the original code was probably avoiding declaring new variables. Note that in the code in read() `projectLocation = ProjectChooser.getProjectsFolder()` retrieves the default value for where to put new projects the other branch of the "if" statement is working with the value of "projdir" and does ["projdir"].getParentFile().

matthiasblaesing: Ok - I might be wrong, but this is awfully lot text to say, that the project parent folder is not the project parent folder. We should at least try to get this straight, there might be more people like me, that read variables and think that they hold what they say. Sorry I can't say this without less irony.

errael: I'm itching to clean this up, glad you brought it up again. Yes, verbose; just saying `123 != 123` seemed inadequate. And I want to record the situation so I don't have to reconstruct it again. It does seem that simply restoring two lines that were mistakenly removed is consistent with a minimal diff philosophy.

PROJECT_PARENT_FOLDER should be renamed, maybe PROJECT_BASE_FOLDER (or just PROJECT_FOLDER); but that's an API change. PARENT could be deprecated in favor of BASE.

How abut this one? (keeping it local)

neilcsmith-net:
@errael I'll open a JIRA ticket, but looks like this change breaks the Maven Create Module Action - https://github.com/apache/netbeans/blob/master/java/maven/src/org/netbeans/modules/maven/nodes/ModulesNode.java#L334

errael:
Thanks for the heads up. The change has the comment
//PROJECT_PARENT_FOLDER confusing, better name is PROJECT_BASE_FOLDER
Rename is an API change. Of course most of the old/ancient code just 
using a literal string "proj.dir" or something like that. Hopefully it's 
usage is pretty local...

I guess some hack for 12.0. I haven't looked yet, maybe a property name 
like "parent.proj.dir" with a suitable CONSTANT is the right idea.

errael:
There's `public static final String PROJECT_PARENT_FOLDER = "projdir";` which is probably the root of my considerable confusion. If you grep this, like '"projdir"', it is used as a wizard property key for the directory of the project, not the parent directory (projectLocation).

It may be that this isn't an issue, and that the new/maven CommonProjectActions won't interact with the old stuff that this is a problem. I noticed that the change about lazy directory creation, that it never got invoked in actual use.

Given what "projdir" means in maven, I think you're right. Use `projectLocationTextField`; fix up read as needed. The original problem with read was after the back button when it was re-initializing the gui fields. Should have a PR later today.

PR #1981, opened 2020-02-25
[NETBEANS-3903] - Netbeans 11.2 fails to detect startup of Tomcat

nam-van-nguyen:
+ Added new line to the HTTP head request to the Tomcat so it would 
    return successfully (according to HTTP specs)


matthiasblaesing:
Change looks sane, is straight forward and was already tested. Thank you.

matthiasblaesing:
I assume that you are aware of the file header and thus agree, that you donate the changeset to the ASF. If not please speak out, the commit will then be reverted.

PR #1982, opened 2020-02-25
[NETBEANS-3807] - cleanup ArrayList rawtype warnings..

BradWalker:
This code was previously reverted because it was checked in by accident during the release window..
It's been re-based and tested.

This change cleans up the following warning..

  [repeat] /home/bwalker/src/netbeans/webcommon/javascript2.model/src/org/netbeans/modules/javascript2/model/ModelVisitor.java:1454: warning: [rawt
ypes] found raw type: ArrayList
  [repeat]             private List<VarNode> declaredVars = new ArrayList();
  [repeat]                                                      ^
  [repeat]   missing type arguments for generic class ArrayList<E>


eirikbakke:
Thank you--done!

PR #1984, opened 2020-02-26
[NETBEANS-3923] - cleanup ThreadLocal rawtype warnings

BradWalker:
This removes all the ThreadLocal rawtype warnings..

   [repeat] /home/bwalker/src/netbeans/platform/masterfs/src/org/netbeans/modules/masterfs/ProvidedExtensionsProxy.java:44: warning: [rawtypes] found ra
w type: ThreadLocal
   [repeat]     private static ThreadLocal  reentrantCheck = new ThreadLocal();
   [repeat]                    ^
   [repeat]   missing type arguments for generic class ThreadLocal<T>
   [repeat]     T extends Object declared in class ThreadLocal

### DIFF
--- a/ide/versioning.util/src/org/netbeans/modules/turbo/Memory.java
+++ b/ide/versioning.util/src/org/netbeans/modules/turbo/Memory.java
@@ -53,7 +53,7 @@ final class Memory {
     public static final Object NULL = new Object();
 
     /** Keep reference to last isPrepared result. */
-    public static final ThreadLocal prepared = new ThreadLocal();
+    public static final ThreadLocal<Entry> prepared = new ThreadLocal<>();
 
     private final Statistics statistics;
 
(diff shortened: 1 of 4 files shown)


ide/versioning.util/src/org/netbeans/modules/turbo/Memory.java
```diff
@@ -53,7 +53,7 @@
     public static final Object NULL = new Object();
 
     /** Keep reference to last isPrepared result. */
-    public static final ThreadLocal prepared = new ThreadLocal();
+    public static final ThreadLocal<Entry> prepared = new ThreadLocal<>();
```
matthiasblaesing: The users of `prepared` in this class could now drop their cast (I think else you are introducing another warning).

BradWalker: Good catch!! It's fixed now..

matthiasblaesing:
Thank you.

PR #1986, opened 2020-02-26
[NETBEANS-3332] Use the latest maven-nbm-plugin 4.4

oyarzun:
This PR updates the maven-nbm-plugin version to 4.4 when a user creates a new Maven based NetBeans project from the wizard. 

 **This PR requires that  nbm-archetype and netbeans-platform-app-archetype are updated, voted on, and published.** 

https://github.com/apache/netbeans-mavenutils-archetype-nbm-archetype/pull/1
https://github.com/apache/netbeans-mavenutils-archetype-nbm-suite-root/pull/1
https://github.com/apache/netbeans-mavenutils-archetype-netbeans-platform-app-archetype/pull/2


ebarboni:
Maybe it's possible to split the PR to hve one with 4.4 and another with archetype.

I think it would be nice to detect the most recent nbm/archetype instead of hardcoding.

oyarzun:
@ebarboni I think it is a good idea to have it detect the latest version.

I've created another PR #1999 to update the archetypes. Still looking into the maven-nbm-plugin version as that static variable is referenced in other classes. Also wanted to get feedback if my implementation in #1999 is the preferred method to solve the problem.

oyarzun:
@ebarboni I've created another PR #2017 to dynamically get the latest nbm-maven-plugin version. I'm going to close this PR.

PR #1987, opened 2020-02-28
Opening ide/jumpto API for everyone

JaroslavTulach:
OracleLabs IGV needs to access _Go to Type_ functionality. Right now I have to use implementation dependency, but that feels kinda hacky. Let's remove the list of friend modules and open up the `ide/jumpto` module to everyone.


sdedic:
shouldn't we update spec versions in formerly-friend modules to enforce the module update ?

jtulach:
I know: http://wiki.apidesign.org/wiki/FriendDependencies - however I guess in this situation (E.g. when opening to public) it is not necessary. The formerly friend modules will work with old as well as new version of jumpto - new modules of course have to use the latest version.

PR #1990, opened 2020-02-29
Fix HtmlRenderer.createLabel()

matthiasblaesing:
As the author or #1447 (@timboudreau) is unresponsive we either need to
drop or fix the PR. Instead of dropping it, I rebased it onto master and added
the missing NOI18N markers.

@junichi11 @neilcsmith-net could you have a look if it is mergeable in this
form?

@timboudreau maybe you could find some time to check if you agree with
this update.

Closes: #1447


matthiasblaesing:
Ok - apart from Mac OS unittests now return cleanly. I'll merge now, then we'll have time to catch problems before 12.0.

PR #1991, opened 2020-02-29
[NETBEANS-3918] defer creation of target folder until/if needed

errael:
Without this change, there is some directory creation in BasicPanelVisual.store, but that method is called in several situations where directory/DataFolder creation should not occur. This PR moves the creation code to TemplateWizard.getTargetFolder(), which defers the creation until needed. Unfortunately, with a breakpoint, I have not seen getTargetFolder() invoked when creating a maven project.

This PR is designed to preserve semantic; it is highly targeted and relatively simple. There may be better/simpler/more-direct fixes; but I am unfamiliar with the overall architecture; revset which introduced the spurious directory creation:

```
    changeset:   301468:76d9cafcee35
    branch:      ArchetypesUI268677
    parent:      299970:0c5aa5cdb86a
    user:        Jaroslav Tulach <jtulach@netbeans.org>
    date:        Tue Oct 25 21:52:49 2016 +0200
    summary:     #268677: Recognizing .archetype template files and using them to instantiate projects via mvn archetype
```

This simplest fix would be to simply remove the directory creation from BasicPanelVisual and not put it in getTargetFolder(). This would fix the cases I have seen; in those cases the directory is created by maven project instantiation and the data fold is not referenced. But this would not preserve semantics around TemplateWizard and DataFolder.

This PR also fixes [NETBEANS-3875] which is about failure of maven artifact creation.

### DIFF
--- a/platform/openide.loaders/src/org/openide/loaders/TemplateWizard.java
+++ b/platform/openide.loaders/src/org/openide/loaders/TemplateWizard.java
@@ -29,6 +29,7 @@ import java.text.MessageFormat;
 import java.util.*;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.swing.*;
@@ -91,6 +92,8 @@ public class TemplateWizard extends WizardDescriptor {
     private DataFolder targetDataFolder;
     /** This is true if we have already set a value to the title format */
     private boolean titleFormatSet = false;
+    /** for deferred directory/DataFolder creation */
+    private Supplier<DataFolder> targetDataFolderCreator;
 
     /** listens on property (steps, index) changes and updates steps pane */
     private PropertyChangeListener pcl;
@@ -243,6 +246,10 @@ public class TemplateWizard extends WizardDescriptor {
      */
     public DataFolder getTargetFolder () throws IOException {
         LOG.log(Level.FINE, "targetFolder={0} for {1}", new Object[] {targetDataFolder, this});
+        if (targetDataFolder == null && targetDataFolderCreator != null) {
+            targetDataFolder = targetDataFolderCreator.get();
+            LOG.log(Level.FINE, "lazy targetFolder={0} for {1}", new Object[] {targetDataFolder, this});
+        }
         if (targetDataFolder == null) {
             throw new IOException(NbBundle.getMessage(TemplateWizard.class, "ERR_NoFilesystem"));
         }
@@ -269,7 +276,22 @@ public class TemplateWizard extends WizardDescriptor {
         if (LOG.isLoggable(Level.FINE)) {
             LOG.log(Level.FINE, "set targetFolder=" + f + " for " + this, new Throwable());
         }
+        if(f == null) {
+            targetDataFolderCreator = null;
+        }
         targetDataFolder = f;
+    }
+
+    /** Use this method for deferred creation of target folder
+     * until and <em>if</em> it is needed.
+     * 
+     * @since 7.76
+     * 
+     * @param folderCreator function which, when called, produces TargetFolder
+     */
+    public void setTargetFolderLazy(Supplier<DataFolder> folderCreator) {
+        // Fix for [NETBEANS-3918] needs dererred creation feature
+        targetDataFolderCreator = folderCreator;
     }
 
     /** Getter for the name of the target template.
--- a/java/maven/nbproject/project.xml
+++ b/java/maven/nbproject/project.xml
@@ -458,7 +458,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>7.61</specification-version>
+                        <specification-version>7.76</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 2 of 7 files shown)


platform/openide.loaders/src/org/openide/loaders/TemplateWizard.java
```diff
@@ -244,7 +252,14 @@ public DataFolder getTemplatesFolder () {
     public DataFolder getTargetFolder () throws IOException {
         LOG.log(Level.FINE, "targetFolder={0} for {1}", new Object[] {targetDataFolder, this});
         if (targetDataFolder == null) {
-            throw new IOException(NbBundle.getMessage(TemplateWizard.class, "ERR_NoFilesystem"));
+            // Fix for [NETBEANS-3918] and [NETBEANS-3875]
+            File targetFolderFile = (File) getProperty(PROP_TARGET_FOLDER_HINT);
```
jtulach: The proposed property `PROP_TARGET_FOLDR_HINT` can hold any value, not just `File`. To prevent `ClassCastException` change this and next line to `if (value instanceof File) { ... }`.

platform/openide.loaders/src/org/openide/loaders/TemplateWizard.java
```diff
@@ -47,6 +48,13 @@
 * @see TemplateRegistration
 */
 public class TemplateWizard extends WizardDescriptor {
+    /** Set to <CODE>File</CODE> to use for deferred creation of target folder.
+     * If <CODE>TemplateWizard.getTargetFolder()</CODE> is called and the
+     * the target folder has not been set then this property value is
+     * used to create the directory and DataFolder. This allows deferral
+     * of creation of DataFolder until and <em>if</em> it is needed.
+     */
```
jtulach: This would be an API change and as such it would need additional attributes of an API change. See the [wiki page](http://wiki.netbeans.org/APIDevelopment) for overall description. Here is the enumeration of what I am missing the most:
- incrementing module version, using `@since` tag
- modifying `apichanges.xml`
- testing the change

platform/openide.loaders/src/org/openide/loaders/TemplateWizard.java
```diff
@@ -244,7 +252,14 @@ public DataFolder getTemplatesFolder () {
     public DataFolder getTargetFolder () throws IOException {
         LOG.log(Level.FINE, "targetFolder={0} for {1}", new Object[] {targetDataFolder, this});
         if (targetDataFolder == null) {
-            throw new IOException(NbBundle.getMessage(TemplateWizard.class, "ERR_NoFilesystem"));
+            // Fix for [NETBEANS-3918] and [NETBEANS-3875]
+            File targetFolderFile = (File) getProperty(PROP_TARGET_FOLDER_HINT);
+            if(targetFolderFile != null) {
+                targetFolderFile.mkdirs();
+                setTargetFolder(DataFolder.findFolder(FileUtil.toFileObject(targetFolderFile)));
```
jtulach: Please note that `TemplateWizard` hasn't yet been using `java.io.File`. This would be the first time and I am not happy about it as it limits the usage of `TemplateWizard` to local disk. Could you consider alternative solution? Could you introduce
```java
public void setTargetFolder(Supplier<DataFolder> toBeCreatedFolder) {
}
```
e.g. give the code a chance to work with `DataFolder` and `FileObject` only, but delay creation of the actual folder as much as possible?

errael:
New commit should be close.

In the comment for setTargetFolder(Supplier) I had following, but seemed to much, took it out
```
     * If <CODE>TemplateWizard.getTargetFolder()</CODE> is called and
     * the target folder is not set then invoke folderCreator
     * to create the directory and <CODE>DataFolder</CODE>.
```
Do you know someplace in the regular NetBeans code where I can use this and see it working in a non-test situation?

Did not error check that both version of setTargetFolder are used, or other weird cases. Seemed like overkill.

A little unclear about apichanges "compatibility". Since no differences if don't use new method, assume binary and source compatible. Not sure what modification means.

errael:
Travis build fails
```
/home/travis/build/apache/netbeans/ide/projectuiapi/src/org/netbeans/spi/project/ui/templates/support/Templates.java:152: error: reference to setTargetFolder is ambiguous
   [repeat]                 ((TemplateWizard)wizardDescriptor).setTargetFolder( null );
```
Ambiguous reference. To stay compatible, I can change the new method to unique name like
`setLazyTargetFolder(Supplier)`
Comment?

errael:
Actually, I like
`setTargetFolderLazy(Supplier)`
better. You would see both methods together in code completion.

jtulach:
Excellent! The last remaining thing: the java/maven module has to depend on 7.76 version of opening.loaders now, otherwise it will not link. Then the change can be merged.

errael:
Some time after seeing the earlier build failure, my subconscious informed me of a minor semantic difference. Previously
`wiz.setTargetFolder(null); wiz.getTargetFolder();`
would throw an assertion. Along with changing method name to setTargetFolderLazy, in setTargetFolder I've added
```
         if (LOG.isLoggable(Level.FINE)) {
             LOG.log(Level.FINE, "set targetFolder=" + f + " for " + this, new Throwable());
         }
+        if(f == null) {
+            targetDataFolderCreator = null;
+        }
         targetDataFolder = f;
     }
```
This maintains semantics AFAICT.

And set java/maven dependency to 7.76

errael:
If you're up to it, there's an almost trivial change PR#1980 which would finish up the obvious/glaring problems with maven project creation.

PR #1996, opened 2020-03-01
Only register LSP based formatting if MimePath has language server

matthiasblaesing:
It was observed that XML formatting did not work after enabling LSP
formatting. The problem was caused by the MimeLookup for application/xml
that returned two ReformatTask.Factory implementations. One was the
right factory, the other the Formatter provided by the LSP Client
module.

Instead of registering the ReformatTask.Factory of the LSP client module
on the empty mime path, it is now dynamically registered by a
MimeDataProvider (LspMimeDataProvider). That MimeDataProvider ensures
that the lsp based formatter is only registered if a
LanguageServerProvider is present on the targetted mime path.

As an implementation note the LspMimeDataProvider is registered with a
position, which ensures, that it is registered with higher priority than
LegacyFormattersProvider which unconditionally registers a formatter.
Having a higher priority (any position has higher priority than no
position) ensures LspMimeDataProvider is queried first.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
@@ -19,7 +19,6 @@
 package org.netbeans.modules.lsp.client.bindings;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
@jlahoda could you please have a look at this? I noticed that after merging the lsp formatting patch my nightly build no longer supports formatting XML files. I suspect that other file types are affected too. The approach I chose might also be interesting for other cases where services should not be registered on the empty mime path.

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/Formatter.java
```diff
@@ -47,12 +45,18 @@
...
-    @MimeRegistration(mimeType = "", service = ReformatTask.Factory.class)
     public static class Factory implements ReformatTask.Factory {
 
         @Override
         public ReformatTask createTask(Context context) {
-            return new Formatter(context);
+            FileObject file = NbEditorUtilities.getFileObject(context.document());
+            if (file != null) {
+                LSPBindings bindings = LSPBindings.getBindings(file);
+                if(bindings != null) {
```
junichi11: Nitpick: missing whitespace after `if`

matthiasblaesing:
Thank you both for checking - merged with updated format.

PR #1998, opened 2020-03-04
Fix typo in gradle License Header menu

sg4e:
This simple PR corrects a typo in the gradle License Header menu: *it's* -> *its*.


matthiasblaesing:
Thank you - we can't accept anonymous changes and as it is a minimal change I pushed it myself.

PR #2000, opened 2020-03-05
Adding full support for Wildfly 17, 18

ehsavoie:
Fixing issue with missing JavaEE 8 web profile
Removing exteernla downloads and updating schemas

This should bring full WildFly support back into the contributed plugin for Apache NetBeans


contrib/javaee.wildfly/licenseinfo.xml
```diff
@@ -47,4 +52,26 @@
...
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-app_4_0.dtd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-service_4_0.dtd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss_4_0.dtd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_7_3.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_4_0.dtd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/datasources_1_4.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_6_0.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-as-messaging-deployment_1_0.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/wildfly-messaging-activemq-deployment_1_0.xsd</file>
+        <license ref="LGPL-2.1" />
```
emilianbold: I think LGPL is category X for Apache. Can you confirm the license?

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/config/gen/package-info.java
```diff
@@ -58,8 +58,38 @@
...
     ),
+//    @Schema2Beans(
+//        schema="../../resources/jboss-ejb3-2_1.xsd",
+//        schemaType=SchemaType.XML_SCHEMA,
+//        outputType=OutputType.TRADITIONAL_BASEBEAN,
+//        validate=true,
+//        attrProp=true,
+//        removeUnreferencedNodes=true,
+//        docRoot="enterprise-beans",
+//        java5=true
```
emilianbold: Are these comments something we could enable in the future?

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-as-messaging-deployment_1_0.xsd
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2011, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
```
emilianbold: Again LGPL, Apache can't use it. No other variant?

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_6_0.xsd
```diff
@@ -0,0 +1,572 @@
...
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified"
+            version="6.0">
+
+   <xsd:annotation>
+      <xsd:documentation> JBoss, Home of Professional Open Source Copyright 2005, JBoss Inc., and
+         individual contributors as indicated by the @authors tag. See the copyright.txt in the
+         distribution for a full listing of individual contributors. This is free software; you can
+         redistribute it and/or modify it under the terms of the GNU Lesser General Public License as
```
emilianbold: LGPL

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/config/gen/package-info.java
ehsavoie: I don't think those are required at all. i should remove them

contrib/javaee.wildfly/licenseinfo.xml
ehsavoie: Yes
I can update some (like wildfly-messaging-activemq-deployment_1_0.xsd can be moved to ASL 2.0) I can also remove some of them but then we would loose support for old JBoss / WildFly versions

emilianbold: It would be useful to comment out the LGPL stuff and replace what you can replace with a better license (like Apache 2.0).

emilianbold:
> What we can do is to keep the ALv2 in the contrib directory, LGPL dependencies at build time and not release the module - some one else would need to do it.

Maybe @ehsavoie  can explain some more but according to hist post [here](https://github.com/apache/netbeans/pull/2000#discussion_r390845524): 

> I can also remove some of them but then we would loose support for old JBoss / WildFly versions

... we can have a plugin that is part of NetBeans proper. NetBeans will just only support more recent JBoss/WildFly version.

Personally I would also prefer to keep and comment out the LGPL schema names/references (not LGPLed content) so it become super easy to make a build of the plugin that supports older version too (in case people really need to search for it).

I think it's important to move this module from contrib/ back into NetBeans and support (modern) JBoss/WildFly out of the box.

ehsavoie:
I think some of the 'offending' schemas are for old versions. I'm not sure they are required for newer versions of WildFly. One solution also could be to download them at build time. Would that be sufficient ?

emilianbold:
> One solution also could be to download them at build time. Would that be sufficient ?

I don't think so, because then the binary would have LGPL stuff, but perhaps @matthiasblaesing can confirm.

I think another option would be to add an user action which does this download. If it's done by the user on the user's machine then Apache is fine with it.

Another option would be to invent some layer location or service provider that provides these schemas. Then users install the plugin with the modern schemas but can easily install a 2nd plugin (eg. `Wilfly extra`) which includes the other schema files. I would prefer this option compared to an user UI action but your call.

matthiasblaesing:
My understand is this: If we only need the Schema files at runtime, we could use them from an installed jboss, but it becomes problematic when we base code on the XSDs. If I remember correctly, the Schemas and DTDs are used to create code that is build and my understanding is, that the license transcends from the Schema to the generated code. This would mean, that we could not release the plugin from Apache NetBeans.

emilianbold:
> If I remember correctly, the Schemas and DTDs are used to create code that is build and my understanding is, that the license transcends from the Schema to the generated code. This would mean, that we could not release the plugin from Apache NetBeans.

As somebody that doesn't even do JavaEE and doesn't even care about any particular web server, let alone JBoss or Wildfly, my angle here is:

1. How can we add Wildfly back to the official NetBeans tree so it gets released to all the users.
2. How can we make it easy so older JBoss / Wildfly versions with problematic schemas are supported for users.

The answer to 1. seems easy: we just don't include the old schemas and we re-add the plugin to NetBeans using only the schemas that are acceptable (under and Apache 2.0 license, or whatever).

For an answer to 2 we have to get creative, the same way we did it for nb-javac and other situations. My idea is that a 2nd "companion" plugin could be added perhaps to the plugin portal. This 2nd plugin would register "somewhere" (if it's just schema, perhaps in a layer file, if it's also generated code then some sort of service in the lookup) the old stuff which the Wildfly plugin would pick up and expand its support to all versions.

In the end, maybe this PR shouldn't even have to fix both 1 and 2. I think just getting 1 would be progress.

So how about @ehsavoie just deletes the license-wise non-compliant schemas and updates the PR? Then we have a PR that is acceptable and we can move the plugin from contrib/ back into NetBeans. Later on @ehsavoie can make a PR for whatever hooks are necessary for the companion plugin.

contrib/javaee.wildfly/licenseinfo.xml
```diff
@@ -47,4 +52,26 @@
...
+        <comment type="CATEGORY_B" />
+    </fileset>
+    <fileset>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_14_1.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-app_8_1.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_8_1.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-ejb3-2_1.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_14_1.xsd</file>
+        <license ref="Apache-2.0-ASF" />
+        <comment type="COMMENT_UNSUPPORTED" />
```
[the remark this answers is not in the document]
ehsavoie: No they are not donated to ASF but are under ASL 2.0 license they can be incorporated and also because of this license not donation is required right ?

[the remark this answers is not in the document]
emilianbold: I think `ref="Apache-2.0-ASF"` means owned by Apache Software Foundation.

I guess you can use `ref="Apache-2.0"` for stuff under Apache 2.0 owned by somebody else.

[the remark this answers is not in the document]
matthiasblaesing: The "ref" points to the license in the nbbuild/licenses folder. For each license there can the full text be found. In this case "Apache-2.0-ASF" starts with "Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements." That statement would be wrong for foreign code, so the correct variant is "Apache-2.0".

contrib/javaee.wildfly/licenseinfo.xml
ehsavoie: I'm almost done :)

contrib/javaee.wildfly/licenseinfo.xml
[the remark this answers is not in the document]
ehsavoie: ok, fixed

ehsavoie:
@emilianbold @matthiasblaesing all licenses have been cleared :)

contrib/javaee.wildfly/licenseinfo.xml
```diff
@@ -47,4 +44,23 @@
         <license ref="W3C2" />
         <comment type="CATEGORY_B" />
     </fileset>
+    <fileset>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/datasources_1_4.xsd</file>
+        <license ref="LGPL-2.1" />
```
matthiasblaesing: LGPL-2.1 is Category-X, i.e. we can't use/redistribute it. https://www.apache.org/legal/resolved.html#category-x

contrib/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/xml.xsd
```diff
@@ -1,111 +1,107 @@
 <?xml version='1.0'?>
-<!--
```
matthiasblaesing: Please keep the license information - it makes it easier to reason about the code and if I remember correctly, we have multiple copies of the xml.xsd and they all should have the header.

contrib/javaee.wildfly/licenseinfo.xml
ehsavoie: this is a mistake it should be 1.4.1 which is EL1.0

ehsavoie:
I've found the icon under EPL 2.0 in jboss-tools-server.

ehsavoie:
@matthiasblaesing may you merge this PR please ? So that we can move this plugin back into the main project ?

JaroslavTulach:
Hello Emmanuel, I have troubles building *wildfly* support behind corporate proxy. Probably the goal of "Removing external downloads and updating schemas" wasn't fully fulfilled. My proxy says that http://java.sun.com/xml/ns/javaee/javaee_6.xsd doesn't exists. When I replace:
```diff
--- a/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_6_0_1.xsd
+++ b/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_6_0_1.xsd
@@ -61,7 +61,7 @@
          consistency with the Servlet API. </xsd:documentation>
    </xsd:annotation>
 
-   <xsd:import namespace="http://java.sun.com/xml/ns/javaee" schemaLocation="http://java.sun.com/xml/ns/javaee/javaee_6.xsd"/>
+   <xsd:import namespace="http://java.sun.com/xml/ns/javaee" schemaLocation="javaee_6.xsd"/>
 
    <xsd:group name="descriptionGroup">
       <xsd:sequence>
```
Then my build succeeds. Btw. I did a grep and there are more places that would deserve replacement:
```bash
netbeans$ grep -r xsd.import.*namespace enterprise/javaee.wildfly/src/
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/javaee_6.xsd:  <xsd:import namespace="http://www.w3.org/XML/1998/namespace"
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_14_1.xsd:    <xsd:import namespace="http://xmlns.jcp.org/xml/ns/javaee" schemaLocation="http://www.jboss.org/schema/jbossas/web-app_4_0.xsd"/>
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_6_0_1.xsd:   <xsd:import namespace="http://java.sun.com/xml/ns/javaee" schemaLocation="javaee_6.xsd"/>
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-common_8_1.xsd:   <xsd:import namespace="http://xmlns.jcp.org/xml/ns/javaee" schemaLocation="http://www.jboss.org/schema/jbossas/javaee_8.xsd"/>
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/javaee_8.xsd:  <xsd:import namespace="http://www.w3.org/XML/1998/namespace"
enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/jboss-web_7_3_1.xsd:   <xsd:import namespace="http://java.sun.com/xml/ns/javaee" schemaLocation="http://www.jboss.org/schema/jbossas/web-app_3_0.xsd"/>
```
however when I tried to replace the other, I got missing file errors like `FileNotFoundException: netbeans/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/resources/web-common_4_0.xsd` - I guess there are additional transitive dependencies that need to be fixed. As such I decided to leave a note here rather than fixing just the one xsd file.

emilianbold:
The related question would be why does the build system resolve online
schemas? Builds should run entirely offline.

--emi

PR #2001, opened 2020-03-06
Get a file of all libraries during download

ebarboni:
little file generation to get all binaries-list in one file.
May save lots of time (15 per build) for Maven artefacts generation

related to: https://github.com/apache/netbeans-mavenutils-nb-repository-plugin/pull/2

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
@@ -33,6 +33,8 @@ import java.math.BigInteger;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
```diff
@@ -161,6 +169,13 @@ public void execute() throws BuildException {
                             } else {
                                 success &= fillInFile(hashAndFile[0], hashAndFile[1], manifest, () -> legacyDownload(hashAndFile[0] + "-" + hashAndFile[1]));
                             }
+                            if (report != null) {
+                                Files.write(
+                                        report.toPath(),
+                                        (hashAndFile[0] + ";" + hashAndFile[1] + ";" + include + "\n").getBytes(),
```
matthiasblaesing: Please use the `getBytes(Charset)` variant. Depending on the default file encoding can lead to bugs.

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
```diff
@@ -161,6 +169,13 @@ public void execute() throws BuildException {
...
+                                        StandardOpenOption.APPEND);
```
matthiasblaesing: The append looks strange here (you have a delete / touch before this). Why not just use `StandardOpenOption.CREATE`?

ebarboni: This would lead to having only last binary in the file.

PR #2002, opened 2020-03-06
[NETBEANS-3961] Save project location for reuse for native maven proj…

errael:
In the instiate methods in both java/java.j2seproject/src/org/netbeans/modules/java/j2seproject/ui/wizards/NewJ2SEProjectWizardIterator.java and java/maven/src/org/netbeans/modules/maven/newproject/ArchetypeWizardUtils.java the new project location is saved.

This PR add that to instantiate in IDENativeMavenWizardIterator


matthiasblaesing:
Looks sane to me. Thank you.

PR #2003, opened 2020-03-06
[NETBEANS-2559] Avoid NPE when openjfx

errael:
Older versions of JavaFX required the debugger to set the flag com.sun.javafx.runtime.SystemProperties.isDebug. Openjfx does not have this flag. When debugging openjfx there is an NPE when trying to set the flag. This PR avoid the NPE by just returning if there is not an isDebug flag.


matthiasblaesing:
Looks sane. Thank you.

PR #2005, opened 2020-03-06
[NETBEANS-3968] Fix convert visibility hint for methods

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-3968

1. Fix convert visibility hint for methods

- Show hints for methods when the caret is only on the line that has visibility

2. Lower the priority of convert visibility hint

- When this hint is shown with other hints(e.g. Initialize Fields), we should lower order than the other hints because other hints are more important

Priority

##### Before

nb-3968-priority-before

##### After

nb-3968-priority-after


tmysik:
@junichi11 

Just to be sure - I can merge this PR right now, correct?

junichi11:
@tmysik Yes, you can. Thank You!

oldbasilbear:
I am wondering why I am still (apparently) unable to turn off these hints in 12.4

Can you advise please?

Product Version: Apache NetBeans IDE 12.4
Java: 13.0.7; OpenJDK 64-Bit Server VM 13.0.7+5-Ubuntu-0ubuntu120.04
Runtime: OpenJDK Runtime Environment 13.0.7+5-Ubuntu-0ubuntu120.04
System: Linux version 5.4.0-84-generic running on amd64; UTF-8; en_GB (nb)
User directory: /home/victor/snap/netbeans/common/data/12.4
Cache directory: /home/victor/snap/netbeans/common/cache/12.4

junichi11:
I don't think it's a big problem because it is shown when the caret is only on the line that has visibility.

oldbasilbear:
I agree, it's not a big problem.
Annoying, yes.

PR #2006, opened 2020-03-06
[NETBEANS-445][NETBEANS-2634] Update CSS display property to CSS display module level 3 + add unicode-range

matthiasblaesing:


matthiasblaesing:
Work in progress as @junichi11 noticed, that there is another problem lurking in the CSS parser. The `org.netbeans.modules.css.lib.api.properties.GrammarResolver` should be able resolve "b c" against a grammar "a? && b && c" (the a value is optional), yet it is rejected. The handling of multiplicities in the resolver does not seem to be correct and if I'm not mistaken, there are more problems.

matthiasblaesing:
Ok - I think I managed to fix the issue of the optional parts in the grammar. I did multiple iterations with varying degree of changes. It turns out, that this fix is not perfect, but should work for the current grammars. A general fix turned out to be much harder and I stepped back from trying to this.

I added unittests to verify the current behavior and ensured that all existing unittests still work.

matthiasblaesing:
@junichi11 thank you for the patch - looks sane to me and is a valued addition. I added the patch as a further commit. As unittests come back cleanly, I'll merge it now.

PR #2007, opened 2020-03-07
Update README.md with Build Notes - JDK Parameter Added

ebresie:
This is in regarding to adding JDK parameter to allow build with newer JDKs


README.md
```diff
@@ -63,7 +63,9 @@ Build the NetBeans Platform:
 $ ant -Dcluster.config=platform
 ```
 
-**Note:** You can also use `php`, `enterprise`, etc. See the [cluster.properties](https://github.com/apache/netbeans/blob/master/nbbuild/cluster.properties) file.
+#### Notes:
+* You can also use `php`, `enterprise`, etc. See the [cluster.properties](https://github.com/apache/netbeans/blob/master/nbbuild/cluster.properties) file.
+* It is possible to compiled with newer JDK versions, by adding the ' -Dpermit.jdk9.builds=true' parameter to the ant build expression
```
mklaehn: typo: compile

JaroslavTulach:
Status?

JaroslavTulach:
Ping? Fix the typo, squash the commits to single one, reopen.

PR #2008, opened 2020-03-07
NETBEANS-3254 Remove allowAmbigousCommand follow up

ebarboni:
We make maven works without allowAmbigousCommand  for 11.3. But we did remove command for safety. Now we have time to test


ebarboni:
phase 2 on plan see

PR #2009, opened 2020-03-08
Notifications module long description

blackleg:
Add module long description for notifications module

### DIFF
--- a/ide/notifications/src/org/netbeans/modules/notifications/Bundle.properties
+++ b/ide/notifications/src/org/netbeans/modules/notifications/Bundle.properties
@@ -21,6 +21,7 @@ OpenIDE-Module-Display-Category=Notifications
 OpenIDE-Module-Name=Notifications
 #module short description
 OpenIDE-Module-Short-Description=Notifications pup-ups and Notifications Center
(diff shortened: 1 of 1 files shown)


ide/notifications/src/org/netbeans/modules/notifications/Bundle.properties
```diff
@@ -21,6 +21,7 @@ OpenIDE-Module-Display-Category=Notifications
 OpenIDE-Module-Name=Notifications
 #module short description
 OpenIDE-Module-Short-Description=Notifications pup-ups and Notifications Center
+OpenIDE-Module-Long-Description=Provides IDE notifications pup-ups and Notifications Center
```
ebarboni: is it pop-ups ?

blackleg: It is for show a description in plugin manager: 
Screenshot from 2020-03-14 19-16-38

Before:
Screenshot from 2020-03-14 19-17-11

ebarboni: ok but the words were not good I guess pup up should be pop-up

PR #2015, opened 2020-03-11
[NETBEANS-3342] Fix Netbeans Installer on macOS with Java 8 and 11

oyarzun:
This pull request fixes the NetBean RCP installer on macOS with Java 11 since `com.apple.eawt` classes were removed with [JEP 272](https://openjdk.java.net/jeps/272)

This implementation is modeled after the applemenu implementation and addresses @matthiasblaesing comment  https://github.com/apache/netbeans/pull/1929#issuecomment-583061799 that it needs to work with both Java 8 & 11 and could be accomplished using reflection.

### DIFF
--- a/nbi/engine/build.xml
+++ b/nbi/engine/build.xml
@@ -20,7 +20,7 @@
 
 -->
 
-<project name="NBI_Engine" default="default" basedir=".">
+<project name="nbi/engine" default="default" basedir=".">
     <property file="build.properties"/>
     <property file="${native.dirname}/build.properties"/>
 
@@ -32,6 +32,23 @@
     ======================================================================== -->
     <target name="-post-clean" depends="native-clean"/>
     <target name="-post-compile" depends="native-build,copy-native,create-jar-contents-list,probe"/>
+    
+    <!-- =======================================================================
+        Support JDK 9+ JEP 272: Platform-Specific Desktop Features. 
+    ======================================================================== -->    
+    
+    <import file="../../nbbuild/templates/projectized.xml"/>
+    <available property="desktop.classes.available" classname="java.awt.desktop.AppEvent"/>
+    
+    <target name="-compile-depend" depends="-javac-init-nbjdk,-download.release.files" unless="desktop.classes.available">
+          <mkdir dir="${build.dir}/desktop-classes-src" />
+          <unzip src="external/applemenu-external-desktop-classes-8.2.zip" dest="${build.dir}/desktop-classes-src"/>
+          <mkdir dir="${build.dir}/desktop-classes-classes" />
+          <javac srcdir="${build.dir}/desktop-classes-src" destdir="${build.dir}/desktop-classes-classes"/>
+          
+          <property name="javac.bootclassarg" value="-bootclasspath"/>
+          <property name="javac.bootclasspath" value="${build.dir}/desktop-classes-classes:${nbjdk.bootclasspath}"/>
+    </target>
     
     <!-- =======================================================================
         NetBeans Installer engine specific targets
@@ -145,8 +162,8 @@
         <available property="probe.javac.target" value="1.6" classname="java.lang.Module"/>
         <available property="probe.javac.source" value="1.6" file="${nbjdk.home}/bin/jmod"/>
         <available property="probe.javac.target" value="1.6" file="${nbjdk.home}/bin/jmod"/>
-        <property name="probe.javac.source" value="1.3"/>
-        <property name="probe.javac.target" value="1.1"/>
+        <property name="probe.javac.source" value="1.4"/>
+        <property name="probe.javac.target" value="1.4"/>
         <mkdir dir="${build.classes.dir}/org/netbeans/installer/utils/applications/"/>
         <javac srcdir="probesrc" destdir="${build.classes.dir}/org/netbeans/installer/utils/applications/" source="${probe.javac.source}" debug="true" deprecation="true" target="${probe.javac.target}"/>
     </target>
--- a/nbi/engine/build.xml
+++ b/nbi/engine/build.xml
@@ -24,6 +24,13 @@
     <property file="build.properties"/>
     <property file="${native.dirname}/build.properties"/>
 
+    <!-- prepend bootclasspath for JDK 9+ JEP 272: Platform-Specific Desktop Features stubs -->
+    <condition else="" property="endorsed.classpath" value="${build.dir}/desktop-classes-classes">        
+      <not>
+        <isset property="permit.jdk9.builds"/>
+      </not>
+    </condition>
+
     <import file="nbproject/build-impl.xml"/>
     <import file="${native.dirname}/build.xml"/>
     
@@ -45,9 +52,6 @@
           <unzip src="external/applemenu-external-desktop-classes-8.2.zip" dest="${build.dir}/desktop-classes-src"/>
           <mkdir dir="${build.dir}/desktop-classes-classes" />
           <javac srcdir="${build.dir}/desktop-classes-src" destdir="${build.dir}/desktop-classes-classes"/>
-          
-          <property name="javac.bootclassarg" value="-bootclasspath"/>
-          <property name="javac.bootclasspath" value="${build.dir}/desktop-classes-classes:${nbjdk.bootclasspath}"/>
     </target>
     
     <!-- =======================================================================
--- a/nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
+++ b/nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.installer.wizard.containers;
+
+import com.apple.eawt.Application;
+import com.apple.eawt.ApplicationAdapter;
+import com.apple.eawt.ApplicationEvent;
+
+/**
+ *
+ * @author christian.oyarzun
+ */
+public class InitializeMacJDK8 {
+    public static void initialize(SwingFrameContainer frameContainer)
+    {
+        final Application application = Application.getApplication();
+        if(application == null) {
+            // e.g. running OpenJDK port via X11 on Mac OS X
+            return;
+        }
+        application.removeAboutMenuItem();
+        application.removePreferencesMenuItem();
+        application.addApplicationListener(new ApplicationAdapter() {
+
+            @Override
+            public void handleQuit(ApplicationEvent event) {
+                frameContainer.cancelContainer();
+            }
+        });
+        
+    }
+}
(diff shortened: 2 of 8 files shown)


nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
```diff
@@ -0,0 +1,49 @@
...
+import com.apple.eawt.ApplicationEvent;
+
+/**
+ *
+ * @author christian.oyarzun
+ */
+public class InitializeMacJDK8 {
+    public static void initialize(SwingFrameContainer frameContainer)
+    {
+        final Application application = Application.getApplication();
```
arusinha, line 33: Is this going to throw java.lang.NoClassDefFoundError in  jdk9+ runtime?

nbi/engine/src/org/netbeans/installer/wizard/containers/SwingFrameContainer.java
```diff
@@ -418,23 +419,28 @@ public void actionPerformed(ActionEvent event) {
     
     private void initializeMacOS() {
         if (SystemUtils.isMacOS()) {
-            final Application application = Application.getApplication();
-            if(application == null) {
-                // e.g. running OpenJDK port via X11 on Mac OS X
-                return;
+            if (!initializeMacJDK("org.netbeans.installer.wizard.containers.initializeMacJDK8")) { // NOI18N
```
arusinha, line 422: If there is not much significant difference in look & feel UI , we can then  use Desktop class in JDk8 also instead of Application class.

nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
oyarzun: Yes, this is the same pattern used in the applemenu module.

 [NbApplicationAdapterJDK8.java](https://github.com/apache/netbeans/blob/master/platform/applemenu/src/org/netbeans/modules/applemenu/NbApplicationAdapterJDK8.java)
[NbApplicationAdapterJDK9.java](https://github.com/apache/netbeans/blob/master/platform/applemenu/src/org/netbeans/modules/applemenu/NbApplicationAdapterJDK9.java)

```java
            if (!installAdapter("org.netbeans.modules.applemenu.NbApplicationAdapterJDK8")) {   // NOI18N
                // JDK 8 failed, try JDK 9
                installAdapter("org.netbeans.modules.applemenu.NbApplicationAdapterJDK9");      // NOI18N
            }
```

nbi/engine/src/org/netbeans/installer/wizard/containers/SwingFrameContainer.java
oyarzun: JDK8 Desktop class does not support a QuitHandler.

nbi/engine/build.xml
```diff
@@ -20,10 +20,17 @@
 
 -->
 
-<project name="NBI_Engine" default="default" basedir=".">
+<project name="nbi/engine" default="default" basedir=".">
```
arusinha, line 23: Please don't change the project name. We tried building NB installer for 12.0 after merging your change set, It failed with below error
BUILD FAILED
netbeans/nbbuild/installer/infra/build/build.xml:133: The following error occurred while executing this line:
nbbuild/installer/infra/build/nbi_all/nbi/engine/build.xml:47: Cannot find /Users/sarillamohanarao/Mohan/Code_Repository/Github_Repository/Installer_Maven_Issue/netbeans/nbbuild/installer/infra/build/nbi_all/nbbuild/templates/projectized.xml imported from /Users/sarillamohanarao/Mohan/Code_Repository/Github_Repository/Installer_Maven_Issue/netbeans/nbbuild/installer/infra/build/nbi_all/nbi/engine/build.xml

nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
arusinha: can you enclose the code snippet with try-catch as it will throw the error on jdk9+ runtime

nbi/engine/src/org/netbeans/installer/wizard/containers/SwingFrameContainer.java
arusinha, line 422: Do we really needs reflection to call the classes?
you can change InitializeMacJDK8.initialize() to return true/false

if(!InitializeMacJDK8.initialize())
   InitializeMacJDK9.initialize()


other alternative approach

boolean isJdk9OrAbove = false
  try {
         SourceVersion.valueOf("RELEASE_9"); //NOI18N
         isJdk9OrAbove = true;
        }
		catch (IllegalArgumentException ex) {
          isJdk9OrAbove = true;
        }
		if(isJdk9OrAbove)
		 (InitializeMacJDK9.initialize())
		 else InitializeMacJDK9.initialize()

ebarboni:
@arusinha  @oyarzun can we merge this ? I cannot check that (no macos)

arusinha:
@ebarboni , there are some comments which needs to be resolved, PR is yet not approved

ebarboni:
ok we can redirect to 12.1  or is this important for 12.0

nbi/engine/src/org/netbeans/installer/wizard/containers/SwingFrameContainer.java
oyarzun: I tried changing the code to not use reflection and wrapping the initialize in a try catch.

```java
public class InitializeMacJDK8 {
    public static boolean initialize(SwingFrameContainer frameContainer)
    {
        try {
            final Application application = Application.getApplication();
            if(application == null) {
                // e.g. running OpenJDK port via X11 on Mac OS X
                return true;
            }
            application.removeAboutMenuItem();
            application.removePreferencesMenuItem();
            application.addApplicationListener(new ApplicationAdapter() {

                @Override
                public void handleQuit(ApplicationEvent event) {
                    frameContainer.cancelContainer();
                }
            });

            return true;
        }
        catch (NoClassDefFoundError e) {
            // we are running Java 9+
        }
        return false;
    }
}
```
This did not work when running the installer on Java 11
```
2020-05-23 15:46:45.507]:     An unexpected exception happened in thread main
[2020-05-23 15:46:45.507]:     java.lang.NoClassDefFoundError: com/apple/eawt/ApplicationListener
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.wizard.containers.SwingFrameContainer.initializeMacOS(SwingFrameContainer.java:422)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.wizard.containers.SwingFrameContainer.initComponents(SwingFrameContainer.java:355)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.wizard.containers.SwingFrameContainer.<init>(SwingFrameContainer.java:165)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.wizard.Wizard.newWizardContainer(Wizard.java:479)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.wizard.Wizard.open(Wizard.java:509)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.Installer.start(Installer.java:124)
[2020-05-23 15:46:45.508]:     	at org.netbeans.installer.Installer.main(Installer.java:61)
[2020-05-23 15:46:45.508]:     Caused by: java.lang.ClassNotFoundException: com.apple.eawt.ApplicationListener
[2020-05-23 15:46:45.508]:     	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
[2020-05-23 15:46:45.508]:     	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
[2020-05-23 15:46:45.508]:     	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
[2020-05-23 15:46:45.508]:     	... 7 more
[2020-05-23 15:46:45.508]:     ... show message dialog
[2020-05-23 15:46:45.508]:         title: Critical Error
[2020-05-23 15:46:45.508]:         message: An unexpected exception happened in thread main
[2020-05-23 15:46:45.508]:
[2020-05-23 15:46:45.509]:         Exception:
[2020-05-23 15:46:45.509]:           java.lang.NoClassDefFoundError:
[2020-05-23 15:46:45.509]:           com/apple/eawt/ApplicationListener
```

nbi/engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java
oyarzun: The try-catch is in the initializeMacJDK method

nbi/engine/build.xml
oyarzun: I had to change the project name to get `projectized.xml` to work. This was needed to download the stubs to compile `InitializeMacJDK9` with JDK8. 

The build error above is not due to changing the project name, instead it is due to the fact that the relative path does not work when building the NB installer.

```xml
```

Do you know how the project can download the stubs `applemenu-external-desktop-classes-8.2.zip` without using `projectized.xml`?

arusinha: @oyarzun , indeed import of projectized.xml is causing the issue. I currently don't have solution for this issue. Is PR is urgent or we can wait for 12.1, as changing something in NB installer script is risky at this moment as we are near to NB 12.0 release.

oyarzun: @arusinha & @ebarboni It can wait until 12.1 since this has been broken since version 10. Plus there are ways around it, create one installer for Java 8 and another for Java 11+. 

I do not use this feature and only started looking at it since we switched our RCPs builds to use maven. In maven there is no goal to build a macOS application like there is with ant.

https://issues.apache.org/jira/browse/NETBEANSINFRA-37

oyarzun: @arusinha 

I've found a work around that allows the NB installer to be built. Since it is not used for macOS we don't need the mac specific classes that are loaded by reflection. Create a copy of the original nbi/engine build.xml in nbbuild/installer/infra/build/build-nbi-engine.xml then patch it's build.xml as follows.

```diff
diff --git a/nbbuild/installer/infra/build/build.xml b/nbbuild/installer/infra/build/build.xml
index 4861fc7092..94134b2634 100644
--- a/nbbuild/installer/infra/build/build.xml
+++ b/nbbuild/installer/infra/build/build.xml
@@ -71,8 +71,13 @@
                 <exclude name="infra/server/**/build/**/*.*"/>
                 <exclude name="infra/server/**/dist/**/*.*"/>
                 <exclude name="infra/server/**/private/**/*.*"/>
+                <!-- macOS uses native installer so ignore initialize macOS classes loaded with reflection -->
+                <exclude name="engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK8.java"/>
+                <exclude name="engine/src/org/netbeans/installer/wizard/containers/InitializeMacJDK9.java"/>
             </fileset>
         </copy>
+        <!--replace nbi/engine build.xml with one that does not build the initialize macOS classes -->
+        <copy file="build-nbi-engine.xml" tofile="${nbi.core.dir}/engine/build.xml" overwrite="true"/>
         <copy todir="${nbi.netbeans.dir}">
             <fileset dir="${hg.root}/installer">
                 <exclude name="infra/server/**/build/**/*.*"/>
```

arusinha: @oyarzun , separate NB installer is released each for Linux/Mac/Windows.  
Conditional import can come handy.
Possible solution can be 
<property name="IS_NB_INSTALLER" value="true" />(this property has to be defined in installer/infra/build/build.xml)


nbi/engine/build.xml  changes
<condition property="relative_path" value="../../../../" else="../../">
 <isset property="IS_NB_INSTALLER"/>
        </condition>
            <import file="${relative_path}/nbbuild/templates/projectized.xml"/>

oyarzun: @arusinha I like the idea of a property, but I tried changing the relative path. I ran into the same problem with projectized.xml which made me change the project name. The new path causes this check to fail when building the NB installer.

```xml
    <property name="ant.file.1" location="${ant.file}"/>
    <property name="ant.file.2" location="${nb_all}/${ant.project.name}/build.xml"/>
    <fail message="Please keep the Ant project name the same as the project's directory name: ${ant.file.1} vs. ${ant.file.2}">
        <condition>
            <not>
                <equals arg1="${ant.file.1}" arg2="${ant.file.2}"/>
            </not>
        </condition>
    </fail>
```

arusinha: @oyarzun I am working on some other things, will revisit this issue once those are done

oyarzun: @arusinha when you have time to circle back to this, I have a new branch that has this fix working without using `projectized.xml`. 

https://github.com/oyarzun/netbeans/commit/cba705bde9c76c0a65cbd5e2dfaac757660d92fe

Let me know if you want me to submit a new PR.

arusinha: John  is now managing mac NB Installer. @mcdonnell-john request you to review the changes.

PR #2016, opened 2020-03-11
[NETBEANS-4002] - clean Iterator raw types

BradWalker:
This change removes the majority of these warnings..

This bug implements changes that clean a whole lot of Iterator raw type warning messages from the compiler..

   [repeat] /home/bwalker/src/netbeans/java/xml.tools.java/src/org/netbeans/modules/xml/tools/java/generator/ElementBindings.java:156: warning: [rawtypes] found
 raw type: Iterator
   [repeat]         Iterator it = values().iterator();
   [repeat]         ^

This change removes the majority of these warnings..


matthiasblaesing:
Looks sane to me. Thank you.

PR #2020, opened 2020-03-16
[NETBEANS-3986] Creating New Class On Pasting Class Text

Akshay-Gupta-Oracle:
This PR supports copying raw java text from anywhere(inside or outside of Netbeans) and pasting directly to a package to create a new class automatically. Copied text should have a valid syntax and should start with a **public class** followed by class name.

### DIFF
--- a/java/java.project.ui/src/org/netbeans/spi/java/project/support/ui/CreateClassAndCopy.java
+++ b/java/java.project.ui/src/org/netbeans/spi/java/project/support/ui/CreateClassAndCopy.java
@@ -1,0 +1,172 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.spi.java.project.support.ui;
+
+import java.awt.Toolkit;
+import java.awt.datatransfer.Clipboard;
+import java.awt.datatransfer.DataFlavor;
+import java.awt.datatransfer.UnsupportedFlavorException;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Set;
+import java.util.StringTokenizer;
+import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


java/java.project.ui/src/org/netbeans/spi/java/project/support/ui/CreateClassAndCopy.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent ev) {
+        try {
+            Clipboard c = Toolkit.getDefaultToolkit().getSystemClipboard();
+            if(!c.isDataFlavorAvailable(DataFlavor.stringFlavor))return;
+            boolean containsClass = false;
+            String data = (String) c.getData(DataFlavor.stringFlavor);
+            String className = "";
```
lahodaj: I wonder if it wouldn't be easier to let some (Java) parser parse the class, and analyze the output. Something as simple as:
((JavacTask) javax.tools.ToolProvider.getSystemJavaCompiler().getTask(...)).parse()
might give much better outcomes, and is unlikely to be problematic w.r.t. performance (parse is typically very fast). The only parameter that really needs to be filled in for getTask is the last one, and it may be a singleton list with a very simple JavaFileObject, like this one:
https://github.com/apache/netbeans/blob/5f675e88a3ae3fee652332f9faf2f1985508c272/java/lib.nbjavac/test/unit/src/org/netbeans/lib/nbjavac/services/NBClassWriterTest.java#L61

(Passing the "out" Writer and passing the output to "/dev/null" is probably a good idea, as well, but not needed for the functionality to work.)

As far as I can tell, the code here won't work if:
-there are any imports
-or annotations
-or the class is non-public
-or if the thing that is being pasted is an interface, enum, annotation type (or record)
-or if there's any other modifier for the class (not a big factor right now, as probably only "strictfp" could be here).

This feels unnecessarily limited.

java/java.project.ui/src/org/netbeans/spi/java/project/support/ui/CreateClassAndCopy.java
```diff
@@ -0,0 +1,172 @@
...
+                JOptionPane.showMessageDialog(null, "Code not valid to create class");
+                return;
+            }
+            Set<FileObject> files = this.context.files();
+            if (files.size() != 1) {
+                return;
+            }
+            String path = files.iterator().next().getPath();
+            String packageName = "";
+            File f = new File(path + "\\" + className + ".java");// NOI18N
```
lahodaj: Using:
            File f = new File(path + "\\" + className + ".java");// NOI18N
will clearly not work on Linux, and probably not on Mac - the file separator for these is '/'. I would suggest to simply using FileObject all the way, unless there is a really compelling reason to not to.

java/java.project.ui/src/org/netbeans/spi/java/project/support/ui/CreateClassAndCopy.java
```diff
@@ -0,0 +1,172 @@
...
+            if (!f.createNewFile()) {
+                JOptionPane.showMessageDialog(null, "Can not create file");
+                return;
+            }
+            ExplorerManager explorerManager = findExplorerManager();
+            if (explorerManager.getSelectedNodes().length != 1) {
+                return;
+            }
+            Node selectedNode = explorerManager.getSelectedNodes()[0];
+            while (!selectedNode.getName().equals("${src.dir}")) {// NOI18N
```
lahodaj: This feels really weird - I don't think anyone ever said anything about the names of the Nodes. Suggest to use ClassPath.getClassPath(SOURCE, target).getResourceName(target) (handling failures along the way).

ebarboni:
my bad did a merge a few hour later. I did a revert. I think this PR has to be recreated.

PR #2028, opened 2020-03-17
Sync master to release 12.0 beta2

ebarboni:
the lotr PR to catch all master commit to the release120


ebarboni:
@jkovalsky @neilcsmith-net  this PR will contains the delta between 12.0 and master

eirikbakke:
I'm sorry to have caused extra work with that stray commit here... I did the same mistake last month and in my head I thought there would be 3 months until the master branch was closed again.

There's no strong reason for this particular commit to be in 12.0--feel free to drop it, or to include it.

ebarboni:
I though it was just a visual display ?

This PR will grow and contains PR to catch all differences, revert should be done on master and will dissapears from here.

GeertjanWielenga:
Indeed, this is about visual display in Plugin Manager only unless I am mistaken.

neilcsmith-net:
As @ebarboni says, this is a sync PR to merge between master and release branch, just like we've done throughout the last 3 releases!  It's not specific to @eirikbakke commit - and I for one am +1 for that change staying there, if not how it was achieved. :smile:

eirikbakke:
> I though it was just a visual display?
Yes, this change does not actually deactivate the SVG Loader module, it just prevents it from showing up as a separate plugin in the "Installed" tab in the "Plugins" dialog on NetBeans Platform applications.

(Also note that the SVG Loader module isn't actually used yet in the IDE--the few scalable icons you may be seeing already are custom-painted. Right now the only SVG icons are the in-menu icons for the Undo/Redo actions, for testing purposes. More will come in the future.)

PR #2031, opened 2020-03-18
[NETBEANS-3942] FlatLaf: fix editor annotations when switching LaF and switching color profile too on restart

DevCharly:
see [NETBEANS-3942](https://issues.apache.org/jira/browse/NETBEANS-3942?focusedCommentId=17061871&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17061871) for details

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/options/LafPanel.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/options/LafPanel.java
@@ -27,6 +27,7 @@ import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.prefs.Preferences;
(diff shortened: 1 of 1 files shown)


platform/core.windows/src/org/netbeans/core/windows/options/LafPanel.java
```diff
@@ -296,7 +297,11 @@ private void switchEditorColorsProfile() {
         try {
             Class klz = cl.loadClass( COLOR_MODEL_CLASS_NAME );
             Object colorModel = klz.newInstance();
-            Method m = klz.getDeclaredMethod( "setCurrentProfile", String.class ); //NOI18N
+            Method m = klz.getDeclaredMethod( "getAnnotations", String.class ); //NOI18N
```
JaroslavTulach: Wow! That is ugly. Isn't the UI misplaced when it needs to use reflection to load a class from other module? If not, then ideally there would be an SPI interface `CurrentProfile` and one would use `Lookup.getDefault().lookupAll(CurrentProfile.class).forEach(it -> it.setProfile(preferredProfile))`.

lkishalmi: That's true, filed [NETBEANS-4039](https://issues.apache.org/jira/browse/NETBEANS-4039) for this. And yes, we blindly followed the pattern was used before us.

PR #2035, opened 2020-03-19
[NETBEANS-3957] fixed "opened documents list" show HTML code instead of rich text

DevCharly:
https://issues.apache.org/jira/browse/NETBEANS-3957

tested with FlatLaf light/dark, Windows, Nimbus and Metal LaFs

### DIFF
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/popupswitcher/SwitcherTable.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/popupswitcher/SwitcherTable.java
@@ -161,7 +161,7 @@ public class SwitcherTable extends JTable {
         if (ren instanceof JLabel) {
             // #199007: Swing HTML renderer does a poor job of truncating long labels
             JLabel prototype = (JLabel) ren;
-            lbl = HtmlRenderer.createLabel();
+            lbl = (JLabel) HtmlRenderer.createRenderer();
             if( lbl instanceof HtmlRenderer.Renderer ) {
                 ((HtmlRenderer.Renderer)lbl).setRenderStyle( HtmlRenderer.STYLE_TRUNCATE );
             }


platform/o.n.swing.tabcontrol/src/org/netbeans/swing/popupswitcher/SwitcherTable.java
```diff
@@ -161,7 +161,7 @@ public Component prepareRenderer(
         if (ren instanceof JLabel) {
             // #199007: Swing HTML renderer does a poor job of truncating long labels
             JLabel prototype = (JLabel) ren;
-            lbl = HtmlRenderer.createLabel();
+            lbl = (JLabel) HtmlRenderer.createRenderer();
```
matthiasblaesing: I think this could be improved if lbl would be declared to be a  `JComponent`. Most calls will still work, and the cases where that is not the case (calls to setIcon/setText) are already null protected and could just as well be protected with `if lbl instanceof JLabel)` and only cast at that point.

matthiasblaesing:
Can you give instructions for reproduction? I tested on windows Nimbus LAF was tested on:

```
Product Version: Apache NetBeans IDE DEV (Build dev-96c12bc480bb621f05e40978fced87f99768873f)
Java: 13.0.2; OpenJDK 64-Bit Server VM 13.0.2+8
Runtime: OpenJDK Runtime Environment 13.0.2+8
```

```
Product Version: Apache NetBeans IDE DEV (Build dev-96c12bc480bb621f05e40978fced87f99768873f)
Java: 1.8.0_232; OpenJDK 64-Bit Server VM 25.232-b09
Runtime: OpenJDK Runtime Environment 1.8.0_232-b09
```

DevCharly:
You have to enable "Sort opened documents list by project" checkbox in "Options > Appearance > Document Tabs"  (at the bottom).

Then the documents popup (down arrow at right side of editor tabs) includes project names (in bold):

image

If you move the mouse over a project name it changes from bold text to HTML text.
There are more screenshots in the JIRA issue.

DevCharly:
I can now confirm that the problem is caused by PR https://github.com/apache/netbeans/pull/1990

When I checkout the commit before PR 1990 was merged (and compile the "UI Utilities API" project) it works fine.

When I checkout the commit where PR 1990 was merged it shows HTML text.

platform/o.n.swing.tabcontrol/src/org/netbeans/swing/popupswitcher/SwitcherTable.java
matthiasblaesing: Ignore this - it is not cleaner, the return of `HtmlRenderer#createRenderer` does not allow this.

matthiasblaesing:
@ebarboni please check if you agree, that this is good to be merged to master+12.0

PR #2044, opened 2020-03-23
Fixed class cast exception when customzing the URLDisplayer

EnricoScantamburlo:
We created a class that extended URLDisplayer and used theNbURLDisplayer as a delegate, but the got a ClassCastException. This is an attempt to fix it.


platform/o.n.core/src/org/netbeans/core/NbURLDisplayer.java
```diff
@@ -105,6 +105,11 @@ private void warmBrowserUp( boolean externalBrowser ) {
         }
     }
 
+    @Override
+    public void clearCache() {
+        super.clearCache(); //To change body of generated methods, choose Tools | Templates.
+    }
+
```
mklaehn: Why do you override a method that is doing nothing except call its super - which in turn does nothing?
Maybe setting htmlViewer to null should have been done here?

EnricoScantamburlo: You are right, my change was not uploaded, fixing it right now

platform/openide.awt/src/org/openide/awt/HtmlBrowser.java
```diff
@@ -965,6 +965,14 @@ public static URLDisplayer getDefault() {
...
             showURL(u);
         }
+        
+        /**
+         * resets the caches used
+         * If the default browser is changed this clears the instantiated objects for the old one
+         */
+        public void clearCache(){
+            
+        }
```
mklaehn: This is a new method on a public API interface. As such it will propably require
* an entry in the api changes XML file,
* the module version needs to be increased and
* the javadoc will need an appropriate `@since` tag.

@matthiasblaesing did I miss something?

platform/openide.awt/src/org/openide/awt/HtmlBrowser.java
```diff
@@ -965,6 +965,14 @@ public static URLDisplayer getDefault() {
         public void showURLExternal(URL u) {
             showURL(u);
         }
+        
+        /**
+         * resets the caches used
+         * If the default browser is changed this clears the instantiated objects for the old one
+         */
+        public void clearCache(){
```
mklaehn: Nitpick: usually there's an whitespace before the curly brace

platform/openide.awt/src/org/openide/awt/HtmlBrowser.java
matthiasblaesing: I think you covered everything. I think the only thing missing is `org.netbeans.core` needs to update its dependency on `org.openide.awt` to the newly set version.

lkishalmi:
@EnricoScantamburlo What shall we do with this one?

EnricoScantamburlo:
What shall I do?

Il giorno dom 20 dic 2020 alle ore 02:41 Laszlo Kishalmi <
notifications@github.com> ha scritto:

matthiasblaesing:
@EnricoScantamburlo there are open comments:
https://github.com/apache/netbeans/pull/2044#discussion_r396772981
https://github.com/apache/netbeans/pull/2044#discussion_r396773002

EnricoScantamburlo:
please tell me if it is as expected

matthiasblaesing:
We need an issue to track this contribution. You need to be aware, that contributors are required to donate their changes to the ASF (see statement at the top of most of the source files). The regular way is to get an ICLA on file, for smaller contributions, we accept if you acknowledge that fact explicitly and the issue is used to track that acknowledgement.

The issue tracker is: https://issue.apache.org/jira - create an issue for this and then update this PR in the following way:

1. Update the title of this PR to read: "[NETBEANS-XXXXX] Fixed class cast exception when customzing the URLDisplayer" (XXXXX is the issue number assigned by the issue tracker)
2. Squash all your commits into one and update the commit message so that the summary line (first line) reads "[NETBEANS-XXXXXX] <Summary>", then leave an empty line and potentially add a longer description if necessary. Force push the combined commit
3. Wait until jira picks up your commit (a bot takes care of that)
4. Acknowledge, that you donate the change to the ASF

For the change itself, the entry to `apichanges.xml` reads strange. It gives a motiviation, but this is not the intention of the file, you describe the past/problem, but what is required is the description of the change. If the issue described above is created, this can be referenced in addition (again see the other entries).

PR #2046, opened 2020-03-25
[NETBEANS-2415]: Bring back support for WildFly server

ehsavoie:
* Moving WildFly plugin back into enterprise cluster
* Fixing missing DTDs form xml.xsd

Jira: https://issues.apache.org/jira/browse/NETBEANS-2415


ebarboni:
Looks good,  all license issue were fixed

matthiasblaesing:
Please use the `datatypes.dtd` and `XMLSchema.dtd` files from:

- `ide/xml.catalog/src/org/netbeans/modules/xml/catalog/resources/datatypes.dtd`
- `ide/xml.catalog/src/org/netbeans/modules/xml/catalog/resources/XMLSchema.dtd`

Both have the W3C license attached, so that we don't need to discuss license headers everytime. That way all four copies in the source tree of the files will be identical.

Apart from that - great.

ehsavoie:
@ebarboni @matthiasblaesing I've referenced the proper XMLSchema.dtd so that I don't need to have XMLSchema.dtd and datatypes.dtd in the license.info

matthiasblaesing:
Sorry, I don't like it. This introduces a cross-cluster dependency, that is not recorded in the project setup. The benefit is minimal and the potential cost is huge. Loading the file at runtime will also most probably fail, as the relative path does not take into account the runtime location of the files.

ehsavoie:
@matthiasblaesing 
ok I copied the files into the wildfly plugin.

matthiasblaesing:
I hope the last nitpick: `src/org/netbeans/modules/javaee/wildfly/resources/xml.xsd` has still the cross-module reference

PR #2047, opened 2020-03-26
[NETBEANS-4070] fixed JavaDoc in tooltip (ctrl+mouse hover) duplicate…

dmochalov:
See the [issue ]( https://issues.apache.org/jira/browse/NETBEANS-4070) for details.
Simple code changes: moved if block `if (element.getKind() == ElementKind.METHOD) {` outside of `for (DocTree tag : tags) {` loop and added some comments.

### DIFF
--- a/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
+++ b/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
@@ -1020,35 +1020,41 @@ public class ElementJavadoc {
                     }
                     break;
                 }
-                if (element.getKind() == ElementKind.METHOD) {
-                    if (!ret.containsKey(DocTree.Kind.RETURN)) {
+            }
+
+            if (element.getKind() == ElementKind.METHOD) {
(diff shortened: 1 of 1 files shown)


lkishalmi:
I think this one is safe to be merged in 12.0

eirikbakke:
What is the behavior in Javadoc generated by the javadoc tool? Does it do the same?

(One could argue that the superclass description _should_ be included, as the subclass is bound to follow the requirements set there. But I would say, do whatever the behavior of the javadoc tool is.)

dmochalov:
@eirikbakke thank you for taking time to review it.

Indeed Javadoc handles it differently. It does not include the Javadoc of a parent method. Also, I checked the Eclipse's tooltip - it doesn't include the parent Javadoc at all.
But in the NetBeans, there is a whole bunch of code to handle it. The bug here is just how NetBeans combines a description of parameters.
Before fixing NetBeans dose it in that way:

image

i.e for the index it combines Javadoc both from parent and from the current method and does it a little awkwardly.
for the c parameter, it only adds Javadoc from the parent method.

After fixing NetBeans dose it in that way:
image

PR #2048, opened 2020-03-26
[NETBEANS-3787] FlatLaf: fixed active tab painting if project tab colors enabled

DevCharly:
Fix for https://issues.apache.org/jira/browse/NETBEANS-3787

When project tab colors are enabled in Options dialog (Appearance > Document tabs), then it is painted over the blue "underline" that FlatLaf uses for active tabs. To avoid this I've moved the project color bar to the top of the tab.

Before:
image

After:
image

Other LaFs are not affected. E.g Windows:
image

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
@@ -161,8 +161,16 @@ public class ProjectColorTabDecorator extends TabDecorator {
         }
         g.setColor( c );
         Rectangle rect = new Rectangle( tabRect );
-        rect.y += rect.height - 3;
-        rect.grow( -1, -1 );
+        int underlineHeight = UIManager.getInt("nb.multitabs.underlineHeight"); // NOI18N
+        if( underlineHeight > 0 ) {
+            // if the selected tab is highlighted with an "underline" (e.g. in FlatLaf)
+            // then paint the project color bar at the top of the tab
+            rect.height = underlineHeight;
+        } else {
+            // bottom project color bar
+            rect.y += rect.height - 3;
+            rect.grow( -1, -1 );
+        }
         g.fillRect( rect.x, rect.y, rect.width, rect.height );
     }
 


platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
```diff
@@ -161,8 +161,16 @@ public void paintAfter( TabData tab, Graphics g, Rectangle tabRect, boolean isSe
         }
         g.setColor( c );
         Rectangle rect = new Rectangle( tabRect );
-        rect.y += rect.height - 3;
-        rect.grow( -1, -1 );
+        int underlineHeight = UIManager.getInt("nb.multitabs.underlineHeight"); // NOI18N
```
eirikbakke: Looking at FlatLaf.properties ( https://github.com/apache/netbeans/blob/3139141059141caab4a6af1158cce7131e1b40f9/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLaf.properties ), the definition is "nb.multitabs.underlineHeight={scaledInteger}3". What does "scaledInteger" mean? How does that get parsed here?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
```diff
@@ -161,8 +161,16 @@ public void paintAfter( TabData tab, Graphics g, Rectangle tabRect, boolean isSe
         }
         g.setColor( c );
         Rectangle rect = new Rectangle( tabRect );
-        rect.y += rect.height - 3;
```
eirikbakke: Realizing this code was already there... what's going on here? Is tabRect the entire tab to start with?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
```diff
@@ -161,8 +161,16 @@ public void paintAfter( TabData tab, Graphics g, Rectangle tabRect, boolean isSe
...
-        rect.grow( -1, -1 );
+        int underlineHeight = UIManager.getInt("nb.multitabs.underlineHeight"); // NOI18N
+        if( underlineHeight > 0 ) {
+            // if the selected tab is highlighted with an "underline" (e.g. in FlatLaf)
+            // then paint the project color bar at the top of the tab
+            rect.height = underlineHeight;
```
eirikbakke: If an underline is used to indicate the active tab, why not just keep the whole tab colored, like the other tabs, with the underline added? E.g.

rect.height - =underlineHeight?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
eirikbakke: Sorry, stray comment... I got it.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
DevCharly: `{scaledInteger}` means that FlatLaf scales this value (when initializing the Laf) using the FlatLaf user scale factor, which is used for Java 8 but also on Linux. When running in Java 9 the FlatLaf user scale factor is usually 1.

In other places I usually invoke method `UIScale.scale( intValue )`, but in this case it would require that Multi-tabs project depends on FlatLaf library project, which I wanted to avoid.

The downside of using `{scaledInteger}` is that it is not updated when the user changes scaling in Windows Settings.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
lkishalmi: @eirikbakke That would look better. Thanks Eirik!

DevCharly:
> Might look nicer to keep the entire tab colored

Yea, I tried this.

For **FlatLaf Light** it works good when all tabs have same color (only one project). When having multiple projects and mixed tab colors it is a little bit more difficult to spot the active tab. But I think its acceptable.

image

image

For **FlatLaf Dark** it is unusable because the project tab colors are darker and the underline is very hard to see.

image

If another view is focused and the underline becomes gray it is even harder to see:

image

image

image

I think that the dark project colors are too bright compared with the dark background.
Will try to make the project colors darker. Hope this makes the underline more visible...

ebarboni:
Is conversation over and can this be merged ?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ProjectColorTabDecorator.java
eirikbakke: Ah, so I get it's FlatLAF that reads the property file and puts the scaled value into UIManager. Sounds good.

> The downside of using {scaledInteger} is that it is not updated when the user changes scaling in Windows Settings.
Presumably it's scaled via the default Graphics2D transform in that case. Sounds correct in any case.

eirikbakke:
Seems fine to merge.

On the color-the-whole-tab-vs.-just-the-top discussion, feel free to merge whichever one you prefer for now. Easy to revisit later.

PR #2049, opened 2020-03-27
Contributing end-of-line switcher.

jlahoda:


DevCharly:
Support for "Mac OS 9 (CR)" could be removed IMHO.
Mac OS 9 is 20 years old. I don't think that NetBeans or Java 8 runs on this OS.

matthiasblaesing:
@DevCharly I would merge as is - that way all the three big groups are covered (\n, \r\n, \r). It does not matter where NetBeans runs, but what files are edited and I can image file formats stuck with old mac conventions.

svenreimers:
So, should we add 12.0 label or postpone and select 12.1?

eirikbakke:
There are plenty of CSV files with \r line endings around--probably due to Excel for Mac generating them (the last time I checked).

eirikbakke:
This will be really useful, by the way... my git repo is configured to require UNIX line endings, but NetBeans always creates new files with Windows line endings, which I have been converting with the dos2unix tools. (At some point I went through the various git settings and decided it was safest to do it this way rather than use autocrlf--I frequently switch between MacOS, Windows, WSL, Cygwin etc.)

In the future, it would also be good to detect the line ending configuration of a git repository, and use that for all new files.

jlahoda:
Regarding 12.0 vs 12.1, I'll leave it to Eric, I don't have a strong opinion either way.

PR #2051, opened 2020-03-28
[NETBEANS-3428] FlatLaf: hover in multi-tabs

DevCharly:
Addition to PR #1865.
Adds the missing hover effect to multi-tabs for FlatLaf.
Other LaFs are not changed. Tested with Metal, Nimbus and Windows LaF.

image

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
@@ -110,7 +110,6 @@ public class TabDataRenderer implements TableCellRenderer {
             renderer.tabData = tab;
             renderer.isSelected = isSelected;
             renderer.isActive = isActive;
-            renderer.isHover = isHover;
             renderer.tabsLocation = (table instanceof TabTable) ? ((TabTable)table).getTabsLocation() : JTabbedPane.TOP;
 
             if( table instanceof TabTable ) {
@@ -196,7 +195,6 @@ public class TabDataRenderer implements TableCellRenderer {
         private TabData tabData;
         private boolean isSelected;
         private boolean isActive;
-        private boolean isHover;
         private int tabsLocation = JTabbedPane.TOP;
 
         public RendererPanel() {
@@ -255,10 +253,8 @@ public class TabDataRenderer implements TableCellRenderer {
             }
 
             // paint tab decorators
-            if (!isHover) {
-                for( TabDecorator td : decorators ) {
-                    td.paintAfter( tabData, g, rect, isSelected );
-                }
+            for( TabDecorator td : decorators ) {
+                td.paintAfter( tabData, g, rect, isSelected );
             }
         }
 
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
@@ -58,6 +58,7 @@ public class TabDataRenderer implements TableCellRenderer {
     private static final Color underlineColor = UIManager.getColor("nb.multitabs.underlineColor"); // NOI18N
     private static final Color inactiveUnderlineColor = UIManager.getColor("nb.multitabs.inactiveUnderlineColor"); // NOI18N
     private static final Color activeBackground = UIManager.getColor("nb.multitabs.activeBackground"); // NOI18N
+    private static final Color hoverBackground = UIManager.getColor("nb.multitabs.hoverBackground"); // NOI18N
 
     private final RendererPanel renderer = new RendererPanel();
     private final List<TabDecorator> decorators = getDecorators();
@@ -97,6 +98,10 @@ public class TabDataRenderer implements TableCellRenderer {
                     icon = i;
                 }
             }
+            boolean isHover = (hoverBackground != null && TabTableUI.isHover(table, row, column));
+            if (isHover) {
+                colBackground = hoverBackground;
+            }
             renderer.label.setText( text );
             renderer.label.setIcon( icon );
             renderer.label.setFont( table.getFont() );
@@ -105,6 +110,7 @@ public class TabDataRenderer implements TableCellRenderer {
             renderer.tabData = tab;
             renderer.isSelected = isSelected;
             renderer.isActive = isActive;
+            renderer.isHover = isHover;
             renderer.tabsLocation = (table instanceof TabTable) ? ((TabTable)table).getTabsLocation() : JTabbedPane.TOP;
 
             if( table instanceof TabTable ) {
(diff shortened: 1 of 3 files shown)


platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
```diff
@@ -58,6 +58,7 @@
     private static final Color underlineColor = UIManager.getColor("nb.multitabs.underlineColor"); // NOI18N
     private static final Color inactiveUnderlineColor = UIManager.getColor("nb.multitabs.inactiveUnderlineColor"); // NOI18N
     private static final Color activeBackground = UIManager.getColor("nb.multitabs.activeBackground"); // NOI18N
+    private static final Color hoverBackground = UIManager.getColor("nb.multitabs.hoverBackground"); // NOI18N
```
eirikbakke: Other LAFs, e.g. Windows, have a hover color as well. Why does FlatLAF need a new property here?

If the new property is needed, should the other LAFs be updated to use it as well?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
```diff
@@ -248,8 +255,10 @@ public void paint( Graphics g ) {
             }
 
             // paint tab decorators
-            for( TabDecorator td : decorators ) {
-                td.paintAfter( tabData, g, rect, isSelected );
+            if (!isHover) {
```
eirikbakke: Why are tab decorators not invoked in the isHover case? TabDecorator is a public API, so imagine having to describe this behavior in its Javadoc: "if the UIManager property nb.multitabs.hoverBackground is defined, then paintAfter is only called when the tab is not hovered over". That would be a really weird caveat.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabTableUI.java
```diff
@@ -173,6 +193,28 @@ public void mouseDragged( MouseEvent e ) {
...
+                }
+
+                if (hoverRow != -1 && hoverColumn != -1) {
+                    table.repaint(table.getCellRect(hoverRow, hoverColumn, true));
+                }
+                if (row != -1 && column != -1) {
+                    table.repaint(table.getCellRect(row, column, true));
+                }
+                hoverRow = row;
+                hoverColumn = column;
```
eirikbakke: Don't you need to set these _before_ calling repaint? The repaint typically ends up being scheduled for later, but it's cleaner to not assume this.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
DevCharly: > Other LAFs, e.g. Windows, have a hover color as well.

There are two implementation for editor tabs in NB:
- "Tab Control" (package org.netbeans.swing.tabcontrol)
- "Multi-tabs" (package org.netbeans.core.multitabs)

Tab-control is used by default and has hover colors in all Lafs (I think).
Multi-tabs is used e.g. if tab placement is left/right/bottom (see #1865 for more options to enable multi-tabs). Multi-tabs did not support hover in the past. This PR adds hover to multi-tabs.

> If the new property is needed, should the other LAFs be updated to use it as well?

Yes, this a good idea. Can be done in another PR.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabTableUI.java
DevCharly: This does not work. The first repaint is necessary to repaint the tab that was hovered before (remove hover background). The second repaint repaints the currently hovered tab. Maybe I should add some comments...

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
DevCharly: > Why are tab decorators not invoked in the isHover case?

The reason was that I wanted to avoid that the "project color bar" (from PR #2048) is painted for hovered tabs. Probably a bad idea. I'll change it back...

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabTableUI.java
eirikbakke: I meant like this:

```
int oldRow = hoverRow;
int oldColumn = hoverColumn;

if (oldRow != -1 && oldColumn != -1) {
    table.repaint(table.getCellRect(oldRow, oldColumn, true));
}
if (row != -1 && column != -1) {
    table.repaint(table.getCellRect(row, column, true));
}
hoverRow = row;
hoverColumn = column;
```
Since regardless of whether you are updating the new area or the old area, you want paintComponent to run according to the _new_ state.

eirikbakke: (Note: I edited this comment--see the latest version.)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
eirikbakke: Ah, I wasn't aware that different tab positions used a different implementation. Indeed on Windows these do not seem to support hover.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabTableUI.java
DevCharly: Ok, now I understand 👍 
You're right.
I'll change it.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
eirikbakke: Wait, but the screenshots show the editor tabs in the "top" position? Isn't that a tabcontrol type component, rather than a multitabs?

DevCharly: Multi-tabs are also used for top position if you enable one of the 4 checkboxes at the bottom of the screenshot below:

image

The screenshot highlights all options that enable multi-tabs.

I think I had enabled "Sort opend docs by project" when capturing the screenshots. Then only the popup menu is changed, but the tabs do not show additional information and look identical to Tab-control in FlatLaf. In other Lafs the Multi-tabs always look slightly different to Tab-control because Tab-control has UI delegates for different LaFs.

eirikbakke: OK, that explains it. I was curious, since I was thinking of doing some tweaks to the Windows LAF tabcontrol myself at some point.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabTableUI.java
DevCharly: Done

PR #2052, opened 2020-03-29
Improving LSP client: ability to share language servers across differ…

jlahoda:
…ent mime-types; improving span of hyperlinks based on TextMate grammars, if available.


jlahoda:
@ebarboni, it would be awesome if this could be considered for 12.0.

Thanks, Jan

PR #2053, opened 2020-03-30
Validations for maven project creation

errael:
Two fixes. Exception trying to create a maven project. Maven project created with invalid package name.


errael:
Funny, the pull request didn't use the commit summary and hook up to the issue, it used the bookmark instead.
```
changeset:   4122:a21deae90ed9
bookmark:    ValidationsForMavenProjectCreation
tag:         tip
parent:      4102:a92abb562f42
user:        Ernie Rael <errael@raelity.com>
date:        Mon Mar 30 06:47:05 2020 +0100
summary:     [NETBEANS-4083] Add missing validation in new maven project dialog
```

errael:
@ebarboni I might know why this PR didn't hook up properly. It's parent is
```
o    changeset:   4102:a92abb562f42
|\   bookmark:    release120
| ~  tag:         12.0-beta2
|    tag:         upstream/release120
```
but probably its parent should be
```
o    changeset:   4101:4881ea11ed7b
|\   tag:         upstream/master
```
I wonder if that could be the problem?

I see some stuff about manually linking to the issue
https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue
but those steps require write access to repository.

ebarboni:
I'm not an expert maybe you created the branch being on release120?

errael:
> Having a unit test verifying the new behavior wouldn't be bad...

My old school self never remembers that...

It looks like that in this case, that would be a major undertaking. I didn't see a single unit test in the whole of the netbeans/java subtree that uses/checks stuff based on the validation.api (at least not directly). And the class `BasicPanelVisual.java` has no existing tests. With significant guidance and time, a test for this simple enhancement to verification of user input to the dialog won't happen.

PR #2056, opened 2020-03-31
Introducing new Plugin Portal 3.0 update center in Apache NetBeans 12.0

jkovalsky:
In order to motivate NetBeans plugin developers to start using the new **Plugin Portal 3.0** [1] we need to register its update center in the **Apache NetBeans IDE 12.0**. This pull request contains necessary changes in this direction. Once this is merged we will also have to modify the URL redirect rules [2] accordingly. This is a fix for #NETBEANSINFRA-185.

[1] [https://netbeans-vm.apache.org/pluginportal/](https://netbeans-vm.apache.org/pluginportal/)
[2] [https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess](https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess)
[3] [https://issues.apache.org/jira/browse/NETBEANSINFRA-185](https://issues.apache.org/jira/browse/NETBEANSINFRA-185)

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -25,12 +25,15 @@ OpenIDE-Module-Short-Description=Declares NetBeans autoupdate centers.
 
 Services/AutoupdateType/distribution-update-provider.instance=NetBeans Distribution
 Services/AutoupdateType/pluginportal-update-provider.instance=NetBeans Plugin Portal
+Services/AutoupdateType/legacy-pluginportal-update-provider.instance=Legacy NetBeans Plugin Portal
 Services/AutoupdateType/82pluginportal-update-provider.instance=NetBeans 8.2 Plugin Portal
 
 #NOI18N
 URL_Distribution=@@metabuild.DistributionURL@@
 #NOI18N
 URL_PluginPortal=@@metabuild.PluginPortalURL@@
+#NOI18N
+URL_LegacyPluginPortal=@@metabuild.LegacyPluginPortalURL@@
 #NOI18N
 URL_82PluginPortal=http://updates.netbeans.org/netbeans/updates/8.2/uc/final/distribution/catalog.xml.gz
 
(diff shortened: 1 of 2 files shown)


nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
```diff
@@ -25,13 +25,16 @@ OpenIDE-Module-Short-Description=Declares NetBeans autoupdate centers.
...
 Services/AutoupdateType/pluginportal-update-provider.instance=NetBeans Plugin Portal
+Services/AutoupdateType/legacy-pluginportal-update-provider.instance=Legacy NetBeans Plugin Portal
 Services/AutoupdateType/82pluginportal-update-provider.instance=NetBeans 8.2 Plugin Portal
 
 #NOI18N
 URL_Distribution=@@metabuild.DistributionURL@@
 #NOI18N
 URL_PluginPortal=@@metabuild.PluginPortalURL@@
 #NOI18N
+URL_LegacyPluginPortal=@@metabuild.LegacyPluginPortalURL@@
```
mcdonnell-john: Does this need to be set somewhere?

Just looking at the other metadata its also used:
https://github.com/apache/netbeans/blob/312f983bf6627e91bdfb7526e67302975d3f7f80/nb/updatecenters/build.xml#L31
and
https://github.com/apache/netbeans/blob/6c3e8d1eee5c8ca49c11027065ca90c8abcbd920/nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java#L177

neilcsmith-net:
-1 to this from me, pending some conversations elsewhere, as I think we need to switch over to PP3 entirely for 12.0 release.

errael:
If I understand correctly, PP3 becomes `NetBeans Plugin Portal` and PP2 becomes `Legacy NetBeans Plugin Portal` after this change and `modify the URL redirect rules`. 

If PP2 has to be there (I'm inclined to agree with Neil, take it out), at least start off disabled/un-checked. Hmm, you could get rid of it in the release, then add a plugin to PP3 that when installed adds legacy PP2 UC.

neilcsmith-net:
@jkovalsky given conversations elsewhere, should this now be closed to get it off the 12.0 list?

jkovalsky:
Yes, we can close this one but need to create a different PR for removal of 8.2 update center instead. I will do that.

PR #2058, opened 2020-04-01
[NETBEANS-3870] FlatLaf and Windows LaF: fix smooth scrolling

DevCharly:
This fixes smooth scrolling in **editors**,  **Output** view and **Debugging** view for FlatLaf and Windows LaF.

Smooth scrolling requires a "precise" trackpad our mouse wheel (e.g. Macs or Surface Books) and allows scrolling by fraction of lines.

Smooth scrolling worked before fine in many views where FlatLaf or Windows LaF does the scrolling. But there are some places in NB where scrolling is done in NB, but precise rotation values are not used.

The problem is that `MouseWheelEvent.getWheelRotation()` or `MouseWheelEvent.getUnitsToScroll()` return zero when a precise trackpad/wheel is moved slowly. So the events are ignored.

Solution is to use `MouseWheelEvent.getPreciseWheelRotation()` instead.

On not precise wheels, `getPreciseWheelRotation()` returns the same value as `getWheelRotation()`.

### DIFF
--- a/ide/editor.lib2/src/org/netbeans/modules/editor/lib2/view/DocumentViewOp.java
+++ b/ide/editor.lib2/src/org/netbeans/modules/editor/lib2/view/DocumentViewOp.java
@@ -1510,7 +1510,7 @@ public final class DocumentViewOp
        
         JTextComponent textComponent = docView.getTextComponent();
         Keymap keymap = textComponent.getKeymap();
-        int wheelRotation = evt.getWheelRotation();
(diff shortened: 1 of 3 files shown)


DevCharly:
Tested with precise trackpad under FlatLaf, Windows Laf, Mac OS Laf, Metal and Nimbus.
Tested with not precise mouse wheel under FlatLaf, Windows Laf, Metal and Nimbus.

PR #2062, opened 2020-04-05
[NETBEANS-4078] Don't install nb-javac on JDK11

jtulach:
When creating new project with IDE running on JDK11, don't attempt to install nb-javac. I did following testing:
```bash
netbeans$ ant build
netbeans$ rm -rf nbuild/testuserdir nbbuild/netbeans/extra
netbeans$ ant tryme -Dnbjdk.home=/jdk8
netbeans$ rm -rf nbuild/testuserdir nbbuild/netbeans/extra
netbeans$ ant tryme -Dnbjdk.home=/jdk11
```
Only download of nb-javac is offered on OracleJDK8. Only download of OpenJFX is offered on JDK-11.

### DIFF
--- a/ergonomics/ide.ergonomics/java.properties
+++ b/ergonomics/ide.ergonomics/java.properties
@@ -20,7 +20,7 @@ project.file.../java.base/share/classes/java/lang/Object.java=org.netbeans.modul
 projectImporter=Eclipse
 
 mainModule=org.netbeans.modules.java.kit
-extra.modules=org.netbeans.modules.nbjavac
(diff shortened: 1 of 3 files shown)


geertjanw:
Only on JDK 11? Or on all JDKs except JDK 8?

neilcsmith-net:
Shouldn't this be matched by somewhat similar logic around https://github.com/apache/netbeans/blob/master/java/java.source/src/org/netbeans/modules/java/source/JBrowseModule.java#L94 ?

PR #2063, opened 2020-04-05
Lookup, add optional support

hectorespert:
Lookup, add optional support

### DIFF
--- a/platform/openide.util.lookup/src/org/openide/util/Lookup.java
+++ b/platform/openide.util.lookup/src/org/openide/util/Lookup.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -228,6 +229,18 @@ public abstract class Lookup {
      *         implementation is found
      */
     public abstract <T> T lookup(Class<T> clazz);
+    
+    /**
+     * Look up an object matching a given interface, optional version.
+     * Method to support Java Optional.
+     * If more than one object matches, the first will be returned.
+     * @param clazz class of the object we are searching for
+     * @return an optional that contains an object implementing the given class 
+     * or an empty optional if no such implementation is found
+     */
+    public <T> Optional<T> lookupOptional(Class<T> clazz) {
+        return Optional.ofNullable(lookup(clazz));
+    }
 
     /** The general lookup method. Callers can get list of all instances and classes
      * that match the given <code>template</code>, request more info about
--- a/platform/openide.util.lookup/test/unit/src/org/openide/util/LookupTest.java
+++ b/platform/openide.util.lookup/test/unit/src/org/openide/util/LookupTest.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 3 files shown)


jtulach:
Hello Hector, thanks for your interest in extending `Lookup` API. It is a core NetBeans API and I am still fond of creating it. That's one aspect.

The other one is that I don't understand why Java has `Optional`!? I believe I am not the only one. Java community seems to be divided into `Optional` believers and haters. I guess I fall rather into the latter category. To make it short, I have nothing against the believers using:
```java
Optional.of(context.lookup(AClass.class)).orElse(...)
```
but adding new method into `Lookup` and writing:
```java
context.lookupOptional(AClass.class).orElse(...)
```
isn't my cup of tea. Adding a new method into subclassable class isn't for free. Moreover please notice that both variants of the code are actually almost of the same length - e.g. I don't see any benefits in adding new `lookupOptional` method into `Lookup`.

That said, I am not going to veto the change, if there are at least three supporters of `Optional` in the `Lookup`. E.g. I am going to continue with the review, assuming you'll find enough supporters to overcome my own dislike of `Optional`.

platform/openide.util.lookup/src/org/openide/util/Lookup.java
```diff
@@ -228,6 +229,18 @@ private static synchronized void resetDefaultLookup() {
...
      */
     public abstract <T> T lookup(Class<T> clazz);
+    
+    /**
+     * Look up an object matching a given interface, optional version.
+     * Method to support Java Optional.
+     * If more than one object matches, the first will be returned.
+     * @param clazz class of the object we are searching for
+     * @return an optional that contains an object implementing the given class 
+     * or an empty optional if no such implementation is found
```
jtulach, line 239: Need `@since 8.42`.

platform/openide.util.lookup/test/unit/src/org/openide/util/LookupTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void setUp() {
+        lookup = new LookupImpl();
+    }
+
+    @Test
+    public void testLookupOptional() {
+        Optional<Integer> optionalInteger = lookup.lookupOptional(Integer.class);
+        assertFalse(optionalInteger.isPresent());
+        
+        Optional<String> optionalString = lookup.lookupOptional(String.class);
```
jtulach, line 47: This test is fine, but there are many implementations of `Lookup` and we have to test them all. There is a base "TCK-like" test `AbstractLookupBaseHid` - if you add a test method in there, it is going to be executed against all the known implementations of lookup - `ProxyLookup`, `AbstractLookup`, etc.

neilcsmith-net:
Unlike @jtulach I'll stick my head above the parapet as an `Optional` believer! :smile: And if `Lookup` was being written now I'd be very much in favour of building around `Optional` and particularly `Stream` (in fact have done so in my own take on the idea, and like JDK ServiceLoader additions).  However, I'm not in favour of retrofitting APIs just for the sake of it.  A mailing list discussion on pros, cons, reasons, etc. might be more appropriate on this one?

chhh:
If you rename the method to `maybe`, then it at least becomes shorter than Jaroslav's suggestion of `Optional.of()`

neilcsmith-net:
As per comment on dev@ if you're looking to shorten the method names (which isn't a bad idea) then I'd propose -

```
Optional<T> find(Class<T> clazz);
Stream<T> findAll(Class<T> clazz);
```

Yes, I'd also like to see a Stream returning method if we have this, and I think `find()` or a similar word mirrors `lookup()` better than `maybe()` does.

hectorespert:
After the conversation about this change, I'm going to close it: https://lists.apache.org/thread.html/r524f48285b6f695cfda47f11a1fb358b02d85f3e41e5394e01049522%40%3Cdev.netbeans.apache.org%3E

PR #2064, opened 2020-04-05
[NETBEANS-4116] Add Java Platform ComboBox Support to the Java Platform UI API

lkishalmi:
There are similar implementations at least 3 places in NetBeans. This PR moves the one used in Gradle to be a reusable part of the Java Project UI API.
Theoretically it would be easy to transform the existing uses to this new API, though I'd have a review first and even after that it might not be an urgent task to do.


lkishalmi:
Just a side note to this change, I think that the JavaPlatform.getProperties.get("platform.ant.name") would be worth to have it's own API in JavaPlatform like a String JavaPlatform.getInternalID().
Maybe it was not planned to be used and exposed that much, however it is "extensively" used in the code as a commonly know "secret trick".

java/java.platform.ui/src/org/netbeans/api/java/platform/PlatformComboBoxModel.java
```diff
@@ -0,0 +1,137 @@
...
+ *     String id = platform.getProperties().get(PROP_PLATFORM_ID);
+ *     ...
+ *     PlatformComboBoxModel model = new PlatformComboBoxModel();
+ *     model.setSelectedItem(id);
+ * </pre>
+ *
+ * @since 1.51
+ * @author Laszlo Kishalmi
+ */
+public final class PlatformComboBoxModel extends AbstractListModel<JavaPlatform> implements ComboBoxModel<JavaPlatform>, PropertyChangeListener {
```
JaroslavTulach, line 45: I suggest to hide this class behind factory method.

groovy/gradle.java/src/org/netbeans/modules/gradle/java/customizer/CompileOptionsPanel.java
```diff
@@ -171,7 +146,7 @@ private void initComponents() {
         lbCompileOnSave = new javax.swing.JLabel();
         cbAugmentedBuild = new javax.swing.JCheckBox();
         lbPlatform = new javax.swing.JLabel();
-        cbPlatform = new javax.swing.JComboBox<>();
+        cbPlatform = PlatformsCustomizer.createPlatformComboBox(new PlatformComboBoxModel());
```
JaroslavTulach, line 149: This usage could be replaced with the no-argument function.

java/java.platform.ui/src/org/netbeans/api/java/platform/PlatformComboBoxModel.java
lkishalmi: The updated PR does that. Good idea. Thank you!

groovy/gradle.java/src/org/netbeans/modules/gradle/java/customizer/CompileOptionsPanel.java
lkishalmi: Done. Thanks!

lkishalmi:
@JaroslavTulach thank you for the advises and the approval! I'd consider to create a Node based model, though unfortunately, I'm in rush finishing the new Gradle project wizards for 12.0 (if the RM would allow to put that in). This thing is just a side work coming out of that.

BTW: Can I have you opinion on the "platform.ant.name" thing I mentioned earlier on this thread?

lkishalmi:
I'm trying to add some unit tests at least for the model.

neilcsmith-net:
@lkishalmi is this 12.0 or is this "do not merge" in your opinion? Inclined to agree with @JaroslavTulach that this should be Node-based.

lkishalmi:
Removed 12.0 for milestone. Work is more demanding than usual, I gave up chasing 12.0. This one would have been required for my Gradle Template improvements which, I still did not have time to finish. Might consider on the Node based solution later...

PR #2067, opened 2020-04-06
Upgrading FreeMarker to Apache FreeMarker

lkishalmi:
Upgrade of Freemarker 2.3.19 to Apache FreeMarker 2.3.30. It might have a few useful feature for template developers, and FreeMarker is a graduated Apache project now.

Useful change: it supports resolving values from Lambdas.


lkishalmi:
It might or may not be included in 12.0. Let's @ebarboni have that call.

PR #2068, opened 2020-04-08
[NETBEANS-4127] Adjust maven profiler to provide jvmargs as separate …

errael:
…options

This change is structured to add some new properties for use in pom.xml. It does not change any existing stuff, either internal to NB or available to the pom.

This change makes the profiling jvmargs available as individual properties. This is required for the javafx-maven-plugin.

The PR also uses updated javafx archetypes which include a profile nbactions and pom execution.


geertjanw:
See this related issue from NetCAT 12.0:

https://issues.apache.org/jira/browse/NETBEANS-4033

PR #2069, opened 2020-04-08
Use current platform for OpenJDK tests

jlahoda:
Without this, the OpenJDK tests do not see classes added in the OpenJDK sources, but rather the runtime platform classes (this is particularly problem for classes for java.base).

Erik, it would be awesome if this could be considered for 12.0.

### DIFF
--- a/java/java.openjdk.project/src/org/netbeans/modules/java/openjdk/project/ClassPathProviderImpl.java
+++ b/java/java.openjdk.project/src/org/netbeans/modules/java/openjdk/project/ClassPathProviderImpl.java
@@ -29,6 +29,7 @@ import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/MoveMembersTransformer.java
```diff
@@ -451,6 +460,9 @@ public Boolean reduce(Boolean r1, Boolean r2) {
                     return (r1 == Boolean.TRUE || r2 == Boolean.TRUE);
                 }
             };
+            if (workingCopy.getTrees().getTree(el) == null) {
+                System.err.println("!!!");
```
JaroslavTulach, line 464: Left over debug message, I guess.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/PushDownTransformer.java
```diff
@@ -62,11 +67,18 @@ public Problem getProblem() {
...
+    public PushDownTransformer(FileObject originFile, MemberInfo<ElementHandle<? extends Element>> members[]) {
+        this.originFile = originFile;
         this.members = members;
         this.translateQueue = new LinkedList<>();
     }
 
+    @Override
+    public void setWorkingCopy(WorkingCopy workingCopy) throws ToPhaseException {
+        SourceUtils.forceSource(workingCopy, originFile);
+        super.setWorkingCopy(workingCopy); //To change body of generated methods, choose Tools | Templates.
```
JaroslavTulach, line 79: Delete the comment.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/MoveMembersTransformer.java
jlahoda: Right. Sorry, I mistakenly based this branch on a different branch, fixed. There should only be one file changed, the ClassPathProvider.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/PushDownTransformer.java
jlahoda: Right. Sorry, I mistakenly based this branch on a different branch, fixed. There should only be one file changed, the ClassPathProvider.

arvindaprameya:
+1 Approved, Thanks

PR #894, opened 2018-09-21
[NETBEANS-1275] Register 8.2 update center disabled.

geertjanw:

PR #2072, opened 2020-04-10
Removing legacy Update Center for NetBeans 8.2 hosted on Plugin Portal 2.0

jkovalsky:
As discussed in PR #2056 [1] Apache NetBeans IDE 12.0 will not rely on Oracle hosted infrastructure so here we remove old update center for NetBeans 8.2. If someone still needs old plugins, s/he can still register copy of legacy update centers hosted on Emilian's box as described in [2] and install the plugins like CND.

[1] https://github.com/apache/netbeans/pull/2056
[2] https://cwiki.apache.org/confluence/display/NETBEANS/Where+to+download+plugins+for+NetBeans+10.0+and+earlier

### DIFF
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/Bundle.properties
@@ -25,14 +25,11 @@ OpenIDE-Module-Short-Description=Declares NetBeans autoupdate centers.
 
 Services/AutoupdateType/distribution-update-provider.instance=NetBeans Distribution
 Services/AutoupdateType/pluginportal-update-provider.instance=NetBeans Plugin Portal
-Services/AutoupdateType/82pluginportal-update-provider.instance=NetBeans 8.2 Plugin Portal
 
 #NOI18N
 URL_Distribution=@@metabuild.DistributionURL@@
 #NOI18N
 URL_PluginPortal=@@metabuild.PluginPortalURL@@
-#NOI18N
-URL_82PluginPortal=http://updates.netbeans.org/netbeans/updates/8.2/uc/final/distribution/catalog.xml.gz
 
 3rdparty=Third Party Libraries
 #NOI18N
--- a/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/mf-layer.xml
+++ b/nb/updatecenters/src/org/netbeans/modules/updatecenters/resources/mf-layer.xml
@@ -46,16 +46,6 @@
          <attr name="instanceCreate" methodvalue="org.netbeans.modules.autoupdate.updateprovider.AutoupdateCatalogFactory.createUpdateProvider"/>
       </file>
       
-      <file name="82pluginportal-update-provider.instance">
-          <attr name="displayName" bundlevalue="org.netbeans.modules.updatecenters.resources.Bundle#Services/AutoupdateType/82pluginportal-update-provider.instance"/>
-          <attr name="iconBase" stringvalue="org/netbeans/modules/updatecenters/resources/updateAction.gif"/>
-          <attr name="url" bundlevalue="org.netbeans.modules.updatecenters.resources.Bundle#URL_82PluginPortal"/>
-          <attr name="category" stringvalue="STANDARD"/>
-          <attr name="enabled" boolvalue="false"/>
-          <attr name="instanceOf" stringvalue="org.netbeans.spi.autoupdate.UpdateProvider"/>
-          <attr name="instanceCreate" methodvalue="org.netbeans.modules.autoupdate.updateprovider.AutoupdateCatalogFactory.createUpdateProvider"/>
-      </file>
-
       <file name="3rdparty.instance">
           <attr name="displayName" bundlevalue="org.netbeans.modules.updatecenters.resources.Bundle#3rdparty"/>
           <attr name="iconBase" stringvalue="org/netbeans/modules/updatecenters/resources/updateAction.gif"/>


vieiro:
Hi @jkovalsky 

New URL for NetBeans 12.0 is https://plugins.netbeans.apache.org/data/12.0/catalog.xml.gz

Would it be possible to update it everywhere? There's no need to add a "pluginportal" directory any more.

Thanks,
Antonio

neilcsmith-net:
@vieiro for the IDE will need changing in https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess

Need to liaise with Eric on that change and this - he's missing on review list. Can't add via mobile UI.

vieiro:
Hi @ebarboni , @neilcsmith-net .

I understand that .htaccess has to be changed, but I don't know **what** needs to change. I need to understand the current setup to that I can make the change. 

So, for instance, is
http://plugins.netbeans.org/nbpluginportal/updates/11.0/catalog.xml
Supposed to be changed to
https://plugins.netbeans.apache.org/data/11.0/catalog.xml ?

Maybe we can meet sometime and talk about this? Or explain the current setup so I can be of help?

vieiro:
Hi again @neilcsmith-net , @ebarboni , @jkovalsky if we're dropping plugins.netbeans.org then the redirection is not needed any more, or  is it?

jkovalsky:
Yes, I think so @vieiro. We can get rid of it unless it breaks redirection for older NetBeans 9.*, 10.* or 11.*.

neilcsmith-net:
-1 to removing the redirection for now. In early video meetings on web infra we discussed having all IDE inbound links under the same path (/nb/*).  It may be less important now as we've almost finished transitioning, but good for consistency, tracking, ease of changing redirects while releases active, etc.

@vieiro @jkovalsky can we have that discussion on dev@ for the 12.1 timeframe rather than changing at this point?

vieiro:
@neilcsmith-net @ebarboni @jkovalsky  I don't know what "early video meetings on web infra ", please send me a link via email so I can join.

neilcsmith-net:
@vieiro you were involved - talking 2 years ago now?!  Just saying that having the redirects through the website .htaccess is still potentially useful, and changing just before release doesn't seem the right approach.  Better to discuss / change in next release cycle?

vieiro:
@neilcsmith-net , @ebarboni , @jkovalsky Regarding the redirections, I think that should be done on a separate PR on the netbeans-website repo.

JaroslavTulach:
If this plugin portal is disabled, then it is going to be hard to for users of Apache NetBeans to install C/C++ support. Until we can provide C/C++ support as part of Apache NetBeans, I suggested to delay the removal of this (disabled) plugin central.

neilcsmith-net:
@JaroslavTulach is documenting how to add the UC for those who want C/C++ support that much more difficult than documenting to enable it?  Surely it really depends on how long Oracle plan on keeping it up?

JaroslavTulach:
> @JaroslavTulach is documenting how to add the UC for those who want C/C++ support 
> that much more difficult than documenting to enable it? 

I remember there was a discussion about setting up the 8.2 update center to get the CND support being too complicated. It resulted in @geertjanw's #894 - I am not sure what has changed since then? Apache NetBeans still lacks CND and that is, in my opinion, a potential acceptance barrier. 

> Surely it really depends on how long Oracle plan on keeping it up?

I have no idea. Knowing corporate policies, I can only assume that until Oracle NetBeans 8.2 is officially EOLed, it has to stay around.

neilcsmith-net:
> I am not sure what has changed since then?

@JaroslavTulach this was initiated by @jkovalsky here and on mailing list partly (as I understand) because there isn't a commitment from Oracle to keep the relevant infrastructure for old plugin centre and 8.2 update centre running for full life of NB 12.0 - ie. until at least May 2021.  And potentially (given other changes) put it behind https.

ebarboni:
do we merge this ?

jkovalsky:
In my opinion we don't want to merge this if @emilianbold helps us transfer modules of NetBeans 8.2 standard distribution from Oracle infrastructure to http://netbeans.osuosl.org/ as discussed here: https://netbeans.slack.com/archives/C2BJDQ77A/p1588447076088100

emilianbold:
I think in Terminator-like movies they use all these terminal-based computers made from scrap hardware because all the more recent software was lost.

It's important to preserve this stuff, but not necessarily provide it as a crutch forever.

Technically, copying some files to OSUOSL should be easy. Still, considering the Apache NetBeans PMC has full control over the OSUOSL account (although, I can't remember if OSUOSL knows this, formally), I wonder how legally that's not an act of the foundation right there and how Apache is not distributing this stuff...

jlahoda:
FWIW, there may be a chance to provide archives of the legacy releases on ASF infra:
https://mail-archives.apache.org/mod_mbox/www-legal-discuss/202001.mbox/%3CCABD8fLU3OcAfKvX5dP1Yx_0C%3DhzVeQkwd4Lidg7VMRYAAARMEA%40mail.gmail.com%3E

Maybe we should try to pursue this direction?

neilcsmith-net:
> nothing changed since then

@JaroslavTulach excepting Oracle's commitment to maintain the necessary infrastructure to continue serving these for the lifetime of 12.0 until end-May 2021?! Either we have that, or a valid alternative option in place / ready to go, or we need to remove IMO.

geertjanw:
Can’t we change the physical location of where these are via htaccess?

neilcsmith-net:
We can certainly redirect now, but might be better to change the link and put through the same redirections / .htaccess as the other UCs?

I'd certainly like to see it stay, but not without a guaranteed plan for a stable and secure location to move to if Oracle hosting of it is going.  Ideally, if not via Oracle, something under PMC control.  If the link is in the IDE, it should be something we can guarantee is available, stable and secure throughout the lifetime of the release.  If not, and it does stay, at least some bloody big caveat in the description, and perhaps not on an "official" (ie. netbeans.org) domain?

Asking @matthiasblaesing for thoughts, given work done on securing other aspects of UCs recently.

geertjanw:
Nobody is saying Oracle hosting of it is going. But Oracle hosting of it is not guaranteed nor will last forever. It’s a question of someone at some point pulling the plug, though in the greater scheme of Oracle things, what we’re doing here is very much under the radar, so as long as things keep going like they are, nothing will change.

neilcsmith-net:
@geertjanw that was not my understanding from conversations on dev@ and elsewhere to why we have this PR in the first place.  Right now, I'm +1 to whichever approach @jkovalsky thinks best to take.  Nothing further to add, and we're just going around in circles! :smile:

jkovalsky:
OK, since we didn't reach consensus, I am closing this PR. We can reconsider this in the future when physical location changes.

PR #2074, opened 2020-04-12
[NETBEANS-4159] Switch to ElementKindVisitor8 as base class (min runtime JDK) and add noop logger for unknown elements

graben:


graben:
Make unit tests work again.

graben:
@jlahoda : Could you plz have a look. Might be good to get into 12.0 to avoid errors with JDK14 features

PR #2076, opened 2020-04-13
[NETBEANS-346] Tabs, characters, and the right Margin don't line up

eirikbakke:
Fix a problem that caused the end-of-line marker (e.g. at 80 chars) to be displayed at an incorrect position (e.g. at 82 chars) on certain zoom levels, on Java 9 or above. This problem stems from rounding error caused by assuming that the width of each (monotype font) character is an integral number of logical pixels wide. The latter is not the case on HiDPI screens (e.g. where 10 device pixels may map to 6.67 logical pixels at 150% scaling), nor on Windows with subpixel antialiasing, and probably some other situations as well.

This also fixes less prominent problems with visual alignment of tabs vs. spaces (e.g. 8-space tabs will get the wrong width due to the same rounding error).

While looking at the related code, I discovered various other places where the same kind of rounding error is likely to cause problem; I created NETBEANS-4166 for these.


eirikbakke:
Bug is confirmed to occur on both MacOS and Windows (with and without HiDPI/Retina scaling), and fixed on both.

PR #2077, opened 2020-04-14
Update used groovy libraries to 2.5.11. Fixes NETBEANS-4172

svenreimers:


svenreimers:
Maybe too late for 12.0 but I think this may be a low risk, nice simple quick win.

PR #2079, opened 2020-04-15
[NETBEANS-4173] FlatLaf: Partial render selected BeanTreeView element

dinix2008:
When using FlatLaf Look and Feel, incorrect rendering of tree item under the context menu in the BeanTreeView component occurs. This problem can be easily reproduced by the following simple example.

```
import com.formdev.flatlaf.FlatLightLaf;
import org.openide.explorer.ExplorerManager;
import org.openide.explorer.view.BeanTreeView;
import org.openide.nodes.AbstractNode;
import org.openide.nodes.Children;

import javax.swing.*;
import java.awt.*;
import java.awt.event.ActionEvent;

public class SampleFrame extends JFrame implements ExplorerManager.Provider {
    private final ExplorerManager explorerManager = new ExplorerManager();

    public SampleFrame() throws HeadlessException {
        setSize(300, 400);
        explorerManager.setRootContext(new SampleNode());
        BeanTreeView beanTreeView = new BeanTreeView();
        getContentPane().add(beanTreeView, BorderLayout.CENTER);
    }

    @Override
    public ExplorerManager getExplorerManager() {
        return explorerManager;
    }


    private static class SampleNode extends AbstractNode {
        public SampleNode() {
            super(Children.LEAF);
            setName("SampleNode");
        }

        @Override
        public Action[] getActions(boolean context) {
            AbstractAction action = new AbstractAction() {
                @Override
                public void actionPerformed(ActionEvent e) {

                }
            };
            action.putValue(Action.NAME, "Sample");
            return new Action[]{action};
        }
    }

        SwingUtilities.invokeLater(() -> {
            FlatLightLaf.install();

            SampleFrame frame = new SampleFrame();
            frame.setDefaultCloseOperation(EXIT_ON_CLOSE);
            frame.setVisible(true);
        });
    }
}
```

The result will be the following
FlatLaf_before_fix

and in Netbeans Services Window

Netbeans Services Windows before fix_FlatLaf

After my fix, the render of the tree item becomes correct

FlatLaf

and in Netbeans Services Window

Netbeans Services Windows after fix_FlatLaf

Additionally, I checked the fix in Metal and Aqua LAF

Metal:
Netbeans Services Windows after fix_Metal

Aqua:
Netbeans Services Windows after fix_Aqua

### DIFF
--- a/platform/openide.awt/src/org/openide/awt/HtmlRendererImpl.java
+++ b/platform/openide.awt/src/org/openide/awt/HtmlRendererImpl.java
@@ -182,7 +182,7 @@ class HtmlRendererImpl extends JLabel implements HtmlRenderer.Renderer {
 
         setSelected(selected);
 
-        if (selected) {
+        if (selected && leadSelection) {
             setParentFocused(checkFocused(target));
         } else {
             setParentFocused(false);


lkishalmi:
BTW what shall we see on those pictures above?

dinix2008:
Sorry, I corrected the description. If this is not enough, tell me what else to add

DevCharly:
I think this more a bug in FlatLaf than in NB. 

The question is whether the selected tree item should be blue (which means focused) if a popup menu is shown or become gray (which means not focused).

In Windows Explorer and macOS Finder the selection stays blue when a popup menu is shown.
This feels more natural than changing selection from blue to gray on right-click.

I'll check whether I can change this in FlatLaf.

dinix2008:
> I think this more a bug in FlatLaf than in NB.
> 
> The question is whether the selected tree item should be blue (which means focused) if a popup menu is shown or become gray (which means not focused).
> 
> In Windows Explorer and macOS Finder the selection stays blue when a popup menu is shown.
> This feels more natural than changing selection from blue to gray on right-click.
> 
> I'll check whether I can change this in FlatLaf.

I thought this was a mistake, because in IntelliJ IDEA in the IntelliJ Light theme that FlatLaf looks like, the selection is inactive. For example

IntelliJ IDEA

This is also the case with Aqua LAF in JTree with DefaultTreeCellRenderer.

DevCharly:
> I thought this was a mistake, because in IntelliJ IDEA in the IntelliJ Light theme that FlatLaf looks like, the selection is inactive.

Yes, I see. But IDEA is not perfect 😉 

Done more testing on Linux (GNOME and KDE) and macOS.
The behavior is always that the focused border and selection stays where it is when opening a popup menu or the main menu. Even a Firefox text area (were I currently write this text) keeps the blue focused border when right-clicking on it.

So I changed the behavior in FlatLaf 0.31.
See PR #2094

Thanks @dinix2008 for pointing out this problem.

dinix2008:
> > I thought this was a mistake, because in IntelliJ IDEA in the IntelliJ Light theme that FlatLaf looks like, the selection is inactive.
> 
> Yes, I see. But IDEA is not perfect 😉
> 
> Done more testing on Linux (GNOME and KDE) and macOS.
> The behavior is always that the focused border and selection stays where it is when opening a popup menu or the main menu. Even a Firefox text area (were I currently write this text) keeps the blue focused border when right-clicking on it.
> 
> So I changed the behavior in FlatLaf 0.31.
> See PR #2094
> 
> Thanks @dinix2008 for pointing out this problem.

Thanks!

dinix2008:
Not needed, task solved in FlatLaf library

PR #2081, opened 2020-04-15
[NETBEANS-3707] Fixed colors used for PHP code completion

KacerCZ:
Colors used for code completion are taken from Java.
New colors are readable when dark theme is used.

Also fixes issue NETBEANS-3706

### DIFF
--- a/ide/csl.api/src/org/netbeans/modules/csl/editor/completion/GsfCompletionItem.java
+++ b/ide/csl.api/src/org/netbeans/modules/csl/editor/completion/GsfCompletionItem.java
@@ -648,16 +648,16 @@ public abstract class GsfCompletionItem implements CompletionItem {
     /** Format parameters in orange etc. */
     private static class CompletionFormatter extends GsfHtmlFormatter {
         private static final String METHOD_COLOR = LFCustoms.getTextFgColorHTML();
-        private static final String PARAMETER_NAME_COLOR = getHTMLColor(160, 96, 1);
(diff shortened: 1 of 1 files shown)


junichi11:
Could you attach screenshots for default and dark theme (before and after)?

junichi11:
Maybe, I think this is also related to other languages. So, we should verify them.

KacerCZ:
Before - NetBeans profile, Windows look and feel
before-light-keyword-crop before-light-property-crop before-light-class-crop

Before - FlatLaF Dark profile, FlatLaF Dark
before-dark-keyword-crop before-dark-property-crop before-dark-class-crop


After - NetBeans profile, Windows look and feel
after-light-keyword-crop after-light-property-crop after-light-class-crop

After - FlatLaF Dark profile, FlatLaF Dark look and feel 
after-dark-keyword-crop after-dark-property-crop after-dark-class-crop

@eirikbakke If PR is approved it would be great to include the fix in version 12.0.

eirikbakke:
The colors look fine. Reading the comments on https://issues.apache.org/jira/browse/NETBEANS-3707 it appears this does not impact Java editing, which makes this less risky of a change.

I note that keywords like "private" have been changed from green to blue. But they are blue for Java code, so that seems like a change that makes things more consistent across languages.

KacerCZ:
@ebarboni Please review PR and if it's not too late include this fix in 12.0.

PR #2082, opened 2020-04-15
[NETBEANS-4177] Add another check for valid parameter list position to avoid exception if sourcecode is incorrect

graben:


graben:
@eirikbakke : Should fix your mentioned issue

eirikbakke:
Reading through the method, this seems straightforward enough. Thank you!

graben:
@jlahoda : Simple test added.

graben:
@jlahoda : Should be considered for 12.0, too?

jlahoda:
Yes, I believe this should be considered for 12.0. Eirik - what do you think?

eirikbakke:
Looks completely safe to include to me.

eirikbakke:
If you still have this patch running on your build, could you see if this also handles the following case?

```
public class FooTest {
  public void someMethod() {
    List<String> someFoo = listOf(transform(null, s -> new Integer(5)));
  }

  public static <E> List<E> listOf(E element) {
    throw new UnsupportedOperationException();
  }

  public static <F, T> Iterable<T> transform(
      Iterable<F> fromIterable, Function<? super F, ? extends T> function)
  {
    throw new UnsupportedOperationException();
  }
}
```

graben:
Hi @eirikbakke, works for me on master build [1578](https://builds.apache.org/view/M-R/view/NetBeans/job/netbeans-linux/1578/).

eirikbakke:
@graben Excellent--thank you!

PR #2083, opened 2020-04-16
Preventing AssertionError from javac due to use of incorrect scope on…

jlahoda:
… JDK12+

There is a random AssertionError, typically from code completion, due to the use of incorrect scope. Sadly, on JDK 11 without nb-javac, we need to use the wrong scope, as otherwise code comletion/code templates would not work in common cases. But we don't need to use the wrong scope on JDK12+ or when running with nb-javac.


arvindaprameya:
+1 Approved, Thank you Jan

PR #2084, opened 2020-04-16
[NETBEANS-4182] Save All Files when switching to Terminal

lkishalmi:
I added another option class finally, instead of extending the TermOptions. As this option does not fit in that context. It might look a bit handy capped, but I think that was better instead of adding an option to a module where that could not be handled.
@ebarboni I think the change is safe, however might be late in the game for 12.0


lkishalmi:
Well let's rethink the feature!

PR #2086, opened 2020-04-17
Fix apisupport.ant tests

hectorespert:
Fix tests in apisupport.ant module


hectorespert:
@JaroslavTulach Could you review this?

PR #2087, opened 2020-04-18
[NETBEANS-4071] Fix plexus class setup for maven embedder

matthiasblaesing:
Updated maven embedder and indexer changed jar structure and in turn
the setup of the classrealm needs to be updated to match.


matthiasblaesing:
The fix was tested locally and by an affected user. The change is minimal and only affects the maven indexing.

ebarboni:
/bow thanks for this fix. Had no idea how to fix that 🥇

PR #2039, opened 2020-03-20
[NETBEANS-4039] Added SPI for Preferred Color Profile Changes

lkishalmi:
A non-hack implementation of #2031

PR #2088, opened 2020-04-18
Revert "[NETBEANS-4039] Added SPI for Preferred Color Profile Changes"

lkishalmi:
Reverts apache/netbeans#2039, as the introduced new dependency made some unit-test load for ever waiting on the UI.


platform/core.windows/manifest.mf
```diff
@@ -6,5 +6,5 @@ OpenIDE-Module-Provides: org.netbeans.core.WindowSystem, org.openide.windows.Win
 OpenIDE-Module-Recommends: org.netbeans.core.windows.nativeaccess.NativeWindowSystem
 AutoUpdate-Show-In-Client: false
 AutoUpdate-Essential-Module: true
-OpenIDE-Module-Specification-Version: 2.92
+OpenIDE-Module-Specification-Version: 2.91
```
jlahoda, line 9: I guess I would suggest to actually increase the spec version (so that the spec versions are monotonically increasing). I.e.:
```suggestion
OpenIDE-Module-Specification-Version: 2.93
```

ebarboni:
done

PR #2089, opened 2020-04-19
[NETBEANS-3311] Convert TextBlock to String, Function call parameter fix

Akshay-Gupta-Oracle:
If a TextBlock was present in a function call, previous hint does not converts it to String. This PR is to fix (and test)that bug.


arvindaprameya:
+1 approved

PR #2090, opened 2020-04-19
[NETBEANS-4037] Don't invalidate global classpath on JAR content change.

JaroslavTulach:
One of the identified reasons for [debugger slowness](https://issues.apache.org/jira/browse/NETBEANS-4037) was the fact that GlobalClassPath registry is invalidated everytime a JAR file is changed. Hence, when you make a change and want to debug, you compile, change the JAR and then the debugger has to wait. Needlessly. I'd like to fix that by not invalidating the global classpath on JAR content change, when the JAR file still exists.

### DIFF
--- a/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
+++ b/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
@@ -1415,7 +1415,9 @@ public final class ClassPath {
 
         @Override
         public void fileChanged(FileEvent fe) {
-            processEvent(fe);
+            if (!fe.getFile().isValid()) {
+                processEvent(fe);
+            }
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
```diff
@@ -1415,7 +1415,9 @@ public void fileDataCreated(FileEvent fe) {
 
         @Override
         public void fileChanged(FileEvent fe) {
-            processEvent(fe);
+            if (!fe.getFile().isValid()) {
```
tzezula, line 1418: This is in fact wrong and incompatible change.
The `ClassPath.PROP_ROOTS` must be fired on content change.

JaroslavTulach:
Thanks for the review @tzezula. In such case we have to do just with #2092 fix.

PR #2092, opened 2020-04-20
[NETBEANS-4037] Attempting to avoid querying the SourceForBinaryQuery…

jlahoda:
… for the same root more than once.


arvindaprameya:
+1 Approved , Thanks for the fix

PR #2095, opened 2020-04-21
If cancel happens while Analyzer is running, AssertionError is thrown…

jlahoda:
…. Rather ignore all exception happening while the parser is cancelled.

While running without nb-javac, the AssertionErrors from Analyzer due to cancellation are about 1/2 of javac-related exceptions I see. So feels like something worthy to solve. Theoretically, this can lead to ignored exceptions - but I don't think we really care about anything if the parser is cancelled.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
@@ -709,9 +709,14 @@ public class JavacParser extends Parser {
         } catch (Abort abort) {
             parserError = currentPhase;
         } catch (RuntimeException | Error ex) {
-            parserError = currentPhase;
-            dumpSource(currentInfo, ex);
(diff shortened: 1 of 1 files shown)


arvindaprameya:
+ 1 Approved, Thanks Jan

java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -709,9 +709,14 @@ Phase moveToPhase (final Phase phase, final CompilationInfoImpl currentInfo, Lis
         } catch (Abort abort) {
             parserError = currentPhase;
         } catch (RuntimeException | Error ex) {
```
eirikbakke, line 711: Perhaps this should be catching AssertionError more specifically, rather than Error in general, giving other kinds of Error the old behavior. Though I'm fine with either approach.

PR #2097, opened 2020-04-22
[NETBEANS-4206] Only set project location once directly from PROJECT_…

errael:
…PARENT_FOLDER

Problem is the settings property key "projdir", aka PROJECT_PARENT_FOLDER, used by maven's new project wizard's "Name and Location" panel. "projdir" has two meaning.

- Maven's CreateNewModule action uses "projdir" property to inform the wizard where to put the project
example: /tmp/parent
- The wizard uses "projdir" property to inform project creation of the project directory
example: /tmp/parent/child

This difference is a problem if back/next are used in the dialog; in this case the output `/tmp/parent/child` becomes the input. Repeated back/next cause the directory to grow each time. See #1980 for details of the first fix and discussion of problem with CreateNewModule action. The first fix took care of the back/next problem, but it did not take CreateNewModule usage into account.

This fix detects the first time the properties are read, and in this case uses "projdir" to indicate the parent directory; subsequent usage (back/next) "projdir" is used to indicate the child directory.

This is a quick fix for 12.0. NOTE that most old code uses literal "projdir", rather than PROJECT_PARENT_FOLDER. I suggest a new property be introduced, for example
```
String PROJECT_BASE_FOLDER = "projdir";
String PROJECT_PARENT_FOLDER = "parent-projdir";
```
@neilcsmith-net , @matthiasblaesing this needs some support to get it into 12.0

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/newproject/BasicPanelVisual.java
+++ b/java/maven/src/org/netbeans/modules/maven/newproject/BasicPanelVisual.java
@@ -567,6 +567,7 @@ public class BasicPanelVisual extends JPanel implements DocumentListener, Window
         });
     }
     
+    private static final String SETTINGS_HAVE_READ = "BasicPanelVisual-read-properties-before"; //NOI18N
     @Messages({
         "# {0} - project count", "TXT_MavenProjectName=mavenproject{0}",
         "TXT_Checking1=Checking additional creation properties..."
@@ -578,13 +579,27 @@ public class BasicPanelVisual extends JPanel implements DocumentListener, Window
                 handle = null;
             }
         }        
-        // PROJECT_PARENT_FOLDER confusing, better name is PROJECT_BASE_FOLDER
+        // PROJECT_PARENT_FOLDER usage is confusing. Sometimes it's the
+        // parent directory, sometimes it's the project directory.
+        // Maybe introduce PROJECT_BASE_FOLDER, to clarify and differentiate.
+        // But for local fix [NETBEANS-4206] keep track of whether
+        // these properties have been read before.
+        boolean haveRead =  Boolean.parseBoolean((String)settings.getProperty(SETTINGS_HAVE_READ)); //NOI18N
         File projectFolder = (File) settings.getProperty(CommonProjectActions.PROJECT_PARENT_FOLDER); //NOI18N
         File projectLocation;
-        if (projectFolder == null || projectFolder.getParentFile() == null || !projectFolder.getParentFile().isDirectory()) {
-            projectLocation = ProjectChooser.getProjectsFolder();
+        if(!haveRead && projectFolder != null) {
+            // First time in here, dialog was started with project folder;
+            // example is creating a project from pom parent
+            projectLocation = projectFolder;
         } else {
-            projectLocation = projectFolder.getParentFile();
+            if (projectFolder == null || projectFolder.getParentFile() == null || !projectFolder.getParentFile().isDirectory()) {
+                projectLocation = ProjectChooser.getProjectsFolder();
+            } else {
+                projectLocation = projectFolder.getParentFile();
+            }
+        }
+        if(!haveRead) {
+            settings.putProperty(SETTINGS_HAVE_READ, "true"); //NOI18N
         }
         this.projectLocationTextField.setText(projectLocation.getAbsolutePath());
         


errael:
> Can you document where and how the "projdir" attribute in the wizard is used elsewhere that requires it to be the project folder and not the parent?

I can probably find it again. Here's an easy way to see it. In `store()` set "projdir"/PROJECT_PARENT_FOLDER property to the same value as picked up in `read()`, "/tmp/parent" in the example. Then after "Finish" and project creation, there is a "/tmp/parent/src" and the parent project is marked with "unloadable" (or somesuch)  (that is the first thing I tried, as @neilcsmith-net  suggested and I agreed)

Note that before the fix for the next/prev bug, "projdir" was set to "/tmp/parent/child" in `store()`, reverting the fix will not change that.  before any changes. This is from 
```
changeset:   1692:b319f80ce20b
summary:     Moving java modules into java subdirectory
```
Here's the code in `BasicPanelVisual.store()`
```
        String name = projectNameTextField.getText().trim();
        String folder = createdFolderTextField.getText().trim();
        final File parentFolder = new File(folder);

        d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, parentFolder);
```
Even though the variable says `parentFolder` it comes from `createdFolderTextField`. I changed the names around to satisfy a request from reviewer in https://github.com/apache/netbeans/pull/1980#discussion_r387210839

neilcsmith-net:
> I can probably find it again. Here's an easy way to see it. In store() set "projdir"/PROJECT_PARENT_FOLDER property to the same value as picked up in read(), "/tmp/parent" in the example. Then after "Finish" and project creation, there is a "/tmp/parent/src" and the parent project is marked with "unloadable"

Yes, not entirely surprising - something is relying on the incorrect (as documented) value.  That is where the fix should be IMO.  Also, are the `PROJECT_PARENT_FOLDER` and the target folder meant to be the same?  I _think_ both are meant to be the parent folder of the created project?

Maybe can be fixed up as part of the wizard instantiate?

Personally, given that the prev/next bug fix introduced something worse, I'd still rather revert and review post 12.0 than risk adding to it and leaving a problem in there. But see what others think.

errael:
Two extreme possibilities are that "as documented" is correct and all
the code is wrong. Or vice-versa. We're probably somewhere in the middle,
but closer to "as documented" wrong I think.

When working on the original fix, I saw how the GUI and 
instantiation were copied/modeled from java projects. In
```
java/java.j2semodule/src/.../NewJ2SEModularProjectWizardIterator.java
```
note the explicit use of "projdir" in the following
```
public Set<FileObject> instantiate (ProgressHandle handle) throws IOException {
    ...
    File dirF = (File)myWiz.getProperty("projdir"); //NOI18N
    ...
    J2SEModularProjectGenerator.createProject(dirF, name, platform);
```
And in this code from
```
java/maven/.../newproject/idenative.IDENativeMavenWizardIterator
```

note the usage of "projdir" (PROJECT_PARENT_FOLDER)
```
public Set<FileObject> instantiate (ProgressHandle handle) throws IOException {
    ...
    File projFile = FileUtil.normalizeFile((File) wiz.getProperty(
                         CommonProjectActions.PROJECT_PARENT_FOLDER)); // NOI18N
    ...
    CreateProjectBuilder builder = createBuilder(projFile, vi, handle);
```

I put a breakpoint on CreateProjectBuilder and

- Did CreateNewModule on pom project /tmp/parent,
- projectName called child,
- dialog shows projectLocation as /tmp/parent
- and projectFolder as /tmp/parent/child

at the breakpoint: projFile is /tmp/parent/child

errael:
> Also, are the `PROJECT_PARENT_FOLDER` and the target folder meant to be the same? _I think_ both are meant to be the parent folder of the created project?

Do you still think this is the case?

> I'm not entirely sure it hasn't broken other things now I've started looking further.

Could you clarify what you've seen from "looking further"?

neilcsmith-net:
> Do you still think this is the case?

Yes, because target folder on the template takes a `DataFolder`, which means the thing you're trying to create must already exist, which seems backwards?!  I have a feeling the `mkdirs()` call that deep is also the cause of another archetype bug report?

Still, you're right about the existing wizard panel contract, so I guess maybe this is the right fix for 12.0 - I think it will fix the issues I've seen - did more digging and not seen another problem, except more suspicious uses of that parameter.

I'm not sure I agree that the contract is flawed though - Gradle support seems to be doing what the Maven wizards should be doing? https://github.com/apache/netbeans/blob/master/groovy/gradle/src/org/netbeans/modules/gradle/spi/newproject/SimpleGradleWizardIterator.java#L85

errael:
> Yes, because target folder on the template takes a DataFolder, which means the thing you're trying to create must already exist, which seems backwards?! I have a feeling the mkdirs() call that deep is also the cause of another archetype bug report?

If you're talking about the `((TemplateWizard) d).setTargetFolderLazy(`, this is recent, and part of an API change introducing the Lazy version. Before this, if Cancel button used, directories were left scattered around; particularly ugly if back/next had been used leaving a chain of empty directories. In BasicPanelVisual
```
         d.putProperty(CommonProjectActions.PROJECT_PARENT_FOLDER, parentFolder);
         if (d instanceof TemplateWizard) {
-            parentFolder.mkdirs();
-            ((TemplateWizard) d).setTargetFolder(DataFolder.findFolder(FileUtil.toFileObject(parentFolder)));
+            ((TemplateWizard) d).setTargetFolderLazy(() -> {
+                parentFolder.mkdirs();
+                return DataFolder.findFolder(FileUtil.toFileObject(parentFolder));
+            });
         }
````
If you're curious about the full change
```
changeset:   3987:b9e1311c588c
user:        Ernie Rael <errael@raelity.com>
date:        Sat Feb 29 21:41:39 2020 +0000
summary:     [NETBEANS-3918] defer creation of target folder until/if needed
```

neilcsmith-net:
> If you're talking about the ((TemplateWizard) d).setTargetFolderLazy(, this is recent, and part of an API change introducing the Lazy version.

No, I'm not. I'm aware of that change. But after looking more at this issue, my reading of TemplateWizard is that this is intended to be the folder _in which_ the new file or project is created, the parent of the project itself. I'm not sure the lazy API actually was needed to fix the prev/next folder creation issue, although useful if typing in parent name.

errael:
> not sure the lazy API actually was needed to fix the prev/next folder creation issue

Yeah, that fix was to avoid directory creation when Cancel. prev/next just made it a spectacular failure.

I've never worked directly with TemplateWizard, hope it stays that way ;-) I'd guess newer stuff, spi.project.ui..., got glued to modified ant java stuff, but wasn't lined up quite right.

The new stuff using "projdir" seems like a mistake; or at least a poor choice.

neilcsmith-net:
> Yeah, that fix was to avoid directory creation when Cancel.

That was my point - it was only creating a directory at all because it was passing the wrong argument.

Still, thanks for fixing my issue with this for the release.  Post 12.0, and if Maven support is moved away from being friend-only, then it should probably be revised to be in spec.

PR #2099, opened 2020-04-23
Opened up multiple final and package private classes and methods in Nodes and Propertysheet APIs

makusuko:
Opened up multiple final and package private classes and methods to support more flexible subclassing and modifications in the Nodes and Propertysheet APIs.

See https://issues.apache.org/jira/browse/NETBEANS-4222.


matthiasblaesing:
As Geertjan said in Jira, this should be discussed first. Opened API can become a burden and this means without concrete usecases, that can't be realized with the current API this won't fly. It is also a long time commitment, so will you be around to fix issues in 2-3 years?

JaroslavTulach:
I guess I scared the reporter enough... Sorry for that, I hope you will not give up on NetBeans, Markus.

PR #2102, opened 2020-04-23
[NETBEANS-4203] for InputOutput.reset() do NbWriter.reset() as needed

errael:
The new `I/O API and SPI`, introduced around 2014, has never had a working "reset()" to clear the output window, AFAICT. 

It looks like this new interface delegates to platform/core.output2 and this in turn is based on the old stuff in platform/openide.windows. The new interface gave us the long promised InputOutput.reset() and hid the older OutputWriter.reset(). The older OutputWriter.reset has some essential code that the newer InputOutput.reset() never invokes.

This PR makes sure there's a full reset and the output window is cleared when using the default api implementation.

Testing is tricky. The tests for the new stuff test that delegation occurs. It never actually tests a real output window. Based on the tests for the old api I've put together some tests as a module that uses reflection for verification. I don't know if there's any way to incorporate these tests into netbeans. I'm cleaning them up and will attach them to the issue.

### DIFF
--- a/platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
+++ b/platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
@@ -25,6 +25,7 @@ import java.io.Reader;
 import java.util.EnumSet;
 import java.util.Set;
 import java.util.WeakHashMap;
+import java.util.logging.Level;
 import javax.swing.Action;
 import org.netbeans.api.io.Hyperlink;
 import org.netbeans.api.io.OutputColor;
@@ -214,7 +215,14 @@ public final class NbIOProvider extends IOProvider implements
     @Override
     public void resetIO(InputOutput io) {
         if (io instanceof NbIO) {
-            ((NbIO) io).reset();
+            try {
+                // ((NbWriter)io.getOut).reset() does OutWriter fixup,
+                // then NbWriter invokes ((NbIO) io).reset()
+                io.getOut().reset();
+            } catch (IOException ex) {
+                Exceptions.attachSeverity(ex, Level.WARNING);
+                Exceptions.printStackTrace(ex);
+            }
         } else {
             throw new IllegalArgumentException();
         }


platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
```diff
@@ -214,7 +214,17 @@ public Lookup getIOLookup(InputOutput io) {
     @Override
     public void resetIO(InputOutput io) {
         if (io instanceof NbIO) {
-            ((NbIO) io).reset();
+            if(io.getOut() instanceof NbWriter) {
+                try {
+                    // NbWriter does OutWriter fixup,
+                    // then NbWriter invokes ((NbIO) io).reset()
+                    io.getOut().reset();
```
eirikbakke, line 221: The comment seems correct. At first I thought maybe io.getIn().reset() would also need to be called, but as the comment says, we the implementation of NbWriter will call NbIO.reset() in turn..

platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
```diff
@@ -214,7 +214,17 @@ public Lookup getIOLookup(InputOutput io) {
     @Override
     public void resetIO(InputOutput io) {
         if (io instanceof NbIO) {
-            ((NbIO) io).reset();
+            if(io.getOut() instanceof NbWriter) {
```
eirikbakke, line 217: space after if

platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
```diff
@@ -214,7 +214,17 @@ public Lookup getIOLookup(InputOutput io) {
...
+                try {
+                    // NbWriter does OutWriter fixup,
+                    // then NbWriter invokes ((NbIO) io).reset()
+                    io.getOut().reset();
+                } catch(IOException ex) {
+                    Exceptions.printStackTrace(ex);
```
eirikbakke, line 223: This seems to be the way errors are handled in this class, I guess.

platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
```diff
@@ -214,7 +214,17 @@ public Lookup getIOLookup(InputOutput io) {
     @Override
     public void resetIO(InputOutput io) {
         if (io instanceof NbIO) {
-            ((NbIO) io).reset();
+            if(io.getOut() instanceof NbWriter) {
+                try {
+                    // NbWriter does OutWriter fixup,
+                    // then NbWriter invokes ((NbIO) io).reset()
+                    io.getOut().reset();
+                } catch(IOException ex) {
```
eirikbakke, line 222: space before (

errael:
Since this was initially pushed, I have determined (convinced myself) that NbIO only has an NbWriter (in fact, there are examples of casting it without first testing its type) and so the patch can be simplified to
```
-            ((NbIO) io).reset();
+            try {
+                // ((NbWriter)io.getOut).reset() does OutWriter fixup,
+                // then NbWriter invokes ((NbIO) io).reset()
+                io.getOut().reset();
+            } catch (IOException ex) {
+                Exceptions.printStackTrace(ex);
+            }
```
Which I've run my manual tests against. Are there any objections to pushing this? (along with some space for @eirikbakke )

eirikbakke:
Yeah, that looks reasonable. Let's keep the PR up for a bit in case others want to chime in. Otherwise it looks safe to me.

platform/core.output2/src/org/netbeans/core/output2/NbIOProvider.java
JaroslavTulach: This will probably show a dialog to the user. If you just want to print errors to console, consider `Exceptions.attachSeverity` first.

errael:
(I still don't have the hang of githug, don't know how to add an additional commit to PR)

There is a test in #2130

I found a way to indirectly test the change...
Although can't test api.io's InputOutput.reset() directly. Can test the NbIOProvider change.

If your curious, the newly added testNbIOReset fails without the fix.

neilcsmith-net:
> (I still don't have the hang of githug, don't know how to add an additional commit to PR)

@errael Just keep pushing to the branch the PR is made from. It's a good idea not to use your master branch for PRs though.

ebarboni:
This Pr is done one master :/ that's maybe why it's not easy to append. Merging this one and the test late

phejl:
> The new `I/O API and SPI`, introduced around 2014, has never had a working "reset()" to clear the output window, AFAICT.
> 
> It looks like this new interface delegates to platform/core.output2 and this in turn is based on the old stuff in platform/openide.windows. The new interface gave us the long promised InputOutput.reset() and hid the older OutputWriter.reset(). The older OutputWriter.reset has some essential code that the newer InputOutput.reset() never invokes.

Can you explain a bit more on what was actually not working?

errael:
This is probably the key line, in the key method
https://github.com/apache/netbeans/blob/de3af4382a2f1d8a907288050f14355cbf41388c/platform/core.output2/src/org/netbeans/core/output2/NbWriter.java#L68
Before the change, this method was never called. And then because of that, OutputTab.reset simply returned because 
https://github.com/apache/netbeans/blob/de3af4382a2f1d8a907288050f14355cbf41388c/platform/core.output2/src/org/netbeans/core/output2/OutputTab.java#L184
...

errael:
In other words, to clear the window a new document (in OutputTab) had to be created with a new empty memory mapped file (in NbWriter).

PR #2103, opened 2020-04-23
[NETBEANS-3972] Draft implementation of GIT GPG signing

Caffe1neAdd1ct:
Initial draft of GIT GPG commit signing, I'm not happy with this implementation and need some guidance on how to improve this implementation.

1. Alteration of the commit() command construct signature breaks a lot of tests and doesn't feel like the right way to pass through the GPG passphrase to jgit/bouncycastle.
2. GPG Passphrase popup doesn't obfuscate the input and isn't a great UX being separate and before the commit window for file selection/author/committer etc
3. Override checkbox on the commit window to enable/disable signing like author/committer would also be useful.

I've built this on the 11.3 stable release of netbeans but will follow guidance on how to raise a new pull request as needed. To build/run at the moment all tests in Git and Git Client Library need disabling for now due to the commit command construct change.

I'm happy to continue attempting to implement this, but Java isn't my primary language (so please go easy).

Calling for suggestions and guidance on implementation.

### DIFF
--- a/ide/git/src/org/netbeans/modules/git/client/GitClient.java
+++ b/ide/git/src/org/netbeans/modules/git/client/GitClient.java
@@ -285,16 +285,16 @@ public final class GitClient {
         }, "clean", roots); //NOI18N
     }
     
-    public GitRevisionInfo commit (final File[] roots, final String commitMessage, final GitUser author, final GitUser commiter, final ProgressMonitor monitor) throws GitException {
-        return commit(roots, commitMessage, author, commiter, false, monitor);
+    public GitRevisionInfo commit (final File[] roots, final String commitMessage, final GitUser author, final GitUser commiter, final String gpgPrivateKeyPassphase, final ProgressMonitor monitor) throws GitException {
+        return commit(roots, commitMessage, author, commiter, false, gpgPrivateKeyPassphase, monitor);
     }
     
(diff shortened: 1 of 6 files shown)


ebarboni:
@Caffe1neAdd1ct, please have discussion on dev list to find some help. I have no idea how to help here.

hotis:
Please, fix this ticket. NB release 14 is still affected.

I tried all instruction for Windows, but it doesn't work in NB, work fine manually in cmd with -S flag on commit.

neilcsmith-net:
@hotis please take that conversation elsewhere - mailing list or discussions - thanks!

FurmigaHumana:
is anyone working on this? I could take over

matthiasblaesing:
@FurmigaHumana at least there are no other discussion on going about this. I think the hard part is to define what you want to accomplish and how that might fit into extensions in the future. For example I would expect the support to hook into an OpenPGP agent, just as the SSH support hooks into the SSH agent, maybe not initially, but at least with an perspective. I would try to avoid adding new API if possible. It might be required to indicate "I want to sign" info, but at least the passphrase could be provided using a custom implemenation of `GitClientCallback` and modifying the `CommitCommand` to install a credentials provider, just as is done in the TransportCommand.

PR #2104, opened 2020-04-24
Fix ant property to exclude tests

hectorespert:
Netbeans build system has a project property to define a list of classes/packages to be excluded when run tests in a module. 
```java
test.config.stableBTD.includes=**/*Test.class

test.config.stableBTD.excludes=\
    **/CachingAndExternalPathsTest.class,\
    **/CachingAndExternalURLTest.class,\
    **/CachingPreventsFileTouchesTest.class,\
    **/LogReaderServiceTest.class,\
    **/NetigsoHasSAXParserTest.class
```
This PR removes the hardcoded value from the property to allow us to exclude tests again.

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -621,7 +621,6 @@
             <depend srcdir="${test.@{test.type}.src.dir}" destdir="${build.test.@{test.type}.classes.dir}" cache="${build.test.@{test.type}.dir}/depcache">
                 <classpath refid="test.@{test.type}.cp"/>
             </depend>
-            <property name="test.excludes" value=""/> <!-- # 113770 -->
             <nb-javac srcdir="${test.@{test.type}.src.dir}" destdir="${build.test.@{test.type}.classes.dir}" excludes="${test.excludes}"
                       debug="true" deprecation="${build.compiler.deprecation}" encoding="UTF-8"
                       source="${javac.source}" target="${javac.target}" optimize="${build.compiler.optimize}" includeantruntime="false">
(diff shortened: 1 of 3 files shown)


jlahoda:
This is awesome - I actually ran into the test.excludes problem a few days ago as well. I wonder if the property on do-test-build shouldn't be renamed (rather than removed), e.g. to "test.javac.excludes". Note it is used at the line following the one that sets the value.

hectorespert:
I'm not sure but default value of property `test.excludes` is setted in this line: https://github.com/apache/netbeans/blob/c47210cc8266257bfec4e502bb80ea964669d0ea/nbbuild/templates/common.xml#L708

jlahoda:
Right, but test.excludes is actually used for two things in the scripts: for test compile excludes and test runtime excludes. But the formats are different for these, so what I am suggesting is to us test.javac.excludes for compilation, and keep test.excludes for runtime.

nbbuild/templates/common.xml
```diff
@@ -621,7 +621,6 @@
             <depend srcdir="${test.@{test.type}.src.dir}" destdir="${build.test.@{test.type}.classes.dir}" cache="${build.test.@{test.type}.dir}/depcache">
                 <classpath refid="test.@{test.type}.cp"/>
             </depend>
-            <property name="test.excludes" value=""/> <!-- # 113770 -->
             <nb-javac srcdir="${test.@{test.type}.src.dir}" destdir="${build.test.@{test.type}.classes.dir}" excludes="${test.excludes}"
```
jlahoda, line 624: Basically, I guess I am suggesting this:
```suggestion
            <property name="test.javac.excludes" value=""/> <!-- # 113770 -->
            <nb-javac srcdir="${test.@{test.type}.src.dir}" destdir="${build.test.@{test.type}.classes.dir}" excludes="${test.javac.excludes}"
```

hectorespert: I changed it

PR #2106, opened 2020-04-27
Run, debug JVM JavaScript with Graal.js as well as Nashorn

JaroslavTulach:
The standard JDK's JavaScript engine *Nashorn* [has already been deprecated](https://www.youtube.com/watch?v=SraYjqB0oX8) for several years. Recently it has been [removed from the JDK](https://openjdk.java.net/jeps/372). Time to find more places where NetBeans functionality depends on Nashorn and make it more flexible.

One of the areas is support for executing and debugging standalone JavaScript files. Here is my pull request that enhances the options UI to allow one to switch between Nashorn and *Graal.js* (already present in NetBeans). My PR also makes sure that whenever Nashorn is missing Graal.js is selected to do the right job.

The PR registers Graal.js as a Java library (including its launcher). Then it modifies the run/debug actions to put that library on classpath - that's all necessary to make sure Graal.js runs on any JDK8+.


java/java.source.ant/src/org/netbeans/modules/java/source/ant/ProjectRunnerImpl.java
```diff
@@ -122,8 +122,7 @@
     private static final RequestProcessor RP = new RequestProcessor(ProjectRunnerImpl.class);
     
     public boolean isSupported(String command, Map<String, ?> properties) {
-        return BuildArtifactMapper.isCompileOnSaveSupported() &&
-               locateScript(command) != null;
+        return locateScript(command) != null;
```
JaroslavTulach, line 125: Without this change, @jlahoda, the Ant runner refused to perform `run` and `debug` actions. I am sure the check is here for a reason, but it is probably not correct to prevent `run`/`debug` just because we don't have `nbjavac` installed.

Mohan-Sarilla:
@JaroslavTulach , I have built the code and tested , The following are observations:

JDK Version | Nashorn Option in ComboBox | Graal.js Option in ComboBox | Run Action with Nashorn | Run Action with Graal.js | Debug Action with Nashorn | Run Action with Graal.js
-- | -- | -- | -- | -- | -- | --
JDK8 | Yes | Yes | Yes | Yes | Yes | Yes
JDK11 | Yes | Yes | Yes | Yes | Yes | Yes
JDK13 | Yes | Yes | Yes | Yes | Yes | Yes
JDK14 | Yes | Yes | Yes | Yes | Yes | Yes
JDK15 | No | Yes | No | Yes | No | Yes


- By using Graal.js debugging is working fine, But variables info is not showing in the Variables tab of Debugging window.

- If Netbeans launch with JDK15, By interchanging JDK Platform on **JS on JVM** Panel to Nashorn supported JDK Version, We could see Nashorn Option populating as expected in Engine ComboBox , By selecting Nashorn Option from that Engine ComboBox, still code is executing with **Graal.js** only.

JaroslavTulach:
Thanks @Mohan-Sarilla for your review and testing.
> variables info is not showing in the Variables tab of Debugging window.
I guess I leave this problem for @entlicher - it'd be an issue in `java/debugger.jpda.truffle` module

I'll take a look at 
> selecting Nashorn  from Engine ComboBox, still code is executing with Graal.js only.
before integrating.

Mohan-Sarilla:
@JaroslavTulach , I have tested the updated fix, It works good.

**If Netbeans launch with JDK15, By interchanging JDK Platform on JS on JVM Panel to Nashorn supported JDK Version, We could see Nashorn Option populating as expected in Engine ComboBox , By selecting Nashorn Option from that Engine ComboBox, still code is executing with Graal.js only.**

arusinha:
> @JaroslavTulach , I have tested the updated fix, It works good for 2nd Observation.

👍

PR #2109, opened 2020-04-28
Mark included 3rd-party update centre as trusted.

neilcsmith-net:
Mark included 3rd-party update centre as trusted and don't sign bundled nbms. This should fix the issue with the signature of the bundled nbms not matching the keystore in the updatecenter nbm - a problem for anyone using the distribution nbms, and for pushing updates to nb-javac or JavaFX to the IDE after release.

Deliberately kept light touch for 12.0.  We can probably drop all the `ide.ks` and `NetBeansKeyStoreProvider` stuff for 12.1?

### DIFF
--- a/nb/updatecenters/build.xml
+++ b/nb/updatecenters/build.xml
@@ -41,8 +41,6 @@
       />
       <subant  target="nbm" inheritall="false">
           <property name="build.dir" location="build/3rdparty-nbms"/>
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@matthiasblaesing thanks! Yes, started pulling out the ide.ks stuff from the build but kept getting more and more things complaining, and not sure if there was a reason we had an empty ide.ks file before in 9.0. I'll do a post-release PR to remove it all.

PR #2110, opened 2020-04-29
[NETBEANS-4280] - cleanup potential security breaches

BradWalker:
There are a few known security breaches in the sample source..

Specifically the following alerts:

CVE-2019-5484
Bower before 1.8.8 has a path traversal vulnerability permitting file write in arbitrary locations via install command, which allows attackers to write arbitrary files when a malicious package is extracted.

CVE-2019-5413
An attacker can use the format parameter to inject arbitrary commands in the npm package morgan < 1.9.1.

CVE-2017-16137
The debug module is vulnerable to regular expression denial of service when untrusted user input is passed into the o formatter. It takes around 50k characters to block for 2 seconds making this a low severity issue.

I'm not saying these are critical. But, it's better we fix them to prevent any possibility of using Netbeans IDE to allow someone to exploit this. As well as set the proper example.

### DIFF
--- a/webcommon/javascript.nodejs/samples_src/MessagesAngular/package.json
+++ b/webcommon/javascript.nodejs/samples_src/MessagesAngular/package.json
@@ -10,7 +10,7 @@
     "dependencies": {
         "express": "^4.10.1",
         "body-parser": "^1.4.3",
-        "bower": "~1.3.12"
+        "bower": "~1.8.8"
     },
     "devDependencies": {
         "gulp": "^3.8.10",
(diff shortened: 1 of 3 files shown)


ebarboni:
Hi @BradWalker do you have a reviewer in mind, I'm not very skilled in this topic

BradWalker:
Hey @ebarboni , I don't really have a reviewer in mind..
This is only a version bump to the packages. In addition, the changes are in sample src. So my thinking was this is very low on the risk scale.

PR #2113, opened 2020-04-30
[TRAVIS] Add missing license to correct failing verification

graben:
@ebarboni : Should be merge to avoid actual travis build to fail.


graben:
Hi @lkishalmi, your latest Freemarker update from #2067 seems to break Travis build at all. Hopefully I fixed the right thing on the fly ;)

graben:
Well, there also unit test seems to fail.

graben:
Ignore failing tests for now. Must be completely rewritten since freemarker has changed it error handling. Much more chatty :)

lkishalmi:
Thanks for fixing my bad. I do not know how could I push that incomplete PR in the first place. I'm sorry.

jlahoda:
Looks good to me. Eric, unless there are objections, or unless you want to do it yourself, I'll merge in a few hours. (The only alternative I see would be to revert #2067 - but I think fixing it is better!)

Thanks @graben for fixing this!

graben:
Just fixed another minor file returning 2.3.19 instead of 2.3.30.

PR #2115, opened 2020-05-01
[NETBEANS-4291] Fix some classes not found from multi-release JARs in Java editor

neilcsmith-net:
If a multi-release JAR contains resources with the same name but a different extension, then only one of these resources is picked up.  This causes errors with projects using eg. FlatLaf where errors are seen in the code editor while compilation works fine.  Files such as `FlatLightLaf.properties` shadows `FlatLightLaf.class`.

This is caused by an error in CachingFileManager and ModuleFileManager using the file name and location without extension as a map key to look up multi-release files. Hopefully changes here fix this without causing problems elsewhere - better tests would be good in this area in future.

See also https://github.com/JFormDesigner/FlatLaf/issues/13

The projects attached to the JIRA issue demonstrate the issue on classpath and modulepath. Both show broken in NB 12.0-beta3. https://issues.apache.org/jira/browse/NETBEANS-4291

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/CachingFileManager.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/CachingFileManager.java
@@ -247,6 +247,8 @@ public class CachingFileManager implements JavaFileManager, PropertyChangeListen
                 Archive archive = provider.getArchive( entry.getURL(), cacheFile );
                 if (archive != null) {
                     final Iterable<JavaFileObject> entries;
+                    // multi-release code here duplicated in ModuleFileManager
+                    // fixes should be ported across, or ideally this logic abstracted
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@jlahoda be grateful if you could check this looks correct?

@DevCharly can't seem to add you as a reviewer, but you might want to check this?

DevCharly:
I've tested/debugged the changes and it fixes the issue. Many thanks.

Don't think that that the fix may cause problems elsewhere.
`fqn2f` is a method local map and the changed map keys are not passed to the outside of the method. Only the values of `fqn2f` are used.

neilcsmith-net:
@DevCharly thanks for testing.  Glad it works for you.  Yes, the map keys are local, so don't escape - only problem would be if the filtering is still incorrect, just in a different way.  Changed the delimiter to `/` because of one concern there.

PR #2120, opened 2020-05-02
Added property to allow the user to specify the max number of objects and details in search:

EnricoScantamburlo:
Some clients of ours had this problem, they had a lot of files to search inside, but NetBeans limited the maximum number of results to 500.

### DIFF
--- a/platform/api.search/src/org/netbeans/modules/search/Constants.java
+++ b/platform/api.search/src/org/netbeans/modules/search/Constants.java
@@ -29,11 +29,11 @@ public final class Constants {
     /**
      * maximum number of found objects
      */
-    public static final int COUNT_LIMIT = 500;
+    public static final int COUNT_LIMIT = Integer.getInteger("netbeans.search.count.limit", 500);
     /**
      * maximum total number of detail entries for found objects
      */
-    public static final int DETAILS_COUNT_LIMIT = 5000;
+    public static final int DETAILS_COUNT_LIMIT = Integer.getInteger("netbeans.search.details.count.limit", 5000);
 
     public enum Limit {
 


lkishalmi (approved):
Seems to be trivial enough.

platform/api.search/src/org/netbeans/modules/search/Constants.java
```diff
@@ -29,11 +29,11 @@
     /**
      * maximum number of found objects
      */
-    public static final int COUNT_LIMIT = 500;
+    public static final int COUNT_LIMIT = Integer.getInteger("netbeans.search.count.limit", 500);
```
JaroslavTulach, line 32: This shall be documented in `arch.xml` as an `<api group="systemproperty" ...>some description</api>`. See [howto](http://wiki.netbeans.org/APIDevelopment#Writing_the_documentation) for more details about `arch.xml`. 

Also one shall version the change. E.g. increase the module version and leave a note in `apichanges.xml` file.

platform/api.search/src/org/netbeans/modules/search/Constants.java
```diff
@@ -29,11 +29,11 @@
...
     /**
      * maximum total number of detail entries for found objects
      */
-    public static final int DETAILS_COUNT_LIMIT = 5000;
+    public static final int DETAILS_COUNT_LIMIT = Integer.getInteger("netbeans.search.details.count.limit", 5000);
```
JaroslavTulach, line 36: Dtto.

JaroslavTulach (commented):
The API change isn't properly documented. Can that be fixed?

PR #2121, opened 2020-05-02
 Adding tests for java.source.base without nb-javac and java.hints tests to Travis; fixing tests and code along the way.

jlahoda:
Some excludes still apply for both java.source.base and java.hints when testing without nb-javac, but that can hopefully be eliminated eventually.


arvindaprameya:
+ 1 Approved, thanks for this change

PR #2122, opened 2020-05-02
[NETBEANS-3413] Implicit inheritance of PHPDoc descriptions

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-3413
Uses description of inherited element when no PHPDoc is present.

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
@@ -32,6 +32,7 @@ import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.netbeans.api.annotations.common.CheckForNull;
+import org.netbeans.api.annotations.common.NullAllowed;
 import org.netbeans.api.lexer.Token;
 import org.netbeans.api.lexer.TokenSequence;
 import org.netbeans.api.project.FileOwnerQuery;
@@ -242,9 +243,10 @@ final class DocRenderer {
         private final PhpElement indexedElement;
         private final List<String> links = new ArrayList<>();
         private final ASTNode node;
+        @NullAllowed
         private PHPDocBlock phpDocBlock;
 
-        public PHPDocExtractor(CCDocHtmlFormatter header, PhpElement indexedElement, ASTNode node, PHPDocBlock phpDocBlock) {
+        public PHPDocExtractor(CCDocHtmlFormatter header, PhpElement indexedElement, ASTNode node, @NullAllowed PHPDocBlock phpDocBlock) {
             this.header = header;
             this.indexedElement = indexedElement;
             this.node = node;
@@ -380,9 +382,6 @@ final class DocRenderer {
         }
 
         private void extractPHPDocBlock() {
-            if (phpDocBlock == null) {
-                return;
-            }
             List<PHPDocVarTypeTag> params = new ArrayList<>();
             List<PHPDocTypeTag> returns = new ArrayList<>();
             StringBuilder others = new StringBuilder();
@@ -391,10 +390,11 @@ final class DocRenderer {
             List<PhpElement> inheritedElements = getInheritedElements();
             List<PHPDocBlock> inheritedComments = getInheritedComments(inheritedElements);
 
-            String description = phpDocBlock.getDescription();
+            String description = phpDocBlock == null ? null : phpDocBlock.getDescription();
             description = composeDescription(description, inheritedComments);
 
-            for (PHPDocTag tag : phpDocBlock.getTags()) {
+            List<PHPDocTag> tags = phpDocBlock == null ? Collections.emptyList() : phpDocBlock.getTags();
+            for (PHPDocTag tag : tags) {
                 AnnotationParsedLine kind = tag.getKind();
                 if (kind.equals(PHPDocTag.Type.PARAM)) {
                     PHPDocVarTypeTag paramTag = (PHPDocVarTypeTag) tag;
@@ -574,7 +574,11 @@ final class DocRenderer {
             return ret;
         }
 
-        private String replaceInheritdocForDescription(String description, String parentDescription) {
+        @CheckForNull
+        private String replaceInheritdocForDescription(@NullAllowed String description, @NullAllowed String parentDescription) {
+            if (description == null) {
+                return parentDescription;
+            }
             if (description != null && hasInlineInheritdoc(description)) {
                 if (parentDescription != null && !parentDescription.trim().isEmpty()) {
                     if (INLINE_INHERITDOC_PATTERN.matcher(description.trim()).matches()) {
@@ -717,7 +721,7 @@ final class DocRenderer {
 
         private boolean needInheritedElements() {
             if (phpDocBlock == null) {
-                return false;
+                return true;
             }
             String description = phpDocBlock.getDescription();
             if (hasInlineInheritdoc(description)) {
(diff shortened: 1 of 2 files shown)


KacerCZ:
@junichi11 @sdedic @tzezula Please review.

junichi11:
@tmysik Maybe, I restricted it for performance when I implemented it. I think that a user should write `{@inheritdoc}` explicitly. What do you think?

tmysik:
@junichi11 I would perhaps try the Java editor and simply copy its behavior.

BTW can this change have any negative performance impact?

junichi11:
@tmysik Thank you!

> I would perhaps try the Java editor and simply copy its behavior.

Agree. Java editor just shows a parent description.

> BTW can this change have any negative performance impact?

AFAIR, a description is gotten from the index until the parent one is found. I'm not sure exactly, but at least, I suppose performance would be bad if there are many items.

OK, Let's show a parent description if there is no PHPDoc. If there is a problem, someone would report about it, maybe.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -391,10 +388,11 @@ private void extractPHPDocBlock() {
             List<PhpElement> inheritedElements = getInheritedElements();
             List<PHPDocBlock> inheritedComments = getInheritedComments(inheritedElements);
 
-            String description = phpDocBlock.getDescription();
+            String description = phpDocBlock == null ? null : phpDocBlock.getDescription();
             description = composeDescription(description, inheritedComments);
 
-            for (PHPDocTag tag : phpDocBlock.getTags()) {
+            List<PHPDocTag> tags = phpDocBlock == null ? new ArrayList<>() : phpDocBlock.getTags();
```
junichi11, line 394: Should use `Collections.emptyList()` instead of `new ArrayList<>()`

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -575,6 +573,9 @@ private String composeDescription(String description, List<PHPDocBlock> comments
         }
 
         private String replaceInheritdocForDescription(String description, String parentDescription) {
+            if (description == null ) {
```
junichi11, line 576: Please remove whitespace before `)`

KacerCZ: Fixed.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
KacerCZ: Changed, thank you.

junichi11:
https://github.com/apache/netbeans/blob/cefa138b237cea1daea709747532cb91945b62c5/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java#L245-L247
Could you add `@NullAllowed` as the following? This is my fault, sorry.
```java
        @NullAllowed
        private PHPDocBlock phpDocBlock;

        public PHPDocExtractor(CCDocHtmlFormatter header, PhpElement indexedElement, ASTNode node, @NullAllowed PHPDocBlock phpDocBlock) {

```

KacerCZ:
@junichi11 I've added the annotation. No problem.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -242,6 +243,7 @@ private static String getLocation(PhpElement indexedElement) {
         private final PhpElement indexedElement;
         private final List<String> links = new ArrayList<>();
         private final ASTNode node;
+        @NullAllowed
         private PHPDocBlock phpDocBlock;
 
         public PHPDocExtractor(CCDocHtmlFormatter header, PhpElement indexedElement, ASTNode node, PHPDocBlock phpDocBlock) {
```
tmysik, line 249: Missing `@NullAllowed` for `phpDocBlock` param, right?

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -575,6 +575,9 @@ private String composeDescription(String description, List<PHPDocBlock> comments
         }
 
         private String replaceInheritdocForDescription(String description, String parentDescription) {
```
tmysik, line 577: Missing `@NullAllowed` for `description` param.

tmysik: OT: And for `parentDescription` as well, could you please add it? Thanks.

tmysik: OT: Also, the whole method should be annotated by `@CheckForNull` (IIRC) since it can return `null` if I understand it correctly. Thanks.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -768,6 +771,9 @@ private boolean isOnlyInheritdoc(PHPDocBlock phpDocBlock) {
         }
 
         static String replaceInlineInheritdoc(String description, String inheritdoc) {
```
tmysik, line 773: OT: Again, missing annotations for `null` values.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
tmysik: Resolving because of no change of this line in this MR (but still, would be nice to have it done).

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
tmysik: Resolving because of no change of this line in this MR (but still, would be nice to have it done).

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
tmysik: Now I see that @junichi11 already requested this change, sorry.

junichi11: Yes, @tmysik is right.

junichi11:
@tmysik Thank you for your review :)

@KacerCZ Could you resolve annotations that @tmysik pointed out if possible? Thank you for your work :)

We can't merge this because the merge window closes yet. Let's merge this after NetBeans 12.0 is released.

KacerCZ:
@junichi11 @tmysik I've added annotations mentioned in review.
Thank you for reviews and accepting this PR.

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
junichi11: @KacerCZ 
Missing `@NullAllowed` yet:
```java
public PHPDocExtractor(CCDocHtmlFormatter header, PhpElement indexedElement, ASTNode node, @NullAllowed PHPDocBlock phpDocBlock) {
```

KacerCZ: Fixed.

junichi11: Thanks!

PR #2123, opened 2020-05-02
[NETBEANS-4300] Typo in "default platform" value for Maven property netbeans.hint.jdkPlatform

anthonyvdotbe:


ebarboni:
hi @anthonyvdotbe  what is the intention ? what is fixed by this change ?

anthonyvdotbe:
hi @ebarboni. NetBeans wrote a property in my POM as follows:

```xml
<netbeans.hint.jdkPlatform>@@DEFAU:T@@</netbeans.hint.jdkPlatform>
```

which is just ugly and an obvious typo. So the intention is to make NetBeans write sensible properties into people's POMs

anthonyvdotbe:
> Why not remove the property from the POM? (override the value from a parent pom?)

Not sure if I understand what you're asking, but if it is: why don't you remove the property from the POM? Then, yes, it doesn't make sense to specify `@@DEFAULT@@` as the value, but if NetBeans writes it to the POM, it should at least do so without typos. And ideally it shouldn't write it to the POM at all, but that's an issue of its own

> How is this story about backwards compatibility?

Imho this is a negligible issue. At worst, people will get an `Unresolved project problem` saying `No such Java Platform: @@DEFAU:T@@`, upon which they'll select a Java Platform of their choosing in the settings and that's it. But I doubt this will impact many users at all.

matthiasblaesing:
@anthonyvdotbe netbeans does not write to the pom.xml, but it writes to a companion file `nb-configuration.xml`. The codepath to change the pom.xml is only taken, if someone added the property to the pom.xml  (or a parent pom) manually (or maybe in an old version of netbeans this was done).

So I put some more work into this (again, this is evaluation of the issue) and my finding:

- the property is inherited from parent poms
- setting the value of the property to `@@DEFAU:T@@`  in a child module indeed restores this modules compiler settings to the default JDK
- removing the property from the pom is thus not a good way to explicitly choose the default JDK

I also checked what happens, when you set `@@DEFAULT@@`  in the pom and it turns out, that the same happens as with `@@DEFAU:T@@`, the project is flagged with a warning reporting a project problem.

So you found the tip of the iceberg. Do you want to go down the rabbit hole and try a full fix or stay with the fix for the typo?

anthonyvdotbe:
> I also checked what happens, when you set `@@DEFAULT@@` in the pom and it turns out, that the same happens as with `@@DEFAU:T@@`, the project is flagged with a warning reporting a project problem.

Not sure I understand what you're saying. Currently `@@DEFAU:T@@` works, and `@@DEFAULT@@` gives a project problem. After this fix, it would be the inverse. Right?

> So you found the tip of the iceberg. Do you want to go down the rabbit hole and try a full fix or stay with the fix for the typo?

I'd rather stay with the fix for the typo. Primarily because of a lack of time. Secondarily because I don't see what a "full fix" would be? The issue is "there's a typo, which might appear in a user's POM" and the current fix resolves exactly that issue.

matthiasblaesing:
This is from netbeans master (the icon shows an exclamation mark):

Screenshot

So it looks as if `@@DEFAU:T@@` is not recognized as a special value. Same goes for the Platform selector:

screenshot2

anthonyvdotbe:
Thanks, I see what you meant now, also w.r.t. the tip of the iceberg.

ebarboni:
thanks for review and discussion. merging so we hava less PR and this is an easy one

PR #2124, opened 2020-05-02
[NETBEANS-4283]:Code Modifications related to finalisation of Switch Expressions in JDK14

Mohan-Sarilla:
Code Modifications of Hints and their tests related to Switch Expressions, As Switch Expressions finalised in JDK14,


Mohan-Sarilla:
@EnricoScantamburlo , Can you please consider it for Netbeans 12.0

PR #2125, opened 2020-05-02
Commit author validation

hectorespert:
Proposal of commit author validation using a GitHub action: https://github.com/marketplace/actions/check-author-name-and-email


ebarboni:
Hi @hectorespert it would be super nice if it can print the name and mail in the log or somewhere. So we can check easily

hectorespert:
> Hi @hectorespert it would be super nice if it can print the name and mail in the log or somewhere. So we can check easily

@ebarboni They are printed now.

ebarboni:
thanks seems ok for me, would be good if @matthiasblaesing find it usefull.

matthiasblaesing:
Thank you for working on this - this solves one part of the problem (direct edits from github without a public email address), but it does not cover:

- international names with non-ascii first letter
- "creative" usernames like "Don't care"
- squash merges (because github decides its more clever, that git itself it replaceses the author information for the squash with the public information of the author)

Idea: We could create a "white" list of known authors and validate against that, the list would be filled with all Committers and known third party authors. If a new author comes along, he must be added to the list. I admit, that this is not a new idea, but taken from project Skara (the project, that prototypes migrating OpenJDK Development to github). The interesting question: How to save such a list to make it not too easy for spam bots.

hectorespert:
> squash merges (because github decides its more clever, that git itself it replaceses the author information for the squash with the public information of the author)

Why not disable squash merges from repo config? https://help.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

ebarboni:
there is some possibility in the apache yml file https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories

neilcsmith-net:
> Why not disable squash merges from repo config?

Surely we'd need to be clear on an alternative and how we work with contributors with PRs that need squashing.  This has the potential to slow down the contribution / release process and make another headache unless everyone is clear what approach needs to be taken?  Something for a thread on dev@ IMO.

hectorespert:
I also think that disable squash merges should be debated in a dev thread. 
I would like to maintain this pr only related to check the commit author data.

hectorespert:
> Would it be possible to have name and mail directly in the PR ?
> Check commits / Verify commits (pull_request) Successful in 3s can be
> Check commits / Verify commits (pull_request) Successful in 3s username: **** mail: ***** so we don't have to browse a lot to get info and can check the false positive

I don't know, probably yes, but it wouldn't be simple to implement

hectorespert:
>  Idea: We could create a "white" list of known authors and validate against that, the list would be filled with all Committers and known third party authors. If a new author comes along, he must be added to the list.

I updated the GitHub action to read a file that contains a list of checked authors.
```yaml
whitelist:
  - username: hectorespert
    md5: 4c49d58f19c10b3055ca3aea46f37b7c #md5("Name:Email")
```

> The interesting question: How to save such a list to make it not too easy for spam bots.

My first approach is to save the author name and email as a md5 hash.

matthiasblaesing:
I experimented also a bit and came up with this (implemented in java, but should be translateable to javascript):

https://github.com/matthiasblaesing/github-action-test
https://github.com/matthiasblaesing/github-action-test/blob/master/src/main/java/eu/doppel_helix/github/action/sanitycheck/Main.java

That implementation uses the github API to add a comment to the PR with the data summarizing the commit. The result can be seen here:

https://github.com/matthiasblaesing/github-action-test/pull/4

The drawback is, that a comment triggers a notification. This might be a way to reflect the real data of the commits. The check runs API might be a better fit for the use case, but I had not yet time to look deeper into that part of the github api.

hectorespert:
> I experimented also a bit and came up with this (implemented in java, but should be translateable to javascript):
> 
> https://github.com/matthiasblaesing/github-action-test
> https://github.com/matthiasblaesing/github-action-test/blob/master/src/main/java/eu/doppel_helix/github/action/sanitycheck/Main.java
> 
> That implementation uses the github API to add a comment to the PR with the data summarizing the commit. The result can be seen here:
> 
> [matthiasblaesing/github-action-test#4](https://github.com/matthiasblaesing/github-action-test/pull/4)

Looks good, I like your approach. It's better than my action, it creates less friction than a whitelist to verify the info.

hectorespert:
@matthiasblaesing What are your plans for your experiment? Are you going to publish the GitHub action?

matthiasblaesing:
@hectorespert yes, I'll revisit it, but currently I'm occupied with other work

hectorespert:
I would close this PR if we are going to use @matthiasblaesing  GitHub action. Or we are going to use both?

matthiasblaesing:
@hectorespert I just polished my implementation to put it into production and now learned, that it only worked, because my PRs were created from the same repository (I have a serious WTF moment) - I'll see what can I find as a alternative but at this point my idea is dead in the water.

hectorespert:
It is a pity, it looked very good. What about if I change this implementation to do little checks in the email and username? Only to check if this contains any reference to github,  a username with a number or any other little check. We can use it while we find an alternative.

hectorespert:
I close it, I'm going to test other approach https://github.com/apache/netbeans/pull/2168

PR #2126, opened 2020-05-02
[NETBEANS-4299] find endPos for synthetic annotation attribute

errael:
Fixes 12.0 regression from 11.3. (might be nb-javac to jdk14 issue)

This fixes "Use @NbBundle.messages" adding a message to
``` 
@Messages("OptionsCategory_Name_OptionsPlayHack=any old")
```
and also adding a message to
```
@Messages({"OptionsCategory_Name_OptionsPlayHack=any old", "OptionsCategory_Keywords_OptionsPlayHack=keyw1"})
```
Not sure it's the best fix, it seems safe; needs review.

Questions about the fix:
- I don't understand getKind() vs instanceof (if/when it matters)
- Not sure if this check fully qualifies "assignment as annotation", but don't think that matters.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
@@ -535,8 +535,14 @@ public class CasualDiff {
             VariableTree vt = fgt.getVariables().get(fgt.getVariables().size() - 1);
             return TreeInfo.getEndPos((JCTree)vt, oldTopLevel.endPositions);
         }
-        return TreeInfo.getEndPos(t, oldTopLevel.endPositions);
+        int endPos = TreeInfo.getEndPos(t, oldTopLevel.endPositions);
+
+        // [NETBEANS-4299], might be a synthetic annotation attribute, try rhs
+        if (endPos == Position.NOPOS && t instanceof JCAssign) {
+            endPos = TreeInfo.getEndPos(((JCAssign)t).rhs, oldTopLevel.endPositions);
         }
+        return endPos;
+    }
 
     private int endPos(com.sun.tools.javac.util.List<? extends JCTree> trees) {
         int result = -1;


java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
```diff
@@ -535,8 +535,18 @@ public int endPos(JCTree t) {
...
             return TreeInfo.getEndPos((JCTree)vt, oldTopLevel.endPositions);
         }
-        return TreeInfo.getEndPos(t, oldTopLevel.endPositions);
+        int endPos = TreeInfo.getEndPos(t, oldTopLevel.endPositions);
+
+        // [NETBEANS-4299], catch a synthetic annotation attribute and try rhs
+        if (endPos == Position.NOPOS && t instanceof JCAssign) {
+            JCAssign assign = (JCAssign)t;
+            if (assign.lhs instanceof JCIdent
+                    && ((JCIdent)assign.lhs).sym instanceof Symbol.MethodSymbol) {
```
jlahoda, line 544: I think the "&& ... instanceof MethodSymbol" is probably unnecessary.

errael: I guess a regular/real assignment would have been in oldTopLevel.endPositions. So since it failed to get endPos as an assignment, all that's needed is to simply check the rhs (the end should be after that). Checking the lhs shouldn't be needed at all. I was being most conservative, but any rhs match should be good. I guess it could just be
```
// [NETBEANS-4299], catch a synthetic annotation attribute and try rhs
if (endPos == Position.NOPOS && t instanceof JCAssign) {
    endPos = TreeInfo.getEndPos(((JCAssign)t).rhs, oldTopLevel.endPositions);
}
```
That feels less conservative and I don't know anything about the quirks in the trees.

@jlahoda  your call. Do you want either change or as is?

jlahoda: I'd say - simplify the code. (We will see if some tests fail.) Overall, touching symbols too much inside CausalDiff is not very good, although in this particular case it probably does not matter much.

errael: Simplified. Since first check is NOPOS, probably doesn't matter how much stuff is done; but I like simple. My manual tests passed.

Do you believe this is safe for 12.0? There's that milestone thingy... And/or might add @ebarboni as reviewer.

PR #2128, opened 2020-05-03
A prototype of CoS for vanilla javac.

jlahoda:
This improves the Java indexer used when nb-javac is not present to more granularly remove erroneous code, so that the part of the code that are correct can be run in the CoS mode.


JaroslavTulach (approved):
Looks like smaller changes than I was expecting. Thanks for the detailed test coverage.

eirikbakke:
I haven't had time to help much myself here, but the work on CoS for vanilla javac is highly appreciated! Looking forward to testing my platform application development workflow without nbjavac. Thanks!!

arvindaprameya:
Approved, thank you so much for doing this Jan

eirikbakke:
I've started using vanilla CoS during my daily development now. It works, including for my NetBeans Platform app and for the "Apply Changes" feature in the debugger! This is very good.

I have reported the first bug relating to CoS on vanilla javac here:
https://issues.apache.org/jira/browse/NETBEANS-5757 (Compile-on-Save drops static initializer)

PR #2129, opened 2020-05-04
Input output reset

errael:


errael:
Although can't test api.io's InputOutput.reset() directly. Can test the NbIOProvider change.

If your curious, the newly added testNbIOReset fails without the fix.

PR #2130, opened 2020-05-04
test NbIOProvider.resetIO(io) as used by api.io InputOutput.reset()

errael:
Test for change in #2102


errael:
@jtulach @ebarboni This PR is the test for #2102. Don't merge this without that fix.

PR #2131, opened 2020-05-05
[NETBEANS-4305] - cleanup dead hardware support

BradWalker:
Sun Microsystems has not produced a 32-bit microprocessor in years. In addition, Solaris is
now strictly 64-bit and has been that way for a LONG time.

So cleanup the platform support to remove support for 32-bit SPARC processors and Solaris.

Java doesn't even run on this equipment!


matthiasblaesing:
Are you sure about this? I have a solaris setup where I rebuild JNA on and I can still build the 32bit variant. You need a Java 7 JVM (at least Oracle/Sun dropped 32bit support on the way to 8), but then you can switch to 32bit.

BradWalker:
Hey Matthias,

Here are my opinions and I'm completely open to changing them because this brings up a topic that needs architectural comments from people like you.

I'm of the opinion that Netbeans has decided that Java 8 is our base supported Java and it's getting more difficult to get 32-bit binaries. In addition, Java on Solaris is deprecated on the Solaris SPARC platform and is slated for removal.

So at this point, I'm only suggesting that we remove support on 32-bit OS platforms. We will still keep 64-bit platform support..

Hopefully this is more clear..

jlahoda:
I'd like to point out that while the IDE itself requires JDK 8, it supports development of applications using older JDKs.

My understanding is that the binaries in nativeexecution and profiler may be used on some kind of remote target machine, they don't need to run on the same machine as the IDE. If there's anyone doing that with a 32-bit SPARC and Solaris, I don't know, but they surely would not need to run the IDE on that machine.

BradWalker:
Hey Jan,

The last 32-bit processor that Sun made was around the mid-90s. Sometime around 2006, Solaris OS removed all support for non 32-bit machines. It would still run 32-bit apps but since then that restriction has mostly been lifted.

So at this point, I think we can safely say that 32-bit processors and Solaris OS support for those are dead and can be removed.

lkishalmi:
I think it is fine for removing solaris 32bit support. If someone needs that for old hardware he might have an old netbeans laying around as well with which he/she can profile the old code.

BTW: I think profiling support for Java 1.5 can be removed as well.

lkishalmi (approved):

BradWalker:
Hey @lkishalmi , I will clean up the profiling support for Java 1.5 in another pull request.. Thanks for the approval.

BradWalker:
Hey  @matthiasblaesing , can I get a blessing from you on this.. I know you might not be a fan of this, given the previous comments, but could you still agree to move forward on it.

matthiasblaesing:
@BradWalker not sure whether you need my blessing :-) - I'm just attached to old architectures, because I was in the situation where a Java 1.4 VM was used in production as an embedded VM, currently a 1.5 VM standalone in production, so removing support without reason is always strange for me.

So I rephrase: If it helps, from my POV 32 bit solaris sparc can be considered obsolete and removed (Oracle as the main vendor of the platform already decided java support is not worth it anymore for the platform).

lkishalmi:
Well about the support of profiling 1.5 JVM. I think even NetBeans 8.5 is perfectly fine for that task. We are less sure about 12.0, hence I'd doubt anyone have tried that, but theoretically possible. One of the gray areas we have is the profiler support. The build does not really rebuild the source code with each release, just downloads some dusty zips and populate it's content. If we would like to move forward from this state, I think it's not bad to leave the rusty parts behind.
Also the state of the profiler worries me a little. Right now a modern JVM often crashes in a few seconds when attaching the profiler (at least on Ubuntu). It might be just some minor binary incompatibility or something else. So it would be at least good to have the profiler binaries rebuilt with each release.
Though I'm going to open a discussion on that on the dev mailing list.

BradWalker:
Hey @matthiasblaesing , I agree with your comments about removing old h/w support. As a Sun person, it really, really bothers me to remove the old support because: 1 - the support will be gone forever, 2 - we loose historical information when this happens, and 3 - it's just one more door closed that reminds me of the past.. 8-)

BradWalker:
Lastly, @matthiasblaesing since Laszlo approved it.. Can you merge the pull request? Please.. 8-)

profiler/lib.profiler.common/nbproject/org-netbeans-lib-profiler-common.sig
```diff
@@ -384,7 +384,6 @@ fld public final static java.lang.String PLATFORM_MAC_OS
 fld public final static java.lang.String PLATFORM_SOLARIS_AMD64_OS
 fld public final static java.lang.String PLATFORM_SOLARIS_INTEL_OS
 fld public final static java.lang.String PLATFORM_SOLARIS_SPARC64_OS
-fld public final static java.lang.String PLATFORM_SOLARIS_SPARC_OS
```
matthiasblaesing, line 387: One last request: Could we rework this, so that compatibility is retained? This should be binary compatible, as "final static"s are resolved at compile time and not runtime, but it is an source incompatible change. Is there harm in retaining the string?

BradWalker:
Hey @matthiasblaesing , I reverted the change in _profiler/lib.profiler.common/nbproject/org-netbeans-lib-profiler-common.sig_ you asked for.. Seem pretty reasonable.. Thanks for the good eyes!

profiler/lib.profiler.common/src/org/netbeans/lib/profiler/common/integration/IntegrationUtils.java
```diff
@@ -60,7 +60,6 @@
     public static final String PLATFORM_LINUX_CVM = messages.getString("IntegrationUtils_PlatformLinuxCvm"); // NOI18N
     public static final String PLATFORM_SOLARIS_INTEL_OS = messages.getString("IntegrationUtils_PlatformSolarisIntelOs"); // NOI18N
     public static final String PLATFORM_SOLARIS_AMD64_OS = messages.getString("IntegrationUtils_PlatformSolarisAmd64Os"); // NOI18N
-    public static final String PLATFORM_SOLARIS_SPARC_OS = messages.getString("IntegrationUtils_PlatformSolarisSparcOs"); // NOI18N
```
matthiasblaesing, line 63: The change in the sig file is a result from the change in this file - the string `PLATFORM_SOLARIS_SPARC_OS` needs to be retained.

matthiasblaesing:
Ok - lets get this in.

PR #2132, opened 2020-05-06
[NETBEANS-4310] - Update asm from 7.2 to 8.0.1

pepness:
Notes:
- New V15 constant
- Experimental support for PermittedSubtypes and RecordComponent
- Bug fixes (317885): SKIP_DEBUG now skips MethodParameters attributes
- Java 14 support (RecordComponent)
- Bug fixes (317896): Performance degradation when using dynamic constants as a static paramet to another InDy/ConDy
- Bug fixes (317898): ASM 8 does not support writing of empty records

[ASM Web Page](https://asm.ow2.io/index.html)
[Release Notes](https://asm.ow2.io/versions.html)

### DIFF
--- a/harness/apisupport.harness/build.xml
+++ b/harness/apisupport.harness/build.xml
@@ -39,13 +39,13 @@
                 <pathelement location="${cluster}/tasks.jar"></pathelement>
             </classpath>
         </taskdef>
-        <echo file="build/binaries-list">FA637EB67EB7628C915D73762B681AE7FF0B9731 org.ow2.asm:asm:7.2</echo>
+        <echo file="build/binaries-list">3F5199523FB95304B44563F5D56D9F5A07270669 org.ow2.asm:asm:8.0.1</echo>
         <TestDownload>
             <manifest dir="build">
                 <include name="binaries-list"/>
             </manifest>
         </TestDownload>
-        <delete file="build/asm-7.2.jar"/>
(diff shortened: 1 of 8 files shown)


ebarboni:
@sdedic safe for 12.0 beta5 ?

neilcsmith-net:
@ebarboni the JIRA ticket has 12.1 on it, which sounds the safer bet!

I wish people would add information from a NetBeans perspective when suggesting library updates - what's it fix / add here?

sdedic:
@ebarboni I'd say it's safe, but AFAIK no bugfix/feature in the new lib is needed for 12.0; @pepness - what's your opinion ?

pepness:
I think it is safe, this version was merged in [JDK-15](https://bugs.openjdk.java.net/browse/JDK-8241627), but it is true that no bugfix/feature in the new lib is needed for 12.0. This library is used by groovy/gradle;@lkishalmi - do you think it is reasonable to update this library every few releases?
@sdedic does a changes like [this](http://mail.openjdk.java.net/pipermail/core-libs-dev/2018-November/056600.html) would have affected the patching code on NetBeans (we already use ASM7)?

lkishalmi:
As of me it is good to do the upgrade in 12.1 What depends on it is Gradle Code coverage support. Wth Java 14 Support needs 8.0.1 ASM for Jacoco, however it is not released yet.

PR #2134, opened 2020-05-11
[NETBEANS-4259] Preventing crash in code completion for a lambda expr…

jlahoda:
…ession stub.


ebarboni:
make sense for 12.0 ?

PR #2141, opened 2020-05-15
[NETBEANS-4343] - remove unused MacroExpanderFactory methods

BradWalker:
I'm trying to cleanup some of the code that deals with native execution. This will be helpful to the CND module code that I'm working on..

Removing these two unused & old methods as there are no consumers and leads to API bloat..

public static MacroExpander getExpander(ExecutionEnvironment execEnv, boolean connectIfNeed)

public static MacroExpander getExpander(ExecutionEnvironment execEnv, ExpanderStyle style)


matthiasblaesing (changes requested):
The `dlight.nativeexecution` module exports the `api.util` package. While it is only a friend-only api, the list of potential consumers is long. The two methods are only convenience helpers, so what problem is solved by removing them?

matthiasblaesing:
No update in > 6 months - assuming dead. Feel free to reopen.

PR #2144, opened 2020-05-17
[NETBEANS-3733] Suggest surefire 2.22.0 for JUnit5 tests.

sdedic:
Small improvement to 073c42c1cd: the IDE properly tests for JUnit5 in `runSingleMethodEnabled`, so it triggers a suggest/status line message if the project is not properly set up. But the display message and POM fix installs JUnit4+older surefire, which does not work in JUnit5 environment.

I've just improved the messages for JUnit5 case + used proper Surefire version.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/ActionProviderImpl.java
+++ b/java/maven/src/org/netbeans/modules/maven/ActionProviderImpl.java
@@ -211,6 +211,8 @@ public class ActionProviderImpl implements ActionProvider {
     
     //TODO these effectively need updating once in a while
     private static final String SUREFIRE_VERSION_SAFE = "2.15"; //2.16 is broken
+    // surefire 2.22 is needed for JUnit 5
+    private static final String SUREFIRE_VERSION_SAFE_5 = "2.22.0";
     private static final String JUNIT_VERSION_SAFE = "4.11";
 
     @Override public void invokeAction(final String action, final Lookup lookup) {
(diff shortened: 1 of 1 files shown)


java/maven/src/org/netbeans/modules/maven/ActionProviderImpl.java
```diff
@@ -304,9 +309,24 @@ private boolean checkSurefire(final String action) {
                         RequestProcessor.getDefault().post(new Runnable() {
                             @Override
                             public void run() {
+                                String surefireVersion = null;
+                                String junitVersion = null;
+                                
+                                if (ju5 && !usingSurefire2_22()) {
+                                    surefireVersion = SUREFIRE_VERSION_SAFE_5;
+                                } else if (usingSurefire28()) {
```
jlahoda, line 317: Not sure about the logic here - should it be:
```suggestion
                                } else if (!usingSurefire28()) {
```

sdedic: Typo - thanks !

sdedic: Works much better ... in 78778ee

mcdonnell-john:
Probably too late for 12.0, so lets put a milestone of 12.1 on it so its not missed later on

neilcsmith-net:
@mcdonnell-john not sure, it's certainly an annoyance.  There's a chance we're going to need one more beta, so putting 12.0 on it and leaving it to @ebarboni to make the call.

neilcsmith-net:
OTOH, @sdedic seems to have gone for it anyway!

sdedic:
OOPS! Feel free to rollback :-/

neilcsmith-net:
@sdedic :laughing: well Eric really will need to review it now!

PR #2145, opened 2020-05-17
Fix macOS QuickSearch toolbar position

hectorespert:
When Netbeans projects is running in macOS using `ant tryme` command, this `WARN` message is logged: `WARNING [org.openide.filesystems.Ordering]: Found same position 700 for both Toolbars/Git and Toolbars/QuickSearch`
Git Toolbar has the same position than QuickSearch toolbar.


ebarboni:
thks for this update

PR #2148, opened 2020-05-20
WIP: Use manifest attribute to automatically install Truffle language NetBeans modules.

JaroslavTulach:
@daemontus works on a dedicated _Heap Truffle Language_ (similar to [heapdump](https://github.com/JaroslavTulach/heapdump) sketch) to speed up execution of _VisualVM OQL_. In order to make such approach common, we need to make sure that one can integrate such language as a **NetBeans** module and use it when running on _regular JVM_ (in a slower mode) as well as when running on _GraalVM_ (at full speed).

_GraalVM_ offers `-Dtruffle.class.path.append` property and this PR plans to base the solution on top of that property when in GraalVM mode. It collects all modules with `Truffle-Class-Path-Append` attribute on start and composes the system property (usually) sooner than first script is executed.

CCing @sdedic, @thurka, @tzezula, @chumer


neilcsmith-net:
@JaroslavTulach this still WIP or still intended for 12.1?

JaroslavTulach:
I am afraid this has some undesirable Truffle compatibility implications. Let's delay progressing on this for now.

PR #2149, opened 2020-05-20
[NETBEANS-4337] Code Sniffer standard is optional

KacerCZ:
Code Sniffer can be run without parameter '--standard'. In that case it is
looking for configuration file in project root.

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java
@@ -73,7 +73,7 @@ public final class AnalysisOptionsValidator {
 
     @NbBundle.Messages("AnalysisOptionsValidator.codeSniffer.standard.empty=Valid code sniffer standard must be set.")
     public AnalysisOptionsValidator validateCodeSnifferStandard(String codeSnifferStandard) {
-        if (!StringUtils.hasText(codeSnifferStandard)) {
+        if (codeSnifferStandard == null) {
             result.addWarning(new ValidationResult.Message("codeSniffer.standard", Bundle.AnalysisOptionsValidator_codeSniffer_standard_empty())); // NOI18N
         }
         return this;
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodeSniffer.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodeSniffer.java
@@ -137,7 +137,7 @@ public final class CodeSniffer {
         assert file.isValid() : "Invalid file given: " + file;
         try {
             Integer result = getExecutable(Bundle.CodeSniffer_analyze(analyzeGroupCounter++), findWorkDir(file))
-                    .additionalParameters(getParameters(ensureStandard(standard), file, noRecursion))
+                    .additionalParameters(getParameters(standard, file, noRecursion))
                     .runAndWait(getDescriptor(), "Running code sniffer..."); // NOI18N
             if (result == null) {
                 return null;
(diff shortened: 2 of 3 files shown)


php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java
```diff
@@ -73,7 +73,7 @@ public AnalysisOptionsValidator validateCodeSnifferPath(String codeSnifferPath)
 
     @NbBundle.Messages("AnalysisOptionsValidator.codeSniffer.standard.empty=Valid code sniffer standard must be set.")
     public AnalysisOptionsValidator validateCodeSnifferStandard(String codeSnifferStandard) {
```
KacerCZ, line 75: I'm not sure about keeping or removing this method.

KacerCZ:
@junichi11 @tmysik Please review.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java
tmysik: You cannot remove it since it is still used in the file below. Try to remove it and compile the module.

tmysik:
@KacerCZ 

One note - please, always add a direct link to NetBeans issue ([this one](https://issues.apache.org/jira/browse/NETBEANS-4337) in this case), just for easier navigation, reviews, etc. Thank you.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java
KacerCZ: Yes, now its used at more places.
Pesonally I would removed this method and all its usages, because I don't see any value testing if standard is not null as in other places null value is replaced by empty string.

KacerCZ: Yes, now its used at more places.
Personally I would removed this method and all its usages, because I don't see any value testing if standard is not null. There are places where null is replaced by empty string.

tmysik: If it worked, I would not be against such a change, of course :)

tmysik: Maybe @junichi11 could also help here.

KacerCZ: Here: https://github.com/apache/netbeans/blob/ac4e72f120fc01463c525f357d25c43c81e34a6f/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptions.java#L168

junichi11: Well, maybe we can't remove this. It can be `null`.

See 
https://github.com/apache/netbeans/blob/2e7a2e98a2908dbed3897565621073ef263b286f/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptionsValidator.java#L38-L42

https://github.com/apache/netbeans/blob/ac4e72f120fc01463c525f357d25c43c81e34a6f/php/php.code.analysis/src/org/netbeans/modules/php/analysis/ui/options/CodeSnifferOptionsPanel.java#L187

https://github.com/apache/netbeans/blob/ac4e72f120fc01463c525f357d25c43c81e34a6f/php/php.code.analysis/src/org/netbeans/modules/php/analysis/ui/options/CodeSnifferOptionsPanel.java#L121-L127

KacerCZ: OK, I'll leave the method as it is.

PR #2151, opened 2020-05-23
[NETBEANS-4356] Code Completion must check also superclass/superinter…

jlahoda:
…face methods when determining if cast is required.

Consider code like:
```package test;
public class Test {
    public interface Base {
        public void test() { }
    }
    public interface Int extends Base {
    }
    public static class Impl implements Int {
         public void test() { }
    }
    private t(Int b) {
        if (b instanceof Impl) {
            b.tes",
```

And invoke code completion, and choose "test()". The code completion then needs to determine if a cast of "b" to "Impl" is needed of not. So it will look into the type of "b", which is "Int", to see if "test()" is declared there. But it is not, and so the cast will be generated, despite the method being declared in a supertype.

I believe it should look not only at methods declared directly in "Int", but on all members of "Int", even those that are inherited. Which is what this patch tries to do.


arvindaprameya:
+1 Approved, Thanks for addressing this .

lkishalmi:
Approved and looks fine, let's merge it then!

PR #2152, opened 2020-05-23
[NETBEANS-3898] skip hidden files in ModuleList.doScanNetBeansOrgSources

errael:
This avoids descending into .git/.hg and other hidden directories.


errael:
@sdedic Reviewer?

ebresie:
Would it be worth considering usage of the .gitignore?
Would allow more than just "." folders be worth adding?

errael:
> considering usage of the .gitignore

Certainly worth considering, but I wouldn't as part of this issue/PR. This PR is a simple tweak of the current code. Referencing/using .gitignore is an algorithmic change, in particular

- The current structure of `...doScan...` only looks at names on the current level. To use .gitignore requires multi-directory-level consideration.
- Reading a file to change scan behavior.
- embedding knowledge of internals of a VCS file.

These issues require extended thought and consideration.

PR #2154, opened 2020-05-23
[NETBEANS-4365] Fix NetBeans New Action Wizard.

lkishalmi:
The new freemarker seems to be do a serialize/unserialize phase when processing parameters map. The original implementation put a non serializable inner class wrapped into a proxy eventually caused that a collection contained a proxy class which referenced a null object.

This solution is simply replace that kind of "smart" way to pass Java annotation value object around by creating an actual serializable model class instead.

### DIFF
--- a/apisupport/apisupport.wizards/src/org/netbeans/modules/apisupport/project/ui/wizard/action/DataModel.java
+++ b/apisupport/apisupport.wizards/src/org/netbeans/modules/apisupport/project/ui/wizard/action/DataModel.java
@@ -21,9 +21,7 @@ package org.netbeans.modules.apisupport.project.ui.wizard.action;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.Method;
-import java.lang.reflect.Proxy;
+import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


lkishalmi:
I've checked the NetBeans API Wizard sources for similar "smart" solution and found no other occurances.

neilcsmith-net:
Not sure this is a reason to hold up 12.0 release, but perhaps we need to bring back / introduce a `Cherry Pick` / `Backport` label for critical fixes that we want to include in a 12.0 update once they've reached master?

Also maybe @JaroslavTulach might want to review this?

geertjanw:
Sounds silly, maybe, but via this problem we'll be able to see who's actually using the NetBeans Platform, from scratch -- i.e., we know what the problem is here and can help anyone immediately as soon as they encounter this, we'll list it in the release notes, etc, though by not fixing it (unless we need another Beta for other reasons), this could be a way to see what kind of new development is done with the NetBeans Platform, since a newbie is likely to encounter this within the 1st hours of using the NetBeans Platform...

lkishalmi:
Well, I removed the 12.0 milestone. This would be a candidate for 12.0.1 patch release if we would do such things.

PR #1659, opened 2019-11-24
[NETBEANS-3257]: Fixed issue on refactoring rename of java method - c…

jlahoda:
…herrypick of 1112b5038827dcb62fc48d5c52916ff45a5ac3ef

PR #2156, opened 2020-05-25
[NETBEANS-4370] Fix Yaml Lexer in-mustache EOF fails

lkishalmi:
I think this one would be a candidate for 12.0.1 patch release, if we would have that.

### DIFF
--- a/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlLexer.java
+++ b/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlLexer.java
@@ -468,6 +468,12 @@ public final class YamlLexer implements Lexer<YamlTokenId> {
             case ISA_CURLY:
                 state = ISI_WHITESPACE;
                 return token(YamlTokenId.TEXT);
+            case ISI_MUSTACHE:
+                state = ISI_WHITESPACE;
+                return token(YamlTokenId.MUSTACHE);
+            case ISI_MUSTACHE_QUOTE:
+                state = ISI_WHITESPACE;
+                return token(YamlTokenId.MUSTACHE);
             case ISI_PHP:
                 state = ISI_WHITESPACE;
                 return token(YamlTokenId.PHP);
--- a/ide/languages.yaml/test/unit/data/testfiles/issue_NETBEANS-4370_1.yaml.txt
+++ b/ide/languages.yaml/test/unit/data/testfiles/issue_NETBEANS-4370_1.yaml.txt
@@ -1,0 +1,3 @@
+.t.e.s.t. SimpleMustache
+k: {{a
+.e.o.f.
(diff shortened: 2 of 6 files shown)


Chris2011:
Since 9.0 we don't have any patch releases, unfortunately. Hope we can have patch releases again. We already talked about it via mailinglist, but this will be a topic for the mailinglist again.

neilcsmith-net:
@Chris2011 not true! We had a patch release for 11.2. And we'll almost certainly have one for 12.0. There has also been changes since 11.2 to make it easier to achieve.

Chris2011:
@neilcsmith-net thx maybe I missed it. I thought it will be 11.2.1 or so to catch this up. Thx for the info :)

neilcsmith-net:
@Chris2011 update was pushed via update centre. https://lists.apache.org/thread.html/6344fac486525e69369bea981e7c800c3e04c3beb2cfc26d0798deb5%40%3Cdev.netbeans.apache.org%3E  Just remembered Laszlo did similar with 11.0 too.

@lkishalmi Process used in 11.2 and discussed on list was PR to master (use of Cherry Pick label), then once merged a cherry picked PR to release branch - release branch specific testing and review.  See eg. #1604 and #1659  Suggest we follow same for 12.0? Up to @ebarboni if he wants to RM a follow up patch release, or someone else could.  Have added new description to Cherry Pick label and will add to this PR.

ebarboni:
this kind of patch after release make all the maven artefacts desync . Would prefer merge for Voting candidate 1

neilcsmith-net:
@ebarboni we agreed after 11.1 to have vc identical to last beta, because we didn't there on first vc and it caused major problem and delay. I know at least one thing that will probably require update - Oracle's nb-javac with Java 14.  We're going to need to be able to push updates to 12.0 in the next year - maybe need to work out a way to handle that on Maven side?

ebarboni:
ok, maven will need full artefact regeneration 800M thats why we have out of sync maven artefacts for the release with patch.

neilcsmith-net:
Yes, I know.  We discussed same with 11.2, and about possible workarounds - something to discuss on dev@.  And reason to maybe batch up and release a patch update as infrequently as possible!  Longer term maybe we need modules to be released under spec version and a BOM for releases?  Shorter term, maybe we can work out a way to just add RELEASE120-1 to changed files?

ebarboni:
Changing modules version is not enough, impacted related dependencies should be done too + cluster eventually. (big mess compared to rebuild all + source )

neilcsmith-net:
That's something we should probably discuss intricacies of on dev@. Be good to find a better way to handle updates in Maven, as Matthias' work on update centres has made it more feasible to ship patch updates to releases.

Agreed, this by itself might not be enough to release patch update unless other things make that necessary.

I'm -0 to merging this for vc - your call.

PR #2158, opened 2020-05-26
Continuous build in Windows

hectorespert:
Use GitHub actions to build and run tests in Windows.

https://lists.apache.org/thread.html/rc3a7ad290b004557c69d8cac7f5bd4825ca3d230b626641a8565164b%40%3Cdev.netbeans.apache.org%3E

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,46 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


JaroslavTulach:
Great work!

geertjanw:
Pretty cool. GitHub actions uses Zulu, at least by default, so speaking as an Azul guy, this is cool. :-)

ebarboni:
do we have limit on workload / usage for githubactions?

hectorespert:
> do we have limit on workload / usage for githubactions?

These are the usage limits: https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits. Seems that there isn't any limitation in build minutes for open projects.

KacerCZ:
@hectorespert Do you plan to run all tests on Windows?
It would be useful because some tests (i.e. php/php.atoum) pass on Linux but fail no Windows.

PR #2160, opened 2020-05-31
[NETBEANS-4339] PHP Mess Detector - added option to use custom rules

KacerCZ:
Add option to select custom rule set file in PHP Mess Detector configuration.
Customizer panel is also updated.

https://issues.apache.org/jira/browse/NETBEANS-4339

Before:
before

After:
after

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorAnalyzerImpl.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorAnalyzerImpl.java
@@ -18,6 +18,7 @@
  */
 package org.netbeans.modules.php.analysis;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -37,11 +38,13 @@ import org.netbeans.modules.php.analysis.ui.analyzer.MessDetectorCustomizerPanel
 import org.netbeans.modules.php.analysis.util.AnalysisUtils;
 import org.netbeans.modules.php.analysis.util.Mappers;
 import org.netbeans.modules.php.api.executable.InvalidPhpExecutableException;
+import org.netbeans.modules.php.api.util.StringUtils;
 import org.netbeans.modules.php.api.validation.ValidationResult;
 import org.netbeans.modules.refactoring.api.Scope;
 import org.netbeans.spi.editor.hints.ErrorDescription;
 import org.netbeans.spi.editor.hints.HintsController;
 import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
 import org.openide.util.NbBundle;
 import org.openide.util.lookup.ServiceProvider;
 
@@ -76,8 +79,8 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
             return Collections.emptyList();
         }
 
-        List<String> messDetectorRuleSets = getValidMessDetectorRuleSets();
-        if (messDetectorRuleSets == null) {
+        MessDetectorParams messDetectorParams = getValidMessDetectorParams();
+        if (messDetectorParams == null) {
             context.reportAnalysisProblem(Bundle.MessDetectorAnalyzerImpl_messDetector_ruleSets_error(), Bundle.MessDetectorAnalyzerImpl_messDetector_ruleSets_error_description());
             return Collections.emptyList();
         }
@@ -92,7 +95,7 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
 
         context.start(totalCount);
         try {
-            return doAnalyze(scope, messDetector, messDetectorRuleSets, fileCount);
+            return doAnalyze(scope, messDetector, messDetectorParams, fileCount);
         } finally {
             context.finish();
         }
@@ -109,7 +112,8 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
         "MessDetectorAnalyzerImpl.analyze.error=Mess detector analysis error",
         "MessDetectorAnalyzerImpl.analyze.error.description=Error occurred during mess detector analysis, review Output window for more information.",
     })
-    private Iterable<? extends ErrorDescription> doAnalyze(Scope scope, MessDetector messDetector, List<String> messDetectorRuleSets, Map<FileObject, Integer> fileCount) {
+    private Iterable<? extends ErrorDescription> doAnalyze(Scope scope, MessDetector messDetector,
+            MessDetectorParams params, Map<FileObject, Integer> fileCount) {
         List<ErrorDescription> errors = new ArrayList<>();
         int progress = 0;
         messDetector.startAnalyzeGroup();
@@ -117,7 +121,7 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
             if (cancelled.get()) {
                 return Collections.emptyList();
             }
-            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(messDetectorRuleSets, root);
+            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(params, root);
             if (results == null) {
                 context.reportAnalysisProblem(Bundle.MessDetectorAnalyzerImpl_analyze_error(), Bundle.MessDetectorAnalyzerImpl_analyze_error_description());
                 return Collections.emptyList();
@@ -131,7 +135,7 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
             if (cancelled.get()) {
                 return Collections.emptyList();
             }
-            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(messDetectorRuleSets, file);
+            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(params, file);
             if (results == null) {
                 context.reportAnalysisProblem(Bundle.MessDetectorAnalyzerImpl_analyze_error(), Bundle.MessDetectorAnalyzerImpl_analyze_error_description());
                 return Collections.emptyList();
@@ -155,7 +159,7 @@ public class MessDetectorAnalyzerImpl implements Analyzer {
             if (dataChildren.isEmpty()) {
                 continue;
             }
-            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(messDetectorRuleSets, dataChildren);
+            List<org.netbeans.modules.php.analysis.results.Result> results = messDetector.analyze(params, dataChildren);
             if (results == null) {
                 context.reportAnalysisProblem(Bundle.MessDetectorAnalyzerImpl_analyze_error(), Bundle.MessDetectorAnalyzerImpl_analyze_error_description());
                 return Collections.emptyList();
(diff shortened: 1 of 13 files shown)


KacerCZ:
@tmysik @junichi11 Please advise/help how to deal with selection of built-in rules (codesize, design, ...).
When rule set file is selected, then built-in rules don't have to be selected. Mess detector will work only with rules in file.

Validation of settings was already modified to handle this case. Problem is with selection of built-in rules. It seems that it is not possible for JList to have nothing selected.

tmysik:
@KacerCZ 

What about just disabling the _Default Rule Sets_ field if _Rule Set File_ is selected?

CC @junichi11

KacerCZ:
@tmysik Default rules can be combined with custom rules in file - see http://phpmd.org/documentation/index.html
My intention is to make Default Rule Sets optional, so using only custom rule set would be possible.

tmysik:
@KacerCZ 

I am not sure I understand. So - what about adding a checkbox that would enable _Default Rule Sets_ field? Would that work? :thinking:

tmysik:
@KacerCZ 

BTW looking at the attached images, I guess that the _Rule Set File_ textfield should be as long as the other 2 text fields above (it means that both the _Browse..._ buttons would be aligned and after the new one, there would be a free space).

What do you think?

CC @junichi11

junichi11:
I would add an empty string or whitespace to the list if that's possible. Or I would change the label to a checkbox as @tmysik wrote. 

> BTW looking at the attached images, I guess that the Rule Set File textfield should be as long as the other 2 text fields above (it means that both the Browse... buttons would be aligned and after the new one, there would be a free space).

Agree.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorAnalyzerImpl.java
```diff
@@ -177,22 +181,49 @@ private MessDetector getValidMessDetector() {
...
+                .setRuleSetFile(getValidRuleSetFile());
+     ValidationResult result = new AnalysisOptionsValidator()
+                .validateMessDetector(messDetectorParams)
+                .getResult();
+        if (result.hasErrors()
+                || result.hasWarnings()) {
+            return null;
+        }
+     return messDetectorParams;
+    }
```
junichi11, line 197: Please format.

KacerCZ: Fixed. Thanks for spotting it.

KacerCZ:
@junichi11 I added whitespace to list of built-in rule sets and made sure it does not goes into command parameters.
@tmysik I fixed alignment of Browse button as you suggested.
Same alignment issue is in configuration for PHPStan, shall I prepare PR to fix it?

Options after changes:
after2

Inspection customization:
after-customizer

KacerCZ:
Please review.

KacerCZ:
@tmysik Yes, I added empty item to list selection of predefined rules. This empty item is ignored when parameters for Mess detector are created.
Valid cases are:
- one or more predefined rule sets and no rule set file
- empty predefined rule set and rule set file
- one or more predefined rule sets and rule set file
Invalid case is when there is no predefined rule selected and also no rule set file is used.

The second screenshot is from inspection customization dialog. 
_Enabled_ checkbox enables running of selected analyzer in given configuration.
1. From main menu select Source > Inspect.
2. In Inspect dialog select Use > Configuration and select Default.
3. Click Manage next to Default. 
4. Select Analyzer > Mess Detector - that is the dialog from screenshot

php/php.code.analysis/src/org/netbeans/modules/php/analysis/ui/options/MessDetectorOptionsPanel.java
```diff
@@ -179,113 +197,137 @@ public ValidationResult getValidationResult() {
...
-        messDetectorBrowseButton = new JButton();
-        messDetectorSearchButton = new JButton();
-        messDetectorHintLabel = new JLabel();
-        messDetectorRuleSetsLabel = new JLabel();
-        messDetectorRuleSetsScrollPane = new JScrollPane();
-        messDetectorRuleSetsList = new JList<String>();
-        noteLabel = new JLabel();
-        minVersionInfoLabel = new JLabel();
-        messDetectorLearnMoreLabel = new JLabel();
+        messDetectorLabel = new javax.swing.JLabel();
```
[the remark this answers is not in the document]
KacerCZ: I changed the setting (had no idea it exists) and fixed the file.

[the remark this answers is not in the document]
tmysik: Great, thank you.

tmysik:
One quick note - maybe instead of an empty item in the list, we could add something like _&lt;ignored>_ or _&lt;unused>_ or similarly. What do you think?

junichi11:
> One quick note - maybe instead of an empty item in the list, we could add something like _<ignored>_ or _<unused>_ or similarly.

Nice idea :) It's easy to understand, I think.

junichi11:
BTW, I added the `Enabled` checkbox before because all analyzers are run when a user customizes a configuration if the checkbox doesn't exist.

KacerCZ:
I replaced empty string with _<none>_ and updated validation.
I added missing listener for rule set file textfield in MessDetectorCustomizerPanel.

tmysik:
@junichi11 

Junichi, please, just to be sure - could you test whether it works? I am sorry, I am really busy now and did not get to try it myself :/

@KacerCZ

Well done!

Thank you.

junichi11:
@tmysik 
Yes, I'll test it later :)

BTW, We should be able to merge PRs again: https://lists.apache.org/thread.html/r619a782b9e828bd62324817b6ae96f93cfab0dd0ea1f67edcfce25c6%40%3Cdev.netbeans.apache.org%3E

tmysik:
@junichi11 

Thanks a lot! :bow:

junichi11:
@KacerCZ 
I can't edit "Rule Set File" of the Configurations dialog.
nb4339-conf-dialog

KacerCZ:
@junichi11 Fixed enabling/disabling of File controls. Thank you.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
```diff
@@ -0,0 +1,46 @@
...
+    private FileObject ruleSetFile;
+
+    public List<String> getRuleSets() {
+        return ruleSets;
+    }
+
+    public MessDetectorParams setRuleSets(List<String> ruleSets) {
+        this.ruleSets = ruleSets;
+        return this;
+    }
```
junichi11, line 36: Should use package private instead of `public`.
Should add the list as a new list. i.e. 
```java
this.ruleSets = new ArrayList<>(ruleSets);
```

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import org.openide.filesystems.FileObject;
+
+public final class MessDetectorParams {
+
+    private List<String> ruleSets;
+    private FileObject ruleSetFile;
+
+    public List<String> getRuleSets() {
+        return ruleSets;
```
junichi11, line 30: Should use `Collections.unmodifiableList()`.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
```diff
@@ -0,0 +1,46 @@
...
+    public MessDetectorParams setRuleSets(List<String> ruleSets) {
+        this.ruleSets = ruleSets;
+        return this;
+    }
+
+    public FileObject getRuleSetFile() {
+        return ruleSetFile;
+    }
+
+    public MessDetectorParams setRuleSetFile(FileObject ruleSetFile) {
```
junichi11, line 42: Should use package private instead of `public`.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/MessDetector.java
```diff
@@ -59,10 +60,12 @@
     private static final String REPORT_FORMAT_PARAM = "xml"; // NOI18N
     private static final String EXCLUDE_PARAM = "--exclude"; // NOI18N
     private static final String SUFFIXES_PARAM = "--suffixes"; // NOI18N
+    public static final String EMPTY_RULE_SET = "<none>"; // NOI18N
```
junichi11, line 63: I would consider `i18N` because it is not a command or parameter.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/ValidatorMessDetectorParameter.java
```diff
@@ -0,0 +1,69 @@
...
+        ruleSets = panel.getSelectedRuleSets();
+        ruleSetFilePath = panel.getRuleSetFile();
+    }
+
+    public String getMessDetectorPath() {
+        return messDetectorPath;
+    }
+
+    public List<String> getRuleSets() {
+        return ruleSets;
```
junichi11, line 62: Should use `Collections.unmodifiableList()`.

junichi11:
@KacerCZ 
BTW, You should submit icla because we would like to avoid troubles for the license.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/MessDetector.java
KacerCZ: It's kind of special rule - ignored during command parameters building, but still written into settings.
Making it translated would also mean to map value saved into settings to translated value shown in list, right?

tmysik: I agree with Junichi, this should be translated. However, the stored value should be just some identificator (so, not the translated string).

KacerCZ: Can you point me to a place in NetBeans where something like this is done or what class/interface to look for, please.

tmysik: Not sure what you mean... Do you mean a custom Renderer for the given JList so this specific option can be translated?

KacerCZ: Yes, that are the keywords I was looking for ;)
I'm novice in Java/NetBeans realm and I try to learn it by observing how these things are done in other parts of the project.

tmysik: This is more about Java and Swing than NetBeans. See [the official UI tutorial](https://docs.oracle.com/javase/tutorial/uiswing/components/list.html). This tutorial is fine IMHO and I am pretty sure that you will be able to find a similar case already used in NetBeans ;)

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
KacerCZ: Changed as suggested.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
KacerCZ: Fixed.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/MessDetectorParams.java
KacerCZ: Fixed.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/MessDetector.java
KacerCZ: I created renderer based on `PHPStanLevelListCellRenderer`.
Value is now only dash '-' and '<none>' is shown in list.
Before testing again you should clean userdir.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/ValidatorMessDetectorParameter.java
KacerCZ: Fixed.

KacerCZ:
@junichi11 I sent signed ICLA to ASF.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/ui/MessDetectorRuleSetsListCellRenderer.java
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.php.analysis.ui;
+
+import java.awt.Component;
+import javax.swing.JList;
+import javax.swing.ListCellRenderer;
+import org.netbeans.modules.php.analysis.commands.MessDetector;
+import org.netbeans.modules.php.analysis.options.AnalysisOptions;
```
junichi11, line 25: Please remove the unused import.

junichi11:
> I sent signed ICLA to ASF.

Thanks!

Let's merge it after the unused import is fixed.

php/php.code.analysis/src/org/netbeans/modules/php/analysis/ui/MessDetectorRuleSetsListCellRenderer.java
KacerCZ: Removed. Sorry I didn't spot it.

junichi11:
@tmysik Could you please merge this if it's OK with you?

tmysik:
@KacerCZ 

Well done! Thanks!

@junichi11 

Thank you too!

KacerCZ:
@tmysik @junichi11 Thank you for your guidance and patience.

PR #2161, opened 2020-06-01
Show detailed information about commits in PRs

matthiasblaesing:
Information shown:

- the public information about the PR requestor
- the author information of each commit as recorded in the commit data
- for each found author the information whether an ICLA could be found
- hints about which special concerns should be applied when considering merging

This is an alternative to: #2125
Closes: #2125

### DIFF
--- a/.github/workflows/check-commits.yml
+++ b/.github/workflows/check-commits.yml
@@ -1,0 +1,29 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
Ok - this won't fly - Actions on PRs can't create comments on themselves. It is even documentated:

https://github.community/t/github-actions-are-severely-limited-on-prs/18179

and ugly work around created:

https://github.com/nyurik/auto_pr_comments_from_forks

Maybe I'll rebuild this as a github app and create an authorized token for this *arg*

hectorespert:
> and ugly work around created:
> 
> https://github.com/nyurik/auto_pr_comments_from_forks

It is an ugly work around, but seems simpler to implement than a GitHub app. On the other hand, a GitHub app is the ideal option.

hectorespert:
@matthiasblaesing I think I have found an alternative: https://github.com/apache/netbeans-jenkins-lib/pull/21

PR #2162, opened 2020-06-03
Warn about custom modifications in build-impl.xml

jlahoda:
The intent is to mitigate the Octopus Scanner Malware at least a little. Will warn the user about modification to build-impl.xml while opening the project.

An alternative would be to re-generate the build-impl.xml before each build action, but the code is trying hard to preserve any user modifications, so this feels dangerous.


geertjanw:
Great!

arvindaprameya:
+1 Approved, Thanks for doing this.

ide/project.ant.ui/manifest.mf
```diff
@@ -1,5 +1,5 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.project.ant.ui/1
-OpenIDE-Module-Specification-Version: 1.73
+OpenIDE-Module-Implementation-Version: 1
```
JaroslavTulach, line 3: It is unfortunate to use implementation dependency.

neilcsmith-net:
What's status on this one?  Merging, fixing, closing?

lkishalmi:
@jlahoda Would you update this one, or close it if this is no longer relevant.

lkishalmi:
@jlahoda Would you update this one, or close it if this is no longer relevant.

PR #2163, opened 2020-06-03
[NETBEANS-4396] Properly disabling Compile on Save when it is not sup…

jlahoda:
…ported.


arvindaprameya:
+1 Approved, Thanks for making this change.

PR #2167, opened 2020-06-05
[NETBEANS-4066] Fix overshooting text in editor tabs on HiDPI displays

eirikbakke:
This PR fixes the problem shown in the editor tab in the screeenshot; on some HiDPI scaling factors (150% in this case), the text in the tab may be incorrectly measured. The bug is fixed by making sure that the right FontRenderContext, with HiDPI scaling taken into account, is used when painting on a particular monitor.

Inaccurate tab text mesaurement

See also https://github.com/apache/netbeans/pull/2025 , which fixed a similar problem in the org.openide.awt.HtmlRenderer class.


lkishalmi:
No conflict, no review. Merging it in not to have it rotten here.

eirikbakke:
Thanks! I read through the patch again now, a month after working on it last, and didn't discover any new issues.

PR #2170, opened 2020-06-08
Do not blindly cast to IndexedClassDecl while reassigning anonymous c…

jlahoda:
…lass numbers, as the class decl may not be IndexedClassDecl.

While running on current JDK 15 without nb-javac, the parser produces ordinary JCClassDecl, not IndexedClassDecl, and then the casts to IndexedClassDecl fail, and various tests (like code completion tests) fail due to this.

Eventually, we probably want to get rid of IndexedClassDecls altogether, but more testing will be needed for that.

Added also a related test to partial reparse test, to verify things work reasonably during partial reparse.


arvindaprameya:
+1 Approved, Thanks for doing this Jan.

PR #2171, opened 2020-06-08
Moved Gradle modules out of Groovy cluster

lkishalmi:
This change is about to move Gradle out of Groovy's shade. My motivation are the following:

- Groovy is not really required by Gradle. Having it next to Maven and Ant, would be more suitable.
- C/C++ Support is coming soon. I'd like to add Gradle Native projects to that support.

This change would move Gradle modules out from the Groovy Cluster to:

- ide: libs.gradle (Gradle Tooling library), gradle (Gradle Project support)
- java: gradle,java, gradle.java.coverage, gradle kit, gradle.persistence, gradle.spring and gradle.test (These are the modules that supporting the Java development)
- enterprise: gradle.javaee (Gradle Web Project support)

Please share your wisdom! I might need to recreate this PR as probably some other changes would be better to be included in the repo, so this one is just to have a glimpse.


JaroslavTulach:
Enabling Gradle without enabling support for Groovy in the editor doesn't make much sense to me. While I know that Kotlin guys are trying hard to replace Groovy, 95% project build scripts is still in Groovy, I'd bet.

C/C++ support is likely to end up in a `cnd` cluster. That cluster never had a dependency on `java` cluster. The `cnd` functionality is using Ant (for historical reasons, I'd say) - however to do that the basic Ant support had to be moved to an artificially created `extide` cluster. The logic was that Ant is too Java oriented to belong into generic `ide` cluster (mostly editor, version control, etc.).

Gradle is certainly also "too Java oriented". Maybe it belongs next to Ant into `extide`?

lkishalmi:
Well, unfortunately the Groovy editor do not offer too much help when editing Gradle files. Also it is not an everyday task for most of the users to edit the build files, mostly just to change/add dependencies and that can be done without Groovy. If someone would enable the Groovy cluster, then it would have more editing options, but that's it.

Thank you for mentioning the extide cluster, yes Gradle could belong there. I just was not aware of the existence of that cluster.

jlahoda:
I'd vote for extide cluster over ide cluster for the base Gradle support - AFAIK we don't have build systems in the ide cluster. C/C++ support should be fine with depending on the extide cluster.

lkishalmi:
Works as I've imagined. Made Travis happy too.

PR #2172, opened 2020-06-08
[NETBEANS-3311] Convert TextBlock to String, Test Case Fix

Akshay-Gupta-Oracle:
Convert TextBlock to String, Function call parameter fix test case fix


arvindaprameya:
+1 Approved, Thank you

PR #2176, opened 2020-06-09
[NETBEANS-4341] Fix unloadable pom regression

oyarzun:
Maven v3.6 introduced more aggressive caching which causes child poms to bu unload-able when a dependency is added to the parents dependency management.

This PR disables the global cache with: 
`-J-Dmaven.defaultProjectBuilder.disableGlobalModelCache=true` 

https://marc.info/?l=maven-dev&m=154695759532149&w=2

https://youtrack.jetbrains.com/issue/IDEA-216851


matthiasblaesing:
This is a great find! However, I don't like the approach to set this from command line. For example if maven support is used in a platform application with a custom launcher it will be broken. So what do you think about this variant:

```java
package org.netbeans.modules.maven;

/**
 * Disable the maven global cache. With the cache enabled updates to the pom
 * or the dependencies are not picked up correctly.
 *
 * @see <a href="https://issues.apache.org/jira/browse/NETBEANS-4341">NETBEANS-4341</a>
 * @see <a href="https://issues.apache.org/jira/browse/MNG-6530">MNG-6530</a>
 */
public class MavenCacheDisabler implements Runnable {

    @Override
    public void run() {
        System.setProperty("maven.defaultProjectBuilder.disableGlobalModelCache", "true");
    }

}
```

I'm waiting for my build to finish, but it should work, as the maven property is reevaluated at each run.

neilcsmith-net:
@matthiasblaesing I like that approach even less. Can you at least check if it's been explicitly set first?

matthiasblaesing:
@neilcsmith-net why? We know, that the maven embedder will not work correctly if the system property is not set. My understanding is, that setting the property has performance implications, but is necessary if used if you are not faced with a static pom.

neilcsmith-net:
@matthiasblaesing I'm aware of that, but hardcoding system properties is not good _in principle_ for end users and/or (platform) developers.  Let's assume with any system property, if it's already set someone has a reason.

matthiasblaesing:
Ok makes sense, so this would be what you mean:

```java
package org.netbeans.modules.maven;

import org.openide.modules.OnStart;

/**
 * Disable the maven global cache. With the cache enabled updates to the pom
 * or the dependencies are not picked up correctly.
 *
 * @see <a href="https://issues.apache.org/jira/browse/NETBEANS-4341">NETBEANS-4341</a>
 * @see <a href="https://issues.apache.org/jira/browse/MNG-6530">MNG-6530</a>
 */
@OnStart
public class MavenCacheDisabler implements Runnable {
    @Override
    public void run() {
        if(System.getProperty("maven.defaultProjectBuilder.disableGlobalModelCache") == null) {
            System.setProperty("maven.defaultProjectBuilder.disableGlobalModelCache", "true");
        }
    }
}
```

neilcsmith-net:
Better, thanks! :+1:  Although aren't system properties writeable?

```
System.getProperties().putIfAbsent("maven.defaultProjectBuilder.disableGlobalModelCache", "true");
```

matthiasblaesing:
Both versions (the get + check for null variant and the putIfAbsent variant). I have no strong feelings either way. It might make a difference if a security manager is active. In that case the get+check variant might trigger less problems. I think that is a theoretical case.

oyarzun:
@matthiasblaesing sounds good to me. After you've tested your build are you going to submit a new PR?

PR #2184, opened 2020-06-12
[NETBEANS-4434] Code Completion doesn't work when script type attribute value is "module"

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4434

#### Before

nb4434-before

#### After

nb4434-after


junichi11:
@sdedic Could you please review it if possible?

junichi11:
@sdedic Thank you for your review.

I declared it as a string constant in `JsEmbeddingProvider` but I didn't declare it in `AttrValuesCompletion` since the other items were not declared as constant. Maybe, we should create a new class for constants (e.g. `HtmlConstants`).

diff: https://github.com/apache/netbeans/compare/38de25fcb8645aad44c076af9939907884b65c92..8b507cbb5b1d61638cfa1422bb51d8f313546d03

PR #2185, opened 2020-06-13
JUnit Ignore annotation support

hectorespert:
Add support for JUnit [@Ignore annotation](http://junit.sourceforge.net/javadoc/org/junit/Ignore.html)  on [NbTestCase](https://bits.netbeans.org/dev/javadoc/org-netbeans-modules-nbjunit/org/netbeans/junit/NbTestCase.html) test case extension.

```java
    @Ignore("Ignored test case")
    public static class IgnoredTestCase extends NbTestCase {

        public IgnoredTestCase(String n) {
            super(n);
        }

        public void testFailed() {
            fail();
        }

    }
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,12 +97,11 @@ matrix:
         - name: Test harness modules
           jdk: openjdk8
           env:
-            - OPTS="-silent -Dcluster.config=platform -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false -Dtest-unit-sys-prop.ignore.random.failures=true -Dvanilla.javac.exists=true"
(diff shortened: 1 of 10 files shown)


JaroslavTulach:
There already is support for `@RandomlyFails` on individual methods. What's the purpose of supporting `@Ignore`?

hectorespert:
> There already is support for `@RandomlyFails` on individual methods. What's the purpose of supporting `@Ignore`?

On the one hand, JUnit annotations support in NbUnit tests could be a good feature for these.

On the other hand, I would like to start to use this annotation to mark the tests that doesn't work to disable these, and use `@RandomlyFails` annotation only for the test that fails randomly.

JaroslavTulach:
Thanks for your contribution, Hector.

PR #2187, opened 2020-06-13
Test Netbeans build system (Ant Tasks)

hectorespert:
Use a GitHub action to run Netbeans Build System (Ant Tasks) tests.


hectorespert:
> One thing that might be worth checking is, if the build call is necessary or of localtest builds what it needs.

`localtest` only builds the bootstrap but some tests have other modules as a dependency

PR #2191, opened 2020-06-15
[NETBEANS-4410] Fixed NPE on Certain Gradle Test Executions.

lkishalmi:
It seems Gradle post output events for tests SUITE-s before it would open a TestCase.
This one is a trivial fix, potential candidate fro 12.0.1 patch release.


lkishalmi:
Just an image after this fix:

image

PR #2194, opened 2020-06-16
[NETBEANS-4458] - enable use of generics

BradWalker:
A lot of places in the code has generic use commented out.. For example, the following in
LayoutModel.java

     public void unsetSameSize(List/*<String>*/ components, int dimension) {
        Iterator i = components.iterator();
        while (i.hasNext()) {
            String cid = (String)i.next();
            LayoutComponent lc = getLayoutComponent(cid);
            removeComponentFromLinkSizedGroup(lc, dimension);
        }
    }

All I've done here is enable it to reduce the warnings and test out the code to make sure it works correctly.

### DIFF
--- a/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/TemplateClientPanelVisual.java
+++ b/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/TemplateClientPanelVisual.java
@@ -75,7 +75,7 @@ public class TemplateClientPanelVisual extends javax.swing.JPanel implements Hel
 
     private WizardDescriptor wizardDescriptor;
 
-    private final Set/*<ChangeListener>*/ listeners = new HashSet(1);
+    private final Set<ChangeListener> listeners = new HashSet<>(1);
 
     private final static String VALUE_NAME = "name";    //NOI18N
 
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
@@ -59,7 +59,7 @@ class ClassDumpSegment extends TagBounds {
     final int superClassIDOffset;
     ClassDump java_lang_Class;
     boolean newSize;
-    Map /*<JavaClass,List<Field>>*/ fieldsCache;
+    Map<JavaClass,List<Field>> fieldsCache;
     private List<JavaClass> classes;
     private Map /*<Byte,JavaClass>*/ primitiveArrayMap;
 
@@ -202,12 +202,12 @@ class ClassDumpSegment extends TagBounds {
         }
     }
 
-    synchronized List /*<JavaClass>*/ createClassCollection() {
+    synchronized List<JavaClass> createClassCollection() {
         if (classes != null) {
             return classes;
         }
 
-        List cls = new ArrayList /*<JavaClass>*/(1000);
+        List<JavaClass> cls = new ArrayList<>(1000);
 
         long[] offset = new long[] { startOffset };
 
--- a/enterprise/j2ee.earproject/src/org/netbeans/modules/j2ee/earproject/ui/customizer/EarProjectProperties.java
+++ b/enterprise/j2ee.earproject/src/org/netbeans/modules/j2ee/earproject/ui/customizer/EarProjectProperties.java
@@ -844,7 +844,7 @@ public final class EarProjectProperties {
             return;
         }
         
-        for( Iterator/*<Project>*/ it = spp.getSubprojects().iterator(); it.hasNext(); ) {
+        for( Iterator<? extends Project> it = spp.getSubprojects().iterator(); it.hasNext(); ) {
             Project sp = (Project) it.next();
             if (ProjectUtils.hasSubprojectCycles(project, sp)) {
                 Logger.getLogger("global").log(Level.WARNING, "There would be cyclic " + // NOI18N
--- a/enterprise/j2ee.genericserver/nbproject/project.properties
+++ b/enterprise/j2ee.genericserver/nbproject/project.properties
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 javac.compilerargs=-Xlint:unchecked
-javac.source=1.6
+javac.source=1.8
--- a/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/ManagedBeanPanel.java
+++ b/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/ManagedBeanPanel.java
@@ -123,7 +123,8 @@ final class ManagedBeanPanel implements WizardDescriptor.Panel, WizardDescriptor
         }
         return component.isAddBeanToConfig();
     }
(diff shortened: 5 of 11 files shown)


platform/openide.nodes/src/org/openide/cookies/InstanceCookie.java
```diff
@@ -31,7 +31,7 @@
  *
  * @author Jaroslav Tulach
  */
-public interface InstanceCookie/*<T>*/ extends Node.Cookie {
+public interface InstanceCookie<T> extends Node.Cookie {
```
JaroslavTulach, line 34: This is an API change.

JaroslavTulach:
Many of the changes aren't API changes (like changing local variables and private or package private field/method types). Those can certainly be made, the other ones would deserve some more thoughts. We have just a single try to get the signatures right!

However let me state that it makes no sense to reduce number of warnings without preventing the  warnings to ever appear again. E.g. for each module that you want to clean up, turn on "warning as error" mode and then fix (or suppress) all the warnings in the module. Otherwise it is just a useless mangling of the source code. In my 2 Kč opinion.

platform/openide.nodes/src/org/openide/cookies/InstanceCookie.java
BradWalker: Reverted back to original.

ide/xml.text/src/org/netbeans/modules/xml/text/navigator/base/AbstractXMLNavigatorPanel.java
```diff
@@ -92,7 +92,7 @@ public void panelDeactivated() {
      * 
      * @param selectedFiles 
      */
-    public void navigate(Collection/*<DataObject>*/ selectedFiles) {
+    public void navigate(Collection<DataObject> selectedFiles) {
```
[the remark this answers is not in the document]
BradWalker: Reverted back to original.

ide/xml.core/src/org/netbeans/modules/xml/api/model/ExtendedGrammarQuery.java
```diff
@@ -32,6 +32,6 @@
 public interface ExtendedGrammarQuery extends GrammarQuery {
     
     /** @return a List of resolved entities System id-s names.*/
-    public List/*<String>*/ getResolvedEntities();
+    public List<String> getResolvedEntities();
```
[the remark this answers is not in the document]
BradWalker: Reverted back to original. We'll figure this out in the future..

platform/openide.nodes/src/org/openide/cookies/InstanceCookie.java
BradWalker: I've reverted this back to the original state..

BradWalker:
> Many of the changes aren't API changes (like changing local variables and private or package private field/method types). Those can certainly be made, the other ones would deserve some more thoughts. We have just a single try to get the signatures right!
> 
> However let me state that it makes no sense to reduce number of warnings without preventing the warnings to ever appear again. E.g. for each module that you want to clean up, turn on "warning as error" mode and then fix (or suppress) all the warnings in the module. Otherwise it is just a useless mangling of the source code. In my 2 Kč opinion.

I'm in agreement with you about this.. But, there are 10s of thousands of warnings.. I'm trying to fix things in a way that only "one thing" gets changed at a time.. That's why I'll do things like just fix the EMPTY_LIST..

I've managed to get our number of warnings down from approx. 26,000 to 19,000.. Once things get more manageable then we can start to turn on "error on warning"..

Hopefully this explains my rational better..

BradWalker:
One more thing.. I do try to stay away from API changes because of the previously mentioned comments.  FYI..

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
```diff
@@ -202,12 +202,12 @@ void addInstanceSize(ClassDump cls, int tag, long instanceOffset) {
         }
     }
 
-    synchronized List /*<JavaClass>*/ createClassCollection() {
+    synchronized List/*<JavaClass>*/ createClassCollection() {
```
matthiasblaesing, line 205: This is package private - I would not consider this API, so could be made generic and then the callsite of this function could be generified.

profiler/lib.profiler/src/org/netbeans/lib/profiler/results/BaseCallGraphBuilder.java
```diff
@@ -42,7 +42,7 @@
 
     //~ Instance fields ----------------------------------------------------------------------------------------------------------
 
-    protected List/*<Runnable>*/ afterBatchCommands = new ArrayList<>();
+    protected List<Runnable> afterBatchCommands = new ArrayList<>();
```
matthiasblaesing, line 45: This is a friend API.

BradWalker: > This is a friend API.

fixed..

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
BradWalker: > This is package private - I would not consider this API, so could be made generic and then the callsite of this function could be generified.

Fixed..

BradWalker:
> -1 as of now. This is not a compatible change. Needs proper review / plan, particularly for anything that is public API. @matthiasblaesing made this point already in #1377

Should be fixed..

enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/TemplateClientPanelVisual.java
```diff
@@ -75,7 +75,7 @@
 
     private WizardDescriptor wizardDescriptor;
 
-    private final Set/*<ChangeListener>*/ listeners = new HashSet(1);
```
jtulach, line 78: `HashSet` of size 1 is probably a non-sense. By default it has `75%` fill ratio and then it is rehashed.E.g. as soon as first item is added, the set is probably rehashed.

jtulach:
The current state is integratable, as far as I can say.

matthiasblaesing (approved):
Looks sane - lets get this in :-)

enterprise/web.jsf/src/org/netbeans/modules/web/jsf/wizards/TemplateClientPanelVisual.java
BradWalker: > `HashSet` of size 1 is probably a non-sense. 

Just a follow-up comment in case someone comes across this bug report in the future.. I completely agree with your comment! My goal is simply to try and stay away from any "potential" source changes that could be behavioral. Even ones as small as this. Hope that explains why I do what I do..

BradWalker:
Hey @junichi11 , how can I add the "code cleanup" label? I don't seem to have permission..

junichi11:
@BradWalker Well, only committer... So, if you would like to add a label, please ask someone. I add it when I notice it is missing.

PR #2192, opened 2020-06-15
Netbeans project proxy configuration

hectorespert:
Seems that part of the problems related with the Apache infrastructure on Travis builds or GitHub actions are related to the proxy configuration introduced in the Netbeans build system to work behind a proxy. https://github.com/apache/netbeans/pull/1092 

```
2020-06-13T17:01:33.9599330Z /Users/runner/runners/2.263.0/work/netbeans/netbeans/enterprise/web.core/build.xml:25: java.io.IOException: Cannot connect to http://netbeans.apache.org
2020-06-13T17:01:33.9599810Z 	at org.netbeans.nbbuild.extlibs.ConfigureProxy.openConnection(ConfigureProxy.java:145)
2020-06-13T17:01:33.9601430Z 	at org.netbeans.nbbuild.extlibs.ConfigureProxy.execute(ConfigureProxy.java:62)
2020-06-13T17:01:33.9608220Z 	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
2020-06-13T17:01:33.9608370Z 	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
2020-06-13T17:01:33.9608490Z 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2020-06-13T17:01:33.9608660Z 	at java.lang.reflect.Method.invoke(Method.java:498)
2020-06-13T17:01:33.9608780Z 	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:99)
2020-06-13T17:01:33.9608910Z 	at org.apache.tools.ant.Task.perform(Task.java:350)
2020-06-13T17:01:33.9609020Z 	at org.apache.tools.ant.Target.execute(Target.java:449)
2020-06-13T17:01:33.9609140Z 	at org.apache.tools.ant.helper.ProjectHelper2.parse(ProjectHelper2.java:184)
2020-06-13T17:01:33.9609270Z 	at org.apache.tools.ant.ProjectHelper.configureProject(ProjectHelper.java:93)
2020-06-13T17:01:33.9609400Z 	at org.apache.tools.ant.taskdefs.Ant.execute(Ant.java:393)
2020-06-13T17:01:33.9609510Z 	at sun.reflect.GeneratedMethodAccessor114.invoke(Unknown Source)
[...]

PR #2197, opened 2020-06-17
Custom ant task to download release properties

hectorespert:
Add custom ant task to download and cache the release json properties file to prevent send a request to the Apache gitbox every time that an ant command is executed in the root project.

This PR is related to this other to prevent requests in the proxy configuration: https://github.com/apache/netbeans/pull/2192


hectorespert:
Closed in favor of https://github.com/apache/netbeans/pull/2417

PR #2199, opened 2020-06-17
[NETBEANS-4464] Fixed unit tests for Atoum on Windows

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4464
Directory separators had to be replaced with system-dependent value.
Because Windows use `\` as separator, replacement text has to be quoted.


KacerCZ:
@junichi11 @tmysik Please review.

hectorespert:
@KacerCZ Could you try to run these tests into the Windows build? https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml

KacerCZ:
@hectorespert I added Atoum tests to Windows build and it passed.

PR #2200, opened 2020-06-17
[NETBEANS-4465] - update RAT license exceptions

BradWalker:
When running the RAT report after a build has taken place, the report
picks up hundreds of build generated files. These files do not have a license
header and it's really not necessary as they are generated by the build process..


BradWalker:
Hey @matthiasblaesing, thanks for approving the changes.. I agree with your comments about running the RAT report after running the "ant clean" task..

I have 1000s of files that need to have the license updated in the C/C++ contribution.. This is why I've been running the RAT report after doing a build to test to make sure I'm not missing something. So it does help me in that regard..

FYI.

matthiasblaesing:
Ok - lets get this in.

PR #2202, opened 2020-06-19
Added tests for PHP modules in Windows build

KacerCZ:
List of tested modules is based on Travis configuration.
Commented tests are failing everywhere or fail only on Windows.


KacerCZ:
@hectorespert Please review.

PR #2203, opened 2020-06-20
[NETBEANS-4478] Fixed tests for PHP Code analysis on Windows

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4478

Paths in log file have to be system-dependent for tests to work as
expected.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,7 +57,7 @@ jobs:
           ant -f php/php.api.testing test
           ant -f php/php.apigen test
           ant -f php/php.atoum test
-          # ant -f php/php.code.analysis test
+          ant -f php/php.code.analysis test
           ant -f php/php.codeception test
           ant -f php/php.composer test
           # ant -f php/php.dbgp test
--- a/php/php.code.analysis/test/unit/data/phpstan/phpstan-log-netbeans-3022-without-workdir.xml
+++ b/php/php.code.analysis/test/unit/data/phpstan/phpstan-log-netbeans-3022-without-workdir.xml
@@ -20,10 +20,10 @@
 
 -->
 <checkstyle>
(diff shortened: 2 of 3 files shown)


KacerCZ:

junichi11:
Maybe, it is a problem of an absolute path?
I would add a file for windows(e.g. phpstan-log-netbeans-3022-without-workdir-win.xml) since this case needs not a correct path although I'm not sure whether it works fine.

e.g.
```xml
<checkstyle>
<file name="C:\test\phpstan\app\Test1.php">
  <error line="13" column="1" severity="error" message="Binary operation &quot;+&quot; between string and 2 results in an error." />
</file>
<file name="C:\test\phpstan\app\Test2.php">
  <error line="9" column="1" severity="error" message="Anonymous function should return string but returns void." />
  <error line="9" column="1" severity="error" message="Result of closure (void) is used." />
</file>
</checkstyle>
```
```java
        String logFileName = Utilities.isWindows()
                ? "phpstan-log-netbeans-3022-without-workdir-win.xml"
                : "phpstan-log-netbeans-3022-without-workdir.xml";
        List<Result> results = PHPStanReportParser.parse(getLogFile(logFileName), root, workDir);
workDir);
```

BTW, you should attach a log related to the failure to the issue page if it exists. Thanks.

KacerCZ:
I was trying to minimize differences between Windows as other systems.
So I didn't want to duplicate whole file because there is no guarantee that both files will keep same structure.

junichi11:
As I wrote above, I don't have a strong opinion  (because it is not overwrite.)

Please attach logs to issue pages. Thanks.

KacerCZ:
I added relevant part of log to issue.

KacerCZ:
`PHPStanReportParserTest` now uses simple string replacement.

PR #2205, opened 2020-06-20
[NETBEANS-4480] Fixed tests for PHP Project on Windows

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4480

File paths in tests have to be normalized to system-dependent values.


KacerCZ:

PR #2206, opened 2020-06-20
[NETBEANS-4483] Fixed unit tests for Nette Tester

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4483

- Directory separators had to be replaced with system-dependent value.
- Because Windows use \ as separator, replacement text has to be quoted.
- Updated test to match line numbers after license change.
- Enabled tests in Travis and Github Actions

### DIFF
--- a/php/php.nette.tester/test/unit/src/org/netbeans/modules/php/nette/tester/coverage/CloverLogParserTest.java
+++ b/php/php.nette.tester/test/unit/src/org/netbeans/modules/php/nette/tester/coverage/CloverLogParserTest.java
@@ -32,10 +32,6 @@ import org.netbeans.junit.NbTestCase;
 import org.netbeans.modules.php.spi.testing.coverage.Coverage;
 import org.netbeans.modules.php.spi.testing.coverage.FileMetrics;
 
-import static junit.framework.Assert.assertEquals;
-import static junit.framework.Assert.assertNotNull;
-import static junit.framework.Assert.assertTrue;
-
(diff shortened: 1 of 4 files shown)


KacerCZ:

php/php.nette.tester/test/unit/src/org/netbeans/modules/php/nette/tester/coverage/CloverLogParserTest.java
```diff
@@ -98,7 +95,10 @@ private void fixContent(File file) throws Exception {
         Path path = file.toPath();
         Charset charset = StandardCharsets.UTF_8;
         String content = new String(Files.readAllBytes(path), charset);
-        content = content.replaceAll("%WORKDIR%", getDataDir().getAbsolutePath());
+        String workdirReplacement = Matcher.quoteReplacement(getDataDir().getAbsolutePath());
+        content = content.replaceAll("%WORKDIR%", workdirReplacement);
+        String separatorReplacement = Matcher.quoteReplacement(File.separator);
+        content = content.replaceAll("%SEP%", separatorReplacement);
```
tmysik, line 101: And once more - cannot `content.replace("%SEP%", File.separator)` be used?

KacerCZ: Right, I changed it as you suggest.

PR #2212, opened 2020-06-23
Initial prototype for COS applies to .java only for now.

andres620x:
Hi everybody how are you.
I would like you to support me by reviewing this proposal for COS on gradle systems.
Currently only works for .java
I have tried it in spring boot and it has worked for me.
I know a lot of things are missing.
But I will add them with your support and help.
Basically I'm using everything the maven system already uses in the cos.
Also I still have to debug a little code.
But I would like you to give me a guide.

This currently uses the BuildArtifactMapperImpl class.
For the update.
As I spoke with Lazlo, he commented that there could be another echema.
But before I would like you to evaluate this.

I remain attentive to any observation.
I would like @lkishalmi to review this topic


andres620x:
Hi Lazlo,
So I cancel the pull and reassemble it without formatting the code?

lkishalmi:
Also, probably shall we do a step back first.
Usually COS as much pain to deal with as not.
Technically you can have COS out of the box in NetBeans with Gradle, just by starting the desired task with --continuous option. It would leave the build output open, but as soon as you save something it would execute the necessary tasks.

lkishalmi:
Also determining the sourceset of the saved file and invoke compile<Java|Groovy|Scala> or prepareResources task might be better. They are usually quite fast. The problem with COS is that you have to maintain a different set of "truth" when working on the projects. It is could especially complicated with JavaEE apps.

andres620x:
Hi Lazlo,
Checking your idea then would be to detect when the person saves the source and invoke the "classes" task for the main sources and "testclasses" when it is detected to be a test class.
Is this the idea?

lkishalmi:
Theoretically (I'm little uncertauin at the begin of this flow), but you register a save cookie, so whenever a file is saved, yo test, if it belongs to a Gradle project, then retrieve the GradleJavaProject from it, so it can tell you the sourceset you are in, as well as the compile task name for the sourceset. from there it is just a custom gradle action to be called. That's the idea.

andres620x:
Postponed for better analysis

PR #2215, opened 2020-06-25
Distribute the GPLv2+CPEx JavaFX binaries in the nb cluster

jtulach:
To celebrate the magnificent success of @jlahoda in the [legal discussion](https://lists.apache.org/thread.html/rfb90043e7aa7296e0d090b424b6e56ee360ce6eb0c094d9ee034a1a9%40%3Clegal-discuss.apache.org%3E) he started, I decided to follow the advice given by Roy T. Fielding and include the JavaFX libraries in the resulting ZIP binary of NetBeans. According to the GPLv2+CPEx license text and Roy's interpretation we can still claim the whole ZIP is licensed as Category A.

Jan Lahoda suggested to put the libraries into the `nb` cluster to indicate that they are not part of the NetBeans Plaform & co. libraries, but really just a bundled 3rd party dependency of the NetBeans IDE.

Roy suggested that "... to make sure the JavaFX libraries are entirely separate from our product should be a PMC discussion.". Let's use this PR to set a precedent showing how to deal with other GPLv2+CPEx libraries.


ergonomics/ide.ergonomics/java.properties
```diff
@@ -22,8 +22,6 @@ projectImporter=Eclipse
 mainModule=org.netbeans.modules.java.kit
 extra.modules=org.netbeans.modules.nbjavac.*
 extra.modules.recommended.min.jdk=9
-extra.modules.1=org\\.netbeans\\.libs\\.javafx\\.(linux|win|macosx)
-extra.modules.1.recommended.max.jdk=11
```
jtulach, line 26: It is no longer necessary to download JavaFX libraries when enabling modules using them.

neilcsmith-net:
Great news!  Although -1 to merging until it's updated at https://www.apache.org/legal/resolved.html otherwise it makes a bit of a headache for next release manager.

neilcsmith-net:
Also, shouldn't this be against master not release120?

matthiasblaesing:
I agree with Neil here. It looks like the plain "GPLvX-CPE" ist Category-X can be reevaluated, but that was a first "looks like it could be Category-A". However I can imaginge, that there are more strings attached and maybe we still need to be able to create binaries, that don't contain the GPLv2-CPE binaries.

In the mail from Roy there is a clear hint, the the PMC should also first discuss how this should be handled and I tend to agree.

jlahoda:
First, just to make sure, I represent only myself personally here, and I am not a lawyer.

So, I think Roy's argument is, that the license of the resulting executable is (can be) ALv2, in which case we are fine, because Category A rules apply. While it would be awesome to have an entry in "resolved", I am not sure it is strictly necessary. In fact, I believe we are already doing something similar - the Windows launchers link against the standard libraries, which are (AFAIK) GPLv3 with an exception. (I don't want to pretend I can quantify how similar the exception are or are not.) And I believe the recommendation back then was to mark the whole launcher as ALv2, and this is what was done.

How to translate all this to NetBeans IDE as a whole modular application written in Java is (I think) the question now. And I do not have firm answers, but I assume this PR was meant to discuss aspects the PMC can discuss. OTOH, feel free to take it to the mailing list.

The specific hint about PMC was:

> The source of their work would still only be under GPL+CE and
would have to be entirely separate from our products. How you would
make sure of that is a PMC discussion.

We can discuss that as well, but especially in case of JavaFX, I am not sure how the sources for JavaFX and NetBeans could be seen to not be entirely separate.

What I like here is the use of the "nb" cluster. Unlike all other clusters, this one creates the application, that is Apache NetBeans IDE, and hence could be described as defining the "executable" (and it also physically includes the actual launcher/executable file). If that's a sufficient - not sure, but hope so.

jtulach:
I've been told to switch the destination branch to `master`: done in d4a1f67

How shall we move on?

I believe the NetBeans IDE ZIP (& the installers) can be considered executable. The JavaFX NBMs still contain only the ".external" file - e..g. we don't distribute the GPLv2CPEx bits in the NBMs. Only in the ZIP - a case that the GPLv2CPEx explicitly allows and gives us permission to label the whole ZIP/Installers/Executables with Apache license.

Do we need to do more? Based on RMS's "printer story" I believe it is essential for any user of NetBeans to be able to fix a bug in GPLv2CPEx Apache NetBeans libraries ownself and use the patch for him and/or others. In the case of JavaFX it is trivial - the source is on Maven central next to the libraries.

matthiasblaesing:
I read the messages on legal as not yet decided. Its unfortunate, that the term "executable" is not defined. Before that happens and we have the full picture, I'd like not to rush things.

neilcsmith-net:
I've just realised (while looking at it for something else) that even if GPL+CPE gets approval we won't be able to merge this as is.  The media module includes GLib and GStreamer code that is licensed under LGPL.  Even if we didn't include that, further checking shows that Webkit is also LGPL.  Shipping without JavaFX-Web seems less useful?!

https://github.com/openjdk/jfx/tree/jfx14/modules/javafx.media/src/main/legal
https://github.com/openjdk/jfx/tree/jfx14/modules/javafx.web/src/main/legal

JaroslavTulach:
The primary motivation of this PR was to bring modern HTML rendering engine into Apache NetBeans. With the latest finding and current Apache licensing policy, it does not seem possible to include WebKit at all. Closing.

PR #2217, opened 2020-06-27
[NETBEANS-2333][NETBEANS-3681] Update validator.jar for html.validation

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-2333
https://issues.apache.org/jira/browse/NETBEANS-3681

- Update validator (https://github.com/validator/validator)
- `HTML_AUTO`, `HTML401_STRICT`, and `HTML401_TRANSITIONAL` are available no longer
- Fix the license file

I've created a new patch(validator-20200626.patch) based on https://github.com/apache/netbeans/blob/master/ide/html.validation/external/validator.patch . I didn't remove directories and files.

#### before

netbeans-2333-3681-before

#### after

netbeans-2333-3681-after


neilcsmith-net:
@junichi11 could you take a look at the beginning of the stack trace at https://issues.apache.org/jira/secure/attachment/13009548/idelog-6.txt  Not sure it's the cause of the actual issue, but I've seen this stack trace myself at one point during testing,  Is it related to this PR?

matthiasblaesing:
From a quick check this looks as if htmlvalidator relies on CSS validator (https://search.maven.org/artifact/org.w3c.css/css-validator/20100131/jar). The old patch (still present as validator.patch) contained changes, that removed CSS checking if I interpret this correctly.

neilcsmith-net:
@matthiasblaesing ah, thanks, hadn't seen the old patch file was still there. Wasn't sure if the patch was the problem or a missing dependency with the update.  Guess need to decide whether it's a release blocker.

junichi11:
@neilcsmith-net Thanks for the heads up. Matthias has already created the PR(#2308) for that. I'll have a look at it ASAP.
@matthiasblaesing Thank you for investigating it and creating the PR.

PR #2219, opened 2020-06-29
[NETBEANS-4520] Enable File Chooser in non-project New File Wizard

SiddheshRane:
If there is no active project, then the New File Wizard does not allow to open the file chooser for selecting the folder. This patch fixes that without affecting unintended side effects. This is more of feature enhancement.

**Before**
before

**After**
after

**Expected funtionality unaffected**
no-sideeffects


matthiasblaesing:
Thank you - The change looks good. Bug could you please create a JIRA issue for this and update the title of this PR to "[<JIRA-ID>] <Title>"?

geertjanw:
This is pretty cool, really supports the scenario of using NetBeans for single files.

PR #2225, opened 2020-07-01
Make configureproxy idenpotent when invoked multiple times

JaroslavTulach:
Backporting #2210 to release 12.0.x branch.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ConfigureProxy.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ConfigureProxy.java
@@ -84,51 +84,21 @@ public final class ConfigureProxy extends Task {
     static URLConnection openConnection(Task task, final URL url, URI[] connectedVia) throws IOException {
         final URLConnection[] conn = { null };
         final List<Exception> errs = new CopyOnWriteArrayList<>();
+        final StringBuilder msgs = new StringBuilder();
         final CountDownLatch connected = new CountDownLatch(1);
         ExecutorService connectors = Executors.newFixedThreadPool(3);
         connectors.submit(() -> {
-            String httpProxy = System.getenv("http_proxy");
-            if (httpProxy != null) {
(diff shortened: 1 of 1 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ConfigureProxy.java
```diff
@@ -84,51 +84,21 @@ public void execute() throws BuildException {
     static URLConnection openConnection(Task task, final URL url, URI[] connectedVia) throws IOException {
         final URLConnection[] conn = { null };
         final List<Exception> errs = new CopyOnWriteArrayList<>();
+        final StringBuilder msgs = new StringBuilder();
```
sdedic, line 87: Isn't `msgs` operated from different threads ? Noticed only now :(

JaroslavTulach: E.g. it should be `StringBuffer`, not builder. Right?

JaroslavTulach: Done in 40584ecfeb52cfd7f8f5bc619ad97ad968303819 - if approved, I also merge it to `master`.

Btw. the code was ready for being used from multiple threads - see all the `msgs.append("x' + a + " y " + b + " z" + c)` concatentations that are done in single thread and only then placed into the `msgs` object. Why did I not used `StringBuffer` then? I know I wanted...

JaroslavTulach:
Thanks for the review. I merged the changes into master as 58bd5a3 - I'll wait with merging this PR into `release120` branch for Laszló's approval. Or feel free to merge it yourself @lkishalmi!

JaroslavTulach:
This is not a change in a module, but in the infrastructure build tasks. I am not able to build NetBeans behind our corporate proxy without it. I don't think there is a need to increase any spec versions.

PR #2227, opened 2020-07-02
[NETBEANS-4420] Fix IAE on tooltip images with unknown size.

lkishalmi:


platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -307,8 +307,16 @@ public static final Image mergeImages(Image image1, Image image2, int x, int y)
...
-        return (image instanceof ToolTipImage)
+        int h = image.getHeight(null);
+        int w = image.getWidth(null);
+        if ((h == -1) || (w == -1)) {
+            // [NETBEANS-4420] Cannot create ToolTipImage from an Image
+            // which size is not yet known.
+            return new ImageIcon(image);
+        } else {
+            return (image instanceof ToolTipImage)
                 ? (ToolTipImage) image : assignToolTipToImageInternal(image, "");
```
eirikbakke, line 318: assignToolTipToImageInternal is actually supposed to handle the case of not-yet-loaded images; it calls ToolTipImage.createNew, which calls ImageUtilities.ensureLoaded(Image) in turn before trying to get the width/height of the image. So the exception probably only occurs when there is some kind of error loading the image.

To preserve handling of not-yet-loaded images, I'd suggest a fix in ToolTipImage.createNew instead. Where it currently says

```
int w = image.getWidth(null);
int h = image.getHeight(null);
```

just do

```
// Handle images that couldn't be loaded (width/height=-1), despite our earlier call to ensureLoaded, gracefully.
int w = Math.max(1, image.getWidth(null));
int h = Math.max(1, image.getHeight(null));

```
instead. Is my thinking right here?

lkishalmi: Well, I can try that.
Would it be good for the 
model.createCompatibleWritableRaster(w, h) (Line:1079)?
Actually that is where the original IAE is coming from if the size of the unknown.
Going to try that anyway...

lkishalmi: It seems to work, at least no IAE exception has been seen so far. I'll mark this one to be backported to 12.0-u1.

eirikbakke: I think it's good. Come to think of it, perhaps also put in the Math.max(1, ...) trick in doMergeImages as well. E.g. do

int w = Math.max(1, Math.max(image1.getWidth(null), x + image2.getWidth(null)));
int h = Math.max(1, Math.max(image1.getHeight(null), y + image2.getHeight(null)));

Sorry for not catching this one earlier.

lkishalmi: No problem. did that as well!

PR #2228, opened 2020-07-02
[NETBEANS-4509] PHP - code completion for function with return type

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4509

Fixed code completion of functions with return type "array" and specific
type in "@return" in DocBlock.

Example:
```php
/**
 * @return \SplFileInfo[]
 */
function getFiles(): array { }

foreach (getFiles() as $file) {
    $file->^ // Offers methods from \SplFileInfo
}
```
Before this fix methods from `\SplFileInfo` will not be offered.


tmysik:
@junichi11 Thanks a lot! :)

PR #2230, opened 2020-07-03
[NETBEANS-4541] Set working directory when running PHP CS Fixer

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4541

Set working directory when PHP CS Fixer is run.
Project configuration file is now found.

Output before: `Loaded config default.`
Output after: `Loaded config default from "C:\Devel\PhpProject\.php_cs.dist".`


tmysik:
Nice! Thanks!

PR #2240, opened 2020-07-06
[NETBEANS-4548] Enable launcher to detect java installed by sdkman

lkishalmi:


trixon:
@lkishalmi Works like a charm, thanks!
Would it make sense to apply this for platform based applications too?

lkishalmi:
Well this change is on the platform launcher, so from 12.1 it does work on platform based apps as well.

trixon:
My bad, was looking in the wrong place.
I see it now, but only for ant based applications.

PR #2241, opened 2020-07-07
[NETBEANS-4503] Code completion does not work in functions declared inside methods

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4503

#### 1st commit

Nested functions are in a Method/FunctionScope.
However, they are added as elements of a NamespaceScope.
i.e. Method/FunctionScope doesn't have nested FunctionScopes. NamespaceScope has them.
So, (not return the child result simply,) check whether the child result contains the parent result in the `VariableScopeFinder.findWrapper()`.

#### 2nd commit

Unit test(testNb4503_01()) failed on GitHub Actions after I pushed the first commit.
However, It passed on my local pc.

After I've tried some things, I've somehow reproduced it.
```java
    public void testNb4503_01() throws Exception {
        Thread.sleep(1000); // added this, then reproduced
        testNb4503("            $e->^test(); // test1");
    }
```
Unfortunately, if I set a breakpoint, I couldn't reproduce it. Just used `println()` to investigate the cause. 

The cause:
A cached `Source` instance may be removed after elements are indexed.
So, the test file is scanned again when the test is run. If there are elements for LazyBuild, we may not be able to get the correct result.
To avoiding that, after elements are scanned lazily, find the scope again.

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
@@ -23,6 +23,10 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import org.netbeans.api.annotations.common.CheckForNull;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.netbeans.api.annotations.common.NullAllowed;
 import org.netbeans.modules.csl.api.OffsetRange;
 import org.netbeans.modules.php.editor.model.impl.LazyBuild;
 import org.openide.filesystems.FileObject;
@@ -32,6 +36,8 @@ import org.openide.filesystems.FileObject;
  * @author Ondrej Brejla <obrejla@netbeans.org>
  */
 public final class VariableScopeFinder {
+
+    private static final Logger LOGGER = Logger.getLogger(VariableScopeFinder.class.getName());
 
     public static VariableScopeFinder create() {
         return new VariableScopeFinder();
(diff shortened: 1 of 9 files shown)


php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
```diff
@@ -45,10 +51,22 @@ public VariableScope find(final Scope scope, final int offset, final ScopeRangeA
...
+        AtomicBoolean isLazilyScanned = new AtomicBoolean(false);
+        VariableScope variableScope = findWrapper(elements, offset, scopeRangeAcceptor, isLazilyScanned).getVariableScope();
+        if (isLazilyScanned.get()) {
+            // some scopes may be added new elements when LazyBuild elements are scanned.
+            // so, find again.
+            // e.g. Source instances are cached as weak references.
+            // so, ParserResult may not be the same instance even if the FileObject is the same.
+            // it means that new Model and new ModelVisitor are created again. in such case, LazyBuild elements may not be scanned yet.
+            LOGGER.log(Level.FINE, "(LazyBuild)Scope is scanned."); // NOI18N
+            isLazilyScanned.set(false);
```
tmysik, line 63: Maybe better [`compareAndSet()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicBoolean.html#compareAndSet-boolean-boolean-)?

tmysik: Ha, it is running in just one thread, right? In such a case, not needed, of course.

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
```diff
@@ -45,10 +51,22 @@ public VariableScope find(final Scope scope, final int offset, final ScopeRangeA
...
+            variableScope = findWrapper(elements, offset, scopeRangeAcceptor, isLazilyScanned).getVariableScope();
+        }
+        return variableScope;
     }
 
-    private VariableScopeWrapper findWrapper(final List<? extends ModelElement> elements, final int offset, final ScopeRangeAcceptor scopeRangeAcceptor) {
+    private VariableScopeWrapper findWrapper(final List<? extends ModelElement> elements, final int offset, final ScopeRangeAcceptor scopeRangeAcceptor, final AtomicBoolean isLazilyScanned) {
```
tmysik, line 69: Missing `@Nullable`?

tmysik:
@junichi11 Let me know, I am ready to merge it as it is now. Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
junichi11: yes, will remove.

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
junichi11: Will add. Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
junichi11: Oops, will change it to your suggestion :)

junichi11:
@tmysik Fixed https://github.com/apache/netbeans/compare/117cdec493f2a6685c94c9635eb8b54e5f1d8f0b..9fa8489abf12899ad57e22bf1d586d645c25f3cb.
Maybe, CI takes about 2 hours. Thank you!

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
tmysik: If it is running in only one thread (as here I guess, in the _parsing_ thread) then `AtomicBoolean` is just a holder for a boolean. So, using `compareAndSet()` is IMHO not necessary.

tmysik:
@junichi11 

Looks good to me, will merge once CI passes.

php/php.editor/src/org/netbeans/modules/php/editor/model/VariableScopeFinder.java
junichi11: I seem to misunderstand something. Sorry... I'll revert it if I modify this class in the future. Thank you for your review :)

tmysik: No problem, my fault because of my misleading/confusing comment.

PR #2242, opened 2020-07-07
[NETBEANS-4185] Support for self return type

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4185

Added test files based on [PHPCodeCompletion269108Test](https://github.com/apache/netbeans/blob/master/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletion269108Test.java) testing same feature but without return type in method.

Fixed case when native return type `self` is used in combination with PHPDoc annotations `@return static` and `@return $this.`

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/model/impl/VariousUtils.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/model/impl/VariousUtils.java
@@ -225,8 +225,9 @@ public final class VariousUtils {
             QualifiedName name = QualifiedName.create(returnType);
             assert name != null : returnType;
             String typeName = name.toString();
-            if (Type.ARRAY.equals(typeName)) {
+            if (Type.ARRAY.equals(typeName) || Type.SELF.equals(typeName)) {
                 // For "array" type PHPDoc can contain more specific definition, i.e. MyClass[]
+                // For "self" type PHPDoc can contain more specific definition, i.e. static or $this
                 String typeFromPHPDoc = getReturnTypeFromPHPDoc(root, functionDeclaration);
                 if (typeFromPHPDoc != null) {
                     return typeFromPHPDoc;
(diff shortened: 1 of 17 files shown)


KacerCZ:
@tmysik @junichi11 There is problem if method has native return type `self` than `@return` annotation is not considered.
Objects `MethodScopeImpl` has set return type to `\self' and I can't find place where to use value from `@return`.
Method `FunctionScopeImpl::getReturnTypesDescriptor()` returns different value than in case when method has not native `self` return type.

Any advice how to fix this issue?

junichi11:
@KacerCZ Unfortunately, I can't provide something for it soon. (I have to read the source code to investigate the problem actually.) If you are not sure, I'll take over it later. If you would leave it to me, please let me know(please put on me to Assignee). Thanks.

tmysik:
@KacerCZ 

Please, have you verified that the reported case is valid? I mean, it is some time now so I "left" the PHP world so just would like to get confirmed that it is a valid PHP (PhpDoc) use-case.

Thank you!

KacerCZ:
@tmysik Yes, it is valid.
Type `static` is subtype of `self`. Using `self` as native return type is still better than leaving it out completely.
It is similar case as when you use type `array` in formal parameter and then use `@param \DateTime[]` to indicate that array should contain `\DateTime` objects.

PHP 8 will have native `static` return type - https://wiki.php.net/rfc/static_return_type

tmysik:
@KacerCZ Great, thank you!

KacerCZ:
The reported bug is fixed.
I had wrong PHP test file - declaration of `A::returnSelf(): self` can't be overiden with `B::returnSelf(): self`.

KacerCZ:

junichi11:
@mvorisek Please report your problem to the JIRA as a new issue with your exact steps to reproduce it.

PR #2244, opened 2020-07-09
[NETBEANS-4652] Change the max level of PHPStan to 8

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4562

- https://phpstan.org/user-guide/rule-levels

netbeans-4652-phpstan-level8

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptions.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/options/AnalysisOptions.java
@@ -56,7 +56,7 @@ public final class AnalysisOptions {
     private static final String PHPSTAN_CONFIGURATION = "phpstan.configuration"; // NOI18N
     private static final String PHPSTAN_MEMORY_LIMIT = "phpstan.memory.limit"; // NOI18N
     public static final int PHPSTAN_MIN_LEVEL = Integer.getInteger("nb.phpstan.min.level", 0); // NOI18N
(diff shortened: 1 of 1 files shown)


tmysik:
@junichi11 

Just thinking whether it would make sense to make it configurable... :thinking: Probably not necessary :)

BTW can I merge it, right? The test failure seems unrelated to me. Or am I wrong?

Thanks!

junichi11:
@tmysik 
> Just thinking whether it would make sense to make it configurable...  Probably not necessary :)

Let's think that after the new level is added in the future again :) e.g. We can add the max level combobox. or change to the text field.

> BTW can I merge it, right? The test failure seems unrelated to me. Or am I wrong?

It should be OK. Maybe this: https://lists.apache.org/thread.html/r680b5a50fc7c8d07c91c0750df6cafa1e49732cbfdfbc2aa15c3c012%40%3Cdev.netbeans.apache.org%3E

Thank you!

tmysik:
@junichi11 

> We can add the max level combobox. or change to the text field.

I was thinking more about a simple configuration (properties) file with a max level. Anyone could update it manually without a code change. And yes, if someone puts there `1000` then it is his/her fault :)

junichi11:
@tmysik 
> I was thinking more about a simple configuration (properties) file with a max level. Anyone could update it manually without a code change. And yes, if someone puts there `1000` then it is his/her fault :)

Makes sense. Good idea!

PR #2245, opened 2020-07-09
[NETBEANS-4507] Fix Composer search when package has no description

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4507

Fixed parsing of `composer search` output when package does not have
description.

This happens when [composer-asset-plugin](https://github.com/fxpio/composer-asset-plugin) is installed. Then output contains also packages without description.


KacerCZ:

tmysik:
@KacerCZ Thanks for fixing it!

PR #2250, opened 2020-07-12
Minor improvements for php.api.phpmodule

junichi11:
- Add the `isSupportedVersion()` method to the PhpVersion

    - Check whether the version is supported yet by PHP official
    - Improve the `getLegacy()` method

- Add the `getFirstError()` and the `getFirstWarning()` methods to the `ValidationResult` class

### DIFF
--- a/php/php.api.phpmodule/manifest.mf
+++ b/php/php.api.phpmodule/manifest.mf
@@ -1,4 +1,4 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.php.api.phpmodule
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/php/api/phpmodule/resources/Bundle.properties
-OpenIDE-Module-Specification-Version: 2.71
+OpenIDE-Module-Specification-Version: 2.72
--- a/php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
+++ b/php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
@@ -46,6 +46,7 @@ public final class ValidationResult {
 
     /**
      * Copy constructor.
+     * @param anotherResult another validation result
      */
     public ValidationResult(ValidationResult anotherResult) {
         merge(anotherResult);
@@ -81,6 +82,20 @@ public final class ValidationResult {
     }
 
     /**
+     * Get the first error.
+     *
+     * @return the first error or {@code null} if there are no errors
+     * @since 2.72
+     */
+    @CheckForNull
+    public Message getFirstError() {
+        if (hasErrors()) {
+            return errors.get(0);
+        }
+        return null;
+    }
+
+    /**
      * Check whether there are some warnings present.
      * @return {@code true} if the validation result contains any warning
      * @see #isFaultless()
@@ -96,6 +111,20 @@ public final class ValidationResult {
     public List<Message> getWarnings() {
         return new ArrayList<>(warnings);
 
+    }
+
+    /**
+     * Get the first warning.
+     *
+     * @return the first warning or {@code null} if there are no warnings
+     * @since 2.72
+     */
+    @CheckForNull
+    public Message getFirstWarning() {
+        if (hasWarnings()) {
+            return warnings.get(0);
+        }
+        return null;
     }
 
     /**
--- a/php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
+++ b/php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
@@ -18,6 +18,7 @@
  */
 package org.netbeans.modules.php.api;
 
+import java.time.LocalDate;
 import org.openide.util.NbBundle;
 
 /**
@@ -117,6 +118,11 @@ public enum PhpVersion {
      */
     public static PhpVersion getLegacy() {
         PhpVersion[] phpVersions = PhpVersion.values();
+        for (PhpVersion phpVersion : phpVersions) {
+            if (phpVersion.isSupportedVersion()) {
+                return phpVersion;
+            }
+        }
         return phpVersions[phpVersions.length - 2];
     }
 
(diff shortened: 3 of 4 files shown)


php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
```diff
@@ -80,6 +81,20 @@ public boolean hasErrors() {
...
+     * @return the first error.
+     * @since 2.72
+     */
+    @CheckForNull
+    public Message getFirstError() {
+        if (hasErrors()) {
+            return errors.get(0);
+        }
+        return null;
+    }
```
junichi11, line 96: I also thought other implementations:

#### Create Message.EMPTY and return it 
```java
    public Message getFirstError() {
        if (hasErrors()) {
            return errors.get(0);
        }
        return Message.EMPTY;
    }

    public static final class Message {
        public static final Message EMPTY = new Message("", ""); 

```
#### Use Optional
```java
    public Optional<Message> getFirstError() {
        if (hasErrors()) {
            return Optional.of(errors.get(0));
        }
        return Optional.empty();
    }
```
Which is better?

tmysik: I think that `Message.EMPTY` would complicate things because the caller would need to (a) check whether it is an empty message (which is similar for the current check for `null`) or (b) adjust UI not to show the empty error as an error.

Using `Optional` would be better but I would perhaps keep the current implementation, it is more _consistent_ with the rest of the NetBeans codebase (and we have `@CheckForNull`). But up to you, feel free to use it if you want to, I definitely have no objections.

Thanks.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
```diff
@@ -80,6 +81,20 @@ public boolean hasErrors() {
         return new ArrayList<>(errors);
     }
 
+    /**
+     * Get the first error.
+     *
+     * @return the first error.
```
tmysik, line 87: Maybe to be added `or {@code null} if there are no errors`?

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
```diff
@@ -98,6 +113,20 @@ public boolean hasWarnings() {
 
     }
 
+    /**
+     * Get the first warning.
+     *
+     * @return the first warning.
```
tmysik, line 119: The same note as above.

php/php.api.phpmodule/manifest.mf
```diff
@@ -1,4 +1,4 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.php.api.phpmodule
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/php/api/phpmodule/resources/Bundle.properties
-OpenIDE-Module-Specification-Version: 2.71
+OpenIDE-Module-Specification-Version: 2.72
```
tmysik, line 4: Don't we want to increase the dependency version in `php.kit`?

tmysik:
@junichi11 Looks good to me, just a few comments/questions. Thanks!

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
junichi11: Makes sense. I agree with you. I'll keep this. Thanks a lot!

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
junichi11: Will fix. Thanks!

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
junichi11: Will fix. Thanks!

php/php.api.phpmodule/manifest.mf
junichi11: It seems that `php.kit` doesn't have `phpmodule.api`as module-dependencies: https://github.com/apache/netbeans/blob/master/php/php.kit/nbproject/project.xml

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
junichi11: Fixed.
https://github.com/apache/netbeans/compare/5685f82c7815b1441ffe2188fa7c327dc79c8685..6c2e311da592782e23c67563514a380699ed4314

junichi11:
@tmysik Thanks a lot for your review and your time!

php/php.api.phpmodule/manifest.mf
tmysik: OK, so some other module has to depend on it (my bet is `php.project`). But that would mean that also version of that module (together with `php.kit`) would need to be increased - probably not needed (I have no idea how "live" NB updates are). So, let's keep it as it is, thanks.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
```diff
@@ -98,6 +113,20 @@ public boolean hasWarnings() {
 
     }
 
+    /**
+     * Get the first warning.
+     *
+     * @return the first warning or {@code null} if there are no errors
```
tmysik, line 119: Should be `[...] no [warnings]` ;)

tmysik: Please, let me know if you are going to change it. If not, I will merge this PR. Thanks.

junichi11: Will fix it, sorry...

php/php.api.phpmodule/manifest.mf
junichi11: > OK, so some other module has to depend on it (my bet is `php.project`).

Yes, `php.project`, `php.editor`, etc.

> So, let's keep it as it is

OK.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/validation/ValidationResult.java
junichi11: Fixed.

PR #2252, opened 2020-07-13
Speeding up build, fixing some dependencies as needed.

jlahoda:
Currently, even incremental build is fairly slow. This is an attempt to improve the speed of the build - it reduces the build time by about 4 minutes on my computer, from ~16mins to ~12mins for a full build and ~10mins to 6mins for an incremental build (which is still a lot, of course).

The basic idea is simple: currently, when doing a build, we build every cluster separately in an ant instance. And every module has (generated) target like "all-java.source.base" depends="all-openide.util.lookup,...". Note that the module depends directly on modules from different clusters. So, when building the "java" cluster, we also do a build of "all-openide.util.lookup", even if we have already built the "platform" cluster. This, of course, does only a timestamp check, but takes some time anyway. And we do this for every cluster, and many more modules, accumulating to a significant chunk of time.

The proposed solution is to generate the dependencies so that "all-<module>" depends on a target that ensures all needed *clusters* are built, and on all-<dependent-module> targets for modules from the current cluster. This avoids the unnecessary timestamp checks and improves build time. The code was actually expected to mostly work like this, but there was a small coding mistake, which prevented it from doing that.

Sadly, the current repository contains several cases where modules have dependencies on modules from other clusters, but there is no sensible dependency between the clusters. Namely:
-extide/gradle depends on java/java.platform, but "java" depends on "extide", so this is a dependency in the wrong direction. I have moved gradle to the java cluster for the purpose of this patch.
-java/debugger.jpda.trufflenode depends on webcommon/javascript.nodejs, but the "java" cluster does not depend on the "webcommon" cluster. I made "java" cluster depend on "webcommon" for the purpose of this patch.
-nb/ko4j.debugging depends on java/maven, but the "nb" cluster does not (and should not) depend on the "java" cluster. I've moved the ko4j.debugging module into the "java" cluster for the purposed of this patch.

I'd be happy to resolve the issues in other ways, as desired, but it would be awesome if we could speed up the build and make our dependencies cleaner!


nbbuild/antsrc/org/netbeans/nbbuild/InsertModuleAllTargets.java
```diff
@@ -136,8 +136,29 @@ public void setUseClusters(boolean b) {
                     continue;
                 }
                 String[] prereqsAsCnb = entry.getBuildPrerequisites();
-                StringBuffer namedDeps = new StringBuffer("init");
-                String myCluster = clustersOfModules.get(path);
+                StringBuilder namedDeps = new StringBuilder("init");
+                String myCluster = clustersOfModules.get(entry.getNetbeansOrgId());
+                if (myCluster != null) {
+                    String clusterDep = "all-cluster-" + myCluster;
+                    if (!prj.getTargets().keySet().contains(clusterDep)) {
```
JaroslavTulach, line 143: I see, this is the fix then.

lkishalmi:
One Question. Why the gradle and libs.gradle ended up in the java cluster?

jlahoda:
Regarding gradle - the gradle module depends on "java.platform" from the "java" cluster. Normally java cluster depends on extide, but here is a module from extide depending on a module from java - that is a dependency in the wrong direction. Moving gradle is probably not the only solution, so if you have a better one, great!

lkishalmi:
Good catch. Thanks. Maybe I should remove the java.platform dependency from the Gradle module instead.

lkishalmi:
Gradle project is now independent from Java Platform.

neilcsmith-net:
Should we merge this now?

PR #2254, opened 2020-07-14
[NETBEANS-4573] PHP CS Fixer doesn't work with a directory and a file

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4573

The relative paths from the working directory are used in XML log if the working directory is set. So, set it as the root directory.

### DIFF
--- a/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodingStandardsFixer.java
+++ b/php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodingStandardsFixer.java
@@ -172,15 +172,17 @@ public final class CodingStandardsFixer {
     public List<Result> analyze(CodingStandardsFixerParams params, FileObject file) {
         assert file.isValid() : "Invalid file given: " + file;
         try {
-            
-            Integer result = getExecutable(Bundle.CodingStandardsFixer_analyze(analyzeGroupCounter++), findWorkDir(file))
+            File workDir = findWorkDir(file);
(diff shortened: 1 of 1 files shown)


php/php.code.analysis/src/org/netbeans/modules/php/analysis/commands/CodingStandardsFixer.java
```diff
@@ -172,15 +172,17 @@ public void startAnalyzeGroup() {
...
                     .runAndWait(getDescriptor(), "Running coding standards fixer..."); // NOI18N
             if (result == null) {
                 return null;
             }
-
-            return CodingStandardsFixerReportParser.parse(XML_LOG, file);
+            // if the project for the file is not found(i.e. if the workDir is not found),
+            // the results are not shown in the inspector window
+            FileObject root = workDir != null ? FileUtil.toFileObject(workDir) : file;
+            return CodingStandardsFixerReportParser.parse(XML_LOG, root);
```
tmysik, line 185: Just to be sure - isn't a problem that before it was `file` but now it can be a directory?

junichi11: I think it's OK. It just worked with the project directory (i.e. when the working directory is the project directory) before I fixed this. 
The case where the project directory(i.e. working directory) cannot be found: a user inspects a file or a directory from a Favorite pane. In such case, the API side doesn't show the result in the inspector window. But the XML result is shown in the Output window. I confirmed that.

tmysik: OK then. Thanks!

PR #2255, opened 2020-07-14
Removed java.platform dependency from the Gradle Projects module

lkishalmi:
Well, this one shall fix the java.platform dependency issue for #2252. Unfortunately it means that the API compatibility is broken. Please review this and point at the weak spots!

### DIFF
--- a/extide/gradle/apichanges.xml
+++ b/extide/gradle/apichanges.xml
@@ -83,6 +83,34 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="gradle-java-independence">
+            <api name="general"/>
+            <summary>Remove Java Platform dependency.</summary>
+            <version major="2" minor="3"/>
+            <date year="2020" month="7" day="15"/>
+            <author login="lkishalmi"/>
+            <compatibility binary="compatible" source="compatible"/>
+            <description>
+                <p>
+                    Due to recent movement and the plan to support Gradle native
+                    projects as well in the future the dependency on
+                    Java Platform module has been removed.
+                </p>
+                <p>
+                    The removal affects <code>org.netbeans.modules.gradle.api.execute.RunUtils</code>
+                    <code>getActivePlatform</code> methods. The method returns a
+                    Pair with <code>null</code> as <code>second</code> which
+                    indicates a broken Platform. This is for keeping the binary compatibility only.
+                    These methods shall be not used at all. There is no
+                    replacement planned for these API functions.
+                </p>
+                <p>
+                    Added <code>org.netbeans.modules.gradle.spi.exectute.GradleJavaPlatformProvider</code>
+                    which can be put into the project lookup, to be able to specify the JDK
+                    to be used for Gradle build execution.
+                </p>
+            </description>
+        </change>
         <change id="gradle-project-trust">
             <api name="general"/>
             <summary>Introduce Trust relationship with Gradle Projects to prevent unintentional Gradle invocation.</summary>
--- a/extide/gradle/manifest.mf
+++ b/extide/gradle/manifest.mf
@@ -3,4 +3,4 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.gradle/2
 OpenIDE-Module-Layer: org/netbeans/modules/gradle/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/gradle/Bundle.properties
-OpenIDE-Module-Specification-Version: 2.2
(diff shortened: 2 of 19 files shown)


extide/gradle/nbproject/project.xml
```diff
@@ -139,15 +139,6 @@
                         <specification-version>1.41.1</specification-version>
                     </run-dependency>
                 </dependency>
-                <dependency>
```
JaroslavTulach, line 142: OK.

extide/gradle/nbproject/org-netbeans-modules-gradle.sig
```diff
@@ -1,1074 +0,0 @@
-#Signature file v4.1
-#Version 2.0
-
```
JaroslavTulach, line 3: I don't think you want to completely remove the file. You may want to remove just the lines that are no longer present.

extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
```diff
@@ -347,41 +337,6 @@ public static ReplaceTokenProvider simpleReplaceTokenProvider(final String token
...
 
- /**
-     * Returns the active platform used by the project or null if the active
-     * project platform is broken.
-     * @param activePlatformId the name of platform used by Ant script or null
-     * for default platform.
-     * @return active {@link JavaPlatform} or null if the project's platform
-     * is broken
-     */
-    public static Pair<String, JavaPlatform> getActivePlatform(final String activePlatformId) {
```
JaroslavTulach, line 358: You can keep these two methods around, just `@deprecated`. Change the return type to `Pair` - e.g. the raw type. JVM is going to link to these two methods without a problem. See more in my [erasure](http://wiki.apidesign.org/wiki/Erasure) blog post.

extide/gradle/manifest.mf
```diff
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 AutoUpdate-Show-In-Client: false
-OpenIDE-Module: org.netbeans.modules.gradle/2
```
JaroslavTulach, line 3: I suggest to keep the removed methods around, just deprecated. Then you don't have to increment the release version and you can stick with `/2`.

Btw. the methods could even delegate to some provider to work the same way that did so far...

lkishalmi: Thanks! Great idea! I could return a <String, null> pair. That's supposed to be handled as a broken/invalid platform it would be sightly better than responding a straight null.

lkishalmi:
Got an approval and a clean CI state before updating the apichanges.xml, merging it now.

extide/gradle/nbproject/org-netbeans-modules-gradle.sig
```diff
@@ -480,8 +480,8 @@ meth public static org.netbeans.modules.gradle.spi.actions.ReplaceTokenProvider
 meth public static org.openide.execution.ExecutorTask executeGradle(org.netbeans.modules.gradle.api.execute.RunConfig,java.lang.String)
 meth public static org.openide.filesystems.FileObject extractFileObjectfromLookup(org.openide.util.Lookup)
 meth public static org.openide.filesystems.FileObject[] extractFileObjectsfromLookup(org.openide.util.Lookup)
-meth public static org.openide.util.Pair<java.lang.String,org.netbeans.api.java.platform.JavaPlatform> getActivePlatform(java.lang.String)
-meth public static org.openide.util.Pair<java.lang.String,org.netbeans.api.java.platform.JavaPlatform> getActivePlatform(org.netbeans.api.project.Project)
+meth public static org.openide.util.Pair getActivePlatform(java.lang.String)
```
JaroslavTulach, line 483: Great.

extide/gradle/apichanges.xml
```diff
@@ -83,6 +83,34 @@ is the proper place.
...
+            <author login="lkishalmi"/>
+            <compatibility binary="compatible" source="compatible"/>
+            <description>
+                <p>
+                    Due to recent movement and the plan to support Gradle native
+                    projects as well in the future the dependency on
+                    Java Platform module has been removed.
+                </p>
+                <p>
+                    The removal affects <code>org.netbeans.modules.gradle.api.execute.RunUtils</code>
```
JaroslavTulach, line 100: Btw. it is possible to use `<a href="@TOP@/org/netbeans/modules/gradle/api/execute/RunUtills.html">RunUtils</a>` to connect to the Javadoc generated classes.

lkishalmi: Thanks! I'm going to try to keep that in mind.

jlahoda:
Thanks Laszlo for doing this! (I apologize for not commenting sooner - too much to do.)

PR #2261, opened 2020-07-17
Improve GraalVM script engines interoperability.

sdedic:
There are a few types of changes:
- the exceptions thrown are better reported: the original exception is attached as the `cause` of the `RuntimeException` (actually `PolyglotException`) thrown from eval/invoke methods.
- better (un)wrapping of `org.graalvm.polyglot.Value` from/to POJOs
- the `ScriptEngine` returned by `Scripting.createManager()` will support `ScriptContext.ENGINE` scope as the top-level language scope.
- global attributes are now supported, even before the first `eval` or other `Context`-creating method on the Engine is called.

I also somewhat checked the testu results agains GraalVM 20.1 (now travis runs 19.3.1)

### DIFF
--- a/ide/libs.graalsdk/src/org/netbeans/libs/graalsdk/impl/GraalEngine.java
+++ b/ide/libs.graalsdk/src/org/netbeans/libs/graalsdk/impl/GraalEngine.java
@@ -31,9 +31,11 @@ import org.graalvm.polyglot.Value;
 final class GraalEngine implements ScriptEngine, Invocable {
 
     private final GraalEngineFactory factory;
+    private final ScriptContext langContext;
 
     GraalEngine(GraalEngineFactory f) {
         this.factory = f;
+        this.langContext = new LangContext(factory.id, f.ctx);
     }
 
     private String id() {
@@ -41,8 +43,8 @@ final class GraalEngine implements ScriptEngine, Invocable {
     }
 
     @Override
-    public GraalContext getContext() {
-        return factory.ctx;
+    public ScriptContext getContext() {
+        return langContext;
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


ide/libs.graalsdk/src/org/netbeans/libs/graalsdk/impl/GraalEngine.java
```diff
@@ -55,20 +57,32 @@ public Object eval(String src, ScriptContext arg1) throws ScriptException {
...
+        public Value run();
+    }
+    
+    private Value handleException(ScriptAction r) throws ScriptException {
         try {
-            return ((GraalContext) arg1).ctx().eval(id(), src);
+            return r.run();
         } catch (PolyglotException e) {
+            if (e.isHostException()) {
+                e.initCause(e.asHostException());
```
JaroslavTulach, line 70: OK.

neilcsmith-net:
Is the Travis failure in the GraalVM tests a concern or coincidence?

sdedic:
> Is the Travis failure in the GraalVM tests a concern or coincidence?

Concern.  Apologies, I've didn't run full tests.

Actually ... it's very interesting. @JaroslavTulach could have more deeper insight in this matter. The failure is actually caused by  **classloader mismatch**. 

`GraalJSTest` - the testcase actually run by `junit` ant task is loaded by AppClassLoader, while hamcrest is being loaded by ModuleSystem$SystemClassLoader. There are two inconsistencies:
- NbModuleSuite only loads `org.junit.**` from AppClassLoader, not `org.hamcrest.**`
- the `lib.junit4` module does not list `org.hamcrest.**` as public packages, although it contains the relevant classes.

JaroslavTulach:
>     * NbModuleSuite only loads `org.junit.**` from AppClassLoader, not `org.hamcrest.**`

cc43414 seems to be the correct fix.

JaroslavTulach:
Sváťa has gone for a vacation. @neilcsmith-net if it is OK to merge, feel free to do so. I am convinced the change prepared by Sváťa improves the exception handling quite significantly.

sdedic:
Yes, actually waiting for re-approval after the changes. Sorry, I was out of Internet connection yesterday

neilcsmith-net:
@sdedic go ahead from my perspective. Your choice whether to squash or not first.  Whatever makes sense.

PR #2262, opened 2020-07-17
Backport of OQL finalizables fix

daemontus:
Backport of OQL finalizables bugfix from VisualVM (see [d936905](https://github.com/oracle/visualvm/commit/d9369059b5017c868e831bd1317381718fbe55c1) and [bd9bf46](https://github.com/oracle/visualvm/commit/bd9bf46131fd9e6c1d25ec475775aefbc6540021)).

The PR corrects previous implementation of `heap.finalizables()` in OQL engine which used to fail with a `ClassCastException` for any heap file. Additionally, it also includes a small fix for JavaScript test cases which called several OQL functions incorrectly (reason why the failing behaviour of `finalizables` was not discovered before).

(cc: @JaroslavTulach)


JaroslavTulach:
Thanks Samuel for discovering and fixing the problem.

PR #2265, opened 2020-07-19
[NETBEANS-4443] PHP 8.0 Support: Non-capturing catches part 

junichi11:

- https://wiki.php.net/rfc/non-capturing_catches
- Fix the ASTPHP5Parser.cup
- Add the `PHP80UnhandledError` class
- Add unit tests for the parser, GotoDeclaration, MarkOccurrences, Formatter, and Code Completion features

nb-php80-non-capturing-catches


junichi11:
@neilcsmith-net I've created a new branch (php80-support). We will merge PRs related to PHP8.0 into it. Thanks.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
```diff
@@ -84,7 +85,12 @@
      * PHP 7.4.
      * @since 2.65
      */
-    PHP_74(Bundle.PhpVersion_PHP_74());
+    PHP_74(Bundle.PhpVersion_PHP_74()),
+    /**
+     * PHP 7.4.
+     * @since 2.74
+     */
+    PHP_80(Bundle.PhpVersion_PHP_80());
```
junichi11, line 93: I didn't increment the spec version because I would like to avoid conflicting. Probably, the version is 2.74(2.74 + 2) when the merge window is opened. So, I fixed only `@since`.
current spec version:
https://github.com/apache/netbeans/blob/eb15c28aa6ee25d5e9c673ad04c0f21c7079b501/php/php.api.phpmodule/manifest.mf#L4

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
```diff
@@ -218,6 +224,7 @@ public String toString() {
         PHP_72(LocalDate.of(2017, 11, 30), LocalDate.of(2019, 11, 30), LocalDate.of(2020, 11, 30)),
         PHP_73(LocalDate.of(2018, 12, 6), LocalDate.of(2020, 12, 6), LocalDate.of(2021, 12, 6)),
         PHP_74(LocalDate.of(2019, 11, 28), LocalDate.of(2021, 11, 28), LocalDate.of(2022, 11, 28)),
+        PHP_80(LocalDate.of(2020, 11, 26), LocalDate.of(2022, 11, 26), LocalDate.of(2023, 11, 26)),
```
junichi11, line 227: This is the scheduled date: https://wiki.php.net/todo/php80

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
```diff
@@ -84,7 +85,12 @@
      * PHP 7.4.
      * @since 2.65
      */
-    PHP_74(Bundle.PhpVersion_PHP_74());
+    PHP_74(Bundle.PhpVersion_PHP_74()),
+    /**
+     * PHP 8.0.
+     * @since 2.74
+     */
+    PHP_80(Bundle.PhpVersion_PHP_80());
```
junichi11, line 93: I didn't increment the spec version because I would like to avoid conflicting. Probably, the version is 2.74(2.72 + 2) when the merge window is opened. So, I fixed only `@since`.
current spec version:
https://github.com/apache/netbeans/blob/eb15c28aa6ee25d5e9c673ad04c0f21c7079b501/php/php.api.phpmodule/manifest.mf#L4

neilcsmith-net:
@junichi11 sounds good! Note that if the current spec version is 2.72, then the spec version for 12.2 should (by default) be 2.73.  It's only automatically incremented once on merge window opening.

junichi11:
@neilcsmith-net 
> Note that if the current spec version is 2.72, then the spec version for 12.2 should (by default) be 2.73. It's only automatically incremented once on merge window opening.

Thanks for the information. I see. Will fix the `@since` tag version to `2.73`. (I thought it's incremented twice (before and after the release).)

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
junichi11: Fixed `@since` to `2.73`.

tmysik: I see. Maybe some TODO file could be added with remaining tasks before merging into `master` (like increasing spec versions). Up to you to decide, of course.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
```diff
@@ -35,7 +35,8 @@
     "PhpVersion.PHP_71=PHP 7.1",
     "PhpVersion.PHP_72=PHP 7.2",
     "PhpVersion.PHP_73=PHP 7.3",
-    "PhpVersion.PHP_74=PHP 7.4"
+    "PhpVersion.PHP_74=PHP 7.4",
+    "PhpVersion.PHP_80=PHP 8.0"
```
tmysik, line 39: Cannot the trailing comma be added? That would make the diffs much cleaner (next time).

php/php.editor/src/org/netbeans/modules/php/editor/model/impl/ModelVisitor.java
```diff
@@ -1117,7 +1117,7 @@ public void visit(CatchClause node) {
         Variable variable = node.getVariable();
         Scope scope = modelBuilder.getCurrentScope();
         List<Expression> classNames = node.getClassNames();
-        if (scope instanceof VariableNameFactory) {
+        if (variable != null && scope instanceof VariableNameFactory) {
```
tmysik, line 1120: Not needed, `null` is not `instanceof`, ever.

tmysik:
Nice! I am ready to merge, let me know whether you want to do anything related to my comments :)

Thanks.

tmysik:
BTW not sure about the test failures...

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
junichi11: In that case, shouldn't we add `@since` at the moment? or fix again when we increase the spec version?

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
junichi11: I'll fix. Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/model/impl/ModelVisitor.java
junichi11: Probably, in this case, needed because variables(`variable` and `scope`) are different?

junichi11:
> BTW not sure about the test failures...

This is something strange... MacOS build is only one actually. I re-ran the jobs, but 2 failed MacOS builds didn't disappear. Maybe, no problem.

php/php.editor/src/org/netbeans/modules/php/editor/model/impl/ModelVisitor.java
tmysik: Sorry, those are 2 different variables... :)

junichi11: No problem :)

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
junichi11: I'll keep this as it is. Let's fix it before merging into the `master` if there is a problem.

php/php.api.phpmodule/src/org/netbeans/modules/php/api/PhpVersion.java
junichi11: Fixed: https://github.com/apache/netbeans/compare/63a46a1b8bc87584f63a0d719b73013e8a583887..31c86e7ab8e9dc3a904b71df620023a7f86ac790

tmysik:
@junichi11 Can we merge? I can see some CI failures... I would vote for merging, it is into `php80-support` anyway.

junichi11:
@tmysik Thanks a lot, Tomas!

@hectorespert I'm not sure the cause but MacOS build is strange sometimes.

PR #2267, opened 2020-07-22
[NETBEANS-4615] Display compiler arg in Gradle Source Customizer Panel

lkishalmi:


lkishalmi:
Well this one dropped out of OpenJFX work, in order to be able to check the compiler arguments.

neilcsmith-net:
I assume you meant 12.1 milestone?!  Sounds good.  Feel free to merge whenever before next Tuesday.

lkishalmi:
Well, it is pretty harmless to go into 12.1, I just did not wanted to create pressure. However this one alone just helping me to see the args. I still have to find out the right way to porcess the module-spurce-path information and feed that back to different NetBeans Java Queries.

neilcsmith-net:
@lkishalmi sorry, long day yesterday! :smile: Misread that slightly as helping to fix OpenJFX build support rather than helping to see information.  Still, feel free to merge when ready for beta2.

Is the module-source-path part likely to make 12.1 or more realistically 12.2?

lkishalmi:
Well, it seems module-source-path has not handled by the IDE at all. I have a "hack" in mind, that would add the module name to the classpath whenever the module-source-path command line arg is used. If that would be enough to make OpenJFX right, then it can happen in 12.1. If we need to really parse the argument and feed back to the IDE then it would take more time.

PR #2268, opened 2020-07-23
[NETBEANS-4621] Color highlighting is not correct when script type attribute value is "module"

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4621

- This is related to the https://issues.apache.org/jira/browse/NETBEANS-4434 

netbeans-4621


junichi11:
@neilcsmith-net This is related to https://github.com/apache/netbeans/pull/2184. So I would like to apply to `12.1` if possible.

junichi11:
Thank you for your review. OK.

PR #2269, opened 2020-07-23
[NETBEANS-4443] PHP 8.0 Support: Allow trailing comma in parameter list part

junichi11:

- https://wiki.php.net/rfc/trailing_comma_in_parameter_list
- Fix `ASTPHP5Parser.cup`
- Add unit test for the parser and formatting features

nb-php80-allow-trailing-comma-in-parameter-list


php/php.editor/src/org/netbeans/modules/php/editor/indent/FormatVisitor.java
```diff
@@ -1172,41 +1166,62 @@ public void visit(FunctionDeclaration node) {
 
     @Override
     public void visit(LambdaFunctionDeclaration node) {
```
junichi11, line 1168: I've rewritten almost of all this implementation because we should consider [psr-12](https://www.php-fig.org/psr/psr-12/). (parameters and lexical variables were not wrapped.)
Formatting problems related to closures are likely to occur in an array/parameters. So especially, I've added unit tests for in arrays.

Existing all unit tests also have passed. I just hope that there is no big problem...

tmysik: Well, we will see :)

If the tests pass, let's wait. It would be fine if there was an easy way to fix possible issues once a new version of NetBeans is released.

junichi11:
Thanks a lot for your review and your time, Tomas !

tmysik:
@junichi11 thank you for your work and time!

PR #2270, opened 2020-07-24
[NETBEANS-4617] Flush Gradle Standard output in 200ms if it stalled

lkishalmi:


neilcsmith-net:
@lkishalmi just noticed the NB12.0-u2 label on this. If considering as an NB12.0 backport, wouldn't it make sense to merge for 12.1 if this is ready and not work-in-progress?

lkishalmi:
Well, I just did not want to overload 12.1 with my changes. The code needs to be reviewed though, as I'm not sure if the technique I used is a common pattern for such a problem.

eirikbakke:
Sorry, hadn't gotten time to look at this yet. I see this patch is gradle-specific--does the same problem occur for Maven/Ant?

lkishalmi:
Well, I'm unsure of Maven, as of Ant they do a heavy by-character output processing so probably not.

PR #2273, opened 2020-07-26
[NETBEANS-4636] Make FileBuilt and SourceForBinary work with Modular Gradle Java Projects

lkishalmi:
Well @jlahoda please check these. My test project is OpenJFX with this change it seems the main inter sub-project dependencies started to work. Though during the indexing there are some assertions and NPE-s thrown.
It seems I still need to make the test sourceset dependencies work.


lkishalmi:
@neilcsmith-net you are right. This might be need to be resolved in the Java support. Though I'm not that skilled to do changes there.

neilcsmith-net:
@lkishalmi retriggering all the GitHub Actions because it seems to annoyingly not support retriggering just the one!

Assuming green, please feel free to merge this, and I'll get it in the next beta so our friends at OpenJFX can test. :smile:

PR #2275, opened 2020-07-27
Better Java file wizards with extension and implementation panels

isarthur:
A more advanced wizard for creating Java classes that let the user to 
specify interfaces and superclasses right in the wizard.


java/java.project.ui/src/org/netbeans/modules/java/project/ui/NewJavaFileWizardIterator.java
```diff
@@ -310,7 +314,12 @@ private void addRequires(FileObject createdFile, Set<String> requiredModuleNames
             createdFile = dobj.getPrimaryFile();
         } else {
             DataObject dTemplate = DataObject.find( template );                
-            DataObject dobj = dTemplate.createFromTemplate( df, targetName );
+            String superclass = (String) wiz.getProperty(SUPERCLASS);
+            String interfaces = (String) wiz.getProperty(INTERFACES);
+            Map<String, String> parameters = new HashMap<>();
+            parameters.put(SUPERCLASS, superclass);
```
JaroslavTulach, line 320: This is a change of API and should be documented as such (incremement module version, enhance `arch.xml`, note in `apichanges.xml`).

isarthur:
Added a note in apichanges and incremented spec version. I also took into account all the wizards from the Java category, but not sure how I can enhance arch.xml.

PR #2277, opened 2020-07-27
[NETBEANS-3825] nb-javac Upgrade to JDK-14

Akshay-Gupta-Oracle:
Updating the nb-javac reference link for new nb-javac supporting JDK-14 features.


arvindaprameya:
Thanks for getting this done Akshay

neilcsmith-net:
We seem to be getting a number of test failures related to nb-javac as well as sigtests.  I've tried retriggering in case of test glitch, but failed again.  Could you take a look at what's failing and see if it's a cause for concern.  Thanks.

java/java.hints/src/org/netbeans/modules/java/hints/suggestions/ExpectedTypeResolver.java
```diff
@@ -1388,4 +1391,16 @@ private static void addTypeAndReplaceMoreSpecific(CompilationInfo info, Collecti
     public List<? extends TypeMirror> visitUses(UsesTree node, Object p) {
         return null;
     }
+   
+    public List<? extends TypeMirror> visitBindingPattern(BindingPatternTree bpt, Object p) {
```
lahodaj, line 1395: I am a little surprised this compiles, but should be safe at runtime.

Should these new methods eventually have an implementation that would produce the expected type?

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertToPatternInstanceOfTest.java
```diff
@@ -140,15 +148,4 @@ public void testNoSoSimpleNameClash() throws Exception {
                               "}\n");
     }
 
-    @Override
```
lahodaj, line 143: We are giving up the ability to run tests without nb-javac on JDK<14. Possibly not that big issue as we are not running the tests in that combination anyway.

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1352,6 +1354,33 @@ protected JCModifiers modifiersOpt(JCModifiers partial) {
...
+
+                    Token peeked = S.token(1);
+
+                    if (peeked.kind == TokenKind.COMMA || peeked.kind == TokenKind.RPAREN) {
+                        nextToken();
+                        return JackpotTrees.createVariableWildcard(ctx, name);
+                    }
+                }
+            }
+                        JCTree.JCVariableDecl result = null;
```
lahodaj, line 1371: Nit: note the indent is wrong on this line.
```suggestion
            JCTree.JCVariableDecl result = null;
```

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1352,6 +1354,33 @@ protected JCModifiers modifiersOpt(JCModifiers partial) {
...
+            try {
+                Class[] paramTypes = {boolean.class, boolean.class};
+                result = (JCTree.JCVariableDecl) MethodHandles.lookup()
+                        .findSpecial(JavacParser.class, "formalParameter", MethodType.methodType(JCTree.JCVariableDecl.class, paramTypes), JackpotJavacParser.class) // NOI18N
+                        .invoke(this, lambdaParam, recordComponents);
+            } catch (Throwable e) {
+                Logger.getLogger(NBParserFactory.class.getName()).log(Level.FINE, null, e);
```
lahodaj, line 1378: I think it should be fine to just fail here, shouldn't it (the method must exist, as the current method overrides it):
```suggestion
                throw new IllegalStateException(ex);
```

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1408,7 +1437,35 @@ protected JCCatch catchClause() {
...
+                }
+            }
 
+            com.sun.tools.javac.util.List<JCTree> result = null;
+            Class[] argsType = {com.sun.tools.javac.util.Name.class, boolean.class, boolean.class};
+            try {
+                result = (com.sun.tools.javac.util.List<JCTree>) MethodHandles.lookup().findSpecial(JavacParser.class, "classOrInterfaceOrRecordBodyDeclaration", MethodType.methodType(com.sun.tools.javac.util.List.class, argsType), JackpotJavacParser.class) // NOI18N
+                        .invoke(this, className, false, false);
+            } catch (Throwable ex) {
+                Logger.getLogger(Utilities.class.getName()).log(Level.SEVERE, null, ex);
```
lahodaj, line 1464: I think it should be fine to just fail here, shouldn't it (the method must exist, as the current method overrides it):
```suggestion
                throw new IllegalStateException(ex);
```

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1427,7 +1484,7 @@ protected JCCatch catchClause() {
             }
             return super.classOrInterfaceBodyDeclaration(className, isInterface);
         }
-
+
```
lahodaj, line 1487: Nit: unnecessary change:
```suggestion

```

neilcsmith-net:
Thanks for the changes. This is still failing `ant check-sigtests`.  Is this a change in nb-javac itself that's causing this?  There are numerous lines that say `fatal error` even when this test passes, so not entirely sure the problem, but seems to be failing in `libs.javacapi`. @lahodaj any thoughts on this?

arvindaprameya:
> Thanks for the changes. This is still failing `ant check-sigtests`. Is this a change in nb-javac itself that's causing this? There are numerous lines that say `fatal error` even when this test passes, so not entirely sure the problem, but seems to be failing in `libs.javacapi`. @lahodaj any thoughts on this?

The issue has now been fixed and committed the tests also have been re-triggered, lets wait for a couple of hours to see if all of these now pass.

neilcsmith-net:
Thanks.  Think my comment overlapped with your changes somehow.  Looks like sigtests is now passing, thanks.  Just retriggered two of the others, which may be the usual random fails.  All being well I'll merge tomorrow.

arvindaprameya:
Absolutely no worries Neil, thank you for following up and ensuring the tests pass, we can now say that nb-javac is now available.

arvindaprameya:
Approved, thanks much for getting this validated.

neilcsmith-net:
@arvindaprameya sorry, looks like #2222 merging caused a conflict with this. Look like the same change.  Can you check, then rebase, squash and force push this PR to one commit and I'll merge.  Thanks.

justmeforyou (approved):

PR #2278, opened 2020-07-28
[NETBEANS-626] Invoke tableDataChanged on CountingTableModel from Ins…

Enroi:
[NETBEANS-626] Invoke tableDataChanged on CountingTableModel from InsertRecordDialog
It is my first attempt.

### DIFF
--- a/ide/db.dataview/src/org/netbeans/modules/db/dataview/table/JXTableRowHeader.java
+++ b/ide/db.dataview/src/org/netbeans/modules/db/dataview/table/JXTableRowHeader.java
@@ -32,6 +32,7 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JTable;
 import javax.swing.RowSorter;
+import javax.swing.SwingUtilities;
 import javax.swing.event.RowSorterEvent;
 import javax.swing.event.RowSorterListener;
 import javax.swing.event.TableModelEvent;
@@ -187,8 +188,10 @@ public final class JXTableRowHeader extends JComponent {
         }
 
         private void tableDataChanged() {
-            setCount(backingTable.getRowCount());
-            fireTableDataChanged();
+            SwingUtilities.invokeLater(() -> {
+                setCount(backingTable.getRowCount());
+                fireTableDataChanged();
+            });
         }
 
         @Override


Enroi:
Hello.
Is new version better?

matthiasblaesing:
@Enroi it is good, that it is a smaller change. But it still involves the UI part manually updating the model part. The problem is still the coupling of the the two datamodels. The counting model relies on the UI being updated. So I suggest to modify this method:

https://github.com/apache/netbeans/blob/709587a5a001f78707c32a32c4272a21ecc85fca/ide/db.dataview/src/org/netbeans/modules/db/dataview/table/JXTableRowHeader.java#L189-L192

When this method is invoked, all events on the EDT have to have been processed. Only when that is true, the UI model of the backing table is in a stable state. `SwingUtilities#invokeLater` is a candidate to handle this, as it will enqueue the new action after the existing events.

Enroi:
> @Enroi it is good, that it is a smaller change. But it still involves the UI part manually updating the model part. The problem is still the coupling of the the two datamodels. The counting model relies on the UI being updated. So I suggest to modify this method:
> 
> https://github.com/apache/netbeans/blob/709587a5a001f78707c32a32c4272a21ecc85fca/ide/db.dataview/src/org/netbeans/modules/db/dataview/table/JXTableRowHeader.java#L189-L192
> 
> When this method is invoked, all events on the EDT have to have been processed. Only when that is true, the UI model of the backing table is in a stable state. `SwingUtilities#invokeLater` is a candidate to handle this, as it will enqueue the new action after the existing events.

Enroi:
Hello, sorry for close your reply. I did not understand what this link does.
About you comment. 
You have written "SwingUtilities#invokeLater is a candidate to handle this" but I can not understand how to use SwingUtilities#invokeLater here.
I tested such code, but it does not work
`private void tableDataChanged() {
            setCount(backingTable.getRowCount());
            SwingUtilities.invokeLater(new Runnable() {
                @Override
                public void run() {
                    fireTableDataChanged();
                }
            });
        }`
Do you propose to replace every invoke of tableDataChanged with SwingUtilities.invokeLater(... tableDataChanged() ... ) ?

matthiasblaesing:
The problem is that the row count of the JTable is updated, _after_ tableDataChanged is reached. So the method reads the row count at the wrong moment. A fix needs to ensure all pending events have happened before the _read_ is done. The read is backingTable.getRowCount() - if that is invoked directly and only the fireTableDataChanged is dispatched into the invokeLater, you are still reading the wrong value.

The whole method body of tableDataChanged most be moved into the invokeLater.

Enroi:
Hello.
It works. It is surprise for me.

matthiasblaesing:
Tested the change and it works. Please squash the commits into one.

For the record: I assume you read the file header and you donate this change to the ASF. If this is not the case, please indicate so, the change can then not be merged.

Enroi:
Can I create new MergeRequest? I can not find how to squash GIT commits in GitHub.

matthiasblaesing:
You can update the PR in place. Rewrite history on your system and then force push the change into the github branch. For the "how to" of rewriting history, please see: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Enroi:
I have found way how to squash and push my changes. 
Now this poll request has only 1 commit.

matthiasblaesing:
Lets get this. Thank you.

matthiasblaesing (approved):

PR #2279, opened 2020-07-29
Improvements to the Java LSP Server

jlahoda:
-proper start/end positions for definition lookup
-fixing problems with zip file caches
-waiting for shutdown command, instead of looping forever after starting the server
-fixes "Add new variable" and "Add import" should work
-JDK 14 is a requirement for the server (does not work well on JDK 11)
-improvements to code completion


arvindaprameya:
Approved, Thanks so much for doing this Jan.

PR #2280, opened 2020-07-29
[NETBEANS-4443] PHP 8.0 Support: throw Expression part

junichi11:

- https://wiki.php.net/rfc/throw_expression
- Convert the `ThrowStatement` to the `ThrowExpression`
- Add unit tests for the parser, the formatter, and the comment generator

nb-php80-throw-expression


php/php.editor/src/org/netbeans/modules/php/editor/verification/UnreachableStatementHint.java
```diff
@@ -246,12 +247,23 @@ public void visit(ContinueStatement node) {
...
 
         @Override
-        public void visit(ThrowStatement node) {
+        public void visit(ExpressionStatement node) {
             if (CancelSupport.getDefault().isCancelled()) {
                 return;
             }
             super.visit(node);
-            processLastStatement(node);
+            // PHP 8.0: throw statement -> throw expression
```
tmysik, line 255: Usually, it is a good idea to add also a number of PR and/or issue.

junichi11: Will do it. Thank you!

tmysik: Great, thanks!

PR #2281, opened 2020-07-30
Netbeans license headers cleanup check-in #1

BradWalker:
Netbeans license headers cleanup check-in #1.. The first of several..
    
    Add missing ALv2 license headers to form files.

### DIFF
--- a/cnd/cnd.makeproject.ui/src/org/netbeans/modules/cnd/makeproject/ui/launchers/actions/LaunchersPanel.form
+++ b/cnd/cnd.makeproject.ui/src/org/netbeans/modules/cnd/makeproject/ui/launchers/actions/LaunchersPanel.form
@@ -1,5 +1,26 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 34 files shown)


junichi11:
@BradWalker `#xx` means issue/pr number in the GitHub. So, should use another one.

junichi11:
@matthiasblaesing 

> I don't think the `#1` in the commit message is a deal breaker. That github decided to try to gather information from plain text insteaf of structured information is a problem of github, not Brad. Try to add some random "@" into messages and a random number of people get CCed.

I didn't mean that. Actually, I approved this. I just would like to avoid unnecessary notification/link. I meant I suggest use another one (from next time; It's OK with me to merge this as it is) if we need not use `#xx` absolutely. It's a problem of GitHub, but it's better if we can avoid that, I think.

BradWalker:
Hey @junichi11 and @matthiasblaesing, thanks for the feedback.. I would NEVER have known that GitHub was trying to parse my pull request for notification.. Thanks.. I'll be more careful next time.. 

Can @junichi11 commit my changes for me?

Thanks to both of you!

junichi11:
@BradWalker Never mind :) Merging.

matthiasblaesing:
Thanks to the both of you. :+1:

PR #2282, opened 2020-07-30
[NETBEANS-3990] Sealed color and formatting

Akshay-Gupta-Oracle:
JDK-14 sealed support for IDE with color for keyword sealed,non-sealed and permits, and auto formatting for classes with sealed.


java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
```diff
@@ -871,7 +871,18 @@ public Void visitClass(ClassTree tree, Void p) {
             }
             scan(tree.getExtendsClause(), null);
             scan(tree.getImplementsClause(), null);
-
+            try{
+            while(!tl.currentToken().text().toString().equals("{")) {
```
junichi11, line 875: Although I'm not an expert of this area, maybe, if you would like to check whether text of a token is the same as given one, should use `org.netbeans.api.lexer.TokenUtilities.textEquals()` instead of `toString().equals` (e.g. `!TokenUtilities.textEquals(tl.currentToken().text(), "{")`), then should add `// NOI18N` to strings.
Please reformat your changes. (the changes below, too)

junichi11: It doesn't seem to be reformatted yet. Please add `// NOI18N` for strings that are not expected translation.

arvindaprameya:
Planned for 12.1

neilcsmith-net:
Is this JDK 14 not 15?!  I was just about to look at merging master to release branch for hopefully final beta.  What's the rush on this one?

arvindaprameya:
This change will help the new feature "Sealed types" to work well when Open JDK 15 is released in Sep 2020

neilcsmith-net:
@arvindaprameya yes, that's my understanding, and the first release to officially support JDK 15 is meant to be 12.2.  This missed freeze on July 15th and still has open reviews, so I'm inclined to bump this to 12.2.  In my opinion, we need to get language features, including the relevant nb-javac support, in much earlier in the release cycle and well before freeze to ensure it's adequately tested.

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/ColoringAttributes.java
```diff
@@ -30,7 +30,9 @@
 public enum ColoringAttributes {
 
     UNUSED,
-
+    SEALED,
```
lahodaj, line 33: These new constants don't seem to be needed, right? The keywords should have attribute KEYWORD, I think, and that should be enough to make them highlighted as a keyword.

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
```diff
@@ -871,7 +872,18 @@ public Void visitClass(ClassTree tree, Void p) {
             }
             scan(tree.getExtendsClause(), null);
             scan(tree.getImplementsClause(), null);
-
+            try {
+                while (!TokenUtilities.textEquals(tl.currentToken().text(), "{")) {// NOI18N
```
lahodaj, line 876: My suggestion here would be to:
1. read the Permitted subclasses
2. if non-null/non-empty, do simply firstIdentifierToken("permits"), and add the token to contextKeywords
3. run scan(<permitted>, null)
(so that the classes inside the permits clause are highlighted).

java/java.lexer/src/org/netbeans/api/java/lexer/JavaTokenId.java
```diff
@@ -133,6 +133,9 @@
     TRUE("true", "literal"),
     FALSE("false", "literal"),
     NULL("null", "literal"),
+    PERMITS("permits", "keyword"),
```
lahodaj, line 136: These new constants do not seem to be really used, right? I.e. the lexer will never produce a token with this kind?

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -954,6 +955,11 @@ public Boolean visitClass(ClassTree node, Void p) {
                         wrapToken(cs.wrapExtendsImplementsKeyword(), 1, id == INTERFACE ? EXTENDS : IMPLEMENTS);
                         wrapList(cs.wrapExtendsImplementsList(), cs.alignMultilineImplements(), true, COMMA, impls);
                     }
+                    List<? extends Tree> perms = TreeShims.getPermits(node);
+                    if (perms != null && !perms.isEmpty()) {
+                        wrapToken(cs.wrapExtendsImplementsKeyword(), 1, PERMITS);
```
lahodaj, line 960: I think it is sensible to use the same settings as for wrapping extends/implements. We may consider tweaking the UI to also mention permits (this applies to both wrapExtendsImplementsKeyword and wrapExtendsImplementsList).

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
```diff
@@ -953,6 +965,25 @@ public Void scan(Tree tree, Void p) {
                 handlePossibleIdentifier(tp, true, info.getTrees().getElement(tp));
                 tl.moveToOffset(sourcePositions.getEndPosition(getCurrentPath().getCompilationUnit(), TreeShims.getBindingPatternType(tree)));
                 firstIdentifier(tp, TreeShims.getBinding(tree).toString());
+            } else if (tree != null && tree.getKind().equals(Kind.MODIFIERS)) {
```
lahodaj, line 968: Here, I would suggest to create a new method "visitModifiers", and put this code into it. Also, it will possibly be a little bit more complex to handle, as "sealed" and "non-sealed" may be intermixed in other modifiers and annotations. I guess I would suggest to avoid doing toString() on the tree - what will happen if the modifiers contain something like:
@SuppressWarnings("non-sealed")
?

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -3524,6 +3530,10 @@ private JavaTokenId accept(JavaTokenId first, JavaTokenId... rest) {
                             contains = true;
                             break;
                         }
+                        if (TokenUtilities.textEquals(tokens.token().text(), "non") && tokens.moveNext() && TokenUtilities.textEquals(tokens.token().text(), "-") && tokens.moveNext() && TokenUtilities.textEquals(tokens.token().text(), "sealed")) {// NOI18N
```
lahodaj, line 3533: What happens if "non" is not followed by "-sealed"?

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -1585,7 +1591,7 @@ public Boolean visitModifiers(ModifiersTree node, Void p) {
                 int lblti = lastBlankLinesTokenIndex;
                 Diff lbld = lastBlankLinesDiff;
                 id = accept(PRIVATE, PROTECTED, PUBLIC, STATIC, DEFAULT, TRANSIENT, FINAL,
-                        ABSTRACT, NATIVE, VOLATILE, SYNCHRONIZED, STRICTFP, AT);
+                        ABSTRACT, NATIVE, VOLATILE, SYNCHRONIZED, STRICTFP, AT, SEALED, NONSEALED);
```
lahodaj, line 1594: I see the token ids are useful here - but I would suggest to add a way to do without them. Enhancing API just to simplify one place a little bit will hurt long-term maintenance.

lahodaj (changes requested):
Added some comments. I would also suggest to add tests for the sealed/non-sealed context keywords (see DetectorTest).

lkishalmi:
@lahodaj @jlahoda would any of you approve this change?

lkishalmi:
@Akshay-Gupta-Oracle Could you update this PR to resolve the merge conflict?

arvindaprameya:
Thank you for addressing all the comments, approved. Please resolve the conflicts in master branch as pointed by @lkishalmi

Akshay-Gupta-Oracle:
@lkishalmi  I've resolve the conflict with master. But after resolving I had to merge master into this branch, so now I can't rebase my local branch and squash the commits. may need help there.

PR #2284, opened 2020-07-31
[NETBEANS-4664] - Upgrade JDBC PostgreSQL from 42.2.10 to 42.2.16

pepness:
Notes:
- Fix security vulnerability CVE-2020-13692 
- Minor bug fixes

[Releases Notes](https://jdbc.postgresql.org/documentation/changelog.html)

NetBeans Testing:

- Full build done
- Started NetBeans and ensure the log didn't have any errors or new warnings
- Checked that the file had been downloaded and referenced inside the Ant Library Manager
`jar:nbinst://org.netbeans.modules.db.drivers/modules/ext/postgresql-42.2.16.jar!/`

### DIFF
--- a/ide/db.drivers/external/binaries-list
+++ b/ide/db.drivers/external/binaries-list
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-338AF492789FD198DBD52735A22B1946030ECC96 org.postgresql:postgresql:42.2.10
+A9EE12F737BD5DC7D046E4C065E391D38D6A3CFC org.postgresql:postgresql:42.2.16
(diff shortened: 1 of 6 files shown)


pepness:
Reversed the specification version bump.

neilcsmith-net:
I've only glanced at this.  Is CVE-2020-13692 something that affects the IDE, and if so, does it need to be in 12.1 and backported for a 12.0 update?  If that answer should be done privately, please do so!

pepness:
This would affect Ant projects that are created in NetBeans and rely on the NetBeans Library. 

In my opinion keeping the libraries up to date gives better quality to Ant based projects, there are cases that people can only use the libraries that NetBeans provide because they do not have access to the internet or the access is restricted, e.g. schools, banking institutions and government dependencies.

pepness:
Upgrade to version 42.2.16 and re-test everything.

PR #2286, opened 2020-08-01
[NETBEANS-4667] Fix broken npm search

SiddheshRane:
IDE npm search (Web project > properties > npm > add) is broken due to output format change in npm.
This commit uses npm --parseable flag to fix and simplify output parsing.
Tested with npm 6.14.4

### DIFF
--- a/webcommon/javascript.nodejs/src/org/netbeans/modules/javascript/nodejs/exec/NpmExecutable.java
+++ b/webcommon/javascript.nodejs/src/org/netbeans/modules/javascript/nodejs/exec/NpmExecutable.java
@@ -215,6 +215,7 @@ public class NpmExecutable {
         List<String> params = new ArrayList<>();
         params.add("search"); // NOI18N
         params.add("--long"); // NOI18N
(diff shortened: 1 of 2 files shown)


Chris2011:
Cool, finally :). I had a look into the code, you removed a lot. So is this fix a general fix and will this work for those who not uses npm >= 6?

Chris2011:
Also we should get rid of those hardcoded flags. But this is another task.

SiddheshRane:
The `--parseable` flag was added to npm search around 4 years ago. Before that it was named  `--tab-seperated`. So I think it covers a lot of npm versions that people are actually using. 

This is a general fix for scraping cli output as was meant to be done.

Chris2011:
Ok great. Thx for the info :)

Chris2011:
From my pov it looks promising :)

matthiasblaesing:
Thank you - change looks sane, unittests come back clean, author information looks good.

@SiddheshRane I assume, that you agree that the changeset is donated to the ASF, if that is not the case, please say so, the merge will then be reverted.

PR #2294, opened 2020-08-04
Adding support for clangd to cpplite

jlahoda:


matthiasblaesing:
A quick observation: I noticed that in the typescript editor the completion for paths in import statements did not work correctly for me. This manifested in the selector opening, but when I got on typing, I got a nullpointer exception. This is gone with this change, so if not anything else, this at least fixes a bug for me.

@jlahoda do you think this is stable enough to get into 12.1?

jlahoda:
@vieiro 

> Just a quick question (don't have to do a thorough review, I'm afraid). Do you think this static final is going to cause threading/memory problems?
> 
> ` private static final TextDocumentSyncServerCapabilityHandler h = new TextDocumentSyncServerCapabilityHandler();`
> `

I hope not. The instance was no free-able anyway (it was added as a listener to EditorRegistry and never removed). And final fields should be safe to access from any thread (static fields are initialized under a classloading lock (non-static final fields are also accessible from anywhere after the constructor finishes, AFAIK)).

jlahoda:
@matthiasblaesing - I wouldn't mind if this got to 12.1, but I can live with it not getting there. We can separate the fix for the NPE if needed.

vieiro:
@jlahoda Full speed ahead then!

neilcsmith-net:
@jlahoda @matthiasblaesing fine with me! If it's working, feel free to merge so we can get in next beta - waiting for a fixed nb-javac anyway.

neilcsmith-net:
Added 12.1 milestone to help tracking. If you decide not to merge, please move to 12.2.

PR #2296, opened 2020-08-05
Fixing removal of a double-quote inside a text block.

jlahoda:
Consider code like:
"""
"|
"""

Where '|' is the caret position. Now press backspace. This will delete the ", but also the next character (newline in this case). This makes editing inside the text block quite annoying.

Not sure if this breaks some other text block editing scenario - tests are passing, but may be cases not covered by tests.


neilcsmith-net:
Seems a sensible fix! Do you want to merge for next 12.1 beta?

neilcsmith-net:
Added 12.1 milestone to help tracking.  If you decide not to merge, please move to 12.2.

PR #2298, opened 2020-08-06
Quicksearch API allows to check for cancel.

sdedic:
I've been working with a Quicksearch provider and I noticed that while the Provider can check whether the search query was cancelled when calling `addResult` to the response, there's no way to check if the search was cancelled if the Provider is unable to find any item. In my case, the SearcProvider continues to search even if a new query has been created (provided the Provider still did not found anything).

This simple addition would allow to cancel even such empty-result searches.

### DIFF
--- a/platform/spi.quicksearch/apichanges.xml
+++ b/platform/spi.quicksearch/apichanges.xml
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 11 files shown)


sdedic:
Yes, the request is, semantically ;) but since there was already an indicator on the `response`, I added the flag method there. I though that an implementation would acquire data from the request (possibly leaving it behind), but I needs to have the response handy in order to accumulate results.

I can move the method, if you people  think it is really unnatural :)

PR #2300, opened 2020-08-07
[NETBEANS-4681] Fixed java record rename & move refactoring issue

singh-akhilesh:
https://issues.apache.org/jira/browse/NETBEANS-4681


singh-akhilesh:
created another PR 2301

PR #2272, opened 2020-07-26
[NETBEANS-4635] - Add JakartaEE 8 Platform module

pepness:
This is an improvement to the [PR-2023](https://github.com/apache/netbeans/pull/2023) from Josh Juneau. I used the `j2ee.platform` module as a reference.

Notes:
- Add module `jakartaee8.platform` that contains the Javadoc
- Fix JakartaEE integration with the NetBeans library (Classpath & Javadoc)
  from both jakartaee-api & jakartaee-web-api libraries.

Classpath - Before
NetBeans-1
Classpath - After
NetBeans-1-after
Javadoc - Before
NetBeans-2
Javadoc - After
Javadoc-After

PR #2302, opened 2020-08-07
Fix library definitions causing warnings at startup

matthiasblaesing:
While running a custom build of NetBeans derived from recent master I noticed, that several warnings popped up relating to invalid library definitions. I ran through the warnings and fixed them.

The individual commits hold explanations for each fix, so please have a look at them - or ask.


matthiasblaesing:
@jlahoda @Akshay-Gupta-Oracle could you please have a look at the nbjavac change (https://github.com/apache/netbeans/commit/b2eedc75b29f60d19690b095874af3ea7f222d4d)
@hectorespert  could you please have a look at the spring framework change (https://github.com/apache/netbeans/pull/2302/commits/5c41416da1e32b9607e05a8d3a8c03524adfd422)
@juneau001 could you please have a look at the change for jakartaee? (https://github.com/apache/netbeans/pull/2302/commits/6e7ce2307e36187db42331e7085c10ceb9ff10ba)

neilcsmith-net:
@matthiasblaesing you added me as reviewer but not the 12.1 milestone - you thinking this should / needs to be done for current release?

There are multiple test failures. Have retriggered everything in case it's an infrastructure issue.  If fails again, please take a look.

matthiasblaesing:
The testfailures are infrastructure problems (that is the main problem with our current setup a belief, that checks get only red when a failure was introduced and an unstable test setup).
The library definitions are useless and if someone adds them to their project it will result in the project being marked as broken. So I added it to the 12.1 milestone.

hectorespert:
> @hectorespert could you please have a look at the spring framework change ([5c41416](https://github.com/apache/netbeans/commit/5c41416da1e32b9607e05a8d3a8c03524adfd422))

It's ok to me. But I don't kown if this artifacts are usefull. Seems that are sources and the pom transformed into jars.

matthiasblaesing:
Ok - there is a real problem in ergonomics (did I say, that I hate it?) - I'll look into it now.

neilcsmith-net:
Firstly, please take a look at #2272 - that is scheduled to be merged before next beta, and I think conflicts with this?  If it does and you want to merge that then rebase this, go ahead. 

> Ok - there is a real problem in ergonomics (did I say, that I hate it?) - I'll look into it now.

Did I say it too?! :smile:  I also wish the nb-javac and JavaFX installation support didn't utilise it.

> that is the main problem with our current setup a belief, that checks get only red when a failure was introduced and an unstable test setup.

Some of the tests are unstable, and network issues will sometimes cause failure too.  But they should go green with a bit of kicking.  Nothing that won't go green is being merged - it's generally a sign that something is wrong.  The lack of ability to retrigger single tests on GitHub is a pain - not a fan of moving too much away from Travis with that problem.

matthiasblaesing:
A crap - yes. I'm surprised we consider new modules in beta, but so be it. That PR is indeed a part of the fix. Once that is merged, I'll update this accordingly. I have change requests for #2272 which I'll add shortly.

neilcsmith-net:
Only considered because it seems a fairly minimal addition to Jakarta support already added.  Not an area I work with - so if there are concerns about it, it can also wait for 12.2!

jlahoda:
Regarding nb-javac, looks reasonable to me. Not only the library is not really working, but it is probably also not particularly useful (anymore?). Thanks for the cleanup!

matthiasblaesing:
@neilcsmith-net I dropped the jakartaee change from this PR as it is better covered in #2272  as you pointed out. The initial build failure of the nbjavac library issue was caused by a dangeling reference in the manifest to the layer file I removed, this was fixed. Both remamaining commits are reviewed. Ok to merge for 12.1?

neilcsmith-net:
@matthiasblaesing OK, feel free to merge. :+1:

PR #2305, opened 2020-08-09
[NETBEANS-3693] Create only one instance of javac

Akshay-Gupta-Oracle:
Use only one instance of javac to parse multiple files when multiple java files needed to be parsed when netbeans is running without nb-javac.


java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -424,9 +434,28 @@ private void parseImpl(
...
+                        javacTask = ciImpl.getJavacTask();
+                        oldParsedTrees=ciImpl.getParsedTrees();
+                    } else {
+                        diagnosticListener = null;
+                        javacTask = null;
+                    }
                     ciImpl = createCurrentInfo(this, file, root, snapshot,
-                        sequentialParsing == null || ciImpl == null ? null : ciImpl.getJavacTask(),
-                        sequentialParsing == null || ciImpl == null ? null : ciImpl.getDiagnosticListener());
+                        javacTask,
```
lahodaj, line 457: I suspect this will disable "sequential" processing when nb-javac is installed, right? Or do I miss something here?

Akshay-Gupta-Oracle: This is handled above on the if-else conditions starting at line 440. If nb-javac is installed then sequential parsing will not be null and remaining code will be same as of before this PR.

arvindaprameya:
Approved, thanks for doing this Akshay.

java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
Akshay-Gupta-Oracle: Is this OK @jlahoda, can this PR be merged now??

lahodaj: I note that the if on line 448 is also checking oneInstanceJava. So it seems to me that is sequentialParsing != null and oneInstanceJava == false, the it will always fall to the else section and go file-by-file. Or what do I miss?

Akshay-Gupta-Oracle: You're correct @jlahoda removed the oninstance variable form second if.
Now this PR is final

lahodaj:
Looks good to me.

PR #2306, opened 2020-08-10
[NETBEANS-3409] Missing popup menu for sql editor tabs

matthiasblaesing:
The Filesystems API compatibility module provides implementations for
Filesystem#getActions for several filesystems. Most filesystem
implementations are covered by the implementation in
AbstractFileSystemCompat, but this does not cover implementations
directly inheriting from Filesystem.

One such filesystem is BinaryFS.

In the SQL Editor the command window function (temp files for SQL
execution) is backed by a file in the SystemFileSystem, which delegates
in the end to BinaryFS:

SystemFileSystem-> 
  ModuleLayeredFileSystem ->
    BinaryFS

As BinaryFS does not implement getActions and no patch is provided for
this class, an AbstractMethodError is raised.

Instead of bailing out with an Error, the compatibility implementation
now returns an empty array to satisfy the Filesystem#getActions contract.

### DIFF
--- a/platform/openide.filesystems.compat8/src/org/openide/filesystems/FileSystemCompat.java
+++ b/platform/openide.filesystems.compat8/src/org/openide/filesystems/FileSystemCompat.java
@@ -43,6 +43,9 @@ import org.openide.util.actions.SystemAction;
  */
 @PatchFor(FileSystem.class)
 public abstract class FileSystemCompat {
+    /** system actions for this FS */
+    private static final SystemAction[] NO_SYSTEM_ACTIONS = new SystemAction[0];
+
     /** Property name giving capabilities state. @deprecated No more capabilities. */
     static final String PROP_CAPABILITIES = "capabilities"; // NOI18N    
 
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@JaroslavTulach I would appretiate a look at this. The title might sound as if this is a problem somewhere in the db support, but it can be traced back into the filesystem API (the compatibility module) and from my POV is the result of an incomplete patch implementation for the the extracted `Filesystem#getActions` method.

sdedic:
Very nice catch! Indeed, the Compat class was (conceptually) broken from the initial refactor. The issue was hidden until 9f47905f7 which removed the (now obsolete) `getActions()` implementation from the `BinaryFS` ... and even after until some old-API client was REALLY loaded at runtime.

So it seems that as a rule of thumb, the compat class should have ***no*** abstract methods in it. The `@PatchFor` processor could enforce that.

matthiasblaesing:
@neilcsmith-net this should be considered for inclusion into 12.1 as it is a regression introduced in the 11 series.

@sdedic thank you for the history analysis, yes the pretty much explains the issue and it also explain why I would have sworn, that the menu worked in past versions.

@JaroslavTulach @sdedic thanks for the quick review both of you.

PR #2309, opened 2020-08-13
Various improvements inspired by working on VSCode extension

JaroslavTulach:
Recently I was working on LSP VSCode extension. Before I could work on the polyglot debugging using NetBeans Truffle Debugger implementation, I had to polish the system to:

* make the TypeScript debuggable
* make the Java code debuggable
* make it run in headless mode
* improve [README](https://github.com/apache/netbeans/blob/1c2668aac2fa72d67a88e5f6c6cb6676e4bcedd8/java/java.lsp.server/vscode/README.md)
* and few more.

My debugging improvements aren't ready yet, however I believe the existing VSCode extension is going to benefit from my changes even without debugging. I propose to integrate these changes into Apache NetBeans 12.2


matthiasblaesing:
We are still in feature freeze and this does not look like Neil reviewed. 

Am 18. August 2020 17:51:20 MESZ schrieb Jaroslav Tulach <notifications@github.com>:
>Merged #2309 into master.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/apache/netbeans/pull/2309#event-3667008101

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

JaroslavTulach:
Opps. I haven't realized that. I am sorry.

matthiasblaesing:
Maybe relevant for @neilcsmith-net does the change set affect the 'normal' NetBeans build?

Could it impact the release?


Am 18. August 2020 17:57:27 MESZ schrieb Jaroslav Tulach <notifications@github.com>:
>Opps. I haven't realized that. I am sorry.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/apache/netbeans/pull/2309#issuecomment-675564653

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

neilcsmith-net:
@JaroslavTulach OK. Are you not following dev@?  (there was a message from @geertjanw about another PR we were hoping might be mergeable, but too late now for beta3).  Luckily this just missed release branch syncing, so might be OK to stay in if beta 3 becomes 12.1 release.  OTOH, if we need to merge fixes into master then this will need reverting out first.  I'll update here.

java/java.lsp.server/vscode/.vscode/tasks.json
```diff
@@ -0,0 +1,20 @@
+// See https://go.microsoft.com/fwlink/?LinkId=733558
+// for the documentation about the tasks.json format
```
matthiasblaesing, line 2: I noticed this in other JSON files aswell. Are these files from upstream somewhere or authored? This is not valid JSON (there is no comment in JSON (rfc7159)).

java/maven/src/org/netbeans/modules/maven/api/execute/RunUtils.java
```diff
@@ -218,4 +219,10 @@ public static boolean hasTestCompileOnSaveEnabled(RunConfig config) {
         return false;
     }
 
+    private static void invokeLaterWithUI(Runnable runnable) {
+        if (GraphicsEnvironment.isHeadless()) {
+            return;
+        }
```
matthiasblaesing, line 225: This looks a bit strange - is this intended, that this invocation silently drops the execution of the runnable? I would be a bit surprised if trying to debug.

neilcsmith-net:
> From my POV we should back it out if we further changes for 12.1

@matthiasblaesing yes, that is what I meant by my comment - I will definitely back it out _if_ we need changes, but at the moment I'm hopeful that the release branch is OK for the final release.

This is assuming I'm correct at quick glance that nothing here will change sig file generation @JaroslavTulach ?

java/java.lsp.server/vscode/.vscode/tasks.json
JaroslavTulach: This is what one gets when generating a skeletal VSCode project. Looks like these `.json` files accept comments, surprising - I might have used that instead of 2413e8b

java/maven/src/org/netbeans/modules/maven/api/execute/RunUtils.java
JaroslavTulach: Yes, the method `invokeLaterWithUI` is used to change cursor on main `JFrame` of the application. That yields an exception in headless mode. Not mangling the cursor shape - e.g. skipping the runnable - is exactly what is needed in headless mode.

JaroslavTulach:
> @JaroslavTulach OK. Are you not following dev@?

I am sorry, I didn't read dev@ recently. I was completely focused on VSCode and elsewhere. Once all this PR checks got green, I integrated without realizing the master is supposed to be frozen. I should be more careful. This is a second time I made such mistake, I should learn my lesson.

Possibly I was influenced by considering my change just a _robustness fix_ that cannot break anything. Running the diff and eliminating all VSCode specific changes:
```bash
$ curl 'https://patch-diff.githubusercontent.com/raw/apache/netbeans/pull/2309.diff' \
  | grep ^diff | grep -v java/java.lsp.server | grep -v nbbuild/misc/prepare-bundles \
  | cut -f3 -d " "
a/ide/spi.debugger.ui/src/org/netbeans/modules/debugger/ui/DebuggerManagerListener.java
a/java/debugger.jpda.truffle/manifest.mf
a/java/debugger.jpda.truffle/nbproject/project.xml
a/java/maven/src/org/netbeans/modules/maven/api/execute/RunUtils.java
a/java/maven/src/org/netbeans/modules/maven/execute/cmd/ExecutionEventObject.java
a/nbbuild/build.xml
a/platform/core.windows/src/org/netbeans/core/windows/services/DialogDisplayerImpl.java
```
leaves us with `isHeadless()` checks:
* [ide/spi.debugger.ui](https://github.com/apache/netbeans/pull/2309/files#diff-b9c79046b9548364065efd8dd282089d)
* [maven](https://github.com/apache/netbeans/pull/2309/files#diff-f038edf85db71cd67a5b2a356647847d)
* [core.windows](https://github.com/apache/netbeans/pull/2309/files#diff-6aff1792e1913ea7946414fd9420635b)

these checks shall not influence the release as NetBeans IDE isn't running in headless mode. Then there is another robustness check:
* [IOFolding.isSupported](https://github.com/apache/netbeans/pull/2309/files#diff-0fc693f406ad5eae6e15b6063d7bdba0)

it shall not influence NetBeans IDE as NetBeans IDE is using `core.output2` implementation of output window which supports folding. One more robustness check is done in buildsystem:
* [check Packages](https://github.com/apache/netbeans/pull/2309/files#diff-fa67ba03971f6d21c0ff2493bf5d922f)

that only fixes problem of running `ant tryme` on GraalVM, other JVMs continue to behave the same. And then there is the change of dependencies in `debugger.jpda.truffle` module which replaces strict dependency on `java.kit` with a recommendation. That's the only change that changes some behavior in the NetBeans IDE - however as the `java.kit` is present in the NetBeans IDE the change shall be idempotent.

With the robustness fixes in master my [debugging in VSCode branch](https://github.com/apache/netbeans/compare/master...JaroslavTulach:java-lsp-server-debugging-attempt2) is much cleaner. Nice, but poor excuse for the organizational problems my integration causes just before release of 12.1. Once again, please accept my apology.

neilcsmith-net:
@JaroslavTulach no worries!  You're not the only one. :smile:  Both Eric and I have used a [NOTICE] tag on dev@ for merge window opening and closing, etc.  If following all of dev@ gets a bit much (which it does!) might be worth filtering on that tag and a few others like that.

java/maven/src/org/netbeans/modules/maven/api/execute/RunUtils.java
matthiasblaesing: Understood - when I first read the method name, I assumed, that is intended to invoke the Runnable on the EDT. That made me wonder why `SwingUtilities#invokeLater` was not used, which led to me reading the method.

java/java.lsp.server/vscode/.vscode/tasks.json
matthiasblaesing: Ok - for me the important part is, that VSCode generates it so. I was just wondering why invalid JSON was created, but it seems web developers again ignored RFCs and invented a new format.

neilcsmith-net:
OK, @matthiasblaesing @JaroslavTulach it looks like we need to merge #2325 for 12.1, which means we may need one more beta - so, do I revert this or keep it in?

JaroslavTulach:
I have just built NetBeans from most recent `master` and started with fresh user directory. I can debug JavaScript in a Maven project that launches JavaScript via `ScriptEngineManager` and I can debug Node.js project on top of GraalVM that calls into Java.

I am fine with leaving my accidental changes in. If you or others feel different, feel free to revert. Once again sorry for the mess.

neilcsmith-net:
Thanks, @JaroslavTulach that's fine with me. Will be in.

PR #2313, opened 2020-08-14
[NETBEANS-4715] Make GlassFish Report Jakarta EE 8.0 Profile Support

lkishalmi:
@juneau001 please check it as well. I've removed the GlassFishV5_1.xml as that was identical to the GlassFishV5.xml.


geertjanw:
But this seems to be changes only for GlassFish, what about the other servers (for none of them when selected could I see Jakarta EE being available in the dropdown below that).

lkishalmi:
Well, what are those other servers? I might can look at them as well.

geertjanw:
I was trying with Payara.

lkishalmi:
Well the current Payara tooling does not report JakartaEE 8 compatibility it's configuration. I'm not really aware of the version compatibility matrix of Payara. the latest we have support is 5.192 on the JakartaEE sute says JakartaEE 8.0 is supported in Payara 5.193.1 probably we have to add the descriptor to that version as well. Though I'd leave that for the Payara folks. They know what version is the version.

juneau001:
Thanks @lkishalmi this looks great!  I have cloned and tested locally with success.  I agree that the Payara folks should probably handle the Payara compatibility.  NetBeans has historically provided compatibility with GlassFish out of the box.

PR #2314, opened 2020-08-15
Switch http://repo1.maven.org to HTTPS

lkishalmi:
This is just a simple search and replace in the whole project, may fix some tests/ functionalities.


lkishalmi:
Fixed the commit message. The change is quite trivial it seems the CI liked it. @neilcsmith-net it is on you whether you would like to include this on 12.1 or not. What I see it shouldn't do any wrong as http url calls would return 501, it shall improve the code, but those are mostly tests, documentation. The real impact is on the GlassFish, Payara support for some javadoc and Maven support for checking if the central repository can be added or not.

PR #2322, opened 2020-08-22
Avoiding use of Nashorn (removed in JDK 15) in J2SE multi-module proj…

jlahoda:
…ects by re-writing the JavaScript code into Java.

Testing would be appreciated!


lkishalmi (approved):
Well, interesting approach.
Though just skimmed through, if it works, let's have this integrated.

PR #2326, opened 2020-08-23
[NETBEANS-4627] FlatLaf: fix wrong background color of search result table in Find Tasks Editor

DevCharly:
Simple fix for https://issues.apache.org/jira/browse/NETBEANS-4627

Before:

image

After:

image


neilcsmith-net:
Need to roll another beta for NPE fix, so inclined to think this might be a good, safe fix to squeeze in?

lkishalmi:
Property file change.It's safe.

neilcsmith-net:
@lkishalmi yes, thanks, why considering. My comment was meant more as a heads up that we could squeeze this in because we need another beta anyway - didn't really need the question mark. :smile:

PR #2327, opened 2020-08-23
[NETBEANS-3428] FlatLaf: fix selection background color in output view

DevCharly:
Simple fix for selection background in output view, which was nearly unreadable. 
Used same selection background colors as in editor.

Before:

image

After:

image

Before:

image

After:

image


neilcsmith-net:
Need to roll another beta for NPE fix, so inclined to think this might be a good, safe fix to squeeze in?

geertjanw:
Nice, I like this, works fine in my build of 12.1 vc1.

PR #2329, opened 2020-08-23
[NETBEANS-4735] Detect JDK-s installed with SDKMan and Debian

lkishalmi:
Just make the life a bit easier...


lkishalmi:
image
Java Platforms started on a fresh user dir.

PR #2331, opened 2020-08-25
[NETBEANS-4718] Disallow FXML controller in default package when modu…

errael:
…lar java

I encouraged someone to file this bug, then they went and assigned it to me, sigh. Anyway, I suspect there are widespread problems with the `JavaFX 2 support` netbeans module related to modular java, but this is a fix for the reported problem (and I'll be filing some other bugs). In the wizard, a validation message is shown, next is disabled when there's module-info and the default package is selected for the controller.

I couldn't find any general module-info handling libraries. It looks like separate projects do their own thing.

### DIFF
--- a/javafx/javafx2.project/src/org/netbeans/modules/javafx2/project/JFXProjectUtils.java
+++ b/javafx/javafx2.project/src/org/netbeans/modules/javafx2/project/JFXProjectUtils.java
@@ -31,6 +31,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.zip.CRC32;
 import java.util.zip.Checksum;
+
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.util.Elements;
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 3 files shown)


javafx/javafx2.project/src/org/netbeans/modules/javafx2/project/JFXProjectUtils.java
```diff
@@ -1776,5 +1779,17 @@ public Void run() throws Exception {
             }
         }
     }
+
+    private static final String MODULE_INFO = "module-info.java"; // NOI18N
+
+    // as in org.netbeans.modules.maven.api.ModuleInfoUtils.hasModuleInfoInSource;
+    public static boolean hasModuleInfo(SourceGroupSupport support) {
+        for (File sourceRoot : support.getSourceGroupsAsFiles()) {
+            if (new File(sourceRoot, MODULE_INFO).exists()) {
```
sdedic, line 1788: I would prefer working with `FileObjects`, rather than converting to `File`s, although it does not make probably big difference here. Consider 
```
for (FileObject sourceRoot : support.getSourceGroups()) {
  if (sourceRoot.getFileObject(MODULE_INFO) != null) {
     ...
  }
}

errael: Excellent. Thanks. Had to go through SourceGroupProxy, but...

@sdedic could you re-review with suggested changes?

sdedic (approved):

matthiasblaesing:
This is reviewed, unittests come back green, lets get this in. Thank you for your work.

PR #2335, opened 2020-08-29
[NETBEANS-4443] PHP 8.0 Support: Static return type

junichi11:

#### Part 1

- https://wiki.php.net/rfc/static_return_type
- Fix the `ASTPHP5Parser.cup`
- Fix the code completion feature
- Add unit tests for the parser and the code completion

##### PHP 8.0 (Project Properties)

nb-php80-static-return-type

##### PHP 7.4 (Project Properties)

nb-php80-static-return-type-php74

#### Part 2

- Fix `IntroduceSuggestion` and `PHP80UnhandledError`
- Fix `UnusableTypesUnhandledError` (Error for redundant combination with the object type)
- Add unit tests for override/implement method, formatting, and the comment generator features

##### Error for redundant combination with the object type

nb-php80-static-return-type-redundant-with-object-type


junichi11:
Regarding the failure of Travis CI:
I've tested `webcommon/javascript2.extjs/test/unit/src/org/netbeans/modules/javascript2/extjs/model/JsStructureTest.java` and `ExtModelTest.java` in my local PC.

```
Testcase: testIssue230177(org.netbeans.modules.javascript2.extjs.model.JsStructureTest):	Caused an ERROR
Server returned HTTP response code: 403 for URL: https://cdnjs.cloudflare.com/ajax/libs/extjs/4.2.1/src/class/ClassManager.js
java.io.IOException: Server returned HTTP response code: 403 for URL: https://cdnjs.cloudflare.com/ajax/libs/extjs/4.2.1/src/class/ClassManager.js
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1894)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
```
```
Testcase: testIssue231923(org.netbeans.modules.javascript2.extjs.model.ExtModelTest):	Caused an ERROR
Server returned HTTP response code: 403 for URL: https://cdnjs.cloudflare.com/ajax/libs/extjs/4.2.1/src/class/ClassManager.js
java.io.IOException: Server returned HTTP response code: 403 for URL: https://cdnjs.cloudflare.com/ajax/libs/extjs/4.2.1/src/class/ClassManager.js
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1894)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
	at java.net.URL.openStream(URL.java:1045)
	at org.netbeans.modules.javascript2.extjs.model.ExtModelTest.setUp(ExtModelTest.java:44)
```

@matthiasblaesing should we upload the test data(ClassManager.js) to `https://netbeans.osuosl.org/binaries/` ?
Maybe, related to this change: https://github.com/apache/netbeans/pull/1451/commits/433ba3a8bb22d8d3d608ca31807422047ff18963

matthiasblaesing:
@junichi11 I run the unittests from javascript2.extjs and for me they run clean (no test failures). I was also able to load the file manually. Is this problem still reproducible for you? 

To be frank: In an ideal world netbeans would be hosted by an entity, that has more flexible policy than the ASF, but we live in the real world. From my POV our use of the osuosl servers already stretches policy. If we don't need to host ourselves, I'd like to keep it like that. I think it is a reasonable expectation, that maven central and cdnjs both keep up with request, as it is their only reason for living.

junichi11:
@matthiasblaesing 

> Is this problem still reproducible for you?

Yes. (I can load the file via a browser.) I tried testing them in my PC because Travis CI failed (reran the job twice)... They also seem to fail in other PRs. I was not sure whether this is a problem temporarily. So, I just asked... I think also it's better if we can keep it. Thanks.

junichi11:
@matthiasblaesing As a workaround: Add "User-Agent" to `URLConnection`?

matthiasblaesing:
@junichi11 maybe we see a local test/change, that requires the setting of a users agent. Your idee sounds good and I think I saw similar problems in the past regarding missing user agent information. So if it works I think this is a good addition. One thing worth exploring is, whether a Browser User Agent string is required or if a more hones user agent:

```java
            connection.addRequestProperty("User-Agent", "NetBeans Unittesting");
```

would also work.

junichi11:
@matthiasblaesing I'll leave it to you :) If you would not like to do it, I'll do it. Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesUnhandledError.java
```diff
@@ -229,6 +258,16 @@ private void checkReturnType(@NullAllowed Expression returnType, boolean isInUni
             } else if (type instanceof UnionType) {
                 ((UnionType) type).getTypes().forEach(unionType -> checkReturnType(unionType, true));
                 checkUnionType((UnionType) type);
+            } else if (type instanceof Identifier) {
+                // method, labmda funciton, and arrow function can use static return type
```
tmysik, line 262: Nitpick: Typos, "la[b]mda func[i]ton".

tmysik (approved):

tmysik:
Looks good to me (again, huge patch so impossible to review properly but - we have tests, they pass so OK for me).

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesUnhandledError.java
junichi11: Thanks for caching them :) Will fix them later. (Just add the `TODO` label)

junichi11:
Yes, it's impossible. Thank you!

PR #2337, opened 2020-09-02
Using modern Java coding style when developing NetBeans

dukescript:
A method of [using modern Java code concepts](https://repo1.maven.org/maven2/org/frgaal/compiler/14.0.2/compiler-14.0.2-README.pdf) while retaining compatibility with JDK 1.8 has recently come to our attention. There has been numerous requests to drop support for running NetBeans on JDK 1.8, but there have always been objections to do so. This PR tries to balance such objections with the need of NetBeans codebase to get the best of modern Java.

First of all there is a [small enhancement to the build system](https://github.com/apache/netbeans/commit/b53e2cf21a421773c2abee51effaeafbfa7b527b). If the in `nbproject/project.properties` specified `javac.source` is higher than `javac.target`, then the [backporting compiler](https://repo1.maven.org/maven2/org/frgaal/compiler/14.0.2/compiler-14.0.2-README.pdf) is triggered to handle the compilation. That allows one to use modern concepts like `var` and still produce JDK 1.8 compatible bytecode - this feature has been used in [maven.htmlui module](https://github.com/apache/netbeans/commit/6066163168eb0e76a5d2f0f40897b1b8f87d52ab) to demonstrate that the compiler works as expected. The biggest benefit has however been observed on [api.html module](https://github.com/apache/netbeans/commit/938b7d26695511809f1ebae00a4cf6763edcc357) - by replacing string concatenation with text blocks, the readability improved a lot.

We believe this PR helps to find balance between those who want to write modern code and those who need to execute it on older JVMs. We'd like to use the [backporting compiler](https://repo1.maven.org/maven2/org/frgaal/compiler/14.0.2/compiler-14.0.2-README.pdf) in the modules we maintain. Once this PR is in, others may opt-in and use this feature by setting `javac.source=14` and `javac.target=1.8` in appropriate module configuration files.


matthiasblaesing:
I did not look at the commit itself, but I have doubts. Who is the creator of this compiler? What is the base of this compiler? For the JDK we have several trustworthy entities (Oracle, Red Hat, SAP, Azul, ...) this project does not even have a domain - `frgaal.org` goes to a site, that looks like a "you can buy this domain if you want" place holder. Even the referenced github repository is gone.

We have such an abandoned project already in the code base. The graaljs copy netbeans uses as javascript parser is a modified copy of the upstream source (or it was in the past upstream itself?). In anycase the local copy was not updated and now we have pretty much a dead end. This special compiler seems to head into the same direction.

I have a very bad feeling with this and for me this is a -1.

lkishalmi:
Is it sound silly to require Java 8 and Java 11 (or 14 maybe) for compilation and set the compiler boot classpath to Java 8?

dukescript:
True, this approach allows NetBeans to eliminate reflection when accessing modern Java APIs. Let's prove that by modifying `applemenu` - a module that desperately wants to use JDK9 API. The current version does that with a bunch of reflection and some adhoc library not even published on Maven. 

Commit 4311d60 shows how to introduce new `applemenu.jdk9` with `javac.target=9` and use the JDK9 API without any hacks.

platform/applemenu/nbproject/project.xml
```diff
@@ -104,7 +104,9 @@
                     </run-dependency>
                 </dependency>
             </module-dependencies>
-            <public-packages/>
+            <public-packages>
+                <package>org.netbeans.modules.applemenu.spi</package>
```
JaroslavTulach, line 108: This is an API change and as such it should be documented with `arch.xml`, `apichanges.xml` and addition of this module to list of Javadocs.

Or turn this into `friend-packages`. I don't like usage of friend packages much but in this case, it is probably fine as nobody is likely to provide other implementation anyway.

JaroslavTulach (approved):
Solve the new SPI package issue; otherwise OK.

neilcsmith-net:
-0 I share @matthiasblaesing misgivings pretty much - how many uncertain forks of javac are we willing to keep relying on?!  Not a block though - I'll leave you to convince Matthias to remove his.

matthiasblaesing:
I had some private exchange regarding this subject. I said that I won't Veto this. And I stand by these words. That does not mean, that I agree with merging this. I think it is an error, but I won't stand in the way.

What concerns me is, that NetBeans _again_ becomes a experiment field and that that will affect the whole code base. Once this is merged and used, our way back is cut or at least removing it will be so painful, that is equivalent to a lock-in. So it comes down how much we trust the entity behind frgaal.

At this point I have no reason to trust into it. I'm not aware of any track record of the authors in the javac compiler area, my concerns were not addressed and I did not find any discussion about it in the OpenJDK mailinglists. I see a single vendor, which is in stark contrast to the OpenJDK javac, which is backed by multiple entities. If one of these entities decides to abandon OpenJDK, other vendors will step up, with frgaal it is EOL. It would help to know about other projects, that are already using it and what will happen with frgaal in the next years.

matthiasblaesing:
Had some time to evaluate the wider consequences. I fear, that this could mean the end of visualvm in Ubuntu/Debian. Debian requires the packages to be boostrapable from source. If this goes according to plan, the platform will not be buildable with a normal javac (openjdk javac and ecj), so debian will need to package frgaal. I don't see this happen, as frgaal pulls in a full JDK and questions will be raised what is so special about netbeans, that we don't use one of the widely used compilers.

eirikbakke:
Agree with the concerns here. (I read over the PR a few days ago but saw that the obvious problems had already been pointed out, so I didn't comment then...)

lkishalmi:
-0.9 as of me. If we move forward I'd rather really set the source and target to Java 11, even if we need to maintain a branch/fork of platform with Java 1.8 support. Even willing to provide some time to release the platform binary compiled with 1.8. Though that would be a different discussion. Though I have no burning urge to go that route at all. I'd be happy if we would just lift the source level to 1.8 first. Probably throwing out some 1.5 related code (profiler, laf hacks). So baby steps...

mcdonnell-john:
How was this merged? I don't remember seeing this asfgit merge something before...

Maybe we should back this out now, as there doesn't appear to be a majorly strong push from the author or anyone else to keep it in compared to the concerns being raised about it.

matthiasblaesing:
@mcdonnell-john asfgit is the bot the asf uses to synchronize gitbox (the apache side of the git hosting) and github. If you see asfgit somewhere, it indicates the changeset came in from the apache side (for example because a committer pushed a commit to gitbox).

mcdonnell-john:
Thanks @matthiasblaesing I saw this earlier and thought it looked odd.

JaroslavTulach:
>  I'd rather really set the source and target to Java 11

Just for the record: -1 against dropping JDK8. Thank you.

lkishalmi:
Well, this is not the "dropping JDK 8" discussion. I think that one would come eventually a few years later. On short term I would be just happy if we can lift the source level on all our modules to Java 8, then remove some ancient support, like Java 5 profiler, some LAF hacks, etc.

This is about whether we would allow to depend on another 3rd party compiler in the exchange for some syntax candy.

JaroslavTulach:
> This is about ...  some syntax candy.

No, it is not! True, original motivation (in 938b7d2 and 6066163) might be the syntax candy. I agree that is the useless part. However 31a7eb9 shows it is not candy at all. It is about ability to **compile against future Java APIs on JDK8**. With such ability...

> Well, this is not the "dropping JDK 8" discussion.

...we would never need to have "dropping JDK 8" discussion again. Think about it, before [rolling back](https://github.com/apache/netbeans/pull/2369).

matthiasblaesing:
This gift comes with a hefty price: Dependence on a single entity and an uncertain future. 

The discussion about the baseline JDK will always come up, because the classpath changes. Consider the new http client libraries, the upcoming additions of project panama (improved JNI), project loom (green threading) and so on. At the price of more complexity in netbeans you can again hide them behind compatibility shims, but that won't scale long.

dukescript:
It is actually very easy to use 

> http client libraries

with the here-in introduced system. Here is a demo: Start by creating a [new module](https://github.com/dukescript/netbeans/commit/d439aee5d5992fd8e6b35bfc57bcc8c013fd6a50) which is using the JDK11 specific [HttpClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html) API:
```java
    public void demo() throws Exception {
        HttpClient client = HttpClient.newBuilder()
             .version(Version.HTTP_1_1)
             .followRedirects(Redirect.NORMAL)
             .connectTimeout(Duration.ofSeconds(20))
             .proxy(ProxySelector.of(new InetSocketAddress("proxy.example.com", 80)))
             .authenticator(Authenticator.getDefault())
             .build();
        HttpRequest request = HttpRequest.newBuilder()
            .uri(URI.create("https://foo.com/"))
            .timeout(Duration.ofMinutes(2))
            .header("Content-Type", "application/json")
            .build();
        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
        System.out.println(response.statusCode());
        System.out.println(response.body());  
    }
```
of course, you can't compile this on JDK8. What do you do? You enable the [future API aware compiler](https://github.com/dukescript/netbeans/commit/72e067467dac6f4f651a048a5cc31ba237d37875) and tell it to use JDK11's API (by setting `javac.target=11`). Then everything compiles even on JDK8. The last step is to [use the module](https://github.com/dukescript/netbeans/commit/b325d3c0297def205c23c781a510f848c4dff1ef) when available - of course it is only available when running on JDK11+, otherwise it is disabled.

matthiasblaesing:
> The module with javac.target=11is only enabled when running on JDK11+, otherwise it is disabled.

So what happens when the module is run on JDK8? Yeah it won't work and thus I can compile against JDK9+ API, but I won't be able to rely on it. My point is, that we still need to decide which is the baseline JDK we expect. That won't magically go away - independently of the javac used.

matthiasblaesing:
So I thought about this some more and looked at the wider ecosystem. My suggestion is this:

- let #2369 go through, reestablishing the working state of the netbeans repository
- drop #2371 as it is followup for #2337 which is being reverted by #2369
- discuss the suggestion on the dev@ mailinglist as it has enough consequences to warrent a wider audience
- see what is the outcome of that discussion and if it goes through reprepare a PR with the then agreed changeset

What I noticed is, that there is in fact a java compiler out there, that has the property to support target < source. It is the eclipse compiler. Given that they base a whole IDE on top of it, that survived a few years, I would say, that it is battle proven. Maybe at some point the OpenJDK javac will reach the same state and the changeset that is used to create frgaal might be pushed upstream and thus available to the wider ecosystem.

neilcsmith-net:
+1 to @matthiasblaesing suggestion, and for what it's worth a retrospective change from -0 to -1 on this PR (was not expecting the veto to be dropped after "private exchange" or the debate to be curtailed).

lkishalmi:
+1 on @matthiasblaesing  from my side as well. I'm going to merge #2369. That will make #2371 obsolete. @dukescript or @JaroslavTulach shall start a discussion thread on the dev@ list. Let's come to an agreement there.
If we end up supporting the intentions of #2337, then @dukescript can file a new PR which combines #2337 with #2371 

We are not in a hurry. Let's sit down and talk.

lkishalmi:
Well just read the change in #2371 it is independent from all of this, and can be supported to be merged.

dukescript:
> > The module with javac.target=11is only enabled when running on JDK11+, otherwise it is disabled.
> 
> So what happens when the module is run on JDK8? 

You cannot run such module on JDK8, that's the point. It contains 
```
OpenIDE-Module-Java-Dependencies: Java > 11
```
in its manifest. As such the [runtime container knows](https://bits.netbeans.org/12.0/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html#std-sec) such module shall not be enabled on JDK8, JDK9, JDK10.

PR #2340, opened 2020-09-03
[NETBEANS-4696] maven project open use no-lock getModuleName

errael:
This change avoids the deadlock by having maven projects open use the non-locking `SourceUtils.parseModuleName(moduleInfo)` instead of invoking the parser.

It should be a safe change. Another change, for the other bad actor in the deadlock, is forthcoming

@tzezula Will you review?


lkishalmi:
Looks good for me, but would you remove the unused imports as well?

errael:
> Looks good for me, but would you remove the unused imports as well?

And the now unused LOG as well.

lkishalmi (approved):
Looks sane to me.

PR #2341, opened 2020-09-03
[NETBEANS-4696] html parser dont wait for projects open

errael:
The html parser, in twisty paths, is used from many areas (for example, braceMatching). Without this change, the parser waits for project open, and this created a deadlock. (#2340 takes care of one side). However, according to analysis by @jtulach in the issue report, the html parser should not be calling into the projectui api.

This PR does not fix that, it does avoid blocking on the project ui api. A careful analysis of the html parsing code, and how it's used, should be done to see if there are any unexpected consequences of this change.

The eventual goal is to completely remove the dependency as outlined in the analysis as seen in issue [NETBEANS-4696].

### DIFF
--- a/webcommon/javascript2.knockout/nbproject/project.xml
+++ b/webcommon/javascript2.knockout/nbproject/project.xml
@@ -97,15 +97,6 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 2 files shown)


errael:
Turns out removing the dependency on projectuiapi.base does not change behavior; see issue for details.

@jtulach will you comment and/or review?

JaroslavTulach (approved):

PR #2346, opened 2020-09-06
Improve @modules hints - transfer open, handle hidden usages.

jlahoda:


arvindaprameya:
Approved, thanks for doing this

PR #2352, opened 2020-09-07
Fix unittest for webcommon caused by download failures from cdnjs

matthiasblaesing:
It seems that for some networks CDNJS requires sending a User-Agent
header. This changeset adds it to the requests.


lkishalmi:
Thank you for this one!

matthiasblaesing:
@junichi11 I still can't reproduce the problem locally and have to rely on the unittests, which now come back green, could you please test if this fixes for you locally too?

lkishalmi (approved):
LGTM. Much better, it seems even Travis is happy with it.

lkishalmi:
Checked the code and merging as Travis likes it.

junichi11:
@matthiasblaesing Works fine. Thank you for fixing it :)

PR #2353, opened 2020-09-08
[NETBEANS-4779] Update to the mac os installer as it was failing on Big Sur

mcdonnell-john:
Adding the fix that was applied to generate the Mac OS installer for Apache NetBeans 12.1


lkishalmi (approved):
Well the 12.1 Mac build seems to get +1-s so this seems to work.

PR #2357, opened 2020-09-09
Add Plugin for BootsFaces in order to support <b:dataTable

t-oster:
Hi,

I am using NetBeans with BootsFaces (https://www.bootsfaces.net/). It works quite good, but the autocomplete fails on tags, which create scoped variables like `<b:dataTable value=[ListOfType<B>] var="x"/>` which creates a scoped variable `x` of type `B`.
I just copied the `@ServiceProvider(service = TagFeatureProvider.class)` from the PrimeFaces plugin, but since the org.netbeans.modules.web.jsfapi does not seem to expose this API I had to add my plugin as friend module.

I failed at creating a standalone module without adding a friend module, so my question is if you would include this module or what would be the correct way to add this functionality?

I also thought about a more generic plugin, where the user can specify a list of tags (namespace, tagName, var, value) which enables autocomplete in EL for those without creating a separate plugin for every taglib.


enterprise/web.bootsfaces/licenseinfo.xml
```diff
@@ -0,0 +1,27 @@
...
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<licenseinfo>
```
lkishalmi, line 22: Empty licenseinfo is not required.

enterprise/web.bootsfaces/src/org/netbeans/modules/web/bootsfaces/BootsFacesTagFeatureProvider.java
```diff
@@ -0,0 +1,108 @@
...
+                    && library.getNamespace().equalsIgnoreCase(iterableTag.getNamespace())) {
+                return iterableTag;
+            }
+        }
+        return null;
+    }
+
+    private enum IterableTag {
+
+        DATA_TABLE(BOOTSFACES_UI_NAMESPACE, "dataTable", VALUE, VAR);
```
lkishalmi, line 78: Trailing //NOI18N is missing

enterprise/web.bootsfaces/src/org/netbeans/modules/web/bootsfaces/BootsFacesTagFeatureProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * {@link TagFeatureProvider} for BootsFaces.
+ * 
+ * @author toster
+ */
+@ServiceProvider(service = TagFeatureProvider.class)
+public class BootsFacesTagFeatureProvider implements TagFeatureProvider {
+
+    private static final String BOOTSFACES_UI_NAMESPACE = "http://bootsfaces.net/ui";
```
lkishalmi, line 38: //NOI18N-s are missing

lkishalmi (changes requested):

enterprise/web.bootsfaces/src/org/netbeans/modules/web/bootsfaces/resources/layer.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
t-oster: I tried to run without it, but it threw an Exception.

t-oster:
Thank you for the review, I will update it asap

lkishalmi:
Just remove the empty layer.xml I've checked out this PR removed the layer.xml, it still builds fine. Then we can finally merge this.

t-oster:
> Just remove the empty layer.xml I've checked out this PR removed the layer.xml, it still builds fine. Then we can finally merge this.

I will do this, but i think it will build fine, but it won't run when you try to install the plugin.

lkishalmi (approved):

matthiasblaesing:
This will not be part of 12.2 as the registration of the module in the cluster is missing: https://issues.apache.org/jira/browse/NETBEANS-5054

t-oster:
@matthiasblaesing is #2559 what I should do? I added it underneath web.primefaces, because that's what it is based on.

PR #2358, opened 2020-09-10
Moving libs.truffleapi to ide cluster. GraalJS impl recommended for JDK15+

sdedic:
During `commit-validation` cleanup, I've noticed some dependencies from `java` to `webcommon` cluster. Several of them is on `truffleapi` module, which should IMHO move into `ide` cluster (similar to `graalsdk` library).

As for `libs.graaljs` required as project dependency for `nashorn.execution`: during discussion with @JaroslavTulach, it seemed that GraalJS impl is only necessary when there is no platform that supports Nashorn (either too old, or JDK15+) where the GraalJS classes are put on classpath and used for execution.

So this PR will change the project dependency into `OpenIDE-Module-Recommends` so GraalJS is enabled automatically; in the case the platform is missing, and the GraalJS library is NOT available (= module not installed), it will ask the user whether to download the module (libs.graaljs). If the user says "No", decision is remembered - just status message prints on unsuccessful exec.

I made the `libs.graaljs` module visible in AU, since the user may need to manually download it.

Note that Java cluster **has still a dependency** on `webcommon`: `debugger.jpda.trufflenode` uses some types from javascript implementation. So `ant -Dcluster.config=java` remains broken (the resulting ZIP does not contain webcommon and therefore `debugger.jpda.trufflenode` is broken and disabled on 1st startup)


java/nashorn.execution/manifest.mf
```diff
@@ -4,4 +4,4 @@ Comment: OpenIDE-Module-Layer: org/netbeans/modules/nashorn/execution/resources/
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/nashorn/execution/Bundle.properties
 OpenIDE-Module-Specification-Version: 1.16
 AutoUpdate-Show-In-Client: false
-
+OpenIDE-Module-Recommends: org.netbeans.libs.graaljs
```
JaroslavTulach, line 7: Suspicious at first sight. I'd expect `cnb.org.netbeans.libs.graaljs` as that is the token exported by default by the module system for the `org.netbeans.libs.graaljs` module. However correct, the module explicitly exports `org.netbeans.libs.graaljs` token as can be seen below.

java/nashorn.execution/src/org/netbeans/modules/nashorn/execution/ModuleRequestException.java
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.nashorn.execution;
+
+/**
+ *
+ * @author sdedic
+ */
+public class ModuleRequestException extends UnsupportedOperationException {
```
JaroslavTulach, line 25: Package private, maybe?

JaroslavTulach: `final`.

webcommon/libs.graaljs/manifest.mf
```diff
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
-AutoUpdate-Show-In-Client: false
+AutoUpdate-Show-In-Client: true
 OpenIDE-Module: org.netbeans.libs.graaljs/1
 OpenIDE-Module-Layer: org/netbeans/libs/graaljs/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/libs/graaljs/Bundle.properties
 OpenIDE-Module-Specification-Version: 1.7
-OpenIDE-Module-Provides: javax.script.ScriptEngine.js
+OpenIDE-Module-Provides: javax.script.ScriptEngine.js,org.netbeans.libs.graaljs
```
JaroslavTulach, line 7: I see. The module explicitly provides `org.netbeans.libs.graaljs`.

webcommon/libs.graaljs/manifest.mf
```diff
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
-AutoUpdate-Show-In-Client: false
+AutoUpdate-Show-In-Client: true
 OpenIDE-Module: org.netbeans.libs.graaljs/1
 OpenIDE-Module-Layer: org/netbeans/libs/graaljs/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/libs/graaljs/Bundle.properties
 OpenIDE-Module-Specification-Version: 1.7
```
JaroslavTulach, line 6: Increase module version when you are changing the API (by increasing set of provided tokens).

JaroslavTulach (approved):
Good. Thanks.

java/nashorn.execution/manifest.mf
sdedic: `cnb.` could be better... it actually depends on the GraalJS library being defined rather than code from the module. Do you think it should rather depend on the module as such (rather than on the 'library' feature "token") ?

java/nashorn.execution/src/org/netbeans/modules/nashorn/execution/ModuleRequestException.java
sdedic: Thrown in `execution' package, caught / processed in 'actions'. But definitely `final`.

java/nashorn.execution/manifest.mf
JaroslavTulach: As you wish.

lkishalmi:
Question: Should we backport this one to 12.0 as a patch so it would be usable running on JDK 15+?

sdedic:
> Question: Should we backport this one to 12.0 as a patch so it would be usable running on JDK 15+?

I don't think so. In nashorn execution has a hard dependency on graaljs lib, so it is actually usable on JDK15. This patch only makes the dependency less explicit, so nashorn.execution could eventually work without libs.graaljs (of course on systems that contain nashort js shell).

lkishalmi:
After restarting  a some test task in Travis a few times, I suspect that this PR fails some tests in ergonomics. @sdedic could you look at that?

sdedic:
> After restarting a some test task in Travis a few times, I suspect that this PR fails some tests in ergonomics. @sdedic could you look at that?

Sorry, seems to be based on slightly older master (hmtl.ojet was removed in the meantime). Rebased, squashed.

sdedic:
> Question: Should we backport this one to 12.0 as a patch so it would be usable running on JDK 15+?

No, I don't think it's necessary. Master declares a hard dependency `nashorn.execution` -> `libs.graaljs`, so it is usable on both JDK < 15 and >= 15.

sdedic:
Some more discussion is needed: the failure is caused by making `libs.graaljs` showing in AU. The containing feature (webcommon) is then enabled, which fails the test.

BUT as I played around with pristine builds of `cluster.config=java`, it seems the patch could be very shorter:
- if there's NO `webcommon` cluster in the distribution, then there's no JS support, so `nashorn.execution` does not activate (no text/javascript MIME recognizer). So no Run or Debug actions and Javascript are plaintext files anyway
- if there's `webcommon` cluster, then `nashorn.execution' loads the `libs.graaljs` and everything is fine
PLUS
- it seems little ridiculous to support hypothetical config that contains pieces of clusters (i.e. javascript, but not graaljs). If some distribution does such packaging, let it solve the problems as well.

That would leave only `OpenIDE-Module-Recommends:` in `nashorn.execution` with no fancy AutoUpdate UI suggestions.

@JaroslavTulach @entlicher -- agreed ? If so, I'll reduce this patch.

entlicher:
I agree wit that simplification, it looks sufficient to me.

sdedic:
OK, edited again, simplified as outlined / agreed on.

entlicher (approved):

PR #2361, opened 2020-09-11
[NETBEANS-4766] Arrange Gradle Source info detail tabs to the right

lkishalmi:
image
Also added a little enhancement to be able to keep the selected detail tab active between selecting sourcesets.


eirikbakke:
Did you try resizing the window and making sure that the new Swing layout behaves properly?

I looked through add*Listener calls, found no obvious memory leaks (no listening on global objects, just components within the same panel).

Seems harmless!

eirikbakke (approved):

lkishalmi:
I've tried to resize. it behaves reasonably good.

PR #2365, opened 2020-09-13
Show types for chained method invocations.

jlahoda:
Adds types after method invocation for chained method invocations.

See:
chained-types

### DIFF
--- a/ide/editor.lib2/src/org/netbeans/modules/editor/lib2/view/ParagraphViewChildren.java
+++ b/ide/editor.lib2/src/org/netbeans/modules/editor/lib2/view/ParagraphViewChildren.java
@@ -168,8 +168,8 @@ final class ParagraphViewChildren extends ViewChildren<EditorView> {
                 view.setRawEndOffset(relEndOffset); // Below offset-gap
                 view.setParent(pView);
                 // Possibly assign text layout
-                if (view instanceof HighlightsView || (view instanceof PrependedTextView && ((PrependedTextView) view).getDelegate() instanceof HighlightsView)) {
-                    HighlightsView hView = (HighlightsView) (view instanceof HighlightsView ? view : ((PrependedTextView) view).getDelegate());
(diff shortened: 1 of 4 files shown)


neilcsmith-net:
Nice! Better info when building / using fluent APIs definitely a plus.

Is the setting for this same as method parameters? Would prefer separate.

JaroslavTulach:
Can we turn "View/Show Inline Hints" on by default?

jlahoda:
Regarding enabling - best would be to have some settings (we can expect more virtual text in the future, and we could have more settings even for the existing features).

For enabling by default, I would prefer to improve the appearance and behavior, before we roll this to everybody.

AlexFalappa:
Great!

homberghp:
It would be nice if that worked when the result of a chain is used. It is helpful in developing, but as soon as you assign the value the hint info is lost.
Once you use the result of a chain, either by assigning it or using as an argument to a method the hints are lost, which is a pitty.
image

Rahul-khandelwal:
I just installed netbeans 12.2, how can I turn this feature on?

AlexFalappa:
@Rahul-khandelwal it's in the _View_ menu

swpalmer:
> Once you use the result of a chain, either by assigning it or using as an argument to a method the hints are lost, which is a pitty.
> image

Is three a bug filed for this?  It really ruins what is otherwise a great freature.

homberghp:
I have gained some more experience with this tool and the way it can be used in teaching and in particular in code review.
Sometimes stream expressions can de quite complex to follow, and this tool is a benefit.
As JaroslavTulach already suggested, this hints should stay on, **even when the result of the stream is assigned** to a variable or when there is a closing ';' Actually this would be reason to envy intelij.  The fact that the hints disappear when the result is assigned or has a terminating ; is a pity, and makes it less usefull in review. Remember that one important aspect of code is readability and this certainly is a readability aid.

PR #2367, opened 2020-09-13
[NETBEANS-2898] Removed obsolete Oracle JET support

lkishalmi:
I hope both @geertjanw and Travis would agree.


geertjanw:
Completely agree. Thanks for doing this.

PR #2369, opened 2020-09-14
Revert "Merge remote-tracking branch 'apache/master' into ModernCode"

lkishalmi:
This reverts commit #2337 814184b7077c77216e70e05685973687a5f817be, reversing
changes made to 6408fe154baceaafc437f99f1fd4dcbaccf99479.

PR #2371, opened 2020-09-15
Fixing nbjdk.home behavior broken by #2337 

dukescript:
>  matthiasblaesing commented 6 hours ago in #2369:
>  @mcdonnell-john I was able to reproduce your problem. It happens if the build JDK and the target JDK are not the same. I had this in my ".nbbuild.properties":
```
#permit.jdk9.builds=true
nbjdk.home=/home/matthias/bin/jdk-8
```
> This will lead to ant finding the desktop classes, while in the compile run they are not present. 

Please accept apology for not spotting this problem sooner. This is an additional fix for #2337 to make it work `nbjdk.home=` set.

Up to @lkishalmi to decide whether to merge this, merge #2369 or merge both.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
@@ -89,6 +89,14 @@ public class CustomJavac extends Javac {
     }
 
     @Override
+    public Path getBootclasspath() {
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
There is talk about reverting #2337, as that commit is at least controversial.

JaroslavTulach (approved):

lkishalmi (approved):
Looks good to me.

PR #2372, opened 2020-09-16
Update dependencies to GraalVM 20.2.0

daemontus:
This PR updates `truffle-api`, `graalsdk` and `graaljs` to their latest version (`20.2.0`). This brings numerous bugfixes and new features as described in [GraalVM Release Notes](https://www.graalvm.org/release-notes/). 

(Specifically, some of the bugfixes are needed to successfully integrate graal-based language libraries such as [Heap Language](https://github.com/daemontus/heap-language), which can replace old Nashorn-based OQL scripting)

Additionally, GraalVM tests running on Travis have been split into two variants: MTS (Medium-Term Support) version and latest version to ensure compatibility with both.

Edit(1): Small test update was also need. Since the new GraalVM versions support automatic Long ⇒ String [conversion](https://github.com/graalvm/graaljs/commit/c09ae303ee6b7d9a1a470cade79b147c15ea6d46), there was a method resolution ambiguity problem. Test was fixed by using `String` argument directly instead of `Long`.

Edit(2): Seems that the `OutOfMemory` error was legitimate. For now, memory limit for `graal-sdk` tests has been slightly increased (480 ⇒ 700mb), but I will try to submit this as a performance problem to graalvm r/python.


lkishalmi:
Probably need to be rebased on the current master as webcommon tests are failing, also please check the failing GraalVM tests!

daemontus:
Sure, rebase coming soon :) Looking into the GraalVM stuff now - not sure what's happening there yet. Seems like GraalJS is fine, but anything that is using R/Python fails with `OutOfMemory` exception... I'll investigate more tomorrow.

.travis.yml
```diff
@@ -752,9 +752,21 @@ matrix:
           script:
             - (cd java/java.lsp.server; ant build-vscode-ext)
 
-        - name: GRAALVM Tests
+        - name: "GraalVM Tests (latest)"
```
hectorespert, line 755: Ok for the Travis changes :+1:

platform/api.scripting/test/unit/src/org/netbeans/api/scripting/JavaScriptEnginesTest.java
```diff
@@ -256,7 +256,7 @@ public void allowLoadAClassInJS() throws Exception {
         // BEGIN: org.netbeans.api.scripting.JavaScriptEnginesTest#allowLoadAClassInJS
         Object fn = engine.eval("(function(obj) {\n"
                 + "  var Long = Java.type('java.lang.Long');\n"
-                + "  return new Long(33);\n"
+                + "  return new Long(\"33\");\n"
```
JaroslavTulach, line 259: Is this change really necessary? It does work in the GraalVM 20.2.0 console (CCing @woess):
```bash
./graalvm-ce-java8-20.2.0/bin/js --jvm
> let Long = Java.type("java.lang.Long");
> new Long(33)
33
> new Long("33")
33
```
It would be surprising that the interop behaves differently in the console and in our `ScriptEngine`...

daemontus: Well, [this is the error](https://travis-ci.org/github/apache/netbeans/jobs/727721920#L8773) that I ran into in CI and also later confirmed locally. Maybe there is a way to turn this off `ScriptEngine` configuration?

sdedic (approved):

platform/api.scripting/test/unit/src/org/netbeans/api/scripting/JavaScriptEnginesTest.java
JaroslavTulach: aa76809 seems to be a workaround. Requesting Nashorn compatibility mode activates [NASHORN_HOST_ACCESS](https://github.com/graalvm/graaljs/blob/vm-20.2.0/graal-js/src/com.oracle.truffle.js.scriptengine/src/com/oracle/truffle/js/scriptengine/GraalJSScriptEngine.java#L270) which contains the Long ⇒ String [conversion](https://github.com/graalvm/graaljs/commit/c09ae303ee6b7d9a1a470cade79b147c15ea6d46) as identified by Samuel. By calling
```java
b.put("polyglot.js.allowHostAccess", true);
```
after that one replaces the Nashorn compatibility host access with `HostAccess.ALL` switching back to the behavior of the engine as of GraalVM 19.3.x. CCing @iamstolis

JaroslavTulach (approved):
If the tests pass, I think we are well setup for an integration.

platform/api.scripting/test/unit/src/org/netbeans/api/scripting/JavaScriptEnginesTest.java
iamstolis: Yes, this is a known consequence of the new conversion and also a known workaround that removes the new conversion see https://github.com/graalvm/graaljs/issues/286

Note that the conversion itself is not incorrect (in the sense that Nashorn has it as well and so some Nashorn users need it). Unfortunately, the inclusion of the conversion discloses that our method resolution (in `truffle`) is different than Nashorn one.

JaroslavTulach (changes requested):
I've just got a build failure locally `ant -f ide/libs.graalsdk/ clean build` yields various signature errors:
```
Class org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl
  "E5.2 - Adding abstract methods" : method public abstract boolean org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl.isMetaInstance(java.lang.Object,java.lang.Object)
  "E5.2 - Adding abstract methods" : method public abstract int org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl.hashCodeImpl(java.lang.Object)
  "E5.2 - Adding abstract methods" : method public abstract java.lang.String org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl.getMetaSimpleName(java.lang.Object)
  "E5.2 - Adding abstract methods" : method public abstract java.lang.String org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl.getMetaQualifiedName(java.lang.Object)
  "E5.2 - Adding abstract methods" : method public abstract boolean org.graalvm.polyglot.impl.AbstractPolyglotImpl$AbstractValueImpl.equalsImpl(java.lang.Object,java.lang.Object)
```
that reminds me:
* we should increase version of each module in `manifest.mf`
* we should increase dependencies of the modules among each other in `project.xml` - it is clear that Truffle API 19.3.x isn't going to link to GraalSDK 20.2
* we should investigate why the travis Signature job didn't fail (and make it fail first)

JaroslavTulach (approved):
Signature check for [revision afc3259](https://travis-ci.org/github/apache/netbeans/jobs/728329754) failed - e.g. we have successfully enabled the test in afc3259. Then, in [revision 714f340](https://travis-ci.org/github/apache/netbeans/jobs/728344575) the test succeeded. E.g. sigtest related things are fixed.

Versions and dependencies among GraalVM modules are updated since eed81f5d058

PR #2376, opened 2020-09-17
Two icons converted from PNG to SVG.

otulach:
I converted start and stop button from PNG to SVG in Inkscape, because i wanted to help you.


lkishalmi:
Do not forget about the licenseinfo.xml and add the apache license into the svg header (if that's possible)

neilcsmith-net:
We might consider a separate place for all the editable SVG files (maybe a resource folder in nbbuild?) and stripping out the unnecessary metadata in the shipped icons using eg. Inkscape's save optimized?  Mind you, if we have to have the license header in each SVG anyway, might not be worth it.

eirikbakke:
I don't think we should be doing these one-by-one. It will not end up looking good, and reviewing will take much longer than if we do 50-60 of these in batch at a time. For these particularly simple icons (a single square and a single circle), making the style consistent with other icons, and ensuring alignment to the pixel grid etc., is 90% of the work.

We need to do a full batch of icons from https://people.csail.mit.edu/ebakke/misc/netbeans-icons/prioritized.html , and ensure they are following some agreed-upon style guide (e.g. as proposed in https://issues.apache.org/jira/browse/NETBEANS-2617 ). Then we need to go in and make sure the vertices in the icons are aligned to pixel boundaries and so on.

(As an example: In these particular two SVG icons, the style is inconsistent with existing icons because the borders are too hard. The proposed style guide in NETBEANS-2617 prescribes 33% transparency for the border, on top of the fill color. The fill color of the "Stop Macro Recording" is also too dark--the shade of the old bitmap icon was better.)

So let's hold off integrating individual icons until we have a large batch done.

errael:
I would like to be able to just look at the icons. For example, an issue associated with this PR which had png files attached.

matthiasblaesing:
@eirikbakke however the svg creation happens initially, at some point in time new icons will be needed and then it will be the same situation as volunteers doing work now. I don't see us holding back new features, just because a blessed icon is not ready or until a full new batch is done.

eirikbakke:
> I don't see us holding back new features, just because a blessed icon is not ready or until a full new batch is done.

The feature here is, "make NetBeans look better". When it comes to icons, having a consistent scheme is an essential part of "looking good". So adding one or two icons only, which will now look different next to the existing ones in the toolbar, actually makes NetBeans look worse. Thus the -1.

To review, we also need to see screenshots before and after, at 100% and 200% DPI scaling. Often you'll see issues at this point, that were not evident when you were just looking at the icons in the drawing software. This means recompiling, which takes a while... longer than the time it took to make two icons.

Unrelated: I also recommend keeping the icons in a single file in the drawing software (in different "artboards" or whatever is the equivalent in Inkscape), so they can be viewed side-by-side while designing them. This also facilitates cut-and-paste of shapes between icons.

eirikbakke:
Also, these two icons have already been drawn... see the attachments to https://issues.apache.org/jira/browse/NETBEANS-2617 . Some coordination is needed to avoid duplicating work.

matthiasblaesing:
@eirikbakke how do you suggest how it should work in the future? I don't have Adobe software, don't own a system that could run it and won't pay the money for the software. Am I not allowed to contribute GUI code anymore? Consider a new feature where 4-5 icons are needed, that are derivates of existing ones. I would have taken the SVGs, modified them and committed them with the new feature. Your approach basicly cuts me off from this way and adds a high reliance on proprietary software and a small (not yet exiting) number of people.

eirikbakke:
@matthiasblaesing For the scenario you describe, I think it would be fine to use Inkscape. It can open SVGs created by other software, and you can then use that to copy shapes. This is, of course, hypothetical until we actually have a large number of existing icons converted...

eirikbakke:
Inkscape can open Illustrator files, by the way--Illustrator files are technically just PDFs.

ebarboni:
If I edit via Inkscape will it be savable on ai to keep metadata for so people using illustrator can modify or is it a one way edition ?

eirikbakke:
@ebarboni Not sure, but I wouldn't count on the PDF import/export process being lossless. I'd also guess the SVG itself works better as a common file format.

eirikbakke:
And sorry to the original poster here, who stepped into the middle of a long discussion :-) Thanks for your efforts! Feel free to join the mailing list ( https://netbeans.apache.org/community/mailing-lists.html#dev ).

eirikbakke:
I've created a Google Sheet here to track progress:

(Happy to add anyone who wants to participate as an editor, so they can add notes and "reserve" icons they want to draw.)

JaroslavTulach:
> Also, these two icons have already been drawn... see the attachments to 
> https://issues.apache.org/jira/browse/NETBEANS-2617 . Some coordination is needed to avoid duplicating work.

The best co-ordination would be to integrate the drawn icons into the codebase, right? Will anyone do that anytime soon?

eirikbakke:
> As the best co-ordination would be to integrate the drawn icons into the codebase, right?

That's later in the process. We need a way to say "I am going to be drawing these 24 icons, please nobody else draw the same ones", and see the icons to be done next to each other while adding comments to individual ones regarding consistency etc.

The PR process is more heavyweight--it requires a bunch of steps: reviewing the graphics in the drawing tool to check for technical problems (pixel alignment, unnecessary anchor points), creating a new git branch, conversion from Inkscape or Illustrator to the actual SVG files, insertion of licenses into SVG files, recompilation to produce screenshots, reviewing new icons side-by-side in the GUI, waiting for comments, and so on. The batch approach is needed (1) to minimize work per icon and (2) to be able to review icons for consistency before they all go into use in the IDE.

If you disagree with the need for a batch approach, taking the above points into account, that would be a useful discussion to resolve. The coordination method follows from that.

> Will anyone do that [integrate the drawn icons into the codebase] anytime soon?

I can do that, but not before February, unfortunately.

JaroslavTulach:
> If you disagree with the need for a batch approach, ... 

I agree with Matthias that we cannot lock ourselves into batch-only approach. 

> We need a way to say "I am going to be drawing these 24 icons

My son wasted his days drawing two icons that have been drawn for weeks/months...

> ...The coordination method follows from that.

...enough to integrate what you get and not shelve it on a webpage or elsewhere. 

We have so little contributions of the SVG images, that coordination is the last problem to resolve - an email to `dev@` is enough. Just try to integrate what has been redrawn asap.

JaroslavTulach:
What shall happen with this PR? Close without integration? Or shall I ask my son to polish the drawings and help him fix the failing validations?

eirikbakke:
@JaroslavTulach You say you disagree with the batch approach, but have not responded to the reasons I laid out for why it is necessary?

JaroslavTulach:
Should I know you insist on batch approach I would never let your `SVGLoader` changes in. Only when all the icons are redrawn, you'd be allowed to integrate first API change. How does that feel?

Many reasons were given why batch approach is insane and cannot work in long term:
> Consider a new feature where 4-5 icons are needed, that are derivates of existing ones. I would have taken the SVGs, modified them and committed them with the new feature. Your approach basicly cuts me off from this way and adds a high reliance on proprietary software and a small (not yet exiting) number of people.

I agree with 95% of what has been written against your batch approach.

Plus I found it totally scary that you have SVG icons, and they are not yet integrated. That's not how an open source project works.

eirikbakke:
> Should I know you insist on batch approach
I won't actually insist--I'm happy to retreat from this work and let others do it as they see fit. Just make sure you understand the considerations involved.

> I would never let your SVGLoader changes in
If you're concerned about the size of the Batik JARs, I'd be fine with deactivating this feature until we have a set of SVG icons ready. The SVGLoader implementation is pluggable--you can just exclude it from the cluster.

> Plus I found it totally scary that you have SVG icons, and they are not yet integrated. That's not how an open source project works.

The story here is that I have a NetBeans Platform application that I develop for my company. Every icon in that application needs to be scalable, so I have drawn those particular icons in SVG. A subset of those icons happen to be icons from the NetBeans IDE. I'm happy to donate those icons, but I have to do that work outside of work time... i.e. during my next vacation.

> Consider a new feature where 4-5 icons are needed
See my earlier response further up in the thread.

> My son wasted his days drawing two icons that have been drawn for weeks/months...
It's not wasted time! The first time you draw an icon, most of the work involved is to get familiar with the drawing software etc. Once you've done a few icons, the next ones are much quicker. That's one of the reasons why I recommend doing 10-12 icons before integrating--with more experience, the later icons end up being better, and you go back and refine the first ones you drew etc.

eirikbakke:
OK, let me at the very least export the SVG files from the Illustrator files in NETBEANS-2617... so they're not "locked up" in the AI files for those who don't have Illustrator.

eirikbakke:
Alright, grumble grumble, I'll make a PR which integrates the currently drawn icons... :-)

eirikbakke:
OK, here's a PR that integrates all the SVG icons that have been drawn so far:
https://github.com/apache/netbeans/pull/2387

JaroslavTulach:
With #2387 being submitted, this PR no longer needs to be integrated. Thanks Ondřeji for your PR - based on this PR NetBeans is going to get better!

PR #2383, opened 2020-09-20
[NETBEANS-4824] - General Availability for javadoc 15

pepness:
Update link to javadoc 15 now that is in general availability

Before:
https://download.java.net/java/early_access/jdk15/docs/api/

After:
https://docs.oracle.com/en/java/javase/15/docs/api/


matthiasblaesing:
Looks sane to me.

PR #2384, opened 2020-09-21
Prevent NPE from JavaToCHyperlinkProvider when no project available.

jlahoda:
Seen e.g. when doing Ctrl-click in a code from a library, which is not a project.


lkishalmi:
Seems to be trivial enough.

PR #2385, opened 2020-09-21
Improving the behavior of cpplite project - building should have prop…

jlahoda:
…er output; ergonomics should not disable the modules while the project is being used.

The output improvements are probably not controversial, but the ergonomics tweak probably is.

The issue is that feature-off-demand will disable "unused" features after 5 sessions. But, it detects used projects based on a file pattern inside the project. For the cpplite project, we have no specific file pattern inside the project, which is fine per Projects API. Please note the cpplite project does not even require "Makefile" or anything like that to work.

So, the workaround used here is that the cpplite project marks itself as "used" on the filesystem.

### DIFF
--- a/cpplite/cpplite.project/nbproject/project.xml
+++ b/cpplite/cpplite.project/nbproject/project.xml
@@ -59,6 +59,15 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.modules.extexecution.base</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>1.14</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
                     <code-name-base>org.netbeans.modules.projectapi</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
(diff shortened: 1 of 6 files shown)


JaroslavTulach (changes requested):
Rather than modifying `cpp` modules to be aware of ergonomics, I suggest to design a way for ergonomics to recognize that a certain project is a cpplite project (`prj.getLookup().lookup(Class.forName(specifiedInCppLite.properties)) != null`?). Ergonomics module is already aware of most of the supported modules, so this direction doesn't add any additional coupling.

> For the cpplite project, we have no specific file pattern inside the project, ... 
> the cpplite project does not even require "Makefile" or anything like that to work.

That is absolutely fine. The patterns as specified in `ergonomics/ide.ergonomics/cpplite.properties` don't have to exhaustive. However some common patterns shall be there, so ergonomics can help turn cpplite cluster on. How would a C/C++ user open a typical project after initial start otherwise?

jlahoda:
> That is absolutely fine. The patterns as specified in `ergonomics/ide.ergonomics/cpplite.properties` don't have to exhaustive. However some common patterns shall be there, so ergonomics can help turn cpplite cluster on. How would a C/C++ user open a typical project after initial start otherwise?

Using New Project, which creates the new project. This is not really different from CND, AFAIK - if I would want to open e.g. the ccls project in CND, I think I would need to go through the New Project dialog as well. So no huge difference, IMO.

JaroslavTulach (approved):

arvindaprameya:
Thanks for doing this, approved.

PR #2386, opened 2020-09-21
Cleanup of not just dependencies

JaroslavTulach:
While working on [java.lsp.server debugging](https://github.com/JaroslavTulach/netbeans/tree/java-lsp-server-debugging-attempt2) various issues with module dependencies, their ability to run in headless mode, to run without `core.windows` module, etc. have been identified. This PR is my attempt to bring them into NetBeans master branch.

The most important (from my architect perspective) is the new "branding API" to prevent open projects to be reopened on the application start. With the `LOAD_PROJECTS_ON_START=false` the system always starts with no projects opened.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
@@ -100,6 +100,9 @@ public final class RunUtils {
             SingleMethod method = methods.iterator().next();
             files.add(method.getFile());
         }
+        if (files.isEmpty()) {
+            files.addAll(lookup.lookupAll(FileObject.class));
+        }
         return files.toArray(new FileObject[files.size()]);
     }
 
--- a/ide/javascript2.debug/nbproject/project.xml
+++ b/ide/javascript2.debug/nbproject/project.xml
@@ -114,6 +114,7 @@
             <friend-packages>
                 <friend>org.netbeans.modules.debugger.jpda.js</friend>
                 <friend>org.netbeans.modules.debugger.jpda.truffle</friend>
+                <friend>org.netbeans.modules.java.lsp.server</friend>
                 <friend>org.netbeans.modules.javascript.v8debug</friend>
                 <friend>org.netbeans.modules.javascript.v8debug.ui</friend>
                 <friend>org.netbeans.modules.javascript2.debug.ui</friend>
--- a/ide/javascript2.debug/nbproject/project.xml
+++ b/ide/javascript2.debug/nbproject/project.xml
@@ -114,7 +114,6 @@
             <friend-packages>
                 <friend>org.netbeans.modules.debugger.jpda.js</friend>
                 <friend>org.netbeans.modules.debugger.jpda.truffle</friend>
-                <friend>org.netbeans.modules.java.lsp.server</friend>
                 <friend>org.netbeans.modules.javascript.v8debug</friend>
                 <friend>org.netbeans.modules.javascript.v8debug.ui</friend>
                 <friend>org.netbeans.modules.javascript2.debug.ui</friend>
--- a/ide/parsing.indexing/manifest.mf
+++ b/ide/parsing.indexing/manifest.mf
@@ -4,4 +4,5 @@ OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/parsing/indexing/Bundle.p
 OpenIDE-Module-Layer: org/netbeans/modules/parsing/impl/indexing/layer.xml
 AutoUpdate-Show-In-Client: false
 OpenIDE-Module-Implementation-Version: 1
+OpenIDE-Module-Needs: org.netbeans.modules.parsing.impl.indexing.implspi.ActiveDocumentProvider
 
--- a/ide/projectapi.nb/manifest.mf
+++ b/ide/projectapi.nb/manifest.mf
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.projectapi.nb
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/projectapi/nb/Bundle.properties
-OpenIDE-Module-Provides: cnb.org.netbeans.modules.projectapi.nb
-OpenIDE-Module-Specification-Version: 1.13
+OpenIDE-Module-Provides: org.netbeans.spi.project.ProjectManagerImplementation
+OpenIDE-Module-Specification-Version: 1.14
 AutoUpdate-Show-In-Client: false
--- a/ide/projectapi/manifest.mf
+++ b/ide/projectapi/manifest.mf
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.projectapi/1
-OpenIDE-Module-Specification-Version: 1.77
+OpenIDE-Module-Specification-Version: 1.78
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/projectapi/Bundle.properties
 OpenIDE-Module-Layer: org/netbeans/modules/projectapi/layer.xml
-OpenIDE-Module-Recommends: cnb.org.netbeans.modules.projectapi.nb
+OpenIDE-Module-Needs: org.netbeans.spi.project.ProjectManagerImplementation
--- a/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
@@ -441,12 +441,18 @@ public final class OpenProjectList {
                 }
             });
         }
-            
+
+        @NbBundle.Messages({
+            "LOAD_PROJECTS_ON_START=true"
+        })
         private void loadOnBackground() {
-            lazilyOpenedProjects = new ArrayList<Project>();
-            List<URL> URLs = OpenProjectListSettings.getInstance().getOpenProjectsURLs();
-            final List<Project> initial = new ArrayList<Project>();
-            final LinkedList<Project> projects = URLs2Projects(URLs);
+            lazilyOpenedProjects = new ArrayList<>();
+            final boolean loadProjectsOnStart = "true".equals(Bundle.LOAD_PROJECTS_ON_START());
+            List<URL> urls = loadProjectsOnStart ?
+                    OpenProjectListSettings.getInstance().getOpenProjectsURLs() :
+                    Collections.emptyList();
+            final List<Project> initial = new ArrayList<>();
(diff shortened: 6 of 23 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/AutoUpdateCatalogParser.java
```diff
@@ -240,6 +241,8 @@ public void endElement (String uri, String localName, String qName) throws SAXEx
                 break;
             case manifest :
                 break;
+            case message_digest:
```
JaroslavTulach, line 244: The `AutoUpdate` task was very verbose with modern catalogs. Assuming this field got in with one of @matthiasblaesing  fixes, I've added it to the set of recognized, but not processed fields.

platform/settings/src/org/netbeans/modules/settings/RecognizeInstanceObjects.java
```diff
@@ -93,24 +94,49 @@ protected Lookup lookupFor(Object obj) {
...
+        public static final MSL DEFAULT = new MSL();
+        static {
             try {
                 ModuleSystem ms = Main.getModuleSystem(false);
                 if (ms != null) {
                     ModuleManager man = ms.getManager();
-                    man.addPropertyChangeListener(WeakListeners.propertyChange(this, man));
+                    man.addPropertyChangeListener(WeakListeners.propertyChange(DEFAULT, man));
                 } else {
                     LOG.log(Level.WARNING, "Not listening on module system");
```
JaroslavTulach, line 107: This message was printed out  everytime new instance of `OverObjects` was created in a unit test. Quite annoying. Enough to warning once per test.

java/maven/src/org/netbeans/modules/maven/spi/actions/AbstractMavenActionsProvider.java
```diff
@@ -88,6 +88,10 @@ public AbstractMavenActionsProvider() {
             files.add(method.getFile());
         }
 
+        if (files.isEmpty()) {
```
JaroslavTulach, line 91: Ant projects honor `FileObject` as a selection in the `Lookup`, but Maven projects didn't. After this change they will use `FileObject` (if no `DataObject` is found) as the indentification of main file to execute.

ide/projectapi.nb/manifest.mf
```diff
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.projectapi.nb
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/projectapi/nb/Bundle.properties
-OpenIDE-Module-Provides: cnb.org.netbeans.modules.projectapi.nb
-OpenIDE-Module-Specification-Version: 1.13
+OpenIDE-Module-Provides: org.netbeans.spi.project.ProjectManagerImplementation
```
JaroslavTulach, line 4: @tzezula, there is no need to `OpenIDE-Module-Provides: cnb.org.netbeans.modules.projectapi.nb` it is provided automatically. Rather than that the module no provides implementation of `org.netbeans.spi.project.ProjectManagerImplementation`.

ide/projectapi/manifest.mf
```diff
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.projectapi/1
-OpenIDE-Module-Specification-Version: 1.77
+OpenIDE-Module-Specification-Version: 1.78
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/projectapi/Bundle.properties
 OpenIDE-Module-Layer: org/netbeans/modules/projectapi/layer.xml
-OpenIDE-Module-Recommends: cnb.org.netbeans.modules.projectapi.nb
+OpenIDE-Module-Needs: org.netbeans.spi.project.ProjectManagerImplementation
```
JaroslavTulach, line 6: @tzezula: Here we need an implementation of `org.netbeans.spi.project.ProjectManagerImplementation`. Previously the module recommended single implementation, but that wasn't enough. If the module was missing a `NullPointerException` was raised. Hence we "need" - e.g. the module isn't going to be enabled without an implementation, but the implementation can come from any module.

ide/web.browser.api/manifest.mf
```diff
@@ -2,5 +2,5 @@ Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.web.browser.api
 OpenIDE-Module-Layer: org/netbeans/modules/web/browser/ui/resources/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/web/browser/api/Bundle.properties
-OpenIDE-Module-Requires: org.openide.windows.WindowManager
+OpenIDE-Module-Recommends: org.openide.windows.WindowManager
```
JaroslavTulach, line 5: This module can run without window manager. Hence just "recommends".

extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
```diff
@@ -100,6 +100,9 @@ public static FileObject extractFileObjectfromLookup(Lookup lookup) {
             SingleMethod method = methods.iterator().next();
             files.add(method.getFile());
         }
+        if (files.isEmpty()) {
+            files.addAll(lookup.lookupAll(FileObject.class));
+        }
```
JaroslavTulach, line 105: This is the same [fix as in Maven](78fbf67)

java/maven/src/org/netbeans/modules/maven/spi/actions/AbstractMavenActionsProvider.java
JaroslavTulach: Also [fixed in Gradle](034a600c1329).

sdedic (approved):

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
```diff
@@ -441,12 +441,18 @@ boolean closeBeforeOpen(final Project[] arr) {
...
+        @NbBundle.Messages({
+            "LOAD_PROJECTS_ON_START=true"
+        })
         private void loadOnBackground() {
-            lazilyOpenedProjects = new ArrayList<Project>();
-            List<URL> URLs = OpenProjectListSettings.getInstance().getOpenProjectsURLs();
-            final List<Project> initial = new ArrayList<Project>();
-            final LinkedList<Project> projects = URLs2Projects(URLs);
+            lazilyOpenedProjects = new ArrayList<>();
+            final boolean loadProjectsOnStart = "true".equals(Bundle.LOAD_PROJECTS_ON_START());
```
tzezula, line 450: It's an interesting usage of resource bundle. :-)
But I think it's rather a hack to use module resource bundle for configuration.
Do we already use such a pattern somewhere?
If not I would prefer a method.

tzezula (approved):
Overall seems good to me. Just please re-evaluate the usage of projectui resource bundle for configuration.

lkishalmi (approved):
Seems to be fine with me.

lkishalmi:
Travis is failing to access Apache's gitbox, restarted the failing jobs.

ide/javascript2.debug/nbproject/project.xml
```diff
@@ -114,6 +114,7 @@
             <friend-packages>
                 <friend>org.netbeans.modules.debugger.jpda.js</friend>
                 <friend>org.netbeans.modules.debugger.jpda.truffle</friend>
+                <friend>org.netbeans.modules.java.lsp.server</friend>
```
matthiasblaesing, line 117: I don't see the reason for this change - am I missing the dependency relationship?

ide/parsing.indexing/manifest.mf
```diff
@@ -4,4 +4,5 @@ OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/parsing/indexing/Bundle.p
 OpenIDE-Module-Layer: org/netbeans/modules/parsing/impl/indexing/layer.xml
 AutoUpdate-Show-In-Client: false
 OpenIDE-Module-Implementation-Version: 1
+OpenIDE-Module-Needs: org.netbeans.modules.parsing.impl.indexing.implspi.ActiveDocumentProvider
```
matthiasblaesing, line 7: Just to  understand this correctly: Is the intention here, that parsing.indexing and parsing.nb are loosely coupled? I.e. parsing.indexing needs and implemenation of that SPI?

matthiasblaesing (approved):
I admit, that I could not follow every change, but looks sane to me. I left two inline questions, but these are just to improve my understanding. Thanks.

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
JaroslavTulach: Yes, it is used at many places, controlling behavior of splash screen, controlling behavior of window system. We even have `<api group='branding'..../>` for that case - here is the current [list of branding APIs](https://bits.netbeans.org/dev/javadoc/branding.html).

ide/parsing.indexing/manifest.mf
JaroslavTulach: I tried to assemble minimal application and I was getting `IllegalStateException`. Turned out that `parsing.indexing` cannot run without `ActiveDocumentProvider`. The `OpenIDE-Module-Needs` just formalizes that dependency. @tzezula may confirm, but the code in `RepositoryUpdater` speaks clearly:
```java
        this.activeDocProvider = Lookup.getDefault().lookup(ActiveDocumentProvider.class);
        if (this.activeDocProvider == null) {
            throw new IllegalStateException("No ActiveDocumentProvider instance in global lookup.");    //NOI18N
        }
```

ide/javascript2.debug/nbproject/project.xml
JaroslavTulach: Good catch. This is a change needed for the [java.lsp.server debugging](https://github.com/JaroslavTulach/netbeans/tree/java-lsp-server-debugging-attempt2), but not really a cleanup. Moreover one should bump module version when changing list of friends. 

Let's remove the change from this PR: 2847f6e

JaroslavTulach:
> Overall seems good to me. Just please re-evaluate the usage of projectui resource bundle for configuration.

There are [Branding APIs](https://bits.netbeans.org/dev/javadoc/branding.html) in the NetBeans code base. It is a better than `System.getProperty` way to configure the system (no need to mangle with command line switches), if the configuration is "per application" (which comes with own branding). It is more lightweight than [API seam](http://wiki.apidesign.org/wiki/APISeam).

> looks sane to me. I left two inline questions, but these are just to improve my understanding.

Answered and fixed in 2847f6e.

entlicher (approved):

PR #2387, opened 2020-09-22
[NETBEANS-4830] Integrate SVG icons drawn so far

eirikbakke:
By popular request, here's a PR that integrates all SVG icons that have been drawn to date, by Pete Whelpton and myself.

These icons have all been drawn according to the style guide on https://issues.apache.org/jira/browse/NETBEANS-2617 , which aims to maintain the same general look as the old NetBeans icons, with some modernizations. The goal is for the new SVG icons to look OK next to the old bitmap icons (since there will always be bitmap icons in the mix for the foreseeable future). If this PR is merged, future icons should adhere to the same style for consistency.

This covers only a tiny subset of unique icons in the NetBeans IDE. Some of these icons occur many times around the codebase; in those cases the SVG files have been copied to each location correspondingly as well.

To see icons before and after, comparing bitmap icons in the master branch to SVG icons in this PR branch, see this spreadsheet:

Also attaching some screenshots. Testing was done on Windows 10 at 2x HiDPI scaling on Java 10.0.2. Also tested briefly on Java 8.0.181 (which does not support HiDPI scaling, but which will still render the SVG icons at 1x resolution). Note that there are several places where getting the SVG icons to show at full resolution will require some future tweaks (using ImageUtilities.image2icon instead of explicitly constructing an ImageIcon, using ImageUtilities.createDisabledIcon instead of the default button disablement mechanics etc.).

I recommend reviewing this offline using "git whatchanged", and with the spreadsheet linked above. Would be good if someone could test this on Linux and MacOS as well (as there is one change in each respective LAF to set folder icons).

screenshot1
screenshot2
screenshot3
screenshot4
screenshot5


JaroslavTulach (approved):
> The goal is for the new SVG icons to look OK next to the old bitmap icons (since there will always be bitmap icons in the mix for the foreseeable future).

Great attitude! Yes, unless a miracle happens and we get someone to sponsor the icon redrawing in batch, we will have to live with both types of icons for a while. It is important that look fine when next to each other on non-HiDPI displays (at least).

JaroslavTulach:
I know there is a discussion raging on about SVG icon registry. However I believe it is an orthogonal problem which shall not be blocked by the way we organize our icons (CCing @lkishalmi) - if we are just replacing `wait.png` by `wait.svg` we should go on.

One enhancement has however recently come to my mind: We could use [@StaticResource annotation](https://bits.netbeans.org/12.0/javadoc/org-netbeans-api-annotations-common/org/netbeans/api/annotations/common/StaticResource.html) - it wasn't there when people started to copy `wait.png` to thirty different places, but it was later designed to avoid doing that. Nobody just had enough incentives to fix all the copies. Could we use the switch to SVG as an opportunity to cleanup the copies?

Just find a common module all other depend on and place the `wait.svg` in there. Let's suppose it ends up in `org/openide/awt/resources/wait.svg`. Then let all other modules declare a dependency on the resource by:
```java
@StaticResource
private static final String WAIT_SVG = "org/openide/awt/resources/wait.svg";
```
That basically forms a "compile time" API, as the annotation's processor is going to fail if somebody removes the icon. No need for too many copies. Enough to share it.

eirikbakke:
@jtulach You could introduce the StaticResource technique, but it doesn't have to wait for the SVG icon. Just have it point to "wait.png", and once we are ready to commit some more SVGs, you just put a file named "wait.svg" in the same folder, and ImageLoader will load that one instead. This keeps the SVG loader module optional.

JaroslavTulach:
Sure, go on and integrate (do you have the priviledge, right?). I am just saying that rather than copying a single `.svg` many times around, one could use the `@StaticResource` technique and fix two issues at once. Completely optional.

eirikbakke:
I prefer if the duplicate cleanup is done separately... this PR is already a lot of commits. I'll give others a few days to object before merging this PR.

This PR does introduce a lot of duplicate SVG icon files, but they are exact duplicates that can trivially be identified at a later time (using checksums as noted at the end of https://issues.apache.org/jira/browse/NETBEANS-2617 ).

matthiasblaesing (approved):
Assuming that @peedeeboy agreed to this and you are in the position to donate his changes this looks good to me. I noticed some differences in the icon styling, but looked ok and if it improved rendering on high-dpi screens great. I'm on a normal screen but did not see a problem from  a tryme run..

Thank you.

peedeeboy:
Yes - please consider my icon contributions thus far (and any future contributions that I upload to the JIRA ticket) as donated, and that I'm happy for  them to be added to a PR and merged by anyone 👍

eirikbakke:
Thanks for your contribution, @peedeeboy!

PR #2390, opened 2020-09-22
Do not report unsatisfied autoload fragments triggered by host module.

sdedic:
The `netbeans.libs.javafx.*` library family print warnings on startup. There are several such libraries, each providing `org.openide.modules.jre.JavaFX`, but with `OpenIDE-Module-Needs: ` that demands a specific OS. Since they are `autoload`, they do not activate on their own on startup (c.f. eager and regular modules), but only when demanded.

Some other module depends on the provided token, enabling the (autoload) fragment. The fragment enables its host (also autoload), and the host attempts to load all the remaining fragments. Since the other autoloads have unsatisfied dependencies, a warning is printed to the log - quite bloated:

```
WARNING [org.netbeans.core.modules]: Module StandardModule:org.netbeans.libs.javafx.macosx jarFile: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/extra/modules/org-netbeans-libs-javafx-macosx.jar had unexpected problems: [Java > 11, needs org.openide.modules.os.MacOSX] (willEnable: [Netigso: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/ide/modules/com-google-gson.jar, Netigso: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/platform/modules/org-apache-commons-codec.jar, Netigso: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/platform/modules/org-apache-commons-logging.jar, StandardModule:org.apache.tools.ant.module jarFile: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/extide/modules/org-apache-tools-ant-module.jar, StandardModule:org.apache.xml.resolver jarFile: /space/src/nb/apache/github/incubator-netbeans/nbbuild/netbeans/ide/modules/org-
....
```
plus complete module configuration.

This warning was originally an assert, up to [NETBEANS-1481](https://issues.apache.org/jira/browse/NETBEANS-1481), but still is rather a diagnostic warning that we should avoid: the scenario used by `javafx` implementation is IMHO +- valid.

This PR disables the warning in the case the autoload fragment is enabled **by its host module**. The warning (or other messages) will be still printed if the module enables from different source, i.e. the only provider for a token, or a direct dependency.

The bad thing is that we **do not have** a mechanism that would only enable a fragment if someone needs it (true autoload); the current autoload fragments are "somewhat eager" because they all attempt to load whenever their host is loaded. I tried to find a way how to enable the "true" autoload behaviour, but it could probably break the existing code -- the runtime behaviour in 11 and 12 was  (attempt) to load all fragments.

I have added some tests to fix the current de-facto behaviour and check the logs that no warning is printed.

### DIFF
--- a/platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
+++ b/platform/o.n.bootstrap/src/org/netbeans/ModuleManager.java
@@ -1279,7 +1279,11 @@ public final class ModuleManager extends Modules {
             }
             for (Module m : testing) {
                 if (!modules.contains(m) && !m.isAutoload() && !m.isEager()) {
-                    throw new IllegalModuleException(IllegalModuleException.Reason.ENABLE_TESTING, m);
+                    // it is acceptable if the module is a non-autoload host fragment, and its host enabled (thus enabled the fragment):
+                    Module maybeHost =  attachModuleFragment(m);
+                    if (maybeHost == null && !testing.contains(maybeHost)) {
+                        throw new IllegalModuleException(IllegalModuleException.Reason.ENABLE_TESTING, m);
+                    }
                 }
             }
         }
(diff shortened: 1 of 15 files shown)


JaroslavTulach (approved):
I see tests. Having some code coverage is good. If they pass, let's integrate.

sdedic:
The error passed unnoticed through commit-validation. Only travis complained about log length. And the other sorting bug passed not only commit-validation, but through several releases as well :)

PR #2392, opened 2020-09-23
[NETBEANS-4762] Fixing Lombok in Gradle Projects

lkishalmi:
It seems supporting MODULE_PROCESSOR_PATH in Gradle breaks annotation processor detection.
I've checked, the Maven plugin does not support this classpath, as well as the support is new in 12.1 for Gradle

See: https://github.com/apache/netbeans/blob/a04b626ec6293b4f0e3006a726ffcc121a2af313/java/java.source.base/src/org/netbeans/modules/java/source/indexing/APTUtils.java#L392

here decided if the MODULE_PROCESSOR_PATH is provided the isModule is true and NB tries to read the information through the Module however that does not provide the Lombok service providers. It maybe a bug in our Module implementation or a bug in Lombok

This PR just simply removes the support of MODULE_PROCESSOR_PATH from the classpath implementation for Gradle.


jlahoda (approved):
I don't know if Gradle supports the module processor path, but providing it as a copy of processor path is not quite right. If module processor path is specified, it will be used instead of processor path, and the processors on the module processor path need to be specified using modular approaches. While automatic modules might help, it is better to not provide it, unless the user has configured it.

PR #2393, opened 2020-09-23
Avoid NPE in BridgingIOProvider when using null color

JaroslavTulach:
I've got a `NullPointerException` while working on [java.lsp.server debugging](https://github.com/JaroslavTulach/netbeans/tree/java-lsp-server-debugging-attempt2):
```
        at org.openide.io.BridgingIOProvider$BridgingIOColorPrint.print(BridgingIOProvider.java:299)
        at org.openide.windows.IOColorPrint.print(IOColorPrint.java:68)
        at org.netbeans.modules.gradle.api.execute.RunUtils.executeGradleImpl(RunUtils.java:206)
        at org.netbeans.modules.gradle.api.execute.RunUtils.executeGradle(RunUtils.java:123)
        at org.netbeans.modules.gradle.ActionProviderImpl.invokeProjectAction(ActionProviderImpl.java:266)
        at org.netbeans.modules.gradle.ActionProviderImpl.invokeAction(ActionProviderImpl.java:137)
        at org.netbeans.modules.java.lsp.server.debugging.launch.NbLaunchDelegate.lambda$nbLaunch$0(NbLaunchDelegate.java:115)
```
The problem is that `BridgingIOProvider` isn't ready for `null` color. This PR fixes it and adds some unit tests that, for example discover unexpected usage of `new StringWriter()` in bridging `OutputWriter` constructor!?

In addition to that I've added a check into Gradle code to verify that both `IOColorPrint` and `IOColors` are supported.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
@@ -199,7 +199,7 @@ public final class RunUtils {
                 new ProxyNonSelectableInputOutput(io));
         if (initialOutput != null) {
             try {
-                if (IOColorPrint.isSupported(io)) {
+                if (IOColorPrint.isSupported(io) && IOColors.isSupported(io)) {
                     IOColorPrint.print(io, initialOutput, IOColors.getColor(io, IOColors.OutputType.LOG_DEBUG));
                 } else {
                     io.getOut().println(initialOutput);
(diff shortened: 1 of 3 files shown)


jhavlin:
Looks fine. Thank you for fixing!

sdedic (approved):

JaroslavTulach:
Thanks for your reviews. I'll merge when I wake up.

PR #2394, opened 2020-09-23
Add asf.yaml file to ease configuration

ebarboni:
According to current configuraiotn https://gitbox.apache.org/schemes.cgi?netbeans we only put PR in the worklog.

I guess that previously we had a label showing pr available. So using asf.yaml file from infra to get it again + configuration

label: Add a 'pull-request-available' label to referenced tickets.
link: Add a link to the GitHub PR/issue when it's opened to the referenced Jira ticket.


matthiasblaesing (approved):
Looks sane to me.

PR #2396, opened 2020-09-23
Enable commit-validation

sdedic:
Finally ... all necessary changes should be in.


JaroslavTulach (approved):
I am thankful the `ant commit-validation` is going to be back!

PR #2397, opened 2020-09-23
rebase cnd  branch with master

BradWalker:
The CND module code branch has not been updated recently with a rebase from master.. This needs to be done to keep things somewhat current.

That's what I've done here.

As for testing, it builds and runs correctly. The C/C++ module is activated.


matthiasblaesing:
I have mixed feelings here. In general I prefer rebases, but that are cases where a single author creates a PR. In this case the development of the CND branch is separated from master, but runs in parallel.

It depends on what we envision as "end-game". What I mean, what happens when we what to get this to master. If we agree, that a plain merge is ok, then I would merge master into cnd. If we want to keep cnd more separate a rebase would be more logical. I looked at this PR and have some doubts, that we gain much by a rebase.

I think merging master might be the better option. @jlahoda what do you think?

jlahoda:
I would incline to a merge (rather than a rebase) as well.

lkishalmi:
Shall we close this PR then?

matthiasblaesing:
Closing this - CND already had update merges from master, so this is obsolete.

PR #2398, opened 2020-09-23
[NETBEANS-4840] - Fix NullPointerException when activating plugins

pepness:
Fresh build of current master and then try to activate any of the java related plugins will end in a `java.lang.NullPointerException`.

Plugins that fail to activate:
- Developing NetBeans
- Groovy
- JavaFX2
- Java SE
- Java Web and EE

This was integrated by [PR-2390](https://github.com/apache/netbeans/pull/2390). 

Do not know if there is a special way of handling this operation that @sdedic would need.


sdedic (approved):

sdedic:
No, just omission that slipped through the previous review. Feel free to merge.

PR #2399, opened 2020-09-25
[NETBEANS-4832] Avoid assertion & NPE in fxml code completion

errael:
@sdedic could you take a look at this fix.

The assertion was due to CompilerInfo not having a list of modules. Fixed in `FXMLCompletion2.Q.Taks.run` with the getPackageElement:
```
    ci = CompilationInfo.get(result);
    // following populates module info, partial fix for [NETBEANS-4832]
    ci.getElements().getPackageElement("java.lang");
```

This fixed code completion inside an UI element, for example<Label.../>"

However there was an NPE inside a list, for example in`<children>...</children>`. This was fixed by what looks like a bug/oversight in `ElementUtils.getTypeElementByBinaryName`.

In FxIncludeCompleter.createCompleter
```
    TypeMirror tm = prop.getType().resolve(ctx.getCompilationInfo());
```
returned null which causes NPE in jdk.  fixed by the identity check in
```
    ElementUtils.getTypeElementByBinaryName
```

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
@@ -60,6 +60,11 @@ public class ElementUtils {
         for (ModuleElement me : allModules) {
             TypeElement found = getTypeElementByBinaryName(task, me, name);
 
+            if (result == found) {
+                // avoid returning null, partial fix for [NETBEANS-4832]
+                continue;
+            }
+
             if (found != null) {
                 if ((ModuleSymbol) me == syms.unnamedModule) {
                     foundInUnamedModule = true;
--- a/java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
@@ -72,7 +72,7 @@ public class ElementUtils {
                 if (result != null) {
                     if (foundInUnamedModule == true) {
                         for (TypeElement elem : new TypeElement[]{result, found}) {
-                            if ((elem.getKind() == ElementKind.CLASS || elem.getKind() == ElementKind.INTERFACE)
+                            if ((elem.getKind().isClass() || elem.getKind().isInterface())
                                     && (((ClassSymbol) elem).packge().modle != syms.unnamedModule)) {
                                 return elem;
                             }
--- a/javafx/javafx2.editor/src/org/netbeans/modules/javafx2/editor/FXMLCompletion2.java
+++ b/javafx/javafx2.editor/src/org/netbeans/modules/javafx2/editor/FXMLCompletion2.java
@@ -30,7 +30,8 @@ import javax.xml.parsers.SAXParserFactory;
 import org.netbeans.api.editor.mimelookup.MimeLookup;
 import org.netbeans.api.editor.mimelookup.MimeRegistration;
 import org.netbeans.api.java.source.ClasspathInfo;
-import org.netbeans.api.java.source.CompilationInfo;
+import org.netbeans.api.java.source.CompilationController;
+import org.netbeans.api.java.source.JavaSource.Phase;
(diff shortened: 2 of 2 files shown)


errael:
Just added stacktrace for the NPE to issue https://issues.apache.org/jira/browse/NETBEANS-4832

java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
```diff
@@ -60,6 +60,11 @@ public static TypeElement getTypeElementByBinaryName(JavacTask task, String name
         for (ModuleElement me : allModules) {
             TypeElement found = getTypeElementByBinaryName(task, me, name);
 
+            if (result == found) {
+                // avoid returning null, partial fix for [NETBEANS-4832]
+                continue;
+            }
+
```
sdedic, line 67: Not related to this bug, but a few lines below: `(elem.getKind() == ElementKind.CLASS || elem.getKind() == ElementKind.INTERFACE)` IMHO does not work for `ANNOTATION_TYPE` and `ENUM`. Could be worth to fix as
`elem.getKind().isClass() || elem.getKind().isInterface` ?

javafx/javafx2.editor/src/org/netbeans/modules/javafx2/editor/FXMLCompletion2.java
```diff
@@ -161,6 +161,8 @@ public void run(ResultIterator resultIterator) throws Exception {
                 }
                 Parser.Result result = resultIterator.getParserResult();
                 ci = CompilationInfo.get(result);
+                // following populates module info, partial fix for [NETBEANS-4832]
+                ci.getElements().getPackageElement("java.lang");
```
sdedic, line 165: I understand the trigger, but couldn't it be fixed right in the getPackagElement(), @jlahoda ?

java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
errael: So you're saying that the `if(result==found)continue` seems OK.

And that there's an additional issue that could arise in some cases; and you've proposed a fix which can be incorporated with this change.

javafx/javafx2.editor/src/org/netbeans/modules/javafx2/editor/FXMLCompletion2.java
errael: Did you mean `getElements()` rather than `getPackageElement()`?

jlahoda: I think the expected way would be to get a CompilationController (instead of just CompilationInfo) and call toPhase(ELEMENTS_RESOLVED), or alike, on it.

java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
errael: Added separate commit for ANNOTATION_TYPE, ENUM, still in this PR.

javafx/javafx2.editor/src/org/netbeans/modules/javafx2/editor/FXMLCompletion2.java
errael: @jlahoda Use CompilationController as suggested. Works, thanks. Please review.

java/java.source.base/src/org/netbeans/modules/java/source/ElementUtils.java
errael: @sdedic Known issues handled, please review.

errael:
ping @sdedic @jlahoda Your comment have been incorporated. Can we finish this?
CC: @lkishalmi

sdedic (approved):

PR #2400, opened 2020-09-25
[NETBEANS-4844] Non-empy modulebootpath shall be provided for non-modular code as well.

lkishalmi:
Well it seems even a non modular project is required to fill out the moduleBoot as
https://github.com/apache/netbeans/blob/5743f22f60aac775a0a2ebb3c8a8f6d6be39ac45/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java#L1134
checks that for detecting Java 9 and up


lkishalmi:
@jlahoda I also would like to ask that is the JavaParser implementation correct? There is a isModuleInfo check earlier
https://github.com/apache/netbeans/blob/5743f22f60aac775a0a2ebb3c8a8f6d6be39ac45/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java#L1095

So the later checks should go for non-modular java setup. Am I right?

PR #2407, opened 2020-09-29
Run commit-validation on apple.

sdedic:
Run `commit-validation` as part of MacOS build in github actions.


JaroslavTulach (approved):
Great. Thank you Sváťo!

PR #2408, opened 2020-09-29
[NETBEANS-4861] nb-javac Upgrade to JDK-15

Akshay-Gupta-Oracle:
Upgraded nb-javac to merge with jdk15 so the new nb-javac will support jdk-15 features as well. All changes in nb-javac-15 are complete.


lkishalmi:
@Akshay-Gupta-Oracle Please check why the signature check is failing.

Akshay-Gupta-Oracle:
> @Akshay-Gupta-Oracle Please check why the signature check is failing.

Some api change did not get registered, trying to add in libs.javacapi.

matthiasblaesing:
I asked on the mailing list, but did not get a reply: Oracle is still using apache controlled infrastructure (netbeans.osuosl.org), how far are plans to find a different distribution channel? Maven central would be the obvious place as we already require plugin developers to use it and it is the established way to distribute libraries for java applications.

neilcsmith-net:
+1 to what @matthiasblaesing said, and as per mailing list conversation, if we can move to Maven Central we should consider publishing as NBMs via plugin portal (and remove from third-party UC) so that IDE updates aren't needed to update nb-javac.

arvindaprameya:
> +1 to what @matthiasblaesing said, and as per mailing list conversation, if we can move to Maven Central we should consider publishing as NBMs via plugin portal (and remove from third-party UC) so that IDE updates aren't needed to update nb-javac.

Let me get back to you all with if and how we can do this .

arvindaprameya:
Approved, thanks for getting this done Akshay

arvindaprameya:
This is the latest nb-javac to support JDK 15 features along for NB 12.2

Akshay-Gupta-Oracle:
@lkishalmi can you please add the milestone as 12.2 please?

lahodaj:
The change looks fine to me.

Regarding publishing on Plugin Portal - not sure if that's desirable (at least unless we can have a separate plugin for every NetBeans release). As happened during the 11 timeframe (with .1 and .2 or .2 and .3, I don't recall), a newer version of NetBeans may require nb-javac that is incompatible with the newer version of NetBeans. So having a single AUC with a single version of nb-javac for both such versions may not be possible.

ebarboni:
I'm not sure for plugin portal bug having nb-javac on maven central instead of netbeans.osuosl.org make easier to use for Mavenized NetBeans. 

By the way, you can't use org.netbeans groupId as it's locked by Apache NetBeans. 
But lots of groupId at com.oracle exists should be possible to add one specific like com.oracle.nbtooling and also maybe there is a global Oracle way to publish. Oracle jdbc are since a year or two in maven central

neilcsmith-net:
@lahodaj yes, that's true, but we do now have unique plugin URLs for each release, which was half the problem. Unfortunately, 12.1 still ends up forwarded to the plugin catalog for 12.0 on the portal itself.  So we would need to fix that for this to happen - a good step forward, but probably not a viable idea for 12.2 at this time.

lkishalmi:
I hope merged early enough to get some Java 15 feedback in the 12.2 cycle.

jglick:
> if we can move to Maven Central we should consider publishing as NBMs

Would this allow nb-javac to be downloaded from Central? See the problem I have in https://github.com/stapler/netbeans-stapler-plugin/pull/7.

netbeansuser2019:
This seems badly broken in Apache Netbeans 12.2 as every field, parameter,... that it is named as "permits" are presented a error even source level is 11 and JDK is 11.

netbeansuser2019:
See also NETBEANS-5055

PR #2410, opened 2020-09-30
[NETBEANS-4857] prevent property change events during tree sync

MKuettner:
Prevent PropertyChangeEvents in ExplorerManager during tree synchronization if multiple OutlineViews share one single ExplorerManager.
See Jira Issue: https://issues.apache.org/jira/browse/NETBEANS-4857

### DIFF
--- a/platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
+++ b/platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
@@ -859,6 +859,10 @@ public class OutlineView extends JScrollPane {
         invalidate();
         validate();
         Node[] arr = manager.getSelectedNodes ();
+        
+        // [NETBEANS-4857]: prevent property change events during synchronization  
+        outline.getSelectionModel().removeListSelectionListener(managerListener);
+        
         outline.getSelectionModel().clearSelection();
         int size = outline.getRowCount();
         int firstSelection = -1;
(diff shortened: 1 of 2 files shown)


platform/openide.explorer/src/org/openide/explorer/view/OutlineView.java
```diff
@@ -876,6 +880,9 @@ final void synchronizeSelectedNodes(boolean scroll, Node... nodes) {
                 }
             }
         }
+        // [NETBEANS-4857]: re-activate property change events
+        outline.getSelectionModel().addListSelectionListener(managerListener);
```
eirikbakke, line 884: But shouldn't there be at least one event processed in the end?

Might it be better to do outline.getSelectionModel().setValueIsAdjusting(true) at the beginning and setValueIsAdjusting(false) at the end? Does this work?

MKuettner: The ExplorerManager.PROP_SELECTED_NODES property change event is already fired before _synchronizeSelectedNodes(..)_ method is called. So i think there is no need for any more property change event within the synchronize method.

But you are right setting _setValueIsAdjusting(true/false)_ is a much better solution. 
I just tested this and it works fine for me.

JaroslavTulach (changes requested):
Can I see a unitest? CCing @timboudreau

JaroslavTulach (approved):
A change with a test is always better:
* the test demonstrates the failure before the fix
* the test proves the fix "fixes something" (at least the test)
* the test prevents regressing to previous behavior in the future

You have my "go". Thanks for contributing to NetBeans.

eirikbakke (approved):
Thanks for all the tests!

eirikbakke:
@MKuettner One technicality before we can merge this--could you ensure that the email address you want associated with this commit is visible on your Github profile (as a "public email" in the "profile" section of your settings)?

(When we merge the commits into the NetBeans codebase, they will become a single commit with author information taken from your github profile.)

MKuettner:
@eirikbakke I changed my profile settings so my email address should be visible now.
Thanks for approving!

PR #2411, opened 2020-10-01
[NETBEANS-4867] Fix inconsistent Gradle SubProject Detection

lkishalmi:
In some cases (like javafx base module) the Gradle plugin fails to detect the project directory using heuristics. Caching sub project info aside of project info makes that information early accessible for project manager, also flushing non-project dir cache makes sure that the subproject can be detected after a project reload.

This PR a bit more than that, it contains disk cache abstraction out of GradleProjectCache. That would be used later for caching other kind of Model information also caching multiple model result.


matthiasblaesing:
It seems this broke maven projects: https://issues.apache.org/jira/projects/NETBEANS/issues/NETBEANS-4885

PR #2413, opened 2020-10-01
[NETBEANS-4819] Applied Emilian Bold's patch that fixes missing time-...

SirIntellegence:
…43213.netbeans.org/bugzilla/attachment.cgi?id=161749

### DIFF
--- a/nbbuild/templates/common.xml
+++ b/nbbuild/templates/common.xml
@@ -82,6 +82,8 @@
             <available file="${nbplatform.active.dir}/platform/core/asm-8.0.1.jar"/>
         </condition>
         <property name="asm.jar" location="${platform/libs.asm.dir}/core/asm-8.0.1.jar"/>
+        <property name="tsaurl" value=""/>
+        <property name="tsacert" value=""/>
     </target>
 
     <target name="-release.dir">
@@ -467,7 +469,7 @@
                  locales="${nbm.locales}"
                  >
             <license file="${license.file.override}"/>
-            <signature keystore="${keystore}" storepass="${storepass}" alias="${nbm_alias}"/>
+            <signature keystore="${keystore}" storepass="${storepass}" alias="${nbm_alias}" tsaurl="${tsaurl}" tsacert="${tsacert}"/>
             <updaterjar>
                 <pathfileset>
                     <path refid="cluster.path.id"/>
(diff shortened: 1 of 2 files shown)


SirIntellegence:
That is odd... I didn't even touch that stuff. Is the current master not valid?

matthiasblaesing:
Because the patch falls into the "trivial" category, I think it is ok to integrate. BUT the commit message can't stand. A commit message should indicate what the change intents, if you insist on the reference, I question whether you consider the patch trivial. If the patch is not trivial, the question is: Are you the author? If not this can't be interated.

SirIntellegence:
Unfortunately not. I think it might have been @emilianbold's patch, but I am not sure. I can confirm that the changes work though. Sorry about the commit message. Do I just change the message locally and force push? I am not too familiar with git yet.

SirIntellegence:
One of the hunks did fail to apply, so I did make some changes to fix that if it is applicable.

nbbuild/templates/common.xml
```diff
@@ -82,6 +82,16 @@
             <available file="${nbplatform.active.dir}/platform/core/asm-8.0.1.jar"/>
         </condition>
         <property name="asm.jar" location="${platform/libs.asm.dir}/core/asm-8.0.1.jar"/>
+       <condition property="tsaurl" value="">
```
lkishalmi, line 85: These conditionals is not needed. Use ```<property name="tsaurl" value=""/>``` instead.
Ant property processing rule is: first set value matters.

SirIntellegence:
That commit was just changing commit text.

matthiasblaesing:
@SirIntellegence please check the comment from @lkishalmi - he's right.

SirIntellegence:
Right, sorry. Oddly enough, that was the one part of the patch that didn't just apply. Just a moment while I take care of that.

SirIntellegence:
There you go. Just a random suggestion, is it possible for you to override the default formatting settings on the projects to use spaces instead of tabs? I generally use tabs and find myself needing to use an external editor if I don't want to change that setting since tabs always get inserted.

matthiasblaesing:
Ok, the patch looks clean, it is simple enough to be considered trivial (parameter passing), Emilian is an apache committer, unittests come back clean. I tested a build and a timestamp was reported from jarsigner after I build with a tsaurl. So lets get this in.

PR #2415, opened 2020-10-02
Check Javadoc build using GitHub action

hectorespert:
Add GitHub action job to check errors on Javadoc generation.


JaroslavTulach (approved):
More automated tests that don't slow down existing checks over 30 minutes is good.

PR #2417, opened 2020-10-02
(Re)enable building netbeans offline

matthiasblaesing:
This PR reestablishes the possibility to build netbeans offline. Offline in this case means, that thirdparty dependencies can be downloaded once and after that is done, the build can be done without a network connection.

This is the sequence:

0. Change into the netbeans source directory
1. Download release properties: `wget -O ../netbeansrelease.json "https://gitbox.apache.org/repos/asf?p=netbeans-jenkins-lib.git;a=blob_plain;f=meta/netbeansrelease.json"`
2. Download external binaries: `ant download-all-extbins`
3. Run the build: `ant -Dmetabuild.jsonurl=file:../netbeansrelease.json`

Steps 1+2 need an internet connection, step 3 can be run offline.

### DIFF
--- a/enterprise/javaee.wildfly/licenseinfo.xml
+++ b/enterprise/javaee.wildfly/licenseinfo.xml
@@ -21,11 +21,14 @@
 -->
 <licenseinfo>
     <fileset>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/javaee_7.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/javaee_8.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/javaee_web_services_client_1_4.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/jsp_2_2.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/jsp_2_3.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/web-app_4_0.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/web-common_3_1.xsd</file>
+        <file>src/org/netbeans/modules/javaee/wildfly/resources/web-common_4_0.xsd</file>
         <file>src/org/netbeans/modules/javaee/wildfly/resources/web-fragment_4_0.xsd</file>
         <license ref="CDDL-1.1" />
         <comment type="CATEGORY_B" />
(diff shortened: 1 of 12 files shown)


matthiasblaesing:
The intention of this PR is to ensure netbeans can be build without requiring a constant connection.

@ehsavoie: The javaee parts were problematic because some of the XSDs imported external XSDs from internet servers. In all cases either the requires XSDs were already in the source base and just needed to be moved to the right place or absolute URLs had to replaced with relative paths. 

@lkishalmi  Another problem was the building of the gradle tooling - I tested building it with the "--offline" option and that works. Do you see a problem here?

@jtulach is this what you meant in your email? Could you please give it a spin?

lkishalmi:
```--offline``` should be fine as far as I could check. Once the external binaries are downloaded everything would be locally resolvable.

neilcsmith-net:
-1 until review from Eric.  I'm dubious about this way of doing it.  Would have thought mirroring how the source release handles this might be better?  eg. netbeansrelease.properties in nbbuild/  If existence of that file doesn't stop request for the release json (haven't checked that yet) that sounds like an error that needs fixing anyway?!

ebarboni:
Hi @matthiasblaesing 
 maybe we could add a step to download-all-extbins to download the json file in correct place to avoid the wget phase.  
AFAIK download-all-extbins is not used in "standard build".

It make me think that step 3 -D parameters  is a hack as json file should be detected and used.

matthiasblaesing:
@neilcsmith-net @ebarboni my target was enabling build offline - the procedure was just to show, that it can be successfully done, without to much fuzz. You can broaden the scope, but it was not intended to be.

The approach to download the json file to some "build" directory will not work. `ant clean` will clean the build directory and your offline buildability is gone.

ebarboni:
For me this is ok to merge

neilcsmith-net:
> The approach to download the json file to some "build" directory will not work. ant clean will clean the build directory and your offline buildability is gone.

@matthiasblaesing I actually said `nbbuild/` not `nbbuild/build/`.  Check the source zip releases - they have two extra files in there - `gitinfo.properties` and `netbeansrelease.properties` to ensure consistent builds.  Copying those from `nbbuild/build/` to `nbbuild/` should allow this to work offline (and if it doesn't that's a bug I think).

Agree with @ebarboni this is OK to merge, but do think we should look at using the above mechanism (or extension of it) in a latter PR, in place of your instructions here.

matthiasblaesing:
Thank you for the review. Let's get this in.

JaroslavTulach:
Today I updated to most recent master and my build succeeded (instead of failing in `javaee.wildfly` module). Very nice surprise. Thank you.

PR #2419, opened 2020-10-03
Ant task to display generated Javadoc

hectorespert:
Add `javadoc-nb` ant task at root project level like the `javadoc-nb` ant task available at module level to display the generated javadoc.


JaroslavTulach (approved):
Maybe document it somewhere?

PR #2421, opened 2020-10-05
cache netbeanrelease.json alongside external library

ebarboni:
This PR is a complement to #2417 

ant -Dneedjsondownload=true will force the download json file otherwise it read it from binaries cache.

### DIFF
--- a/nbbuild/build.xml
+++ b/nbbuild/build.xml
@@ -109,39 +109,8 @@
       <zipfileset dir="${nb_all}/nbbuild" includes="notice-stub.txt" fullpath="META-INF/NOTICE"/>
     </jar>
     <taskdef name="createlicensesummary" classname="org.netbeans.nbbuild.extlibs.CreateLicenseSummary" classpath="${nbantext.jar}"/>
-    <!-- get all json info we have on apache gitbox  -->
-    <property name="metabuild.releasejson" value="${nb_all}/nbbuild/build/netbeansrelease.json"/>
-    <condition property="metabuild.jsonurl" value="https://gitbox.apache.org/repos/asf?p=netbeans-jenkins-lib.git;a=blob_plain;f=meta/netbeansrelease.json">
-        <not>
-            <isset property="metabuild.jsonurl"/>
-        </not>
-    </condition>
-    <condition property="needjsondownload" value="true" else="false">
-        <not>
-            <available file="${binaries.cache}/netbeansrelease.json" />
-        </not>
-    </condition>
-    <sequential if:true="${needjsondownload}" >
-        <configureproxy connectTo="${metabuild.jsonurl}" hostProperty="proxyHost" portProperty="proxyPort"/>
-        <setproxy proxyhost="${proxyHost}" proxyPort="${proxyPort}"/>
-        <get dest="${metabuild.releasejson}" skipexisting="false" src="${metabuild.jsonurl}" />
-        <copy file="${metabuild.releasejson}" tofile="${binaries.cache}/netbeansrelease.json" />
-    </sequential>
-    <copy file="${binaries.cache}/netbeansrelease.json" tofile="${metabuild.releasejson}" />
-    <!-- read info from gitinfo.properties , if present in source bundle copy gitinfo-->
-    <copy file="${nb_all}/nbbuild/gitinfo.properties" tofile="${nb_all}/nbbuild/build/gitinfo.properties" failonerror="false"/>
-    <copy file="${nb_all}/nbbuild/netbeansrelease.properties" tofile="${nb_all}/nbbuild/build/netbeansrelease.properties" failonerror="false"/>
-    <!-- get branches and git information as previous not overrided this will take new-->
-    <antcall target="getgitinformation" />
-    <!-- javadoc content filtering -->
-    <taskdef name="setjavadoc" classname="org.netbeans.nbbuild.SetApidocClustersConfig" classpath="${nbantext.jar}"/>
-    <property file="${nb_all}/nbbuild/build/gitinfo.properties"/>
-    <property file="${nb_all}/nbbuild/build/netbeansrelease.properties"/>
-    <setjavadoc branch="${metabuild.branch}"/>
-    <echo message="Building branch: ${metabuild.branch}"/>
-    <property name="releasejson" value="${nb_all}/nbbuild/build/netbeansrelease.json"/>
-    <property name="xmlrelease" value="${nb_all}/nbbuild/build/netbeansrelease.xml"/>
-    <property name="propertiesrelease" value="${nb_all}/nbbuild/build/netbeansrelease.properties"/>
+    
+    <!-- prepare task to create properties files-->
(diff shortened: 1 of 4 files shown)


neilcsmith-net:
I think we should consider pulling some of this into a separate target with conditional?

Is there any need for the json file if `netbeansrelease.properties` exists in `build`?  Not sure a source release build should ever need to touch the json file either.

Perhaps move copy earlier, and only trigger target downloading and parsing json if `netbeansrelease.properties` doesn't already exist?

matthiasblaesing:
I see the problem, that global state (binaries cache) influences the build. If you are not aware of the file, rebuilding netbeans for a newer tag will not yield the same result. Even running "git clean -f -x -d" will not clean that up.

I also think a separate target would be good - then it would be trivial to combine `download-all-extbins` and this task.

ebarboni:
I try to alter the process, but in fact download-all-extbins hide a lot in the antsrc and is offline because it check the cache in java.

Trying to embed the json in release bundle. 
If I'm right
If found use it. If you build a release you will stay with the release you have.

On a normal dev you will be on master by default, you may miss new release tag for javadoc construction (version navigation).
If you try do ant -metabuild.branch=releaseXXX (not in your local json)  ant will tell: 
 No Release Information found for branch releaseXXX, update json file section with ant -Dneedjsondownload=true

Hope it's better workflow

nbbuild/build.xml
```diff
@@ -109,48 +109,59 @@
...
+    <!-- javadoc content filtering -->
+    
+    <setjavadoc branch="${metabuild.branch}"/>
+  </target>
+  <target name="getjsonfile">
+    <!-- copy all related release properties -->
+    <copy file="${nb_all}/nbbuild/netbeansrelease.json" tofile="${nb_all}/nbbuild/build/netbeansrelease.json" failonerror="false"/>
+    <!-- read info from gitinfo.properties , if present in source bundle copy gitinfo-->
+    <copy file="${nb_all}/nbbuild/gitinfo.properties" tofile="${nb_all}/nbbuild/build/gitinfo.properties" failonerror="false"/>
+    <copy file="${nb_all}/nbbuild/netbeansrelease.properties" tofile="${nb_all}/nbbuild/build/netbeansrelease.properties" failonerror="false"/>
```
neilcsmith-net, line 145: Do we still need `netbeansrelease.properties` in `nbbuild`? Won't `gitinfo.properties` be enough to derive the correct info from `netbeansrelease.json` in a source bundle now (with change below)?

nbbuild/build.xml
```diff
@@ -1752,6 +1772,7 @@ It is possible to use -Ddebug.port=3234 -Ddebug.pause=y to start the system in d
       <zipfileset file="${source.zip.readme}" fullpath="README.md" erroronmissingarchive="false" />
       <zipfileset file="${nb_all}/nbbuild/build/netbeansrelease.properties" fullpath="nbbuild/netbeansrelease.properties" erroronmissingarchive="false"/>
       <zipfileset file="${nb_all}/nbbuild/build/gitinfo.properties" fullpath="nbbuild/gitinfo.properties" erroronmissingarchive="false" />
+      <zipfileset file="${nb_all}/nbbuild/build/gitinfo.properties" fullpath="nbbuild/netbeansrelease.json" erroronmissingarchive="false" />
```
neilcsmith-net, line 1775: Wrong file name?!

nbbuild/build.xml
```diff
@@ -141,7 +119,49 @@
...
+    <condition property="needjsondownload" value="true" else="false">
+        <not>
+            <available file="${nb_all}/nbbuild/build/netbeansrelease.json" />
+        </not>
+    </condition>
+    <sequential if:true="${needjsondownload}" >
+        <configureproxy connectTo="${metabuild.jsonurl}" hostProperty="proxyHost" portProperty="proxyPort"/>
+        <setproxy proxyhost="${proxyHost}" proxyPort="${proxyPort}"/>
+        <get dest="${metabuild.releasejson}" skipexisting="false" src="${metabuild.jsonurl}" />
+        <copy file="${metabuild.releasejson}" tofile="${nb_all}/nbbuild/netbeansrelease.json" />
```
neilcsmith-net, line 163: This will cache it in `nbbuild/` on any build? I would have thought that's better done with explicit targets that are not invoked in normal build eg. `ant cache-release-json` and `ant clean-release-json`?  Also, probably needs adding to `.gitignore` just in case?

neilcsmith-net (commented):
A few thoughts.  Changing anything here, would be good to make sure source zips are tested with first beta too.

nbbuild/build.xml
ebarboni: yep sorry changed

nbbuild/build.xml
ebarboni: maybe ant clean-release-json pruning the nbbuild stuff maybe enough to have clean build.

Explicit caching seems a bit overkill. forking / developping netbeans dev will likely be on master or on their branches that will not be detected and will fallback to informations and branding the build to Dev

nbbuild/build.xml
ebarboni: I'm not sure, netbeansrelease.properties can be generated of course but it contains the branding information used for this particular release.

neilcsmith-net: Up to you.  My preference would be not to duplicate the information so it's a little simpler, and both source bundle and checkout do the same thing.  But it's minor.

nbbuild/build.xml
neilcsmith-net: Possibly. It seems a little weird to me that a normal ant build will cache there, but an ant clean won't remove it.  And obviously we don't want ant clean to remove it, or we remove the benefit for offline build.  Either way, pretty sure we want it git ignored though.

nbbuild/build.xml
ebarboni: I tried not duplicating.

nbbuild/build.xml
ebarboni: may still need polish but caching by default is not hurting for me. Implemented 2 explicit target by the way.

lkishalmi:
```
/home/travis/build/apache/netbeans/scratch/nbbuild/build.xml:2253: Failed Rat test(s):
Unapproved license in 1 file(s) 
https://github.com/apache/netbeans/tree/master/nbbuild/netbeansrelease.json
```

lkishalmi:
Is there anything against merging this one?

neilcsmith-net:
OK, @ebarboni  adding this to NB13 milestone as possible fix for #3456 but going to rebase on delivery when that's available so we can review and test by rc2 rather than rushing in now.

neilcsmith-net:
@ebarboni just working on squashing this and testing against requirements for #3456 Will open another PR for your review if need be.

neilcsmith-net:
@ebarboni OK, just squashed, fixed a bit of formatting and updated the commit message. Tested with building a source zip, editing gitinfo.properties and json manually, then passing fake metabuild url to ant - builds fine and no json download.  Fantastic!  Should be no need for anything else for #3456.  Should I force push to your PR or open another one?

PR #2423, opened 2020-10-06
[NETBEANS-4559] Web Project Properties Dialog fails to save Relative URL

jGauravGupta:
This is a bug fix. 
- Delegates action provider to `J2eeActionsProvider` for non-Payara Micro projects. By introducing "micro" packaging via payara-micro-maven-plugin, a separate ActonProvider will be introduced in another PR.
- Currently `-Dnetbeans.deploy.clientUrlPart` is not supported by the Payara Micro Maven plugin, will be fixed should soon in another PR.


lkishalmi (approved):
I'm not an expert in this area. Nothing hits me at the first sight. Some test would be useful, though I know, those are not trivial to write.

PR #2424, opened 2020-10-07
Modularizing Javascript/HTML

sdedic:
Adjustments that lead to better reuse of javascript / HTML support. Javascript modules are quite tangled with HTML (and vice versa) since they have been always developed in parallel. With `nodejs` and similar technologies, javascript can be used independently without embedding into HTML (PHP) web pages. This PR partially loosens the dependencies, but many of them still remain. Some minor issues / warnings were fixed during the course.

Outline of main changes:
- `HtmlIndex` API separated from the HTML Editor module. Many modules have dependencies on the Index, querying files, but not necessarily on HTML editing features. Becomes more important in headless modes.
- JavaScript HTML support separated from basic Javascript editing (just completion at the moment). This is a beginning of the effort that could de-entagle JS/HTML supports

These changes allowed to reduce module dependencies, especially non-visual modules to *.editor modules (which are more or less UI-oriented). IMHO it's good to separate infrastructure/evaluators from presentation layer - if not for better architecture, then for better testing.

I've tried to consolidate the work into several topic-oriented commits for easier review.

### DIFF
--- a/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
@@ -212,7 +212,7 @@ public final class OpenProjectList {
     }
     
     public static void waitProjectsFullyOpen() {
-        getDefault().LOAD.waitFinished();
+        getDefault().LOAD.waitFinished(0);
     }
 
     static void preferredProject(final Project lazyP) {
@@ -340,14 +340,25 @@ public final class OpenProjectList {
             progress = ProgressHandleFactory.createHandle(CAP_Opening_Projects());
         }
 
-        final void waitFinished() {
+        final boolean waitFinished(long timeout) {
             log(Level.FINER, "waitFinished, action {0}", action); // NOI18N
             if (action == 0) {
                 run();
             }
             log(Level.FINER, "waitFinished, before wait"); // NOI18N
-            TASK.waitFinished();
+            if (timeout == 0) {
+                TASK.waitFinished();
+            } else {
+                try {
+                    if (!TASK.waitFinished(timeout)) {
+                        return false;
+                    }
+                } catch (InterruptedException ex) {
+                    return false;
+                }
+            }
             log(Level.FINER, "waitFinished, after wait"); // NOI18N
+            return true;
         }
         
         @Override
@@ -580,7 +591,7 @@ public final class OpenProjectList {
 
         @Override
         public Project[] get() throws InterruptedException, ExecutionException {
-            TASK.waitFinished();
+            waitFinished(0);
             try {
                 enteredGuard.lock();
                 while (entered > 0) {
@@ -595,7 +606,7 @@ public final class OpenProjectList {
         @Override
         public Project[] get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
             long ms = unit.convert(timeout, TimeUnit.MILLISECONDS);
-            if (!TASK.waitFinished(timeout)) {
+            if (!waitFinished(timeout)) {
                 throw new TimeoutException();
             } 
             try {
@@ -682,7 +693,7 @@ public final class OpenProjectList {
         "# {0} - project path", "OpenProjectList.deleted_project={0} seems to have been deleted."
     })
     public void open(Project[] projects, boolean openSubprojects, ProgressHandle handle, AtomicBoolean canceled) {
-        LOAD.waitFinished();
+        LOAD.waitFinished(0);
             
         List<Project> toHandle = new LinkedList<Project>();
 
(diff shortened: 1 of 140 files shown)


ppisl:
I was writing HTML and JavasScript support in NetBeans probably more than for three years. What Svata is suggesting and doing was on my TODO list for long time, but as usually there were still other priorities. So I agree with Svata's PR and I'm supporting him to finish splitting the indexing from editor.

java/maven/src/org/netbeans/modules/maven/execute/AbstractMavenExecutor.java
```diff
@@ -141,15 +142,13 @@ protected final void processInitialMessage() {
...
-            @Override public void run() {
-                createNewTabActions();
-                tabContext.rerun.setEnabled(false);
-                tabContext.rerunDebug.setEnabled(false);
-                tabContext.overview.setRoot(null);
-                tabContext.resume.setFinder(null);
-                tabContext.stop.setEnabled(true);
-            }
+        updateUILater(false, () -> {
+            createNewTabActions();
```
JaroslavTulach, line 146: These are typical UI updating actions and it makes no sense to call them when the system is running in headless mode. Thanks for introducing the `updateUILater` (if ever) utility!

platform/openide.dialogs/src/org/openide/DialogDisplayer.java
```diff
@@ -133,13 +133,19 @@ public Dialog createDialog(DialogDescriptor descriptor, Frame parent) {
      * @see "#30031"
      */
     private static final class Trivial extends DialogDisplayer {
+        @Override
         public Object notify(NotifyDescriptor nd) {
+            if (GraphicsEnvironment.isHeadless()) {
+                return NotifyDescriptor.CLOSED_OPTION;
```
JaroslavTulach, line 139: Showing dialog would yield an exception in headless mode. Returning `CLOSED_OPTION` is the best this default, dummy implementation can do, in my opinion.

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
```diff
@@ -595,7 +606,7 @@ public boolean isDone() {
         @Override
         public Project[] get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
             long ms = unit.convert(timeout, TimeUnit.MILLISECONDS);
-            if (!TASK.waitFinished(timeout)) {
+            if (!waitFinished(timeout)) {
```
JaroslavTulach, line 609: This is the important change. Unit tests that call `OpenProjectList.getDefault().get()` could wait indefinitely long. Now they trigger the actual loading mechanism by calling `waitFinished` and not `TASK.waitFinished`. As a result it is more comfortable to write tests that invoke code relying on the list of open projects being loaded.

sdedic:
Updated original commits with missing licenses (for the new module).

JaroslavTulach (approved):
I am glad someone as skilled as Sváťa has found time to do cleanup like this. Making module dependencies more [lightweight](http://wiki.apidesign.org/wiki/HeavyWeight) is a good thing especially when [backward compatibility](http://wiki.apidesign.org/wiki/BackwardCompatibility) isn't compromised. Thank you for your work Sváťo!

php/php.latte/nbproject/project.xml
```diff
@@ -147,7 +147,7 @@
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
-                        <specification-version>2.19</specification-version>
+                        <specification-version>2.66</specification-version>
```
JaroslavTulach, line 150: For @junichi11 to review, but likely OK.

lkishalmi:
Impressive amount of work! Thank you for that!
I have a naive question. Would it fix the Java SE cluster dependency on the HTML5 cluster or that would be another story?

sdedic:
> Impressive amount of work! Thank you for that!
> I have a naive question. Would it fix the Java SE cluster dependency on the HTML5 cluster or that would be another story?

Not sure, it wasn't (yet) the primary goal; but I've tried to reduce dependencies from java and javascript to HTML. Can you give an example of the offending dependency ? Might be easier to get rid of it now.

php/php.latte/nbproject/project.xml
junichi11: I suppose it's OK.

webcommon/javascript2.editor/src/org/netbeans/modules/javascript2/editor/hints/ErrorCheckingSupport.java
```diff
@@ -86,42 +84,27 @@ public static void setErrorCheckingEnabledForMimetype(String mimeType, boolean e
         }
     }
 
-    public static String getMimeType(Parser.Result result) {
-        SyntaxAnalyzerResult syntax = getSyntaxAnalyzerResult(result);
-        if (syntax != null) {
-            return Utils.getWebPageMimeType(syntax);
+    public static String getMimeType(Parser.Result info) {
```
junichi11, line 87: @sdedic Could you please have a look at https://issues.apache.org/jira/browse/NETBEANS-5155

PR #2425, opened 2020-10-07
[NETBEANS-4880] Remove Derby integration from Payara Server modules

jGauravGupta:
H2 replaced Derby as the default database in Payara 5 and Derby that can be started & managed by the Payara Platform, has been removed starting from version 5.201. So Derby integration can be removed from Payara Server modules of the Apache NetBeans IDE like the creation of Derby JDBC resource and connection pool from the Apache NetBeans IDE to Payara Server.

Note: 
- This PR does not impact the derby integration of the GlassFish modules. 
- Payara Server modules initially forked from the GlassFish modules, so in this PR few common classes are removed from Payara modules and reused via GlassFish modules.


lkishalmi (commented):
I see sig file changes that is not followed by version increments in manifest.mf. That might be ok as the master automatically got a version bump after the 12.1 release. So this is just a warning.

lkishalmi (approved):
Let's give this a go.

PR #2426, opened 2020-10-07
[NETBEANS-2644] Gradle buildSrc projects are recognized and displayed in the Build Scripts node

lkishalmi:
It looks like something like this:
image

### DIFF
--- a/extide/gradle/apichanges.xml
+++ b/extide/gradle/apichanges.xml
@@ -83,6 +83,20 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="gradle-buildsrc-project">
+            <api name="general"/>
+            <summary>GradleFiles SPI has the methods to deal with the buildSrc project.</summary>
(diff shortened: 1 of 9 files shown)


hectorespert:
It would be good to add the `extide/gradle` module in Travis config to allow Travis run the tests

ebarboni (approved):
Looks sane to me (not a gradle user) and tested and documented

lkishalmi:
@hectorespert That's valid as actually this one makes some test fails.

.travis.yml
```diff
@@ -335,6 +335,7 @@ matrix:
             - ant $OPTS build
           script:
             - ant $OPTS -f extide/o.apache.tools.ant.module test
+            - ant $OPTS -f extide/gradle test
```
hectorespert, line 338: 👍

lkishalmi: Well I finally I had to remove this one, as it seems we have a circular cluster dependency on tests, so it fails. I need to fix that in a different PR. But definitely thanks for pointing this out!

PR #2427, opened 2020-10-08
Maven branding configuration API and robustness improvement

JaroslavTulach:
I am helping to build an [application on top of the NetBeans](https://github.com/JaroslavTulach/netbeans/tree/java-lsp-server-debugging-attempt2/java/java.lsp.server/nbcode) Platform and Java support (both Maven and Gradle) which has a completely different UI.

The default behavior of "reusing output tabs" in Maven doesn't suite my use-case. I'd prefer each Maven execution to request new tab. Also the "compile on save" behavior is problematic - my application doesn't have the project configuration options to turn it on/off (yet), so it is safer to just disable it altogether: this is made possible (via branding) in bac26b5

The other problem is related to debugging (and I have witnessed it in NetBeans from time to time): When you launch a debugging, but the Maven script execution fails, the system continues to listen for the debuggee to attach. If you just fix the problem and try to debug again, it doesn't work - the previous "listening" has to be stopped by the "Finish Debugging" action. Quite annoying even in NetBeans. Moreover, my application doesn't have the "Finish Debugging" action at all. I want to solve it by 68d7ceb - if the above situation is detected, the previous "listening" on the same Maven project is cancelled and new one started automatically.

### DIFF
--- a/java/maven/arch.xml
+++ b/java/maven/arch.xml
@@ -169,6 +169,25 @@
           in case reparsing starts too soon (wasting CPU) or too late (impeding editing).
       </api>
   </p>
+  
+  <p>
+      <api category="devel" group="branding" name="DEFAULT_REUSE_OUTPUT" type="export">
+          Brand the <code>DEFAULT_REUSE_OUTPUT</code> key in a 
+          <code>org.netbeans.modules.maven.options.Bundle</code> file
+          with one of the values <code>true</code> or <code>false</code>
+          to specify the default behavior of reusing output by your application.
+          Use <code>never</code> value, if the reuse shall never be done,
+          regardless of the settings value.
+      </api> 
+      <api category="devel" group="branding" name="DEFAULT_COMPILE_ON_SAVE" type="export">
+          Brand the <code>DEFAULT_COMPILE_ON_SAVE</code> key in a 
+          <code>org.netbeans.modules.maven.api.execute.Bundle</code> file
+          with one of the values <code>all</code> or <code>node</code>
+          to specify the default behavior of compile on save in Maven 
+          projects.
+      </api> 
+  </p>
+  
  </answer>
  
  <answer id="resources-file">
--- a/java/maven/arch.xml
+++ b/java/maven/arch.xml
@@ -182,7 +182,7 @@
       <api category="devel" group="branding" name="DEFAULT_COMPILE_ON_SAVE" type="export">
           Brand the <code>DEFAULT_COMPILE_ON_SAVE</code> key in a 
           <code>org.netbeans.modules.maven.api.execute.Bundle</code> file
-          with one of the values <code>all</code> or <code>node</code>
+          with one of the values <code>all</code> or <code>none</code>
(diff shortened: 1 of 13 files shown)


java/maven/arch.xml
```diff
@@ -169,6 +169,25 @@
...
+          <code>org.netbeans.modules.maven.options.Bundle</code> file
+          with one of the values <code>true</code> or <code>false</code>
+          to specify the default behavior of reusing output by your application.
+          Use <code>never</code> value, if the reuse shall never be done,
+          regardless of the settings value.
+      </api> 
+      <api category="devel" group="branding" name="DEFAULT_COMPILE_ON_SAVE" type="export">
+          Brand the <code>DEFAULT_COMPILE_ON_SAVE</code> key in a 
+          <code>org.netbeans.modules.maven.api.execute.Bundle</code> file
+          with one of the values <code>all</code> or <code>node</code>
```
sdedic, line 185: Nitpick - typo: "node".

java/maven/src/org/netbeans/modules/maven/debug/MavenJPDAStart.java
```diff
@@ -69,139 +72,126 @@
...
+            } else {
+                io.getOut().println("failed"); //NOI18N
             }
         }
-        return (String)lock[0];
+        io.getOut().println("JPDA Listening Start..."); //NOI18N
+        Future<String> future = RP.submit(() -> {
+            return startDebugger(io);
+        });
+        lastFuture = future;
```
sdedic, line 107: May need some synchronization/flush when assigning to the shared field.

lkishalmi (approved):
Seems to be fine at first sight.

java/maven/src/org/netbeans/modules/maven/debug/MavenJPDAStart.java
```diff
@@ -69,139 +72,126 @@
...
+        properties.put("name", getName()); //NOI18N
+        properties.put("jdksources", jdkSourcePath); //NOI18N
+        properties.put("baseDir", FileUtil.toFile(project.getProjectDirectory())); // NOI18N
+        if (RunUtils.isCompileOnSaveEnabled(project)) {
+            properties.put ("listeningCP", "sourcepath"); // NOI18N
+        }
+
+        final ListeningConnector flc = lc;
         
+        lastFuture = RP.submit(() -> {
```
sdedic, line 182: Flush the `lastFuture` to other threads, next `execute` will run in other than RP.

java/maven/src/org/netbeans/modules/maven/debug/MavenJPDAStart.java
JaroslavTulach: Done in 0e6accd

java/maven/arch.xml
JaroslavTulach: Thanks for spotting this, done in 869353c

sdedic (approved):
Great, thanks!

PR #2428, opened 2020-10-08
Do not tear down debugger UI in a headless mode.

entlicher:
This is a headless check, corresponding to the existing one in `engineAdded` method.


JaroslavTulach (approved):
No brainer.

PR #2429, opened 2020-10-08
Representation of frames in DebuggingView API.

entlicher:
`DebuggingView` extended with API for frames.

### DIFF
--- a/ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
+++ b/ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
@@ -29,7 +29,9 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
+import java.net.URI;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
@@ -170,6 +172,17 @@ public final class DebuggingView {
          * @return the thread display name
          */
         public abstract String getDisplayName(DVThread thread);
+        
+        /**
+         * Get the display name of the frame. It can contain more information
+         * than a frame name, like source location, etc.
+         * @param frame the frame
+         * @return the frame display name
+         * @since 2.65
+         */
+        public String getDisplayName(DVFrame frame) {
+            return frame.getName();
+        }
         
         /**
          * Get the thread icon.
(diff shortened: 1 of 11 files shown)


ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
```diff
@@ -326,6 +338,37 @@ public final void removePropertyChangeListener(PropertyChangeListener pcl) {
...
+            return 0;
+        }
+
+        /**
+         * Get the stack frames of this thread. Stack frames are provided when the thread
+         * is suspended only.
+         *
+         * @return a list of stack frames, it's empty when the thread is running.
+         */
+        public default List<DVFrame> getFrames() {
```
JaroslavTulach, line 356: Recently I wrote a [philippic against using default methods](http://wiki.apidesign.org/wiki/BinaryCompatibleDefaultMethods), but given the [one to many proximity](http://wiki.apidesign.org/wiki/Proximity) - e.g. there is very likely just a single implementation of these interfaces in the debugger impl modules - I expect no clashes and `default` methods are the easiest way to extend the interfaces.

ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
```diff
@@ -401,6 +444,51 @@ public final void removePropertyChangeListener(PropertyChangeListener pcl) {
         public DVThreadGroup[] getThreadGroups();
     }
     
+    /**
+     * Representation of a stack frame in debugging view.
+     * Nodes representing a stack frame in debugging view model should implement this
+     * interface.
```
JaroslavTulach, line 450: Missing `@since` tag.

JaroslavTulach (changes requested):
I am fine with the concept.

Please add `apichanges.xml`, `@since` tags and increase the spec version of this module. Don't forget to increase dependencies in the modules that are going to use this new API in the future as well.

ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
entlicher: Thanks for the reminder of the apichanges, etc. Updated.

ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
entlicher: Right.

JaroslavTulach (approved):

PR #2433, opened 2020-10-09
[NETBEANS-4818] Using equals instead of == in multi-module J2SE proje…

jlahoda:
…ct's tasks, as suggested in the bug.

Using equals, as suggested in the bug.


lkishalmi (approved):
Trivial, approved.

arvindaprameya:
Approved, thank you

PR #2434, opened 2020-10-09
Add support for static imports in code templates

isarthur:
This PR is my attempt to add support for static imports in code templates. I've added an additional hint `staticImport` for the Java Code Template Processor. When processed by the processor, a static import declaration and static member identifier will be added to the source code. From the user's perspective, this will work something like this:

static-import-in-code-template


java/java.editor/nbproject/project.properties
```diff
@@ -17,7 +17,7 @@
 
 javadoc.title=Java Editor
 
-spec.version.base=2.78.0
+spec.version.base=2.79.0
```
lkishalmi, line 20: What is the reason behind increasing the spec version? Are we changing something public in API/SPI?

lkishalmi (approved):
If we are changing the API/SPI then the change shall be documented. At least a @since javadoc entry shall be placed on the new code.
Other than that, this one is a really fine PR with tests! Thank you!

java/java.editor/nbproject/project.properties
isarthur: Thanks @lkishalmi for the review. I'm not sure if this change is actually an API change, and if so, where this API change should be documented (there are no apichanges.xml and arch.xml files in the java/java.editor, all changes in the production code apply only to private methods)? Should I add the since tag to private methods or not increment the specification version at all?

lkishalmi:
@isarthur I'm sorry but I have force pushed your branch with the requested changes. If Travis agrees this one will go into 12.2-beta2

PR #2436, opened 2020-10-09
Improved stability of Truffle debugger and implemented step from scripts to Java.

entlicher:
- Stability of Truffle debugger is improved, especially the initialization when a polyglot `Engine` class is detected.
- Stepping from Truffle scripts into Java code is implemented. This is important when e.g. JavaScript calls into java.
- JPDAStep API is enhanced with filters, this is necessary for user-friendly step from scripts into Java.


JaroslavTulach (approved):
Assuming the existing Truffle tests pass, everything else looks good.

PR #2438, opened 2020-10-09
NETBEANS-4893 - Added node_modules and bower_components to the option

Chris2011:
ignore files pattern in the misc tab.

It will prevent indexing those folders while using NetBeans for the
first time and opening a web project with a package.json and/or a
bower.json.

### DIFF
--- a/platform/core.ui/src/org/netbeans/core/ui/options/filetypes/IgnoredFilesPreferences.java
+++ b/platform/core.ui/src/org/netbeans/core/ui/options/filetypes/IgnoredFilesPreferences.java
@@ -42,7 +42,7 @@ final class IgnoredFilesPreferences {
     private static final String PROP_IGNORE_HIDDEN_FILES_IN_USER_HOME
             = "IgnoreHiddenFilesInUserHome";                           // NOI18N
     /** Default ignored files pattern. Pattern \.(cvsignore|svn|DS_Store) is covered by ^\..*$ **/
-    static final String DEFAULT_IGNORED_FILES = "^(CVS|SCCS|vssver.?\\.scc|#.*#|%.*%|_svn)$|~$|^\\.(git|hg|svn|cache|DS_Store)$|^Thumbs.db$"; //NOI18N
+    static final String DEFAULT_IGNORED_FILES = "^(CVS|SCCS|vssver.?\\.scc|#.*#|%.*%|_svn)$|~$|^\\.(git|hg|svn|cache|DS_Store)$|^(node_modules|bower_components)$|^Thumbs.db$"; //NOI18N
     static private String syntaxError;
 
     private IgnoredFilesPreferences() {
--- a/platform/masterfs/src/org/netbeans/modules/masterfs/GlobalVisibilityQueryImpl.java
+++ b/platform/masterfs/src/org/netbeans/modules/masterfs/GlobalVisibilityQueryImpl.java
@@ -134,7 +134,7 @@ public class GlobalVisibilityQueryImpl implements VisibilityQueryImplementation2
 
     protected String getIgnoredFiles() {
         // \.(cvsignore|svn|DS_Store) is covered by ^\..*$
-        String retval = getPreferences().get(PROP_IGNORED_FILES, "^(CVS|SCCS|vssver.?\\.scc|#.*#|%.*%|_svn)$|~$|^\\.(git|hg|svn|cache|DS_Store)$|^Thumbs.db$");//NOI18N;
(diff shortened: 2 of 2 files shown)


lkishalmi:
Seems to be trivial enough. Merging.

matthiasblaesing:
If I understand correctly, this will remove code completion for javascript, if the corresponding library is in node_modules? Or what I'm missing?

Chris2011:
@matthiasblaesing unfortunately you are right. But what speaks against it is the following case:

I have 5 - 6 Web projects open (next to a lot of others, web independent projects).
One project is the biggest, that I'm working on, it is from my company. The node_modules folder has 115.605 files and 18.964 folders. The node_modules folder is 963 mb big. It is a mono repo with nrwl-nx - Angular application. After I removed the node_modules from the ignored files pattern, NetBeans started indexing. This project, needed round about 1h to finish and the web projects needed round about 1 more hour. So all in all it ~ 2h for indexing/background scanning the node_modules folder. Which seems waaaaaay to long.

My environment is a Surface Laptop 3, Windows 10, latest version with an i7, 16GB RAM and a M2 SSD. Afaik, not even Windows need that much time to indexing 100k files. I know this is another case but there is also another application called everything: https://en.wikipedia.org/wiki/Everything_(software) It took round about 30s - 1m to index the whole SSD with 3.638.195 files on it. It is written in C/C++ and assembler. Just to let you know, that I don't understand, why NetBeans took so long so as you see it gets faster. Btw Windows Defender is disabled, so this can't be the problem.

So long story short, whether there is possible to be as fast as the everything tool, which would be pretty awesome in any case and second, if not possible my solution for indexing the node_modules would be to read out all packages from the package.json and bower.json, that should be fast enough I guess. Also, indexing all modules, that are used by require() and the ES 6 modules with import etc. So that would be a good improvement to still have code completion for all imported modules and maybe sub modules like incrementialy index.

Is one of those solutions possible? IMHO, it is not needed to index all files in node_modules. Or someone knows how VS Code and WebStorm does it. WebStorm is also fast enough for my point of view. It just needed ~2min to index the company project which is quite good.

matthiasblaesing:
@Chris2011 so this is a big hammer that helps you, but might be a problem for "normal" JS users. 

I can explain why indexing in netbeans is slow: Because the JS indexer tries to deal with the mess, that is the JS ecosystem. The JS files are parsed in context and that is a dramatic slowdown.

By your description your "fix" basicly breaks all index based support for every JS project, that uses npm to manage its dependencies.

Chris2011:
what is a "normal" js user? If you use npm you install your dep and it installs more deps, etc. The other 4 projects are not that big but still have > 20k files inside of it and those are normal JS projects created with Vue CLI or Angular CLI etc. So some other people complained about that in Stack Overflow too "How can I prevent node_modules from indexing in NetBeans". And as I said, why does WebStorm not has this problem with JS parsing?

> I can explain why indexing in netbeans is slow: Because the JS indexer tries to deal with the mess, that is the JS ecosystem. The JS files are parsed in context and that is a dramatic slowdown.

Just in NetBeans, not in any other major IDEs/Editor. I tried it couple of minutes ago with WebStorm, the result is significant and this is what I expect for NetBeans too. WebStorm also indexes the node_modules folder as I can see, when I open a web project, what my company project was. So as I understand it right, just the JS parser in NetBeans is slow and not a JS parser in general.

Anyway, thx for the hint with the indexing we can revert it, but we need a solution for that. Here again:

* Make the JS paraser as fast as possible to make it comparable with other major IDEs. 2h is to long and it is even possible.
* Just index the packages from the package.json and bower.json because those are the packages that are used via require() and import
* Make the indexing in NetBeans as fast as possible and take a look into the everything tool.

Yes I can understand that not just indexing, but also parsing is done here. So everything could be just a hint for indexing and not for parsing.

Chris2011:
So everyne who uses npm and a major cli to create a project is just a "normal" JS user BTW.

emilianbold:
I wonder how parallel the indexer is. I know we have multithreading there
but I forgot how much: an i7 means 8 threads at least.

It would be very cool to have pre-computed indexes which are just
downloaded for a node module. Then there's no CPU waste, just bandwidth.

In an alternate universe it would be very cool to look into the indexing
infra and decouple it a bit, maybe throw it to a separate process, etc.
Lots of indexing problems with other languages too.

--emi

matthiasblaesing:
@Chris2011 you are barking at the wrong tree:

a) I'm not the author of javascript and/or javascript2 implementation in JS and even if I were, they were conceived in a different time
b) Yes I noticed the problem too
c) Breaking index based JS features is not a solution, but just another bug

You could run a profiler an see where the time is burned. If I remember correctly the biggest problem was the support for one special framework.

Chris2011:
@matthiasblaesing you are right, so let us revert the change :)

PR #2439, opened 2020-10-10
[NETBEANS-4895] Improve archive detection of a non-existent file.

lkishalmi:
Well, this one will be required to javafx internal classpath resolution to work.


matthiasblaesing (approved):
I don't know the details of the classpath handling. But the "guessing" whether a virtual file is an archive looks saner this way. Looking at the netbeans directories it is not that uncommon to have dots in a directory name.

PR #2443, opened 2020-10-11
Netbeans CI using Apache Jenkins infra

hectorespert:


hectorespert:
Jenkins, re test this please

nbbuild/jenkins/github/Jenkinsfile.groovy
```diff
@@ -0,0 +1,103 @@
...
+                wrap([$class: 'Xvfb']) {
+                    sh "ant commit-validation"
+                }
+            }
+        }
+        
+        stage("Run harness unit tests") {
+            steps {
+                wrap([$class: 'Xvfb']) {
+                    sh "ant test -Dtest.type=test-unit -Dtest.modules=harness/* -Dtest-unit-sys-prop.ignore.random.failures=true"
```
jglick, line 82: As a matter of course I would recommend using `'` rather than `"` which allows for Groovy string interpolation, which is unwanted for `sh` steps.

nbbuild/jenkins/github/Jenkinsfile.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+    triggers {
+        issueCommentTrigger('.*test this please.*')
+    }
+ 
+    environment { 
+        ANT_OPTS = '-Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json'
+    }
+
+    stages {
```
lkishalmi, line 40: Can't these run in parallel?

jglick: To be useful, that would require separate checkouts, and separate agents.

Since I presume `commit-validation` is by far the longest part, it would not buy you much to parallelize.

nbbuild/jenkins/github/Jenkinsfile.groovy
```diff
@@ -0,0 +1,87 @@
...
+                wrap([$class: 'Xvfb']) {
+                    sh 'ant commit-validation'
+                }
+            }
+        }
+    }
+ 
+    post { 
+        always { 
+            junit testResults: 'nbbuild/build/rat/*.xml,nbbuild/build/verifylibsandlicenses.xml,**/test/*/results/TEST*.xml,**/sigtest/check-release/*.xml', healthScaleFactor: 0.0
```
jglick, line 82: Why not
```suggestion
            junit 'nbbuild/build/rat/*.xml,nbbuild/build/verifylibsandlicenses.xml,**/test/*/results/TEST*.xml,**/sigtest/check-release/*.xml'
```

? Are you deliberately suppressing test results or something?

jglick (commented):
**Checks** seems to be confused, perhaps by all the force-pushes. Hard to follow what went on here.

hectorespert:
> **Checks** seems to be confused, perhaps by all the force-pushes. Hard to follow what went on here.

Sorry, I've been testing some changes

ebarboni:
hi @hectorespert you should discard old build maybe
 options {
            buildDiscarder(logRotator(numToKeepStr: '2')) 
}

nbbuild/jenkins/github/Jenkinsfile.groovy
jglick: Oddly, seems that running sigtests takes _longer_ than commit validation. Seems like something is wrong there.

jglick: Remember that if you break up targets into `stage`s, you are potentially rerunning the same dependent targets and preventing Ant from using its dependency graph effectively. Generally recommend minimizing the total number of `sh` steps if possible: ideal to do the whole build in one line. At least compare times to see if that is contributing overhead.

hectorespert:
> hi @hectorespert you should discard old build maybe
> options {
> buildDiscarder(logRotator(numToKeepStr: '2'))
> }

`buildDiscarder` added

jglick:
Despite often helping to maintain it, Ant is so rarely used these days that I totally forgot to suggest https://plugins.jenkins.io/ant/

```groovy
withAnt {
  sh 'ant …'
}
```

which will improve display in the standard console log.

nbbuild/jenkins/github/Jenkinsfile.groovy
hectorespert: I prefer to run ant tasks in stages for a better debug when it fails but I change the order to prevent run build task twice.
The other tasks doesn't depend to buid task.

hectorespert: Build task is required but is only executed in his step, the other tasks doesn't rerun any task except bootstrap or init task.

hectorespert:
Jenkins test this please

ebarboni:
@hectorespert that at the end is intended to replace GA and travis  to check for PR ?

I would suggest removing the content of the Jenkinsfile.groovy and make it call a new libraries (example here:
https://github.com/apache/netbeans/blob/master/nbbuild/jenkins/Jenkinsfile.groovy)

And you can add the real content here.
https://github.com/apache/netbeans-jenkins-lib/tree/master/vars

That may allow "independent" tweaking of the job and may save your time adjusting.

jglick:
Libraries are for sharing Pipeline script across projects. I would especially warn against loading a library from an implicit `master` branch as it means that pushes to another repo could cause builds to suddenly start breaking in this one, or conversely that PRs to this repo would fail until another repo was updated. Better to keep the Pipeline script here, so that for example a refactoring to `build.xml` to change Ant target names could simply make the corresponding change to `Jenkinsfile`.

ebarboni:
I see, on the particular case of netbeans I used it to avoid always  copying pipeline modification because we build release branch to have javadoc menu updated on prior release (for forward navigation). 
But for PR seems more tricky.

hectorespert:
> @hectorespert that at the end is intended to replace GA and travis to check for PR ?

Yes, it was a try to start replacing GA and Travis due the limitations of these.

hectorespert:
For PRs, if the Jenkins file is defined in project, it allows to adjust o improve the pipeline in an specific branch before merge it, like the travis.yaml config and GA config files.

ebarboni:
@hectorespert, maybe for testing/adjusting you may use empty https://github.com/apache/netbeans-temp (@JaroslavTulach created it but maybe we can recycle ) and configure jenkins to see how it may scale behave on several parell PR / master

ebarboni:
@mbien I think the new way with GA may allow to close this one. travis is out.

mbien:
> @mbien I think the new way with GA may allow to close this one. travis is out.

@ebarboni agreed, closing.

PR #2445, opened 2020-10-11
Adding breadcrumbs for LSP languages.

jlahoda:
(This is an LSP-only client change.)


matthiasblaesing (approved):
Did a quick test with a sample typescript project and found that it worked quite well. I looked over the code and looked sane to me. I left two inline comments, which might improve the code, but are not critical.

junichi11:
@jlahoda Could you please have a look at https://issues.apache.org/jira/browse/NETBEANS-4976 ?

PR #2448, opened 2020-10-12
Revert "NETBEANS-4893 - Added node_modules and bower_components to the option"

lkishalmi:
Reverts apache/netbeans#2438


lkishalmi:
Let's revert it then.

PR #2451, opened 2020-10-13
NETBEANS-4533 : use derive font bold for sorting in ETableHeader

adrienDG:


adrienDG:
Hi @lkishalmi here is the new PR

lkishalmi:
Seems to be trivial enough.

PR #2453, opened 2020-10-13
Bugfixes for VSCode and Gradle

JaroslavTulach:
Invoking the "Java: Compile Workspace" on a Maven project prints empty lines. Removing `"\n"` lines altogether. The EOL character is implicitly present in the previous line.

Avoiding usage of `invokeLater` when in headless mode in Gradle: edc2ff9

Branding API (like the [one in Maven](https://github.com/apache/netbeans/pull/2427)) to disable output reuse: e29dfe1

### DIFF
--- a/extide/gradle/apichanges.xml
+++ b/extide/gradle/apichanges.xml
@@ -83,6 +83,20 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="reusetabs">
+            <api name="general"/>
+            <summary>Default for reusing tabs</summary>
+            <version major="2" minor="5"/>
+            <date day="15" month="10" year="2020"/>
+            <author login="jtulach"/>
+            <compatibility semantic="compatible"/>
+            <description>
+                <p>
+                    Introducing <a href="@TOP@architecture-summary.html#group-branding">branding APIs</a>
+                    to configure default value for <b>reusing tabs</b> in Gradle projects.
+                </p>
+            </description>
+        </change>
         <change id="gradle-buildsrc-project">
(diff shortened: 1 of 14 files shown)


dbalek (approved):
Looks fine.

JaroslavTulach:
Laszlo, I have a question about `JavaActionProvider`. Why does it not run/debug single file?
```
   case COMMAND_DEBUG_SINGLE:
                            case COMMAND_RUN_SINGLE:
                                if (RunUtils.isAugmentedBuildEnabled(project)) {
```
unless the project is augmented? Btw. what does it mean "augmented"?

lkishalmi:
Well run and debug single requires a runSingle taskwhich would take runClassName and runArgs. That task is injected by the NetBeans Gradle Tooling if build augmentation is enabled.
Well the situation can be improved to enable those actions if we have a runSingle task defined in the project. I'm going to create a PR for that one.

JaroslavTulach:
I see. Gradle tooling does augmentation! I was wandering why I cannot find `runSingle` when launching gradle on the command line. The augmentation approach defeats the idea of NetBeans always doing internally the same thing as from command line and it beats the purpose of printing the command line at the beginning of output. It makes no sense to copy it to terminal.

> we have a runSingle task defined in the project

If we have this, then I have to modify the [testing project](https://micronaut.io/launch/) to have `runSingle` action.

Or I need to find a way to open a project and make it augmented.

lkishalmi:
Well, I thought that the truth in the augmented command line would freak out an average user. Though it provides NetBeans binding for runSingle and deploy for web projects. 

I'm almost ready with the PR that accepts runSingle task if that's registered.

PR #2454, opened 2020-10-13
Upgrade payara-micro-maven-plugin to v1.3.0

jGauravGupta:
https://mvnrepository.com/artifact/fish.payara.maven.plugins/payara-micro-maven-plugin is used by pom.xml's freemarker template in Payara Micro modules which is generated during conversion of simple maven web application to Payara Micro maven application. The version upgrade is in the template file.


matthiasblaesing (approved):
Looks sane to me.

PR #2455, opened 2020-10-13
Fix: Multiple CustomIndexer with same Mime-Type use the priority of the IndexerFactory + Tooltip of columns in result-table

ses1112:
If you have multiple CustomIndexer with the same Mime-Type and you override the method "getPriority" in the corresponding IndexerFactory, this priority is ignored/never used.
So there is the bug, that sometimes it can be that this CustomIndexers are executed in the wrong order.

If you execute a query on the db, a result-table appears. There you can hover about the column-name und you can see some additional information about the structure of the column. If the column is of the type string, it also shows the attribute "length" with the precision of the column. This doesn't make sense, because by string there is no precision.
It occured with MariaDB and with a column of the type char(36).

### DIFF
--- a/ide/db.dataview/src/org/netbeans/modules/db/dataview/util/DataViewUtils.java
+++ b/ide/db.dataview/src/org/netbeans/modules/db/dataview/util/DataViewUtils.java
@@ -190,10 +190,11 @@ public class DataViewUtils {
 
         if (isString(column.getJdbcType())) {
             strBuf.append("<tr> <td>&nbsp;").append(NbBundle.getMessage(DataViewUtils.class, "TOOLTIP_column_length")).append("</td> <td> &nbsp; : &nbsp; <b>");
+            strBuf.append(column.getDisplaySize()).append("</b> </td> </tr>");
         } else {
             strBuf.append("<tr> <td>&nbsp;").append(NbBundle.getMessage(DataViewUtils.class, "TOOLTIP_column_precision")).append("</td> <td> &nbsp; : &nbsp; <b>");
+            strBuf.append(column.getPrecision()).append("</b> </td> </tr>");
         }
-        strBuf.append(column.getPrecision()).append("</b> </td> </tr>");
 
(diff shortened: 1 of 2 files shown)


matthiasblaesing (changes requested):
I verified the database issue and the change looks sane. 

The change to the indexer sound sane but would get preferably a test to verify the change.

The changes can't be merged as is. PRs should be grouped by theme/area, which is not the case here. In addition we need to track contributions and record keeping is done in jira. Please create an issue on the netbeans jira (https://issues.apache.org/jira/secure/Dashboard.jspa) and reference the issue. It is referenced by fomatting the subject line of the commit as "[NETBEANS-XXXX] YYYY", where XXXX is the issue number and YYYY is the summary.

lkishalmi:
@ses1112  please address the requested changes by EOB tomorrow (PDT). Otherwise I have to move this PR out of the 12.2 release train.

neilcsmith-net:
Still marked as changes requested from @matthiasblaesing and needs the base branch changed to master if to be merged prior to freeze.

ses1112:
> 
> 
> Still marked as changes requested from @matthiasblaesing and needs the base branch changed to master if to be merged prior to freeze.

I don't have the knowledge to write a test.

matthiasblaesing:
@ses1112 block results from this:

> The changes can't be merged as is. PRs should be grouped by theme/area, which is not the case here. In addition we need to track contributions and record keeping is done in jira. Please create an issue on the netbeans jira (https://issues.apache.org/jira/secure/Dashboard.jspa) and reference the issue. It is referenced by fomatting the subject line of the commit as "[NETBEANS-XXXX] YYYY", where XXXX is the issue number and YYYY is the summary.

This is independed of a test. The changes themselves are fine, this is just a matter of getting the tracking right.

ses1112:
Thanks for your help, i created two issues and two pull requests. For a better overview, i close this request.

PR #2458, opened 2020-10-15
[NETBEANS-4908] Enable run/debug single when the suitable task is provided by the Gradle project.

lkishalmi:
Well it is a bit more than I thought at first. At the start I would have checked only the existence of runSingle task, though if I remember someone would have liked run task defined for certain executable class in his project. With this pr run.single action can be redefined to use a ```run${selectedClassName}``` task.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/spi/actions/ProjectActionMappingProvider.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/spi/actions/ProjectActionMappingProvider.java
@@ -23,6 +23,20 @@ import java.util.Set;
 import org.netbeans.modules.gradle.api.execute.ActionMapping;
 
 /**
+ * An implementation of this interface can be registered in the Gradle project
+ * lookup. Gradle support already has an implementation using {@link GradleActionsProvider}-s.
+ * <p>
+ * This interface also can be used to query the Gradle command line arguments used
+ * for a specific action. Like:
+ * <pre>
+ *     Project project = some Gradle project
+ *     String action = some IDE action
+ *     Lookup context = Lookups.singleton(project); // The context of the action, it is advised to have the project in the context.
+ *     ProjectActionMappingProvider pamp = project.getLookup().lookup(ProjectActionMappingProvider.class);
+ *     ActionMapping actionMapping = pamp.findMapping(action);
+ *     GradleCommandLine cli = new GradleCommandLine(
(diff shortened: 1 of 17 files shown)


extide/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/RunUtilsTest.java
```diff
@@ -70,4 +72,15 @@ public void testIncludeOpenProjects2() throws Exception {
         assertEquals(0, params.size());
     }
 
+    public void testEvaluateArgs1() throws Exception {
+        FileObject a = createGradleProject("projectA",
+                "apply plugin: 'java'\n", "");
+        Project prjA = ProjectManager.getDefault().findProject(a);
+        ProjectTrust.getDefault().trustProject(prjA);
+        openProject(a);
+        String[] args = RunUtils.evalueteActionArgs(prjA, null, "Project: ${projectPath}${projectName}", Lookups.singleton(prjA));
```
JaroslavTulach, line 81: Typo in `evalueteActionArgs`.

extide/gradle/src/org/netbeans/modules/gradle/spi/actions/ProjectActionMappingProvider.java
```diff
@@ -17,13 +17,14 @@
  * under the License.
  */
 
-package org.netbeans.modules.gradle.actions;
+package org.netbeans.modules.gradle.spi.actions;
 
 import java.util.Set;
 import org.netbeans.modules.gradle.api.execute.ActionMapping;
 
 /**
```
JaroslavTulach, line 25: Consider adding better Javadoc. Description/example of "where/how to register this interface" was something I'd have expected here.

JaroslavTulach (approved):
Please fix the typo. Otherwise OK.

However I am not sure how can I benefit from this change in the `java.lsp.server` and/or the associated `nbcode` suite? Preferably the language server support is to be project type independent. Clearly `java.lsp.server` shall not depend on Gradle. But of course, I can add an integration module like `nbcode/gradleintegration` to the suite, if that helps to achieve our goals.

extide/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/RunUtilsTest.java
lkishalmi: Thanks for spotting the typo!
Well, I do not know if this support ```java.lsp.server``` in any way. That would be your call to decide.

PR #2459, opened 2020-10-15
Propagate a launch error to the client.

entlicher:
This change assures that an error during launch is propagated to the client.
The ResponseError's message is displayed in an error dialog in VSCode.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
@@ -18,17 +18,17 @@
  */
 package org.netbeans.modules.java.lsp.server.debugging.launch;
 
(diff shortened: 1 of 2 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
```diff
@@ -63,7 +63,9 @@
         Pair<ActionProvider, String> providerAndCommand = findTarget(toRun, debug);
         CompletableFuture<Void> launchFuture = new CompletableFuture<>();
         if (providerAndCommand == null) {
-            launchFuture.completeExceptionally(new CancellationException("Cannot find " + (debug ? "debug" : "run") + " action!"));
+            launchFuture.completeExceptionally(new ResponseErrorException(new ResponseError(
```
JaroslavTulach, line 66: Good idea.

JaroslavTulach (approved):

PR #2462, opened 2020-10-16
[lsp] partial fix of invalidation of breakpoints during file open

sdedic:
Since the whole document text is removed and re-added, all breakpoints were renumbered to line 1. This fix fixes case where there are no changes between saved file and the didOpen state; there's still slight chance that didOpen will report a different content to the LSP server. Fixing that will probably require some trickery in openide.text ...


JaroslavTulach (approved):

lkishalmi (commented):
Looks fine.

PR #2464, opened 2020-10-16
Using Incomplete classpath to indicate unloadable Gradle project

JaroslavTulach:
When a Gradle project is initially created/opened, it usually misses its JAR libraries. It makes little sense to parse its sources. For that one can use `getFlags() => Flag.INCOMPLETE` to prevent the Java parsing infrastructure from attempting to parse the source.

Moreover the infrastructure adds a hint to the first line of the source file to show `ProjectProblems` dialog or, in case of VSCode, to build the workspace.

After the (priming) build, the project state should be re-checked and the classpath should remove its `INCOMPLETE` flag.

### DIFF
--- a/java/gradle.java/src/org/netbeans/modules/gradle/java/classpath/AbstractGradleClassPathImpl.java
+++ b/java/gradle.java/src/org/netbeans/modules/gradle/java/classpath/AbstractGradleClassPathImpl.java
@@ -27,11 +27,15 @@ import java.io.File;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import org.netbeans.api.annotations.common.NonNull;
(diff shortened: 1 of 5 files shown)


JaroslavTulach:
I can write try to write a test for this behavior, if there is some suitable infrastructure to re-use for testing behavior of Gradle Java projects.

lkishalmi (approved):
Seems to be fine! Thanks for the improvements!

lkishalmi:
About the test, I wish there would be sufficient infrastructure in the test. If a test would involve java parsing, then it would crash. However what's there might be enough to test the classpath implementation. At least it is able to load a project also with and without invoking Gradle.
Any help improving the test infrastructure would be a great help!

JaroslavTulach:
After opening a fresh [micronaut cli](http://micronaut.io/launch) project I see

image

in VSCode. Good, we can start the build. However when the build is finished, the Gradle project doesn't fully refresh - one cannot invoke "Debug File" action. Investigating.

PR #2466, opened 2020-10-17
Stack trace analyzer needs to understand modular stack traces.

jlahoda:
Since JDK 9, stack traces may contain optional module names, like:
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.enter(JavacTaskImpl.java:282)

But, the stack trace analyzer in NetBeans does not understand these lines. This patch tweaks the analyzer to understand lines with modules. It does not use the module name, though - so far does not seem to be really necessary.


JaroslavTulach (approved):

lkishalmi (approved):
Looks fine.

PR #2467, opened 2020-10-17
Read "exec.mainClass" property from POM file, before asking for the main class.

entlicher:
When the pom.xml contains the "exec.mainClass" property, use that as the main class and do not ask the user.
The "exec.mainClass" property is set in some existing pom.xml files, like Micronaut projects, for instance.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
+++ b/java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
@@ -82,7 +82,7 @@ public class RunJarPrereqChecker implements PrerequisitesChecker {
         if ((ActionProvider.COMMAND_RUN.equals(actionName) ||
                 ActionProvider.COMMAND_DEBUG.equals(actionName) ||
                 ActionProvider.COMMAND_PROFILE.equals(actionName))) {
-            String mc = null;
+            String mc = config.getMavenProject().getProperties().getProperty("exec.mainClass");
             for (Map.Entry<? extends String,? extends String> entry : config.getProperties().entrySet()) {
                 if (entry.getValue().contains("${packageClassName}")) { //NOI18N
                     //show dialog to choose main class.


JaroslavTulach:
Is `exec.mainClass` a standardized Maven property? I guess [it is](https://www.mojohaus.org/exec-maven-plugin/usage.html).

However I am not sure this kind of property reading covers all the possible cases. For example the "source level" as obtained in `MavenSourceLevelImpl` is trying to get the actual plugin parameters, not just global `pom.xml` the property. 

I know many projects that configure the [main class differently](https://github.com/jtulach/minesweeper/blob/master/client/pom.xml#L99). How does the `exec` configuration looks in your case?

Should we go with this simplistic approach, then at least please enhance the `RunJarPrereqCheckerTest` to cover your case. If we rewrite the main-class locating, we want to be sure this use-case stays supported.

One additional (diverging) idea: if there is just a single main class in the project, then the selection dialog could automatically select it (at least in the headless mode).

matthiasblaesing:
The idea sound sane, however it is only one way to declare the main class. Another variant can be found in dropwizard projects:

https://github.com/dropwizard/dropwizard/blob/master/dropwizard-archetypes/java-simple/src/main/resources/archetype-resources/pom.xml#L23

The property there is "mainClass". We could define a list of properties as candidates for the "main class", which could be ["exec.mainClass", "project.mainclass", "mainClass"] and scan them sequentially and use the first non empty value.

The thing I'd like to know: Can I override the value with the project properties dialog? There is a mainclass setting available, which should be the ultimate value.

matteodg:
Not sure looking for properties is the safest approach: what about checking the `configuration/mainClass` tag in the `exec-maven-plugin` configuration (and there resolve if any property is used) or the `configuration/archive/manifest/mainClass` tag in the `maven-jar-plugin` configuration (See [Make The Jar Executable](http://maven.apache.org/shared/maven-archiver/examples/classpath.html#Make))?

entlicher:
Thanks for the comments. I realized I need to put this into the delivery branch. Thus I'll create a new PR with the comments incorporated.

entlicher:
@matteodg I've created https://github.com/apache/netbeans/pull/2474 instead.

PR #2468, opened 2020-10-18
Annotation processors may generate class files: 

jlahoda:
-that have a name different from any source file in the source roots
-even while running in the editor

But neither of these seems to be working - fixing.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/OutputFileManager.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/OutputFileManager.java
@@ -21,14 +21,12 @@ package org.netbeans.modules.java.source.parsing;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/parsing/OutputFileManager.java
```diff
@@ -116,22 +114,28 @@ public OutputFileManager(
             File activeRoot = getClassFolderForSource(l, sibling, baseName);
             if (activeRoot == null) {
                 activeRoot = getClassFolderForApt(l, sibling, baseName);
+            }
+            if (activeRoot == null && siblings.hasSibling()) {
```
jlahoda, line 118: This is the meat of the fix - the above attempts to compute the target class folder might have failed, because, if sibling is null, it will try to find the original source root based on "baseName" - but if the newly generated classfile's name does not correspond to any source file (which is valid, of course), this search fails. So adding a new layer, using the "siblings" utility.

lkishalmi (approved):
Looks fine to me.

JaroslavTulach (approved):

PR #2469, opened 2020-10-18
[NETBEANS-4910] Correcting open and close events sent from the LSP client to the LSP server.

jlahoda:
The LSP client was not properly sending the didClose event, and when a file was re-opened, it wasn't fully registered e.g. to provide hyperlinks. Attempting to correct.


lkishalmi (commented):
Isn't this one is an API change?

jlahoda:
> Isn't this one is an API change?

Not in the formal sense - the test is using internal accessors (which was enhanced), but I don't think that formally qualifies as API change.

lkishalmi (approved):
Approved then.

PR #2470, opened 2020-10-18
[NETBEANS-4919] Avoiding dependency on javac internal classes from javadoc/.../Analyzer, using NetBeans' HTML information instead.

jlahoda:
Prevent a crash on JDK 16, which does not have the com.sun.tools.doclint.HtmlTag (internal) class.


lkishalmi (approved):
Looks Good!

PR #2471, opened 2020-10-19
Refresh unloadable Gradle/Maven project after successful build

JaroslavTulach:
Continuation of #2464 - once a build on unloadable Gradle project successfully finishes, then let's schedule a reload to bring the module to loadable state:
* opening unloadable project shall show a fix https://user-images.githubusercontent.com/26887752/96404352-fff94100-11da-11eb-931b-d6cdcbf1ed95.png 
* after invoking it, the build is started
* when finished, the project is refreshed

Ideally the fix disappears and one can invoke "Start Debugging".

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
@@ -269,7 +269,8 @@ public class ActionProviderImpl implements ActionProvider {
                     boolean canReload = project.getLookup().lookup(BeforeReloadActionHook.class).beforeReload(action, outerCtx, task.result(), out1);
                     if (needReload && canReload) {
                         String[] reloadArgs = RunUtils.evaluateActionArgs(project, mapping.getName(), mapping.getReloadArgs(), outerCtx);
-                        prj.reloadProject(true, maxQualily, reloadArgs);
+                        RequestProcessor.Task reloadTask = prj.reloadProject(true, maxQualily, reloadArgs);
+                        reloadTask.waitFinished();
                     }
                     project.getLookup().lookup(AfterBuildActionHook.class).afterAction(action, outerCtx, task.result(), out1);
                     for (AfterBuildActionHook l : context.lookupAll(AfterBuildActionHook.class)) {
--- a/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
@@ -76,6 +76,7 @@ import org.netbeans.api.project.FileOwnerQuery;
 import org.netbeans.api.project.ProjectUtils;
 import org.netbeans.modules.gradle.api.GradleBaseProject;
 import org.netbeans.modules.gradle.api.execute.RunConfig.ExecFlag;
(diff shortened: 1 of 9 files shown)


lkishalmi (changes requested):
Well, let's talk about this one.
The functionality you are looking for is there. Through it is provided by using actions, more properly ActionMapping-s. It is not the duty of the executor to re-evaluate the internal GradleProject structure, that is decided at around:
https://github.com/apache/netbeans/blob/0ec46b545cc9d1f1b5649fa2fbb735bf9dc56aa3/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java#L243

Also if you're using the built-in actions, make sure that the project is opened, that's the only way to set the NbGradleProjectImpl's AimedQuality property to FULL, which means it can turn to Gradle to fetch the information. (though if this patch is working, that means the project aims are higher than FALLBACK)

JaroslavTulach:
OK, I'll investigate the `getReloadRule()` and why it is not in action in my case.

JaroslavTulach:
The problem with the current approach is that the
```
     task.addTaskListener((Task t) -> {
                try {
                    OutputWriter out1 = task.getInputOutput().getOut();
                    boolean canReload = project.getLookup().lookup(BeforeReloadActionHook.class).beforeReload(action, outerCtx, task.result(), out1);
                    if (needReload && canReload) {
                        String[] reloadArgs = RunUtils.evaluateActionArgs(project, mapping.getName(), mapping.getReloadArgs(), outerCtx);
                        prj.reloadProject(true, maxQualily, reloadArgs);
                    }
                    project.getLookup().lookup(AfterBuildActionHook.class).afterAction(action, outerCtx, task.result(), out1);
                    for (AfterBuildActionHook l : context.lookupAll(AfterBuildActionHook.class)) {
                        l.afterAction(action, outerCtx, task.result(), out1);
                    }
                } finally {
                    task.getInputOutput().getOut().close();
                    task.getInputOutput().getErr().close();
                }
            });
```
code is invoke too late!

I'd like the reload of the project to happen immediately after the build is over. However the task listener is notified tens of seconds later.

JaroslavTulach:
d7d4faf comes with a solution that (tries to) honor the existing `getReloadRule()`. As soon as the build succeeds, the `ExecutorTask` wrapper listeners are notified and may immediately refresh the build state.

Btw. the refresh is also made synchronous and only then the build is claimed to be over.

lkishalmi (approved):
Let's give it a go! Thank you!

JaroslavTulach:
Thanks Laszlo, however I have added few more changes meanwhile, so let's review again:
* support for `ActionProgress` added to Gradle project, that's for @lkishalmi to review
* `java.lsp.server` modified to invoke `BUILD` when the project seems misconfigured - a bit of `CompletableFuture` for @sdedic and @entlicher 
* Maven project changes to refresh its state before build is over to be consistent with Gradle.

With these changes running VSCode "Start Debugging" on broken Gradle (just create new project) and broken Maven (remove some `.m2/repository` entries) seems to work.

entlicher (approved):
The changes look fine to me. Thanks.

JaroslavTulach:
OK, let's try. Everything is green (except on `java.hints` batch), so I assume the change passes existing tests. Let's integrate.

PR #2474, opened 2020-10-20
Read mainClass configuration or property. 

entlicher:
Before asking the user, we try to read main class configuration, or property from pom.xml.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
+++ b/java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
@@ -24,9 +24,13 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import javax.swing.JButton;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.model.Plugin;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.netbeans.api.java.project.JavaProjectConstants;
 import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectUtils;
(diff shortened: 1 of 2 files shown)


java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
```diff
@@ -99,6 +105,58 @@
         return true;
     }
 
+    private static String findMainClass(RunConfig config) {
+        // Read main class from the manifest property:
+        String mainClass = getConfiguration(config, "maven-jar-plugin", "archive", "manifest", "mainClass");
```
JaroslavTulach, line 110: maybe add `// NOI18N` as these texts shall never be localized.

java/maven/test/unit/src/org/netbeans/modules/maven/runjar/RunJarPrereqCheckerTest.java
```diff
@@ -63,4 +69,103 @@ public void testWriteMapping() throws Exception {
...
+                "    <version>1.0</version>\n" +
+                "    <build>\n" +
+                "        <plugins>\n" +
+                "            <plugin>\n" +
+                "                <groupId>org.apache.maven.plugins</groupId>\n" +
+                "                <artifactId>maven-jar-plugin</artifactId>\n" +
+                "                <configuration>\n" +
+                "                    <archive>\n" +
+                "                        <manifest>\n" +
+                "                            <mainClass>com.mycompany.Main2</mainClass>\n" +
```
JaroslavTulach, line 87: Does this also work with a property? E.g.  `<mainClass>${clazz}</mainClass>` - I assume it does, but it would be good to verify that.

JaroslavTulach (approved):
Thanks for the tests. Looks good from my perspective.

matthiasblaesing (approved):
I did not test this, but the approach looks sane and covers the sensible sources for defaults from my POV.

java/maven/src/org/netbeans/modules/maven/runjar/RunJarPrereqChecker.java
entlicher: Added.

java/maven/test/unit/src/org/netbeans/modules/maven/runjar/RunJarPrereqCheckerTest.java
entlicher: Yes, it does. A test added.

PR #2475, opened 2020-10-21
[NETBEANS-4925] - cleanup Map raw type warnings..

BradWalker:
Hey, it's me again.. The Netbeans Rust Remover..

There are a bunch of warnings related to raw type usage of the Map interface..

   [repeat] /home/bwalker/src/netbeans/enterprise/websvc.rest/src/org/netbeans/modules/websvc/rest/support/SourceGroupSupport.java:170: warning: [rawtypes] found raw type: Map
   [repeat]         Map result;
   [repeat]         ^
   [repeat]   missing type arguments for generic class Map<K,V>
   [repeat]   where K,V are type-variables:
   [repeat]     K extends Object declared in interface Map
   [repeat]     V extends Object declared in interface Map

This change cleans some of this up..

### DIFF
--- a/java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/SchemaElementImpl.java
+++ b/java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/SchemaElementImpl.java
@@ -343,17 +343,17 @@ public class SchemaElementImpl extends DBElementImpl implements SchemaElement.Im
                 rs = dmd.getImportedKeys(catalog, schema, tables.get(i).toString());
 
             if (rs != null) {
-                Map rset = new HashMap();
+                Map<Integer, String> rset = new HashMap();
                 String c1, c2, s1, s2;
                 while (rs.next()) {
                     if (bridge != null) {
                         rset = bridge.getDriverSpecification().getRow();
                         
                         //test references between two schemas
-                        c1 = (String) rset.get(new Integer(1));
-                        s1 = (String) rset.get(new Integer(2));
-                        c2 = (String) rset.get(new Integer(5));
-                        s2 = (String) rset.get(new Integer(6));
+                        c1 = rset.get(new Integer(1));
+                        s1 = rset.get(new Integer(2));
+                        c2 = rset.get(new Integer(5));
+                        s2 = rset.get(new Integer(6));
                         
                         if (comp(c1, c2)) {
                             if (! comp(s1, s2))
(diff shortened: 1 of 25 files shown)


java/dbschema/src/org/netbeans/modules/dbschema/jdbcimpl/SchemaElementImpl.java
```diff
@@ -592,18 +592,18 @@ private void initViews(ConnectionProvider cp, LinkedList views, DDLBridge bridge
                                     rset = bridge.getDriverSpecification().getRow();
                                     
                                     //test references between two schemas
-                                    c1 = (String) rset.get(new Integer(1));
-                                    s1 = (String) rset.get(new Integer(2));
-                                    c2 = (String) rset.get(new Integer(5));
-                                    s2 = (String) rset.get(new Integer(6));
+                                    c1 = rset.get(new Integer(1));
```
lkishalmi, line 595: Do we really need the new Integer() here?

matthiasblaesing: If we do we'll have a bigger problem. The primitive wrapper constructors will go away (they are deprecated for removal), so we need to get rid of them.

To my knowledge there is only one valid reason to use the constructors: You want an object, that you can synchronize on, that is not shared with others. But that is also flawed and will break in the future, when the primitive wrappers will be turned into value base objects, that don't have identity and don't hold a monitor.

TL;DR: `c1 = rset.get(new Integer(1));` should be `c1 = rset.get(1);`

But that might be worth fixing in another iteration.

lkishalmi: Ok, then let's have this in.
I've checked the code there are more than 5000 instances of new Integer, there also there are a few Long-s, Byte, Short-s and Chars as well. So that would be probably a few additional PR-s

BradWalker: Just wanted to respond for completeness sake, but it would be reasonable to change like @matthiasblaesing  says. But, I didn't want to "confuse" or "cloud" the issue as I'm really trying to just reduce the warning messages here..

I'll pick that up for another change request..

BTW, thanks to @lkishalmi  for committing the change!

PR #2478, opened 2020-10-21
StatusDisplayer messages remoted to LSP client.

sdedic:
StatusDisplayer messages displayed in VSCode's 'Extension Status' statusbar item - or anywhere the LSP client likes. This is different to `showMessage` defined by the LSP protocol, which typically displays alerts or notifications that require an action; status bar messaes are less obtrusive. Also different from `logMessage` that only appears in an output together with e.g. debug messages.

Relatively unrelated change: moved `Lookups.lookupWith` from the build action to an interceptor for all messages received by the LSP endpoint.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
@@ -40,4 +40,10 @@ public interface NbCodeLanguageClient extends LanguageClient {
      */
     @JsonNotification("window/showStatusBarMessage")
     public void showStatusBarMessage(@NonNull ShowStatusMessageParams params);
+    
+    /**
+     * Returns extended code capabilities.
+     * @return code capabilities.
+     */
+    public NbCodeClientCapabilities getNbCodeCapabilities();
 }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.java.lsp.server.protocol;
+
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.eclipse.lsp4j.services.LanguageClient;
+
+/**
+ * An extension to the standard LanguageClient that adds several messages missing
+ * from the official LSP protocol.s
+ * @author sdedic
+ */
+public interface NbCodeLanguageClient extends LanguageClient {
+    
+    /**
+     * Shows a message in the status bar. Log- and Info-type messages are shown "as is".
+     * The other message types can be decorated by an icon according to {@link MessageParams#getType}.
+     * The message will be hidden after specified number of milliseconds; 0 means the client
+     * controls when the message is hidden.
+     * 
+     * @param params message type and text.
+     */
+    @JsonNotification("window/showStatusBarMessage")
+    public void showStatusBarMessage(@NonNull ShowStatusMessageParams params);
+}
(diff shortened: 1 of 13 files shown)


java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -197,6 +199,9 @@ export function activate(context: ExtensionContext) {
         client.start();
         client.onReady().then((value) => {
             commands.executeCommand('setContext', 'nbJavaLSReady', true);
+            client.onNotification(StatusMessageRequest.type, (mp) => {
```
JaroslavTulach, line 202: `client.onNotification(StatusMessageRequest.type, showStatusBarMessage)` would be shorter. Shall this registration really be done in `onReady` callback, shall not it be done sooner without any callback?

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -240,6 +245,29 @@ export function activate(context: ExtensionContext) {
 
 }
 
+function showStatusBarMessage(params : ShowStatusMessageParams) {
+    let decorated : string = params.message;
+    let timeout : number = 10000;
+    
+    if (params.type) {
```
JaroslavTulach, line 252: Maybe you don't need the `if` check. `switch (undefined)` goes to default.

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -240,6 +245,29 @@ export function activate(context: ExtensionContext) {
 
 }
 
+function showStatusBarMessage(params : ShowStatusMessageParams) {
+    let decorated : string = params.message;
+    let timeout : number = 10000;
```
JaroslavTulach, line 250: Shoudn't the timeout be read from `params.timeout`, if present?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Shows a message in the status bar. Log- and Info-type messages are shown "as is".
+     * The other message types can be decorated by an icon according to {@link MessageParams#getType}.
+     * The message will be hidden after specified number of milliseconds; 0 means the client
+     * controls when the message is hidden.
+     * 
+     * @param params message type and text.
+     */
+    @JsonNotification("window/showStatusBarMessage")
+    public void showStatusBarMessage(@NonNull ShowStatusMessageParams params);
```
JaroslavTulach, line 42: Not urgent, but shouldn't there be some client/server capabilities handshake to find out if this message is supported on both sides? If this message isn't supported, then send some standard message like `logMessage` or `showMessage`? @jlahoda tends to repeat that VSCode isn't the only client for this server.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -77,14 +85,56 @@ public static void launchServer(InputStream in, OutputStream out) {
...
         }
     }
+    
+    private static Launcher<NbCodeLanguageClient> createLauncher(LanguageServerImpl server, InputStream in, OutputStream out) {
+        return new LSPLauncher.Builder<NbCodeLanguageClient>()
+            .setLocalService(server)
+            .setRemoteInterface(NbCodeLanguageClient.class)
+            .setInput(in)
+            .setOutput(out)
+            .wrapMessages(new ConsumeWithLookup(server.getSessionLookup())::attachLookup)
```
JaroslavTulach, line 95: `wrapMessages`! That's the trick. Cute.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -185,6 +236,15 @@ public WorkspaceService getWorkspaceService() {
         @Override
         public void connect(LanguageClient client) {
             this.client = client;
```
JaroslavTulach, line 238: Wouldn't it be simpler if the type of `client` field was changed to `NbCodeLanguageClient`?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ShowStatusMessageParams.java
```diff
@@ -0,0 +1,77 @@
...
+        super(type, message);
+    }
+
+    public ShowStatusMessageParams(MessageType type, String message, int timeout) {
+        super(type, message);
+        this.timeout = timeout;
+    }
+
+    @Pure
+    public int getTimeout() {
```
JaroslavTulach, line 42: The TypeScript definition says `timeout` is optional. Shouldn't this method return `Integer` and `null` if the timeout isn't set?

JaroslavTulach (approved):
Please consider some of my comments, but otherwise, great! Amazing improvement.

java/java.lsp.server/vscode/src/extension.ts
sdedic: I've read in the documentation that the registration shall be done after onReady(), Will shorten the call.

java/java.lsp.server/vscode/src/extension.ts
sdedic: victim of the code cleanup :(

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeLanguageClient.java
sdedic: Need still to learn :)

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
sdedic: If the capabilities are negotiated ?

JaroslavTulach: All the time. We can have the rich interface `NbCodeLanguageClient`, but only send messages that are allowed according to the capabilities.

sdedic:
Fixed review comments. Added capability negotiation with the client; adapted tests. Rebased on latest `delivery`

lkishalmi (approved):
Seems to be fine with me.

PR #2479, opened 2020-10-22
[NETBEANS-4930] Upgrade Gradle Tooling 6.7 with JDK-15 Support

lkishalmi:
Well, the final Gradle 6.7 just came before the branching of NetBeans 12.2. Though it seems fortunately it does not change the public API.


extide/libs.gradle/external/binaries-list
```diff
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-74D642F331F4893E34F5063EE9A152D8C29D8323 gradle-tooling-api-6.6.jar
+566891D3FE72F69C9038A921BBE6CE255A77D4A1 gradle-tooling-api-6.7.jar
```
JaroslavTulach, line 17: This change is fine, it is very likely compatible.

extide/gradle/external/binaries-list
```diff
@@ -16,4 +16,4 @@
 # under the License.
 
 A271DEA739D4EBBD1A51F282AA130EB9314AD830 gradle-wrapper-4.10.2.jar
-2564503471E1FC18A4E7D2A120BC567EA3B26AA4 gradle-4.10.2-bin.zip
+E2E8BEBFCC5BA21D6FE9A27325D10EC1DB5D83FA gradle-6.7-bin.zip
```
JaroslavTulach, line 19: Is this change really compatible?

JaroslavTulach (approved):
As far as I know Gradle team adheres to semantic versioning. E.g.:
* minor version updates are compatible
* major version updates are usually not compatible at all

I have recently witnessed the breaking change when applying annotation processors (see [NETBEANS-4848](https://issues.apache.org/jira/browse/NETBEANS-4848)). It is clear that version 6.x is not functionally compatible with 4.x. I assume there will be some consequences (Gradle/Java Frontend App broken?). Up to you to take the risk.

It is probably good to upgrade to recent version of Gradle anyway. Better do it now than doing it just before LTS version.

extide/gradle/external/binaries-list
lkishalmi: Well this one is just to compile the netbeans-gradle-tooling. It is needed to be able to debug the tooling jar inside NetBeans when using a modern JDK. The tooling jar still on Java 8 source (and with that target) level.
Basically this is just a compile-only dependency.

PR #2480, opened 2020-10-23
Search for JDK in typical location

JaroslavTulach:
It all started with a bug report: Java Language Server ignores `JAVA_HOME` environment variable. Then it continued with a discussion and the conclusion was made to mimic behavior of the existing VSCode Java extension. E.g. to search locations:

- `netbeans.jdkhome` setting (workspace then user settings)
- `java.home` setting (workspace then user settings)
- `JDK_HOME` environment variable
- `JAVA_HOME` environment variable
- current system path

`netbeans.jdkhome` is checked to give the user chance to exactly specify JDK just for NbCode. Other locations are checked for compatibility reasons.

As soon as `netbeans.jdkhome` or  `java.home` settings are set the NbCode Java part is restarted.

### DIFF
--- a/java/java.lsp.server/vscode/src/extension.ts
+++ b/java/java.lsp.server/vscode/src/extension.ts
@@ -272,20 +272,22 @@ function activateWithJDK(specifiedJDK: string | null, context: ExtensionContext,
             }
         }
 
-        // Create the language client and start the client.
-        client = new LanguageClient(
-                'java',
-                'NetBeans Java',
-                connection,
-                clientOptions
-        );
+        if (!client) {
+            // Create the language client and start the client.
+            client = new LanguageClient(
+                    'java',
+                    'NetBeans Java',
+                    connection,
+                    clientOptions
+            );
 
-        // Start the client. This will also launch the server
-        client.start();
-        client.onReady().then(() => {
-            commands.executeCommand('setContext', 'nbJavaLSReady', true);
-            client.onNotification(StatusMessageRequest.type, showStatusBarMessage);
-        });
+            // Start the client. This will also launch the server
+            client.start();
+            client.onReady().then(() => {
+                commands.executeCommand('setContext', 'nbJavaLSReady', true);
+                client.onNotification(StatusMessageRequest.type, showStatusBarMessage);
+            });
+        }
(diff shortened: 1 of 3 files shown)


java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -65,33 +65,112 @@ function findClusters(myPath : string): string[] {
...
+        let jHome: any = process.env.JAVA_HOME;
+        if (jHome) {
+            return jHome as string;
+        }
+        return null;
+    }
+
+    let currentJdk = find();
+    workspace.onDidChangeConfiguration(params => {
+        if (!params.affectsConfiguration('java') && !params.affectsConfiguration('netbeans')) {
```
Ondrej-Douda, line 92: I would rather check on full property name: `java.home` and `netbeans.jdkhome`...

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -65,33 +65,112 @@ function findClusters(myPath : string): string[] {
...
+        let newJdk = find();
+        if (newJdk !== currentJdk) {
+            onChange(newJdk);
+        }
+    });
+    onChange(currentJdk);
+}
+
+export function activate(context: ExtensionContext) {
+    vscode.extensions.all.forEach((e) => {
```
Ondrej-Douda, line 104: This asks for `vscode.extensions.getExtension('redhat.java') !== undefined`...

JaroslavTulach: OK, thanks: Done in e38771f

java/java.lsp.server/vscode/src/extension.ts
JaroslavTulach: `ConfigurationChangeEvent` doesn't seem to have a way to check that.

dbalek (approved):
Looks fine to me.

sdedic (approved):

jlahoda (approved):

JaroslavTulach:
Can we merge, Laszlo?

JaroslavTulach:
I've just noticed another integrated PR. I guess the integration is open.

PR #2481, opened 2020-10-23
Prevent from a deadlock between debugger and debuggee when breakpoint is resolved right during it's submission.

entlicher:
Breakpoints into GraalVM scripts are submitted through a call into `JPDATruffleAccessor`, which uses Truffle debugger APIs to install the breakpoint. It does not need to notify about breakpoint resolution synchronously, as it returns the breakpoint object resolved already, when it occurs synchronously. The `breakpointResolvedAccess` is for asynchronous resolution only.


lkishalmi:
Seems to be fine. BTW would it be worth to backport this one for a NetBeans 12.0-u2 ?

lkishalmi (approved):

PR #2483, opened 2020-10-24
Prevent NullPointerException from LSP HyperlinkProvider

matthiasblaesing:
The LSP HyperlinkProvider is registered for every document. It expects,
that documents always provide token sequences. This assumption is
invalid and the calls need to be guarded agains null token sequences.

Observed backtrace:

```
SEVERE [global]
	at org.netbeans.modules.lsp.client.bindings.HyperlinkProviderImpl.getHyperlinkSpan(HyperlinkProviderImpl.java:76)
	at org.netbeans.modules.lsp.client.bindings.HyperlinkProviderImpl.isHyperlinkPoint(HyperlinkProviderImpl.java:62)
	at org.netbeans.lib.editor.hyperlink.HyperlinkOperation.findProvider(HyperlinkOperation.java:266)
	at org.netbeans.lib.editor.hyperlink.HyperlinkOperation.performHyperlinking(HyperlinkOperation.java:224)
	at org.netbeans.lib.editor.hyperlink.HyperlinkOperation.keyPressed(HyperlinkOperation.java:382)
	at java.desktop/java.awt.AWTEventMulticaster.keyPressed(AWTEventMulticaster.java:258)
	at java.desktop/java.awt.AWTEventMulticaster.keyPressed(AWTEventMulticaster.java:257)
```

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/HyperlinkProviderImpl.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/HyperlinkProviderImpl.java
@@ -72,14 +72,16 @@ public class HyperlinkProviderImpl implements HyperlinkProviderExt {
         try {
             //XXX: not really using the server, are we?
             int[] ident = Utilities.getIdentifierBlock((BaseDocument) doc, offset);
+            if (ident == null) {
+                return null;
+            }
             TokenSequence<?> ts = TokenHierarchy.get(doc).tokenSequence();
+            if (ts == null) {
+                return null;
+            }
             ts.move(offset);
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
The stacktrace mentioned above was observed when editing properties files, but I was also able to reproduce it with a simple txt file. Just open a "txt" file, edit and try to save it. You will see a NullPointerException being reported.

With a debugger it can be observed, that the HyperlinkProviderImpl in the lsp.client project is invoked, tries to get the token sequence from the document, gets null and fails.

I consider this a release blocker, so this should go into delivery before we release.

lkishalmi (approved):
Looks good.

PR #2485, opened 2020-10-25
[NETBEANS-4938] Fixed Gradle Java Frontend Application generation

lkishalmi:
Well, I've made the template copy straight. It loads all three projects (root, web, desktop) also the Demo.java into the editor.

### DIFF
--- a/java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
+++ b/java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
@@ -18,59 +18,74 @@
  */
 package org.netbeans.modules.gradle.htmlui;
 
-import java.io.IOException;
+import java.io.File;
 import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
-import org.netbeans.api.project.ProjectManager;
-import org.netbeans.api.templates.FileBuilder;
-import org.netbeans.api.templates.FileBuilder.Mode;
+import org.netbeans.modules.gradle.spi.newproject.TemplateOperation;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
 import org.openide.util.MapFormat;
 
 public final class GradleArchetype {
     private final FileObject templates;
-    private final FileObject projectFo;
+    private final File rootDir;
     private final Map<String, Object> params;
 
-    public GradleArchetype(FileObject templates, FileObject projectFo, Map<String, Object> params) {
+    public GradleArchetype(FileObject templates, File rootDir, Map<String, Object> params) {
         this.templates = templates;
-        this.projectFo = projectFo;
+        this.rootDir = rootDir;
         this.params = params;
     }
 
-    public final void copyTemplates() throws IOException {
+    public final void copyTemplates(TemplateOperation ops) {
         MapFormat mf = new MapFormat(params);
         mf.setLeftBrace("${"); // NOI18N
         mf.setRightBrace("}"); // NOI18N
+        List<File> projectDirs = new LinkedList<>();
+        projectDirs.add(rootDir);
+        
         Enumeration<? extends FileObject> en = templates.getChildren(true);
         while (en.hasMoreElements()) {
             FileObject template = en.nextElement();
-            if (!template.isData()) {
-                continue;
+            String relativePath = FileUtil.getRelativePath(templates, template);
+            if (template.isFolder()) {
+                File dir = new File(rootDir, relativePath);
+                ops.createFolder(dir);
+                Object projectAttr = template.getAttribute("project"); // NOI18N
+                if (Boolean.TRUE == projectAttr) {
+                    projectDirs.add(dir);
+                }
+            } else if (template.isData()) {
+                Object packageAttr = template.getAttribute("package"); // NOI18N
+                if (packageAttr instanceof String) {
+                    String relativeParent = FileUtil.getRelativePath(templates, template.getParent());
+                    String packageName = mf.format(packageAttr);
+                    File sourceRoot = new File(rootDir, relativeParent);
+                    ops.createPackage(sourceRoot, packageName);
+                    File packageDir = new File(sourceRoot, packageName.replace('.', '/'));
+
+                    Map<String, Object> pparams = new HashMap<>(params);
+                    pparams.put("package", packageName); //NOI18N
+                    copyDataTemplate(ops, template, new File(packageDir, template.getNameExt()), pparams);
+                } else {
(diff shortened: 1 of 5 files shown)


java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
```diff
@@ -18,59 +18,74 @@
...
-import org.netbeans.api.templates.FileBuilder;
-import org.netbeans.api.templates.FileBuilder.Mode;
+import org.netbeans.modules.gradle.spi.newproject.TemplateOperation;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
 import org.openide.util.MapFormat;
 
 public final class GradleArchetype {
     private final FileObject templates;
-    private final FileObject projectFo;
```
jtulach, line 33: I always prefer the `FileObject` abstraction over using plain `File`. It makes me sad when I see code moving to the opposite direction.

What is the reason for making such change? The fact that modifications using `File` aren't notified as soon as they are made to various listeners?

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
```diff
@@ -18,59 +18,74 @@
...
-            String relativeParent = FileUtil.getRelativePath(templates, template.getParent());
-            Object packageAttr = template.getAttribute("package"); // NOI18N
-            if (packageAttr instanceof String) {
-                String packageName = mf.format(packageAttr).replace('.', '/');
-                relativeParent += "/" + packageName;
-            }
-            FileObject destinationFolder = FileUtil.createFolder(projectFo, relativeParent);
+            String relativePath = FileUtil.getRelativePath(templates, template);
+            if (template.isFolder()) {
+                File dir = new File(rootDir, relativePath);
```
jtulach, line 55: Relative path is using `/` as a directory separator. `File` is supposed to be using `File.separator`.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
```diff
@@ -18,59 +18,74 @@
...
 
-            FileBuilder fb = new FileBuilder(template, destinationFolder);
-            fb.withParameters(params);
-            fb.defaultMode(Mode.COPY);
-
-            FileObject copied = fb.build().iterator().next();
-
-            assert copied != null && copied.getNameExt().equals(template.getNameExt()) : "Created " + copied;
+    private static void copyDataTemplate(TemplateOperation ops, FileObject template, File target, Map<String, Object> params) {
+        Object importantAttr = template.getAttribute("important");
```
jtulach, line 84: The `important` attribute enhances the API, but I am not sure if the previous API was documented anywhere.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaApplicationProjectWizard.java
```diff
@@ -57,47 +53,24 @@ public HtmlJavaApplicationProjectWizard() {
...
 
         FileObject folder = ((TemplateWizard)this.getData()).getTemplate().getPrimaryFile();
-        ops.addConfigureProject(projectDir, new CopyTree(folder, projectDir, mainParams));
-    }
-
-    private static class CopyTree implements TemplateOperation.ProjectConfigurator {
-        private final FileObject templateFolder;
-        private final File projectDir;
-        private final Map<String, Object> params;
+        GradleArchetype ga = new GradleArchetype(folder, rootDir, params);
```
jtulach, line 66: This certainly looks like a simplification.

jtulach (approved):
The `collectOperations` logic has been quite simplified and that is probably the most important benefit this change provides.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
lkishalmi: The reason using File object during the template steps, as they are non-existent ones while the operation steps are collected, so in that case FileObjects would be null.
The templating has two major phase: collecting the steps to be executed and then execute the steps.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
lkishalmi: Well, that's true for the correctness, though ```/``` as separator works on every platform.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
lkishalmi: Yes that's true, though this module is not providing any public API/SPI at the moment. Maybe some next iterations when the GradleArtifact would be made available for reuse. Infortunately I have at least two branches open to improve the Gradle Wizards, but I have not got the chance to bring them in a presentable state.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
jtulach: I know that OpenVMS isn't using `/` as separator. Long time ago we even got a bug report from someone using NetBeans on such system. But let it be...

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/GradleArchetype.java
jtulach: When you think of improving the wizard, there are two things I'd suggest to address the infamous Gradle problems:
* fix the Gradle version - e.g. always generate `gradlew` - Gradle isn't trying to keep any backward compatibility. The only way to create a Gradle project and make it work after one, two, three years is to fix Gradle version with `gradlew`.
* the wizard should select (a recommended) JDK - Gradle is well known to not work with newest JDKs - the wizard should push user strongly to provide a JDK which is known to work for the project and/or selected Gradle version

Unrelated to this issue, of course. Just my 2 Kč experience of making Geertjan's [java frontend application](https://github.com/JaroslavTulach/oraclejet4j) working again after twelve (only!) months.

lkishalmi: The JDK selection is already on my branch.
I can set the wrapper on by default right now, that would be a trivial change and improve the compatibility especially as Gradle 7.0 is already on the design desk.
It seems I have to fix the tests as well.

JaroslavTulach:
Great, it works: 
obrazek

PR #2488, opened 2020-10-25
[NETBEANS-3990] Sealed hints

Akshay-Gupta-19:
Adding hints for putting sealed and permits clause on classes.


lkishalmi (changes requested):
Nice work!
This one is marked with API change, however those changes are not documented. Please add the documentation into the ```apichanges.xml```

arvindaprameya:
Approved

lahodaj (changes requested):
First, I think we need to see how fast we can compute this, and if the value we gain is big enough for the speed we get. Also, not sure if this should be a "hint" (i.e. warning), it might be enough if this is a "suggestion". Note that currently a slow hint will slow down all other hints/warnings in the editor, so I think we need to be a little bit careful when introducing potentially slow hints.

I've also made some comments in the implementation.

One more general comments - I see not tests for the changes here, that does not seem right. While I understand it might be tricky to test this specific hint, the codegen changes should be easy to test, many of the hint aspects should also be easy to test, except for not providing the hint for types that have subtypes in different named modules.

geertjanw:
Changing milestone to 12.5.

ebarboni:
comment still need to be adressed postponed to 12.6 for now

mbien:
needs a rebase, but the diff doesn't look too bad

neilcsmith-net:
@Akshay-Gupta-Oracle unresolved conversations and merge conflicts, so removing from all milestones and marking stale. Please address or close as appropriate.

matthiasblaesing:
No reply in about 6 months. @Akshay-Gupta-19 feel free to reopen if it is still relevant.

PR #2490, opened 2020-10-25
[NETBEANS-3789] Fix missing artitact nodes in Gradle Project Configurations.

lkishalmi:
Well, this one brings back the module dependencies under the configuration node in Gradle Projects.
Half of this PR allows the Gradle side of the tooling library debugged.
The GradleModuleFileCache21 is a bit odd ball here using NIO Path instead of File or File Objects, that might be reworked later, I just had it on hand from another project of mine.
Probably this PR finishes my Gradle work for 12.2.

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
@@ -99,6 +99,7 @@ class NbProjectInfoBuilder {
         model.info.project_buildDir = project.buildDir;
         model.info.project_projectDir = project.projectDir;
         model.info.project_rootDir = project.rootDir;
+        model.info.gradle_user_home = project.gradle.gradleUserHomeDir;
 
         def visibleConfigurations = configurationsToSave()
         model.info.configurations = storeSet(visibleConfigurations.collect() {conf -> conf.name})
--- a/extide/gradle/src/org/netbeans/modules/gradle/GradleProjectCache.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/GradleProjectCache.java
@@ -88,6 +88,8 @@ public final class GradleProjectCache {
     private static AtomicLong timeInLoad = new AtomicLong();
     private static AtomicInteger loadedProjects = new AtomicInteger();
 
+    private static final boolean DEBUG_GRADLE_INFO_ACTION = Boolean.getBoolean("netbeans.debug.gradle.info.action"); //NOI18N
+
     private GradleProjectCache() {
     }
 
@@ -245,7 +247,10 @@ public final class GradleProjectCache {
     private static BuildActionExecuter<NbProjectInfo> createInfoAction(ProjectConnection pconn, GradleCommandLine cmd, CancellationToken token, ProgressListener pl) {
         BuildActionExecuter<NbProjectInfo> ret = pconn.action(new NbProjectInfoAction());
         cmd.configure(ret);
-
+        if (DEBUG_GRADLE_INFO_ACTION) {
+            // This would start the Gradle Daemon in Debug Mode, so the Tooling API can be debugged as well
+            ret.addJvmArguments("-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5006");
+        }
         if (token != null) {
             ret.withCancellationToken(token);
         }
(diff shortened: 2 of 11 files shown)


extide/gradle/src/org/netbeans/modules/gradle/GradleProjectCache.java
```diff
@@ -88,6 +88,8 @@
     private static AtomicLong timeInLoad = new AtomicLong();
     private static AtomicInteger loadedProjects = new AtomicInteger();
 
+    private static final boolean DEBUG_GRADLE_INFO_ACTION = Boolean.getBoolean("netbeans.debug.gradle.info.action"); //NOI18N
```
JaroslavTulach, line 91: Isn't this an API? If so, I'd better be documented in `arch.xml` as `<arch group="systemproperty" stability="private" .../>`

extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
```diff
@@ -99,6 +99,7 @@ class NbProjectInfoBuilder {
         model.info.project_buildDir = project.buildDir;
         model.info.project_projectDir = project.projectDir;
         model.info.project_rootDir = project.rootDir;
+        model.info.gradle_user_home = project.gradle.gradleUserHomeDir;
```
JaroslavTulach, line 102: Is `gradle_user_name` also some kind of API as it is read elsewhere?

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProjectBuilder.java
```diff
@@ -125,6 +127,9 @@ void processTasks() {
 
     void processDependencies() {
 
+        File gradleUserHome = (File) info.get("gradle_user_home");
```
JaroslavTulach, line 130: Here the `gradle_user_home` value is read...

JaroslavTulach (approved):
I have to say, I don't have a good overview of what is happening. My approval is more a process approval.

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProjectBuilder.java
lkishalmi: Technically it is an API, a protocol behind a serialization between the netbeans-gradle-tooling which is actually a trojan horse that lets serialize the build information out of Gradle. On NetBeans side this data is de-serialized by the different GradleWhateverProjectBuilders. These builders then constructs public API objects from the de-serialized information.

This is one of the ugliest part of my Gradle integration. There are two reasons for implementing it this way. The ready to use Gradle Models were really not provided enough information in the days of Gradle 1.x, 2.x and 3.x series. The other one is that the remoting which would make these rich interfaces easily consumable were/are memory grubbing proxies. So I reduced the remoting to HashSet-s, HashMaps with Strings and Files.
Having that once NetBeans get the information proxies back, basically a few hashmaps, it rewrites the data into real JVM objects, so those can be serialized to disk for caching, and also can built API objects out of it. AFAIK, Attila had a nice abstraction between Gradle and NetBeans and the objects (proxies) returned by  Gradle were ready to use for NetBeans. Unfortunately they were seldom reclaimed by the GC. I guess Attila fixed that later, but I have not followed his plugin that closely.

extide/gradle/src/org/netbeans/modules/gradle/cache/ProjectInfoDiskCache.java
```diff
@@ -83,7 +86,7 @@ public QualifiedProjectInfo(Quality quality, NbProjectInfo pinfo) {
 
         @Override
         public Map<String, Object> getExt() {
-            return Collections.emptyMap();
+            return ext != null ? ext : Collections.emptyMap();
```
lkishalmi, line 89: Actually this is the primary fix. The transient data (I called that ext for whatever reason) shall be not cached to the disk, but when NetBeans process the project information it holds important data on binary, source and javadoc location which are stored in the GradleAtrifactStore. Returning an empty map when the project is not restored from serialization, made that these location data were never stored in the artifact store.

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProjectBuilder.java
```diff
@@ -242,6 +244,23 @@ void processDependencies() {
         problems.addAll(unresolvedProblems.values());
     }
 
+    private ModuleDependency resolveModuleDependency(File gradleUserHome, String c) {
+        GradleModuleFileCache21 moduleCache = GradleModuleFileCache21.getGradleFileCache(gradleUserHome.toPath());
+        GradleModuleFileCache21.CachedArtifactVersion artVersion = moduleCache.resolveModule(c);
+        Set<File> binaries = artifactSore.getBinaries(c);
```
lkishalmi, line 250: This makes the module dependency resolution more stable. If the ArtifactStore has no information on a module for whatever reason, NetBeans tries to resolve it using the Gradle Module Cache directory structure.

PR #2492, opened 2020-10-26
Avoid conflicts with other Java extensions

JaroslavTulach:
Apache NetBeans Language Server extension isn't the only Java supporting
extension. To avoid duplicated code completion and other misleading clashes
the extension disables certain functionality known to cause problems. This
behavior can be disabled by setting `netbeans.conflict.check` setting to `false`.

### DIFF
--- a/java/java.lsp.server/vscode/README.md
+++ b/java/java.lsp.server/vscode/README.md
@@ -104,3 +104,10 @@ following locations:
 - current system path
 
 As soon as one of the settings is changed, the NbCode Java part is restarted.
+
+## Conflicts with other Java Extensions
+
+Apache NetBeans Language Server extension isn't the only Java supporting
+extension. To avoid duplicated code completion and other misleading clashes
+the extension disables certain functionality known to cause problems. This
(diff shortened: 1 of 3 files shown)


JaroslavTulach:
The biggest problem this PR fixes is following "code lens" hint:
image
We want users of Apache to debug using "Java (Polyglot)" configuration. However such action isn't (yet) available as a code lens. To avoid confusion this PR disables the hint provided by [java debug](https://github.com/microsoft/vscode-java-debug) and [java test](https://github.com/microsoft/vscode-java-test) extensions.

Ideally we want such hint/lens to be available, but it is too late for 12.2. What needs to happen to make this work smoothly?
1. Java support needs to identify methods that can be executed (`main`, `@Test`, etc.)
2. * This needs to be displayed in NetBeans editor somehow
    * This needs to be displayed in VSCode as Code lens
3. When invoked by user appropriate action needs to be executed using `ActionProvider`

Anything else?

entlicher (approved):

PR #2493, opened 2020-10-26
Infrastructure to display simple confirmations/questions in LSP client.

sdedic:
I've implemented a `AbstractDialogDisplayer` that, if registered will remote `notify()` and `notifyLater` through `UIContext` into the remote LSP client. I've tested locally on Gradle projects; but it has still some limitations, like Project Problems tend to display a customizer ... so I've commented out the final registration of the service.

I'd like to pull in the infrastructure for collaboration.

### DIFF
--- a/java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDialogDisplayer.java
+++ b/java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDialogDisplayer.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.nbcode.integration;
+
+import org.netbeans.modules.java.lsp.server.ui.AbstractDialogDisplayer;
+import org.openide.DialogDisplayer;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ *
+ * @author sdedic
+ */
+//@ServiceProvider(service = DialogDisplayer.class, position = 1000)
+public class LspDialogDisplayer extends AbstractDialogDisplayer {
+}
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
@@ -155,8 +155,8 @@ public final class Server {
         }
     }
     
-    private static final RequestProcessor SERVER_INIT_RP = new RequestProcessor(LanguageServerImpl.class.getName(), 10);
-    private static final Map<NbCodeClientWrapper, Future> initializations = new WeakHashMap<>();
+    // change to a greater throughput if the initialization waits on more processes than just (serialized) project open.
+    private static final RequestProcessor SERVER_INIT_RP = new RequestProcessor(LanguageServerImpl.class.getName());
     
     
     private static class LanguageServerImpl implements LanguageServer, LanguageClientAware {
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
@@ -26,17 +26,24 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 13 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -216,7 +242,42 @@ Lookup getSessionLookup() {
             capabilities.setDocumentHighlightProvider(true);
             capabilities.setReferencesProvider(true);
             capabilities.setExecuteCommandProvider(new ExecuteCommandOptions(Arrays.asList(JAVA_BUILD_WORKSPACE, GRAALVM_PAUSE_SCRIPT)));
-            return CompletableFuture.completedFuture(new InitializeResult(capabilities));
+            return new InitializeResult(capabilities);
+        }
+        
+        @Override
+        public CompletableFuture<InitializeResult> initialize(InitializeParams init) {
```
sdedic, line 249: Note this change: some events are fired synchronously from Open Project. If the code asks for something like DialogDisplayer.notify (which some code really does), the initialization request processing would block - and the notify() round-trip would be never completed, freezing LSP client-server communication completely.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDialogDisplayer.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.netbeans.modules.java.lsp.server.ui.AbstractDialogDisplayer;
+import org.openide.DialogDisplayer;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ *
+ * @author sdedic
+ */
+//@ServiceProvider(service = DialogDisplayer.class, position = 1000)
```
JaroslavTulach, line 29: Reasonable, given we after feature freeze. This functionality will certainly be needed in the future. For 12.2 we'd rather delay using it, unless necessary.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -86,39 +109,56 @@ public static void launchServer(InputStream in, OutputStream out) {
...
+                        });
+                    } finally {
+                        DISPATCHERS.remove();
+                    }
                 }
             };
         }
     }
     
+    private static final RequestProcessor SERVER_INIT_RP = new RequestProcessor(LanguageServerImpl.class.getName(), 10);
```
JaroslavTulach, line 158: Throughput 10!? Isn't opening of project sequential anyway?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -86,39 +109,56 @@ public static void launchServer(InputStream in, OutputStream out) {
...
+    private static final Map<NbCodeClientWrapper, Future> initializations = new WeakHashMap<>();
```
JaroslavTulach, line 159: Is the `initializations` field read anywhere? I can't find it (and that'd be good, as I don't think it is needed).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/NotifyDescriptorAdapter.java
```diff
@@ -0,0 +1,322 @@
...
+                return MessageType.Log;
+        }
+    }
+    
+    /**
+     * Strip HTML from the message; VSCode standard showMessage does not support HTML.
+     * @param original
+     * @return 
+     */
+    private String translateText(String original) {
```
JaroslavTulach, line 113: I believe the translation of the text away from HTML shall only happen if it starts with `<html>`.

JaroslavTulach (approved):

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDialogDisplayer.java
lkishalmi: I'm bit confused. The PR is against delivery, but no 12.2 milestone set, I'm not among the reviewers. Am I assuming right upon this comment, that you would like to float this here in case of need? So it is tentative for 12.2? If yes, please decide it till beta3 (29th Oct, late PDT) this week. If that one is messed this one have to go to master. I've put the milestone 12.2 anyway.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
sdedic: Hmm ... OK; I thought about multiple instances that may have been initializing (multiple LSP clients), but you're right as far as there's just project open operation being waited on. Will change.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
sdedic: leftover, will remove.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDialogDisplayer.java
JaroslavTulach: Is `beta3` supposed to become VC1? Oct 29, so close! 

I asked Sváťa to propose his infrastructure changes in this PR for 12.2 as they are harmless (until we enable `@ServiceProvider`), but we may desperately need them. I am afraid we may need to download `nbjavac` if support for JDK8 is needed. Downloading shall of course only happen after asking a question - preferably through this Sváťa's infrastructure.

That's my rationale. but Oct 29, is so close!

sdedic: @lkishalmi I would added you after the PR passed through Jarda ;) it's easier to review cleaned-up code.

lkishalmi: The VC is or I'd better call it RC is due to 7th of November. Previously we had 2 week period between beta1 and beta2, I've switched it  to one week as usually the forts week in the release mode is when the most commits are happening "oh, this is a minor, but needs to be there". Also the weekly delivery -> master merge is hopefully small enough change to deal with.

lkishalmi (approved):

PR #2494, opened 2020-10-26
Where is the NbCode userdir?

JaroslavTulach:
Sváťa has just asked: where can I find `messages.log`? The answer is `$HOME/.config/Code/User/globalStorage/jlahoda.apache-netbeans-java/userdir/var/log/messages.log` but the problem is - the log isn't there!

Turns out I have deleted the `--userdir` option in one of the refactorings. This PR fixes that.

### DIFF
--- a/java/java.lsp.server/vscode/src/nbcode.ts
+++ b/java/java.lsp.server/vscode/src/nbcode.ts
@@ -59,9 +59,11 @@ export function launch(
     }
 
     let clusterPath = info.clusters.join(path.delimiter);
-    let ideArgs: string[] = [];
+    let ideArgs: string[] = [
+        '--userdir', userDir
+    ];
     if (info.jdkHome) {
-        ideArgs = ['--jdkhome', info.jdkHome as string];
+        ideArgs.push('--jdkhome', info.jdkHome as string);
     }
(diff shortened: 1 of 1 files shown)


java/java.lsp.server/vscode/src/nbcode.ts
```diff
@@ -86,7 +88,8 @@ if (typeof process === 'object' && process.argv0 === 'node') {
     }
     let clusters = fs.readdirSync(nbcode).filter(c => c !== 'bin' && c !== 'etc').map(c => path.join(nbcode, c));
     let args = process.argv.slice(2);
-    let globalStorage = path.join(os.homedir(), '.config', 'Code', 'User', 'globalStorage', 'jlahoda.apache-netbeans-java');
+    let json = JSON.parse("" + fs.readFileSync(path.join(extension, 'package.json')));
+    let globalStorage = path.join(os.homedir(), '.config', 'Code', 'User', 'globalStorage', json.publisher + '.' + json.name);
```
JaroslavTulach, line 92: While at it, let also read the `publisher` and extension `name` from the actual `package.json` file.

lkishalmi (approved):
This is fine with me.

PR #2495, opened 2020-10-27
[NETBEANS-4955] Add Support for Firefox Developer & Nightly

Vitexus:
Basic support for Firefox Developer edition & Nightly

obrazek

obrazek


matthiasblaesing:
This looks like much code duplication - could the common code be refactored into base classes, so that only the differing elements are visible (basepath, icons)?

neilcsmith-net:
Why do we need this anyway?  Surely anyone who needs this can just use the `Add` button?

Vitexus:
If is there any chance to be merged, then i'll provide updated version soon.

matthiasblaesing:
@Vitexus Neil raises a valid point. This adds much code for little gain. Till I read this PR I was not even aware that there was a Firefox developer edition and did not miss it. Based on the existing firefox entry it is trivial to add entries for developer and more variants.

mbien:
closing due to inactivity. I do also agree that this is probably too much code for the purpose of registering browser dev branches. The UI does allow registering more browsers with a few clicks.

will close jira issue too (https://issues.apache.org/jira/browse/NETBEANS-4955) since jira went read only by now.

PR #2497, opened 2020-10-27
Avoid registering the LSP client's hyperlink provider unless the given file is handled by the LSP client.

jlahoda:
This is a tweak to:
https://github.com/apache/netbeans/pull/2483

The LSP client's hyperlink provider should ignore files that are not handled by the LSP client (for performance reasons, at least), this patch moves the flag at a more correct place. Sorry I didn't have time to look at 2483, and for trouble with this bug.


matthiasblaesing (approved):
I admit I don't understand what is the connection between the hyperlink provider and the search action, but ok, the change looks sane and more important a build showed that searching works again in plain text files. Cut-n-Paste also worked. Not installing the HyperlinkProviderImpl on unsupported files should make this even saver.

Thanks.

jlahoda:
Not sure what exactly is the connection either - the thrown exception probably cancelled the action handling.

neilcsmith-net:
@jlahoda is this also the reason URL's in Java source (in Javadoc) are no longer working in 12.2-beta2?  I noticed this working yesterday - stack trace has LSP code in it.

lkishalmi (approved):

PR #2499, opened 2020-10-28
[NETBEANS-4535] Better way to detect sources and javadocs in Gradle Projects

lkishalmi:
Just to utilize the GradleModuleFileCache21 a bit more, added some code which help find the sources and javadocs more reliably.
This fixes the issue with not finding those artifacts on JavaFx projects.


JaroslavTulach (approved):

JaroslavTulach:
I've just tried to work with configurations and their sources in a project of mine and I can confirm there is a significant improvement between beta2 and beta3! 12.2 Beta3 shows all the configurations and is capable to download the sources. I was able to place there breakpoints and get the information I needed.

The only drawback I found is that Ctrl+B doesn't work. However Ctrl+O does, so it is not as big problem.

PR #2502, opened 2020-10-30
[NETBEANS-4970] Incorrect formatting for lambda functions

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-4970

Before:
```php
<?php

function test(): callable {
    return function ($test): void {
                echo "test";
            };
}
```

After
```php
<?php

function test(): callable {
    return function ($test): void {
        echo "test";
    };
}
```


junichi11:
@lkishalmi This is a regression. Can we include this change to 12.2?

junichi11:
@tmysik Let's leave merging to @lkishalmi  :)

tmysik (approved):
Thanks for the test!

lkishalmi (approved):
Looks sane, and with test!

junichi11:
Thank you for your reviews!

tmysik:
> Thank you for your reviews!

No problem, of course!

PR #2505, opened 2020-10-30
[NETBEANS-4646] Fix completionInvoke hint in code templates

isarthur:
https://issues.apache.org/jira/browse/NETBEANS-4646

Before:

completion-invoke-hint-before

After:

completion-invoke-hint-after


lkishalmi:
Nice work, though I would not add this one into 12.2 at this point of the release. please rebase it on master, so it could go early into 12.3.

lkishalmi:
I guess closed in favor of #2519

PR #2509, opened 2020-10-31
[NETBEANS-819] partial fix for cursor. Text selection does't work

hlavki:
Partial fix for incorrect HiDPI handling in terminal. This patch fixes only cursor position, not text selection.

### DIFF
--- a/ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/MyFontMetrics.java
+++ b/ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/MyFontMetrics.java
@@ -77,7 +77,7 @@ class MyFontMetrics {
      *
      * CacheFactory doles out WidthCaches.
      *
-     * These caches are 64Kb (Character.MAX_VALUE+1) and we don't really want 
+     * These caches are 64Kb (Character.MAX_VALUE+1) and we don't really want
      * Each interp to have it's own. So we share them in a map using FontMetrics
      * as a key. Unfortunately stuff will accumulate in the map. A WeakHashMap
      * is no good because the keys (FontMetrics) are usually alive. For all I
@@ -91,7 +91,7 @@ class MyFontMetrics {
      * it's reference. To make the reference count go down CacheFactory.disposeBy()
      * is used. And that is called from MyFontMetrics.finalize().
      *
-     * NOTE: The actual WidthCache's instances _will_ accumulate, but they are small and 
+     * NOTE: The actual WidthCache's instances _will_ accumulate, but they are small and
      * there are only so many font variations an app can go through. As I
      * mentioned above using a WeakHashMap doesn't help much because WidthCache's
      * are keyed by relatively persistent FontMetrics.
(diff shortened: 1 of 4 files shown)


ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/MyFontMetrics.java
```diff
@@ -142,18 +142,22 @@ public MyFontMetrics(Component component, Font font) {
         height -= leading;
         leading = 0;
 
-        cwidth_cache = CacheFactory.cacheForFontMetrics(fm);
+        cwidth_cache = CacheFactory.cacheForFontMetrics(getFm());
+    }
+
+    public FontMetrics getFm() {
+        return this.cmp.getFontMetrics(cmp.getFont());
```
sdedic, line 149: Is it necessary to store `cmp` in order to obtain `getFont` - couldn't be the font obtained / stored rather than the Component ? Not a big deal, but FontMetrics > Component seem as a backward dependency.

lkishalmi: @hlavki anything on this?

lkishalmi (approved):
Well reading the comment, it shall be good.

ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/MyFontMetrics.java
hlavki: @lkishalmi, sorry for delay. I didn't have time to fix @sdedic's suggestion. Maybe in next round. thanks

lkishalmi:
It seems this one cause a regression with the text selection in the Terminal window.

lkishalmi:
TerminalSelectionIssue

matthiasblaesing:
@hlavki could you please indicate if/when you look into the regression this change caused? I noticed this today and was irritated about the broken behavior.

matthiasblaesing:
This is now tracked as https://issues.apache.org/jira/browse/NETBEANS-5070

ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/MyFontMetrics.java
hlavki: @sdedic this is very interesting think, because in previous implementation `FontMetrics` was created in `MyFontMetrics` constructor, but it didn't work. Cursor was shifted. Would be great if somebody with strong swing knowledge could explain this behaviour.

matthiasblaesing:
For the future, please only remove whitespace on lines you change (Options -> Editor -> On Save -> All Languages -> Remove Trailing Whitespace From: "Modified lines only"). I tried to see what really changed in this PR and the netbeans diff does not automatically remove whitespace changes (or at least I don't want it to) and I had to wade through dozens of whitespace only changes, which have no meaning.

PR #2504, opened 2020-10-30
Snapshot of APIs for 12.2

neilcsmith-net:
As per mailing list thread, snapshot of APIs for 12.2 for review prior to release vote.  Built from beta 3, ideally would keep similar PR open through all betas.  To be updated if need be after beta4/vc and merged to master post release (and then cherry picked to release branch now merging not an option?).

Generated with following steps (adapted from https://issues.apache.org/jira/browse/NETBEANSINFRA-229 ) - please check that's sane too.

```
# checkout branch (or release tag)
git checkout release122
ant clean
ant build
ant gen-sigtests-release
git add -A
git stash
git checkout master
git checkout -b SnapshotOfAPIs122
git stash pop
git commit -m "Snapshot of APIs as of <TAG>"
# remove getPeer() calls
find . -name "*.sig" -exec sed -i '/java.awt.peer.ComponentPeer/{N;d;}' {} \;
# check git diff
git add -A
git commit --amend
```

PR #2510, opened 2020-11-01
Removing unwanted public class from API

sdedic:
Fixes API bug discovered by #2504 -- utility class should not be visible in the API.


sdedic:
Bad analysis, please ignore.

PR #2512, opened 2020-11-01
[NETBEANS-4469] Fix NoSuchFieldError from refactorings on JCInstanceOf.clazz which was renamed

timboudreau:
The field name of `JCInstanceOf.clazz` was changed to `JCInstanceOf.pattern` in late 2019 in the JDK sources of javac.

CasualDiff.java  already contains a method to look up the correct field via reflection, but
one case of directly referencing the original field was not changed, and causes a
NoSuchFieldError in the move inner to outer level refactoring (and likely elsewhere).

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
@@ -5895,7 +5895,7 @@ public class CasualDiff {
     }
 
     private boolean matchTypeTest(JCInstanceOf t1, JCInstanceOf t2) {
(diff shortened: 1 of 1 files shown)


lkishalmi:
Made a search in JIRA. I'm not sure though that this PR is fixing this [NETBEANS-4469](https://issues.apache.org/jira/browse/NETBEANS-4469) as the missing filed there says "pat", the same issue has been reported as [NETBEANS-4319](https://issues.apache.org/jira/browse/NETBEANS-4319)

jlahoda (approved):
Looks good. Thanks!

PR #2513, opened 2020-11-01
ioContainer in IOProvider.getIO(...) is allowed to be null.

jlahoda:
But Lookups.fixed does not allow null.

Would be ideal for 12.2, but I realize it is quite late. Would help running OpenJDK tests in the VS Code.

### DIFF
--- a/platform/openide.io/src/org/openide/io/BridgingIOProvider.java
+++ b/platform/openide.io/src/org/openide/io/BridgingIOProvider.java
@@ -22,6 +22,7 @@ import java.awt.Color;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Reader;
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.Set;
 import java.util.logging.Level;
@@ -92,9 +93,16 @@ public class BridgingIOProvider<IO, S extends PrintWriter, P, F>
     @Override
     public InputOutput getIO(String name, boolean newIO, Action[] actions,
             IOContainer ioContainer) {
+        Object[] lookupContent;
+        if (ioContainer != null) {
+            lookupContent = Arrays.copyOf(actions, actions.length + 1, Object[].class);
+            lookupContent[actions.length] = ioContainer;
+        } else {
+            lookupContent = actions;
+        }
         return new BridgingInputOutput(
                 providerDelegate.getIO(name, newIO,
-                        Lookups.fixed((Object[]) actions, ioContainer)));
+                        Lookups.fixed(lookupContent)));
     }
 
     @NbBundle.Messages({"LBL_STDOUT=Standard output"})
--- a/platform/openide.io/test/unit/src/org/openide/io/BridgingIOProviderTest.java
+++ b/platform/openide.io/test/unit/src/org/openide/io/BridgingIOProviderTest.java
@@ -19,12 +19,17 @@
 package org.openide.io;
 
 import java.awt.Color;
+import java.awt.event.ActionEvent;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
 import java.util.Set;
+import javax.swing.AbstractAction;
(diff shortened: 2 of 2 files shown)


platform/openide.io/src/org/openide/io/BridgingIOProvider.java
```diff
@@ -92,9 +93,16 @@ public InputOutput getIO(String name, Action[] actions,
...
+        Object[] lookupContent;
+        if (ioContainer != null) {
+            lookupContent = Arrays.copyOf(actions, actions.length + 1, Object[].class);
+            lookupContent[actions.length] = ioContainer;
+        } else {
+            lookupContent = actions;
+        }
         return new BridgingInputOutput(
                 providerDelegate.getIO(name, newIO,
-                        Lookups.fixed((Object[]) actions, ioContainer)));
```
JaroslavTulach, line 97: Do I read the change correctly?

Previously the `Lookup` had two elements one of the type `Action[]` and one of the type `IOContainer`. The new `Lookup` is going to have n-elements of type `Action` and optionally one element of type `IOContainer`.

platform/openide.io/test/unit/src/org/openide/io/BridgingIOProviderTest.java
```diff
@@ -78,9 +84,37 @@ public void bridgeToNullColor() throws IOException {
...
+    @Test
+    public void hasIOContainer() throws IOException {
+        MockIOP mock = new MockIOP();
+        IOProvider provider = BridgingIOProvider.create(mock);
+        Action testAction = new AbstractAction("test") {
+            @Override public void actionPerformed(ActionEvent e) {}
+        };
+        IOContainer ioContainer = IOContainer.create(new MockIOProvider());
+        InputOutput io = provider.getIO("test", false, new Action[] {testAction}, ioContainer);
+        assertSame(testAction, mock.last.lookup.lookup(Action.class));
```
JaroslavTulach, line 107: I'd expect a failure here in the previous version of the code.

JaroslavTulach (commented):
I am not sure I understand the wording: Is "IOContainer allowed to be null" or not? I don't see any comment [about null being allowed](https://bits.netbeans.org/12.0/javadoc/org-openide-io/org/openide/windows/IOProvider.html#getIO-java.lang.String-boolean-javax.swing.Action:A-org.openide.windows.IOContainer-) in the Javadoc of the method.

I assume the observation comes from a `NullPointerException` failure. Maybe sharing a stacktrace would say more than thousand of words.

As noted inline: Isn't there a change in semantics?

jlahoda:
> I am not sure I understand the wording: Is "IOContainer allowed to be null" or not? I don't see any comment [about null being allowed](https://bits.netbeans.org/12.0/javadoc/org-openide-io/org/openide/windows/IOProvider.html#getIO-java.lang.String-boolean-javax.swing.Action:A-org.openide.windows.IOContainer-) in the Javadoc of the method. That doesn't mean much, but...

There is: `@NullAllowed IOContainer ioContainer` in the javadoc. `@NullAllowed` means null is allowed, right?

> 
> I assume the observation comes from a `NullPointerException` failure. Maybe sharing a stacktrace would say more than thousand of words.
> 
> As noted inline: Isn't there a change in semantics?

Yes. Because the existing semantics is quite likely wrong. Look at the method variant that does not take IOContainer - it does: `Lookups.fixed((Object[]) actions)`. So each individual Action will be in the Lookup. Now, if I provide an IOContainer, suddenly the array of actions becomes a Lookup element on its own. That sounds unlikely to be the intent, but I admit it might be correct  - where is the expected Lookup content documented?

JaroslavTulach (approved):
> There is: `@NullAllowed IOContainer ioContainer` in the javadoc.

;-) Who pays attention to these annotations? I searched for "ioContainer - parent container accessor" and there I saw no "can be null" warning...

>   A change in semantics? Yes. Because the existing semantics is ... wrong

OK, then this is an important bugfix!

lkishalmi (approved):

PR #2516, opened 2020-11-02
Ensure document state is consistent after unnamed is saved.

sdedic:
In a scenario when the user initially presses `CTRL-N` in VS Code, then saves a file with some contents, a sequence `didOpen`, `didChange` is fired that requires the server to create an empty content file in its internal buffer although the file is already saved on the disk with non-empty content.

Fixed in `TextDocumentServiceImpl` and added a relevant testcase.


dbalek (approved):
Looks fine. Thanks.

JaroslavTulach (approved):

lkishalmi (approved):

PR #2517, opened 2020-11-03
[NETBEANS-4981] Install nbjavac when needed

JaroslavTulach:
The best way to solve [NETBEANS-4981](https://issues.apache.org/jira/browse/NETBEANS-4981) and bring the Apache NetBeans Language Server VSCode experience to users of JDK8 is to mimic the same functionality the NetBeans IDE provides on JDK8 - e.g. to ask the user to download `nbjavac`.

This PR is my attempt to do so.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
@@ -65,6 +65,7 @@ import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectUtils;
 import org.netbeans.api.project.Sources;
 import org.netbeans.api.project.ui.OpenProjects;
+import org.openide.LifecycleManager;
 import org.openide.filesystems.FileObject;
 import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
@@ -213,10 +214,15 @@ public final class Server {
             }
         }
         
-        private void showIndexingCompleted() {
+        private JavaSource showIndexingCompleted(Project[] opened) {
             try {
-                JavaSource.create(ClasspathInfo.create(ClassPath.EMPTY, ClassPath.EMPTY, ClassPath.EMPTY))
-                          .runWhenScanFinished(cc -> {
+                final ClasspathInfo info = ClasspathInfo.create(ClassPath.EMPTY, ClassPath.EMPTY, ClassPath.EMPTY);
+                final JavaSource source = JavaSource.create(info);
+                if (source == null) {
+                    client.showMessage(new ShowStatusMessageParams(MessageType.Error, NO_JAVA_SUPPORT, 0));
+                    return null;
+                }
+                source.runWhenScanFinished(cc -> {
                   if (client.getNbCodeCapabilities().hasStatusBarMessageSupport()) {
                         client.showStatusBarMessage(new ShowStatusMessageParams(MessageType.Info, INDEXING_COMPLETED, 0));
                   } else {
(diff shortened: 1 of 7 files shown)


JaroslavTulach:
To verify this change does what it should. Build the extension:
```
netbeans$ ant build
netbeans$ java/java.lsp.server/
netbeans/java/java.lsp.server$ ant build-vscode-ext
netbeans/java/java.lsp.server$ cd vscode
```

Clean the locally installed nbjavac if any and launch on JDK8
```
netbeans/java/java.lsp.server/vscode$ rm -rf $HOME/.config/Code/User/globalStorage/jlahoda.apache-netbeans-java/userdir/
netbeans/java/java.lsp.server/vscode$ JAVA_HOME=/jdk1.8.0/ code --extensionDevelopmentPath=`pwd`
```
Make sure VSCode doesn't have any other `nbjavac` providing extensions. Make sure VSCode really runs on JDK8 - check `settings.xml` for `java.home`, `netbeans.jdkhome` properties as defined by #2480

sdedic:
Note - you can also start with `--user-data-dir` parameter to point vscode to a fresh directory

sdedic:
apologies - mouse-misclicked Close :( instead of Comment

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -274,54 +282,82 @@ function activateWithJDK(specifiedJDK: string | null, context: ExtensionContext,
                 'nbcodeCapabilities' : {
                     'statusBarMessageSupport' : true
                 }
+            },
+            errorHandler: {
+                error : function(_error: Error, _message: Message, count: number): ErrorAction {
+                    return ErrorAction.Continue;
+                },
+                closed : function(): CloseAction {
+                    activateWithJDK(specifiedJDK, context, log, false);
```
JaroslavTulach, line 291: This should contribute a lot to the robustness of the NetBeans backend integration. Previously it was not guaranteed the launch process really goes thru the `activateWithJDK` booting sequence. By default (if there is no `errorHandler`) the NetBeans client is restarted automatically by trying to connect to dead NetBeans instance.

With this change a restart should replay the whole booting sequence in `activateWithJDK`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -213,35 +211,49 @@ private void asyncOpenSelectedProjects(CompletableFuture f, List<FileObject> pro
...
-            completionOptions.setResolveProvider(true);
-            completionOptions.setTriggerCharacters(Collections.singletonList("."));
-            capabilities.setCompletionProvider(completionOptions);
-            capabilities.setCodeActionProvider(true);
-            capabilities.setDocumentSymbolProvider(true);
-            capabilities.setDefinitionProvider(true);
-            capabilities.setDocumentHighlightProvider(true);
-            capabilities.setReferencesProvider(true);
-            capabilities.setExecuteCommandProvider(new ExecuteCommandOptions(Arrays.asList(JAVA_BUILD_WORKSPACE, GRAALVM_PAUSE_SCRIPT)));
+            if (src != null) {
```
JaroslavTulach, line 244: If there is no `JavaSource` (e.g. no Java infrastructure installed), then our server is started, but it has no code editing capabilities - e.g. no additional queries needed `JavaSource` shall arrive.

platform/autoupdate.cli/src/org/netbeans/modules/autoupdate/cli/ModuleOptions.java
```diff
@@ -482,14 +482,22 @@ public CLIInternalHandle(String displayName, Env env) {
...
+
         @Override
         public void processSelectedProgressEvent(ProgressEvent event) {
-            env.getOutputStream().println(event.getMessage());
+            printEvent(event);
+        }
+
+        private void printEvent(ProgressEvent event) {
+            final String msg = event.getMessage();
+            if (msg != null && msg.length() > 0) {
```
JaroslavTulach, line 498: Without this check the output was full of `null`s:
obrazek

JaroslavTulach (commented):
I am satisfied. This is probably the best we can get for 12.2.

There are some remaining issues:
* installing `nbjavac` requires restart and right now the code calls `System.exit` in middle
* that sometimes breaks the communication channel and yields `EOFException`
* piggy backing on the status message isn't nice either, but it does the trick

Overall it seems to work relatively fine.

sdedic:
As for the code, thumbs up !
However I've tested the branch, and *sometimes* the `nbjavac` install fails with errors. The bug appears +- randomly. I filed [NETBEANS-4991](https://issues.apache.org/jira/browse/NETBEANS-4991) to track the issue.

sdedic (approved):

JaroslavTulach:
Thanks Sváto. It seems that [NETBEANS-4991](https://issues.apache.org/jira/browse/NETBEANS-4991) is independent of the VSCode changes. Let's get them in and then continue investigating the probable problems in `autoupdate.*` modules.

PR #2518, opened 2020-11-03
Enable limited DialogDisplayer emulation over LSP.

sdedic:
The implementation has been already integrated into `origin` (and then merged into `master`), this commit just enables `DialogDisplayer` remoting implementation for the next release cycle.


JaroslavTulach:
I wish I could use this support in #2517, but safer to enable it only in master and observe the possible side-effects.

JaroslavTulach (approved):

sdedic:
re. usage of DialogDisplayer: it would be safer, if the remote support could be selectively enabled, i.e. after mainline code is reviewed, potential `isHeadless` removed. Ideally the Descriptor could be somehow marked to be 'safe' - one of obvious detections is that the `message` of the descriptor is just a String (not JPanel). But Notify/DialogDescriptor does not have such tagging capability at the moment.

PR #2521, opened 2020-11-04
Wait for DAP server init before run/debug.

dbalek:
When trying to run/debug a Java program before the Apache NetBeans DAP Server finishes its initialization a cryptic error message is shown in VSCode UI. Run/debug actions should wait for DAP server init.


JaroslavTulach:
Example of the cryptic message is here:
obrazek

JaroslavTulach (approved):
OK, this seems like a nicer message.

lkishalmi (approved):

PR #2524, opened 2020-11-06
Increase the major version for php.editor

junichi11:
There are incompatible API changes. e.g.
- `ASTPHP5Parser.createDispatch*`
- `ASTPHP5Symbols.T_*`

See: #2504

### DIFF
--- a/php/php.dbgp/nbproject/project.xml
+++ b/php/php.dbgp/nbproject/project.xml
@@ -171,7 +171,8 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.2</specification-version>
+                        <release-version>2</release-version>
+                        <specification-version>2.0</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/php/php.editor/nbproject/project.properties
+++ b/php/php.editor/nbproject/project.properties
@@ -20,7 +20,7 @@ build.compiler=extJavac
 nbjavac.ignore.missing.enclosing=**/CUP$ASTPHP5Parser$actions.class
 javac.compilerargs=-J-Xmx512m
 nbm.needs.restart=true
-spec.version.base=1.89.0
+spec.version.base=2.0
 release.external/predefined_vars-1.0.zip=docs/predefined_vars.zip
 sigtest.gen.fail.on.error=false
 
--- a/php/php.refactoring/nbproject/project.xml
+++ b/php/php.refactoring/nbproject/project.xml
@@ -135,7 +135,8 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.43</specification-version>
+                        <release-version>2</release-version>
+                        <specification-version>2.0</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/php/php.dbgp/manifest.mf
+++ b/php/php.dbgp/manifest.mf
@@ -3,5 +3,4 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.php.dbgp
 OpenIDE-Module-Layer: org/netbeans/modules/php/dbgp/resources/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/php/dbgp/resources/Bundle.properties
-OpenIDE-Module-Specification-Version: 1.50
-
+OpenIDE-Module-Specification-Version: 1.51
--- a/php/php.editor/manifest.mf
+++ b/php/php.editor/manifest.mf
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 AutoUpdate-Show-In-Client: false
-OpenIDE-Module: org.netbeans.modules.php.editor
+OpenIDE-Module: org.netbeans.modules.php.editor/2
 OpenIDE-Module-Layer: org/netbeans/modules/php/editor/resources/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/php/editor/resources/Bundle.properties
 OpenIDE-Module-Implementation-Version: 2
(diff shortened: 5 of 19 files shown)


junichi11:
Do we have to increase the Implementation-Version, too?

https://github.com/apache/netbeans/blob/aaae1fab9f04e87e855836b6795b3c3f8027449e/php/php.editor/manifest.mf#L6

php/php.editor/nbproject/project.properties
```diff
@@ -20,7 +20,7 @@ build.compiler=extJavac
 nbjavac.ignore.missing.enclosing=**/CUP$ASTPHP5Parser$actions.class
 javac.compilerargs=-J-Xmx512m
 nbm.needs.restart=true
-spec.version.base=1.89.0
+spec.version.base=2.0
```
junichi11, line 23: Increased the spec version.

lkishalmi:
Well, not happy about this one, but it's not all about happiness after all. I think the corresponding code is already in beta3. @jtulach , @neilcsmith-net could you take a second look on this one?

junichi11:
Just close this if I'm wrong something or we need not this. Thanks.

tmysik (approved):

tmysik:
> Do we have to increase the Implementation-Version, too?
> 
> https://github.com/apache/netbeans/blob/aaae1fab9f04e87e855836b6795b3c3f8027449e/php/php.editor/manifest.mf#L6

I guess @jtulach will know. Thanks.

php/php.dbgp/nbproject/project.xml
```diff
@@ -171,7 +171,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.2</specification-version>
```
JaroslavTulach, line 174: I would start with:
```
diff --git a/php/php.editor/manifest.mf b/php/php.editor/manifest.mf
index 9e9be53ff210..9509c51eab9a 100644
--- a/php/php.editor/manifest.mf
+++ b/php/php.editor/manifest.mf
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
-OpenIDE-Module: org.netbeans.modules.php.editor
+OpenIDE-Module: org.netbeans.modules.php.editor/2
 OpenIDE-Module-Layer: org/netbeans/modules/php/editor/resources/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/php/editor/resources/Bundle.properties
 OpenIDE-Module-Implementation-Version: 2
diff --git a/php/php.editor/nbproject/project.properties b/php/php.editor/nbproject/project.properties
index a7de1b9d9229..160762e689b5 100644
--- a/php/php.editor/nbproject/project.properties
+++ b/php/php.editor/nbproject/project.properties
@@ -20,7 +20,7 @@ build.compiler=extJavac
 nbjavac.ignore.missing.enclosing=**/CUP$ASTPHP5Parser$actions.class
 javac.compilerargs=-J-Xmx512m
 nbm.needs.restart=true
-spec.version.base=1.89.0
+spec.version.base=2.0
 release.external/predefined_vars-1.0.zip=docs/predefined_vars.zip
 sigtest.gen.fail.on.error=false
```
which would then require this change to be:
```
--- a/php/php.dbgp/nbproject/project.xml
+++ b/php/php.dbgp/nbproject/project.xml
@@ -171,7 +171,8 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.2</specification-version>
+                        <release-version>2</release-version>
+                        <specification-version>2.0</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
```

JaroslavTulach (changes requested):
Increasing major release version results in a change of module name to `OpenIDE-Module: org.netbeans.modules.php.editor/2` and it is the signal for the runtime container to reject any linkages that don't specify `/2`.

JaroslavTulach:
> Do we have to increase the Implementation-Version, too?

Specifying implementation version is only good if other module use implementation dependency. It would be better to get rid of such implementation dependency in any case. Then you can remove the fixed `Implementation-Version` and it will be autogenerated to today's date, unless I am mistaken.

junichi11:
@JaroslavTulach Thank you for your help.

> Increasing major release version results in a change of module name to `OpenIDE-Module: org.netbeans.modules.php.editor/2` and it is the signal for the runtime container to reject any linkages that don't specify `/2`.

Will do that. 

> Specifying implementation version is only good if other (external) module use implementation dependency. It would be better to get rid of such implementation dependency in any case. Then you can remove the fixed `Implementation-Version` and it will be autogenerated to today's date, unless I am mistaken.

Currently, `php.zend`, `php.zend2`, `php.symfony`, and some external plugins use implementation version. 
External plugins may work fine no longer with the new version because they are not API friends of `php.editor`. (Just add them to  `php.editor` as API friends?)

neilcsmith-net:
> Well, not happy about this one, but it's not all about happiness after all.

@lkishalmi yes, that was my concern about having sigtests PR land too late during the release phase. I think we need to think post-release how to do this, and also why certain issues were not picked up earlier (tests?). I don't have much comment on this in particular, but do think @JaroslavTulach comments on #2504 need resolving for release.  That was in my mind from comments on previous releases, but bit surprised how many concerns in this one.

> Then you can remove the fixed Implementation-Version and it will be autogenerated to today's date, unless I am mistaken.

This should be generated from git hash rather than date now, and is why the source releases have that gitinfo file.

php/php.refactoring/nbproject/project.xml
```diff
@@ -135,7 +135,8 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.43</specification-version>
+                        <release-version>2</release-version>
+                        <specification-version>2.0</specification-version>
```
JaroslavTulach, line 139: Yup, I think this is the correct dependency.

JaroslavTulach (approved):

junichi11:
It seems that `php.zend`, `php.zend2`, and `php.symfony` require the implementation version. (cc @tmysik)

tmysik:
@junichi11 

> It seems that `php.zend`, `php.zend2`, and `php.symfony` require the implementation version. (cc @tmysik)

Yes, that is true (they are using some PHP Editor internal classes).

lkishalmi (approved):
Seems to be right now.

PR #2528, opened 2020-11-07
Help Needed: Searchable Open Tab Popup

lkishalmi:
This one is bothering me for a long time. It is not uncommon that I'm working with 100+ files opened and switching between them can be cumbersome. A filter field on the popup switcher would be a great help. I think there is even a JIRA issue on this by @Chris2011.
Well, this PR is far from being complete. Has many issues, but my deep Swing Fu is almost 20 years old and rusty. So it is about this far I could get.

image
The current issues are:
 * The search component is just sitting at the top, no proper borders, layout.
 * The height of the search field is not calculated when determining the size of the table (so the last row of the table could go off screen)
 * Keyboard event handling in the search field and overall component. Well I'd say it does not exist, I merely was able to hack some key events to the search textfield.
 * It would be nice to have some placeholder text and a good border for the search field

So I'd like if somebody with more Swing Experience could take over this one. @DevCharly ?

### DIFF
--- a/platform/o.n.swing.tabcontrol/nbproject/project.properties
+++ b/platform/o.n.swing.tabcontrol/nbproject/project.properties
@@ -17,7 +17,7 @@
 
 is.autoload=true
 javac.compilerargs=-Xlint:unchecked
-javac.source=1.6
+javac.source=1.8
 javadoc.arch=${basedir}/arch.xml
 javadoc.apichanges=${basedir}/apichanges.xml
 
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/ButtonPopupSwitcher.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/ButtonPopupSwitcher.java
@@ -26,10 +26,13 @@ import java.awt.Toolkit;
 import java.awt.event.AWTEventListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseEvent;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import javax.swing.BorderFactory;
 import javax.swing.JComponent;
 import javax.swing.JPopupMenu;
+import javax.swing.JTextField;
 import javax.swing.SwingUtilities;
 import javax.swing.event.*;
 import org.netbeans.swing.popupswitcher.SwitcherTableItem;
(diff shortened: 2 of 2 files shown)


eirikbakke:
I usually use "Go to Symbol" (Ctrl+Shift+O) or "Go to File" (Ctrl+Alt+O on Windows) for this workflow. Do you use those? Does that workflow solve the same problem for you? Still, the search field is a nice addition that can coexist with Go to Symbol/File.

(There are some problems with the ordering of results in Go to Symbol/File... I've been wanting to fix that for a long time, but haven't gotten around to it yet. There's also the problem that the symbol index frequently gets corrupted for the former, but that's a deeper problem.)

eirikbakke:
The code seems nice and simple so far; you have pointed out the remaining issues yourself. Would be a nice addition if someone decides to finish it.

Chris2011:
Wow thx @lkishalmi this is what I was searching and also trying by my own for so long. I checked the code and I couldn't add a shortcut for that. I know this feature in eclipse and I like it. @eirikbakke it is not like go to file Go to file has a lot of more problems. So this feature "Opened Files" needed a search field and a shortcut. Here are the tickets, that I created some years ago: https://issues.apache.org/jira/browse/NETBEANS-1263 (For adding a shortcut to open that field) and https://issues.apache.org/jira/browse/NETBEANS-1264 (Searchfield inside that popup).

We need more and more shortcuts for keyboard users and unfortunately there is no action on this Opened files popup to open it. It is to small to beeing accurate with mouse and with tabs to change focus, is to complex for that. So all in all great start @lkishalmi will check the source too :)

Chris2011 (approved):
Looks good to me :). I dunno whether I'm allowed to give the approval so take it as a comment only. I checked the code and it looks straightforward.

lkishalmi:
Well usually Go to File and Symbol works when I have the Developer hat on. Though my daywork is a DevOps engineer working mostly outside of NetBeans projects, usually that's when this search would be handy.

Chris2011:
Hey @lkishalmi what is missing here to implement?

lkishalmi:
Someone who has a better Swing knowledge. Maybe @DevCharly can look at it.
This implementation is very raw, as far as I remember it can break the event processing a bit in the popup.

Chris2011:
Hey @DevCharly can you please try to help here, that would be awesome :)

eirikbakke:
> This implementation is very raw, as far as I remember it can break the event processing a bit in the popup.

Yeah, there seems to be a serious bug in ButtonPopupSwitcher, unrelated to this PR:
[NETBEANS-5855 Closing unsaved editor tab from dropdown can put IDE in unusable state](https://issues.apache.org/jira/browse/NETBEANS-5855)

Might be good to get this fixed before doing additional changes around this class.

There's also a more cosmetic bug in the same class, filed here:
[NETBEANS-5853 Repaint problems in tabcontrol ButtonPopupSwitcher](https://issues.apache.org/jira/browse/NETBEANS-5853)

PR #2530, opened 2020-11-09
First attempt for Java getter/setter generator for the Java LSP server.

jlahoda:
There are two (three) ways to run the generators:
-if a piece of code is selected, and contains fields, the proposal to add them is sent using code actions (a light bulb should appear)
-there is a new "Generate" sub-menu in the editor context menu in VS Code (note you need a fairly new VS Code to see it)
-via the Command Palette in VS Code


JaroslavTulach:
CC @jisedlac

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -1134,6 +1140,93 @@ public void logMessage(MessageParams arg0) {
         }
     }
 
+    public void testCodeActionGetterSetting() throws Exception {
```
JaroslavTulach, line 1143: Great, there is a test!

JaroslavTulach (approved):

dbalek (approved):
Good job!

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -1129,7 +1171,115 @@ private JavaSource getSource(String fileUri) {
...
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    public static int getOffset(Document doc, Position pos) {
+        return LineDocumentUtils.getLineStartFromIndex((LineDocument) doc, pos.getLine()) + pos.getCharacter();
+    }
+
+    private static String toUri(FileObject file) {
+        if (FileUtil.isArchiveArtifact(file)) {
```
JaroslavTulach, line 1201: I've noticed this code during #2587 and asked why we aren't using `URLMapper.findURL(EXTERNAL)`.

PR #2531, opened 2020-11-10
[NETBEANS-5007] Prevent endless restarts when changing JDKs.

sdedic:
When changing JDKs using GraalVM's extension UI, the vscode kills Apache Language Server (ALS), but then the ALS extensions notices that the channel to ALS broke and restarts it again ... killing the instance. The channel closes again etc etc.

This patch should prevent restarts if the process was killed *because of* a deliberate restart.

### DIFF
--- a/java/java.lsp.server/vscode/src/extension.ts
+++ b/java/java.lsp.server/vscode/src/extension.ts
@@ -251,7 +251,9 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
     let ideRunning = new Promise((resolve, reject) => {
         let collectedText : string | null = '';
         function logAndWaitForEnabled(text: string) {
-            activationPending = false;
+            if (p == nbProcess) {
+                activationPending = false;
+            }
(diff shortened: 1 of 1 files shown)


JaroslavTulach (approved):

JaroslavTulach:
Testing of [12.2-rc1 build of VSCode extension](https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/173/) revealed unwanted number of restarts as @sdedic describes. The fix is essential for proper behavior of the VSCode extension when one updates `netbeans.jdkhome` property.

The fix is in TypeScript part of the extension - e.g. it doesn't influence the behavior of NetBeans IDE at all - can we get the fix into release 12.2 source code without requiring retesting of NetBeans IDE 12.2-rc1, @lkishalmi?

lkishalmi:
The way to get it into 12.2 is the following. Create a blocker issue in JIRA, also post it on the dev mailing list. I need to cut an rc2 probably without affecting the target date.

PR #2532, opened 2020-11-11
[NETBEANS-5010] Go to File search based on folder path too (bz #222697)

emilianbold:


ide/jumpto/src/org/netbeans/modules/jumpto/file/FileSearchAction.java
```diff
@@ -220,6 +220,7 @@ public boolean setListModel(final FileSearchPanel panel, String text ) {
                 itemsComparator.setUsePreferred(panel.isPreferedProject());
                 itemsComparator.setText(text);
                 filterFactory.setLineNumber(lineNr);
+                //TODO: EMI: searchByFolder?
```
Chris2011, line 223: Does this mean anything? If not, please remove, otherwise it will be hard later to figure out again what that means :).

Chris2011 (commented):
IMHO it is good to go, maybe someone else could have a look to. @lkishalmi maybe?

Chris2011:
Didn't test the patch yet, will do it later this week.

Chris2011:
So I tested it. 2 things here:

* the matched folders should also highlighted, like the characters from the file name
* There is no action on the checkbox. Nothing happens when I check/uncheck it. I expect that the search will triggered again. This happens, when you check/uncheck the Case Sensitive checkbox.

Chris2011:
The highlighting is essential for the user to see what is highlighted from the search.

emilianbold:
Perhaps you'll find this more acceptable...

--emi

Chris2011:
Thx, I checked it and it is working great :)

lkishalmi:
All good, let's merge this one in!

PR #2534, opened 2020-11-14
API Signature Snapshot of NetBeans 12.2

lkishalmi:
This one would eventually replace #2504 as the API versions has changed since then.
Thank you @neilcsmith-net  for the guideline. I'm going to add a Step in the release Schedule that the creation and review of the API Snapshot shall happen 1 week before branching.

### DIFF
--- a/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
+++ b/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
@@ -1,5 +1,5 @@
 #Signature file v4.1
-#Version 2.81
+#Version 2.82
 
 CLSS public java.lang.Object
 cons public init()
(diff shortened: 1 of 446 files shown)


neilcsmith-net:
@lkishalmi had intended to look at updating the existing PR this weekend but hadn't had chance since rc yet, sorry.  It should be open to pushing to if you want to keep discussion and review in one place, as well as diff to previous?  But as it needs squashing before merging anyway, pros / cons of that.

Comparing two across separate tabs still looks like some requests for change from @JaroslavTulach remain?  I'll leave him to comment / withdraw those.

Agree wholeheartedly with the need to kick this off before freeze / beta1.  And you, Eric and I could probably spend a bit of time reviewing the current release docs after 12.2 is out?

lkishalmi:
> @lkishalmi had intended to look at updating the existing PR this weekend but hadn't had chance since rc yet, sorry. It should be open to pushing to if you want to keep discussion and review in one place, as well as diff to previous? But as it needs squashing before merging anyway, pros / cons of that.

Well, yes I'm going to cherry-pick this change to your branch, and updating the previous PR. I don't mind squashing. There is value in keeping things together.

PR #2535, opened 2020-11-14
[NETBEANS-5027] PHP - improved display of array parameter

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5027

Improves display of array used as default parameter value.
Uses short array format, displays first item in array, indicates more elements in array.

Before:
netbeans5027-before

After:
netbeans5027-after

### DIFF
--- a/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/CodeUtilsTest.java
+++ b/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/CodeUtilsTest.java
@@ -22,6 +22,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import org.netbeans.junit.NbTestCase;
+import org.netbeans.modules.php.editor.parser.astnodes.ArrayCreation;
+import org.netbeans.modules.php.editor.parser.astnodes.ArrayElement;
+import org.netbeans.modules.php.editor.parser.astnodes.FormalParameter;
+import org.netbeans.modules.php.editor.parser.astnodes.Scalar;
 
 public class CodeUtilsTest extends NbTestCase {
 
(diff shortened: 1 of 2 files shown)


junichi11 (approved):

tmysik:
@KacerCZ 

Could you please check the failed build? Thanks.

php/php.editor/test/unit/src/org/netbeans/modules/php/editor/CodeUtilsTest.java
```diff
@@ -179,4 +183,43 @@ public void testCommonNamespacePrefixes11() {
...
+        FormalParameter param = createFormalParameterWithDefaultArray(array);
+        assertEquals("[3 => 'a']", CodeUtils.getParamDefaultValue(param));
+    }
+
+    private FormalParameter createFormalParameterWithDefaultArray(List<ArrayElement> arrayContent) {
+        ArrayCreation defaultValue = new ArrayCreation(1, 1, arrayContent, ArrayCreation.Type.NEW);
+        FormalParameter param = new FormalParameter(1, 1, null, null, defaultValue);
+        return param;
+    }
+
```
tmysik, line 224: Could you please add a test where the array value is an object? I would like to see how the tooltip will look like in such a case. Thanks!

tmysik (changes requested):
Please, see my comment. Thanks.

php/php.editor/test/unit/src/org/netbeans/modules/php/editor/CodeUtilsTest.java
KacerCZ: If I understand you correctly you mean this case?
<code>
function printDate(\DateTime $date = new \DateTime()) {}
</code>

PHP does not support object as default value.

KacerCZ:
@tmysik Build failure is related to network issue during build setup. Could you please trigger build again?

junichi11:
@hectorespert Could you have a look at the following?
```
Error: Unable to process command '::add-path::/home/runner/.composer/vendor/bin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

php/php.editor/test/unit/src/org/netbeans/modules/php/editor/CodeUtilsTest.java
tmysik: No. I meant his the tooltip would look like for array e.g. `[0 => new DateTime()]`. Just to verify that it still looks ok. Thanks and sorry for not being clear.

tmysik: Ha, sorry, I was too fast, now I see it. It is only about the parameters, not all arrays "everywhere".

tmysik (approved):

KacerCZ:
@junichi11 This issue was fixed by @hectorespert in PR #2543

junichi11:
I know that. It occurred again when I reran jobs. (See timestamp of build job.)

tmysik:
So, I would merge it. Any objections?

junichi11:
@tmysik Just in case, let's wait for a reply from @hectorespert .
I reran the jobs after that change was merged into the master branch. That should pass if there is no problem, I suppose.

hectorespert:
For me it's ok, linux tests are failing due the fix was not applied in this PR, but the job to run php tests in windows pass this change.

PR #2537, opened 2020-11-15
[NETBEANS-5029] PHP - generated methods are public by default

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5029

Changes default value for generating public modifier of constructor/getters/setters to enabled.

constructor-generator


junichi11 (approved):

junichi11:
@tmysik Could you please merge this if it's OK with you?

tmysik:
Looks good to me. Perhaps, as a possible enhancement, we could replace the given checkbox with a combo box (protected/public). And one more thing - it would be nice if the value could be remembered (in user properties) and restored whenever this dialog appears.

Thanks!

KacerCZ:

Regarding protected/public selection - I needed protected getter only in rare occasions and it was always easier to change generated code than change setting in dialog and then to remember change it back.

If I understand this code correctly, then setting is saved to project:
https://github.com/apache/netbeans/blob/f986c03ac36c77c077165a4e0e3d7e4b680bc562/php/php.editor/src/org/netbeans/modules/php/editor/codegen/CGSGenerator.java#L414-L445

tmysik:
@KacerCZ 

I see. Project settings seem fine to me. And those were just possible enhancements, as I wrote.

PR #2538, opened 2020-11-16
Turning on lexer.nbbridge enables Go to declaration in VSCode

JaroslavTulach:
Using F12 on String & co. didn't open the String source. With `org.netbeans.modules.lexer.nbbridge` included in `nbcode` suite, it works again.


jlahoda (approved):
Looks good.

PR #2539, opened 2020-11-16
[NETBEANS-5031] - remove use of deprecated APIs in MiniEdit.java

BradWalker:
I'm trying to remove the use of a deprecated API that hinders updating the Java version.

MiniEdit.java uses a really old deprecated API. This API prevents us from upgrade the Java
source version in the project properties file.

So step #1 is to remove the use of the old deprecated API.

### DIFF
--- a/java/performance/actionsframework/src/org/netbeans/actions/examples/MiniEdit.java
+++ b/java/performance/actionsframework/src/org/netbeans/actions/examples/MiniEdit.java
@@ -25,7 +25,6 @@
 package org.netbeans.actions.examples;
 
 import java.awt.Component;
-import java.awt.Dimension;
 import java.awt.FileDialog;
 import java.awt.FlowLayout;
 import java.awt.KeyboardFocusManager;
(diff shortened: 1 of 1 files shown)


BradWalker:
I would like to ask if @JaroslavTulach could have a look at this as it's his code that I'm changing.. That would be helpful if possible..

JaroslavTulach (approved):
Consider also removing dependency on `org.openide.util.enum` from the `project.xml` file.

BradWalker:
> Consider also removing dependency on `org.openide.util.enum` from the `project.xml` file.

Hey @JaroslavTulach , I can't seem to find this.. Can you point to where I should look so that I'm more familiar in the futures. Thanks..

jtulach:
I cannot find it.

Btw. is this file even compiled? If I make an error in the file, rebuilding the `java/performance` module still succeeds. Doing `java/performance$ grep -r actionsframework .` yields no results...

BradWalker:
> I cannot find it.
> 
> Btw. is this file even compiled? If I make an error in the file, rebuilding the `java/performance` module still succeeds. Doing `java/performance$ grep -r actionsframework .` yields no results...

Hey @JaroslavTulach, I don't think this file is even compiled. So I would be willing to even remove it from the tree? My change is simply to try and keep things "neat & complete" as I clean up the old rust..

What do you think I should do?

PR #2541, opened 2020-11-17
Gradle projects should recognize Kotlin source directories.

jpeseknb:


java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
```diff
@@ -42,12 +42,13 @@
...
         "LBL_GROOVY=Groovy",
         "LBL_SCALA=Scala",
+        "LBL_KOTLIN=Kotlin",
         "LBL_RESOURCES=Resources",
         "LBL_GENERATED=Generated"
     })
     public static enum SourceType {
 
-        JAVA, GROOVY, SCALA, RESOURCES,
+        JAVA, GROOVY, SCALA, KOTLIN, RESOURCES,
```
lkishalmi, line 51: Always add enum at the end of the list when extending an API, aslo add javadoce with @since annotation. see GENERATED

lkishalmi (changes requested):
Looks sane. Beside the small ones, the only change is missing is the api-changes.xml

extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
```diff
@@ -199,12 +199,15 @@ class NbProjectInfoBuilder {
         boolean hasJava = project.plugins.hasPlugin('java-base')
         boolean hasGroovy = project.plugins.hasPlugin('groovy-base')
         boolean hasScala = project.plugins.hasPlugin('scala-base')
+        boolean hasKotlin = project.plugins.hasPlugin('org.jetbrains.kotlin.android') ||
```
lkishalmi, line 202: I'd add these plugins to the detectPlugins (line 140) as well, so it other NetBeans modules can enrich the GradleProject Lookup when they are present.

lkishalmi:
Thanks for your contribution!
Also a JIRA enhancement issue would be really nice!

jpeseknb:
@lkishalmi I've filled https://issues.apache.org/jira/browse/NETBEANS-5050 and updated the patch in cfd8de3.

lkishalmi (approved):
Looks good now! Thank you!

lkishalmi:
I'd squash and merge this change however you need to have a public email on your GitHub account.

lkishalmi:
@matthiasblaesing I think this contribution is a natural extension of the code we already have, however I'd like you have a look. Do we require a signed ICLA in this case?

matthiasblaesing:
@lkishalmi I agree with your assessment and would consider this an evolution. Without diminishing the work done here one might argue, that from observing the surrounding source, you could reach the same conclusion/implementation. So @jpeseknb if you intend to contribute more work (certainly appreciated!) I suggest to get an ICLA on file and there is no doubt, if you are unsure, please acknowledge here or in the issue (they are linked and thus a record will be retained with the ASF) that this code proposed here is yours and you understand, that you donate the code to the ASF (as stated in the license header of the files you edited).

neilcsmith-net:
@lkishalmi moving to 12.4 milestone, but happy for you to retarget back if you think we should and we can get this in by beta2.  We can squash and merge to the PR branch without requiring changes by the author if need be.  And disagree on need for ICLA in this case.

geertjanw:
There are conflicts to be resolved here. Changing milestone to 12.5.

ebarboni:
still no answer changing to milestone 12.6

lkishalmi:
Rebased on latest master, seems to work as the attached screenshot shows. Ready to merge (if CI agrees as well).

PR #2542, opened 2020-11-17
[NETBEANS-5034] fix wrong javadoc error hint

jherkel:
https://issues.apache.org/jira/browse/NETBEANS-5034

### DIFF
--- a/java/javadoc/src/org/netbeans/modules/javadoc/hints/Analyzer.java
+++ b/java/javadoc/src/org/netbeans/modules/javadoc/hints/Analyzer.java
@@ -633,7 +633,7 @@ final class Analyzer extends DocTreePathScanner<Void, List<ErrorDescription>> {
         if (t == null) {
             errors.add(ErrorDescriptionFactory.forSpan(ctx, start, end, TAG_UNKNOWN(treeName)));
         } else {
-            if (!t.hasOptionalEndTag() && !isVoid(t)) {
+            if (!isVoid(t)) {
                 tagStack.push(node);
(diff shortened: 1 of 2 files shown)


lkishalmi:
It would be good to reason, why the original problem happened and how that condition removal fixes it.

jherkel:
I think probably this PR https://github.com/apache/netbeans/pull/2470 caused this behaviour. After some studying of code I think that correct place where we can omit some html tags with optional end tag is in visitEndElement method, not visitStartElement. And there is a code :
```
while (!tagStack.isEmpty()) {
....
else if (tag != null && tag.hasOptionalEndTag()) {
tagStack.pop();
}
...
```
which processes these start tags and remove them from stack.

jlahoda:
I think this fix (removing `!t.hasOptionalEndTag()` in start element) is great.I'd suggest to add a test, like:

```
    public void testOptionalEndTag() throws Exception {
        HintTest.create()
                .input(
                "package test;\n" +
                "/**\n" +
                " * <ul><li>One<li>Two</li></ul>\n" +
                " */\n" +
                "class Zima {\n" +
                "}\n")
                .preference(AVAILABILITY_KEY + true, true)
                .preference(SCOPE_KEY, "private")
                .run(JavadocHint.class)
                .assertWarnings();
    }
```

in `java/javadoc/test/unit/src/org/netbeans/modules/javadoc/hints/Analyzer2Test.java`.

Sorry for the trouble!

jlahoda (approved):

PR #2544, opened 2020-11-17
[NETBEANS-1309] add support for @summary javadoc tag

jherkel:
see https://issues.apache.org/jira/browse/NETBEANS-1309

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -186,7 +186,7 @@ class NbProjectInfoBuilder {
 
         public ValueAndType(Class type, Object value) {
             this.type = type;
-            this.value = Optional.of(value);
+            this.value = Optional.ofNullable(value);
         }
 
         public ValueAndType(Class type) {
@@ -290,7 +290,7 @@ class NbProjectInfoBuilder {
         Map<String, Object> taskProperties = new HashMap<>();
         Map<String, String> taskPropertyTypes = new HashMap<>();
         
-        Map<String, Task> taskList = project.getTasks().getAsMap();
+        Map<String, Task> taskList = new HashMap<>(project.getTasks().getAsMap());
         for (String s : taskList.keySet()) {
             Task task = taskList.get(s);
             Class taskClass = task.getClass();
@@ -307,7 +307,7 @@ class NbProjectInfoBuilder {
     private void detectTaskDependencies(NbProjectInfoModel model) {
         Map<String, Object> tasks = new HashMap<>();
         
-        Map<String, Task> taskList = project.getTasks().getAsMap();
+        Map<String, Task> taskList = new HashMap<>(project.getTasks().getAsMap());
         for (String s : taskList.keySet()) {
             Task task = taskList.get(s);
             Map<String, String> taskInfo = new HashMap<>();
@@ -684,7 +684,7 @@ class NbProjectInfoBuilder {
                     }
 
                     NamedDomainObjectContainer nc = (NamedDomainObjectContainer)value;
-                    Map<String, ?> m = nc.getAsMap();
+                    Map<String, ?> m = new HashMap<>(nc.getAsMap());
                     List<String> ss = new ArrayList<>(m.keySet());
                     propertyTypes.put(prefix + propName + COLLECTION_KEYS_MARKER, String.join(";;", ss));
                     for (String k : m.keySet()) {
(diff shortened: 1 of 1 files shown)


java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
```diff
@@ -1280,6 +1282,22 @@ private StringBuilder inlineTags(List<? extends DocTree> tags, TreePath docPath,
                 case TEXT:
                     TextTree ttag = (TextTree)tag;
                     sb.append(ttag.getBody());
+                    break;
+                default : {
+                    // process tags that we cannot add directly because they are not accessible during compilation
+                    DocTree.Kind values[] = tag.getKind().getClass().getEnumConstants();
```
jlahoda, line 1288: For consideration here, how about, instead of finding the summaryTag, there would be just `if ("SUMMARY".equals(tag.getKind().name())) ..` or `switch (tag.getKind().name()) { case "SUMMARY":`?

java/java.editor/src/org/netbeans/modules/java/editor/javadoc/TagRegistery.java
```diff
@@ -48,18 +50,17 @@ public static TagRegistery getDefault() {
...
-        List<TagEntry> selection = new ArrayList<TagEntry>();
-        for (TagEntry te : tags) {
-            if (te.isInline == inline && te.whereUsed.contains(kind)) {
-                selection.add(te);
-            }
-        }
+    public List<TagEntry> getTags(ElementKind kind, boolean inline,SourceVersion sourceVersion) {
+        List<TagEntry> selection = tags.stream().filter(te -> (te.isInline == inline 
+            && te.whereUsed.contains(kind)
+            && sourceVersion.ordinal() >= te.minJavaVersion
```
jlahoda, line 56: The use of integer versions and ordinals is not particularly nice, but I suspect this unfortunatelly cannot be avoided here.

jlahoda (approved):
I think this looks good overall. One comment on detecting the tags. Would be good to have tests - I don't think there are tests for ElementJavadoc, which is sad, but at least for the code completion (`java/java.editor/test/unit/src/org/netbeans/modules/java/editor/javadoc/JavadocCompletionQueryTest.java`).

java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
jherkel: I will change it in accordance with your proposal. But I don't know how to add unit test that is executed only with JDK 11 for example. Is there any example how to do it?

java/java.editor/src/org/netbeans/modules/java/editor/javadoc/TagRegistery.java
jherkel: I couldn't use SourceVersion but I think (and hope) that SourceVersion enum is constructed in a way that ordinal method equals java version. Also I found this bug (https://bugs.openjdk.java.net/browse/JDK-8253545) with comment from Joe Darcy confirming my assumption.

java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
jherkel: I added unit test, but need to change some additional files because I have to set correct source level.

geertjanw:
Any reason not to merge?

geertjanw:
Changing milestone to 12.5.

ebarboni:
this PR is a bit old but maybe @jlahoda may say it's ok to merge for 12.5.

ebarboni:
postponing until decision

java/java.editor/src/org/netbeans/modules/java/editor/javadoc/JavadocCompletionItem.java
```diff
@@ -156,10 +157,10 @@ public String toString() {
                 txt, substitutionOffset);
     }
 
-    public static List<CompletionItem> addBlockTagItems(ElementKind kind,
+    public static List<CompletionItem> addBlockTagItems(SourceVersion sourceVersion,ElementKind kind,
             String prefix, int startOffset) {
         
-        List<TagEntry> tags = TagRegistery.getDefault().getTags(kind, false);
+        List<TagEntry> tags = TagRegistery.getDefault().getTags(kind, false,sourceVersion);
```
mbien, line 163: nit: please use a space behind the "," to be consistent with the rest of the file

mbien (commented):
@jherkel since this is an older PR, it would be good to rebase this to latest master so that we get a fresh CI run using the new workflow and travis setup.

java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementJavadoc.java
```diff
@@ -1280,6 +1282,24 @@ private StringBuilder inlineTags(List<? extends DocTree> tags, TreePath docPath,
...
+                    break;
+                default : {
+                    // process tags that we cannot add directly because they are not accessible during compilation
+                    switch(tag.getKind().name()) {
+                        case "SUMMARY" : 
+                            try {
+                                Method getSummaryMethod = tag.getClass().getDeclaredMethod("getSummary");
+                                List<? extends DocTree> summaryList = (List<? extends DocTree>)getSummaryMethod.invoke(tag);
+                                sb.append(inlineTags(summaryList, docPath, doc, trees, null));
+                            } catch(NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException ex) {
```
mbien, line 1294: could be: `(ReflectiveOperationException | ... rest)`

`ReflectiveOperationException` is a common supertype which makes those lines shorter.

mbien (approved):
@jherkel I added remarks where the code could be potentially improved. I think it would be good to remove the reflective code before integration if possible.

Ran some manual tests and it worked as advertised.

Please squash commits before merge. Code freeze is April 20th (tomorrow) for the next release.

lkishalmi:
@jherkel I've just done a rebase on current master!

matthiasblaesing:
Removing milestone as this is stale and needs a rebase.

matthiasblaesing:
@jherkel something went wrong here - it seems you pushed an empty branch.

jherkel:
@matthiasblaesing You are right, I wanted to rebase my branch but I did it via UI and not cmd line, my mistake. I would like to update my commit (address points from review) , so I will prepare the new commit. But probably I will prepare this as a new pull request.

PR #2546, opened 2020-11-19
End user oriented README.md and dedicated BUILD.md

JaroslavTulach:
Improving `README.md` to look better in VS Code user interface:

obrazek

Plus backporting fix for #2538

PR #2548, opened 2020-11-19
[NETBEANS-5043] Turning on lexer.nbbridge enables Go to declaration in VSCode

lkishalmi:
Well piggybacking on #2546 is a nasty thing to do.


JaroslavTulach:
The intention wasn't to piggyback, but rather have a single "VSCode fixes for RC3" issue. This way - e.g. #2546 and #2548 - it is cleaner. Thank you.

PR #2550, opened 2020-11-21
[NETBEANS-5048] PHP - cancel Find Usages action

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5048

Allows to cancel Find Usages action in PHP editor.

### DIFF
--- a/php/php.refactoring/src/org/netbeans/modules/refactoring/php/findusages/PhpWhereUsedQueryPlugin.java
+++ b/php/php.refactoring/src/org/netbeans/modules/refactoring/php/findusages/PhpWhereUsedQueryPlugin.java
@@ -40,6 +40,7 @@ public class PhpWhereUsedQueryPlugin extends ProgressProviderAdapter implements 
 
     protected AbstractRefactoring refactoring;
     private WhereUsedSupport usages;
+    private volatile boolean cancelled;
 
     public PhpWhereUsedQueryPlugin(AbstractRefactoring refactoring) {
         this.refactoring = refactoring;
@@ -62,6 +63,11 @@ public class PhpWhereUsedQueryPlugin extends ProgressProviderAdapter implements 
             for (FileObject fileObject : relevantFiles) {
                 if (fileObject == null) {
                     continue;
+                }
+                if (cancelled) {
+                    // Reset cancelled state for repeated search to work.
+                    cancelled = false;
+                    break;
                 }
                 usages.collectUsages(fileObject);
                 fireProgressListenerStep();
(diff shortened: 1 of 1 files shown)


php/php.refactoring/src/org/netbeans/modules/refactoring/php/findusages/PhpWhereUsedQueryPlugin.java
```diff
@@ -63,6 +64,10 @@ public Problem prepare(final RefactoringElementsBag elementsBag) {
                 if (fileObject == null) {
                     continue;
                 }
+                if (cancelled) {
+                    cancelled = false;
```
junichi11, line 68: Maybe, need not this. `cancelled` is used only here at the moment, but I think we should keep the cancelled state.

KacerCZ: If state is left, then it is impossible to repeat action by clicking green arrows in Usages window.

junichi11: Ah, I see. You are right. Thanks.

php/php.refactoring/src/org/netbeans/modules/refactoring/php/findusages/PhpWhereUsedQueryPlugin.java
```diff
@@ -63,6 +64,10 @@ public Problem prepare(final RefactoringElementsBag elementsBag) {
                 if (fileObject == null) {
                     continue;
                 }
+                if (cancelled) {
```
junichi11, line 67: @tmysik Instead, should use `AtomicBoolean.compareAndSet()`?

junichi11: We should leave a comment about why we change it to `false`.

KacerCZ: Added comment why `cancelled` is reset.

For `cancelled` I used same type as in OccurencesFinderImpl in PHP Editor:
https://github.com/apache/netbeans/blob/514e170f43416aec1a25c1c443837bf1ef9043f0/php/php.editor/src/org/netbeans/modules/php/editor/csl/OccurrencesFinderImpl.java#L62

tmysik: Volatile is OK here, IMHO.

tmysik (approved):

junichi11 (approved):

KacerCZ:

tmysik:
@KacerCZ 

You are welcome. Thanks for your work!

PR #2551, opened 2020-11-21
[NETBEANS-5049] Prevent NPEs in SemanicHighlighter.

jlahoda:


lkishalmi (approved):
Looks sane. BTW shall this one be backported to 12.0 later?

PR #2552, opened 2020-11-21
Add dialogs for inserting/editing code template parameters

isarthur:


Chris2011:
This is cool and I have a different and natural approach for this. It would be better to have code completion inside of that area. Code completion while typing or while pressing ctrl + space. This is more natural and you can type and no need to switch with mouse or tab key.

isarthur:
Thank you for your comment @Chris2011. Basically, I can register a MIME type for code templates and associate it with this text pane. Upon completion of the code templates, options from predefined parameter sets will be suggested, but code completion in this area that currently exists, for example for JS, will not be suggested. If no one is against it, I can probably do it.

Chris2011:
Sounds good to me :)

mbien:
@arsadykov if you want to resurrect this PR, please fix the conflicts and rebase to latest master.

PR #2557, opened 2020-11-23
[NETBEANS-5057] fix for indexing JSF composite components

fcorneli:
We have several JSF libraries that contain both JSF components and JSF composite components.
When the JSF library is open in NetBeans, all JSF components are found within the editor.
When the JSF library is referenced as external JAR, only the JSF components are recognised by the editor, not the JSF composite components.

The fix as suggested in:
https://stackoverflow.com/questions/21607320/jsf-composite-components-in-jar-not-recognised-in-netbeans
works for us.

https://issues.apache.org/jira/browse/NETBEANS-5057

### DIFF
--- a/enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/CompositeComponentModel.java
+++ b/enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/CompositeComponentModel.java
@@ -270,7 +270,7 @@ public class CompositeComponentModel extends JsfPageModel {
             if (folder.getName().equalsIgnoreCase("resources")) { //NOI18N
                 //check if its parent is META-INF
                 FileObject parent = folder.getParent();
-                if (parent != null && parent.getNameExt().startsWith("META-INF")) { //NOI18N
+                if (parent != null && parent.getName().startsWith("META-INF")) { //NOI18N
                     //the folder seems to be the right resources folder
                     return folder;
                 }
--- a/enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/JsfBinaryIndexer.java
+++ b/enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/JsfBinaryIndexer.java
@@ -48,14 +48,14 @@ import org.openide.util.Exceptions;
  * @author marekfukala
  */
 @ConstrainedBinaryIndexer.Registration(
-namePattern = ".*\\.tld|.*\\.taglib\\.xml",
+namePattern = ".*\\.tld|.*\\.taglib\\.xml|.*\\.xhtml",
 indexVersion = JsfBinaryIndexer.INDEXER_VERSION,
 indexerName = JsfBinaryIndexer.INDEXER_NAME)
 public class JsfBinaryIndexer extends ConstrainedBinaryIndexer {
 
     private static final Logger LOGGER = Logger.getLogger(JsfBinaryIndexer.class.getSimpleName());
     private static final String CONTENT_UNKNOWN = "content/unknown";    //NOI18N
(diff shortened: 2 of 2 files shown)


matthiasblaesing:
@fcorneli thank you for your work. This issue troubles me @work also. It is nice to see this getting fixed. I rebuild a nightly and try to test this in the next few days. In the meantime, I have to things I noticed right now:

- please ensure an issue exists, that covers your changes
- make sure the title of the PR and the commit summary start with the issue number in the format "[NETBEANS-????] <Summary>"
- either you should have an ICLA on file or I have to ask you whether you understand and agree, that you donate the change to the ASF. At this point in time this means, that the code is covered by the ALv2, but the ASF is free to relicense in the future und any license it chooses. With an ICLA this is part of the agreement, if you plan to contribute more (this would be great!), an ICLA is preferred, if not, please comment with the acknowledgement into the corresponding issues.

If you have questions, feel free to ask here or (preferred on the dev@netbeans.apache.org  mailinglist (http://netbeans.apache.org/community/mailing-lists.html)

fcorneli:
@matthiasblaesing Made the suggested changes. Also sent over the signed ICLA.

enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/CompositeComponentModel.java
```diff
@@ -270,7 +270,7 @@ private static FileObject getResourcesDirectory(FileObject file) {
             if (folder.getName().equalsIgnoreCase("resources")) { //NOI18N
                 //check if its parent is META-INF
                 FileObject parent = folder.getParent();
-                if (parent != null && parent.getNameExt().startsWith("META-INF")) { //NOI18N
+                if (parent != null && parent.getName().startsWith("META-INF")) { //NOI18N
```
matthiasblaesing, line 273: This should not be necessary. FileObject#getName would be better called "getNameWithoutExtension" and FileObject#getNameExt returns the normal filename. Could you please double check?

matthiasblaesing (commented):
My quick test today was positive - I already found several places where invalid attributes where used, that were not declared. This was previously masked by the general error.

I left an inline comment, please have a look at that.

Thank you for getting an ICLA on file. Please indicate once you received confirmation, that the ICLA was received.

enterprise/web.jsf.editor/src/org/netbeans/modules/web/jsf/editor/index/CompositeComponentModel.java
fcorneli: The inclusion of the "extension" is irrelevant here I guess, as you're only looking for `META-INF` at the beginning. That's probably why the original patch, as suggested on stackoverflow, changed from `getNameExt()` to `getName()`. Additional benefit is that you also get rid of a string concat by not using `getNameExt()`.

As for `FileObject` itself: do you really want to change such a "core" class? This would trigger a lot of changes all over the place, no?

matthiasblaesing: No I don't want to change FileObject, that sentence was meant as an explanation. I think FileObject is a broken abstraction, but indeed it so far into the base of netbeans, that it won't be changed. I prefer changes to be minimal. Not needing this change, dropping it is IMHO a good idea.

fcorneli: `FileObject` indeed looks a bit, how to put it, hairy.

fcorneli:
> Please indicate once you received confirmation, that the ICLA was received.

@matthiasblaesing NETBEANS project should have received the signed ICLA.

matthiasblaesing:
Thank you - looks sane to me.

PR #2563, opened 2020-11-26
[NETBEANS-5004] Replace script with native macOS executable

oyarzun:
When running NetBeans on mac with JDK 11.0.9 and/or JDK 15.0.1, Netbeans and any of its subprocesses are not allowed to access any of the personal files and folders (e.g.: Downloads, Documents...). 

Replacing the `CFBundleExecutable` in the app bundle from a script to a native executable fixes the issue.

https://developer.apple.com/forums/thread/132986

This PR removes the old unused mac IDE launcher, with a new one written in Swift with similar logic to the netbeans shell script. 

The native executable is built when creating the mac installer package.

When using the native executable the user is now ask for permission to access the files and folder and show up in the mac's Files and Folders Privacy settings.

Screen Shot 2020-11-26 at 2 36 13 PM

@mcdonnell-john when you get a chance, can you review this PR?


oyarzun:
@matthiasblaesing Thanks, fixing it now.

mcdonnell-john:
Hi @oyarzun 

I had a look at this today, had to work out all the steps that Jenkins does, when we're generating an Installer, so took me a bit of time and have been fairly busy recently so couldn't get around to this as soon as I would have liked.

I generated a new installer for a DEV version of Netbeans from this PR, I'll need to make a small change in the apache/netbeans-jenkins-lib so that we can copy over the `nb/ide.launcher/macosx` folder into the distpreparation(Generated as part of the Jenkins release builds) folder to make the installer process seemless - I'll raise a PR for this in the morning.  I assume these scripts are in this location as its also used when building a macosx installer for NetBeans Platform apps from within NetBeans, which I assume you've tested? (Forgive my ignorance, I've only really generated Apache NetBeans installer)

I noticed that when I build a new installer, and installed it, called Apache NetBeans XX.X I got the same thing as you got in your screenshot of the Security & Privacy app:
Screenshot 2020-12-30 at 01 24 52

But so long as the other PR makes it in, I think this could be tried out for Apache NetBeans 12.3

oyarzun:
@mcdonnell-john thanks for taking a look at the PR. 

The application name in the Security & Privacy panel, is coming from `CFBundleName` in the Info.plist. It gets populated with [nbbuild/installer/newbuild/netBeants/nbide/build.xml](https://github.com/apache/netbeans/blob/master/nbbuild/installer/mac/newbuild/netBeans/nbide/build.xml) and [nbbuild/installer/mac/newbuild/build.pro](https://github.com/apache/netbeans/blob/0415bfa6e3abd4ae8881d966c279e2156b2763d5/nbbuild/installer/mac/newbuild/build.pro#L32) from the [NB_VER_NUMBER](https://github.com/apache/netbeans-jenkins-lib/blob/ff762b4a31c244a1ae05db7d2ace8a93c5e611e3/resources/org/apache/netbeans/installermac.sh#L130). So I imagine you had `NB_VER_NUMBER=X.XX` when you built the installer? 

I was not familiar with apache/netbeans-jenkins-lib, I thought the mac installers were built manually. The other option could be to have the native Mac launcher already prebuilt, like the Windows launcher. But I was not sure how submit a PR that has a new external binary.

This launcher is only for the NetBeans app bundle [ide.launcher](https://github.com/apache/netbeans/tree/master/nb/ide.launcher), not RCP app bundles [api.harness](https://github.com/apache/netbeans/tree/master/harness/apisupport.harness). 
I was planning on submitting another PR to address the RCP launcher. I believe it will need to be a prebuilt external binary like Window. This would allow the mac RCP installer to be build on any platform since it does not use Apple's `pkgbuild`.

neilcsmith-net:
What's the status of this?  If it's really going in to 12.3 will need merging soon, and we need to consider how best to test it?  Built like the Windows launcher may be better there, or we need to find a way to provide beta installers too.

oyarzun:
@neilcsmith-net 

> we need to consider how best to test it?

macOS users could test the launcher using their NetBeans 12.2 installation by replacing the symlink (to the shell script) with the native launcher.

1. Get the launcher source from this PR in `nb/ide.launcher/macosx`.
2. Install Xcode if not already installed.
3. Compile the swift launcher.
    ```
    cd nb/ide.launcher/macosx
    swift build -c release
    ```
4. Replace launcher.
    ```
    sudo rm /Applications/NetBeans/Apache\ NetBeans\ 12.2.app/Contents/MacOS/netbeans
    sudo cp .build/x86_64-apple-macosx/release/NetBeansLauncher /Applications/NetBeans/Apache\ NetBeans\ 12.2.app/Contents/MacOS/netbeans
    ``` 


> Built like the Windows launcher may be better there

For the NetBeans launcher, I do not see this as a better approach right now. Currently, there are bug fixes to native code that have been merged, but have never been rebuilt and uploaded to https://netbeans.osuosl.org/binaries/.

For the RCP launcher, this is a must to allow RCP developers to be able to package macOS app bundles on Windows or Linux.

Thankfully, @lbruun has begun to work on a solution to leverage GitHub Actions to build the native code in PR #2700.
But there is still the issue of uploading the binaries to https://netbeans.osuosl.org/binaries/. A manual procedure that needs to be performed by a PMC member.


**NOTE apache/netbeans-jenkins-lib#23 also needs to be merged for the native NB launcher to be built when generating the macOS installer.**

neilcsmith-net:
@oyarzun I think you're slightly missing what I meant by like the Windows launcher.  Ideally if we have a new launcher for macOS it would be alongside the existing ones in `bin/` for zip builds, which means being usable from the Jenkins build.  At the moment we only link zip builds of betas for testing, and some people still prefer using that anyway.

Obviously, that doesn't seem feasible for 12.3, so may need an interim plan for this that just involves the installer?

oyarzun:
@neilcsmith-net the macOS native launcher is specific to a macOS Application bundle [1] it will not work out of the bin directory. A user could still execute the shell script from the bin dir to start NetBeans from the command line. Running from the command line inherits the permission of the Terminal app so NETBEANS-5004 does not manifest itself. But the average Mac NetBeans user will download the installer and run NetBeans from Finder, Dock or Spotlight Search.

The swift code uses the Bundle class [2] 

```swift
let brandingToken = Bundle.main.object(forInfoDictionaryKey: "CFBundleExecutable") as? String
let appName = Bundle.main.object(forInfoDictionaryKey: "CFBundleName") as? String
let confFile = Bundle.main.path(forResource: brandingToken, ofType: "conf", inDirectory: brandingToken! + "/etc")
let clustersFile = Bundle.main.path(forResource: brandingToken, ofType: "clusters", inDirectory: brandingToken! + "/etc")
let iconFile = Bundle.main.path(forResource: brandingToken, ofType: "icns")
let nbexecURL = Bundle.main.url(forResource: "nbexec", withExtension: "", subdirectory: brandingToken! + "/platform/lib")
```

1. [https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/AboutBundles/AboutBundles.html](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/AboutBundles/AboutBundles.html)

2. [https://developer.apple.com/documentation/foundation/bundle](https://developer.apple.com/documentation/foundation/bundle)

neilcsmith-net:
@oyarzun OK, makes sense, sorry.  Understand bundles, just hadn't looked enough at where this was going!  Also understand the bug, but confused at overlap with shell launcher behaviour.  Is replicating the job of the shell script, rather than delegating to it, necessary to fix the permission issue?

My key concern is whether we're trying to get this in 12.3, and having a way to get it tested, through.

matthiasblaesing:
Review was requested from me - sorry I don't feel qualified for reviewing mac OS software.

mcdonnell-john:
Hi Guys,

I initially tested this locally rebuilding a new Mac OSX Installer for Apache NetBeans and it worked. I didn't merge it at the time as I wanted to see if @rtaneja1 had any comments.

Though there appears to be new a commits since that point in time.  If this [commit](https://github.com/apache/netbeans/pull/2563/commits/5a7738fc122257f54303b963bdd93c602cb5fb96) @oyarzun was reverted and moved into its own PR I'd be comfortable mergng this PR without that additional commit as I've tested it to that point.

I'd say this is needed to merged otherwise we'll get more issues because of BigSur, but not comfortable merging it with that additional commit at this late stage.

oyarzun:
@neilcsmith-net 

> Is replicating the job of the shell script, rather than delegating to it, necessary to fix the permission issue?

I'm not sure it is necessary. The current launcher shell script just call another script `nbexec`,  so I chose to cut out the middleman and take advantage of macOS APIs to get default user and cache dirs.

mcdonnell-john:
@oyarzun  That last commit doesn't interfere with the previous work does it?  i.e. Ultimately I think we need this merged for the Apache NetBeans installer, and that commit won't affect what I've already tested.

So I wouldnt stand in the way of this being merged as is.

mcdonnell-john (approved):

oyarzun:
@mcdonnell-john 

> Though there appears to be new a commits since that point in time. If this commit @oyarzun was reverted and moved into its own PR I'd be comfortable merging this PR without that additional commit as I've tested it to that point.

> I'd say this is needed to merged otherwise we'll get more issues because of BigSur, but not comfortable merging it with that additional commit at this late stage.

NETBEANS-5004 is not specific to Big Sur, it is also an issue on Catalina.

The code in that commit is not compiled or used until we solve the binary upload to https://netbeans.osuosl.org/binaries/. At which point it will be included in the harness. Currently with the PR as is, there is no chance that commit will break anything. 

But if you want me to revert it, I will and submit a separate PR.

mcdonnell-john:
@oyarzun no need to revert...

neilcsmith-net:
> I'm not sure it is necessary. The current launcher shell script just call another script nbexec, so I chose to cut out the middleman and take advantage of macOS APIs to get default user and cache dirs.

Pros and cons. Being able to update the launcher script without needing to recompile the native executable has benefits, it's less to have to keep in sync and test (given both must now work on macOS), and in terms of the platform people might adapt the launcher script.

PR #2564, opened 2020-11-26
Added extension mechanism to LSP module

rcano:
Custom plugins integrating with the LSP module require different
extension methods to leverage custom mechanism in LSP implementations
(which is common practice).
This PR enables one such extension by exposing the WorkspaceService api
in an abstracted manner.


sdedic:
Note: there has been massive development in the LSP area. Will you update the PR to the latest codebase ? A general extension mechanism could be useful; but the code herein may be (partially)  obsolete already.

rcano:
I'll have to take a look, I haven't kept up with the changes. Maybe this PR is no longer needed.

PR #2565, opened 2020-11-27
[NETBEANS-5073] Add space after data type

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5073

Fixes formatting of PHP code by adding space between data type and parameter or property.

Source

```php
class Foo {
    public string$name;
    public function isValid(float$length): bool {}
}
```

is formatted as

```php
class Foo {
    public string $name;
    public function isValid(float $length) {}
}
```


junichi11 (approved):

tmysik (approved):

tmysik:
@junichi11 Can we meet it now, right?

junichi11:
@tmysik Yes. Merging.

PR #2566, opened 2020-11-28
[NETBEANS-5074] - cleanup ArrayList warnings..

BradWalker:
This work cleans up a bunch of warnings surrounding the use of ArrayList..

For example:

   [repeat] /home/bwalker/src/netbeans/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabLayoutManager.java:245: warning: [unchecked] unchecked conversion
   [repeat]             List<Integer>[] rowIndexes = new ArrayList[rowCount];
   [repeat]                                          ^
   [repeat]   required: List<Integer>[]
   [repeat]   found:    ArrayList[]

ActionsManager.java was especially involved since one change would require another. And keeping track of
that, so as not to break things, was hard.

### DIFF
--- a/java/java.sourceui/src/org/netbeans/modules/java/source/ui/OpenProjectFastIndex.java
+++ b/java/java.sourceui/src/org/netbeans/modules/java/source/ui/OpenProjectFastIndex.java
@@ -254,7 +254,7 @@ class OpenProjectFastIndex implements ClassIndexManagerListener {
     }
 
     private Collection<FileObject>  getFileRoots(Iterable<? extends URL> roots) {
-        Collection c = new ArrayList<FileObject>(5);
+        Collection<FileObject> c = new ArrayList<>(5);
         for (Iterator<? extends URL> it = roots.iterator(); it.hasNext(); ) {
             URL rootURL = it.next();
             FileObject fo = URLMapper.findFileObject(rootURL);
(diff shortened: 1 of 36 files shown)


java/java.sourceui/src/org/netbeans/modules/java/source/ui/OpenProjectFastIndex.java
```diff
@@ -550,7 +550,7 @@ private PI getCacheInfo(Project p) {
         private final int         size;
         private final Project     project;
         
-        NameIndex(Project p, FileObject root, String files, ArrayList indices) {
+        NameIndex(Project p, FileObject root, String files, List indices) {
```
mklaehn, line 553:
```suggestion
        NameIndex(Project p, FileObject root, String files, List<Object[]> indices) {
```

mklaehn (approved):
Nice work.
Commented on a place where the type of the list could be added.

java/java.sourceui/src/org/netbeans/modules/java/source/ui/OpenProjectFastIndex.java
BradWalker: Hey @mklaehn , thanks for the sharp eyes!

It's been changed and integrated..

mklaehn (approved):
thank you for the work

PR #2568, opened 2020-11-29
[NETBEANS-5062] PHPDoc static type incorrectly resolved when returned by another class

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-5062

```php

<?php
class Test1
{

    private function getTest2() {
        return Test2::returnStatic();
    }

    public function testMethod() {
        echo "Test1" . PHP_EOL;
    }

    public function test() {
        $static1 = $this->getTest2(); // Before: $static1 is recognized as Test1, After $static1 is recognized as Test2
        var_dump($static1); // object(Test2)#2 (0) {}
        $static1->testMethod();
    }
}

class Test2
{

    public static function returnStatic() {
        return new static();
    }

    public function testMethod() {
        echo "Test2" . PHP_EOL;
    }

}

$test = new Test1();
$test->test();
```


tmysik (approved):
Well done! Can we meet it now, right?

junichi11:
@tmysik Thank you. Yes.

PR #2571, opened 2020-11-30
VSCode: Auto import of types on code completion.

dbalek:
- Auto import of types on code completion
- Completion items sorting
- Override/implement method code completion


jlahoda (approved):

jlahoda:
Looks fine to me. I would prefer some tests, but can live without them.

JaroslavTulach (approved):

dbalek:
> Looks fine to me. I would prefer some tests, but can live without them.

Test added.

PR #2572, opened 2020-11-30
Display sources loaded from String and find their MIME type.

entlicher:
Sources that have no File associated are put to a special virtual SourceFS filesystem. We add the MIME type information and use these files in NetBeans. It's also assured that breakpoints submitted into these files are resolved correctly.
A simple API change needs to be done: `DVFrame.getSourceMimeType()`.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
@@ -18,7 +18,11 @@
  */
 package org.netbeans.modules.java.lsp.server.debugging;
 
+import java.io.File;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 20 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
```diff
@@ -286,11 +290,33 @@ public NbProtocolServer(DebugAdapterContext context) {
...
                         Source source = new Source();
-                        source.setName(Paths.get(sourceURI).getFileName().toString());
-                        source.setPath(sourceURI.getPath());
-                        source.setSourceReference(0);
+                        String scheme = sourceURI.getScheme();
+                        if (null == scheme || scheme.isEmpty() || "file".equalsIgnoreCase(scheme)) {
+                            source.setName(Paths.get(sourceURI).getFileName().toString());
+                            source.setPath(sourceURI.getPath());
+                            source.setSourceReference(0);
+                        } else {
```
JaroslavTulach, line 300: A test for this path would be nice. Maybe extend `ServerTest`?

JaroslavTulach (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
entlicher: Right. I've created https://issues.apache.org/jira/browse/NETBEANS-5084

PR #2573, opened 2020-12-01
[NETBEANS-5075] Remove consecutive empty lines when formatting PHP code

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5075

Removes multiple empty lines from PHP code.
Adds setting under Formatting/PHP/Blank Lines.

Formatting options:
netbeans-5075

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
@@ -430,3 +430,6 @@ FmtWrapping.groupUseListLabel.text=G&roup Use List:
 FmtWrapping.groupUseBracesCheckBox.text_1=Group &Use Braces
 FmtBraces.groupUseLabel.text=G&roup Use:
 FmtAlignment.noteLabel.text=<html><i>Note: These options depend on Wrapping settings</i></html>
+FmtBlankLines.maxPreservedBlankLabel.AccessibleContext.accessibleDescription=Maximal preserved consecutive empty lines
+FmtBlankLines.maxPreservedBlankLabel.AccessibleContext.accessibleName=Maximal Preserved Lines
+FmtBlankLines.maxPreservedBlankLabel.text=Max Preserved Lines:
--- a/php/php.editor/src/org/netbeans/modules/php/editor/indent/CodeStyle.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/indent/CodeStyle.java
@@ -236,6 +236,10 @@ public final class CodeStyle {
         return preferences.getInt(BLANK_LINES_BEFORE_CLOSE_PHP_TAG, getDefaultAsInt(BLANK_LINES_BEFORE_CLOSE_PHP_TAG));
     }
 
+    public int getBlankLinesMaxPreserved() {
+        return preferences.getInt(BLANK_LINES_MAX_PRESERVED, getDefaultAsInt(BLANK_LINES_MAX_PRESERVED));
+    }
+
     // Spaces ------------------------------------------------------------------
 
     public boolean spaceBeforeWhile() {
--- a/php/php.editor/src/org/netbeans/modules/php/editor/indent/FmtOptions.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/indent/FmtOptions.java
@@ -108,6 +108,7 @@ public final class FmtOptions {
     public static final String BLANK_LINES_AFTER_OPEN_PHP_TAG = "blankLinesAfterOpenPHPTag"; //NOI18N
     public static final String BLANK_LINES_AFTER_OPEN_PHP_TAG_IN_HTML = "blankLinesAfterOpenPHPTagInHTML"; //NOI18N
     public static final String BLANK_LINES_BEFORE_CLOSE_PHP_TAG = "blankLinesBeforeClosePHPTag"; //NOI18N
+    public static final String BLANK_LINES_MAX_PRESERVED = "blankLinesMaxPreserved"; //NOI18N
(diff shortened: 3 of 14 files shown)


php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
```diff
@@ -430,3 +430,8 @@ FmtWrapping.groupUseListLabel.text=G&roup Use List:
 FmtWrapping.groupUseBracesCheckBox.text_1=Group &Use Braces
 FmtBraces.groupUseLabel.text=G&roup Use:
 FmtAlignment.noteLabel.text=<html><i>Note: These options depend on Wrapping settings</i></html>
+FmtBlankLines.jTextField1.text=jTextField1
```
junichi11, line 433: Remove?

junichi11:
I would add the option to enable this feature. Some users may add new lines intentionally.

KacerCZ:
This option removes blank lines only if there are more blank lines than configured value.
In default setting it keeps only blank single line and removes all extra lines.
This formatting is used in Symfony framework - https://cs.symfony.com/ rule "no_extra_blank_lines".

If user wants to keep more lines, then it can be set to higher value.
Only setting this value to 0 will remove blank lines from code completely.

Other lines around properties, methods, classes are not affected by this change.

junichi11:
Ah, I see now. OK.Thanks.

php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
KacerCZ: Yes, removed.

junichi11 (approved):

php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
```diff
@@ -430,3 +430,7 @@ FmtWrapping.groupUseListLabel.text=G&roup Use List:
 FmtWrapping.groupUseBracesCheckBox.text_1=Group &Use Braces
 FmtBraces.groupUseLabel.text=G&roup Use:
 FmtAlignment.noteLabel.text=<html><i>Note: These options depend on Wrapping settings</i></html>
+FmtBlankLines.maxPreservedField.text=
```
tmysik, line 433: Nitpick: The given text field can be set to be "NOI18N" in its properties (this line would not be here than).

php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
```diff
@@ -430,3 +430,7 @@ FmtWrapping.groupUseListLabel.text=G&roup Use List:
...
+FmtBlankLines.maxPreservedLabel.text=Max Preserved Lines:
```
tmysik, line 434: Please, add "blank" everywhere here as it is not apparent what exactly it does. The constant as well as the Symfony rule contains it. Thanks.

CC @junichi11

tmysik (approved):

php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
KacerCZ: Fixed.

php/php.editor/src/org/netbeans/modules/php/editor/indent/ui/Bundle.properties
KacerCZ: Changed name to `maxPreservedBlankFiled` and `maxPreservedBlankLabel`.

tmysik: Well, I was talking about the text of the label, not the label name. But thanks for changing it too and sorry for not being clear. Please, improve the label so it is obvious what lines are going to be preserved. Thanks!

junichi11: @tmysik Maybe, I suppose @KacerCZ did not add it because Category is Blank Lines.

nb2573-formatting-option

KacerCZ: Changing text to "Max Preserved Blank Lines:" widens panel.
I hope meaning of this setting will be clear from category as @junichi11 says.

tmysik: I see now... OK then. Thanks!

tmysik:
@KacerCZ Could you please check the CI failures whether they are unrelated? Thank you.

junichi11:
@tmysik Failed job is "Build the Visual Studio Code extension for Java". So, maybe I think it is not related this change. (I reran the job but it failed.)

@hectorespert Do you know anything about this? We can ignore it, right?

hectorespert:
> @hectorespert Do you know anything about this? We can ignore it(merge this PR), right?

Yes, that job is failing in other branches too

KacerCZ:

tmysik:
@KacerCZ Thanks for your time and work! And for your patience with my comments :)

junichi11:
@hectorespert Thank you!

terax6669:
Those are my current settings. I believe I've had issues with properties touching the traits or constants. I can't remember - it was quite a while ago.
netbeans64_WUCImgWJsp

Anyway, this way I am able to keep the inside of the class tidy, but the unfortunate side effect is that the beginning and end of class have additional blank lines. Will this pr fix this? (If it's already possible please instruct me how to fix my settings)

junichi11:
@terax6669 If you have a problem, please report it to the JIRA: https://issues.apache.org/jira/projects/NETBEANS/issues
Thanks.

PR #2574, opened 2020-12-02
Inform VSCode that application/javascript is also a MIME type of JavaScript.

entlicher:
This change is necessary to have files with `application/javascript` MIME type recognized as JavaScript files. VSCode recognizes `text/javascript` only by default.
Sources that come from evaluations may have just `<eval>` name, without an extension. In that case just the associated MIME type defines the language of the source.


dbalek (approved):
Fine.

PR #2575, opened 2020-12-03
Open GraalVM sources out of the box

JaroslavTulach:
The [GraalVM](http://graalvm.org) sources use a fine tuned build system called [mx](http://github.com/graalvm/mx) (shorthand of [MaXine](https://en.wikipedia.org/wiki/Maxine_Virtual_Machine)). The `mx` command seems to work fine for the special purposes of building own enhanced JVM and it looks like it is going to stay. The goal of this PR is to establish a solid infrastructure to make sure the _Apache NetBeans IDE_ as well as products based on top of it (let's name [IGV](https://www.graalvm.org/tools/igv/) and [VSNetBeans](https://cwiki.apache.org/confluence/display/NETBEANS/Apache+NetBeans+Extension+for+Visual+Studio+Code)) understand the `suite.py` metadata files and can use the `mx` command to build and run/debug unit tests.

The `mx` build system integrates with various IDEs via `mx ideinit` command. That command generates essential metadata for Eclipse, IntelliJ as well as NetBeans. However the NetBeans solution is based on *Ant* scripts. It has some limitations and moreover the Apache NetBeans project decided to focus on Maven (and Gradle) rather than promoting the old fashioned Ant based projects. As such, I'd like to deprecate the `mx netbeansinit` command in `mx` and rather propose all NetBeans users to switch to native *mx projects* support proposed by this PR.

The `mx` build system isn't in wide spread use (however it *is* used by independent groups like at the [Postdam university](https://github.com/hpi-swa/trufflesqueak) for its unique features allowing to easily extend the JVM). Providing support for such exotic build system may seem unnecessary, but there are reasons to do it. We already have support for _OpenJDK projects_, which is also kind of special as well. As far as I know there were no problems with it. It fully builds on the modular architecture of NetBeans and comes as a single module. Moreover it gives NetBeans a fantatic story for OpenJDK developers. I'd like to repeat the same with the *mx projects* support - a single, isolated module causing no harm, handy (out of the box) for those who use `mx` as their build system. Acceptable?

The PR isn't ready to be merged right now, I still need to find out what to do with the tests. Still, I wanted to bring this idea to your attention. If we agree this code is _"integratable"_ I'll polish the code and tests to comply with the Apache requirements.

PS: My hope is that the approach of this PR is found acceptable for Apache NetBeans project. If that happens the code of this module is going to be more easily approachable to other people building NetBeans and together we can improve experience of NetBeans users even when working with other projects (I dream about improved _compile on save capability_, better API for _external execution_ & co.).


neilcsmith-net:
Thanks for the request to look at this.  I was thinking about alternative build systems and NetBeans earlier this year but not had time to look further (after [Bach](https://github.com/sormuras/bach) talk at our room at FOSDEM).  I do wonder whether trying to add internal support for other specific build systems is the right approach.  Maybe a standard declarative project form (with optional JS, JShell, etc.) scripting and/or support for [Build Server Protocol](https://build-server-protocol.github.io/) might be a better way forward?

JaroslavTulach:
>  [Build Server Protocol](https://build-server-protocol.github.io/) might be a better way forward?

Interesting approach. Thanks for sharing the link. After skimming through the docs three questions puzzle in my mind:

1. While promised in the teaser pages, there is no specification for **debugging** - that's tough one to abstract
2. I was hoping to have the *mx projects* support in 12.3 - yet the BSP looks very **experimental** (like the missing debugging support)
3. Should NetBeans acts as a **server or client**? The NetBeans infrastructure is actually ideal for abstracting over various build systems and responding to the BSP queries!

Ideally we get this PR in, someone sits and implements BSP server based on NetBeans project infrastructure and then we'll have BSP for `mx`! Did I turned it all upside down?

lkishalmi (approved):
Interesting piece of work. Seems sane.

lkishalmi:
```
https://github.com/apache/netbeans/tree/master/java/java.mx.project/src/org/netbeans/modules/java/mx/project/mx-suite.py
https://github.com/apache/netbeans/tree/master/java/java.mx.project/src/org/netbeans/modules/java/mx/project/Bundle.properties
https://github.com/apache/netbeans/tree/master/java/java.mx.project/release/org.eclipse.jdt.core.prefs
https://github.com/apache/netbeans/tree/master/java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/suitepy/compiler-suite.py
```

JaroslavTulach:
> Unapproved license in 4 file(s) 
> java/mx/project/Bundle.properties
> release/org.eclipse.jdt.core.prefs

Both files fixed by adding the license header.

> modules/java/mx/project/suitepy/compiler-suite.py

Only a test class, but removed, the parsing is now tested as part of the 977d1eb integration tests

> mx-suite.py

Removed in f1fadbb

matthiasblaesing:
I only skimmed the code. I still have a WTF feeling when I consider, that once again a build system was invented.

If I'm not mistaken, it has the same problems, that gradle exhibits, i.e. the configuration format is an executable file (at least I understand, that the suite.py is the configuration and the file ending indicates full blown python file). This raises for me the question: Will the approach to statically parse that file scale to the future?

In `CoreSuite.java` a big set of libraries seems to be generated. Are these dependencies set in stone? It looks like a strange approach to dependency management.

The one question in my mind: I read mx as a special system invented in the context of graal, does it pull enough weight to be implemented in the core of netbeans or could it live a s a plugin in the plugin portal?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/CoreSuite.java
```diff
@@ -0,0 +1,732 @@
...
+        return new MxLibrary.Arch() {
+            @Override
+            public MxLibrary amd64() {
+                return amd64;
+            }
+        };
+    }
+
+    static final MxSuite CORE_5_279_0;
+    static {
```
jlahoda, line 221: I wonder - can we parse this at runtime from something? Like, maybe, mx/suite.py? It feels weird to have this information hardcoded.

java/java.mx.project/release/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,442 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
jlahoda, line 1: What is the specific use of this file? Is it better to have it hardcoded, or read the version from mx?

jlahoda (approved):
I guess I concur with Matthias' feeling about another build system invented. But, it is not use who is inventing it, we just add support for it.

I'll leave detailed answers on reading the suite.py on Jaroslav, but my understanding is that this is "mostly declarative" - i.e. there appears to be a (semi-)JSON embedded in the Python file. We can parse the (semi-)JSON without running the code, hopefully.

Regarding the BSP - I wonder how (if) it solved the problem this project type is solving. Yes, part of this project type is a way to actually do a build action, and start debugger, etc. But, there is also a part that determines the source roots, their relationship and classpaths. That is needed by the editor, and I am not sure if BSP provides some configuration.

I may be biased, but I think having the project type here in the Apache NetBeans project would allow joint development, which is likely needed.

I only skimmed through the patch (didn't check line-by-line), but overall looks reasonable. I tried to open an mx project, and it did something that seemed reasonable. Added comments on two places that may need some tweaking. I think I am not worried too much of this causing technical problems: unless someone opens an mx project, the project type should have minimal, I'd hope.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/CoreSuite.java
JaroslavTulach: That would require `mx.mx/suite.py` being around. While `mx` is needed for execution/debugging of tests, it is not needed for opening the projects and navigating the sources right now. However, to properly resolve "standard mx libraries" we need to have information about them. That is what this file provides.

Having the information here improves out of box experience. One can `git clone https://github.com/oracle/graal` and immediately open and inspect the sources.

Of course, when new version of `mx` introduces new "standard libraries", we might need to release an update to this `CoreSuite` data. Possibly we could use this as default/fallback. Once `mx` is found/specified we could switch to its `mx.mx/suite.py` definitions.

JaroslavTulach:
Thanks for the review.

> I only skimmed the code. I still have a WTF feeling when I consider, that once again a build system was invented.

I share your feelings. Looks to me every bigger project ends up inventing own build system (NetBeans included). I spent my first year in OracleLabs trying to remove `mx` (at least it now integrates with Maven central), but when it got to _either me or mx_ decision point I gave up and let `mx` be. Having `mx` creates some weird situations (like the need for this PR), but overall one can get used to it.

> If I'm not mistaken, it has the same problems, that gradle exhibits, i.e. the configuration format is an executable file (at least I understand, that the suite.py is the configuration and the file ending indicates full blown python file). 

Right the `suite.py` configuration is written in Python syntax. However there is 1:1 mapping to JSON and that's what this module is using to parse the configuration files in a declarative way. E.g. we can avoid the Gradle situation. 

> This raises for me the question: Will the approach to statically parse that file scale to the future?

* There are other systems in OracleLabs infrastructure that use the conversion to JSON and subsequent parsing. 
* There seems to be an overall consensus that this is desirable. 
* There are _gate checks_ that run the code used in this PR - should that ever get broken integration into any Graal repository gets rejected
* Once this PR is accepted into NetBeans I modify the gates to use the NetBeans to do the actual parsing/verification - btw. it would be great if there was `--verifyprojectconsistency dir` option to run NetBeans in headless mode and verify project structure is still without any errors from a NetBeans IDE point of view, CCing @jlahoda! Recently also [requested on a mailing list](https://lists.apache.org/thread.html/r799decd155c9197bc2dbe6fc1185e090ce1dc7fafcdb91e55fc0b0e1%40%3Cdev.netbeans.apache.org%3E).

Overall the conversion to JSON and parsing approach shall be safe forever.

> In `CoreSuite.java` a big set of libraries seems to be generated. Are these dependencies set in stone? It looks like a strange approach to dependency management.

These libraries are mostly infrastructure JARs used by the actual `mx` Python code to perform unit testing, benchmarking, signature testing, code coverage tasks. These libraries may grow over time (just [recently a newer version of sigtest](https://github.com/graalvm/mx/commit/68fd66578b5b13f79fbf72a2c1e96164f5f43dc7) was added). But overall they are kept stable and immutable.

> The one question in my mind: I read mx as a special system invented in the context of graal, does it pull enough weight to be implemented in the core of netbeans or could it live a s a plugin in the plugin portal?

I'd like to get this PR in for 12.3 in its simple form, but long term I have to agree - it would be better to provide this module as a standard Apache NetBeans extension on plugin portal. We need:
* build system changes to create a cluster(?) that is not part of final ZIP, but its NBMs are part of the release
* make sure ergonomics module scans these NBMs for various registrations
* polish the UX of downloading NBMs via ergonomics - for example when `mx.*/suite.py` folder is found
* ultimately we can modify installer to install just `platform`, `ide`, `ergonomics` clusters by default and let the rest be downloaded on a fly!

I always wanted [ergonomics to allow downloads](http://wiki.netbeans.org/FitnessForever), decrease the download size, promote features to install and I am thrilled to see this being considered again. However, assuming this PR doesn't negatively affects other modules when people don't use `mx`, I'd prefer to integrate it for 12.3 in its simple form and let the great visions materialize later. By integrating this PR as is, the module is immediately available in VSNetBeans - it would be great if this support could get into VSNetBeans as soon as possible.

java/java.mx.project/release/org.eclipse.jdt.core.prefs
JaroslavTulach: This file was added by @dbalek to support [Eclipse Formatter by Benno](http://plugins.netbeans.org/plugin/50877/eclipse-code-formatter-for-java-eclipse-mars-4-5).

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
```diff
@@ -0,0 +1,245 @@
...
+                                    }
+                                }));
+                            }
+                        }
+                        return null;
+                    };
+                });
+        ProcessBuilder processBuilder = ProcessBuilder.getLocal();
+        processBuilder.setWorkingDirectory(suiteDir.getPath());
+        processBuilder.setExecutable("mx"); // NOI18N
```
sdedic, line 206: This seems to assume `mx` is on the IDE process'  `PATH`. Shouldn't it be configurable in Tools | Options, similar to maven/ant, especially when NB does not distribute `mx` ? If reasonable, I recommend to track it as JIRA enhancement.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
```diff
@@ -0,0 +1,100 @@
...
+    private final SuiteProject project;
+    private final ClassPath bootCP;
+
+    public SuiteClassPathProvider(SuiteProject project) {
+        this.project = project;
+        List<ClassPath.Entry> entries = JavaPlatformManager.getDefault().getDefaultPlatform().getBootstrapLibraries().entries();
+        List<URL> roots = new ArrayList<>();
+        for (ClassPath.Entry entry : entries) {
+            URL root = entry.getURL();
+            if (root.getPath().contains("/graal-sdk.jar")) {
```
sdedic, line 45: Maybe this exclusion is worth a comment in source; why the exclusion is needed.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public ClassPath findClassPath(FileObject file, String type) {
+        SuiteSources.Group g = project.getSources().findGroup(file);
+        if (g == null) {
+            return null;
+        }
+        if (ClassPath.BOOT.equals(type)) {
+            return bootCP;
```
sdedic, line 66: I am not sure about this fixed bootclasspath, as there may be different `JAVA_HOME` defined in `~/.mx/env` or `mx.{projectname}/env`

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+                    Group g = new Group(name, mxPrj, srcDir, srcGenDir, binDir, prgName, displayName);
+                    arr.add(g);
+                    if (firstGroup == null) {
+                        firstGroup = g;
+                    }
+                    prevName = displayName;
+                }
+            }
+            if (firstGroup != null) {
+                fillDeps.put(name, firstGroup);
```
sdedic, line 160: Silly question, but why just the 1st group ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+            return sources;
+        }
+        if (imp.subdir()) {
+            for (FileObject subDir : dir.getParent().getChildren()) {
+                sources = findSuiteSources(subDir, imp.name());
+                if (sources != null) {
+                    return sources;
+                }
+            }
+            for (FileObject subDir : dir.getParent().getParent().getChildren()) {
```
sdedic, line 264: Is the 2-level parent limit hardocded in `mx` as well ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+        Set<Project> prjs = new HashSet<>();
+        for (SuiteSources imp : imported.values()) {
+            prjs.add(imp.prj);
+        }
+        return prjs;
+    }
+
+    @Override
+    public URL[] findUnitTests(FileObject fo) {
+        return new URL[0];
```
sdedic, line 530: no unit tests reported ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+        public String getName() {
+            return this.name;
+        }
+
+        @Override
+        public Set<ClassPath.Flag> getFlags() {
+            return exists ? Collections.emptySet() : Collections.singleton(ClassPath.Flag.INCOMPLETE);
+        }
+
+        private FileObject getJar(boolean ignore) {
```
sdedic, line 597: Maybe an useless parameter

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+        }
+
+        @Override
+        public Collection<Dep> allDeps() {
+            return allDeps;
+        }
+
+        private void computeClassPath(Map<String, Dep> transDeps) {
+            for (Dep d : transDeps.values()) {
+                d.owner().computeTransitiveDeps();
```
sdedic, line 805: `processTransDeps` computes transitive deps on `d.owner()`, isn't it computed 2nd time here for all deps not just those used in classpath construction ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+            this.lib = getOSSLibrary(lib);
+        }
+
+        final MxLibrary getOSSLibrary(MxLibrary lib) {
+            if (lib.sha1() == null && !lib.os_arch().isEmpty()) {
+                Map<String, MxLibrary.Arch> os_dep_libs = lib.os_arch();
+                String os = System.getProperty("os.name").toLowerCase();
+                for (Map.Entry<String, MxLibrary.Arch> entry : os_dep_libs.entrySet()) {
+                    if (os.contains(entry.getKey())) {
+                        return entry.getValue().amd64();
```
sdedic, line 935: Uh :) I meant `mx` already supports `aarch64`, maybe even `i386` ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+        }
+
+        @Override
+        public void setAllDeps(Collection<Dep> set) {
+            this.allDeps = set;
+        }
+
+        @Override
+        public Set<ClassPath.Flag> getFlags() {
+            return exists ? Collections.emptySet() : Collections.singleton(ClassPath.Flag.INCOMPLETE);
```
sdedic, line 1003: better check exists != null as well

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
```diff
@@ -0,0 +1,1195 @@
...
+        }
+
+        @Override
+        public List<? extends PathResourceImplementation> getResources() {
```
sdedic, line 1117: Can be somehow shared with `Library` implementation ?

matthiasblaesing:
Just for the record: I expressed my concerns, but this change has minimal cross cutting changes and does not bring to much code in. So if mx support gets extracted/removed at some point, it can be done with minmal fuzz and thus the potential negative impact is minimal.

lkishalmi:
It seems Travis is complaining on some jaxb stuff for Java 8 tests, I guess that's unrelated with this PR. @JaroslavTulach please feel free to Squash and merge this one.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
JaroslavTulach: True, the UI should be improved in the future. Recorded as [NETBEANS-5158](https://issues.apache.org/jira/browse/NETBEANS-5158).

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
JaroslavTulach: Unless I am mistaken this is because GraalVM (based on JDK8) contains these JARs on boot class path and we want to avoid them.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
JaroslavTulach: Let's put this on TODO list: [NETBEANS-5159](https://issues.apache.org/jira/browse/NETBEANS-5159). If you face a buggy behavior, please describe reproducer there and turn that issue into bug.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: The existing sources that I am aware of are find with 2-level limit. I am not sure if the 2-level parent limit is in the {{mx}} itself or not. Scanning more deeply might have negative effect on performance. Let's leave it as it is unless 
@gilles-duboscq wants to comment.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: Done in 3933d300360

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: This is one of the reasons why special `mx` support is desirable. Typical NetBeans projects assume two kinds of directories - main and test. This is not true for `mx` there is no association of main code and test code. Tests are regular source roots with classes annotated by `@Test`. As such this code returns `new URL[0]`.

Feel free to improve it, if you know how.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: This silently assumes all developer laptops and workstations are AMD64. True for now, until Apple own processor becomes popular among devs. I guess it is going to take a bit of time and it will need deeper fixes in `mx` itself then.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: Running the tests I haven't seen a case where there was more groups than `firstGroup`. But I don't want to replace it with `assert` now.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: Good catch: 9e9567e4729

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
gilles-duboscq: There is indeed a 2-level limit in mx: starting from the directory in which the various repositories are checked out, the "importer" specifies if the imported suite is found at depth 1 or 2, no other depth can be specified (default is depth 1 and a `"subdir" : True` attribute switches the depth to 2).

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: The transitive deps are processed just once. Let's make it clearer by better naming 5b589ee5930f

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteSources.java
JaroslavTulach: Excellent idea: af12108

JaroslavTulach:
Thanks a lot for your reviews. As soon as the gates seem fine, I merge.

java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteCheck.java
```diff
@@ -0,0 +1,257 @@
...
+            if (fo.isFolder()) {
+                continue;
+            }
+            Project prj = FileOwnerQuery.getOwner(fo);
+            assertSame("FileOwnerQuery returns the right project", project, prj);
+            if (!fo.hasExt("java")) {
+                nonJavaCount++;
+                continue;
+            }
+            JavaSource source = JavaSource.forFileObject(fo);
```
JaroslavTulach, line 183: Alas, this test seems to [randomly fail](https://travis-ci.org/github/apache/netbeans/jobs/749982297).

java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteCheck.java
```diff
@@ -0,0 +1,257 @@
...
+                SourceForBinaryQuery.Result2 res2 = SourceForBinaryQuery.findSourceRoots2(root);
+                assertTrue("Has to prefer sources", res2.preferSources());
+                for (FileObject src : res2.getRoots()) {
+                    if (src.equals(dir)) {
+                        break OK;
+                    }
+                }
+                fail("Expecting to find " + dir + " among:\n" + Arrays.toString(res2.getRoots()));
+            }
+            source.runUserActionTask((CompilationController p) -> {
```
JaroslavTulach, line 200: Is there a better way to verify project sources are fine from a NetBeans Java Source support perspective?

java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteCheck.java
JaroslavTulach: Hopefully it fails only if the `.mx/cache` doesn't contain necessary JAR libraries (locally I have them). Let's see whether this gets fixed after 4cbe7aa

PR #2579, opened 2020-12-04
Avoid to call add/addAll on a potentially unmodified collection

digiovinazzo:
The JDK (AdoptOpenJDK in this case) returns an unmodifiable collection, so we should not call add/addAll on it, but create a new ArrayList instead.

### DIFF
--- a/platform/core.network/src/org/netbeans/core/network/proxy/NbProxySelector.java
+++ b/platform/core.network/src/org/netbeans/core/network/proxy/NbProxySelector.java
@@ -67,12 +67,12 @@ public final class NbProxySelector extends ProxySelector {
         int proxyType = ProxySettings.getProxyType ();
         switch (proxyType) {
             case ProxySettings.DIRECT_CONNECTION:
-                res = Collections.singletonList (Proxy.NO_PROXY);
+                res.add(Proxy.NO_PROXY);
                 break;
             case ProxySettings.AUTO_DETECT_PROXY:
                 if (useSystemProxies ()) {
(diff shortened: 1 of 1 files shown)


platform/core.network/src/org/netbeans/core/network/proxy/NbProxySelector.java
```diff
@@ -145,7 +145,7 @@ public NbProxySelector() {
             case ProxySettings.AUTO_DETECT_PAC:
                 if (useSystemProxies ()) {
                     if (original != null) {
-                        res = original.select(uri);                   
+                        res = new ArrayList<>(original.select(uri));
```
matthiasblaesing, line 148: Minimal nitpick: The other branches of the conditional add their result to the list, that is initialized in line 66. Would it be an option to use the same approach here? If not, this is ok, just indicate so.

digiovinazzo: Yes, it makes sense: now all uses of `res` variable just add items to it.

matthiasblaesing:
Thank you @digiovinazzo. I'll merge now.

Normally we require a Jira issue to track contributions from non-committers, but in this case the change is small enough to be merged without the extra fuzz. Just to be clear: The statement at the head of the changed file will also be applied to the contributed change, i.e. if the ASF ever needs to relicense the file, it will still be considered to be owned by the ASF.

matteodg:
OK, thanks sounds good.

PR #2582, opened 2020-12-05
[NETBEANS-5057] Improved support for JSF composite components

fcorneli:
The JSF tag attribute help now shows the default value, if available.


matthiasblaesing:
Looks sane to me. I would consider moving all the key-value information together into a table, but that is just a suggestion.

fcorneli:
@matthiasblaesing Was also playing with the idea to put all those attributes in a table. Maybe for my next PR...

PR #2584, opened 2020-12-08
Removing debug output.

jlahoda:
As noted by @junichi11 in #2324. Sorry for trouble!


junichi11 (approved):
Thank you for fixing it!

PR #2585, opened 2020-12-09
Enable full VSNetBeans Maven build test.

JaroslavTulach:
There has been many intermittent (or even persistent) errors when running the [VSNetBeans integration test](https://travis-ci.org/github/apache/netbeans/jobs/747857479). The test has been disabled in c9dc923f39d61b28f73dc886443956c8fe4d6d6d - this PR is an attempt to investigate what is wrong and make the test more robust.


JaroslavTulach:
My integration test for VSNetBeans (to compile a Maven project) seems to be working again with most recent `master`. I don't know what was the instability problem caused by, but it seems to be gone. Moreover there is now yet another test by @sdedic (since #2679) which also compiles (plus runs) a Maven project and it seems to be working fine as well.

I will have just run the [Build the Visual Studio Code extension for Java](https://travis-ci.org/github/apache/netbeans/jobs/755640803) test for three times. Shall it be successful I suggest to merge.

PR #2589, opened 2020-12-12
[NETBEANS-5142] LSP Client creates excessive processes

matthiasblaesing:
The caching of the LSP client that is rooted in project2MimeType2Server
does not work. In the WeakHashMap the keys are help by weak references
and the key is invalidated, once it is GCed. The URIs that are used
as keys become immediately eligible for GC and thus the cache is not
used at all.

To fix this, use a regular HashMap with WeakReference values.

The LSP Clients themselves are either strongly referenced from other
components or can be collected. To not create excessive start/stops of
servers, an additional strong reference is retained for a certain time
after the last bind request was issued. Only after the timeout has
occured, that extra reference is released and the bindings could become
eligible for GC.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
@@ -83,6 +83,7 @@ import org.openide.util.Lookup;
 import org.openide.util.NbBundle.Messages;
 import org.openide.util.RequestProcessor;
 import org.openide.util.RequestProcessor.Task;
+import org.openide.util.Utilities;
 import org.openide.util.WeakListeners;
 import org.openide.util.lookup.Lookups;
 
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
@jlahoda @JaroslavTulach you both worked in the area of the lsp client. I documented my reasoning in the commit message and the issue report but the TL;DR version is: I worked with angular applications and noticed, that the whole system became sluggish, investigating I found that a massive number of node processes was running as children of the netbeans process. That started my investigation and my finding is, that the caching/reuse of the LSP client does not work and every request for LSP bindings results in a new process being started.

jlahoda (approved):
The direction seems sensible to me. Instead of using a custom ReferenceQueue, I'd suggest to consider `Utilities.activeReferenceQueue()`, unless there is a reason to avoid it.

matthiasblaesing:
@jlahoda thank you for the pointer. The 'activeReferenceQueue`  looks indeed like the better (and simpler) solution.

JaroslavTulach:
> worked with .... applications and noticed, that the whole system became sluggish, investigating I found that a massive number of node processes was running as children of the netbeans process

I have actually seen the enormous multiplication of the `node` processes as well. Alas, I haven't concluded it is my own error.

JaroslavTulach (approved):

matthiasblaesing:
Thank you both - I'm pretty sure I introduced my own set of bugs :-)

PR #2594, opened 2020-12-14
Removing the need for a default JS engine in JDK to call ant tryme.

jlahoda:
nbbuild/build.xml uses JavaScript to construct tryme.args property from tryme.arg.* properties. This, however, does not work for JDK 15+, as there is no default JS engine for it. I was thinking of various ways to fix it, but, in the end, this does not seem like a particularly important feature, and my proposal is to simple remove the feature to joing tryme.arg.* properties into tryme.args property.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
@@ -19,6 +19,7 @@
 package org.netbeans.nbbuild;
 
 import java.util.Map.Entry;
+import java.util.TreeMap;
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Task;
 
@@ -29,7 +30,7 @@ import org.apache.tools.ant.Task;
 public class MergeTrymeArgs extends Task {
     public @Override void execute() throws BuildException {
         StringBuilder tryMeArgs = new StringBuilder();
-        for(Entry<String,Object> e: getProject().getProperties().entrySet()) {
+        for(Entry<String,Object> e: new TreeMap<>(getProject().getProperties()).entrySet()) {
             if(e.getKey().startsWith("tryme.arg.")) {
                 tryMeArgs.append(" ");
                 tryMeArgs.append(e.getValue());
--- a/nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
@@ -36,6 +36,6 @@ public class MergeTrymeArgs extends Task {
                 tryMeArgs.append(e.getValue());
             }
         }
-        getProject().setProperty("tryme.args", tryMeArgs.toString());
+        getProject().setNewProperty("tryme.args", tryMeArgs.toString());
     }
 }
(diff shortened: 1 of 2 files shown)


jlahoda:
Well, I was considering writing a task - but, frankly, what would be the point. We say all the Ant-based projects are effectively deprecated (we didn't even bother including any of them in the VS Code extension, AFAIK), and this is only the project for building NetBeans itself - not user applications. And I doubt this is a feature broadly used. I guess we should decide that either the Ant-base projects are deprecated, and we are allowed to keep shortcuts in maintaining them, or they are not deprecated, and then we need to treat them properly on all places.

matthiasblaesing:
I agree with @JaroslavTulach and I'm putting my work where my mouth is:

https://github.com/matthiasblaesing/netbeans/tree/pr-2594

I added:

https://github.com/matthiasblaesing/netbeans/commit/bee5912cd664a9747bcefca78b3412f73ceee392

Feel free to copy/pull. That should be a drop-in replacement and I think it prevents a regression where the "tryme.arg.hack" would not be picked up.

jlahoda:
Thanks Matthias! I've included your patch in this PR.

nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
```diff
@@ -0,0 +1,41 @@
...
+public class MergeTrymeArgs extends Task {
+    public @Override void execute() throws BuildException {
+        StringBuilder tryMeArgs = new StringBuilder();
+        for(Entry<String,Object> e: new TreeMap<>(getProject().getProperties()).entrySet()) {
+            if(e.getKey().startsWith("tryme.arg.")) {
+                tryMeArgs.append(" ");
+                tryMeArgs.append(e.getValue());
+            }
+        }
+        getProject().setProperty("tryme.args", tryMeArgs.toString());
```
JaroslavTulach, line 39: The JavaScript code is using `setNewProperty` - e.g. don't override existing previous definition. This code is using `setProperty`.

JaroslavTulach (approved):
I am glad the `<script language="javascript">` is going to be gone. It is not working properly with GraalVM either (too secure sandbox for JavaScript execution by default). Java version is going to be way more portable across JVMs. Thanks.

nbbuild/antsrc/org/netbeans/nbbuild/MergeTrymeArgs.java
```diff
@@ -0,0 +1,41 @@
...
+package org.netbeans.nbbuild;
+
+import java.util.Map.Entry;
+import java.util.TreeMap;
+import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.Task;
+
+/**
+ * Merge the arguments supplied via the tryme.arg.* properties into the
+ * property tryme.args.
```
jglick, line 28:
```suggestion
 * Merge the arguments supplied via any {@code tryme.arg.*} properties into the
 * property {@code tryme.args}.
```

nbbuild/build.xml
```diff
@@ -1189,22 +1189,10 @@ PKGINST=${pkg.svr4.pkginst}</echo>
...
-                e = i.next()
-                if (e.getKey().startsWith('tryme.arg.')) {
-                    v += ' ' + e.getValue()
-                }
-            }
-            project.setNewProperty('tryme.args', v)
-        }
-    </script>
+    <taskdef name="mergetrymeargs" classname="org.netbeans.nbbuild.MergeTrymeArgs" classpath="${nbantext.jar}" />
+    <mergetrymeargs />
```
jglick, line 1193:
```suggestion
    <taskdef name="mergetrymeargs" classname="org.netbeans.nbbuild.MergeTrymeArgs" classpath="${nbantext.jar}"/>
    <mergetrymeargs/>
```

jglick (approved):
Thanks!

PR #2595, opened 2020-12-14
[NETBEANS-2387] Gradle Task Navigator with Favorites

lkishalmi:
Several fixes around the Gradle Task Navigator. It has a Favorites parts now.
See the concept here: https://issues.apache.org/jira/secure/attachment/13017035/GradleFavoritesCut.mp4
I'm not sure if this would be the best design. 
Maybe another Group in the Navigator Tree as of "Favorites" or Just keep the last few executed tasks as "Recently Used"


lkishalmi:
Well this one is with the In-Tree design. The Favorite Task are persisted as well. I'm unsure of the FavoriteTaskManager implementation. It might not worth to have it. Will see it in the future.

PR #2596, opened 2020-12-15
[NETBEANS-5151] - update java.source in the project properties

BradWalker:
There are hundreds of warnings that are generated because the java.source
specifies version 1.6 in the project.properties files. The warning looks like this:

...
-compile-ant:
    [javac] Compiling 2 source files to /home/bwalker/src/netbeans/ergonomics/ide.ergonomics/build/classes
 [repeat] warning: [options] bootstrap class path not set in conjunction with -source 1.6
   [repeat] 1 warning
     [copy] Copying 6 files to /home/bwalker/src/netbeans/ergonomics/ide.ergonomics/build/classes

-generate-proxies:
...

I've cleaned this up. There are still a few of these warnings left. I will
clean them up when I have time to go in and inspect each one individually.

Since Java 8 is our default, cleaning these up makes those warnings go away.


matthiasblaesing:
If I remember correctly, not the complete netbeans code base targets java 8, there are parts, that are loaded into the target vm (debug+profiling) and these should be as low as reasonable possible. Could you please ensure, that this PR only touches the parts, that are build into the platform, but not used in the target VM?

matthiasblaesing:
@BradWalker is this still relevant? I think it was superseeded by other PRs. Feel free to reopen and rebase it.

PR #2597, opened 2020-12-15
[NETBEANS-5070] fix errors on text selection

hlavki:
I think the best solution would be to rewrite a lot of `ide/lib.terminalemulator` code. Anyway this PR fixes text selection problem which was brought by #2509.

### DIFF
--- a/ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/Line.java
+++ b/ide/lib.terminalemulator/src/org/netbeans/lib/terminalemulator/Line.java
@@ -156,13 +156,13 @@ final class Line {
     }
 
     public int stringWidth(MyFontMetrics metrics, int at) {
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
Thank you - handling is much better with this on "normal res" (scale 1x). I see a slight problem in selection - it seems there is an of-by-one problem:

Bildschirmfoto_2020-12-17_21-33-25

As you can see the dot in "dummy.bpmn" is not select, but it is part of the exported data:

```
workspace   Öffentlich     sketchbook  'System Volume Information'   Tracks  'VirtualBox VMs'
 Bilder   dokumentationen   dummy.
```

lkishalmi:
Tested. Works.

lkishalmi:
Thank you!

matthiasblaesing:
I just reopened the issue, there is still a problem on windows.

PR #2598, opened 2020-12-16
[NETBEANS-4550] Handling tests that have a module-info overriding module-info from sources.

jlahoda:
So, the problem here is that some Maven projects have a `module-info` for module `$m` in `src/main/java` and another for `$m` in `src/test/java`, which overrides the main one. (Note the module names are the same.) The `maven-compiler-plugin` will then put the `src/main/java` on patch path for (test) `$m`:
https://github.com/apache/maven-compiler-plugin/blob/1b206fe40156bb9814ed8b804a23362ec2b41daa/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java#L335

This is an attempt to simulate this inside NetBeans, although I am by far not sure if it is correct. It:
-does the tweak in the module projects to include the `--patch-module` option
-tweaks `PatchModuleManager`, so that if we have caches for the main sources, we use the caches
-tweaks the `ClassPath` content so that the "IDE-view" classpath for tests was a reference to the main sources, so that navigation from tests to sources works.

### DIFF
--- a/java/java.api.common/src/org/netbeans/modules/java/api/common/classpath/ModuleClassPaths.java
+++ b/java/java.api.common/src/org/netbeans/modules/java/api/common/classpath/ModuleClassPaths.java
@@ -870,6 +870,7 @@ final class ModuleClassPaths {
                                     if (myModule != null) {
                                         dependsOnUnnamed = dependsOnUnnamed(myModule, true);
                                         requires.addAll(collectRequiredModules(myModule, myModuleTree, true, false, modulesByName));
+                                        requires.addAll(modulesByName.getOrDefault(myModule.getQualifiedName().toString(), Collections.emptyList()));
                                     } else if (base == systemModules) {
                                         //When module unresolvable add at least java.base to systemModules
                                         Optional.ofNullable(modulesByName.get(MOD_JAVA_BASE))
--- a/java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/classpath/ModuleClassPathsTest.java
+++ b/java/java.api.common/test/unit/src/org/netbeans/modules/java/api/common/classpath/ModuleClassPathsTest.java
@@ -102,9 +102,11 @@ public class ModuleClassPathsTest extends NbTestCase {
                                 .anyMatch((ed) -> ed.getTargetModules() == null);
     
     private ClassPath src;
+    private ClassPath testSrc;
     private ClassPath systemModules;
     private FileObject automaticModuleRoot;
     private FileObject jarFileRoot;
+    private FileObject target;
     private TestProject tp;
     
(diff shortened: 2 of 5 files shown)


matthiasblaesing:
@jlahoda the issue author tested the change and found it fixed the issue for him. I had a look at the change and I did not see obvious problems, but I'm not an expert in that area. Should we hold back merging or can it go in from your POV?

PavelKastornyy:
Please, don't hold back merging. This feature is really necessary. Maven surefire and failsafe have been been supporting in-module testing for a long time.

jlahoda:
Ok, lets try with this patch, and fix if it breaks something.

sormuras:
I'll give it try, too. Soon™️.

PavelKastornyy:
@sormuras Apache NetBeans 12.3-beta1 available for testing. If you want to take a look, it is the right time.

sormuras:
Tried to start Apache NetBeans 12.3-beta3 and got an error.

image

Is Java 16 (it's on the PATH) too high?

geertjanw:
Yes, try with 15 or lower to see if you can reproduce.

sormuras:
Same same.

image

sormuras:
My fault. Setting `netbeans_jdkhome="D:\dev\jdk\jdk-15.0.1"` in `etc/netbeans.conf` did the trick.

geertjanw:
Great. Try the same for JDK 16, though not supported yet, should work too.

sormuras:
Even starts with `netbeans_jdkhome="D:\dev\jdk\jdk-16"`. 👍

sormuras:
Back on topic.

Project under test: https://github.com/sormuras/testing-in-the-modular-world/
Created `src/test/java/module-info.java` with the following content:

```java
open /*test*/ module foo {
  exports foo;
  requires org.junit.jupiter.api;
}
```
... which is what I invented [module-info.test](https://github.com/sormuras/testing-in-the-modular-world/blob/master/src/test/resources/module-info.test) for, back in the days.

And... it looks good to me.

image

Caveat: I'm not sure why Maven/NB says "1 tests skipped." Did look into the project for years...

sormuras:
Methinks, this is a feature that even IntelliJ doesn't support: see https://youtrack.jetbrains.com/issue/IDEA-171419 and linked issues.

/cc @akozlova

geertjanw:
Tell the world, tweet it, @sormuras. :-)

sormuras:
Done: https://twitter.com/sormuras/status/1362465668076044299

PashaTurok:
@jlahoda  Could you take a look at this [issue ](https://issues.apache.org/jira/browse/NETBEANS-5704)? It is a small bug.

PR #2599, opened 2020-12-16
Illegal accesses are not allowed by default now, adding javac-related…

jlahoda:
… add-exports/add-opens.


matthiasblaesing:
Looks harmless, but one question remains: What happens once illegal access is really shut down? Are you aware of efforts to expose the API for javac in a supported manner?

jlahoda:
> Looks harmless, but one question remains: What happens once illegal access is really shut down? Are you aware of efforts to expose the API for javac in a supported manner?

I am not aware of any plans to remove "--add-exports"/"--add-opens", so using them is likely to be less of a problem for some time. Of course the non-API opened by those can change. I am not aware of systemic efforts being done to avoid the dependencies on the non-APIs.

matthiasblaesing:
You are right, I mangled the options, it is `--illegal-access` which changed its default to `deny` (JEP 396). Sorry for the confusion.

PR #2600, opened 2020-12-16
First attempt add LSP-based rename refactoring to the Java LSP Server and the LSP Client.

jlahoda:


matthiasblaesing:
I think this is worth merging. I looked only at the client part though. I suggest to apply the commits from this branch:

https://github.com/matthiasblaesing/netbeans/tree/pr-2600:

https://github.com/matthiasblaesing/netbeans/commit/4c865767568f091041c03d511757a533099d4427: Merge master
https://github.com/matthiasblaesing/netbeans/commit/52feaf755c926ee8af6a56993580c15b85df5397: Implement basic support for cancelability (primary intention: don't raise an `UnsupportedOperationException`.
https://github.com/matthiasblaesing/netbeans/commit/26487ddd3a7285fbbcd25eb8c624536af016870b: Register the rename refactoring into the menu of the filetypes
https://github.com/matthiasblaesing/netbeans/commit/1618fb4bfd9c5c8060b5ea4d2b6e065e5824671f: Prevent exceptions about locking

@jlahoda It would be great to get this in.

jlahoda:
Thanks @matthiasblaesing! I've merged your changes into this PR!

matthiasblaesing:
I would like to see this merged - I think the lsp.client part looks good to me, but I can't speak about the server part - @jlahoda do you want to give it more time? I see, that you are already working on the next addition, which is great, but we should also see, that we get things in.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -192,12 +205,14 @@
 
     private static final RequestProcessor BACKGROUND_TASKS = new RequestProcessor(TextDocumentServiceImpl.class.getName(), 1, false, false);
     private static final RequestProcessor WORKER = new RequestProcessor(TextDocumentServiceImpl.class.getName(), 1, false, false);
+    private static final Set<TextDocumentServiceImpl> ALL_INSTANCES = Collections.newSetFromMap(new WeakHashMap<>());
```
dbalek, line 208: What is `Set<TextDocumentServiceImpl> ALL_INSTANCES` good for? It is only used from `static void refreshAllEditors()` that seems to be unused.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -1410,20 +1556,37 @@ public static int getOffset(Document doc, Position pos) {
...
-            edits.add(new TextEdit(new Range(Utils.createPosition(lm[0], diff.getStartPosition().getOffset()),
-                                             Utils.createPosition(lm[0], diff.getEndPosition().getOffset())),
+            edits.add(new TextEdit(new Range(offset2Position.apply(diff.getStartPosition().getOffset()),
+                                             offset2Position.apply(diff.getEndPosition().getOffset())),
                                    newText != null ? newText : ""));
         }
         return edits;
     }
+    
+    static void refreshAllEditors() {
```
dbalek, line 1581: Seems to be unused.

dbalek (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
jlahoda: Sorry, this was part of a (failed) attempt to fix an unrelated problem and should not be here. Removed in 5987977.

PR #2601, opened 2020-12-16
LSP: Hoover the mouse over identifier shows JavaDoc.

dbalek:
- Hoover the mouse over identifier shows JavaDoc
- Generate Constructor source action throws InvocationTargetException


java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -888,6 +891,73 @@ public void logMessage(MessageParams arg0) {
...
+        Hover hover = server.getTextDocumentService().hover(new HoverParams(new TextDocumentIdentifier(toURI(src)), new Position(2, 10))).get();
+        assertNotNull(hover);
+        assertTrue(hover.getContents().isRight());
+        MarkupContent content = hover.getContents().getRight();
+        assertNotNull(content);
+        assertEquals(content.getKind(), "markdown");
+        assertEquals(content.getValue(), "**[java.​lang](*0)**\n" +
+                "\n" +
+                "```\n" +
+                "public final class System\n" +
```
JaroslavTulach, line 943: Isn't this test too JDK specific? What if the wording of the Javadoc changes between the various JDKs? Then the test would fail.

It might be safer to create own class with Javadoc and test against it. That one would be stable.

JaroslavTulach (approved):

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
dbalek: Good catch. Fixed.

dbalek:
Failing PHP Debugger test is not related to changes in this PR.

PR #2603, opened 2020-12-17
[NETBEANS-5155] retain fallback to file's MIME type.

sdedic:
When reviewing, please compare also the [original semantics (before the refactoring)](https://github.com/apache/netbeans/commit/580d925e0608690eb3462aad37de15cdf0259853#diff-37427eeda3bcf06f91cce8d11efead9489dbe75a90f4d7433c183846c8a43577L89) with this fixed behaviour.


junichi11 (approved):
Verified that NPE doesn't occur. Thank you for your quick fix!

junichi11:
Regarding the failure of PHP debugger tests: It should be fixed in #2604 .

junichi11:
Merging. Thanks!

PR #2605, opened 2020-12-18
Existing CreateElement fixes exposed as CodeActions in LSP

dbalek:
- Create missing local variable, parameter, field, method, and class fixes added.
- Generate Overrride method puts in misleading comment - fixed.


java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateFixBase.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.IOException;
+import org.netbeans.api.java.source.ModificationResult;
+import org.netbeans.spi.editor.hints.Fix;
+
+/**
+ *
+ * @author Dusan Balek
+ */
+public abstract class CreateFixBase implements Fix {
```
JaroslavTulach, line 29: This might deserve some versioning, but there is implementation dependency in java.lsp.server anyway:
```
                    <code-name-base>org.netbeans.modules.java.hints</code-name-base>
                    <build-prerequisite/>
                    <compile-dependency/>
                    <run-dependency>
                        <release-version>1</release-version>
                        <implementation-version/>
                    </run-dependency>
```

JaroslavTulach (approved):

PR #2606, opened 2020-12-20
[NETBEANS-5161] Prevent IAE when resolving composite project dependencies

lkishalmi:
This one just prevents the IAE from happening on composite projects. I need a testcase and find out how to handle the dependency issues.


neilcsmith-net:
@lkishalmi as it has work-in-progress on it, just removing milestone rather than adding 12.4.  Let us know if it is actually intended for release.

lkishalmi:
Finally I had the chance to test and polish this fix. Would be great to be included in the 12.3 as right now those few Gradle projects which has included builds are not loadable at all. Les worse side effect that the included build won't show up in the configuration nodes, though the classpath would be corecly evaluated for such projects.

neilcsmith-net (approved):

neilcsmith-net:
Great, thanks, let's get it in!

PR #2607, opened 2020-12-21
[NETBEANS-5122] fix problem when selecting selenium test framework immediately modifies pom.xml

jherkel:
see https://issues.apache.org/jira/projects/NETBEANS/issues/NETBEANS-5122


matthiasblaesing:
Change looks sane to me. Are you aware, that the change is required to be donated to the ASF (see the comment at the top of the source files)? If you agree, I'm willing to merge.

jherkel:
I'm OK with that.

matthiasblaesing:
Thank you.

PR #2608, opened 2020-12-25
[NETBEANS-5185] - cleanup imports in the Enterprise Web.Monitor module

BradWalker:
In cleaning up Vector raw types, I noticed a bunch of imports in the
Enterprise Web.Monitor module that could use some cleanup.

It's bad form to use the "*" when importing. It's pulls in every class in the import.

In addition, it can break  modularization.


matthiasblaesing:
Thank you, looks sane.

PR #2609, opened 2020-12-26
update default imports for JShell

jherkel:
When I tried some code with JShell I found out that there were differencies in default imports between JShell from jdk and internal JShell from Netbeans. Default imports from JDK JShell :
jshell> /imports
|    import java.io.*
|    import java.math.*
|    import java.net.*
|    import java.nio.file.*
|    import java.util.*
|    import java.util.concurrent.*
|    import java.util.function.*
|    import java.util.prefs.*
|    import java.util.regex.*
|    import java.util.stream.*

Netbeans JShell :
[1]-> /imports
|    import java.util.*
|    import java.io.*
|    import java.math.*
|    import java.net.*
|    import java.util.concurrent.*
|    import java.util.prefs.*
|    import java.util.regex.*

I think it would be better if these default imports would be the same.

### DIFF
--- a/java/jshell.support/src/org/netbeans/modules/jshell/tool/JShellTool.java
+++ b/java/jshell.support/src/org/netbeans/modules/jshell/tool/JShellTool.java
@@ -215,13 +215,16 @@ public class JShellTool implements MessageHandler {
 
     static final String DEFAULT_STARTUP =
             "\n" +
-            "import java.util.*;\n" +
             "import java.io.*;\n" +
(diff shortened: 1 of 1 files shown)


jlahoda (approved):
I agree, seems sensible.

@matthiasblaesing - do you see a problem if we integrate this? Thanks!

matthiasblaesing:
Looks sane to me - aligning with upstream sounds like a good idea.

PR #2610, opened 2020-12-27
add extra cases to SizeEqualsZero java hints

bradh:
Adds support for size() > 0, size() < 1, and size() <=0.

I think it would be useful to have unit tests for this, but I need a hint on how to write them.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
@@ -51,10 +51,29 @@ public class SizeEqualsZero {
     @BooleanOption(displayName = "#LBL_org.netbeans.modules.java.hints.perf.SizeEqualsZero.CHECK_NOT_EQUALS", tooltip = "#TP_org.netbeans.modules.java.hints.perf.SizeEqualsZero.CHECK_NOT_EQUALS", defaultValue=CHECK_NOT_EQUALS_DEFAULT)
     public static final String CHECK_NOT_EQUALS = "check.not.equals";
 
+    @BooleanOption(displayName = "#LBL_org.netbeans.modules.java.hints.perf.SizeEqualsZero.CHECK_GREATER_ZERO", tooltip = "#TP_org.netbeans.modules.java.hints.perf.SizeEqualsZero.CHECK_GREATER_ZERO", defaultValue=CHECK_NOT_EQUALS_DEFAULT)
+    public static final String CHECK_GREATER_ZERO = "check.greater_zero";
+
     @TriggerPatterns({
         @TriggerPattern(value="$subj.size() == 0", 
             constraints = @ConstraintVariableType(type = "java.util.Collection", variable = "$subj")),
         @TriggerPattern(value="$subj.size() == 0", 
+            constraints = @ConstraintVariableType(type = "java.util.Map", variable = "$subj")),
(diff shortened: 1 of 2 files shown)


java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
```diff
@@ -51,11 +51,22 @@
...
         @TriggerPattern(value="$subj.size() == 0", 
             constraints = @ConstraintVariableType(type = "java.util.Map", variable = "$subj")),
+        @TriggerPattern(value="$subj.size() <= 0",
+            constraints = @ConstraintVariableType(type = "java.util.Collection", variable = "$subj")),
+        @TriggerPattern(value="$subj.size() <= 0",
+            constraints = @ConstraintVariableType(type = "java.util.Map", variable = "$subj")),
+        @TriggerPattern(value="$subj.size() < 1",
+            constraints = @ConstraintVariableType(type = "java.util.Collection", variable = "$subj")),
+        @TriggerPattern(value="$subj.size() < 1",
+            constraints = @ConstraintVariableType(type = "java.util.Map", variable = "$subj")),
```
sdedic, line 69: Shouldn't be here also the flipped conditions like `1 > $subj.size()` ?

java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
```diff
@@ -82,6 +93,23 @@ public static ErrorDescription sizeNotEqualsZero(HintContext ctx) {
...
+        @TriggerPattern(value="0 < $subj.size()",
+            constraints = @ConstraintVariableType(type = "java.util.Collection", variable = "$subj")),
+        @TriggerPattern(value="0 < $subj.size()",
+            constraints = @ConstraintVariableType(type = "java.util.Map", variable = "$subj")),
+    })
+    public static ErrorDescription sizeGreaterZero(HintContext ctx) {
+        if (!ctx.getPreferences().getBoolean(CHECK_GREATER_ZERO, CHECK_NOT_EQUALS_DEFAULT)) {
+            return null;
+        }
+        return sizeEqualsZeroHint(ctx, true);
```
sdedic, line 110: is `not == true` correct in this invocation ?

java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
bradh: I will add them.

java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
bradh: I believe so. Its essentially the same logic as the previous set of trigger patterns - another way of saying `sizeNotEqualsZero`.

java/java.hints/src/org/netbeans/modules/java/hints/perf/SizeEqualsZero.java
bradh: Now added.

mbien:
@bradh unfortunately I overlooked this PR when https://github.com/apache/netbeans/pull/3164 got integrated :(

It is lacking the size() < 1 case though. If you still would like to contribute you could rebase to master and add the case + update the test just like the linked PR did.

You could start a new PR or reuse this one, whatever you prefer.

PR #2612, opened 2020-12-28
[NETBEANS-4274] fix refactoring of class member when a new name is the same as the name of a local variable

jherkel:
see https://issues.apache.org/jira/browse/NETBEANS-4274

### DIFF
--- a/java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/RenameTransformer.java
+++ b/java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/RenameTransformer.java
@@ -254,17 +254,11 @@ public class RenameTransformer extends RefactoringVisitor {
             String useThis = null;
             String useSuper = null;
 
-            if (elementToFind!=null && elementToFind.getKind().isField() && tree.getKind() != Tree.Kind.MEMBER_SELECT) {
+            if (elementToFind!=null && elementToFind.getKind().isField() && tree.getKind() == Tree.Kind.IDENTIFIER) {
                 Scope scope = workingCopy.getTrees().getScope(elementPath);
                 for (Element ele : scope.getLocalElements()) {
                     if ((ele.getKind() == ElementKind.LOCAL_VARIABLE || ele.getKind() == ElementKind.PARAMETER) 
                             && ele.getSimpleName().toString().equals(newName)) {
-                        if (tree.getKind() == Tree.Kind.MEMBER_SELECT) {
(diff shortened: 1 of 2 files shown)


java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/RenameTransformer.java
```diff
@@ -254,7 +254,7 @@ public boolean accept(Element e, TypeMirror type) {
             String useThis = null;
             String useSuper = null;
 
-            if (elementToFind!=null && elementToFind.getKind().isField()) {
+            if (elementToFind!=null && elementToFind.getKind().isField() && tree.getKind() != Tree.Kind.MEMBER_SELECT) {
```
matthiasblaesing, line 257: Please add reasoning why you propose a change. In this case, the change looks wrong, as it contradicts the check in line 262. That branch can't be reached anymore with your change in place.

jherkel: Good point. I think the whole if statement started from line 262 can be deleted. Because this part of code (started from line 257) checks if a new name for refactored field doesn't have the same name as some local variables do. For this case it is adding an outer scope (this or class name for static fields). But I think if field is a member i.e. tree.getKind() == Tree.Kind.MEMBER_SELECT we don't need to do anything because we have a scope.

matthiasblaesing:
@sarveshkesharwani @rtaneja1 @jlahoda could you please have a look at this?

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/RenameTransformer.java
jlahoda: I think I agree with @jherkel - the code on lines 262-267 does not make much sense to me. If a reference to a field is qualified, it cannot clash with a local variable (AFAIK), even if the qualifier is different from `this`. I agree we should delete the whole `if` starting at 262. Maybe instead of `tree.getKind() != Tree.Kind.MEMBER_SELECT` do `&& tree.getKind() == Tree.Kind.IDENTIFIER`, because I believe that is the only case where the clash can occur.

geertjanw:
Any reason not to merge?

mbien:
added to NB13 milestone to put it back on the radar

neilcsmith-net:
Thanks @mbien Let's merge before branch and check in rc for any issues.

PR #2613, opened 2020-12-28
[NETBEANS-5114] add missing namespace for facelets composition template client

jherkel:
see https://issues.apache.org/jira/browse/NETBEANS-5114

### DIFF
--- a/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/facelets/resources/templates/compositionTemplateClient.template
+++ b/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/facelets/resources/templates/compositionTemplateClient.template
@@ -1,6 +1,7 @@
 <?xml version='1.0' encoding='UTF-8' ?>
 <!DOCTYPE composition PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<ui:composition xmlns:ui="__NS_LOCATION__/jsf/facelets"
+<ui:composition xmlns="http://www.w3.org/1999/xhtml"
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
I could not reproduce the problem - a reproducer on the issue would be good. While I understand, that the namespace declaration is necessary to get valid xml, the issue is about a different problem statement.

geertjanw:
Seems to be no clarity on this yet, are we sure this fixes the stated problem, and won't lead to side effects? Changing milestone to 12.5.

brooksie:
Geertjan, what specifically do you need to see in order to make you happy? 
The proposed fix is as simple as it gets - a 1 line fixed-text addition to a template file. It would be easy enough to amend the reproducer to demonstrate that the error goes away when the namespace is added - that's exactly what I do manually (and repeatedly) at the moment to resolve the issue myself as a netbeans user. Including it in the default generated template takes away those manual, repeated edits. But your question is wider than that - are there any situations where including this in the generated ui:composition header would break the code. Perhaps you need to pose your question to your Netbeans JSF expert, or to someone in the expert panel working on the JSF specifications? Bauke Scholtz comes to mind.
FYI here is a couple of postings on Stackoverflow from Bauke Scholtz, known there as BalusC, showing  the correct usage of ui:composition as a facelets template client. They all include the basic xmlns: 
https://stackoverflow.com/questions/4792862/how-to-include-another-xhtml-in-xhtml-using-jsf-2-0-facelets
https://stackoverflow.com/questions/8500894/how-to-change-head-elements-of-a-page-when-using-uicomposition
From wikipedia: https://en.wikipedia.org/wiki/Facelets#Templating (see "template client" heading)

matthiasblaesing:
Lets get this in - the xmlns is not strictly necessary, but people will likely add html, so it keeps them happy. As delivery has not yet branched, still merging.

PR #2614, opened 2020-12-29
[NETBEANS-5172] Fix anonymous function formatting

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5172

- Enabled and fixed tests for spaces around keyword.
- Fixed formatting of anonymous functions.
- Updated expected formatting results in rest of tests.

Sample code:
```php
$lambda = function($param) use($parent) {
    echo "$param\n";
};
```

Formatted after fix:
```php
$lambda = function ($param) use ($parent) {
    echo "$param\n";
};
```


junichi11 (approved):
I assume that you have run all tests in your local environment because GitHub actions are not available.

KacerCZ:
Yes, I run all tests in "php.editor.indent" package.

tmysik (changes requested):
Personally, I don't like this change (yes, I don't use this space in my code) - looking at the number of the changed files in this patch, this will have a big impact; moreover, there was no issue reported for the current behavior so far (it means for the last couple of years), right?

Can we make this configurable? If there was an option for it, I would not be against merging this PR.

KacerCZ:
Configuration is now bound to PHP > Spaces > Other > After Keywords.
I will add dedicated setting as PHP > Spaces > Before Parentheses > Anonymous Function (there is already setting for Anonymous Class).

By default it will be enabled because it is the way how anonymous functions are formatted in PHP manual - https://www.php.net/manual/en/functions.anonymous.php

tmysik:
@KacerCZ Great, thank you!

KacerCZ:
I added new formatting setting and related tests.
Also added anonymous function to code sample in options panel.

tmysik (approved):

tmysik:
@KacerCZ Thanks a lot!

@junichi11 Feel free to merge it if it is OK for you. Thanks!

PR #2617, opened 2020-12-29
First version of semantic highlight support for both the LSP client and Java LSP server.

jlahoda:
This is a new feature in the protocol version 3.16.0.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -154,7 +158,18 @@ public static NbLspServer launchServer(Pair<InputStream, OutputStream> io, LspSe
             .setInput(io.first())
             .setOutput(io.second())
             .wrapMessages(processor)
-//                .traceMessages(new java.io.PrintWriter(System.err))
+            .configureGson(gb -> {
```
dbalek, line 161: `configureGson` is already called couple of lines above - registrations should be merged.

dbalek (approved):
Looks fine. Thanks.

Chris2011:
Could this be already ready for NB 13?

PR #2621, opened 2020-12-30
[NETBEANS-5148] Added javafx:run to supported execgoals

mcdonnell-john:
The Gluon JavaFX templates use the org.openjfx:javafx-maven-plugin maven plugin to run, as a result, the maven goal javafx:run was never treated the same as a normal Java application which uses the org.codehaus.mojo:exec-maven-plugin maven plugin.

I've added the goal javafx:run to the list of supported exec goals in `ExecPluginOutputListenerProvider.java`


matthiasblaesing:
Change looks sane to me. The failing unittest is known to be flaky (I restarted the run for java.hints batch2).

svenreimers (approved):

PR #2628, opened 2020-12-31
[NETBEANS-819] Fix terminal scaling at scale > 2

matthiasblaesing:


matthiasblaesing:
With 172a9cf the fixed width grid can be displayed by running netbeans with:

```
bin\netbeans64.exe -J-Dorg.netbeans.lib.terminalemulator.Term.debug=32
```

or from a custom build:

```
ant -Dtryme.args="-J-Dorg.netbeans.lib.terminalemulator.Term.debug=32" tryme
```

hlavki:
@matthiasblaesing this is great! Thanks for this, it works as expected on my 4k display.

matthiasblaesing:
@hlavki thanks for checking.

PR #2629, opened 2021-01-01
[NETBEANS-5204] New Gradle Sub-project handling

lkishalmi:
It also shall fix some non-resolvable configuration processing, though that needs to be experimented further.
Also the documentation of the new API function is not ready yet...


lkishalmi:
image

extide/gradle/src/org/netbeans/modules/gradle/api/GradleProjects.java
```diff
@@ -88,6 +93,36 @@ public static File getJavadoc(File binary) {
...
+     * otherwise returns the Project which is assigned for the given project
+     * root directory.
+     *
+     * @param project the project to determine of its root.
+     *
+     * @return the root project or <code>null</code> if the given project is not
+     *         a Gradle project.
+     * @since 2.7
+     */
+    public static Project rootOf(Project project) {
```
JaroslavTulach, line 108: It might be more useful to return `project` if no parent/root project is known. Especially if:

This might be a useful concept in generic Project API. There already is [SubprojectProvider](https://bits.netbeans.org/12.0/javadoc/org-netbeans-modules-projectapi/org/netbeans/spi/project/SubprojectProvider.html) and [ProjectContainerProvider](https://bits.netbeans.org/12.0/javadoc/org-netbeans-modules-projectapi/org/netbeans/spi/project/ProjectContainerProvider.html). There could be `ParentProvider` and/or `RootProjectProvider` and the `rootOf` method could be moved to [ProjectUtils](https://bits.netbeans.org/12.0/javadoc/org-netbeans-modules-projectapi/org/netbeans/api/project/ProjectUtils.html). Then it could/should be implemented by Maven as well.

I can imagine a "Build All" action that would build the root project to make sure all bits are compiled and then continue with some other operation.

lkishalmi: That's a good idea, I could implement that as well.

lkishalmi:
So there is two new interfaces in the Project API ```RootProjectProvider``` and ```ParentProjectProvider```. I'm not sure if I shall have added change listener registration methods to them. I'd trust @JaroslavTulach experience on that matter, so if you see it would be wise to include those, I'd do that.
It is a bit unfortunate that this PR and the commit is entangled with two changes. If really required I could split those into two PR-s.

ide/projectapi/apichanges.xml
```diff
@@ -83,6 +83,23 @@ is the proper place.
...
+            <summary>Added <code>ParentProjectProvider</code> and <code>RootProjectProvider</code> to improve project hierarchy discovery</summary>
+            <version major="1" minor="79"/>
+            <date day="2" month="1" year="2021"/>
+            <author login="lkishalmi"/>
+            <compatibility addition="yes"/>
+            <description>
+                <p>
+                    <code><a href="@TOP@/org/netbeans/spi/project/ParentProjectProvider.html">ParentProjectProvider</a></code> and
+                    <code><a href="@TOP@/org/netbeans/spi/project/RootProjectProvider.html">RootProjectProvider</a></code> with correspondig
+                    utility methods in <code><a href="@TOP@/org/netbeans/api/project/ProjectUtils.html">RootProjectProvider</a></code>
```
JaroslavTulach, line 97: Here is a typo. The utility method is in `ProjectUtils`! CCing @sdedic

sdedic:
I filed [NETBEANS-5455](https://issues.apache.org/jira/browse/NETBEANS-5455) for myself; I think I need a clarification on what the API introduced in this PR means -- how to properly transpose it into Maven world. Pls. comment in NETBEANS-5455.

pekuz:
There is a strange asymmetry in the provider interfaces: 
  - get parent can return null 
  - get root can not return null 
  
Assuming a root should be reachable by traversing parents chain then allowing the null in the parents chain but not allowing the null for the root itself smells.

sdedic:
Well, that's a question: for `root` project, returning self makes perfect sense. From each project, one hop through `RootProjectProvider` leads to the root, whatever it is. In a tree-like structure, it's quite common (IMHO) that a parent is null. 

To make developer's life easier, there could be e.g. `ProjectUtils.projectContainment` that could return the whole sequence including self = never empty.

pekuz:
Then parent provider interface JavaDoc:

     The immediate parent of this project or <code>null</code> if there is no parent known.

could become:

     The immediate parent of this project or <code>null</code> if this project is the root.

no?

Originally I got it like null stays for unknown, e.g. because of incomplete source tree.

sdedic:
I guess the interface will need some love anyway, we have released it with nice typo right in the method's name :)

PR #2630, opened 2021-01-01
[NETBEANS-377] Add PSR-4 hints

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-377

- Added hints for [PSR-4 Autoloader](https://www.php-fig.org/psr/psr-4/) standard.
- Checks that class/interface/trait declaration matches file name.
- Checks that namespace matches directory structure.
- Highlights only non-matching part of namespace.

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/verification/Bundle.properties
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/verification/Bundle.properties
@@ -31,6 +31,7 @@ ImmutableVariablesCustomizer.jLabel1.text=Number of Allowed Assignments per Vari
 csl-hints/text/x-php5/hints/braces=Braces
 csl-hints/text/x-php5/hints/psr0=PSR-0: Autoloading Standard
 csl-hints/text/x-php5/hints/psr1=PSR-1: Basic Coding Standard
+csl-hints/text/x-php5/hints/psr4=PSR-4: Autoloader
 csl-hints/text/x-php5/hints/superglobals=Superglobals
 csl-hints/text/x-php5/hints/toomanylines=Too Many Lines
 NestedHintsCustomizer.jLabel1.text=Number of Allowed Nested Blocks:
--- a/php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
@@ -1,0 +1,262 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.php.editor.verification;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ListIterator;
+import org.netbeans.editor.BaseDocument;
+import org.netbeans.modules.csl.api.Hint;
+import org.netbeans.modules.csl.api.OffsetRange;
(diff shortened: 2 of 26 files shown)


junichi11:
We should leave a comment for unsupported cases somewhere.

Add test for the following case?
``` php
// file: Vendor/App/ClassName.php
namespace Vendor\App;

class classname {

}

```

php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
```diff
@@ -0,0 +1,258 @@
...
+        public CheckVisitor(PSR4Hint psr4hint, FileObject fileObject, BaseDocument baseDocument) {
+            this.psr4hint = psr4hint;
+            this.fileObject = fileObject;
+            this.baseDocument = baseDocument;
+            this.file = FileUtil.toFile(fileObject);
+            this.hints = new ArrayList<>();
+        }
+
+        public List<Hint> getHints() {
+            return hints;
```
junichi11, line 223: Use `Collections.unmodifiableList(hints)`?

KacerCZ:
I added tests where casing is different.

Also added note about case-sensitivity to hint descriptions:
- "Namespace MUST match (case-sensitive) file directory." 
- "Type name MUST match (case-sensitive) file name."

php/php.editor/src/org/netbeans/modules/php/editor/verification/Bundle.properties
```diff
@@ -31,6 +31,7 @@ ImmutableVariablesCustomizer.jLabel1.text=Number of Allowed Assignments per Vari
 csl-hints/text/x-php5/hints/braces=Braces
 csl-hints/text/x-php5/hints/psr0=PSR-0: Autoloading Standard
 csl-hints/text/x-php5/hints/psr1=PSR-1: Basic Coding Standard
+csl-hints/text/x-php5/hints/psr4=PSR-4: Autoloader
```
tmysik, line 34: Should not it be "Autoloading Standard", similarly to PSR-0?

php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
```diff
@@ -0,0 +1,259 @@
...
+            public void visit(NamespaceDeclaration node) {
+                if (CancelSupport.getDefault().isCancelled()) {
+                    return;
+                }
+                NamespaceName namespaceName = node.getName();
+                if (namespaceName != null) {
+                    int endOffset = namespaceName.getEndOffset();
+                    File currentDir = getFile().getParentFile();
+                    ListIterator<Identifier> segmentsIterator = namespaceName.getSegments().listIterator(namespaceName.getSegments().size());
+                    while (segmentsIterator.hasPrevious()) {
```
tmysik, line 99: Perhaps calling `CancelSupport.getDefault().isCancelled()` in this loop could be useful?

php/php.editor/src/org/netbeans/modules/php/editor/verification/Bundle.properties
KacerCZ: "Autoloader" is official name of PSR-4 - https://www.php-fig.org/psr/psr-4/

php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
KacerCZ: One iteration of the loop will be done for one part of namespace.
I expect no more than 10 iterations even if someone uses very deep structure.

Can traversing directories can take long time and getting its name take so much time it will be noticeable?

php/php.editor/src/org/netbeans/modules/php/editor/verification/Bundle.properties
tmysik: OK.

php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
tmysik: In general, this check should be present in every loop to avoid unneeded work to be done. The check is very cheap and IO operations are slow, in general (yes, I heard about SSD :) so I think it should be there.

KacerCZ: Right, I added it.

tmysik: Thanks!

tmysik (approved):

junichi11 (approved):

junichi11:
Will re-run jobs.

KacerCZ:

ahajzik-crido:
Does this feature support psr-4 declaration from composer.json?

KacerCZ:
@ahajzik-crido No, mapping from composer.json is separate task - https://issues.apache.org/jira/browse/NETBEANS-5341

PR #2631, opened 2021-01-01
[NETBEANS-5206] Enable LazyOpen on Gradle Projects

lkishalmi:
And deprecate experimental settings.


lkishalmi:
I'm merging this one as it's trivial and I need it to be a base of another change.

PR #2634, opened 2021-01-03
[NETBEANS-4783] Unselect added files from CDNJS by default

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4783

When adding new files from CDNJS new files are not selected by default.


KacerCZ:
@matthiasblaesing Thank you for review.

PR #2636, opened 2021-01-04
[NETBEANS-3428] FlatLaf: fix selection background color in output view

lkishalmi:
Back ported from #2327 (12.1)


matthiasblaesing (commented):
Looks sane to me.

matthiasblaesing (approved):

PR #2637, opened 2021-01-04
[NETBEANS-4627] FlatLaf: fix wrong background color of search result …

lkishalmi:
…table in Find Tasks Editor
Backported from #2326


matthiasblaesing (approved):
Looks sane.

PR #2638, opened 2021-01-04
[NETBEANS-4722] Add JAVA_HOME to Gradle build environment variables

lkishalmi:
Backport of #2320

### DIFF
--- a/groovy/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
+++ b/groovy/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
@@ -34,6 +34,8 @@ import java.io.OutputStream;
 import java.io.StreamCorruptedException;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


groovy/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
```diff
@@ -156,6 +158,9 @@ public void run() {
             if (!activePlatform.second().getInstallFolders().isEmpty()) {
                 File javaHome = FileUtil.toFile(activePlatform.second().getInstallFolders().iterator().next());
                 buildLauncher.setJavaHome(javaHome);
+                Map<String, String> envs = new HashMap<>(System.getenv());
+                envs.put("JAVA_HOME", javaHome.getCanonicalPath());
```
matthiasblaesing, line 162: This can throw an IOException, which is not yet caught and thus breaks the build.

matthiasblaesing (changes requested):

groovy/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
lkishalmi: Yes, that's true. Thank you for the check! It shall be fixed now.

PR #2639, opened 2021-01-04
[NETBEANS-5214] Fix handling of global values

matthiasblaesing:
When checking the global values, the input state needs to be reset after
the possible values are checked. Before this change the right grammar
element was found, but then could not be processed, because the state
of the input was already changed by the checking.


junichi11 (approved):
Looks good to me. Thank you!

PR #2640, opened 2021-01-05
[NETBEANS-4443] PHP 8.0 Support: Attribute Syntax

junichi11:

- https://wiki.php.net/rfc/attributes_v2
- https://wiki.php.net/rfc/shorter_attribute_syntax
- https://wiki.php.net/rfc/shorter_attribute_syntax_change

#### Part 1

- Fix the parser and lexers
- Add/Fix unit tests for the parser and the lexers

##### PHP 8.0

nb-php80-attribute-syntax-01
nb-php80-attribute-syntax-02

##### PHP 7.4

nb-php80-attribute-syntax-01-php74
nb-php80-attribute-syntax-02-php74

#### Part 2

- Fix the formatting feature

nb-php80-attribute-syntax-formatting-indent

#### Part 3

- Fix the braces matching feature

nb-php80-attribute-syntax-brace-matching


php/php.editor/tools/ASTPHP5Parser.cup
```diff
@@ -382,6 +409,7 @@ terminal T_POW_EQUAL;
 terminal T_ELLIPSIS;
 terminal T_COALESCE;
 terminal T_COALESCE_EQUAL;
+terminal T_ATTRIBUTE;
```
junichi11, line 412: To avoid the incompatible api change, always add new one to the last position.

php/php.editor/tools/ASTPHP5Parser.cup
```diff
@@ -2419,6 +2547,13 @@ is_variadic:isVariadic expr_without_variable:var
     RESULT = paramsList;
 :}
 
+| error:expr
+{:
+    List paramsList = new LinkedList();
+    paramsList.add(new ASTErrorExpression(exprleft, exprright));
+    RESULT = paramsList;
+:}
+
```
junichi11, line 2556: To avoid `ASTError`, add `ASTErrorExpression`. 
e.g. 
If we don't add `ASTErrorExpression`, the following (whole statement)is recognized as `ASTError`. In such case, some features(e.g. formatting, CC) may not work correctly. Plus, we have to sanitize the code and parse it again.
```
#[A(1, self::)]
class Test {}
```

php/php.editor/tools/ASTPHP5Parser.cup
```diff
@@ -2445,6 +2580,12 @@ is_variadic:isVariadic expr_without_variable:var
     paramsList.add(var_ref);
     RESULT = paramsList;
 :}
+
+| non_empty_function_call_parameter_list:paramsList T_COMMA error:expr
+{:
+    paramsList.add(new ASTErrorExpression(exprleft, exprright));
+    RESULT = paramsList;
+:}
```
junichi11, line 2588: The same as the above.

php/php.editor/tools/ASTPHP5Parser.cup
```diff
@@ -192,6 +192,28 @@ parser code {:
...
+            attributedStatement = FieldsDeclaration.create((FieldsDeclaration) statement, attributes);
+        } else if (statement instanceof ConstantDeclaration) {
+            attributedStatement = ConstantDeclaration.create((ConstantDeclaration) statement, attributes);
+        } else if (statement instanceof MethodDeclaration) {
+            attributedStatement = MethodDeclaration.create((MethodDeclaration) statement, attributes);
+        } else {
+            assert false;
+        }
+        return attributedStatement;
+    }
```
junichi11, line 215: I've added the static create methods to each node to add attributes (recreate the node with attributes).

php/php.editor/test/unit/data/testfiles/formatting/broken/issue197074_04.php.formatted
```diff
@@ -21,25 +21,28 @@ class companyAddressesActions extends myFrontModuleActions {
         }
 
         if ($this->getPage()->getModuleAction() == 'supplier/editCompanyAddress')
-            $this->getRequest ()->setParameter ('cancel_url', $this->getHelper ()main/checkout')
+        $this->getRequest ()->setParameter ('cancel_url', $this->getHelper ()main/checkout')
```
junichi11, line 24: The cause of this change is I added `ASTErrroExpression` to the parameter (grammar file).
Original issue is https://bz.apache.org/netbeans/show_bug.cgi?id=197074

formatting behavior is changed, but original code has syntax error and the purpose of the original issue is to prevent AssertionError. So, I suppose there is no problem.

junichi11:
I have to improve/fix other features of Attribute Syntax (e.g. CC, model). However, I'm going to fix other syntaxes before that.

KacerCZ:
**Thank you for your work on this big feature.**

I tested typing, brace matching, different formatting options and it worked as expected.

The only thing that worked differently than I expected was blank lines (more their lack of) between fields with attributes.
Box for Formatting > PHP > Blank Lines > Group Fields with PHP Doc is checked.
I expected that fields with attribute will be treated as fields with PHPDoc and will have blank line between them.
The same applies to constants

Expected result:
```php
class Person {

    #[Required]
    public string $name;

    #[Required]
    public string $login;

    /**
     * ISO code of country
     */
    public string $country;
    public string $street;
    public string $city;
```

Actual result:
```php
class Person {

    #[Required]
    public string $name;
    #[Required]
    public string $login;

    /**
     * ISO code of country
     */
    public string $country;
    public string $street;
    public string $city;
```

tmysik (approved):
The patch is huge because the change is huge. So, again, hard to review carefully :) However, we have plenty of tests so we should be safe.

php/php.editor/tools/ASTPHP5Parser.cup
tmysik: I see.

php/php.editor/tools/ASTPHP5Parser.cup
tmysik: Great.

php/php.editor/test/unit/data/testfiles/formatting/broken/issue197074_04.php.formatted
tmysik: Agreed.

tmysik:
@junichi11 

I can only repeat @KacerCZ comment above - thanks a lot, Junichi! This is a lot of work and I cannot image that anyone else than you could do it. Grealy done!

tmysik:
@junichi11 

Some checks are still not finished - feel free to merge it or I will do it later. Thanks!

junichi11:
@tmysik 
> The patch is huge because the change is huge. So, again, hard to review carefully :) However, we have plenty of tests so we should be safe.

Yes, I think so :)

Thank you for reviewing this, Tomas!

Merging.

PR #2643, opened 2021-01-06
Display Java + Truffle guest frames correctly.

entlicher:
There are two important changes:
1) We request a combination of Java and guest language frames from Truffle via `DebuggreSession.setShowHostStackFrames`, when available.
2) When suspended in Java code, we search for guest language frames underneath via `DebuggerSession.suspendHere`, when available and when Truffle backend was installed.

This functionality is available on GraalVM 20.3 and newer only.
In order to intercept the guest language frames from within the `DebuggerSession.suspendHere` call, we need to intercept a breakpoint event from a breakpoint submitted in `onSuspend` callback. This is why we're adding `Operator.addEventInterceptor`.

Since we're adding guest language frames always when they're available, even when suspended in Java code, we need an explicit way how to uncover all the underlying Java frames, when necessary. This is why we're adding `Show All Host Frames` context action on threads with guest frames.


java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/TruffleStackInfo.java
```diff
@@ -41,19 +41,21 @@
...
     private final JPDADebugger debugger;
     private final JPDAThread thread;
     private final ObjectVariable stackTrace;
+    private final boolean supportsJavaFrames;
     private TruffleStackFrame[] stackFrames;
     private boolean includedInternalFrames;
     private boolean areInternalFrames;
 
-    public TruffleStackInfo(JPDADebugger debugger, JPDAThread thread, ObjectVariable stackTrace) {
+    public TruffleStackInfo(JPDADebugger debugger, JPDAThread thread, ObjectVariable stackTrace, boolean supportsJavaFrames) {
```
JaroslavTulach, line 54: Sigtest reports:
```
 "E5.4 - Removing constructor" : 
constructor public org.netbeans.modules.debugger.jpda.truffle.frames.TruffleStackInfo.init(
  org.netbeans.api.debugger.jpda.JPDADebugger,
  org.netbeans.api.debugger.jpda.JPDAThread,
  org.netbeans.api.debugger.jpda.ObjectVariable
)
```
please add new constructor, keep the old one.

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/models/DebuggingTruffleActionsProvider.java
```diff
@@ -96,20 +122,60 @@ public void performDefaultAction(NodeActionsProvider original, Object node) thro
             }
             return actions;*/
         } else {
-            return original.getActions(node);
+            Action[] actions = original.getActions(node);
+            if (node instanceof DebuggingView.DVThread && hasGuestInfo(node)) {
+                Action showAction;
+                if (!isShowAllHostFrames((DebuggingView.DVThread) node)) {
+                    showAction = SHOW_HOST_ACTION;
```
JaroslavTulach, line 129: There is going to be an action "Show/Hide Java Frames", great!

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/source/Source.java
```diff
@@ -118,14 +124,16 @@ public static URI getTruffleInternalURI(FileObject fo) {
...
                                    String path,
                                    URI uri,
                                    StringReference codeRef) {
-        return getSource(debugger, id, name, path, uri, null, codeRef);
+        return getSource(debugger, id, name, hostMethodName, path, uri, null, codeRef);
     }
 
     public static Source getSource(JPDADebugger debugger, long id,
                                    String name,
+                                   String hostMethodName,
```
JaroslavTulach, line 136: Class org.netbeans.modules.debugger.jpda.truffle.source.Source
Sigtest reports:
```
  "E1.2 - API type removed" : method 
public final static org.netbeans.modules.debugger.jpda.truffle.source.Source Source.getSource(
  org.netbeans.api.debugger.jpda.JPDADebugger,
  long,
  java.lang.String,java.lang.
  String,java.net.URI,
  com.sun.jdi.StringReference
)
```
consider keeping the old method and passing `null` as the `hostMethodName`.

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/models/DebuggingTruffleTreeModel.java
```diff
@@ -47,15 +50,15 @@
 import org.netbeans.spi.viewmodel.TreeModelFilter;
 import org.netbeans.spi.viewmodel.UnknownTypeException;
 
-@DebuggerServiceRegistration(path="netbeans-JPDASession/"+TruffleStrataProvider.TRUFFLE_STRATUM+"/DebuggingView",
```
JaroslavTulach, line 50: Introducing new path for Truffle? That forms an API and might be good to be documented somewhere - for example in `arch.xml` with `group="layer"`. But I understand that the content of `netbeans-JPDASession` registration hasn't been documented in great detail yet...

JaroslavTulach (changes requested):
I suggest to keep binary compatibility in exported packages. Consider also running 
```bash
java/debugger.jpda.truffle$ ant build gen-sigtest
``` 
and checking the result is sane by:
```bash
java/debugger.jpda.truffle$ git diff nbproject/org-netbeans-modules-debugger-jpda-truffle.sig
```
no need to commit the `.sig` file, it is going to be updated during next 12.3 release.

Btw. while trying to do the above I realized there are references to non-existing sdk and truffle-api JARs:
```
java/debugger.jpda.truffle$ grep -r 1.0.0-rc6 .
./external/binaries-list:93A352D21C1D95061BDFE5C86A91F946F3A3E605 com.oracle.truffle:truffle-sl:1.0.0-rc6
./external/truffle-1.0.0-license.txt:Files: truffle-sl-1.0.0-rc6.jar
./nbproject/project.properties:truffle-backend.cp.extra=external/graal-sdk-1.0.0-rc6.jar:external/truffle-api-1.0.0-rc6.jar
./nbproject/project.properties:truffle.sl=external/antlr4-runtime-4.7.2.jar:external/truffle-sl-1.0.0-rc6.jar
```
I don't see any test. There is a Travis job testing various modules against GraalVM:
```bash
netbeans$ cat .travis.yml  | grep GRAAL
            - GRAAL_VERSION="20.2.0"
            - GRAAL_VERSION="19.3.3"
```
 we could certainly also check against 20.3 (or replace 20.2 with 20.3) to be sure the behavior of the agent is as expected.

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/TruffleStackInfo.java
entlicher: Fixed.

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/models/DebuggingTruffleActionsProvider.java
entlicher: :-)

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/source/Source.java
entlicher: Resolved.

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/frames/models/DebuggingTruffleTreeModel.java
entlicher: Well, this is generally understood as a friend api. The user needs to be knowledgeable about how the individual services work and any changes need to take that into account.
This produces UI at the end, we need to have some flexibility in changing it as needed.

entlicher:
I've removed the old unused sdk and truffle-api JARs.
But I've retained truffle-sl, it's needed by tests.
We should probably create a lib with truffle-sl, like `ide/libs.trufflesl` right? It might be handy for tests elsewhere.
For tests of this I'll need 20.3 truffle API and SL. I can add tests as soon as we upgrade the Truffle version.

JaroslavTulach:
> I've removed the old unused sdk and truffle-api JARs.

+1

> But I've retained truffle-sl, it's needed by tests.

+1

> We should probably create a lib with truffle-sl, like `ide/libs.trufflesl` right? It might be handy for tests elsewhere.

I am not sure there is a general need for `truffle-sl`. There is `graal-js` available as a library and very likely it is better to test against JavaScript rather than SL in majority of cases. I no longer remember why we selected `truffle-sl` rather than JavaScript, but that is OK, `truffle-sl` can stay "hidden" as a test library in this module. 

> For tests of this I'll need 20.3 truffle API and SL. I can add tests as soon as we upgrade the Truffle version.

Upgrading to 20.3 is desirable. Let's aim to have it for NetBeans 12.3.

JaroslavTulach (approved):

JaroslavTulach:
Good enough for NetBeans mileston `12.3` (e.g. `master` branch), probably also good candidate for `12.2-u1` - another PR with target `release122` has to be created to backport the changes.

PR #2645, opened 2021-01-07
[NETBEANS-4996] Refresh errors and hints in the LSP server after indexing.

jlahoda:


sdedic (approved):

JaroslavTulach (approved):
Can the implementation dependency be avoided in the future?

JaroslavTulach:
In clash with #2648, btw.

PR #2647, opened 2021-01-07
Make IntroduceElement fixes available as LSP CodeActions.

dbalek:
On proper selection, the following CodeActions should appear:
- Introduce Variable...
- Introduce Constant...
- Introduce Field...
- Introduce Method...


JaroslavTulach (approved):

Chris2011:
As I can see it on in the code, this is just for Java, right? Is there a possibility to have such functionality global? I mean we now have the implementation of LSP. The functionality should be somehow generic, right? Just guessing.

dbalek:
> As I can see it on in the code, this is just for Java, right? Is there a possibility to have such functionality global? I mean we now have the implementation of LSP. The functionality should be somehow generic, right? Just guessing.

Currently, language server for Java is provided by the NetBeans `java.lsp.server` module. This patch makes some Java specific hints (provided by the `java.hints` module) available to LSP clients.

PR #2649, opened 2021-01-08
LSP: Go to definition enhanced with possible sources download

dbalek:
Go to definition modified to download missing sources if necessary.


jlahoda (approved):
Seems reasonable.

PR #2653, opened 2021-01-08
Support executor ProgressHandle cancel over DAP protocol.

sdedic:
Followup to the previous ProgressHandle PRs. Changes the way `OperationContext` records created handles: they are recorded as soon as the `InternalHandle` instance is created (so it may not be started at the time someone calls `getHandles`).

The DAP server is then able to collect ProgressHandles created during the course of `invokeAction` call; hopefully the 1st will be the handle to the Executor.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/DebugAdapterContext.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/DebugAdapterContext.java
@@ -33,6 +33,8 @@ import org.eclipse.lsp4j.debug.services.IDebugProtocolClient;
 
 import org.netbeans.modules.java.lsp.server.debugging.breakpoints.BreakpointsManager;
 import org.netbeans.modules.java.lsp.server.debugging.launch.NbDebugSession;
+import org.netbeans.modules.java.lsp.server.progress.LspInternalHandle;
+import org.netbeans.modules.progress.spi.InternalHandle;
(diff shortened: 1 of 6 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/DebugAdapterContext.java
```diff
@@ -116,6 +119,19 @@ public int getDebuggerLine(int clientLine) {
     void setClientPathsAreUri(boolean clientPathsAreUri) {
         this.clientPathsAreUri = clientPathsAreUri;
     }
+    
+    public boolean requestProcessTermination() {
+        if (processExecutorHandle != null) {
```
JaroslavTulach, line 124: Only correct if `requestProcessTermination` and `setProcessExecutorHandle` are always called by the same thread.

JaroslavTulach (approved):

JaroslavTulach:
Setting milestone to `12.3`. Another PR shall be created to backport this to `12.2-u1` milestone.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/DebugAdapterContext.java
sdedic: I wouldn't count on that

dbalek (approved):

PR #2652, opened 2021-01-08
Project problems are resolved in headless environment.

sdedic:
This PR aims to solve issues with newly opened projects in headless environment  (required by LSP server). Project problems UI demand an interactive GUI, which can be partly handled by `DialogDisplayer` implementation for headless env. In an absence of better interaction with the user, all reported problems that can be fixed are attempted to be fixed. Results are reported using StatusDisplayer.

PR #2654, opened 2021-01-09
Let Mutex.EVENT methods co-operate with Lookups.executeWith(...)

JaroslavTulach:
While working on #2652 it turned out that `OpenProjectList` class is using `Mutex.EVENT` to deliver changes and that the call disassociates the event deliver from the originating request. One option is to modify `OpenProjectList` to deliver events synchronously (at least in `GraphicsEnvironment.isHeadless()` mode, the other is to make sure effective lookup set by `Lookups.executeWith` is properly propagated through `Mutext.EVENT` methods. That's what this PR proposes.


sdedic (approved):

tzezula (commented):
Looks good to me.

PR #2656, opened 2021-01-09
[NETBEANS-5195] Change icon for JSON files

mhalachev:
https://issues.apache.org/jira/browse/NETBEANS-5195

Change icon for JSON file type, as it is currently denoted by the same icon as JavaScript files.


matthiasblaesing:
Looks ok to me - three questions/requests:

- [ ] You understand, that you are donating this changeset to the ASF, which is then free to do with it as it sees fit (see the comment of the source files for the longer version). TL;DR: If the ASF ever decides to relicense the code it will treat this code as if it was the owner of it.
- [ ] Would it be possible to add an SVG version of the icon? There was some work done to convert older icons to svg, it would be great if new additions would directly be present in png and svg versions.
- [ ] As the PNG can't hold a license header, please add an entry to  https://github.com/apache/netbeans/blob/96bbee4b557e6ef99645bd7a5c0de6e2554f0f3d/webcommon/javascript2.editor/licenseinfo.xml (see https://github.com/apache/netbeans/blob/master/ide/git/licenseinfo.xml as reference). That way the license check can pick up the license information. If an SVG added, the license information should be in the XML, as it supports comments.

mhalachev:
Thank you for your comment and suggestions!
> Looks ok to me - three questions/requests:
> 
> * [ ]  You understand, that you are donating this changeset to the ASF, which is then free to do with it as it sees fit (see the comment of the source files for the longer version). TL;DR: If the ASF ever decides to relicense the code it will treat this code as if it was the owner of it.
* [x] I understand the aforementioned.
> * [ ]  Would it be possible to add an SVG version of the icon? There was some work done to convert older icons to svg, it would be great if new additions would directly be present in png and svg versions.
* [x] Added SVG version of the icon.
> * [ ]  As the PNG can't hold a license header, please add an entry to  https://github.com/apache/netbeans/blob/96bbee4b557e6ef99645bd7a5c0de6e2554f0f3d/webcommon/javascript2.editor/licenseinfo.xml (see https://github.com/apache/netbeans/blob/master/ide/git/licenseinfo.xml as reference). That way the license check can pick up the license information. If an SVG added, the license information should be in the XML, as it supports comments.
* [x] Updated license info for the PNG.

matthiasblaesing:
Looks good to me. Thank you. Could you please squash the commits into one?

mhalachev:
All right, it’s now as a single commit.

matthiasblaesing:
Thank you - lets get it in.

mhalachev:
Thanks a lot!

PR #2660, opened 2021-01-09
LSP codeAction fixes.

dbalek:
- [GR-28621] Introduce Field & Constant replaces all occurrences in all classes in single .java file
- [GR-28623]  IllegalStateException thrown from CreateOuterClassFix

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateClassFix.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateClassFix.java
@@ -320,9 +320,7 @@ public abstract class CreateClassFix extends CreateFixBase implements EnhancedFi
                         default: throw new IllegalStateException();
                     }
                     CompilationUnitTree cut = make.CompilationUnit(targetSourceRoot, packageName.replace('.', '/') + '/' + simpleName + ".java", Collections.<ImportTree>emptyList(), Collections.singletonList(source));
-                    ClassTree nue = createConstructor(working, new TreePath(new TreePath(cut), source));
                     working.rewrite(null, cut);
-                    working.rewrite(source, nue);
                 }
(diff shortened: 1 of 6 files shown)


java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateClassFix.java
```diff
@@ -320,9 +320,7 @@ public void run(final WorkingCopy working) throws IOException {
                         default: throw new IllegalStateException();
                     }
                     CompilationUnitTree cut = make.CompilationUnit(targetSourceRoot, packageName.replace('.', '/') + '/' + simpleName + ".java", Collections.<ImportTree>emptyList(), Collections.singletonList(source));
-                    ClassTree nue = createConstructor(working, new TreePath(new TreePath(cut), source));
                     working.rewrite(null, cut);
-                    working.rewrite(source, nue);
```
sdedic, line 325: Apologies, I don't quite follow this change - I thought it's desirable to generate a constructor (?). If the constructor is not wanted here, the `implement` method below (also generates one) may need to be modified as well.

JaroslavTulach (approved):

java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateClassFix.java
dbalek: The language server uses `getModificationResult()` to get changes that need to send back to LSP client as a response to codeAction request. These changes are applied at the client side. On the other hand, `implement` is used by ordinary NetBeans UI and uses createFromTemplate mechanism to create new outer class. But you are right - constructor should be generated even if called via LSP. Fixed.

sdedic (approved):

PR #2663, opened 2021-01-10
Readme changelog version

JaroslavTulach:
Martin Balín prepared updated documentation for the 12.2.1 update. This is a PR which merges that into master.


java/java.lsp.server/vscode/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Change Log
```
entlicher, line 1: There is no license. One build fails on it.

entlicher (changes requested):

entlicher (approved):

PR #2665, opened 2021-01-10
[NETBEANS-2843] PHP debugger - support for resolved breakpoints

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-2843

- Adds PHP debugger option to enable/disable support of resolved breakpoints in Xdebug.
- Adds new Xdebug feature 'resolved_breakpoints'.
- Requests use of resolved breakpoints during debug session initialization.
- Supported by XDebug 2.8 or newer.

### DIFF
--- a/php/php.dbgp/nbproject/project.xml
+++ b/php/php.dbgp/nbproject/project.xml
@@ -180,7 +180,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>2.87</specification-version>
+                        <specification-version>2.149</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpDebuggerPanel.form
+++ b/php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpDebuggerPanel.form
@@ -52,6 +52,20 @@
                   <Component id="requestedUrlsCheckBox" alignment="0" min="-2" max="-2" attributes="0"/>
                   <Component id="debuggerConsoleCheckBox" alignment="0" min="-2" max="-2" attributes="0"/>
                   <Component id="errorLabel" alignment="0" min="-2" max="-2" attributes="0"/>
+                  <Group type="102" alignment="0" attributes="0">
+                      <Group type="103" groupAlignment="0" attributes="0">
+                          <Component id="portLabel" alignment="0" min="-2" max="-2" attributes="0"/>
+                          <Component id="sessionIdLabel" alignment="0" min="-2" max="-2" attributes="0"/>
+                          <Component id="maxDataLengthLabel" alignment="0" min="-2" max="-2" attributes="0"/>
+                      </Group>
+                      <EmptySpace max="-2" attributes="0"/>
+                      <Group type="103" groupAlignment="0" max="-2" attributes="0">
+                          <Component id="portTextField" linkSize="2" min="-2" pref="50" max="-2" attributes="0"/>
+                          <Component id="sessionIdTextField" min="-2" pref="176" max="-2" attributes="0"/>
+                          <Component id="maxDataLengthTextField" linkSize="2" alignment="0" min="-2" pref="50" max="-2" attributes="0"/>
+                      </Group>
+                  </Group>
+                  <Component id="resolveBreakpointsCheckBox" min="-2" max="-2" attributes="0"/>
                   <Group type="102" attributes="0">
                       <EmptySpace min="21" pref="21" max="-2" attributes="0"/>
                       <Group type="103" groupAlignment="0" attributes="0">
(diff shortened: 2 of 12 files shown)


KacerCZ:
For basic test use this code:
```php
$array = [
    'a',
    'b',
    'c',
];

var_dump($array);
```
Set breakpoint at line `$array = [` and debug program.
Without breakpoint resolution program will not stop.
With breakpoint resolution program will stop at line `'a'`.

junichi11:
I had to uncheck the `Stop at First Line` option to verify it.
Please increment spec version of `php.dbgp` and fix the dependency of `php.project` (project.xml of `php.dbgp`).
cc: @tmysik

KacerCZ:
@junichi11 I updated spec and dependency.

When spec version needs to be changed?
Is it after change in public package (Project properties > API Versioning > Public Packages)?

Probably spec version of `php.dbgp` should not be changed.

junichi11:
@KacerCZ 

> When spec version needs to be changed?
> Is it after change in public package (Project properties > API Versioning > Public Packages)?
> 
> Probably spec version of `php.dbgp` should not be changed.

Increasing it for Auto update. However, I'm not sure whether UC for dev build is available currently. So, maybe, we should not do it to avoid conflicting with other PRs. Please feel free to revert it. Sorry for my wrong point out.

KacerCZ:
OK, I reverted change of spec for `php.dbgp`.

junichi11 (approved):

php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpDebuggerPanel.form
```diff
@@ -349,5 +352,19 @@
...
       </AccessibilityProperties>
     </Component>
+    <Component class="javax.swing.JCheckBox" name="resolveBreakpointsCheckBox">
+      <Properties>
+        <Property name="text" type="java.lang.String" editor="org.netbeans.modules.i18n.form.FormI18nStringEditor">
+          <ResourceString bundle="org/netbeans/modules/php/project/ui/options/Bundle.properties" key="PhpDebuggerPanel.resolveBreakpointsCheckBox.text" replaceFormat="org.openide.util.NbBundle.getMessage({sourceFileName}.class, &quot;{key}&quot;)"/>
+        </Property>
+      </Properties>
+    </Component>
+    <Component class="javax.swing.JLabel" name="resolveBreakpointsInfoLabel">
```
tmysik, line 362: OT: It is usually "better" (more readable) to avoid FQNs in the generated code.

tmysik: Ignore it :arrow_up: , it is XML file... :man_facepalming:

php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpOptions.java
```diff
@@ -52,6 +53,7 @@
     public static final boolean DEFAULT_DEBUGGER_SHOW_CONSOLE = false;
     public static final boolean DEFAULT_DEBUGGER_STOP_AT_FIRST_LINE = true;
     public static final boolean DEFAULT_DEBUGGER_WATCHES_AND_EVAL = false;
+    public static final boolean DEFAULT_DEBUGGER_RESOLVE_BREAKPOINTS = true;
```
tmysik, line 56: This is probably fine... maybe we could check what Xdebug version is present in the latest Ubuntu LTS?

php/php.dbgp/nbproject/project.xml
```diff
@@ -180,7 +180,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>2.87</specification-version>
+                        <specification-version>2.149</specification-version>
```
tmysik, line 183: Once this is done, typically also specification version of the module should be increased as well.

Why? If someone uses a daily build which is updated from DEV update center, he will get the new version of PHP project - that is fine. However, since PHP Debugger module has still the same specification version, no update is done. Therefore, functionality from this PR will not work. If the user downloads a fresh daily build, it will work. Therefore, the specification of the PHP Debugger module should be increased too so the user gets its updated version automatically too.

I am not sure where there is any DEV update center but in general, my comment applies.

tmysik (approved):

tmysik:
@KacerCZ @junichi11 

Just added a comment about increasing of specification version. No matter whether there is or is not a DEV update center _currently_, the specification version of PHP Debugger should be increased too (because that is simply correct :)

Thanks.

php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpOptions.java
KacerCZ: Ubuntu 20.04 is latest LTS and it contains Xdebug 2.9.2.
https://packages.ubuntu.com/focal/php-xdebug

php/php.dbgp/nbproject/project.xml
KacerCZ: Thanks for the explanation.
I increased specification version of `php.dbgp` module.

php/php.project/src/org/netbeans/modules/php/project/ui/options/PhpOptions.java
tmysik: Great, thanks for checking it.

php/php.dbgp/nbproject/project.xml
tmysik: You are welcome.

tmysik:
Once CI finishes, let's merge this PR. Thank you guys.

KacerCZ:

PR #2672, opened 2021-01-11
[NETBEANS-4693] new nb-javac with records in maven projects fix

lkishalmi:
I'm a bit unsure in this. Theoretically NB 12.0 comes with Java 14 Support, so it'd make sense.


geertjanw:
Yes, this would be good for the reason Laszlo mentions.

lkishalmi:
Something is still missing from this PR. If it should be in 12.0-u2 Then I need help!

matthiasblaesing:
If I remember nb-javac introduced regressions (Exception in normal operation, syntax requirements for Java 14 applied to older code), so I'm reluctant to cheer this in.

geertjanw:
I would revert my support for this one then in light of @matthiasblaesing's comment.

lkishalmi:
Well, I've tested this PR. Seems to be right, though it needs some work on the specification versions. Finally I've decided not to include it into the 12.0-u2 rather add for u3

Akshay-Gupta-Oracle (approved):
seems correct to me.

mbien:
are update releases for NB 12 still going to happen? If not we can probably close this.

PR #2673, opened 2021-01-11
[NETBEANS-5234] PHP debugger - fix display of long strings

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5234

Fixed processing of `property_value` command response.
Because response does not contain variable name, it has to be copied from command.
For details see [Xdebug documentation](https://xdebug.org/docs/dbgp#property-get-property-set-property-value) 

Fixed `Property.setName()` because code for adding attribute did not work.

### DIFF
--- a/php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
+++ b/php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
@@ -32,6 +32,7 @@ import org.netbeans.modules.php.dbgp.SessionId;
 import org.netbeans.modules.php.dbgp.SessionManager;
 import org.netbeans.modules.php.dbgp.UnsufficientValueException;
 import org.openide.util.Exceptions;
+import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 
 /**
@@ -39,6 +40,7 @@ import org.w3c.dom.Node;
  *
  */
 public class Property extends BaseMessageChildElement {
+
     static final String PROPERTY = "property"; // NOI18N
     private static final String NUMCHILDREN = "numchildren"; // NOI18N
     static final String ENCODING = "encoding"; // NOI18N
@@ -64,12 +66,11 @@ public class Property extends BaseMessageChildElement {
     }
 
     public void setName(String value) {
-        Node node = getNode().getAttributes().getNamedItem(NAME);
-        if (node == null) {
-            node = getNode().getOwnerDocument().createAttribute(NAME);
-            getNode().appendChild(node);
+        Node node = getNode();
+        if (node instanceof Element) {
+            Element element = (Element) node;
+            element.setAttribute(NAME, value);
         }
-        node.setNodeValue(value);
     }
 
     public String getFullName() {
--- a/php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/PropertyValueResponse.java
+++ b/php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/PropertyValueResponse.java
@@ -33,7 +33,7 @@ public class PropertyValueResponse extends DbgpResponse {
     }
 
     public Property getProperty() {
-        Node node = getChild(getNode(), Property.PROPERTY);
+        Node node = getNode();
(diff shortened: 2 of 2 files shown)


php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
```diff
@@ -64,12 +66,9 @@ public String getName() {
...
 
     public void setName(String value) {
-        Node node = getNode().getAttributes().getNamedItem(NAME);
-        if (node == null) {
-            node = getNode().getOwnerDocument().createAttribute(NAME);
-            getNode().appendChild(node);
-        }
-        node.setNodeValue(value);
+        assert getNode() instanceof Element;
+        Element element = (Element) getNode();
```
junichi11, line 70: Please check the type to avoid the ClassCastException.

KacerCZ: Check added.
Hopefully it was what you meant.

junichi11 (approved):
I've verified that a value is shown instead of "Evaluating...".

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
```diff
@@ -64,12 +66,10 @@ public String getName() {
     }
 
     public void setName(String value) {
-        Node node = getNode().getAttributes().getNamedItem(NAME);
-        if (node == null) {
-            node = getNode().getOwnerDocument().createAttribute(NAME);
-            getNode().appendChild(node);
+        if (getNode() instanceof Element) {
+            Element element = (Element) getNode();
```
tmysik, line 70: Nitpick: calling `getNode()` twice.

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
```diff
@@ -64,12 +66,10 @@ public String getName() {
     }
 
     public void setName(String value) {
-        Node node = getNode().getAttributes().getNamedItem(NAME);
-        if (node == null) {
-            node = getNode().getOwnerDocument().createAttribute(NAME);
-            getNode().appendChild(node);
+        if (getNode() instanceof Element) {
```
tmysik, line 69: What if node is not an `Element`? Can it happen? Is NOOP in such a case expected? I don't know it, just asking :)

tmysik (approved):

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
tmysik: One note to the `assert` - imagine that it would happen - what it will take tell you? Just that the node is not an `Element`... Not great, probably, and the assertion would need to be likely improved to something like:

```
assert getNode() instanceof Element : "expected Element but got " + getNode().getClass().getName();
```

tmysik:
@KacerCZ @junichi11 

Just a comment added, not sure how relevant it is... Feel free to merge, of course.

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
KacerCZ: If I `node` is not `Element` then also `getName()` will not work and it will return `null`.
So no effect in this case is OK.

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
KacerCZ: Fixed.

php/php.dbgp/src/org/netbeans/modules/php/dbgp/packets/Property.java
tmysik: OK, thanks for the explanation.

tmysik:
@junichi11 @KacerCZ 

Let's merge this PR once CI finishes. Thank you, guys!

PR #2675, opened 2021-01-13
Fix of tooltip and evaluations of Truffle languages through LSP.

entlicher:
Corrections of Truffle evaluations through LSP.


JaroslavTulach (approved):
A test would be nice.

PR #2678, opened 2021-01-15
[NETBEANS-5244] Add the project settings for XML configuration of phpDocumentor

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-5244

- Add settings for an XML configuration file
- Add parameters (`--config /path/to/configuration/file.xml`)

netbeans-5244-phpDocumentor-configuration-path

### DIFF
--- a/php/php.phpdoc/src/org/netbeans/modules/php/phpdoc/PhpDocScript.java
+++ b/php/php.phpdoc/src/org/netbeans/modules/php/phpdoc/PhpDocScript.java
@@ -23,6 +23,7 @@ import java.awt.EventQueue;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


junichi11:
@tmysik Sorry for the many review requests. Don't worry, I'll wait anyway :) Thanks.

KacerCZ (approved):
Tested, works.

php/php.phpdoc/src/org/netbeans/modules/php/phpdoc/PhpDocScript.java
```diff
@@ -145,13 +149,24 @@ private String sanitizePath(String path) {
...
     }
 
-    private List<String> getParameters(String sanitizedPhpDocTarget, PhpModule phpModule) {
+    private List<String> getAllParameters(String sanitizedPhpDocTarget, PhpModule phpModule) {
+        List<String> params = new ArrayList<>(getDefaultParameters(sanitizedPhpDocTarget, phpModule));
+        if (PhpDocPreferences.isConfigurationEnabled(phpModule)) {
+            String configurationPath = PhpDocPreferences.getPhpDocConfigurationPath(phpModule);
+            if (!StringUtils.isEmpty(configurationPath)) {
+                params.add(PARAM_CONFIG);
+                params.add(sanitizePath(PhpDocPreferences.getPhpDocConfigurationPath(phpModule)));
```
tmysik, line 158: If the path is incorrect, PhpDoc itself will complain, I guess, so we don't need to run the validator here. Or do we? :thinking: 

AFAIR we validate project properties before each run and if they are incorrect, the Project Properties dialog is opened.

@junichi11 @KacerCZ What do you think, guys?

tmysik (approved):

tmysik:
@junichi11 Just added one comment. Feel free to merge it as it is now, it is more a question. Thank you!

php/php.phpdoc/src/org/netbeans/modules/php/phpdoc/PhpDocScript.java
KacerCZ: Properties are opened only if target directory for generated documentation is not valid.
`sanitizePath()` only replaces backslashes on Windows with slashes.

tmysik: > Properties are opened only if target directory for generated documentation is not valid.

So, should not it be opened also if the enabled configuration file is invalid, typically it does not exist?

junichi11: I'll improve it.

tmysik: Thank you.

junichi11: Fixed.
netbeans-5244-phpdocumentor-invalid-settings

tmysik: Nicely done, thank you.

junichi11:
Thank you for your reviews.

PR #2679, opened 2021-01-15
Properly stop Maven execution from LSP/DAP

sdedic:
Maven executor forks the execution into a separate thread, and only then it creates a `ProgressHandle`. The handle is likely to be created **after** the _DAP_ call finishes, so it is not collected by the current code. I've implemented a event source  + listener interface to listen on create/finish of progress-generating tasks. _DAP_ operation can now establish a listener and receive a ProgressHandle even after the original launch request returns; it can connect the `ProgressHandle` to the `DebugContext`.

The default execution engine implementation should support default `Lookup` passing, so the threads it creates operate in the caller's context.

The tests of LSP server must run in headless mode; without that property, some code takes wrong path, blocking test execution on UI that nobody can click on.

### DIFF
--- a/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
+++ b/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
@@ -151,6 +151,7 @@ public class ServerTest extends NbTestCase {
 
     @Override
     protected void setUp() throws Exception {
+        System.setProperty("java.awt.headless", Boolean.TRUE.toString());
         super.setUp();
         clearWorkDir();
         ServerSocket srv = new ServerSocket(0, 1, InetAddress.getLoopbackAddress());
--- a/java/java.lsp.server/vscode/src/test/suite/extension.test.ts
+++ b/java/java.lsp.server/vscode/src/test/suite/extension.test.ts
@@ -1,6 +1,7 @@
 import * as assert from 'assert';
 import * as fs from 'fs';
 import * as path from 'path';
+import * as ps from 'ps-node';
 import { spawn, ChildProcessByStdio, spawnSync, SpawnSyncReturns } from 'child_process';
 import { Readable } from 'stream';
 
@@ -8,6 +9,7 @@ import { Readable } from 'stream';
 // as well as import your extension to test it
 import * as vscode from 'vscode';
 import * as myExtension from '../../extension';
+import { TextDocument, TextEditor, Uri } from 'vscode';
 
 suite('Extension Test Suite', () => {
     vscode.window.showInformationMessage('Cleaning up workspace.');
@@ -34,7 +36,7 @@ suite('Extension Test Suite', () => {
 
     test('Find clusters', async () => {
         let clusters = myExtension.findClusters('non-existent');
-        assert.strictEqual(clusters.length, 6, 'six clusters found: ' + clusters);
+        assert.strictEqual(clusters.length, 7, 'seven clusters found: ' + clusters);
         let nbcode = vscode.extensions.getExtension('asf.apache-netbeans-java');
         assert.ok(nbcode);
         for (let c of clusters) {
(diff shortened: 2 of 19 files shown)


JaroslavTulach:
> Maven executor forks the execution into a separate thread, and only then it creates a `ProgressHandle`. 

Wouldn't changing the Maven to create the progress sooner be simpler? The changes in `java.lsp.server` are quite complicated and beyond my ability to review their correctness( which [would rather be verified](https://github.com/apache/netbeans/blob/master/java/java.lsp.server/vscode/src/test/suite/extension.test.ts) somehow).

> The default execution engine implementation should support default `Lookup` passing, so the threads it creates operate in the caller's context.

+1, please add there some unit tests for the platform changes

> The tests of LSP server must run in headless mode; without that property, 
> some code takes wrong path, blocking test execution on UI that nobody can click on.

Wouldn't it be better to enforce the behavior with branding? E.g. have a property for various dialogs with values `Ask|Yes|No` which `java/java.lsp.server/nbcode` project would set to `No`? Value of `isHeadless()` may mean many things, it doesn't necessarily have to mean _running in VSCode extension_. Btw. I have a WIP replacing dialogs in `DataEditorSupport` with such branding.

sdedic:
Re: earlier creation of the handle. Possibly yes. But still I believe it is more correct (for the LSP) to allow collection of handles from an operation even if the actual process executes asynchronously / creates the handle after the initiator finishes.

Yes, tests should be added.

sdedic:
Re. branding: This should rather become a sort of public API of the platform, right ? Some brandable token used to decide whether to ask a question or take some default course. I would prefer some abstraction in platform / maven / javadoc support than direct "running in vscode extension" ... any suggestion how to specify the token meaning ? Or should there be really a separate token for each such (now)isHeadless occurrence ?

platform/openide.execution/src/org/openide/execution/ExecutionEngine.java
```diff
@@ -143,12 +151,14 @@ public InputOutput getInputOutput() {
             }
             
             public void run() {
-                try {
-                    super.run();
-                } catch (RuntimeException x) {
-                    x.printStackTrace();
-                    resultValue = 1;
-                }
+                Lookups.executeWith(originalLookup, () -> {
```
JaroslavTulach, line 154: Unit test this, please.

platform/core.execution/src/org/netbeans/core/execution/RunClassThread.java
```diff
@@ -65,14 +68,20 @@ public RunClassThread(TaskThreadGroup base,
...
         // #33789 - this thread must not be daemon otherwise it is immediately destroyed
         setDaemon(false);
         this.start();
     }
 
     /** runs the thread
     */
+    @Override
     public void run() {
+        Lookups.executeWith(originalLookup, this::doRun);
```
JaroslavTulach, line 81: Unit test this, please.

JaroslavTulach (approved):
> it is more correct (for the LSP) to allow collection of handles from an operation ...
...until operation's `CompletionStage` is completed. OK.

> Did mean that a functional test

Yes, I suggested (not required) an integration test that would launch an endless Maven process and then killed it.

> suggestion how to specify the token meaning?

Unless you want an API named `isRunningVSNetBeans()` then you need a separate API for each instance of `isHeadless()` that decided anything else then simple usage of `GraphicsEnvironment`.

sdedic:
@JaroslavTulach  I've added a functional test: it uses `ps-node` to check for the user Java process the Maven executor actually starts, then checks that the process has been terminated. Please re-review the JS test.

Tthe `extension.test.js` still needs some polishing, though: if there are cluster in other extensions (i.e. installed for the `code` globally), the test will break on cluster number check.

java/java.lsp.server/vscode/src/test/suite/extension.test.ts
```diff
@@ -34,7 +36,7 @@ suite('Extension Test Suite', () => {
 
     test('Find clusters', async () => {
         let clusters = myExtension.findClusters('non-existent');
-        assert.strictEqual(clusters.length, 6, 'six clusters found: ' + clusters);
+        assert.strictEqual(clusters.length, 7, 'seven clusters found: ' + clusters);
```
JaroslavTulach, line 39: This can certainly be changed by knowing the minimal list of clusters to find and ignoring others.

java/java.lsp.server/vscode/src/test/suite/extension.test.ts
```diff
@@ -97,6 +99,88 @@ class Main {
...
+                    }, (err, list ) => {
+                        let success : boolean = (list && list.length > 0) == running;
+                        if (success) {
+                            cb();
+                        } else {
+                            if (cnt == 0) {
+                                reject("Timeout waiting for user application");
+                                return;
+                            }
+                            setTimeout(() => waitUserApplication(cnt - 1, running, cb), 100);
```
JaroslavTulach, line 160: Expecting the application gets started in 500ms? We'll see how well that behaves on the CI servers.

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -151,6 +151,7 @@ public ServerTest(String name) {
 
     @Override
     protected void setUp() throws Exception {
+        System.setProperty("java.awt.headless", Boolean.TRUE.toString());
```
JaroslavTulach, line 154: I was hoping that, if we integrated #2685, this change would no longer be needed. But that is probably not true, as the branding would only be active in integration tests of `nbcode` suite.

One would have to turn the branding on manually in the test...

JaroslavTulach (approved):
Thanks Sváťo for all the additional improvements. I am really glad the set of people who know how to write VS Code integration tests has doubled its size!

PR #2682, opened 2021-01-16
 Adding folding range via LSP support, both client and server.

jlahoda:
Reorganizing LSP client background tasks, so that the tasks are run even if the server is restarted.


dbalek (approved):
Nice feature. Thanks.

PR #2685, opened 2021-01-18
Don't show reload/save dialogs in VSNetBeans

JaroslavTulach:
VS Code manages files itself. Rename, copy, delete is handled by their UI and only later notified to the NetBeans backend. This PR provides branding API, to disable showing dialogs to the user and provide a default answer. This approach builds on previous branding APIs introduced for example in #2427, #2386, etc.

### DIFF
--- a/java/java.lsp.server/nbcode/branding/modules/org-openide-loaders.jar/org/netbeans/modules/openide/loaders/Bundle.properties
+++ b/java/java.lsp.server/nbcode/branding/modules/org-openide-loaders.jar/org/netbeans/modules/openide/loaders/Bundle.properties
@@ -15,13 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# DataEditorSupport
-# Cases when modified source file was changed to [read-only]
-MSG_FileReadOnlyClosing=File {0} was changed to read-only! All changes will be discarded.
-MSG_FileReadOnlySaving=File {0} was changed to read-only! Can not save it.
-MSG_FileReadOnlyChanging=File {0} was changed to read-only!
-
 # yes/no/ask
-ASK_OnSaving=ask
+ASK_OnSaving=no
 # yes/no/ask
-ASK_OnClosing=ask
+ASK_OnClosing=no
--- a/platform/openide.loaders/arch.xml
+++ b/platform/openide.loaders/arch.xml
@@ -542,6 +542,21 @@ for more information about this.
         underlaying <a href="@TOP@/architecture-summary.html#script">scripting and templating
         engines</a>.
     </api>
+
+    <api name="ASK_OnSaving" group="branding" type="export" category="stable">
+        Control on save yes/no dialog in
+        <a href="@TOP@/org/openide/text/DataEditorSupport.html">DataEditorSupport</a>
+        by setting the <code>ASK_OnSaving</code> key in
+        <code>org/netbeans/modules/openide/loaders/Bundle.properties</code>
+        to <code>yes</code> or <code>no</code> in a branding file in your application.
+    </api>
+    <api name="ASK_OnClosing" group="branding" type="export" category="stable">
+        Control on close yes/no dialog in
+        <a href="@TOP@/org/openide/text/DataEditorSupport.html">DataEditorSupport</a>
+        by setting the <code>ASK_OnClosing</code> key in
+        <code>org/netbeans/modules/openide/loaders/Bundle.properties</code>
+        to <code>yes</code> or <code>no</code> in a branding file in your application.
+    </api>
 </answer>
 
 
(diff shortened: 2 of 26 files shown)


platform/openide.loaders/src/org/openide/text/DataEditorSupport.java
```diff
@@ -1353,19 +1331,8 @@ public void run() throws IOException {
                 try {
                     des.superSaveDoc();
                 } catch (UserQuestionException ex) {
-                    if (GraphicsEnvironment.isHeadless()) {
```
JaroslavTulach, line 1356: This check has been added by Martin Entlicher half a year ago, before it become possible to show `NotifyDescriptor` dialogs in VSCode. Now the dialog could be shown (but we don't want it anyway). As such I am removing the headless check and replacing it with a branding API that allows `java/java.lsp.server/nbcode` suite to explicitly say 'no'.

sdedic: Note that this code was released with 12.2; IMHO in headless environment the code should default to 'yes' setting to remain compatible. Otherwise declare an incompatible change.

java/java.lsp.server/nbcode/branding/modules/org-openide-loaders.jar/org/netbeans/modules/openide/loaders/Bundle.properties
```diff
@@ -0,0 +1,21 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# yes/no/ask
+ASK_OnSaving=no
+# yes/no/ask
+ASK_OnClosing=no
```
sdedic, line 21: Q: shouldn't be there some dependency on loaders >= 1.80 ?

platform/openide.loaders/arch.xml
```diff
@@ -542,6 +542,21 @@ for more information about this.
         underlaying <a href="@TOP@/architecture-summary.html#script">scripting and templating
         engines</a>.
     </api>
+
+    <api name="ASK_OnSaving" group="branding" type="export" category="stable">
```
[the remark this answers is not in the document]
JaroslavTulach: Looking at [list of branding APIs](https://bits.netbeans.org/dev/javadoc/branding.html), it seems there used to be some convention. For example the `org.netbeans.core.netigso.FRAMEWORK_START_LEVEL` key & similar seem to spell out the package name of the bundle first and then some further info.

I can follow that convention in this PR too. Fixing others may be better for some another PR.

platform/openide.loaders/src/org/openide/text/DataEditorSupport.java
JaroslavTulach: I'm opting for declaring the change as an incompatible change in this case. The previous change was neither documented or tested and it is there just a three months...

java/java.lsp.server/nbcode/branding/modules/org-openide-loaders.jar/org/netbeans/modules/openide/loaders/Bundle.properties
JaroslavTulach: Right. I did check `java/java.lsp.server/nbcode/integration/nbproject/project.xml` and there is no dependency on `org.openide.loaders`. I can add one.

JaroslavTulach: Right. I did check `java/java.lsp.server/nbcode/integration/nbproject/project.xml` and there is no dependency on `org.openide.loaders`. I can add one.

platform/openide.loaders/src/org/openide/text/DataEditorSupport.java
JaroslavTulach: Done in 4f42da8398.

java/java.lsp.server/nbcode/branding/modules/org-openide-loaders.jar/org/netbeans/modules/openide/loaders/Bundle.properties
JaroslavTulach: Fixed in cc1eb58196680

platform/openide.loaders/arch.xml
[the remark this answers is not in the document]
JaroslavTulach: Fixed globally in 1ee8dfee3601

PR #2686, opened 2021-01-18
Fix broken links to images in the extension's readme.

dbalek:
Adding `--baseImagesUrl` option to the extension's build script.

### DIFF
--- a/java/java.lsp.server/build.xml
+++ b/java/java.lsp.server/build.xml
@@ -60,6 +60,13 @@
         <ant dir="nbcode" target="clean" inheritall="false" inheritrefs="false"/>
     </target>
     <target name="build-vscode-ext" depends="build-lsp-server" description="Build the Visual Studio Code extension.">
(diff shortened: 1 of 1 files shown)


java/java.lsp.server/build.xml
```diff
@@ -96,6 +96,8 @@
         </exec>
         <exec executable="${build.dir}/vsce/node_modules/vsce/out/vsce" failonerror="true" dir="${basedir}/vscode">
             <arg value="package" />
+            <arg value="--baseImagesUrl" />
+            <arg value="https://github.com/apache/netbeans/raw/master/java/java.lsp.server/vscode" />
```
JaroslavTulach, line 100: Using `master` means that such images can never be removed. Wouldn't it be more proper to obtain the actual commit ID and use it in the URL?

JaroslavTulach (changes requested):

java/java.lsp.server/build.xml
dbalek: Fixed as requested.

JaroslavTulach (approved):

PR #2688, opened 2021-01-18
[NETBEANS-5258]-Fix integration with scene builder 15.0.0.0

moacirrf:
JavaFX Scene Builder 15.0.0 Integration Linux
https://issues.apache.org/jira/browse/NETBEANS-5258

### DIFF
--- a/javafx/javafx2.scenebuilder/src/org/netbeans/modules/javafx2/scenebuilder/impl/Bundle.properties
+++ b/javafx/javafx2.scenebuilder/src/org/netbeans/modules/javafx2/scenebuilder/impl/Bundle.properties
@@ -32,5 +32,5 @@ MAC_GLUON_HOME=/Applications/SceneBuilder.app
 
 #UN*X
 UX_WKIP=/opt/JavaFXSceneBuilder$ver$;~/JavaFXSceneBuilder$ver$;/opt/SceneBuilder
-UX_LAUNCHER=JavaFXSceneBuilder$ver$;SceneBuilder
-UX_PROPERTIES=app/scenebuilder.properties;app/SceneBuilder.cfg
+UX_LAUNCHER=JavaFXSceneBuilder$ver$;SceneBuilder;bin/SceneBuilder
+UX_PROPERTIES=app/scenebuilder.properties;app/SceneBuilder.cfg;lib/app/SceneBuilder.cfg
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
Thank you. Change looks sane, three points:

- [ ] are you aware, that when this change set is integrated it is assumed, that it is a donation to the ASF? (See license header to get a clearer picture)
- [ ] Is my interpretation correct, that the properties define search paths inside the target folder and you solve the issue by adding additional directories, that might hold the binaries.
- [ ] This fixes only unix systems - would you be willing to have a look at the other os packages?

moacirrf:
> Thank you. Change looks sane, three points:
> 
>     * [x ]  are you aware, that when this change set is integrated it is assumed, that it is a donation to the ASF? (See license header to get a clearer picture)
> 
>     * [ x]  Is my interpretation correct, that the properties define search paths inside the target folder and you solve the issue by adding additional directories, that might hold the binaries.
> 
>     * [ x]  This fixes only unix systems - would you be willing to have a look at the other os packages?
Hello,
You are right, are paths of the folders, separated by ";".
I tested with windows, and no problem, Netbeans recognize scene builder folder,  when i try open an fxml file, scene builder opens as expected.

But with mac os i think its broken, but i can't test on a real mac.
I extracted dmg file of scenebuilder with p7zip and executed only the code responsible of that part, and didn't recognize the extracted folder.
I change the line above
`MAC_PROPERTIES=Contents/Java/scenebuilder.properties;Contents/Resources/Java/scenebuilder.properties`
by
`MAC_PROPERTIES=Contents/Java/scenebuilder.properties;Contents/Resources/Java/scenebuilder.properties;Contents/Java/SceneBuilder.cfg;Contents/app/SceneBuilder.cfg`
Build Netbeans again and the folder was recognized, but  i cant test on a real mac.

**For Scene Builder 8.5:** Contents/Java/SceneBuilder.cfg: 
**For Scene Builder 15.0.0:** Contents/app/SceneBuilder.cfg:
 
Do you think is a good a idea include this patch for MacOS in this Pull Request too?

Maybe this patch fix this issue:
[https://issues.apache.org/jira/browse/NETBEANS-1823](https://issues.apache.org/jira/browse/NETBEANS-1823)

matthiasblaesing:
Yes I think this is a good idea. In the worst case 2 further paths are checked. This sounds sane to me.

errael:
@neilcsmith-net @geertjanw Seems like a 12.3 candidate; needs reviewers

matthiasblaesing:
@moacirrf I verified that with this change I could integrate SceneBuilder 15 into NetBeans (select installation and open the fxml) on Linux and macOS, I verified, that the original problem is reproducible on macOS with a build from master.

If I can persuade travis to accept this, I'll squash and merge.

Thank you.

matthiasblaesing:
Integrated via https://github.com/apache/netbeans/commit/ed02fd2a9c250021f62269e4374df60f9f50a773

moacirrf:
good!
Thanks @matthiasblaesing

neilcsmith-net:
Thanks for merging @matthiasblaesing but if you squash and push to the upstream PR branch it can be merged and register as such, which is also useful.

PR #2689, opened 2021-01-18
[NETBEANS-2260] Welcome Page header graphics

mhalachev:
https://issues.apache.org/jira/browse/NETBEANS-2260

- Update content banner image asset
- Set bicubic interpolation before drawing


matthiasblaesing:
Thank you, this indeed improves the look and even saves space :-)

I assume, that you read and understood the license information at the header of the edited source file. You are donating the changeset to the ASF. If you don't agree to this, please indicate so, the merge will then be reverted.

mhalachev:
Thank you for your review!

Actually, some space is being gained as well :-)

PS. Yes, it's all right regarding the license information.

PR #2691, opened 2021-01-18
[NETBEANS-5261] Shared settings for PHP code generators

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5261

Changes settings of PHP code generators to shared in project.
Style of generated getters/setters/constructors is part of coding standard so it should be part of shared settings.

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/codegen/CGSGenerator.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/codegen/CGSGenerator.java
@@ -418,7 +418,8 @@ public final class CGSGenerator implements CodeGenerator {
         Preferences preferences = null;
         Project project = FileOwnerQuery.getOwner(fo);
         if (project != null) {
-            preferences = ProjectUtils.getPreferences(project, CGSGenerator.class, false);
+            // Share settings because style of generated code is part of project coding standard.
+            preferences = ProjectUtils.getPreferences(project, CGSGenerator.class, true);
             try {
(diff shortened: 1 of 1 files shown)


php/php.editor/src/org/netbeans/modules/php/editor/codegen/CGSGenerator.java
```diff
@@ -418,7 +418,7 @@ public void invoke() {
         Preferences preferences = null;
         Project project = FileOwnerQuery.getOwner(fo);
         if (project != null) {
-            preferences = ProjectUtils.getPreferences(project, CGSGenerator.class, false);
+            preferences = ProjectUtils.getPreferences(project, CGSGenerator.class, true);
```
junichi11, line 421: I would write the reason as a comment.

tmysik: Trying to remember why it is done so... no idea.

tmysik (approved):

php/php.editor/src/org/netbeans/modules/php/editor/codegen/CGSGenerator.java
KacerCZ: Comment added.

junichi11: Thanks.

junichi11 (approved):

PR #2693, opened 2021-01-19
Update FlatLaf from 0.31 to 1.0-rc1

DevCharly:
This PR updates FlatLaf from 0.31 (from April 2020) to latest version.
A lot of improvements and fixes have been made in the last 9 months...
https://github.com/JFormDesigner/FlatLaf/releases

Some notable/visible changes:

Complete rewrite of menu item layout with stable left margins, right aligned accelerators (as in Windows and macOS) and larger gap between text and accelerator.

image

The text and accelerator are no longer layouted in "columns", which makes menu width smaller.

image

All popups (menu, tooltip, combobox, etc) now have drop shadows, which makes them better recognizable:

image

image

File chooser now show system icons:

image

Scrollbars use round thumb on macOS and Linux to make it look similar to native platform scroll bars:

image

Also changed FlatLaf properties files recognition from service to explicit registration in module installer
(this fixes [NETBEANS-4723](https://issues.apache.org/jira/browse/NETBEANS-4723)).


neilcsmith-net:
Seems to be sigtest failures on Travis, due to changing APIs between 0.31 and 1.0 presumably, but need to handle in wrapper somehow?

lkishalmi:
There are two options. We either bump up the release version of the library module, or disable sigtest for the library wrapper.

neilcsmith-net:
Thanks @lkishalmi Yes, former was in my head but wasn't sure if that seemed the right option?  Haven't looked at config for sigtest - where to look?  Bumping release version is probably the better idea?  Am I right in reading the default isn't 1, so we could use that? eg. http://wiki.netbeans.org/VersioningPolicy

@DevCharly can you look at adding the release version into the wrapper?  Assuming test then passes we can merge for 12.3.

DevCharly:
> @DevCharly can you look at adding the release version into the wrapper? Assuming test then passes we can merge for 12.3.

Sure.
Is it the version in `OpenIDE-Module-Specification-Version` in `manifest.mf`?
https://github.com/apache/netbeans/blob/2de44ca42571c3c719f9183c176522c2fc4fbe8b/platform/libs.flatlaf/manifest.mf#L4

Seems this version was already increased from 1.3 to 1.4 in commit 1104ec1d48321cc409db5c9c74d61d109d69edb6

Even if I increase this version to 1.5, the error remains (in a local build).

Do I have to generate the `org-netbeans-libs-flatlaf.sig` file somehow and check it in?

jtulach:
I guess the flat guys should start using Sigtest themselves... unless they want to fix their incompatibilities in RC2, we have no other option than to update the `.sigtest` file, right? Also increment the `releaseversion` of the module as this is clearly binary incompatible and there could be linkage problems.

neilcsmith-net:
@DevCharly check the old wiki page I linked to - but think you'll want to change `OpenIDE-Module: org.netbeans.libs.flatlaf` to `OpenIDE-Module: org.netbeans.libs.flatlaf/1`

See for reference https://github.com/apache/netbeans/blob/master/platform/libs.jna/manifest.mf#L3 although I think 1 is right?

Noticed also JNA has https://github.com/apache/netbeans/blob/master/platform/libs.jna/nbproject/project.properties#L32

Maybe @jtulach can advise here.  Also, this *is* the flat guys! :wink: And it's not that unusual for incompatible changes on major version changes - something normally handled by not exposing third-party APIs in the past?

DevCharly:
Ok, changed the major release version of the FlatLaf library to `/1` and updated the module dependency in the "FlatLaf look and feel" module. Hope it is correct now.

JaroslavTulach:
The information on release version behavior is in [module system documentation](https://bits.netbeans.org/12.0/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html) (which isn't easy to find) and also in Javadoc of [ModuleInfo](https://bits.netbeans.org/12.0/javadoc/org-openide-modules/org/openide/modules/ModuleInfo.html#getCodeNameRelease--) which mentiones that if the release version isn't defined, it defaults to `-1`...

> See for reference https://github.com/apache/netbeans/blob/master/platform/libs.jna/manifest.mf#L3 although I think /1 is right here?

As such changing to `/1` properly changes the release version from `/-1` to new value.

> Also, this _is_ the flat guys! 

I see. Consider setting [sigtest](https://github.com/jtulach/netbeans-apitest) up.

> And it's not that unusual for incompatible changes on major version changes - 
> something normally handled by not exposing third-party APIs in the past?

`platform/libs.flatlaf/nbproject/project.xml` declares public packages because `platform/o.n.swing.laf.flatlaf/nbproject/project.xml` depends on it. As NetBeans project tries to prevent usage of friend packages, the API is exposed to anyone. That is good for reuse, but also a problem with 3rd party dependencies that don't stick to backward compatibility that much.

platform/o.n.swing.laf.flatlaf/nbproject/project.xml
```diff
@@ -54,7 +54,8 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.0</specification-version>
+                        <release-version>1</release-version>
```
JaroslavTulach, line 57: According to [module documentation](https://bits.netbeans.org/12.0/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html#how-vers) this could be a range if you know this module works with previous as well as new incompatible versions.

JaroslavTulach (approved):

neilcsmith-net:
@DevCharly looks like we're there, but ideally need to squash the 3 commits into 1 on the PR. If you're able to do that on your branch that would be great. Otherwise I'll look into it before branching. Thanks.

And thanks @jtulach  Also I didn't realise sigtest was available as a Maven plugin either - that's great!

DevCharly:
@neilcsmith-net have squashed and rebased the 3 commits

@jtulach many thanks for the help.
Will have a look at sigtest.

The API changes in FlatLaf were done in 0.x development phase, which should be acceptable.
Will avoid such changes in the future 😉

PR #2694, opened 2021-01-20
Allow "Attach sources/javadoc" UIs to be branded out.

sdedic:
This PR is somewhat inspired by #2685: LSP tests now hang on displaying 'attach sources' for a Java platform, which is traditionally handled by `GraphicsEnvironment.isHeadless()`. But @JaroslavTulach approach shows a way how to do that more politely and allow the application to select which parts of the interaction are applicable in the final environment.

I have slightly modified the support made by @dbalek for 12.3; the `SourceJavadocAttacher.attachSources()` call could (IMHO) fail in a headless environment, since it got no `plugin` from the library + platform Attachers. @dbalek please review the behaviour: the other attachers now do nothing in 'yes' mode - they could possibly use the `Definer` SPI as well.

The actual bundle keys may still need to be modified based on discussion in #2685. After this PR merges, the headless property in #2679 can be replaced by branding out the dialogs.

### DIFF
--- a/java/java.j2seplatform/arch.xml
+++ b/java/java.j2seplatform/arch.xml
@@ -340,9 +340,54 @@
     </api>
    </li>
   </ul>
+  
+  <api name="API_Ask_attachSourcesQuestion" group="branding" type="export" category="stable">
+      <p>
+        Controls interaction when <b>sources</b> are about to be attached to a Classpath root, in
+        <a href="@org-netbeans-api-java@/org/netbeans/spi/java/queries/SourceJavadocAttacherImplementation.html">SourceJavadocAttacherImplementation</a>
+        provided by this module. By setting <code>API_Ask_attachSourcesQuestion</code> key in 
+        <code>org/netbeans/modules/java/j2seplatform/api/Bundle.properties</code> one can control the function:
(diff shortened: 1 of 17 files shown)


java/java.j2seplatform/arch.xml
```diff
@@ -340,10 +340,55 @@
     </api>
    </li>
   </ul>
+  
+  <api name="API_Ask_attachSourcesQuestion" group="branding" type="export" category="stable">
```
JaroslavTulach, line 344: Let's agree on following naming policy: `pck.name.KEY_NAME`. I'll update #2685 too.

JaroslavTulach (approved):

java/java.j2seplatform/arch.xml
sdedic: BTW It's a question if the `<api name=""` attribute should not contain the Bundle resource path as well, since the `path/to/Bundle:resourceKey` is the actual API identifier.

dbalek (approved):
Seems to me that banding for Language Server is missing.

JaroslavTulach (approved):

PR #2695, opened 2021-01-20
[NETBEANS-5269] Tooltip of columns in result-table

ses1112:
If you execute a query on the db, a result-table appears. There you can hover about the column-name und you can see some additional information about the structure of the column. If the column is of the type string, it also shows the attribute "length" with the precision of the column. This doesn't make sense, because by string there is no precision.
It occured with MariaDB and with a column of the type char(36).


matthiasblaesing:
I reread the spec and no, NetBeans behaves correctly. See here (https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSetMetaData.html#getPrecision-int-):

> Get the designated column's specified column size. For numeric data, this is the maximum precision. **For character data, this is the length in characters**. For datetime datatypes, this is the length in characters of the String representation (assuming the maximum allowed precision of the fractional seconds component). For binary data, this is the length in bytes. For the ROWID datatype, this is the length in bytes. 0 is returned for data types where the column size is not applicable.

So netbeans does the right thing - sorry, that I did not find that on the first pass, but on second read this is an issue in the DB driver.

ses1112:
It was an error in the MariaDB-Driver:
https://jira.mariadb.org/browse/CONJ-850

matthiasblaesing:
Thank you for pursuing this and getting it fixed :thumbsup:

PR #2696, opened 2021-01-20
[NETBEANS-5270] Multiple CustomIndexer same Mime-Type: priority ignored

ses1112:
If you have multiple CustomIndexer with the same Mime-Type and you override the method "getPriority" in the corresponding IndexerFactory, this priority is ignored/never used.
So there is the bug, that sometimes it can be that this CustomIndexers are executed in the wrong order.

### DIFF
--- a/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexerCache.java
+++ b/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexerCache.java
@@ -464,7 +464,7 @@ public abstract class IndexerCache <T extends SourceIndexerFactory> {
 
                 // the comparator instance must not be cached, because it uses data
                 // from the default lookup
-                Collections.sort(_orderedInfos, new C());
+                Collections.sort(_orderedInfos, new C().thenComparing(IIC));
                 sortInfosByMimeType(_infosByMimeType);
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
I reread it and I fail to see how the change fixes the issue. The per mimetype indexers are sorted:

https://github.com/apache/netbeans/blob/1a6621378cbf31bec4fcf0a79ce33a4a06430c89/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexerCache.java#L468

So why does your change fixes the original problem?

ses1112:
Yes, _infosByMimeType are sorted by mimetype. 
And _orderedInfos are also sorted by mimetype. But if two Indexer has the same mimetype, they should be sorted by the priority of the IndexerFactory.

https://github.com/apache/netbeans/blob/1a6621378cbf31bec4fcf0a79ce33a4a06430c89/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexerCache.java#L790

matthiasblaesing:
It would help if you could show an example and explain how the problem manifests. I can guess or debug myself through the netbeans codebase, but I won't do that.

matthiasblaesing:
No reply after nearly 12 months. Closing.

PR #2697, opened 2021-01-21
Completing the CPPLite debugger: threads, variables, watches, tooltips, etc...

entlicher:
Adapted to work with multiple threads, all necessary views are provided, functionality improved.


cpplite/cpplite.debugger/test/unit/src/org/netbeans/modules/cpplite/debugger/StepTest.java
```diff
@@ -75,7 +75,7 @@ public void testStepOver () throws Exception {
         LineCookie lc = DataObject.find(source).getLookup().lookup(LineCookie.class);
         assertNotNull(lc);
         DebuggerManager.getDebuggerManager().addBreakpoint(new CPPLiteBreakpoint(lc.getLineSet().getCurrent(4)));
-        CPPLiteDebugger d = CPPLiteDebugger.startDebugging(new CPPLiteDebuggerConfig(Arrays.asList(new File(wd, "main").getAbsolutePath()))).first();
+        CPPLiteDebugger d = CPPLiteDebugger.startDebugging(new CPPLiteDebuggerConfig(Arrays.asList(new File(wd, "main").getAbsolutePath()), wd)).first();
```
JaroslavTulach, line 78: I guess this test could be expanded to cover bits of the new functionality.

JaroslavTulach (approved):
Ok from an architecture perspective. I don't see any changes in the dependencies, any new API. Just an implementation. Good. I guess the implementation shall be tested a bit more deeply. There already is a `StepTest`, it might be good to expand it a bit. Anyway that can wait.

Thanks for improving NetBeans native support!

cpplite/cpplite.debugger/test/unit/src/org/netbeans/modules/cpplite/debugger/StepTest.java
entlicher: Right. I plan to improve the tests, hopefully soon after the 12.3 feature-freeze.

neilcsmith-net:
@entlicher this just reminded while you're looking at CPPLite debugger for 12.3, is it at all possible to change the external dependency name to remove `SNAPSHOT`? It stops this cluster being included in the Maven publishing.

entlicher:
@neilcsmith-net O.K. I'll do that...

entlicher:
@neilcsmith-net Well, the file in question is in binaries: `cpplite-mi-1.0-SNAPSHOT.jar`. It's not anything from Maven. However, I didn't find that file at http://netbeans.osuosl.org/binaries/ I'd rather solve this in a separate PR given the approaching feature freeze.

PR #2698, opened 2021-01-21
Restart NBLS when an extension providing some NB cluster is installed/uninstalled.

dbalek:
Language Server should be restarted whenever an extension providing its part (e.g. nbjavac) is installed/uninstalled.

### DIFF
--- a/java/java.lsp.server/vscode/src/extension.ts
+++ b/java/java.lsp.server/vscode/src/extension.ts
@@ -45,6 +45,7 @@ let client: Promise<LanguageClient>;
 let nbProcess : ChildProcess | null = null;
 let debugPort: number = -1;
 let consoleLog: boolean = !!process.env['ENABLE_CONSOLE_LOG'];
+let currentClusters: string[] = [];
 
 function handleLog(log: vscode.OutputChannel, msg: string): void {
     log.appendLine(msg);
@@ -66,7 +67,7 @@ export function enableConsoleLog() {
 }
 
 export function findClusters(myPath : string): string[] {
-    let clusters = [];
+    currentClusters = [];
     for (let e of vscode.extensions.all) {
         if (e.extensionPath === myPath) {
             continue;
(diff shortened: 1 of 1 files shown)


java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -66,7 +67,7 @@ export function enableConsoleLog() {
 }
 
 export function findClusters(myPath : string): string[] {
```
JaroslavTulach, line 69: This is an exported function, I'd prefer to keep it pure (e.g. without side effects).

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -156,6 +157,13 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
 
     // find acceptable JDK and launch the Java part
     findJDK((specifiedJDK) => {
+        context.subscriptions.push(vscode.extensions.onDidChange(() => {
+            const currentClustersSorted = currentClusters.slice().sort();
```
JaroslavTulach, line 161: keep the global state somewhere here, if possible.

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -156,6 +157,13 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
+            const newClusters = findClusters(context.extensionPath);
+            if (newClusters.length !== currentClustersSorted.length || newClusters.sort().find((value, index) => value !== currentClustersSorted[index])) {
```
JaroslavTulach, line 163: Wow. That's a functional programming!

JaroslavTulach (approved):
Please keep the `findClusters` function without side effects. I'd be nice to expand the [findClusters test](https://github.com/apache/netbeans/blob/983cae8922d52e11657661fab95a5420349df10a/java/java.lsp.server/vscode/src/test/suite/extension.test.ts#L35). Otherwise OK.

java/java.lsp.server/vscode/src/extension.ts
dbalek: Fixed.

java/java.lsp.server/vscode/src/extension.ts
dbalek: Fixed.

PR #2701, opened 2021-01-21
[NETBEANS-5272] Fixed grouping of documents in document switcher

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5272

Fixes issue when files from different projects with same name are not grouped correctly.
Adds sorting by project path to ensure files from same project are together.

Sample project reproducing the issue is available in JIRA.

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
@@ -237,6 +237,11 @@ class DocumentSwitcherTable extends SwitcherTable {
                     int res = project.getDisplayName().compareTo( otherProject.getDisplayName() );
                     if( res != 0 )
                         return res;
+
+                    res = project.getPath().compareTo(otherProject.getPath());
+                    if (res != 0) {
+                        return res;
+                    }
                 }
             }
             return super.compareTo( o );


platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -225,21 +225,27 @@ public TabData getTabData() {
         }
 
         @Override
-        public int compareTo( Object o ) {
-            if( o instanceof Item ) {
-                ProjectProxy otherProject = ((Item)o).project;
-                if( null == project && null != otherProject )
+        public int compareTo(Object o) {
```
DevCharly, line 228: I think you should avoid all whitespace changes.
They make the review (here on GitHub) harder because it is difficult to find the real changes.

DevCharly (changes requested):
Have tested the PR with two equal projects (in different folders) and it fixes the order:

Before:

image

After:

image

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
KacerCZ: I reverted formatting of method.
Thanks for review.

DevCharly (approved):

KacerCZ:
@neilcsmith-net @geertjanw Can this fix be included in 12.3, please?

PR #2702, opened 2021-01-21
Update .gitattributes - support for shell scripts

lbruun:
Added support for Windows shell scripts (`.bat` and `.cmd`) 
Added support for Linux/MacOS shell scripts (`.sh`)


matthiasblaesing:
Looks sane to me.

PR #2705, opened 2021-01-22
LSP code lens for running tests and main methods.

entlicher:
In cooperation with @jlahoda we propose pre-computation of test methods and code lens annotations in VSCode allowing to easily Run/Debug test methods and main methods.


jlahoda (approved):
We (I) may need a cleanup pass later, but looks OK to me.

PR #2708, opened 2021-01-24
Migrate build from Travis CI to GitHub Actions.

KacerCZ:
First attempt to migrate build to GitHub Actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,21 +15,24 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Apache Netbeans Continuous Integration
+name: Apache NetBeans Continuous Integration
 
-on:
-  push:
-  pull_request:
+on: workflow_dispatch
 
 jobs:
-  windows:
-    name: Windows build
-    runs-on: windows-2019
+  compile-all-modules-openjdk11:
+    name: Compile all modules with OpenJDK11
+    runs-on: ubuntu-18.04
     env:
-      ANT_OPTS: -Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json
+      OPTS: -Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json -quiet -Dcluster.config=full -Dpermit.jdk9.builds=true -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
+      - name: Setup OpenJDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
 
       - name: Caching dependencies
         uses: actions/cache@v2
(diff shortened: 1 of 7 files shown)


geertjanw:
Great effort, what needs to still be done to complete this work?

KacerCZ:
I migrated only a few steps just to check how to use different JDK version.
If it will be successful I'll transfer other jobs.

lbruun:
Hi @KacerCZ. Great work!

Yours tests for building Java8 vs Java11 seem similar to me. Perhaps you can benefit from using [Matrix Testing](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix)

Perhaps the PR also need to be reviewed in the light of the ASF's resource limitation wrt GitHub Action Runners ? For example some tests may need to only be carried out on a cron schedule and not for every commit?  Just a thought. 

Apart from that the PR clearly shows signs of optimization. Excellent.

.github/workflows/main.yml
```diff
@@ -309,3 +309,806 @@ jobs:
...
+  compile-all-modules-openjdk8:
+    name: Compile all modules with OpenJDK8
+    runs-on: ubuntu-18.04
+    env:
+      OPTS: -Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json -quiet -Dcluster.config=full -Djavac.compilerargs=-nowarn -Dbuild.compiler.deprecation=false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Caching dependencies
```
lbruun, line 430: Add "setup-java" step for Java8 rather than relying on Java8 to be the default for the Runner. (making it explicit rather than implied)

KacerCZ:
@lbruun Yes, some tests are similar and I plan to use matrix.
My goal is to do conversion of jobs first.

Next step will be improvement of workflow
- use matrix when all parameters are same and there is just different environment, 
- add dependencies between jobs
- join jobs with same settings under to spare runner.

neilcsmith-net:
Thanks @KacerCZ and great work!

> Perhaps the PR also need to be reviewed in the light of the ASF's resource limitation wrt GitHub Action Runners ?

Yes, this is a major concern at the moment - this email on users @ infra (private, unfortunately) worth a read - https://lists.apache.org/thread.html/r748a0e034256f22c240d6a00cece0e005724a3173cfbb478b548d979%40%3Cusers.infra.apache.org%3E

The short version is that ASF is already close to maxing out available runners at times, and NetBeans is already one of the top users (although not really that high in comparison to some).

So, I'm saying -1 to merging this (not the implementation!) until conversations on dev@ and with infra and we can be assured we can do this without negative impact.

> For example some tests may need to only be carried out on a cron schedule and not for every commit

An annoyance I have with our GitHub Actions as opposed to the Travis CI config is that at the moment a failure in GitHub Actions needs all tasks to be re-run, whereas Travis allows for retriggering the failing test only.  (At least I've never found a way to retrigger part of a GH workflow?).  So, the tests on Travis that have a tendency to fail at random would definitely be good ones to move off PR testing.  Let's try and keep PR tests to the ones where going red is a sign of serious issue.  I'm certainly a fan of no merging until green!  But that can be a little awkward right now.

lbruun:
> 
> So, I'm saying -1 to merging this (not the implementation!) until conversations on dev@ and with infra and we can be assured we can do this without negative impact.
> 

Perhaps merge now but with a manual trigger only  (`workflow_dispatch`) ?  This will allow to move on and get some experience. And then improve gradually. The person who wishes to merge can start the workflow on a need-basis then.


> An annoyance I have with our GitHub Actions as opposed to the Travis CI config is that at the moment a failure in GitHub Actions needs all tasks to be re-run, whereas Travis allows for retriggering the failing test only. (At least I've never found a way to retrigger part of a GH workflow?). So, the tests on Travis that have a tendency to fail at random would definitely be good ones to move off PR testing. Let's try and keep PR tests to the ones where going red is a sign of serious issue. I'm certainly a fan of no merging until green! But that can be a little awkward right now.

Speaking of that I've experienced _intermittent_ problems with the Ant [DownloadBinaries ](https://github.com/apache/netbeans/blob/master/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java) task when executed in GA. Part of the problem is that the task seems to swallow some exceptions which makes it hard to debug. 

And yes, I too have been annoyed by the need-to-rerun-all-jobs.  See https://github.com/actions/runner/issues/432.

neilcsmith-net:
> Perhaps merge now but with a manual trigger only (workflow_dispatch) ? This will allow to move on and get some experience. And then improve gradually. The person who wishes to merge can start the workflow on a need-basis then.

@lbruun certainly!  My -1 is there purely to ensure we don't land a massive amount of extra call on GH runners without liaison with infra.

> And yes, I too have been annoyed by the need-to-rerun-all-jobs. See actions/runner#432.

Thanks for the link.  I guess the other option might be to split into multiple workflows?  Pros and cons?

KacerCZ:
I extracted parts of `main.yml` into smaller files to speed-up testing.
Workflows are not running automatically, but they were green.

KacerCZ:
Status of individual workflows:
* `macos` - extracted from `main`, passing
* `main` - added jobs from Travis, not working
* `test-enterprise` - converted from Travis, passing
* `test-groovy` - converted from Travis, not working
* `test-passing` - converted from Travis, passing
* `test-php` - extracted from `main`, passing
* `windows` - extracted from `main`, passing

Notes on Groovy workflow:
* I was not able to successfully run tests with different JDK even on my local machine.
* Test matrix is not used because build is done only once with Java 8 and then tests are run with different JDK versions.
* Contains action which allows to install different JDK on test runner.

According to [current limitations](https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status) of GH Actions for ASF projects I don't plan to work on conversion any more.

mbien:
closing in favor of #4817

PR #2709, opened 2021-01-24
Fix versioning of external dependency cpplite-mi

matthiasblaesing:
Publishing fails with "-SNAPSHOT" dependencies. This PR uses the id of
the originating mercurial commit as version.

### DIFF
--- a/cpplite/cpplite.debugger/external/binaries-list
+++ b/cpplite/cpplite.debugger/external/binaries-list
@@ -14,4 +14,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-7E2B3E2AB5EBCABF3971F118A4D288A7E5DAC1DF cpplite-mi-1.0-SNAPSHOT.jar
+2BF984B5372F003D3E6C2E96A34F45D659168534 eu.doppel-helix.netbeans.lib:cpplite-mi:f8f8250283be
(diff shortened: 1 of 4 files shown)


matthiasblaesing:
I think we had two releases now with the same problem when the maven artifacts are published: The cpplite cluster fails publication.

12.1: https://lists.apache.org/thread.html/r9e685dbceb684edd88d54f2c40416c5c42e546459361d58ed2ca13bb%40%3Cdev.netbeans.apache.org%3E
12.2: https://lists.apache.org/thread.html/rc00921f6dceacac959bacdc0e5b13af8f65cc2a8eb1061f6e1424da5%40%3Cdev.netbeans.apache.org%3E

The reason from my POV is that, -SNAPSHOT versions are treated specially by maven. They are distributed via different repositories and follow different rules (have a look at the "real" artifact versions of snapshots after upload), so I assume that sonatype (or the nexus instance of apache) block artifacts with "-SNAPSHOT" versions and/or dependencies.

The artifact for the dependency is already packaged on maven central. To do that, I had a look at the artifact @jlahoda created and replicated the work based on his. I updated the pom to the requirements of maven central and pushlished the artifact there using my personal groupId.

I targeted this to the delivery branch, as I think that this should help easing release.

neilcsmith-net:
Thanks @matthiasblaesing Yes, that is exactly the reason. Had asked @entlicher to have a look in context of other recent CPPLite debugger PR. Have tagged for review (and stop duplication), then will look to merge.

jlahoda (approved):
Looks good to me, thanks Matthias!

matthiasblaesing:
After the failed checks from github, I rechecked the commit and there was an error in the project.properties (`cpplite-mi-f8f8250283be.jar.jar` instead of `cpplite-mi-f8f8250283be.jar`. I ran commit validation locally and indeed it failed. I fixed the issue, commit validation now runs clean.

neilcsmith-net (approved):
OK, let's merge this. Thanks!

PR #2710, opened 2021-01-25
Don't use local .m2 cache in build by default.

neilcsmith-net:
Always download Maven binaries by default rather than relying on local .m2 cache (issues with Jenkins builds).
Expose binaries.repos property so source repositories of binaries can be configured, including local .m2 if desired.

Passing eg. `-Dbinaries.repos="file:///<HOME>/.m2/repository/ https://repo1.maven.org/maven2/"`to ant still allows use of local binaries if necessary.


lkishalmi (approved):
Looks sane.

neilcsmith-net:
Thanks @lkishalmi It's a simple change, but let's hope no-one / nothing is relying on the previous default.  Merging - let's see if we can get a beta built!

PR #2711, opened 2021-01-25
Popup tab switcher should hide when another application steals focus

timboudreau:
Fixes a problem I've run into several times on Mac OS - if another application steals focus
while the popup switcher is open, the switcher can wind up in a state where it is listening
twice, and is hidden as soon as you release the mouse button.

Fix is to add AWTEvent.WINDOW_EVENT_MASK to the set of AWTEvents listened for, check that
there really is no active window (to avoid confusion around heavyweight AWT popups) and
hide the popup in that case.

### DIFF
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/ButtonPopupSwitcher.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/ButtonPopupSwitcher.java
@@ -21,11 +21,13 @@ package org.netbeans.swing.tabcontrol;
 
 import java.awt.AWTEvent;
 import java.awt.Component;
+import static java.awt.KeyboardFocusManager.getCurrentKeyboardFocusManager;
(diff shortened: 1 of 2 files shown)


platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/ButtonPopupSwitcher.java
```diff
@@ -21,11 +21,13 @@
 
 import java.awt.AWTEvent;
 import java.awt.Component;
+import static java.awt.KeyboardFocusManager.getCurrentKeyboardFocusManager;
```
sdedic, line 24: Unusual ;) in NB coding style.

sdedic (approved):

sdedic:
@timboudreau should this go to 12.4 (attach to a milestone) ?

geertjanw:
Seems no reason not to merge this.

PR #2717, opened 2021-01-28
[NETBEANS-5004] Update macOS native launchers to call launcher shell script rather than nbexec

oyarzun:
Update macOS Native Launchers to call the unix launcher shell scripts instead of duplicating the logic in the scripts.

As @neilcsmith-net pointed out in https://github.com/apache/netbeans/pull/2563#issuecomment-765627457 this will avoid having to update the logic in yet another place as we do with the Windows Launcher.

### DIFF
--- a/harness/apisupport.harness/macosx-launcher-src/Sources/AppLauncher/main.swift
+++ b/harness/apisupport.harness/macosx-launcher-src/Sources/AppLauncher/main.swift
@@ -19,150 +19,11 @@
 
 import Foundation
 
-func getDefaultDir(for directory: FileManager.SearchPathDirectory) -> String {
-    let urls = FileManager.default.urls(for: directory, in: .userDomainMask)
-    let defaultDir = URL(string: brandingToken!, relativeTo: urls[0])
-
-    return defaultDir!.path
-}
-
-func processClusters(file: String) -> String {
-    let appDir = Bundle.main.path(forResource: brandingToken, ofType: "")
-    var clusters = String()
-    do {
-        let contents = try String(contentsOfFile: file)
-        let lines = contents.components(separatedBy: .newlines)
-        for line in lines {
-            if (line.starts(with: "#")) {
-                continue;
-            }
-            clusters = clusters + appDir! + "/" + line + ":"
-        }
-    } catch {
-        print("error reading" + brandingToken! + ".clusters")
-    }
-
-    return clusters
-}
-
-func processConf(confFile: String, confDict: inout Dictionary<String, String> )  {
-    //set default userdir and cachedir
-    let userdir_root = getDefaultDir(for: .applicationSupportDirectory)
-    let cachedir_root = getDefaultDir(for: .cachesDirectory)
-    
-    do {
-        let contents = try String(contentsOfFile: confFile)
-        let lines = contents.components(separatedBy: .newlines)
-        for line in lines {
-            if (line.starts(with: "#")) {
-                continue;
-            }
-        
-            if let idx = line.firstIndex(of: "=") {
-                let varRange = line.startIndex..<idx
-                let idx2 = line.index(after: idx)
-                let valRange = idx2..<line.endIndex
-                let nbvar = String(line[varRange])
-                let nbval = String(line[valRange])
-                
-                // strip quotes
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
Thanks @oyarzun To be clear, the duplication of logic is not the only concern I have.  The Windows launcher is the same launcher used in the zip bundles, and thus daily builds, betas, etc.  It's having two methods of launching on macOS, both requiring updates and testing, one of which more likely to be used by devs, one by users, and which we're not (at the moment) in a good position to test.

This looks a good change to me, and I like passing in the dock name as an environment variable.  That actually covers one use I've had in platform apps for forking the launch script.

In the original PR you mentioned using Apple APIs to look up userdir and cache locations.  Are the values in the script always correct?  If not, another use for environment variables?

oyarzun:
> In the original PR you mentioned using Apple APIs to look up userdir and cache locations. Are the values in the script always correct? If not, another use for environment variables?

The values in the script are correct, so there is no need to add it to the binary.

```shell
        # set default userdir and cachedir on Mac OS X
        DEFAULT_USERDIR_ROOT="${HOME}/Library/Application Support/${APPNAME}"
        DEFAULT_CACHEDIR_ROOT=${HOME}/Library/Caches/${APPNAME}
```

> This looks a good change to me, and I like passing in the dock name as an environment variable. That actually covers one use I've had in platform apps for forking the launch script.

What are your other changes to the launch script? Anything that should be added to the binary?

neilcsmith-net:
Thanks @oyarzun 

> The values in the script are correct, so there is no need to add it to the binary.

Good to know.

> What are your other changes to the launch script? Anything that should be added to the binary?

Nothing usefully generic.

neilcsmith-net (approved):

oyarzun:
@mcdonnell-john when you get a chance can you review this PR to see if can be merged into master?

mcdonnell-john:
Hi @oyarzun, Yeah no problems, I'll have a look at the weekend.

mcdonnell-john (approved):
Managed to create an Installer for Apache NetBeans with this change.

No issues from me

neilcsmith-net:
@oyarzun @mcdonnell-john in case you haven't noticed, can you check the issue at https://issues.apache.org/jira/browse/NETBEANS-5481and see if there's a cause that needs addressing in this PR?  Or is this a build issue locally for the installer?  Something that could be fixed in an installer update for 12.3?  Thanks!

oyarzun:
@neilcsmith-net it will need to be addressed in this PR since the Swift 5 runtime is need to macOS prior to 10.14.4. We can add some checks in the installer logic to check if it needs to be installed. 

https://support.apple.com/kb/dl1998?locale=en_US


I imagine we can not package the runtime due to potential license issues? Would the installer have to download the runtime?

neilcsmith-net:
Is there overlap between where Swift runtime is required and where NETBEANS-5004 requires to not use a shell script?  Because if going as far as the installer checking macOS version, would it make sense to revert to the old shell script on < 10.14.4 rather than requiring an extra download?  Just considering what might be easier here or for users.  Somehow needs a fix for 12.4, and probably a description and link on the 12.3 download page?

oyarzun:
IMHO, the easiest would be to add a link on the webpage and note that macOS < 10.14.4 requires the user to download the Swift 5 runtime library. But adding it in the installer would avoid those who don't read the note on the webpage.

I'm not familiar enough with the mac Package installers to know how easy it would be to add logic to use the script for earlier versions instead of the swift launcher.

oyarzun:
@neilcsmith-net given the discussion on the [dev mailing list](https://lists.apache.org/thread.html/r099b047a8b9d7dc9b52c8b8d3e181aa641ead28b7cd6907b8267ab1e%40%3Cdev.netbeans.apache.org%3E) should we go ahead and merge this for 12.4?

neilcsmith-net:
@oyarzun yes, let's do that before beta - cc @geertjanw 

@mcdonnell-john we're going to vote on beta this time around.  Once that's through, could we get a macOS installer with this for testing purposes?  Probably good to widen the pool of people who know how to handle that process too!

geertjanw (approved):

PR #2718, opened 2021-01-29
Fixing failing build.

dbalek:
- `PrepareBundles` task should use the same versions of NPMs as the extension itself
- stop LSP client before killing server while restarting NBLS

### DIFF
--- a/java/java.lsp.server/vscode/src/extension.ts
+++ b/java/java.lsp.server/vscode/src/extension.ts
@@ -41,7 +41,7 @@ import * as launcher from './nbcode';
 import { StatusMessageRequest, ShowStatusMessageParams, QuickPickRequest, InputBoxRequest } from './protocol';
 
 const API_VERSION : string = "1.0";
-let client: Promise<LanguageClient>;
+let client: Promise<LanguageClient> | null = null;
 let nbProcess : ChildProcess | null = null;
 let debugPort: number = -1;
 let consoleLog: boolean = !!process.env['ENABLE_CONSOLE_LOG'];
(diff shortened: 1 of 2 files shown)


JaroslavTulach (approved):
The "VSNetBeans gate" succeeded with this change, so there must be some improvement!

There are some failing "javac"-releated gates. However they cannot be failing because of this PR (as it only changes `.ts` files). I suggest to integrate anyway.

Btw. don't you want to integrate into `delivery` branch? So the change ends up both in 12.3 and `master`?

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -41,7 +41,7 @@ import * as launcher from './nbcode';
 import { StatusMessageRequest, ShowStatusMessageParams, QuickPickRequest, InputBoxRequest } from './protocol';
 
 const API_VERSION : string = "1.0";
-let client: Promise<LanguageClient>;
+let client: Promise<LanguageClient> | null = null;
```
JaroslavTulach, line 44: Ideally this `client` promise could be set soon to be defined (not `undefined`) and resolved later. That would mean to make sure the `client` promise is set sooner then `registerCommand` is registered. E.g. 
```
let c : LanguageClient = await client
```
can only be called when the `client` is defined.

dbalek: Agreed. Fixed.

MartinBalin:
This is the change to build so we have 12.3 Beta binaries of VSCode for testing, available https://ci-builds.apache.org/job/Netbeans/view/vscode/job/netbeans-vscode/

neilcsmith-net:
@MartinBalin @dbalek we'll be looking to merge today to trigger beta2 for tomorrow.  I assume this is now ready? Please can you squash and force push your branch into a single commit for merging.

In terms of the beta for VSCode, that task seems to be building master?  We need a build off of release123, and we should do the VSCode beta2 off of the same commit as the rest of beta2.

MartinBalin:
@neilcsmith-net - I assume apache:delivery is enough.

JaroslavTulach:
This PR has been merged into `delivery`. It is going to become part of 12.3 once @neilcsmith-net merges that branch into 12.3 line. All is good.

PR #2720, opened 2021-01-30
Fixing a mistake in ergonomics project detection, so that it can detect OpenJDK projects.

jlahoda:
Basically, if the project marker file relative path contains `../`, we attempt to use the parent file. But, the `../` is never stripped from the marker file relative path. Trying to fix here.

### DIFF
--- a/ergonomics/ide.ergonomics/src/org/netbeans/modules/ide/ergonomics/fod/FeatureProjectFactory.java
+++ b/ergonomics/ide.ergonomics/src/org/netbeans/modules/ide/ergonomics/fod/FeatureProjectFactory.java
@@ -150,6 +150,8 @@ implements ProjectFactory, PropertyChangeListener, Runnable {
                 return false;
             }
 
+            relative = relative.substring(pos);
+
             if (relative.contains("*")) {
                 for (String segment : relative.split("/")) {
                     FOUND: if (segment.contains("*")) {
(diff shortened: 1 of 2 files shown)


yasharya2901:
Hello, I am having some problems with Netbeans. I tried sending an email to users@netbeans.apache.org but couldn't send the email. It said email couldn't be sent to the given address. I didn't know where and who to ask and so I am writing it here. I am simply pasting the email I wrote here:

I downloaded the Netbeans from the official sh script from the website in my latest KDE Neon operating system.
Here's the text from the About option in my netbeans:
Product Version: Apache NetBeans IDE 12.0

Java: 15.0.2; OpenJDK 64-Bit Server VM 15.0.2+7-27

Runtime: OpenJDK Runtime Environment 15.0.2+7-27

System: Linux version 5.4.0-65-generic running on amd64; UTF-8; en_IN (nb)



In the screenshot I have attached, you can see that no Buttons or labels are visible to be dragged into the window. Please help!    

Netbeans Swing not shown

JaroslavTulach:
CCing @jlahoda

JaroslavTulach (approved):
Integrating together with a unit test would be great.

neilcsmith-net:
Thanks for test, but can you squash the commits and force push for merging?

geertjanw (approved):

PR #2722, opened 2021-01-31
[NETBEANS-5304] CSS: Prevent UI blocking on undocumented properties

matthiasblaesing:
When the documentation for the target property could not be found, the
whole page of the CSS module was passed as help text. The rendering of
the HTML in combination with a popup menu caused problems on hide.

The documentation could be rendered, but on hide, the EDT is stuck in a
call to:

javax.swing.PopupFactory$MediumWeightPopup.hide()

Instead of rendering the whole documentation, the information "there is
no documenation" is rendered. While not optimal, it at least prevents a
heard freeze.


matthiasblaesing:
The failing unittest are the one I would expect to fail - I don't see how these failures could connect to the changeset.

geertjanw:
Great research and sounds like the right solution.

junichi11 (approved):
Looks good to me. I've verified that the problem doesn't occur. Thank you, Matthias!

junichi11:
Will re-run jobs.

neilcsmith-net:
Thanks, will merge for beta2.  Guess this is also showing a bug in Swing?  And that whole parsing looks a bit of a nightmare! 😄 

Don't worry about the particular failing tests - known test failures being looked at, and re-running won't help at the moment.

geertjanw (approved):
Great that this is fixed.

PR #2725, opened 2021-02-03
[NETBEANS-5326] LSP codelens output fixes

entlicher:
This is a fix of NETBEANS-5326

If Run test/Debug test code lenses are used through LSP protocol, the actual test output is not visible. In case of Debug tes, the output is visible when it succeeds, however, it is significantly cut when it fails, preventing from seeing the test failure.

Correction of output from Run/Debug code lens actions.
- In case of Debug, the TerminatedEvent was sent too soon. Now, similar to Run, the TerminatedEvent is sent when the application terminates.
- Run now runs in the same was as "Run Without Debug" - through startDebugger with noDebug = true. This is necessary for consistency and proper output display.
- A NPE associated with a missing source path in case of Run is addressed

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
@@ -39,6 +39,7 @@ import org.eclipse.lsp4j.debug.OutputEventArguments;
 import org.eclipse.lsp4j.debug.Source;
 import org.eclipse.lsp4j.debug.TerminatedEventArguments;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
+import org.netbeans.api.java.classpath.ClassPath;
 import org.netbeans.modules.java.lsp.server.debugging.DebugAdapterContext;
 import org.netbeans.modules.java.lsp.server.debugging.NbSourceProvider;
 import org.netbeans.modules.java.lsp.server.debugging.utils.ErrorUtilities;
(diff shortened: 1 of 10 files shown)


JaroslavTulach (approved):
Sharing the _Run_ and _Debug_ and _single_ and _all_ execution is certainly good idea.

jlahoda (approved):

neilcsmith-net:
At the moment, we don't have another beta scheduled, although discussions elsewhere might lead to one.  If not, I don't feel in a position to assess this one against the [bug priority guidelines](https://cwiki.apache.org/confluence/display/NETBEANS/Bug+Priority+Guidelines). Can we get assessment of the necessity to get this in to 12.3? Thanks!

entlicher:
@neilcsmith-net I've updated the description and created NETBEANS-5326 based on our internal report. That should allow to asses the importance for 12.3. I consider the fix to be of a very low risk, especially comparing to the currently broken behavior.
The modifications are in the area of the LSP server implementation only.

neilcsmith-net:
@entlicher thanks.  It's more about criticals and blockers generally requiring another beta for testing. Quite happy for major (or minors) to come in then where available, particularly when they're low risk!  @geertjanw and I put forward an optimistic schedule with only two betas initially due to late start on the release (no RM in place).  At least one more beta is looking likely now anyway.

I've updated the title - hopefully Apache infra will link the two.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
```diff
@@ -161,7 +176,7 @@ private static OutputEventArguments convertToOutputEventArguments(String message
     protected void handleTerminatedEvent(DebugAdapterContext context) {
         CompletableFuture.runAsync(() -> {
             try {
-                waitForDebuggeeConsole.get(5, TimeUnit.SECONDS);
+                waitForDebuggeeConsole.get(1, TimeUnit.SECONDS);
```
sdedic, line 179: Hi, incidentally I attempted to solve this delay by removing it :) completely in apache/netbeans#2754 . See the discussion in that PR.

sdedic:
Merged per Neil's advice in #2754

PR #2732, opened 2021-02-04
Makes java.j2seplatform API / SPI public

emilianbold:
foojay has a nice plugin in development which cannot call java.j2seplatform APIs without being a friend.

This PR adds the plugin to the friend list.

### DIFF
--- a/java/java.j2seplatform/apichanges.xml
+++ b/java/java.j2seplatform/apichanges.xml
@@ -82,6 +82,19 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="public-api">
+            <api name="j2selibraries"/>
+            <summary>Make API / SPI packages public.</summary>
+            <version major="1" minor="55"/>
+            <date day="5" month="2" year="2021"/>
+            <author login="emi"/>
+            <compatibility binary="compatible" source="compatible" deprecation="no" deletion="no" addition="yes" />
+            <description>
+                <p>
+                    The API / SPI packages are public and any module can use them, not only friend modules.
+                </p>
+            </description>
+        </change>
         <change id="ask.on.volume.attach">
             <api name="j2selibraries"/>
             <summary>Rebrand defaults for attach source and javaddoc user questions</summary>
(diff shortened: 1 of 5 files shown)


geertjanw (commented):
Excellent. 

foojay.io is a vendor-neutral site focused on info (and APIs) promoting OpenJDK distributions, with Emilian working on integration into NetBeans.

JaroslavTulach:
CCing @tzezula - yes, it would be better to make the necessary API public rather than adding next friend.

emilianbold:
I made the friend packages into public packages. Builds locally, let's see what the CI says.

I wonder:

* [ ] Is these still some API review process? The existing API/SPI packages are quite old.
* [ ] would `org-netbeans-modules-java-j2seplatform.sig` need any tweaks? I would seem not, but I never looked into that.
* [ ] should I go and remove impl dependencies for existing friends modules? This seems optional to me.

JaroslavTulach:
I am seeing following classes in the API: `J2SEPlatformCreator`, `J2SEPlatformDefaultJavadoc`, `J2SEPlatformDefaultSources`, that's a reasonable size, I guess.

> Is these still some API review process?

Just a regular review before integration with more attention to API aspects.

> would org-netbeans-modules-java-j2seplatform.sig need any tweaks?

No. Keep it unchanged.

> should I go and remove impl dependencies for existing friends

Hopefully there are no "implementation" dependencies. Friend modules use a regular dependency. You can keep it unchanged, but:

- you need to version your change - see [this essay](http://wiki.apidesign.org/wiki/FriendDependencies) - otherwise your module would get `NoClassDefFoundError`
- add an `apichanges.xml` entry describing the module went public
- your module has to depend on the new version that made this module public
- change `arch.xml` to say the API is public
- make sure the module is listed among public API modules in `nbbuild/build.properties`
- try to build `ant build-javadoc` to verify everything looks sane at the end

I am assuming your module is going to download and install various JDKs. Good luck with it! I hope to see it on our plugin portal soon.

emilianbold:
@JaroslavTulach I hope the apichanges are fine.

The Javadoc looks OK to me, no more red/'friend' link.

I don't know what to do about the new lines with `javadoc.docs.org-netbeans-modules-java-j2seplatform` added to these files:

* nbbuild/javadoctools/links.xml
* nbbuild/javadoctools/properties.xml
* nbbuild/javadoctools/replaces.xml

Should they also be committed?

JaroslavTulach (approved):

emilianbold:
Wondering if there's something more that needs to be done here? I see @JaroslavTulach is pleased with the PR.

PR #2734, opened 2021-02-05
[GR-28352] Java: Compile Workspace compiles more than workspace.

dbalek:
Modified to compile only the project that correspond to the current workspace folders.


JaroslavTulach (approved):
What happens when a new project is added to the workspace?

dbalek:
> What happens when a new project is added to the workspace?

That is the general problem of the current language server implementation. Currently, scan for projects contained in the workspace is done only once as a part of `initialize` request. Projects found are opened and indexed. But there is IMHO no code that opens and index a project added to the workspace later - which is beyond the scope of this PR and should be solved separately.

sdedic (approved):

JaroslavTulach (approved):

PR #2739, opened 2021-02-06
[NETBEANS-4587] Code completion for HTML forms

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4587

* Updated list of values for `method` attribute in forms.
* Added completion for `enctype` and `method` attributes of `<form>` element.
* Added completion of `target` attribute of `<a>`, `<button>`, `<form>` and `<input>` elements.
* Improved names of `FORM_ENCTYPE_SUPPORT` and `FORM_METHOD_SUPPORT`.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method

For easier review I made two commits. The first one groups definitions for same element together and removes duplicates. The second one makes actual changes in completion.


geertjanw:
Great work.

KacerCZ:
@geertjanw Thanks for review.

KacerCZ:
@geertjanw Please, set milestone to 12.4. I don't have permissions to set it by myself.

PR #2741, opened 2021-02-06
[NETBEANS-4637] Remove obsolete PHP code templates

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-4637

Removes templates for obsolete MySQL extension.
MySQL extension which was deprecated in PHP 5.5 and removed in PHP 7.0.


junichi11:
Frankly, this is an irritating PR. I won't review and won't approve this. It would be a bad example if this were merged.

KacerCZ:
You are right. I'm sorry.

PR #2742, opened 2021-02-07
[NETBEANS-5340] Prevent NPE on TaskNode when opening Gradle Projects

lkishalmi:
I've seen this issue only once so far, when closed some Gradle projects, having no project open in the IDE.
Then opening new other Gradle projects were throwing that NPE.


lkishalmi:
The fix is trivial I, hope it can go into 12.3.

neilcsmith-net (approved):
Sure! Only had a quick glance in context - curious at concurrent access between EDT and not, but assume all setting of `currentP` is on EDT?

lkishalmi:
Well, as far as I could teh the currentP is written on the EDT. The issue is that it can set to be null, by closing  a project, while a weak referenced listener use the previous FavoriteTaskManager. Probably needs a better implementation in the future.

PR #2743, opened 2021-02-07
[NETBEANS-5090] fix missing Serializable interface when a @ViewScoped  CDI bean is created via JSF CDI Bean Wizard

jherkel:
see https://issues.apache.org/jira/browse/NETBEANS-5090


sdedic (approved):

matthiasblaesing:
Looks good to me.

PR #2744, opened 2021-02-07
Options panel use GTK default colors

hectorespert:
Use default colors in options panel when GTK laf is used.

Tested in Ubuntu 20.04 with default theme. 

Before:
Screenshot from 2021-02-07 22-39-09

After:
Screenshot from 2021-02-07 22-40-04

### DIFF
--- a/platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/LFCustoms.java
+++ b/platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/LFCustoms.java
@@ -423,4 +423,11 @@ public abstract class LFCustoms {
     public static final String PROGRESS_CANCEL_BUTTON_ICON = "nb.progress.cancel.icon";
     public static final String PROGRESS_CANCEL_BUTTON_ROLLOVER_ICON = "nb.progress.cancel.icon.mouseover";
     public static final String PROGRESS_CANCEL_BUTTON_PRESSED_ICON = "nb.progress.cancel.icon.pressed";
+
+    /**
+     * Keys used by the options dialog module.
+     */
+    public static final String OPTIONS_USE_UI_DEFAULT_COLORS = "nb.options.useUIDefaultsColors";
+    public static final String OPTIONS_CATEGORIES_SEPARATOR_COLOR = "nb.options.categories.separatorColor";
+    public static final String OPTIONS_CATEGORIES_BUTTON_USE_NIMBUS = "nb.options.categories.button.useNimbusCategoryButton";
 }
--- a/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
+++ b/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
@@ -78,7 +78,6 @@ import javax.swing.border.Border;
 import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
 import javax.swing.border.LineBorder;
-import javax.swing.plaf.metal.MetalLookAndFeel;
 import javax.swing.text.JTextComponent;
 import org.netbeans.modules.options.CategoryModel.Category;
 import org.netbeans.modules.options.advanced.AdvancedPanel;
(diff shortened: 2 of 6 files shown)


neilcsmith-net:
Sorry, -1 from me on this.  Although +1 to fixing it.  We can't keep adding exceptions for look and feels over the code base.  Better to fix this so that a single key works for all LAFs.  If `separator.foreground` doesn't work for all, maybe a specific key for this panel that does?

hectorespert:
> Sorry, -1 from me on this. Although +1 to fixing it. We can't keep adding exceptions for look and feels over the code base. Better to fix this so that a single key works for all LAFs. If `separator.foreground` doesn't work for all, maybe a specific key for this panel that does?

I added a keys to modify this panel to allow remove LAFs exceptions but this PR only removes exceptions for the GTK laf.
I can't test other LAFs and they could be modified later.

lkishalmi:
A step in the good direction, though if we are adding new LAF keys, probably shall add constants for it in LFCustoms.
And by introducing that it would be good to check if all supported LAF-s are providing defaults for the new key.

A sidenote: It would be really great if you could detect somehow when the GTK skin is set to dark, if it does, it would be great at least add the ```"nb.dark.theme"``` to true.

hectorespert:
> A step in the good direction, though if we are adding new LAF keys, probably shall add constants for it in LFCustoms.

I added the new constants in LFCustoms.

> And by introducing that it would be good to check if all supported LAF-s are providing defaults for the new key.

New keys are only used by GTK theme, I would like to fix the problem with that theme before start to modify the other themes.

> A sidenote: It would be really great if you could detect somehow when the GTK skin is set to dark, if it does, it would be great at least add the `"nb.dark.theme"` to true.

A GTK configuration property to detect that is available: https://developer.gnome.org/gtk3/3.8/GtkSettings.html#GtkSettings--gtk-application-prefer-dark-theme

geertjanw:
Changing milestone to 12.5.

ebarboni:
@neilcsmith-net @lkishalmi  are the changes better now ? Can we consider merging ?

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/LFCustoms.java
```diff
@@ -423,4 +423,11 @@ private void disposeValues (Object[] arr) {
     public static final String PROGRESS_CANCEL_BUTTON_ICON = "nb.progress.cancel.icon";
     public static final String PROGRESS_CANCEL_BUTTON_ROLLOVER_ICON = "nb.progress.cancel.icon.mouseover";
     public static final String PROGRESS_CANCEL_BUTTON_PRESSED_ICON = "nb.progress.cancel.icon.pressed";
+
+    /**
+     * Keys used by the options dialog module.
+     */
+    protected static final String OPTIONS_USE_UI_DEFAULT_COLORS = "Nb.options.useUIDefaultsColors";
```
lkishalmi, line 430: Why protected instead of public?
"Nb" or "nb"? Yes I've seen both prefixes, however it would be good to stick with "nb.options"

platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
```diff
@@ -877,7 +869,7 @@ private void addCategoryButtons () {
                 
     private CategoryButton addButton (CategoryModel.Category category) {
         int index = buttons.size ();
-        CategoryButton button = isNimbus || isGTK 
+        CategoryButton button = UIManager.getBoolean("Nb.options.categories.button.useNimbusCategoryButton") //NOI18N
```
lkishalmi, line 872: Would be nicer to reintroduce the same constants as in LFCustoms at the top of the file again (now it can be not public)

lkishalmi (changes requested):
Generally I'm Ok with the implementation, just those constants...

ebarboni:
change resquested moving to 12.6 for now

matthiasblaesing:
This seems to be a working implementation, which even looks like a cleanup. Any reason not to merge now? (the unittests that fail are known to be flaky).

hectorespert:
> Any reason not to merge now?

It's only waiting for a positive review.

lkishalmi (approved):
Seems to be fine now.

PR #2752, opened 2021-02-10
Adding exports so that opening classfiles without sources works.

jlahoda:
Attempts to open a class for which source file is not available leads to an IllegalAccess exception on JDK 16+. At least temporary solution is to add some exports, although we may want some better solution over time:
```
-J--add-exports=jdk.jdeps/com.sun.tools.classfile=ALL-UNNAMED -J--add-exports=jdk.jdeps/com.sun.tools.javap=ALL-UNNAMED
```

Not sure if there's time for 12.3, but that would be ideal, I think. Thanks!


matthiasblaesing:
Looks safe to me.

PR #2755, opened 2021-02-10
[NETBEANS-5350] Improved Fix Uses dialog

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5350

Added buttons "Change Nothing" and "Restore Defaults" to Fix Uses dialog.
This is useful when user just wants to remove unused `use` statements or wants to fix only a few imports and dialog contains large number of items.

Initial state  and state after "Restore Defaults":
netbeans-5350-defaults

State after "Change Nothing":
netbeans-5350-none

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/actions/Bundle.properties
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/actions/Bundle.properties
@@ -33,14 +33,14 @@ ImportChoices=Import choices:
 ToggleCommentAction_shortDescription=Toggle Comment
 toggle-comment=Toggle Comment
 
-FixDupImportStmts_Title=Fix Uses in Current Namespace
-FixDupImportStmts_IntroLbl=<html>Select the fully qualified name to use in the use statement.</html>
-FixDupImportStmts_Header=Use Statements:
+FixDupImportStmts_Title=Fix Imports in Current Namespace
+FixDupImportStmts_IntroLbl=<html>Select the fully qualified name to import.</html>
+FixDupImportStmts_Header=Import:
 FixDupImportStmts_Combo_ACSD=Choose appropriate fully qualified name
 FixDupImportStmts_Combo_Name_ACSD=Fully Qualified Names
-FixDupImportStmts_UnusedImports=&Remove unused uses
+FixDupImportStmts_UnusedImports=&Remove unused imports
 FixDupImportStmts_CannotResolve=<html><font color='#FF0000'>&lt;cannot be resolved&gt;
 FixDupImportStmts_NothingToFix=<nothing to fix>
-FixDupImportStmts_checkUnusedImports_a11y=Remove unused uses checkbox
+FixDupImportStmts_checkUnusedImports_a11y=Remove unused imports checkbox
 
-Actions/Source/org-netbeans-modules-php-editor-actions-FixUsesAction$GlobalAction.instance=Fix Uses...
+Actions/Source/org-netbeans-modules-php-editor-actions-FixUsesAction$GlobalAction.instance=Fix Imports...
--- a/php/php.editor/src/org/netbeans/modules/php/editor/actions/FixDuplicateImportStmts.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/actions/FixDuplicateImportStmts.java
@@ -28,18 +28,22 @@ import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.Rectangle;
 import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
 import java.awt.event.FocusEvent;
 import java.awt.event.FocusListener;
 import java.awt.event.KeyEvent;
 import java.util.ArrayList;
 import java.util.List;
 import javax.swing.AbstractAction;
+import javax.swing.Box;
 import javax.swing.Icon;
 import javax.swing.InputMap;
+import javax.swing.JButton;
 import javax.swing.JCheckBox;
 import javax.swing.JComboBox;
 import javax.swing.JLabel;
 import javax.swing.JList;
+import javax.swing.JPanel;
 import javax.swing.KeyStroke;
 import javax.swing.ListCellRenderer;
 import javax.swing.UIManager;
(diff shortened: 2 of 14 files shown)


php/php.editor/src/org/netbeans/modules/php/editor/actions/FixDuplicateImportStmts.java
```diff
@@ -147,6 +153,58 @@ private static String getBundleString(String s) {
...
+            dontUseVariants[i] = dataItem.getDefaultVariant();
+            for (ItemVariant variant : dataItem.getVariants()) {
+                if (!variant.canBeUsed()) {
+                    dontUseVariants[i] = variant;
+                }
+            }
+        }
+
+        JPanel buttonsPanel = new JPanel(new BorderLayout(5, 3));
+        bottomPanel.add(buttonsPanel, BorderLayout.LINE_END);
```
KacerCZ, line 176: I used same way as "Remove unused uses" is added to `bottomPanel`.
I don't know why it was not created in designer.

sdedic:
A few comments (gathered with a help of an UI experienced colleague). 
Visual: The action buttons should be separated from the checbox by some visual spacing; like at least width of the button. They appear to be funcionally connected to the checkbox on the screenshot, which is not true.

Wording:
> Select fully qualified names to use in the use statement

- Either select different work for "use" verb, or somehow highlight (bold, syntax-highlight) the "use" statement name. Similar in the checkbox "unused uses" ( -> unneeded **use** statements, unnecessary uses,...)

- Don't use type  -- suggest that the type is not going to be used in the source; instead, the selection means 'do not produce use statement'. Consider rewording to "Do nothing" or "Skip"
- Change nothing -- seems more like 'reset' or 'cance'. Consider other wording like "Clear suggestions", "Clear all actions".

KacerCZ:
[PHP documentation](https://www.php.net/manual/en/language.namespaces.importing.php) uses "alias" or "import" for this usage of "use" statement.

PHP editor has already "Fast Import" action which inserts "require"/"include" statements. It could be confusing for user.

"Remove unused aliases" sounds good.

KacerCZ:
@junichi11 What is your opinion on naming "alias" vs "import"?

KacerCZ:
Original code used "import" so I decided to use it also in labels.

Here are new dialogs with updated texts:
netbeans-5350-2-defaults
netbeans-5350-2-none

sdedic (approved):
FIne from my perspective. If no objections (@junichi11  ? Other PHP _real_ reviewer ?) I'd close this after 72hrs grace period.

sdedic (changes requested):
Whoops ! I was again fooled by routinely failing builds. Please fix golden files in tests: `Don't import` does not match the expected content.
Sorry :(

KacerCZ:
@sdedic Tests are hopefully fixed.
@tmysik Could you review this PR, please?

tmysik:
@KacerCZ

Would it be possible to try the visual separation as Sváťa suggested?

CC @sdedic

php/php.editor/src/org/netbeans/modules/php/editor/actions/Bundle.properties
```diff
@@ -33,14 +33,14 @@ ImportChoices=Import choices:
...
-FixDupImportStmts_UnusedImports=&Remove unused uses
+FixDupImportStmts_UnusedImports=&Remove unused imports
 FixDupImportStmts_CannotResolve=<html><font color='#FF0000'>&lt;cannot be resolved&gt;
 FixDupImportStmts_NothingToFix=<nothing to fix>
-FixDupImportStmts_checkUnusedImports_a11y=Remove unused uses checkbox
+FixDupImportStmts_checkUnusedImports_a11y=Remove unused imports checkbox
 
-Actions/Source/org-netbeans-modules-php-editor-actions-FixUsesAction$GlobalAction.instance=Fix Uses...
\ No newline at end of file
+Actions/Source/org-netbeans-modules-php-editor-actions-FixUsesAction$GlobalAction.instance=Fix Uses...
```
tmysik, line 46: Perhaps better `Fix Imports...` then?

tmysik:
@KacerCZ 

I am not sure that I understand what this PR is solving exactly, what situation - can you please explain? When exactly I want to click any of those 2 new buttons?

Looking into the ticket, I see:

> When user wants to fix only import of one class using "Fix Uses" action, then he has to manually go through all popup menus and select "Don't use type".

When, as a user, I want to fix import of only one class? What happens with the other imported classes?

Thanks.

KacerCZ:
I'll try to explain my use cases.

I work on projects where classes and interfaces from root namespace are not imported.
For example `\DateTime` or `\Countable` are never imported and fully qualified name is always used.
Also there are other types which are not imported - typically objects passed only in constructor or classes used infrequently.
This leads to situations where list of suggested imports contains types which I don't want to import.

Running "Fix Uses" action just to sort existing and remove unused imports I have to change all suggestions to "Don't use type" one by one. "Clear Suggestions" button simplifies this change to one click.

Another use case is when I want to import only a few of suggested types in existing code.
Again "Clear Suggestions" changes all proposed changes to "Don't use type" and those few for import I change back to import manually. Types marked as "Don't use type" are left in source code without change and only the selected types are imported.
Its useful in case class is used on many places in code - action replaces all occurrences with alias.

php/php.editor/src/org/netbeans/modules/php/editor/actions/Bundle.properties
KacerCZ: Changed.

KacerCZ:
> Would it be possible to try the visual separation as Sváťa suggested?

I made space between checkbox and buttons even larger.
netbeans-5350-button-spacing

tmysik:
@KacerCZ 
Thanks for the explanation.

@sdedic 
Sváťo, feel free to comment the UI change, if needed.

Thanks.

tmysik (approved):

sdedic:
re visual separation - I think it's clear enough. I'd suggest to add mnemonics to the controls in the dialog, and check tab-order (combos > checkbox > buttons). Sorry :( I missed that earlier (again).

sdedic (approved):

KacerCZ:
@sdedic "C" for "Clear Suggestions" and "D" for "Restore Defaults" is OK?

KacerCZ:
@sdedic I added mnemonics and verified tab order is exactly as you write.

tmysik:
@KacerCZ @junichi11 

I would wait for Junichi for a couple of days and then merge; is that OK for you?

junichi11:
@tmysik Please feel free to merge it. Thanks.

tmysik:
Just to be sure - PRs can be merged now, right? 😅

junichi11:
Yes :)

KacerCZ:
Thanks for the review.

Please set 12.4 milestone.

PR #2756, opened 2021-02-11
[NETBEANS-5354] - update public interface for a few users of LinkedList

BradWalker:
I've wanted to expand my knowledge about how to change the interface. So I've picked a simple one to change..

I've changed the interface to use List instead of LinkedList. One should use interfaces and not implementations in the public interface.. Changing this will allow me to fix the following warnings easily.

   [repeat] /home/bwalker/src/netbeans/enterprise/websvc.manager/src/org/netbeans/modules/websvc/manager/ui/TestWebServiceMethodDlg.java:390: warning: [rawtypes] found raw type: LinkedList
   [repeat]     public void methodFinished(final Object inReturnedObject,final LinkedList inParamList) {
   [repeat]                                                                    ^
   [repeat]   missing type arguments for generic class LinkedList<E>
   [repeat]     E extends Object declared in class LinkedList

### DIFF
--- a/java/dbschema/manifest.mf
+++ b/java/dbschema/manifest.mf
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.dbschema/1
-OpenIDE-Module-Implementation-Version: 4
+OpenIDE-Module-Implementation-Version: 5
 OpenIDE-Module-Layer: org/netbeans/modules/dbschema/jdbcimpl/resources/mf-layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/dbschema/resources/Bundle.properties
 AutoUpdate-Show-In-Client: false
--- a/enterprise/websvc.manager/manifest.mf
+++ b/enterprise/websvc.manager/manifest.mf
@@ -3,5 +3,5 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.websvc.manager
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/websvc/manager/Bundle.properties
 OpenIDE-Module-Install: org/netbeans/modules/websvc/manager/WebServiceModuleInstaller.class
-OpenIDE-Module-Specification-Version: 1.39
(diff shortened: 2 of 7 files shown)


java/dbschema/manifest.mf
```diff
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 OpenIDE-Module: org.netbeans.modules.dbschema/1
-OpenIDE-Module-Implementation-Version: 4
+OpenIDE-Module-Implementation-Version: 5
```
sdedic, line 3: Here you probably want to increment `specification-version` (see `project.properties`). You're changing `friend-public` API in a binary-incompatible way (but that should not matter as all friends are recompiled in the tree - right @JaroslavTulach  ?)

JaroslavTulach: Thanks for CCing me, Sváťo!

I see that the [signature test](https://travis-ci.com/github/apache/netbeans/jobs/482358349) failed with:
```
Class org.netbeans.modules.dbschema.jdbcimpl.IndexElementImpl

  "E1.2 - API type removed" : method protected void org.netbeans.modules.dbschema.jdbcimpl.IndexElementImpl.initColumns(java.util.LinkedList)

Class org.netbeans.modules.dbschema.jdbcimpl.SchemaElementImpl

  "E1.2 - API type removed" : method public void org.netbeans.modules.dbschema.jdbcimpl.SchemaElementImpl.initTables(org.netbeans.modules.dbschema.jdbcimpl.ConnectionProvider,java.util.LinkedList,java.util.LinkedList)
```

I suggest to not change any public signature incompatibly.  While...

> changing friend-public API in a binary-incompatible way (but that should not matter as all friends are recompiled in the tree

...is possible. You'd have to increase dependencies of all friend modules. It is easier to keep `LinkedList` where it was. Even I agree with [code against interfaces not implementation](http://wiki.apidesign.org/wiki/Code_Against_Interfaces%2C_Not_Implementations) advice, I value backward compatibility more.

sdedic (changes requested):
per Jarda's evaluation (I tend to agree with), I'd like to request relevant changes - also please check the versioning.

PR #2757, opened 2021-02-12
Avoid autoupdating existing NetBeans modules.

JaroslavTulach:
Gregor Kováč [reported problems on subsequent start of VSNetBeans](https://lists.apache.org/thread.html/r07d0861858590e3a047d284d1443b8794912f4caeb231a3de19efedc%40%3Cusers.netbeans.apache.org%3E) and we have tracked down the problem to autoupdate!

There is `updatecenters` module in VSNetBeans since 90d2af9 - it is needed to install `nbjavac` on JDK8. However it also provides other update centers.

There is code checking whether Apache NetBeans `java.lsp.server` module is properly started. It [uses a regexp](https://github.com/apache/netbeans/blob/master/java/java.lsp.server/vscode/src/extension.ts#L350) to parse the output and searches for:

```ts
stdOut.match(/org.netbeans.modules.java.lsp.server.*Enabled/)
```

However, since recently there has been a release of updates and the output now says:
```
org.netbeans.modules.java.lsp.server 1.6.1.1.1.12.1.17.32.5.2.1.1.1.13 
Upgrade to 1.7.0.1.1.12.1.17.32.5.2.1.1.1.13
```

the existing regexp doesn't match this output and as such VSCode never connects to the Apache NetBeans LSP server now!

This PR disables the other update centers and changes the the regexp.

### DIFF
--- a/java/java.lsp.server/build.xml
+++ b/java/java.lsp.server/build.xml
@@ -139,7 +139,11 @@
 
         <zip destfile="${build.dir}/apache-netbeans-java-${vsix.version}.vsix" basedir="${build.dir}/vscode-mandatory/" update="true" />
     </target>
-    <target name="test-vscode-ext" description="Tests the Visual Studio Code extension built by 'build-vscode-ext' target.">
+    <target name="test-lsp-server" description="Tests the LSP server behavior">
+        <ant dir="nbcode" target="test" inheritall="false" inheritrefs="false" />
+    </target>
+
+    <target name="test-vscode-ext" depends="test-lsp-server" description="Tests the Visual Studio Code extension built by 'build-vscode-ext' target.">
         <exec executable="npm" failonerror="true" dir="vscode">
             <arg value="--allow-same-version"/>
             <arg value="run" />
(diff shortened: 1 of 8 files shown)


dbalek (approved):

sdedic (approved):

MartinBalin:
I've seen this issue and it is completely breaking VSCode NetBeans ext. It was there around 10am here and it was gone 1 hour later.
At 10am I could see this output fro VSCode NB
`org.netbeans.modules.java.lsp.server        1.6.1.1.1.12.1.17.32.5.2.1.1.1.13 Upgrade to 1.7.0.1.1.12.1.17.32.5.2.1.1.1.13`
When I downloaded distribution center catalog from NB 12.2 now https://netbeans.apache.org/nb/updates/12.2/updates.xml.gz it  lists `1.6.0.1.1.12.1.17.5.2.13` version.
Seems like someone is playing with catalogs?

neilcsmith-net:
OK, as mentioned on dev@ it looks like the 12.2.1 update was released with a link to the dev UC, which as of a few weeks ago points to the daily builds - it was pointing at 12.2.  That shouldn't happen, and is probably because an entry was missed in https://github.com/apache/netbeans-jenkins-lib/blob/master/meta/netbeansrelease.json#L307

So, yes, getting this fix in looks good assuming you don't want external UCs, even if it should be mitigated by correct release process.

cc/ @geertjanw this answers the nb-javac in VSCode question from the other day then! :smile:

neilcsmith-net (approved):

JaroslavTulach:
There are some failing javac related builds, but those failures are unlikely associated with this PR. I am merging this change.

PR #2760, opened 2021-02-13
[NETBEANS-5362] Update FlatLaf from 1.0-rc2 to 1.0 to fix NPE in Jvi plugin when showing popup

DevCharly:
This FlatLaf update fixes a regression in FlatLaf 1.0-rc2 (and older)
that breaks the Jvi plugin as soon as a popup should be shown. 
E.g. pressing `:w` in editor.
https://issues.apache.org/jira/browse/NETBEANS-5362

FlatLaf 0.31, which was used in NB 12.2 (and before), did not have this issue.
So only NB 12.3 betas are affected.

The NPE occurs when invoking `PopupFactory.getSharedInstance().getPopup( owner, contents, x, y )`
with `null` as `owner`, which is actually unusual.
`PopupFactory` is e.g. used to show popups for code completion.

There are about 30 source files in NB repo that use `PopupFactory`:
https://github.com/apache/netbeans/search?q=popupfactory

So it is possible that also some popups in NB core are affected by this Flatlaf bug.

Hope we can get this fix into 12.3 beta 3.

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-5507827B39B40EED2942652D4792A51A5C49D67C com.formdev:flatlaf:1.0-rc2
+A807C984F00B683C65503803720821F12E136E93 com.formdev:flatlaf:1.0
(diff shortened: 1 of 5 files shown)


neilcsmith-net (approved):

neilcsmith-net:
Thanks. Let's get this in.

PR #2731, opened 2021-02-04
Support for VM/app argument injection to (maven) run and debug project actions

sdedic:
The need originated from LSP server, where the client wants to pass explicit arguments for executed application. As it turned out, this feature is generally handy in other scenarios, too.

In addition, Maven project support provides all nice pieces just in a single `exec.args` property suitable directly for `exec:exec` plugin: this is a pain for custom setups, which could eventually accept VM arguments and app parameters from the IDE (LSP server) but cannot extract them from the single command line.

The idea is to run the project action using `Lookups.executeWith()` and supply the lower layers with instructions how to modify the execution environment. @jtulach it may be also possible to use `context` parameter of `ProjectActionProvider.invokeAction()` ?

Here's the summary of changes:
- `ExplicitProcessParameters` immutable holder + builder API introduced in `extexecution` module. Serves as an instruction to modify process arguments, must be processed by the executor. 
- Maven executor refactored "a bit". 
- VM arguments and program arguments are now defined as properties; they are visible in Maven POM, too. `exec.args` now references the other properties.
- `RunJarPanel` logic extracted to a helper class, unit tests written.
- JAR packaging processing now combines POM or action mapping parameters with `ExplicitProcessArguments` to produce the result.

During the course, I've fixed some tests in Maven that lacked test data.

I also caught a race condition in Maven project implementation: if the project uses interpolated `packaging` attribute (i.e. `${packaging}`), that non-interpolated value will appear in the project and screws up the project Lookup contents: checkers for the real packaging will not be included, only afer 1st run. Fixed by forcing a reload after the real POM model is loaded by `MavenEmbedder`.

In this PR, just Maven support jar packaging is changed. I'd like to apply similar change to Gradle projects, but would probably need help from @lkishalmi .

PR #2761, opened 2021-02-14
Towards building on JDK11 and running on JDK8

jtulach:
As observed by Jan: [lookup module requires 1.8](https://lists.apache.org/thread.html/r082d9d71ed5227a0f6610cd03ddfd64bf7b7661dfe848f885fccf850%40%3Cdev.netbeans.apache.org%3E) - e.g. every module that depends on it, requires JDK8 as well.

1. let's update the `javac.target` to `1.8` - 7dbc4f9
2. some scripts define just `javac.source` - let's set the default target to 1.8: 4661771
3. Modify our `javac` task to use `setRelease(${javac.target})` to avoid using new than 1.8 APIs 
4. Update to Ubuntu 18.4 LST to get version of Ant that understands `setRelease`
5. Such update switches from Oracle JDK8 to OpenJDK8 - e.g. there is no JavaFX anymore and tests had to be updated
6. Drop `permit.jdk9.builds` flag, allow building on any JDK8+
7. Enhance `ant commit-validation` to verify all generated classfiles have JDK8's `major/minor` version

Things seem to be working. Approvals welcomed.


platform/libs.javafx/manifest.mf
```diff
@@ -38,5 +38,4 @@ OpenIDE-Module-Provides: javafx.animation,
     javafx.util,
     javafx.util.converter,
     netscape.javascript
-Class-Path: ${java.home}/lib/ext/jfxrt.jar
-
+Class-Path: %24%7Bjava.home%7D%2F/lib/ext/jfxrt.jar
```
lkishalmi, line 41: Well, this is ugly. Cant we just have JAVA_HOME/lib/ext/jfxrt.jar ?

platform/o.n.bootstrap/src/org/netbeans/StandardModuleData.java
```diff
@@ -78,7 +80,12 @@ public StandardModuleData(Manifest mf, StandardModule forModule) throws InvalidE
...
             while (tok.hasMoreTokens()) {
-                String ext = tok.nextToken().replace("%20", " "); // NOI18N
+                String ext;
+                try {
+                    ext = URLDecoder.decode(tok.nextToken(), "UTF-8");
+                } catch (UnsupportedEncodingException ex) {
+                    throw new IllegalStateException(ex);
+                }
                 File extfile;
                 if (ext.equals("${java.home}/lib/ext/jfxrt.jar")) { // NOI18N
```
lkishalmi, line 90: It seems we do not really require that $ sign.
Also there are some JDK 1.7 checks which could be removed as well.

platform/libs.javafx/manifest.mf
jtulach: Very ugly, I agree. I was also considering defining some new token instead of `${java.home}`, but that would mean API changes, versioning, etc. At the end I decided to apply a [simple bugfix](https://github.com/apache/netbeans/pull/2761#discussion_r577187116).

platform/o.n.bootstrap/src/org/netbeans/StandardModuleData.java
jtulach: Yes, the `${java.home}` string is totally artificial. I have chosen it as it had no meaning for `javac` (in JDK8, who could expect JDK11 starts to recognize URLs in `Class-Path` attribute!).

There already was some URI-like processing (replacing `%20` by ` `) and using `URLDecoder` (not available on JDK7 at the time of writing this code) can be seen as a "proper fix" of processing the URI component.

Right, the JDK7 specific code can be removed.

jlahoda: FWIW, you can probably use `java.nio.charset.StandardCharsets.UTF_8` instead of `"UTF-8"`, and avoid the need to handle the `UnsupportedEncodingException`.

jtulach: Alas, I cannot. That method has been added in JDK10! While it seems available in the editor, when compiling with the new `--release 8` the compilation fails. I tried and had to revert back to specifying the encoding as string and catching the exception.

neilcsmith-net:
+1 to the changes so far, -1 to the title.  Not sure we're ready for dropping building on JDK 8?  And so far this does build fine with it.

matthiasblaesing:
> +1 to the changes so far, -1 to the title. Not sure we're ready for dropping building on JDK 8? And so far this does build fine with it.

I understood, that the aim is to build with the "release" option from javac. That was added with java 9 and is not present on 8. It would be possible that ant emulates release 8 on JDK 8 by just using the classpath, but it would IMHO not be correct.

Bumping target version is not that interesting to me.

neilcsmith-net:
@matthiasblaesing yes, Ant basically ignores it (unlike Maven which requires fun with profiles).  So it should be pretty safe to allow both, seen as 8 is the only option without "release" that will work?  Dropping support entirely for compiling on JDK 8 seems a bigger step to take, particularly given we produce source releases.

jtulach:
To use the Ant's `<javac release="8"/>` tag I had to upgrade to Ubuntu 18.04 which provides newer Ant. That triggered a lot of new failures as the "bionic" docker image is not longer using OracleJDK, but OpenJDK. OpenJDK doesn't have JavaFX installed. A lot of tests were failing due to that. Now everything seems to be fine.


1067e68 adds new commit validation test - it checks if all the generated `.class` files are executable on JDK8. I am going to fix it now.

matthiasblaesing:
This looks interesting and an improvement. I have the following remarks/questions:

- There are adjustments to `javac.target` but not `javac.source`. This I don't understand - is the intention here to allow javac to generate more modern code, that uses features only present on Java 8+, but not suffer from code incompatibilites by using old source parser/behavior?
- Why is `javac.default.target` overriden to `1.8`, instead of relying on the default behavior in `nbbuild/templates/common.xml` - there the default target is initialized to `javac.source`, which sounds sane to me?
- I noticed some changes to `build.xml` files, which add `release="8" target="1.8" ` - on JDK 9+ target is not necessary, as release substitutes it. I would expect target to default to 8 on JDK 8, so this feels like a duplicate.

The bump of the baseline linux distribution for unittests looks sane to me. Thank you for working on this.

jtulach:
> This looks interesting and an improvement.

It is a small step forward (e.g. we still support compilation with JDK8 as @neilcsmith-net  requested) but I believe it is a step in the right direction. The most visible improvement is the dropping of `permit.jdk9.builds` flag. Now we can compile with JDK8+ and always get the same result (except maybe few modules where we continue to use `bootclasspath` and the `--release 8` API isn't enforced).

>  I noticed some changes to `build.xml` files, which add `release="8" target="1.8" ` - on JDK 9+ target is not necessary, as release substitutes it. I would expect target to default to 8 on JDK 8, so this feels like a duplicate.

The `release` flag is ignored on JDK8, so I am specifying both `release` and `target`. It feels safer to be explicit than to rely on some defaults. Anyway the outcome is important - and that's what is verified by the commit validation  `ValidateClassFilesTest`.

>  Why is `javac.default.target` overriden to `1.8`, instead of relying on the default behavior in `nbbuild/templates/common.xml` - there the default target is initialized to `javac.source`, which sounds sane to me?

@lkishalmi  [decided we want our modules to use JDK8](https://lists.apache.org/thread.html/r134fbdcee2583ceffab545827a260a2a8983a77a00c3bcec06271a3c%40%3Cdev.netbeans.apache.org%3E) whenever there is a [clash between the `javac.target` and the used API](https://lists.apache.org/thread.html/r42038bdaa1f66e44f20bc4f22482e8a246e8e28e61b6f5c2e92b837c%40%3Cdev.netbeans.apache.org%3E). One option to achieve that is to set source and target explicitly to `1.8` in all such modules. The other option was to make sure the default for NetBeans is `1.8`. Laszlo wrote _"let the past go"_, so I decided to go with the second option. 

>  There are adjustments to `javac.target` but not `javac.source`. This I don't understand - is the intention here to allow javac to generate more modern code, that uses features only present on Java 8+, but not suffer from code incompatibilites by using old source parser/behavior?

The value of `javac.target` influences not just generated bytecode, but also the API one compiles against (via the `--release` flag). I bumped `javac.target=1.8` with a `sed`-like utility. But that wasn't enough, so I also updated the default. The `javac.source` hasn't been changed as @jlahoda warned that it could change the semantics.

java/api.debugger.jpda/nbproject/project.properties
```diff
@@ -21,6 +21,7 @@ javac.compilerargs=-Xlint:unchecked
 javac.source=1.8
 javadoc.arch=${basedir}/arch.xml
 javadoc.apichanges=${basedir}/apichanges.xml
+bootclasspath.prepend=${build.dir}/disable-release
```
matthiasblaesing, line 24: I assume this line is intended to disable setting the release flag as the module links against internal classes?

profiler/profiler/nbproject/project.properties
```diff
@@ -20,6 +20,7 @@ is.autoload=true
 cp.extra=${tools.jar}
 
 javac.source=1.8
+bootclasspath.prepend=${build.dir}/disable-release
```
matthiasblaesing, line 23: I assume this is to disable setting "release" version as linking against com.sun classes happens.

matthiasblaesing (commented):
I left a few inline comments. In general this looks good. I don't see the benefit of bumping the `javac.target` without bumping `javac.source` as this will still fail hard on newer JDKs. For me this add noise without real benefit.

platform/o.n.core/test/qa-functional/src/org/netbeans/core/validation/ValidateClassFilesTest.java
```diff
@@ -0,0 +1,120 @@
...
+    public ValidateClassFilesTest(String name) {
+        super(name);
+    }
+
+    public static Test suite() {
+        return NbModuleSuite.createConfiguration(ValidateClassFilesTest.class).
+                clusters(".*").gui(false).honorAutoloadEager(true).suite();
+    }
+
+    public void testCheckClassFiles() throws IOException {
```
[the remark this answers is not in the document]
jtulach: Yes, the goal is to ensure class version is at most 52 (Java 8) unless otherwise stated. Right now there is an exception for `extra/modules/*javafx*.jar` which are only available/used on JDK11.

java/api.debugger.jpda/nbproject/project.properties
jtulach: Right: 94aebf0

profiler/profiler/nbproject/project.properties
jtulach: Right: 94aebf0

enterprise/web.monitor/build.xml
```diff
@@ -27,7 +27,7 @@
 
     <target name="compile-server" depends="init,compile">
         <mkdir dir="build/server-classes"/>
-        <javac srcdir="serversrc" destdir="build/server-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="${javac.source}" includeantruntime="false">
+        <javac srcdir="serversrc" destdir="build/server-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="${javac.source}" release="8" target="1.8" includeantruntime="false">
```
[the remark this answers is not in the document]
jtulach: Your concern may be valid, but the actual problem is the opposite. Make sure `nbjdk.home` isn't set and then try (on `12.3` without this PR):
```bash
enterprise/web.monitor$ JAVA_HOME=/jdk-8 ant clean compile-server
enterprise/web.monitor$ /jdk-11/bin/javap -v -cp build/server-classes/ org/netbeans/modules/web/monitor/server/Logger | grep -i Major -B 1
  minor version: 0
  major version: 52
```
e.g. in all version of Apache NetBeans up to 12.3 the http monitor could only be executed on JDK8+. Now, try to compile on JDK11:
```bash
enterprise/web.monitor$ JAVA_HOME=/jdk-11 ant -Dpermit.jdk9.builds=true clean compile-server
enterprise/web.monitor$ /jdk-11/bin/javap -v -cp build/server-classes/ org/netbeans/modules/web/monitor/server/Logger | grep -i Major -B 1
  minor version: 0
  major version: 55
```
See? If you compile on JDK11, the http monitor can only be executed on JDK11! According to documentation of [Ant Javac task](https://ant.apache.org/manual/Tasks/javac.html) it is highly desirable to set `target` - which is what this PR does.

Btw. I have only discovered this problem by introducing the 1067e68 test. E.g. the purpose of the `ValidateClassFilesTest` isn't to _bump the target up to 1.8_, but prevent us to accidentally compile with _target higher than expected_.

lkishalmi (approved):
Well, looks fine enough. I'd vote to have this in, if there would be some shortcomings detected we could fix that later. As of setting all javac.source to 1.8, AFAIK there could be some differences in spec between the spec, like the Override annotation in Java 5 and Java 6. So I'd leave that for another day.
BTW If you read that @lkishalmi decided in the comments, I'd mark that I do not decide, I merely suggest.

jtulach:
> BTW If you read that @lkishalmi decided in the comments, I'd mark that I do not decide, I merely suggest.

Changing directions when the code is ready is costly. As such I [discussed the approach](https://lists.apache.org/thread.html/r94d421ce16f609687c769125425b46f2322033879120b00afddbe17b%40%3Cdev.netbeans.apache.org%3E) first, listened to the suggestions and implemented what seemed to be a mutually agreeable intersection.

As soon as I started coding _the suggestions_ (hopefully correctly selected by me) have become _decisions_. I have attributed the _decisions_ to you guys to stress, that I did listen to your _suggestions_ before trying to move the project to the right direction without pushing it inappropriately too far.

ebresie:
Once this rolls out, recommend update the readme.md to account for changes (i.e. sounds like the jdk9 ant parameter not needed, minimum ant version, if any new parameters are needed, etc.)

neilcsmith-net:
> Once this rolls out, recommend update the readme.md to account for changes

Actually, that's a good point - that doc change might be better included in this PR?

ebresie:
As long as considering readme items...readme says needs "Ant 1.9.9 or above" but commit 140cb7e reference to ant 1.9.8.  Does the 1.9.8 need to be changed to 1.9.9 to be in line with the doc?

matthiasblaesing:
@ebresie I had a look at the history of the README and the ant version is there since the beginning. I also remember, that there were broken ant installations in various linux distributions, I would keep it as is. People reading the README have no problem, all people with recent ant version have no problem, only people on ant version which should never  have been used to build netbeans will see the message.

Lets get this in.

jtulach:
The 3695501 fix is needed to accommodate new tests added by #2731

lkishalmi:
Geat job! Thank you!

matthiasblaesing:
@jtulach @jlahoda I'm not sure, but I think I seen an issue with the module path in netbeans with this change (I don't think the change is broken, just another adjustment, that is needed).

I can build NetBeans with JDK 8 and JDK 11 and system ant just fine. So it is not a general problem, but when opening netbeans modules in the IDE, I see issues.

I run a build from current master and open the project `openide.util.lookup`. When the IDE is loaded on JDK 8, the modules can be rebuild using "Clean and Build" from the context menu. 

```
Compiling 31 source files to /home/matthias/src/netbeans/platform/openide.util.lookup/build/classes
Ignoring source, target and bootclasspath as release has been set
error: release version 8 not supported
Usage: javac <options> <source files>
use --help for a list of possible options
```

I tracked this down into the java compiler source in (I used git sources from jdk-11+28):

src/jdk.compiler/share/classes/com/sun/tools/javac/platform/PlatformUtils.java

If I understand it correctly, the list of PlatformProviders is empty, while it should at least provide `com.sun.tools.javac.platform.JDKPlatformProvider`.

I have some hope that one of you two has an idea how to fix this.

Rareitor:
On a related note, should the instructions for building from source on the website (https://netbeans.apache.org/download/dev/index.html) be updated to note that the -Dpermit.jdk9.builds=true flag is not needed anymore?
I did the last modification, but missed the chance to do it again.
It would be an update for the next release anyway, since it still needs the flag to build.

PR #2762, opened 2021-02-15
[NETBEANS-5353] Use cleanTest instead of rerun-tasks to force test to run

sdedic:
See [NETBEANS-5353](https://issues.apache.org/jira/browse/NETBEANS-5353) description; this PR attempts to use gradle's up-to-date checks so unnecessary compilations are avoided during test runs/debugs.

@lkishalmi  please review.

### DIFF
--- a/java/gradle.java/src/org/netbeans/modules/gradle/java/action-mapping.xml
+++ b/java/gradle.java/src/org/netbeans/modules/gradle/java/action-mapping.xml
@@ -23,16 +23,16 @@
 <actions>
     <apply-for plugins="java">
         <action name="test.single">
-            <args>--rerun-tasks test --tests "${selectedClass}"</args>
+            <args>cleanTest test --tests "${selectedClass}"</args>
         </action>
         <action name="run.single.method">
-            <args>--rerun-tasks test --tests "${selectedMethod}"</args>
+            <args>cleanTest test --tests "${selectedMethod}"</args>
         </action>
         <action name="debug.single.method">
-            <args>--rerun-tasks test --debug-jvm --tests "${selectedMethod}"</args>
+            <args>cleanTest test --debug-jvm --tests "${selectedMethod}"</args>
         </action>
         <action name="debug.test.single">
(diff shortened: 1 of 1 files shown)


lkishalmi (approved):
Seems to be good. I tried to look up since when Gradle recommends using the cleanTest task rule, but could not find the origin. Let's have this as default, and whoever is running a Gradle version where this would be not supported, he/she can still customize it.

sdedic:
@lkishalmi also not sure the **clean**_TaskName_ support is documented in Gradle 5.0, so assuming Test task has well-defined outputs (should have, for the results caching to work), **cleanTest** must have worked from that time as well.

JaroslavTulach (approved):

lkishalmi:
FYI. We do not use the output files from tests to display the results, but rather listening Gradle's test progress events.
Also further down the road I'd like to use the dedicated test executor functionality from Gradle, but maybe i won't get that far in this year...

sdedic:
> FYI. We do not use the output files from tests to display the results, but rather listening Gradle's test progress events.
> Also further down the road I'd like to use the dedicated test executor functionality from Gradle, but maybe i won't get that far in this year...

I admin I don't know gradle :) so maybe I guessed wrong: But the docs seemed to suggest that `cleanTest` purges results (outputs) of `test` task from gradle whatever storage, which subsequently leads to reexecution of `test` task (since gradle-cached outputs are gone). This seemed to be the case from early Gradle versions.

sdedic:
BTW - you were very quick ! with the review. In the meantime, I've discovered one situation my patch affects adversely - the project customizer has a two-state checkbox, which may lead to a temporarily-trusted project becoming permanently trusted after using project customizer.

I'll try to prepare another PR that makes that checkbox tri-state, the indeterminate state for a situation like "the trust is not permanent, but the buildscript is trusted somehow". The user then may leave it asi it is (no change), turn the trust to permanent, or indicate that trust should be revoked.

PR #2764, opened 2021-02-16
[NETBEANS-5318] Temporary script trust.

sdedic:
If the user acknowledges run of the gradle build script in the IDE, the script may be reloaded from time to time, for example as result of __Fix project problems__ or other occasions. If the user did not check "Trust permanently" in the question, the IDE would ask the user repeatedly even during the same IDE session.
This change in 932c71c will change the behaviour that after user's acknowledgement, the IDE trust the buildscript within this IDE run only - and won't ask again.

The other commits that were originally part of this PR were separated to PR #2765 per @lkishalmi request.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
@@ -26,8 +26,10 @@ import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
 import java.util.prefs.Preferences;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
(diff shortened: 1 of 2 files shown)


lkishalmi:
@sdedic Would I be rude if I'd ask to split this PR into two? One for the Temporal Project Trust, the other for the reload work?
As of #2747 I'm working on the project loading at the moment. We might collide there, but generally I'm for getting the internal GradleProject synchronized.

sdedic:
@lkishalmi No problem; later in the evening.

lkishalmi (approved):
Looks good.

sdedic:
BTW - you were very quick ! reviewing this PR :) In the meantime, I've discovered that I bugged Project Customizer: there's a checkbox, that gets turned on from the temporary trust, and after changes are written from the customizer, will promote the trust to permanent. 
I'll try to fix this by using a tri-state checkbox, the intedeterminate state reserved for "trusted, but not permanently" so the user can decide (permanent, revoke, no change).

extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
```diff
@@ -65,14 +72,31 @@
...
+
+    /**
+     * Returns true if the specified project is trusted <b>permanently</b>.
+     * You should be probably using {@link #isTrusted} to avoid duplicate questions
+     * during one IDE run.
+     *
+     * @param project of the trust check.
+     * @return true if the given project is trusted.
+     */
+    public boolean isTrustedPermanetly(Project project) {
```
neilcsmith-net, line 99: Just noticed this method name has a typo (it was highlighted while using `git show` :smile: ).

neilcsmith-net (commented):
Just noticed a method name typo.

PR #2765, opened 2021-02-16
[NETBEANS-5318] Prevent unnecessary Gradle model reloads

sdedic:
Originally submitted in PR #2764, but separated for easier integration into its own PR.

## Prevent unnecessary reloads
When the project problem is being fixed, the project model is loaded again, but then the model is thrown away and loaded once again (unnecessarily). In 22424fb I've changed the behaviour to reuse the model - if another action did not alter the model in the meantime.

## Synchronization fixes
The `project` variable is accessed from several threads, at least from RELOAD_RP and any random thread that calls `NbGradleProjectImpl` getters. I've tried to synchronize that so the reference is replaced, but the `loadProject` is still executed outside mutexes and events are fired in RELOAD_RP.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
@@ -188,7 +188,7 @@ public final class NbGradleProjectImpl implements Project {
         }
         p = loadProject();
         synchronized (this) {
-            // avod unnecessary project replacements.
+            // avoid unnecessary project replacements.
             if (project != null) {
                 return project;
             }
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
@@ -181,6 +181,45 @@ public final class NbGradleProjectImpl implements Project {
         }
         return project;
     }
+    
+    /**
+     * Records a reloaded project instance. Indicates whether the project was replaced.
+     * The method refuses to replace a project that has been changed since the reload procedure
+     * started. Will also not replace the project, if the original has a better quality
+     * than the reloaded one.
+     * <p>
+     * 
+     * @param original the original project data instance.
+     * @param reloaded the reloaded data.
+     * @param reloadIfMismatch will fire new reload if the project differs from original or quality is not satisfactory.
+     * @return {@code null}, if the project was accepted. {@code Boolean.TRUE}, if the project has
+     * been changed independently; {@code Boolean.FALSE}, if the reloaded project's quality is lower than
+     * the current.
+     */
+    Boolean recordOrReloadProject(GradleProject original, GradleProject reloaded, boolean reloadIfMismatch) {
+        Boolean result = null;
+        synchronized (this) {
+            if (project != original) {
+                result = true;
+            } else if (original.getQuality().betterThan(reloaded.getQuality())) {
+                result = false;
+            } else {
+                this.project = reloaded;
+            }
+        }
+        if (result != null) {
+            if (reloadIfMismatch) {
+                // will fire appropriate events after reload
+                fireProjectReload(false);
+            }
+        } else {
+            // notify listeners asynchronously as in fireProjectReload
+            RELOAD_RP.submit(() -> {
+                ACCESSOR.doFireReload(watcher);
+            });
+        }
+        return result;
+    }
 
     public void fireProjectReload(boolean wait) {
         reloadTask.schedule(0);
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
@@ -74,8 +74,7 @@ public final class NbGradleProjectImpl implements Project {
     private final RequestProcessor.Task reloadTask = RELOAD_RP.create(new Runnable() {
         @Override
         public void run() {
-            project = loadProject();
-            ACCESSOR.doFireReload(watcher);
+            recordOrReloadProject(null, loadProject(), false);
         }
     });
 
@@ -90,7 +89,8 @@ public final class NbGradleProjectImpl implements Project {
     @SuppressWarnings("MS_SHOULD_BE_FINAL")
     public static WatcherAccessor ACCESSOR = null;
 
-    GradleProject project;
+    // @GuardedBy(this)
+    private GradleProject project;
 
     static {
         // invokes static initializer of ModelHandle.class
@@ -106,7 +106,7 @@ public final class NbGradleProjectImpl implements Project {
     private final GradleFiles gradleFiles;
 
     public boolean isGradleProjectLoaded() {
-        return project != null;
+        return getProjectInternal() != null;
     }
 
     public static abstract class WatcherAccessor {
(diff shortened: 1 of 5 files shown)


extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -174,14 +174,76 @@ private Lookup createBasicLookup(ProjectState state, GradleAuxiliaryConfigImpl a
...
+     * 
+     * @param original the original project data instance.
+     * @param reloaded the reloaded data.
+     * @param reloadIfMismatch will fire new reload if the project differs from original or quality is not satisfactory.
+     * @return {@code null}, if the project was accepted. {@code Boolean.TRUE}, if the project has
+     * been changed independently; {@code Boolean.FALSE}, if the reloaded project's quality is lower than
+     * the current.
+     */
+    Boolean recordOrReloadProject(GradleProject original, GradleProject reloaded, boolean reloadIfMismatch) {
+        Boolean result = null;
```
JaroslavTulach, line 216: Why is `result` tri-state? Can it be named `mismatchOfOldAndNewProject`, right?

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -174,14 +174,76 @@ private Lookup createBasicLookup(ProjectState state, GradleAuxiliaryConfigImpl a
...
+        synchronized (this) {
+            if (original != null) {
+                if (project != original) {
+                    result = true;
+                } else if (original.getQuality().betterThan(reloaded.getQuality())) {
+                    result = false;
+                }
+            }
+            if (result == null) {
```
JaroslavTulach, line 225: this could be `} else if` and then `result` can be just plain `boolean` I think.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -174,14 +174,76 @@ private Lookup createBasicLookup(ProjectState state, GradleAuxiliaryConfigImpl a
...
+     * The method refuses to replace a project that has been changed since the reload procedure
+     * started. Will also not replace the project, if the original has a better quality
+     * than the reloaded one.
+     * <p>
+     * Pass {@code original = null} to just replace the project without checking.
+     * 
+     * @param original the original project data instance.
+     * @param reloaded the reloaded data.
+     * @param reloadIfMismatch will fire new reload if the project differs from original or quality is not satisfactory.
+     * @return {@code null}, if the project was accepted. {@code Boolean.TRUE}, if the project has
```
JaroslavTulach, line 211: Returned value seems to be unused in all invocations.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -220,7 +282,9 @@ void detachAllUpdater() {
     }
 
     void dumpProject() {
```
JaroslavTulach, line 284: dump!? clear? release?

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -174,14 +174,76 @@ private Lookup createBasicLookup(ProjectState state, GradleAuxiliaryConfigImpl a
...
     public GradleProject getGradleProject() {
-        if (project == null) {
-            project = loadProject();
+        GradleProject p = getProjectInternal();
+        if (p != null) {
+            return p;
+        }
+        p = loadProject();
+        synchronized (this) {
+            // avod unnecessary project replacements.
```
JaroslavTulach, line 191: typo

JaroslavTulach (approved):
Hard to say whether the new behavior is correct. It however seems better than the previous one.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -174,14 +174,76 @@ private Lookup createBasicLookup(ProjectState state, GradleAuxiliaryConfigImpl a
...
+     * Pass {@code original = null} to just replace the project without checking.
+     * 
+     * @param original the original project data instance.
+     * @param reloaded the reloaded data.
+     * @param reloadIfMismatch will fire new reload if the project differs from original or quality is not satisfactory.
+     * @return {@code null}, if the project was accepted. {@code Boolean.TRUE}, if the project has
+     * been changed independently; {@code Boolean.FALSE}, if the reloaded project's quality is lower than
+     * the current.
+     */
+    Boolean recordOrReloadProject(GradleProject original, GradleProject reloaded, boolean reloadIfMismatch) {
```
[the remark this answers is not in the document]
sdedic: possibly. I was never strong in naming things.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
sdedic: Originally I wanted to propagate the action to the caller: accepted (null), changed by other party (true), original valid, but replacement rejected (false).

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
sdedic: `dumpProject` name was not changed in this PR.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
sdedic: tri-state was explained below. If seems reasonable I would just leave it as it is; javadoc documents all three possible return values.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
sdedic: Fixed in 2177f44a16

sdedic:
.... 26 days without review from the component owner; I remember there's a work underway that may cause conflicts; clashes - any ETA where the competing PR could be merged so I can finish this one ?

lkishalmi (changes requested):
Well, I do not see answers on my comments, no new changes. Till I do not see the complete logic fits, I would not approve this PR.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
sdedic: OK, so degrading after reload is permitted. That's broken with this implementation - will revisit. Thanks.

sdedic: @lkishalmi I've checked the code again, recalling what I was thinking about that month ago: the code was written in such a way, as there can be multiple `loadProject` calls (potentialy) that would lead to `impl.project` replacement. The logic in the `recordOrReloadProject` basically ensures that if a project has been replaced **since the loadProject started** (the caller needs to save a reference), the `impl.project` will not be replaced for a worse version. But this is just for **competing** reloads.

Imagine a user clicking on Resolve problems, which will not make the project just `FULL` (whatever reason), but at the same time the background reload task will load the project and reaches `FULL_ONLINE`. If the final calls to `recordOrReloadProject` are reordered (different parallel threads) in an unfortunate sequence, the FULL_ONLINE read `impl.project` first, then it would degrade to FULL unnecessarily. The method's logic prevents that: if two replace processes are competing (caller has recorded `project` instance and passed it), the 'better of them' wins. Similar to AtomicReference.compareAndSet().

To be conservative, I've enabled this logic just from `GradleProjectProblemProvider` call. The other invocations pass `null` for `original`, so the comparison is disabled. So Reload action, reload based on desired quality change will always replace the project: I did not want to affect those occurrences, since I don't fully understand their semantics. I thought that it would not be good if "priming build" actually degraded already well-loaded project.

We can chat over the IM as you suggested or continue writing here. The logic to solve competing operations may be unnecessary.

sdedic:
Changes in this PR became obsolete after #2747 - must be redone from scratch.

PR #2766, opened 2021-02-17
Support for visualizing test results in VSCode extension over LSP.

dbalek:
- Common base class for handlers displaying test results (`org.netbeans.modules.gsf.testrunner.ui.api.TestResultDisplayHandler`) created.
- Its subclass (`org.netbeans.modules.java.lsp.server.progress.TestProgressHandler`) created that sends the test results over LSP to client.
- `Test Explorer UI` VSCode extension is used on the client side to display test results.


ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/api/TestResultDisplayHandler.java
```diff
@@ -0,0 +1,101 @@
...
+     *
+     * @return number of tests
+     */
+    public abstract int getTotalTests();
+
+    /**
+     * Interface providing factory method for creating {@link TestResultDisplayHandler}s.
+     * Instances should be registered in the default lookup.
+     */
+    public static interface Provider {
```
JaroslavTulach, line 98: How do you plan to evolve these two types?

I'd suggest to use [singletonizer pattern](http://wiki.apidesign.org/wiki/Singletonizer). E.g.
* make `TestResultDisplayHandler` constructor package private
* add package private `class Impl<T> extends TestResultDisplayHandler` that will have two fields
* `final T data` and `final Provider<T> provider`
* Add parameter `<T>` to `Provider` and let the `create` method `T`
* for each method in `TestResultDisplayHandler` method create the same method in `Provider` with first parameter being `T`.

Here is a simplified example:
```java
public static interface Provider<T> {
  T create(TestSession session);
  int getTotalTests(T data);
}
```
the `Impl` class should then have
```java
class Impl<T> extends TestResultDisplayHandler {
  private final T data;
  private final Provider<T> provider;

  public int getTotalTests() { return provider.getTotalTests(data); }

  static <T> Impl create(Provider<T> p, TestSession s) {
    return new Impl(p, p.create(s));
  }
}
```
the `get` method in the `TestResultDisplayHandler` would then become:
```java
public static final TestResultDisplayHandler get(TestSession session) {
       TestResultDisplayHandler.Provider provider = Lookup.getDefault().lookup(TestResultDisplayHandler.Provider.class);
        if (provider != null) {
            return Impl.create(provider, session);
        }
        return Impl.create(new ResultDisplayHandler());
    }
```

Evolution story would then be:
- add new method into the `TestResultDisplayHandler` (safe as the class isn't publicly subclassable)
- add another `Provider2 extends Provider` SPI interface with the new method(s)

ide/gsf.testrunner/manifest.mf
```diff
@@ -3,5 +3,5 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.gsf.testrunner/2
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/gsf/testrunner/Bundle.properties
 OpenIDE-Module-Layer: org/netbeans/modules/gsf/testrunner/layer.xml
-OpenIDE-Module-Specification-Version: 2.21
+OpenIDE-Module-Specification-Version: 2.22
```
JaroslavTulach, line 6: Please make a note in the `apichanges.xml` file.

ide/gsf.testrunner/nbproject/project.xml
```diff
@@ -79,6 +79,7 @@
                 <friend>org.netbeans.modules.groovy.support</friend>
                 <friend>org.netbeans.modules.gsf.testrunner.ui</friend>
                 <friend>org.netbeans.modules.hudson.ui</friend>
+                <friend>org.netbeans.modules.java.lsp.server</friend>
```
JaroslavTulach, line 82: I suggest to open up the API to public rather than adding yet another friend.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/progress/TestProgressHandler.java
```diff
@@ -0,0 +1,140 @@
...
+import org.netbeans.modules.java.lsp.server.protocol.NbCodeLanguageClient;
+import org.netbeans.modules.java.lsp.server.protocol.TestProgressParams;
+import org.netbeans.modules.java.lsp.server.protocol.TestSuiteInfo;
+import org.openide.filesystems.FileObject;
+
+/**
+ *
+ * @author Dusan Balek
+ */
+public final class TestProgressHandler extends TestResultDisplayHandler implements TestResultDisplayHandler.Provider {
```
JaroslavTulach, line 41: If the _singletonizer_ idea is implemented, this could become `class TestProgressHandler implements TestResultDisplayHandler.Provider<TestProgressHandler>`, but it is a bit strange that `create` method returns `this`! Why do we have `Provider` factory pattern then?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -452,6 +451,11 @@ public void showStatusBarMessage(ShowStatusMessageParams params) {
             return CompletableFuture.completedFuture(params.getValue());
         }
 
+        @Override
+        public void notifyTestProgress(TestProgressParams params) {
+            logWarning(params);
+        }
+
         @Override
         public NbCodeClientCapabilities getNbCodeCapabilities() {
```
JaroslavTulach, line 460: Shouldn't the additional messages in the LSP protocol be reflected in the capabilities? CCing @sdedic ...

JaroslavTulach (commented):
I think this is great work. I left few comments for your consideration.

ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/api/TestResultDisplayHandler.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public abstract class TestResultDisplayHandler {
+
+    /**
+     * Get the {@link TestResultDisplayHandler} for the test session.
+     *
+     * @param session test session
+     * @return {@link TestResultDisplayHandler} instance
+     */
+    public static final TestResultDisplayHandler get(TestSession session) {
```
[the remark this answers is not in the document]
dbalek: This method should create TestResultDisplayHandler for each test session. Going to rename it to `create`.

ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/api/TestResultDisplayHandler.java
dbalek: Changed as requested.

ide/gsf.testrunner/manifest.mf
dbalek: Done.

ide/gsf.testrunner/nbproject/project.xml
dbalek: Done.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
dbalek: Done. Thanks for pointing this out.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/progress/TestProgressHandler.java
dbalek: Default `ResultDisplayHandler` needs to be created per test session to display test results in different output tabs - that is why the factory pattern is needed. However, the LSP implementation crates `TestResultDisplayHandler.Spi` instance per client test and puts in default lookup for the request execution.

JaroslavTulach (approved):
Thanks for the [singletonizer](http://wiki.apidesign.org/wiki/Singletonizer) changes and opening up the module to everyone.

ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/api/TestMethodController.java
```diff
@@ -36,16 +36,22 @@ public static void setTestMethods(Document doc, List<TestMethod> methods) {
     }
 
     public static final class TestMethod {
+        private final String testClassName;
         private final SingleMethod method;
         private final Position start;
         private final Position end;
 
-        public TestMethod(SingleMethod method, Position start, Position end) {
+        public TestMethod(String testClassName, SingleMethod method, Position start, Position end) {
```
JaroslavTulach, line 44: Looks like this is causing a [sigtest failure](https://github.com/apache/netbeans/pull/2788#issuecomment-791267713) - probably return the original constructor back as part of #2788

PR #2767, opened 2021-02-17
Broken project auto-resolution repeats actions

sdedic:
During LSP testing with gradle, I've found out that if problem resolution fails, the action may be repeated again. Along the way, I've also noticed that the LSP `DialogDisplayer` implementation does not stip `&` that mark mnemonic characters in String options (choices) - LSP protocol does not support such pattern (yet).

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/NotifyDescriptorAdapter.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/NotifyDescriptorAdapter.java
@@ -37,13 +37,13 @@ import java.util.stream.Collectors;
 import javax.accessibility.Accessible;
 import javax.accessibility.AccessibleContext;
 import javax.accessibility.AccessibleIcon;
-import javax.accessibility.AccessibleText;
 import javax.swing.Icon;
 import javax.swing.JButton;
 import org.eclipse.lsp4j.MessageActionItem;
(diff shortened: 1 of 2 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/NotifyDescriptorAdapter.java
```diff
@@ -222,6 +222,11 @@ private void mapDescriptorOptions() {
                 text = mapDescriptorOption(o);
             }
             if (text != null) {
+                // Discard ampersands, LSP spec does not support such pattern.
+                int mn = Mnemonics.findMnemonicAmpersand(text);
```
JaroslavTulach, line 226: There is [cutAmpersand](https://bits.netbeans.org/12.0/javadoc/org-openide-awt/org/openide/awt/Actions.html#cutAmpersand-java.lang.String-) method.

JaroslavTulach (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/NotifyDescriptorAdapter.java
sdedic: `cutAmpersand` is better (handles japanese locale). Thanks ! Fixed in 17ddace

sdedic:
Squashed the review commit(s)

PR #2769, opened 2021-02-17
NonUI compatibility: Trust project question uses standard dialogs.

sdedic:
The current dialog uses a custom layout w/ checkbox. The LSP client has some issues :) with it (can't properly discover / display the option ;)) so I propose to change the dialog to a standard `NotifyDescriptor.Confirmation`. 

The wording remains mostly the same, but there are 3 closing buttons + cancel. This dialog can be remoted to LSP or processed by other `DialogDisplayer` implementations. 

I've added one more option: Trust All, which changes the Gradle execution policy to `run always`.

Please review / comment.


lkishalmi:
Can't we just detect LSP (or headlessness), then use a different dialog in that case?

sdedic:
I had two other solutions before I've decided to take this shortcut:
- use "branding API" to modify the behavior. But I would have to code the variants (permitted actions) in Gradle anyway, so it's not less complicated than the current
- introduce a TrustProvider SPI (instances placed in default Lookup), the current `master` implementation being the default if no Provider decided otherwise. The decisions could be 'trust-persist', 'trust-session', 'abstain' and 'reject'. Gradle core would then take care of the persistent decisions as it does now. LSP can register its own provider and provide the alternative implementation. Would be slightly complicated as I'd probably need to introduce a Gradle integration module (so the base LSP does not bloat with dependencies), but doable

sdedic:
Since no feedback appeared since 22th Feb 2021, I followed @lkishalmi advice to keep trust management strictly inside the module. I've changed the dialog implementation to use 'branding API' that eventually allows to expose "Trust all Gradle" for a distribution. 
The behaviour in Apache NetBeans is intended to stay as it was (OK (=default), Permanent, Cancel).

Please review.

sdedic:
Ping ... this PR is over a week without response or review, 3 weeks from latest input that has been processed. @lkishalmi, @JaroslavTulach

JaroslavTulach (approved):

lkishalmi (approved):
Let's have this in.

PR #2770, opened 2021-02-17
Sync delivery to master post 12.3-beta3 with conflicts resolved

neilcsmith-net:
This replaces #2749 for merging delivery to master resolving conflicts in the files - 

```
java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java 
```

As @matthiasblaesing noticed, using git's merge resolution results in an uncompileable state, so both these files were retained as is from the master branch.


entlicher:
The two files needs to be merged and not replaced. If this is merged as it is, I can follow-up with a PR that merges these two files.

entlicher (approved):

matthiasblaesing:
I would merge as is, as @entlicher already indicated, that he'll "fix" the merged variant.

neilcsmith-net:
OK, merging, thanks.  Tests all OK.  @entlicher - for info, I left as is in master, so they weren't replaced by the delivery branch - not sure there is anything to merge in, but also didn't look in depth why the subsequent PR to delivery had removed those methods.

PR #2771, opened 2021-02-18
Completing the merge of delivery to master.

entlicher:
Completing the merge of https://github.com/apache/netbeans/pull/2770/commits/cf30670c24d446134b12f32efa21a21ab0896fe0 into the master.


neilcsmith-net (approved):
Thanks for looking at that - bit awkward from the outside to work out what exactly was intended to be kept.  Let's get it merged, and hopefully we're done syncing the two branches now!

entlicher:
Right. It was not apparent what's the correct merge as I was deleting this stuff on the delivery, since it was replaced with a different approach.
The merge should be fully done now. Thanks.

PR #2772, opened 2021-02-18
[NETBEANS-5367] Changed PHP Annotation API to public

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5367

First step to be able to implement annotations in module not listed in friend packages.

### DIFF
--- a/nbbuild/build.properties
+++ b/nbbuild/build.properties
@@ -155,7 +155,8 @@ config.javadoc.stable=\
     print,\
     extexecution,\
     lib.uihandler,\
-    uihandler
+    uihandler,\
+    php.api.annotation
 
 #FIXME: changes for retouche merge:
 #   editor/codetemplates,\
--- a/php/php.api.annotation/apichanges.xml
+++ b/php/php.api.annotation/apichanges.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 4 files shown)


geertjanw:
Is there any objection to this, @junichi11?

junichi11:
@tmysik Is this OK?

tmysik:
IMHO, one should be *really* sure that the API is good (usable and stable, at least) before making it public. Also, it seems to me quite unexpected to make the first public PHP API from this module exactly... But I won't be blocking this PR, of course; however, I will not approve it either.

@jtulach 
It would be definitely good to know what Jarda thinks about this. Thanks!

KacerCZ:
My goal is to implement PHPStan annotations in "php.code.analysis" module (because this module implements PHPStan support).
PHP Annotations API can be made public (as I understand it is prefered for Java APIs) or "php.code.analysis" module will need to become friend.
I'd like to know what NetBeans veterans think is better.

matthiasblaesing:
I know that everyone says friend modules are evil, but I see only two options:

- accept that fact or
- be willing to break API

I think both options should be considered (yes, there are reasons API should be broken) - in this case we are late in the cycle, so why not add a friend and plan for publication of the API in the next cycle?

tmysik:
Definitely a friend, IMHO. Exactly the same as for all the other PHP modules. Making any API public really needs to be planned and reviewed properly, it is a big commitment.

neilcsmith-net:
They're not evil, just should never have been used to mark unstable APIs IMO.  (JDK has right approach on opt-in).

However, agree with adding a friend on a one release basis.  Push this to 12.5 for review.

KacerCZ:
OK, I'll make it friend module.
Thanks for your time.

KacerCZ:
@mvorisek Support for PHPStan annotations is in progress and is separate from array format.
Relevant issue is https://issues.apache.org/jira/browse/NETBEANS-4983

PR #2773, opened 2021-02-19
VSCode BUILD.MD update for production build.

MartinBalin:
Correction of how to build for release etc. with prerequisities.

### DIFF
--- a/java/java.lsp.server/vscode/BUILD.md
+++ b/java/java.lsp.server/vscode/BUILD.md
@@ -20,6 +20,13 @@
     under the License.
 
 -->
+## Prerequisities
+It is necessary to have installed:
+- Ant, latest version
+- Maven, latest version
+- node.js, latest LTS (to build VSIX)
+
+It is recommended to build using JDK 8.
 
 ## Getting the Code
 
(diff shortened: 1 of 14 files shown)


dbalek (approved):
Looks fine.

java/java.lsp.server/vscode/BUILD.md
```diff
@@ -35,10 +42,9 @@ To build the VS Code extension invoke:
 ```bash
 netbeans$ ant build
 netbeans$ cd java/java.lsp.server
-java.lsp.server$ ant build-vscode-ext -D3rdparty.modules=.*nbjavac.*
+java.lsp.server$ ant build-vscode-ext -Dvsix.version=x.y.z
 ```
-
-The `3rdparty.modules` property doesn't have to be set at all.
```
jtulach, line 41: Either the `3rdparty.modules` property remain to be documented, or it shall be removed. If it doesn't work with `m2://` protocol (which it didn't), that should be fixed.

JaroslavTulach:
#2777 makes the `-D3rdparty.modules=.*nbjavac.*` work again. We don't have to promote it as the default option, but I'd suggest to still mention it in the readme. It is the way to make the extension work on JDK8.

MartinBalin:
Options better documented now in 2nd commit.

jtulach:
OK. Just squash the commits into one.

PR #2774, opened 2021-02-22
[GR-29499] Fixing bug in updaing the list of discovered tests.

dbalek:
- Fixing bug in updaing the list of discovered tests.
- Fixing fails in ServerTest

### DIFF
--- a/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
+++ b/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
@@ -720,6 +720,7 @@ public class ServerTest extends NbTestCase {
     public void testGoToDefinition() throws Exception {
         File src = new File(getWorkDir(), "Test.java");
         src.getParentFile().mkdirs();
+        try (Writer w = new FileWriter(new File(src.getParentFile(), ".test-project"))) {}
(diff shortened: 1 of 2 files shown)


java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -750,7 +752,12 @@ public void publishDiagnostics(PublishDiagnosticsParams params) {
             }
 
             @Override
-            public void showMessage(MessageParams arg0) {
+            public void showMessage(MessageParams params) {
+                if (Server.INDEXING_COMPLETED.equals(params.getMessage())) {
+                    indexingComplete.countDown();
+                } else {
+                    throw new UnsupportedOperationException("Unexpected message.");
```
sdedic, line 759: Is it necessary to be this strict ? If some code attempts to e.g. `StatusDisplayer` just to note something, this test will fail.

sdedic (approved):

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
dbalek: It behaves in the same way as the rest of the tests here.

PR #2776, opened 2021-02-22
[NETBEANS-5373] Fixed compatibility with Phing 3

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-5373

Removed parameter `-logger` because its value is incompatible with Phing 3.

### DIFF
--- a/php/php.phing/src/org/netbeans/modules/php/phing/exec/PhingExecutable.java
+++ b/php/php.phing/src/org/netbeans/modules/php/phing/exec/PhingExecutable.java
@@ -63,8 +63,6 @@ public class PhingExecutable {
     @org.netbeans.api.annotations.common.SuppressWarnings(value = "MS_MUTABLE_ARRAY", justification = "No need to worry, noone will change it") // NOI18N
     public static final String[] PHING_NAMES;
 
-    private static final String LOGGER_PARAM = "-logger"; // NOI18N
-    private static final String LOGGER_COLOR_VALUE = "phing.listener.AnsiColorLogger"; // NOI18N
     private static final String LIST_PARAM = "-list"; // NOI18N
     private static final String QUIET_PARAM = "-quiet"; // NOI18N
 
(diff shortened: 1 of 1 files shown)


php/php.phing/src/org/netbeans/modules/php/phing/exec/PhingExecutable.java
```diff
@@ -163,10 +161,7 @@ private PhpExecutable getExecutable(String title) {
     }
 
     private List<String> getRunParams(String[] args) {
-        List<String> params = new ArrayList<>(args.length + 2);
-        params.add(LOGGER_PARAM);
-        params.add(LOGGER_COLOR_VALUE);
-        params.addAll(Arrays.asList(args));
+        List<String> params = Collections.unmodifiableList(Arrays.asList(args));
```
matthiasblaesing, line 164: this changes the semantics of getRunParams - before the change the returned list was modifiable, now its not anymore.

KacerCZ: I changed it to normal list.

php/php.phing/src/org/netbeans/modules/php/phing/exec/PhingExecutable.java
```diff
@@ -163,10 +161,7 @@ private PhpExecutable getExecutable(String title) {
     }
 
     private List<String> getRunParams(String[] args) {
-        List<String> params = new ArrayList<>(args.length + 2);
-        params.add(LOGGER_PARAM);
-        params.add(LOGGER_COLOR_VALUE);
-        params.addAll(Arrays.asList(args));
+        List<String> params = Arrays.asList(args);
```
matthiasblaesing, line 164: Sorry, still different semantics - before you got a `java.util.ArrayList`, after the change it will be `java.util.Arrays.ArrayList`. The latter is modifiable, but has fixed length, which is not true for the first.

KacerCZ: Hopefully it is correct this time.

matthiasblaesing:
Looks good.

KacerCZ:
Thank you for review.

PR #2777, opened 2021-02-23
Let AutoUpdate task recognize m2:/ URL scheme

JaroslavTulach:
Let `AutoUpdate` task recognize m2:/ URL scheme, so VSCode build can install additional nbjavac modules from Maven.

The [BUILD.md](https://github.com/apache/netbeans/blob/234392d4fc125d4e20a116598821b2cfd378dfc6/java/java.lsp.server/vscode/BUILD.md) file suggests to use:
```bash
java/java.lsp.server$ ant build-lsp-server -D3rdparty.modules=.*nbjavac.*
```
to build the LSP server with additionally installed `nbjavac` modules. The download no longer works since #2759 - the `AutoUpdate` task hasn't be updated to understand the `m2:/` protocol. This PR:
* calls from `AutoUpdate` task to `DownloadBinaries` to handle the `m2:/` protocol
* modifies the travis check to use the `-D3rdparty.modules` property, so we have some test coverage
* the [release job](https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/) **is not** modified, ASF doesn't want to distribute nbjavac right now
* the `m2:/` URL definitions had to be adjusted to [gradle scheme](https://github.com/apache/netbeans/blob/234392d4fc125d4e20a116598821b2cfd378dfc6/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java#L424)

If this PR is accepted, then #2773 shall be adjusted accordingly.

PR #2778, opened 2021-02-23
Restore original format of m2:/ URLs modified by aa256463dcefb3b95bec6b7ac74e04a960015bc9

matthiasblaesing:
m2:/ became a defacto API for the autoupdate mechanism. Both OpenJFX and
nb-javac rely on being able to be downloaded via the ".external" files
mechanism and third party modules could also rely on the exact syntax
of the URLs.

Instead of modifying the URL format, a new ad-hoc parser is introduced
to convert the m2 URL into maven coordinates.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
@@ -35,13 +35,11 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
-import java.nio.file.Files;
-import java.nio.file.StandardOpenOption;
+import java.net.URLDecoder;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Consumer;
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.Project;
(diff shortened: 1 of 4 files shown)


matthiasblaesing:
@JaroslavTulach could you please have a look at this change?

JaroslavTulach (approved):
Yup. Extending the parser to handle `m2:/` format was the second option for #2777 - but then I decided to prefer using the same format for `binaries-files` as well as `m2:/`. It did seem to work and was enough for the `-D3rdparty.modules=*.nbjavac.*` use-case.

neilcsmith-net:
+1 to this (and/or -1 to #2777 )

> but then I decided to prefer using the same format for binaries-files as well as m2:/.

I don't understand this.  `m2:/` had a format?  eg. as specified at https://github.com/apache/netbeans/blob/master/java/maven.embedder/src/org/netbeans/modules/maven/embedder/impl/MavenProtocolHandler.java and (later) by https://github.com/apache/netbeans/blob/master/platform/core.startup/src/org/netbeans/core/startup/MavenRepoURLHandler.java

> a new ad-hoc parser is introduced to convert the m2 URL into maven coordinates.

Well, I guess would be nice if it wasn't parsed in slightly different ways in multiple places, but I'm not sure there's a way around that?!  Then again, shame we're not using an external format like Package URL too ... :shrug:

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
```diff
@@ -478,6 +472,47 @@ public static MavenCoordinate fromGradleFormat(String gradleFormat) {
             }
             return new MavenCoordinate(group, artifact, version, extension, classifier);
         }
+
+        /**
+         * The maven coordinate is supplied in the form:
+         *
+         * <p>{@code m2:/group:name:version:extension:classifier}</p>
+         *
+         * <p>For the DownloadBinaries task the parts group, name and version
```
neilcsmith-net, line 481: Not sure what the "canonical" format is, but should extension be required, as in https://github.com/apache/netbeans/blob/master/java/maven.embedder/src/org/netbeans/modules/maven/embedder/impl/MavenProtocolHandler.java ?

matthiasblaesing:
- Travis is happy
- github actions mostly (the PHP cluster test runs for > 1 hour), as nothing in the PHP area depends on external urls, I consider this good enough
- the parser was modified, to also require the extension

PR #2779, opened 2021-02-24
[NETBEANS-5141] Support Gradle Source Groups from Alien Projects

lkishalmi:
This change allows Gradle Source groups to display a source group from another project ownership. Previously an error node was created in this case, as having such entities in a project is generally a bad practice. With this change such source groups are coming with an error badge and some warning text.
I was thinking about putting this under an experimental feature flag, though I hope the number of such Gradle projects are really low.


lkishalmi:
Just to see what's changed:
image

The new behavior:
image

PR #2781, opened 2021-02-25
The BindingPatternTree API has been updated, reflecting the change.

jlahoda:
The API for BindingPatternTree was not stable before JDK 16, and was changed in JDK 16, need to update NetBeans to work with the new API.


jlahoda:
Any objections against this change?

JaroslavTulach (approved):
I've seen these problems in #2801, please integrate so Toni & co. can merge these changes in.

PR #2784, opened 2021-02-27
Register JDKPlatformProvider when creating javac in ant's ClassLoader.

jlahoda:
As noted here:
https://github.com/apache/netbeans/pull/2761#issuecomment-787094888

When compiling using `ant` inside the IDE, `--release` does not work properly. The reason is that `javac` for the compilation is load using a (somewhat weird) ClassLoader magic, and is missing the registration for `JDKPlatformProvider`. The registration is normally in module-info, but the ClassLoader magic load javac in an unnamed module, and that requires META-INF/services registration.

The proposed patch here is only to inject the registration. Eventually, we may want to try a simpler way to use `javac` from `ant` while running inside the IDE.

### DIFF
--- a/extide/o.apache.tools.ant.module/src/org/apache/tools/ant/module/bridge/AntBridge.java
+++ b/extide/o.apache.tools.ant.module/src/org/apache/tools/ant/module/bridge/AntBridge.java
@@ -28,6 +28,7 @@ import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


JaroslavTulach (approved):
Thanks a lot for identifying the root cause of the failure! With the herein proposed patch I am able to compile `java/java.hints` when running the IDE on JDK15.

matthiasblaesing (approved):
Looks good to me. I also verified, that the new test fails without the fix and runs clean with it. Thank you!

PR #2785, opened 2021-02-28
Java chained types improvement feb 2020

jlahoda:
As noted here:
https://github.com/apache/netbeans/pull/2365#issuecomment-763579047

When the Java editor shows return types for chained method invocations, it could possibly show them in additional contexts. This PR is trying to show the return types in most/all contexts.

chained-calls

Feedback/opinions are welcome!

Thanks!

### DIFF
--- a/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
+++ b/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
@@ -61,6 +61,7 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.swing.text.Document;
 import org.netbeans.api.java.lexer.JavaTokenId;
@@ -223,7 +224,9 @@ public abstract class SemanticHighlighterBase extends JavaParserResultTask {
             return true;
         
         if (computeUnusedImports) {
-            setter.setHighlights(doc, extraColoring, v.preText);
+            Map<int[], String> preTextWithSpans = new HashMap<>();
+            v.preText.forEach((pos, text) -> preTextWithSpans.put(new int[] {pos, pos + 1}, text));
+            setter.setHighlights(doc, extraColoring, preTextWithSpans);
         }
 
         setter.setColorings(doc, newColoring);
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
I think this is a nice update to the parameter name annotations. I'm not sure, whether I prefer the types at the end or if it would be easier to follow if the return types would be showed inline between the braces.

jlahoda:
Thanks for the comment, @matthiasblaesing . I've tried to sketch putting the inline types after the braces (which, I think, also allows to eliminate some that are probably not so interesting, as they are not part of chained method invocation). I've put that here:
https://github.com/jlahoda/netbeans/tree/java-chained-types-improvement-feb-2020-var2

A sample screenshot:
chained-types-var2

What do you think?

Thanks!

geertjanw:
How ready is this for 12.4?

jlahoda:
@geertjanw - I think the main task here is to decide which of the variants we want to follow (or if there is a different variant we want to follow). The code is hopefully more or less fine.

geertjanw:
Should this go in for 12.4 or should we delay for 12.5? I.e., @jlahoda, feel free to make a choice above and just let's go with that.

jlahoda:
Let's wait, I guess.

matthiasblaesing:
Sorry for the late reply. I had another look at this tonight and found some issues:

It looks as if the return type output only works once:

Bildschirmfoto von 2021-04-16 20-48-05

For the second prototype, I like the inline display, but the alignment looks strange (left Marker) and it seems the intermediate returns displayed in the first prototype (above) are missing here (right marker):

image

Thinking over this, having the types at the end of the line has less impact on the layout, though the parameters already take size.

One thing I noticed with the cursor handling is, that it now feels a bit strange with the types:

image

At this point I'd have expected the cursor behind the closing brace, not behind the type. Just an idea: place the cursor on the side of the type/parameter signature, that has no whitespace?

ebarboni:
@matthiasblaesing  @jlahoda, can we merge this. And later make UI improvement or the current is not usable.

jlahoda:
@ebarboni - I think the existing state is acceptable, so this should only be merged when we are satisfied with the result. I apologize I didn't have time to look through Matthias' comments yet.

ebarboni:
ok moving to 12.6

netbeansuser2019:
Nothing against that inline hints, but sometimes it takes to much space on screen so need scrolling.
Maybe one more suggestion if it could not be possible to be added one extra option to view menu to show only on active e.g. when (typing) cursor is in "chained tree" and mouse is over it.

mbien:
eagerly waiting for the feb 2022 version

matthiasblaesing:
Merged via https://github.com/apache/netbeans/pull/5013

PR #2789, opened 2021-03-02
[NETBEANS-5380] Override generated ant projects for MX with NetBeans support

matthiasblaesing:
In profiles from repositry updater runs a huge CPU load originating 
from SuiteFileOwnerQueryImpl was found. In parallel excessive runtimes
for project scanning were observed.

Comparing SuiteFileOwnerQueryImpl and SimpleFileOwnerQueryImplementation
the latter has an aggressive caching system, which results in much less
cpu usage. Just removing SuiteFileOwnerQueryImpl is not a solution, as
it is used to override the default project detection from
SimpleFileOwnerQueryImplementation.

Explanation from Jaroslav Tulach:

[...] Historically the Graal project provided mx netbeansinit support
which generates 8.2 Ant nbproject/project.xml projects based on content
of suite/mx.suite/suite.py. They were usually located in
suite/src/*/nbproject directories.

To fix this, the idea is, not only to use the current MX project
detection (it requires a folder `mx.proj` to be found in the project
folder `proj`), but also look two level ups from the current directory.
If that directory is recognised as a MX project, that project will
be returned as the project for the requested directory.


matthiasblaesing:
@JaroslavTulach thank you for your check and explanation in https://issues.apache.org/jira/browse/NETBEANS-5380. This is a PR, that brings in the changeset, that was referenced in the comment of the mentioned issue.

JaroslavTulach (approved):
Thanks a lot for finding this way of eliminating global `SuiteFileOwnerQueryImpl`.

matthiasblaesing:
I'll let travis and github actions do their job and if everything comes back green, I'll merge.

PR #2790, opened 2021-03-02
fix travis link in readme

premek:


neilcsmith-net:
Thanks!  A missing part of the Travis transition.

PR #2791, opened 2021-03-04
[NETBEANS:4447]:Fixed NullPointerException while computing switch expression 

mishrasandeep:
Fixed Formatting issues for the Hint SwitchToSwitchExpression and added test cases for the same.

### DIFF
--- a/java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
+++ b/java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
@@ -875,6 +875,71 @@ public class ConvertSwitchToRuleSwitchTest extends NbTestCase {
                               "     }\n" +
                               "}\n");
     }
+    
+    public void testSwitchToRuleSwitchFormattingMultiple() throws Exception {
+        if(!ConvertSwitchToRuleSwitchTest.isJDK14())
+            return;
+        HintTest.create()
+                .input("package test;" +
+                        "public class Test {\n" +
+                        "     private void test(int p) {\n" +
+                        "         String result;\n" +
+                        "         switch (p) {\n" +
+                        "            case 0:\n" +
+                        "            case 1:\n" +
+                        "            case 2:\n" +
+                        "            case 3: result=\"a\"; break;\n" +
+                        "            default: System.err.println(\"No.\"); break;" +
+                        "         }\n" +
+                        "     }\n" +
+                        "}\n")
+                .sourceLevel(SourceVersion.latest().name())
+                .run(ConvertSwitchToRuleSwitch.class)
+                .findWarning("3:9-3:15:verifier:" + Bundle.ERR_ConvertSwitchToRuleSwitch())
+                .applyFix()
+                .assertCompilable()
+                .assertVerbatimOutput("package test;" +
+                        "public class Test {\n" +
+                        "     private void test(int p) {\n" +
+                        "         String result;\n" +
+                        "         switch (p) {\n" +
+                        "            case 0, 1, 2, 3 -> result=\"a\";\n" +
+                        "            default -> System.err.println(\"No.\");\n" +
+                        "         }\n" +
+                        "     }\n" +
+                        "}\n");
+    }
+    
+    public void testSwitchToRuleSwitchFormattingSimple() throws Exception {
+        if(!ConvertSwitchToRuleSwitchTest.isJDK14())
+            return;
+        HintTest.create()
+                .input("package test;" +
+                        "public class Test {\n" +
+                        "     private void test(int p) {\n" +
+                        "         String result;\n" +
+                        "         switch (p) {\n" +
+                        "            case 0: result = \"a\"; break;\n" +
+                        "            default: System.err.println(\"No.\"); break;\n" +
+                        "         }\n" +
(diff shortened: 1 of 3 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
```diff
@@ -4151,7 +4164,7 @@ private int diffList(
                         match = wasInFieldGroup;
                         for (Iterator<JCTree> it = deletedItems.iterator(); !match && it.hasNext(); ) {
                             ld = it.next();
-                            if (match == treesMatch(item.element, ld, false)) {
+                            if (match = treesMatch(item.element, ld, false)) {
```
sdedic: Since this expression seems confusing, I'd recomment to make a comment note that the assignment is really meant :)

sdedic (approved):

java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
mishrasandeep: Thanks for the input. Have made the change.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
```diff
@@ -875,6 +875,73 @@ public void testSwitch2SwitchExpressionNestedInnerSwitchStatement() throws Excep
...
+                        "public class Test {\n" +
+                        "     private void test(int p) {\n" +
+                        "         String result;\n" +
+                        "         switch (p) {\n" +
+                        "            case 0, 1, 2, 3 -> result=\"a\";\n" +
+                        "            default -> System.err.println(\"No.\");\n" +
+                        "         }\n" +
+                        "     }\n" +
+                        "}\n";
+        assertEquals(expected, output);
```
jlahoda, line 911: A partly stylistic comment there: it is usually better to have testcases for the code generation under `java/java.source.base/test/unit/src/org/netbeans/api/java/source/gen`, so that regressions are found without running `java.hints` tests. The `java.hints` tests then can usually use `assertOutput`, which ignores formatting changes, but if formatting check is desired, `assertVerbatimOutput` can be used instead of calling `getOutput()` and `assertEquals`.

java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
```diff
@@ -1954,6 +1961,14 @@ protected int diffCase(JCCase oldT, JCCase newT, int[] bounds) {
             endpos = endPos(oldPatterns.get(oldPatterns.size() - 1));
 
             if (newPatterns.isEmpty()) {
+                if(copyTo > localPointer){
```
jlahoda, line 1964: Should this be something like `copyTo(localPointer, copyTo);`?

jlahoda (commented):
Please see two comments line.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertSwitchToRuleSwitchTest.java
mishrasandeep: Thanks, have added the regression test in java/java.source.base/test/unit/src/org/netbeans/api/java/source/gen and made changes to java.hints test.

java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
mishrasandeep: Here I am trying to print the spaces before the default keyword as the formatting was wrong on using the hint. `copyTo(localPointer, copyTo) ` will copy the entire thing which could either be the case or default keyword.

lahodaj: I guess you might need to use the lexer (there is a tokensequence in this class that can be used, IIRC) to find the end offset to copy. The code below will copy only spaces - but indent may be done using tabs, or even may contain comments. Finding the keyword using lexical tokens is probably easiest.

mishrasandeep: Thanks for the input. Have made the changes.

lahodaj (approved):
LGTM.

Akshay-Gupta-Oracle (approved):
+1 These changes are correct.

singh-akhilesh (approved):
Looks good to me.

matthiasblaesing:
The author of this PR is not a committer, so two things are required:

1. A committer, that actually does the commit and
2. Ensures, that the author is aware, that he donates the change to the ASF (as the license headers of the files state) or if there is not enough content there to warant copyright.

It would be good if one of the reviewers could take this task. Thank you!

mishrasandeep:
> 
> 
> The author of this PR is not a committer, so two things are required:
> 
>     1. A committer, that actually does the commit and
> 
>     2. Ensures, that the author is aware, that he donates the change to the ASF (as the license headers of the files state) or if there is not enough content there to warant copyright.
> 
> 
> It would be good if one of the reviewers could take this task. Thank you!

Thanks for the comment. I understand that I am donating to ASF. Akshay and Akhilesh from the team have reviewed the PR and will be doing the commit.

PR #2794, opened 2021-03-04
Support Freeform projects with source Level 9+

matthiasblaesing:
Freeform projects can not support source levels greater than 1.8. The XML schema does not support the new integer version numbers, as a leading number is expected. This PR adds a new XML schema, that supports integer version numbers, updates the selectable java versions till 17 and make the source level an editable combo box to support manually entered version in the future.


matthiasblaesing:
I'll merge this by the end of the week end if noone objects till then.

PR #2795, opened 2021-03-05
Priming build action support

sdedic:
Sometimes it is good idea to **automate** the priming build. So far, the only way how to do it is to  `ProjectProblemsProvider`s and "somehow" detect a priming build is required. So I have defined `COMMAND_PRIME` Project action to be implemented by specific project types that require some sort of 'initialization' before they can be fully used. Currently they are `maven` and `gradle`. 

Maven support comes in the second commit. I tried to avoid multiple runs of the 'priming build', so if the build is run, an additional request will just piggyback on the execution. But once the execution completes, another ProblemResolver invocation will re-build the project.

In maven, I didn't want to pollute even more `ActionProviderImpl` that already has very complex dependencies. At the same time, I didn't want to break the closure of `ActionProvider`s: since Maven does not use LookupMerger for `ActionProvider`s, there can be only that single one (maven core - provided) and others may only plug through providing configurations for maven executions, not by custom `invokeActions`.
So I have decided to make `InternalActionProvider` for use strictly within Maven module. May be used to decompose `ActionProviderImpl` in the future.

Last, in LSP, I have little optimized project opening: if a project **is being** opened, a new request to open the project will wait on the 1st one to complete. Since priming build is now part of the project open task, it may take a little longer.
It was necessary to prime the projects **before project-open event is fired**: otherwise NB support would kick in and start to display dialogs (Project XYZ has problems, ...). These dialogs will be still displayed, if the priming build fails.

### DIFF
--- a/ide/projectapi/apichanges.xml
+++ b/ide/projectapi/apichanges.xml
@@ -83,6 +83,21 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="project-priming-action">
+            <api name="general"/>
+            <summary>Added <code>ActionProvider.COMMAND_PRIME</code> that initializes the project for IDE use.</summary>
+            <version major="1" minor="80"/>
+            <date day="5" month="3" year="2021"/>
+            <author login="sdedic"/>
+            <compatibility addition="yes"/>
+            <description>
+                <p>
+                    <code><a href="@TOP@/org/netbeans/spi/project/ActionProvider.html">ActionProvider.COMMAND_PRIME</a></code> was introduced, so that
+                    projects can declare (and enable/disable as needed) an action that "primes" or initializes the project for the IDE. That often includes
+                    downloads, or partial build, which may produce messages, and could be automated by project-opening code.
+                </p>
+            </description>
+        </change>
         <change id="parent-and-root-project-provider">
             <api name="general"/>
             <summary>Added <code>ParentProjectProvider</code> and <code>RootProjectProvider</code> to improve project hierarchy discovery</summary>
--- a/java/maven/src/org/netbeans/modules/maven/problems/MavenModelProblemsProvider.java
+++ b/java/maven/src/org/netbeans/modules/maven/problems/MavenModelProblemsProvider.java
@@ -23,13 +23,17 @@ import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 import java.io.File;
+import java.lang.ref.Reference;
+import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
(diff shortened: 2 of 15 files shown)


ide/projectapi/apichanges.xml
```diff
@@ -83,6 +83,21 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="project-priming-action">
+            <api name="general"/>
+            <summary>Added <code>ParentProjectProvider</code> and <code>RootProjectProvider</code> to improve project hierarchy discovery</summary>
```
JaroslavTulach, line 88: Summary isn't correct.

ide/projectapi/apichanges.xml
```diff
@@ -83,6 +83,21 @@ is the proper place.
...
+            <version major="1" minor="80"/>
+            <date day="5" month="3" year="2021"/>
+            <author login="sdedic"/>
+            <compatibility addition="yes"/>
+            <description>
+                <p>
+                    <code><a href="@TOP@/org/netbeans/spi/project/ActionProvider.html">ActionProvider.COMMAND_PRIMIE</a></code> was introduced, so that
```
JaroslavTulach, line 95: Typo `PRIMIE`

java/maven/src/org/netbeans/modules/maven/InternalActionProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.netbeans.modules.maven;
+
+import org.netbeans.spi.project.ActionProvider;
+
+/**
+ * The same as ActionProvider, but won't be exported outside the module.
+ * @author sdedic
+ */
+public interface InternalActionProvider extends ActionProvider {
```
JaroslavTulach, line 27: Anyone who queries `Lookup.lookupAll(ActionProvider.class)` may get instances of `InternalActionProvider`. Is that intended? If not, I suggest to use delegation:
```java
public interface InternalActionProvider {
  ActionProvider getProvider();
}
```

java/maven/src/org/netbeans/modules/maven/problems/MavenModelProblemsProvider.java
```diff
@@ -328,4 +392,58 @@ private String getDescriptionText(Throwable e) {
...
     }
+    
+    //-------------------------------------------------------------------------
+    // ActionProvider implementation
+    
+    private static final String[] PROBLEM_ACTIONS = { ActionProvider.COMMAND_PRIME };
+
+    @Override
+    public String[] getSupportedActions() {
+        return PROBLEM_ACTIONS;
```
JaroslavTulach, line 403: Please add a test to verify the action is present, possibly also that it behaves kind of expected.

JaroslavTulach (changes requested):
Can you also implement the `PRIME` action for `java.mx.projects`? That would be `mx build` as that builds all the binaries and the IDE's Java Source infrastructure can then pick them up rather than building them again.

ide/projectapi/apichanges.xml
sdedic: Stupid copy/paste error, sorry :(
Fixed in 9fb5597448

ide/projectapi/apichanges.xml
sdedic: Fixed in 9fb5597448

java/maven/src/org/netbeans/modules/maven/InternalActionProvider.java
sdedic: Ok, done in 04297683b3

JaroslavTulach: Still a bit [dangerous](https://github.com/apache/netbeans/pull/2795#commitcomment-47981828).

sdedic:
> I guess real delegation has be used.

True. Fixed in a1c0839

java/maven/src/org/netbeans/modules/maven/problems/MavenModelProblemsProvider.java
sdedic: Tests added as 7d140d0; tried to test all 3 cases: 
- action enabled since problems were not identified yet, but the priming build itself won't run
- action not enabled on project analyzed to been OK
- action enabled on broken project, priming build relly executes

JaroslavTulach (approved):
I'd still like to see the `PRIME` command to be implemented not only in Maven, Gradle, but also in `java.mx.projects` to smooth the user experience in VSNetBeans when working with graal/compiler sources.

sdedic:
@jtulach Yes, good idea. Tracked as [NETBEANS-5430](https://issues.apache.org/jira/browse/NETBEANS-5430] and [NETBEANS-5431](https://issues.apache.org/jira/browse/NETBEANS-5431)

sdedic:
Changes from review squashed to the original commits. @dbalek:  Last change to object before merge.

PR #2797, opened 2021-03-06
[NETBEANS-5471] Add MIMEResolver annotation using regex

digiovinazzo:
Adding `MIMEResolver.PatternRegistration` annotation to help users register a mime-type using regex and flags for `java.util.regex.Pattern` to match the filename.
Integrating using the same existing approach (see `MIMEResolver.ExtensionRegistration`) and related classes.

See [NETBEANS-5417](https://issues.apache.org/jira/browse/NETBEANS-5417)


matthiasblaesing:
I assume, that it is a conscious decision to mark this PR as draft, so I'll wait with review until it is considered done.

digiovinazzo:
Yes, exactly and I'm not sure it will be ever done: it seems the serialization code of the MIME resolver XML rules is not extensible... so adding a new `MIMEResolver.PatternRegistration` annotation will not be straightforward...

platform/openide.loaders/src/org/openide/loaders/DataNode.java
```diff
@@ -540,6 +540,9 @@ protected Sheet createSheet () {
                 ss.put(new ExtensionProperty());
             }
             ss.put(new SizeProperty());
+            ss.put(new MIMETypeProperty());
```
jtulach, line 543: Deserves a note in `apichanges.xml` of `openide.loaders`.

Wouldn't it be (from a usability standpoint) better to have just a single property (mime type) with custom property editor to show the details? Three properties for MIME type, it's display name and a filter pattern seem a bit too overkill for majority of users.

platform/openide.loaders/src/org/openide/loaders/DataNode.java
```diff
@@ -611,6 +614,60 @@ public Long getValue() {
...
+            return getDataObject().getPrimaryFile().getMIMEType();
+        }
+
+    }
+    
+    private String getMIMEResolverAttribute(FileObject primaryFile, String attributeName) {
+        final String mimeType = primaryFile.getMIMEType();
+        if (mimeType != null) {
+            FileObject mimeResolversFolder = FileUtil.getConfigFile("Services/MIMEResolver"); // NOI18N
+            if (mimeResolversFolder != null) {
```
jtulach, line 633: Isn't this a bit time consuming to iterate all resolvers for every `DataNode` displayed? 

I am not sure what this method is good for, but it doesn't feel right.

platform/openide.filesystems/src/org/netbeans/modules/openide/filesystems/declmime/FileElement.java
```diff
@@ -205,6 +221,18 @@ private void readExternal(DataInput in) throws IOException {
                     names.add(new FileName(in));
                 }
             }
+            int fileNamePatternsSize = -1;
+            try {
+                fileNamePatternsSize = in.readInt();
```
jtulach, line 226: I'd like to see a test covering this part - e.g. to read old format as well as new format.

jtulach (commented):
Neither this PR or NETBEANS-5471 explains why it is important to have the regex pattern. Where it is needed and the current system isn't sufficient? The test show some Eclipse related MIME type - is that the use-case?

Anyway the change looks good at first sight. It is a compatible extension of existing concepts and after some polishing it is going to be fine from API perspective.

Chris2011:
Hey @digiovinazzo what is the progress here? Can you please check the conversation that was started? :)

PR #2798, opened 2021-03-06
Update nbbuild freeform project to open without errors

matthiasblaesing:


matthiasblaesing:
I'll merge this by the end of the week end if noone objects till then.

PR #2800, opened 2021-03-10
Graceful handling of renamed/deleted files.

sdedic:
When a file is renamed on disk by the client, and the client opens it in a new view, a `didClose` is fired on the old and `didOpen` on the new file's URI (location). But the LSP server failed to process properly the old file and return back empty diagnostics. So after renaming or deleting the file, the client still thinks there are problems/hints in the no-longer-existing filename.

I have changed the behaviour by tracking what files were reported the diagnostics to, and if the file vanishes, the LSP server will actively call `publishDiagnostics` with an empty list to clear the client. 

During the course, I have found many places where the code relied on that the file actually exists, but for example VSCode happily calls Code Lens service even for files that were deleted by GIT operation, but VSCode has still the file contents opened in an editor. I attempted to fix all places in `TextDocumentServiceImpl` by returning an empty/neutral value. Except rename, where I really wanted to know the client that the request has failed.


JaroslavTulach (approved):

dbalek (approved):
Looks fine.

PR #2801, opened 2021-03-10
Updating to automatically generated JDK-16+36 nb-javac - WIP

dukescript:
Using [nb-javac JDK-16+36](https://github.com/oracle/nb-javac/pull/12) from a [staged DukeScript repository](https://oss.sonatype.org/content/repositories/comdukescript-1139/com/dukescript/nbjavac/nb-javac/jdk-16+36/) to verify whether such approach can work.


mbien:
i think we can close this since jackpot-backported nb-javac is already in use for a while. If some of this can be salvaged, feel free to reopen.

PR #2802, opened 2021-03-11
Enabling using single instance of javac for multiple files refactoring

Akshay-Gupta-19:


mbien:
closing since the flag doesn't exist anymore

PR #2804, opened 2021-03-11
[NETBEANS-5008] : Support for finalization of pattern-matching-instance-of feature

singh-akhilesh:


java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToPatternInstanceOf.java
```diff
@@ -89,7 +90,9 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
                 TypeMirror typeITM = ctx.getInfo().getTrees().getTypeMirror(typeI);
                 List<String> varNameCandidates = org.netbeans.modules.editor.java.Utilities.varNamesSuggestions(typeITM, ElementKind.LOCAL_VARIABLE, EnumSet.noneOf(Modifier.class), null, null, ctx.getInfo().getTypes(), ctx.getInfo().getElements(), Collections.emptyList(), CodeStyle.getDefault(ctx.getInfo().getFileObject()));
                 String varName = Utilities.makeNameUnique(ctx.getInfo(), ctx.getInfo().getTrees().getScope(ctx.getPath()), varNameCandidates.get(0));
-                Fix fix = new FixImpl(ctx.getInfo(), ctx.getPath(), varName, false, convertPath).toEditorFix();
+                IfTree it = (IfTree) ctx.getPath().getLeaf();
```
lahodaj, line 93: FWIW, I'd be fine if this hint only supported JDK 16+, and the final version of pattern matching in instanceof - hopefully that would make things a little easier. But not that strong opinion.

java/java.source.base/src/org/netbeans/api/java/source/TreeMaker.java
```diff
@@ -1224,6 +1224,10 @@ public Tree BindingPattern(CharSequence name,
                                Tree type) {
         return delegate.BindingPattern(name, type);
     }
+    
+    public Tree BindingPattern(VariableTree vt) {
```
lahodaj, line 1228: It would be good to have `@since` flag here, and the module's specification version updated, + ideally apichanges.xml updated.
Also, possibly, deprecate the old API method above?

lahodaj (changes requested):
I'd suggest to add `@since` and do a few other changes are mentioned inline.

java/java.source.base/src/org/netbeans/api/java/source/TreeMaker.java
lyndseyBeil, line 1228: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

singh-akhilesh: Thanks for providing the comment. This method name cannot be changed because we are following the same naming convention present in the JDK 16 TreeMaker. https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeMaker.java. 
public JCBindingPattern BindingPattern(JCVariableDecl var) {
        JCBindingPattern tree = new JCBindingPattern(var);
        tree.pos = pos;
        return tree;
    }

singh-akhilesh: Addressed code review comments.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToPatternInstanceOf.java
singh-akhilesh: I believe its better to keep this at least for this release. As many people using below JDK-16 version. Thanks.

java/java.source.base/src/org/netbeans/api/java/source/TreeMaker.java
lyndseyBeil: Notced. 
Thanks.

lahodaj (approved):
Looks good.

PR #2805, opened 2021-03-11
[NETBEANS-5438] Fix incorrect PHP xdebug3 port config when copying to clipboard

saulotoledo:
https://issues.apache.org/jira/browse/NETBEANS-5438

Fixes an issue in which Netbeans copies an incorrect configuration for the PHP xdebug port, despite showing the correct configuration to the user.

The xdebug `remote_port` configuration was replaced by `client_port`, as described in [the upgrade guidelines](https://xdebug.org/docs/upgrade_guide#changed-xdebug.remote_port). Netbeans suggests the configuration correctly and allows the user to copy it to the clipboard:

Netbeans showing correct config

However, when the user copies the values, the IDE copies the old value instead of the new one, as shown in the image below:
Incorrectly copied values

The current changes fix such behavior.

### DIFF
--- a/php/php.dbgp/src/org/netbeans/modules/php/dbgp/ConnectionErrMessage.java
+++ b/php/php.dbgp/src/org/netbeans/modules/php/dbgp/ConnectionErrMessage.java
@@ -311,7 +311,7 @@ public class ConnectionErrMessage extends JPanel {
         StringBuilder sb = new StringBuilder();
         sb.append("xdebug.mode=debug").append(NEWLINE); // NOI18N
         sb.append("xdebug.client_host=localhost").append(NEWLINE); // NOI18N
-        sb.append("xdebug.remote_port=").append(String.valueOf(PhpOptions.getInstance().getDebuggerPort())).append(NEWLINE); // NOI18N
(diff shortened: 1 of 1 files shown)


junichi11:
@saulotoledo First of all, please read this: https://netbeans.apache.org/participate/submit-pr.html
You should submit it to the JIRA and should add the issue number to the commit message. Could you please do that? 
Then, we assume that you agree with the license (i.e. https://www.apache.org/licenses/LICENSE-2.0) when I merge this. If you don't agree with it, we can't merge it.

Thanks for catching it.

saulotoledo:
Hi @junichi11. Yes. I do agree with the licence.
I will create the JIRA ticket in the next minutes and I will update the PR as you requested.

saulotoledo:
Hi again @junichi11.
JIRA ticket created and PR title updated.
I also submitted the ICLA form yesterday, in case it is necessary. But I confirm I agree with the license also from here.

Thanks and please let me know if you request any more updates.

junichi11:
@saulotoledo Please use your correct full name (not Saulo) as the author name in your commit. Please update not only the PR title but also the commit message. Thanks.

saulotoledo:
@junichi11 Commit message and author name updated.

junichi11 (approved):
Thanks!

PR #2806, opened 2021-03-12
Auto-open project owner of the opened file.

sdedic:
Especially with gradle subproject, it happens that if the gradle project is not opened, diagnostics run on the project's sources report missing identifiers. The LSP server can internally automatically open project owner when the LSP client opens a file from that project.

This PR performs `asyncOpenSelectedProjects` using the file being opened. If the project is *being opened* (including priming build), the actual `OpenedProjects.open` call will be postponed until after that process completes.
For projects that whose open operation completed already, the Future in `beingOpened` is completed, so calling `asyncOpenSelectedProjects` on an opened project will also complete immediately.

I've changed the semantic of `beingOpened` to keep the completed Futures as a sign that a priming build is not needed (again).

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspServerState.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspServerState.java
@@ -19,6 +19,7 @@
 package org.netbeans.modules.java.lsp.server;
 
 import java.util.List;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import org.eclipse.lsp4j.services.TextDocumentService;
 import org.netbeans.api.project.Project;
@@ -49,6 +50,23 @@ public interface LspServerState {
      * @return opened projects.
      */
     public CompletableFuture<Project[]>   asyncOpenSelectedProjects(List<FileObject> fileCandidates);
+    
+    /**
+     * Opens project on behalf of a file. This makes the project 'second-class citizen' in LSP: it will be
+     * opened in OpenProjects to be reachable for all supports, but will track it separately from projects
+     * opened by {@link #asyncOpenSelectedProjects}. 
+     * <p/>
+     * The user may be asked, if the opened project is not part of existing workspace projects or opened
+     * projects. If the user cancels, the returned future completes exceptionally with {@link CancellationException}.
+     * <p/>
+     * If the file is not owned by a project, or the project open fails, the returned future will return
+     * {@code null}.
+     * 
+     * @param file file owned by a project
+     * @return future that completes when the project is opened, or opening cancelled.
+     * @see CancellationException
+     */
+    public CompletableFuture<Project> asyncOpenFileOwner(FileObject file);
     
     /**
      * Accesses TextDocumentService instance.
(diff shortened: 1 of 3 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -1784,7 +1784,11 @@ public void didOpen(DidOpenTextDocumentParams params) {
                 client.logMessage(new MessageParams(MessageType.Error, ex.getMessage()));
             }
             openedDocuments.put(params.getTextDocument().getUri(), doc);
-            runDiagnoticTasks(params.getTextDocument().getUri());
+            
+            // attempt to open the directly owning project, delay diagnostics after project open:
```
JaroslavTulach, line 1788: Possibly this could open a completely unrelated project - will the user be asked for a consent to trust the project in that case?

Btw. there is [NETBEANS-5204](https://github.com/apache/netbeans/pull/2629/) to help you find whether a subproject is related to already opened project or not.

JaroslavTulach (approved):
Could this be done for NetBeans as well? It would be great to see a bubble in top right corner asking _"Hey, you are working on a file without opened project. Open xyz?"_

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
sdedic: Good point with the project relationships; will try to add checks.

sdedic:
> Could this be done for NetBeans as well? It would be great to see a bubble in top right corner asking "File xyz belongs to project abc. Open it and analyze it?"

Yes, this would be nice. In NetBeans we have `EditorRegistry`, so Projects UI could eventually track the opened files and ask as you suggested. 

Sadly this must be duplicated (somehow) for LSP server, as we do not have some abstraction for an actively-worked-on file except the EditorRegistry that is tied to Swing.

Tracked as  [NETBEANS-5454](https://issues.apache.org/jira/browse/NETBEANS-5454)

sdedic:
Suspending. To work with Gradle projects, I need to implement [NETBEANS-5430](https://issues.apache.org/jira/browse/NETBEANS-5430) first. Maven projects, which already support priming, lacks the support for project hirerarchy queries -- [NETBEANS-5455](https://issues.apache.org/jira/browse/NETBEANS-5455)

sdedic:
Rebased on master to get Maven project parent and root queries to the branch.

I have added the user questions when a project is implicitly opened. The question is asked only once per LSP server instance, so if the user rejects, he won't be bothered again for the project, but the diagnostics etc won't be working much (but the user wanted so).
The question is aked if **the project and no of its parent projects** are already opened **in the workspace**. So if the user e.g. opens a folder with some parent in VSCode, the subprojects (as they are contained in + built with the parent) are opened without questions.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspServerState.java
```diff
@@ -50,6 +51,23 @@
      */
     public CompletableFuture<Project[]>   asyncOpenSelectedProjects(List<FileObject> fileCandidates);
     
+    /**
+     * Opens project oo behalf of a file. This makes the project 'second-class citizen' in LSP: it will be
```
JaroslavTulach, line 55: Typo `oo`

JaroslavTulach (approved):

sdedic:
Squashed + typo fixed. Waiting for build to proceed before merge.

sdedic:
Windows & Mac builds have connectitivy issues. Travis build  for Graalvm (latest) as well. The only 'normally' failing test is `Test java.hints batch1 without nb-javac on Java 14` - but not from this PR.

PR #2808, opened 2021-03-13
Fixing codegen handling of both rewrite of an import and an adding of an import due to qual ident.

jlahoda:
This should fix the failing re-writes of import in the backporting nb-javac. Basically, without this patch, manual changes to import statements are thrown away when import analysis is adding any imports.


JaroslavTulach (approved):

geertjanw:
OK, merging.

PR #2809, opened 2021-03-15
Avoid displaying unnecessary `<classLoader>` field.

JaroslavTulach:
Recently I was working on generating `.hprof` files manually, without dumping the Java heap by the JVM. I have created a [builder library](https://www.graalvm.org/tools/javadoc/org/graalvm/tools/insight/heap/HeapDump.html) to provide nice Java API around the [Java Profiler Heap Dump Format](http://hg.openjdk.java.net/jdk6/jdk6/jdk/raw-file/tip/src/share/demo/jvmti/hprof/manual.html) and I am using it in the [GraalVM Insight](https://github.com/oracle/graal/blob/master/tools/docs/Insight-Manual.md#Heap-Dumping) to heap dump JavaScript, Ruby, Python and other dynamic languages.

When analyzing the generated heap dump with NetBeans 12.3 I noticed `<classLoader> = null` almost every where - see the [original snapshot](https://github.com/oracle/graal/blob/04e2a3c014c524e8de804365cd5c17a11676e68b/tools/docs/Insight-HeapInspect.png) of the dump. The `<classLoader>` is an extra field added to the view useful only when there are some classloaders. However the dynamic languages may not need to expose `java.lang.ClassLoader` objects at all. With the changes in this PR I was able to [improve the snapshot](https://github.com/oracle/graal/blob/master/tools/docs/Insight-HeapInspect.png).

I tried to avoid changes in the `lib.profiler` library that would form a visible API change. Rather I handled everything in the `profiler.heapwalker` module (using a [magic string](http://wiki.apidesign.org/wiki/MagicalStrings) API check, alas). The number of `java.lang.ClassLoader` subclasses is counted when analyzing the heap and if there is no more than one classloader, information to remove the `classLoader` field is passed down through the heap walker nodes to remove the field named `<classLoader>`. I have created some tests to stiffen the intended behavior a bit.

With this PR am I donating this change to Apache NetBeans. I am making the same patch available under the original NetBeans 8.2 license - e.g. CDDL and GPLv2+CPex - should there be an interest to backport this change to previous version of NetBeans & derived apps.

### DIFF
--- a/profiler/lib.profiler/test/unit/src/org/netbeans/lib/profiler/heap/HeapSegmentTest.java
+++ b/profiler/lib.profiler/test/unit/src/org/netbeans/lib/profiler/heap/HeapSegmentTest.java
@@ -21,7 +21,10 @@ package org.netbeans.lib.profiler.heap;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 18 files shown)


sdedic (approved):

JaroslavTulach:
All gates are OK, except the unrelated sigtest failure. Let's merge.

PR #2810, opened 2021-03-15
Fixed invalid branding directory name.

sdedic:
Since the directory name does not follow the branding spec, the resources are not picked up by NB classloader.


dbalek (approved):
Thanks!

PR #2813, opened 2021-03-18
Execute commit validation on JDK15 and JDK11

JaroslavTulach:
I am in a need to test `java.mx.projects` module on JDK15+ and I realized that our testing framework isn't really easy to use on JDK11+. As such I decided to return back to basics and fix at least commit validation. The first commit enables `ant commit-validation` on JDK15. Subsequent commits will fix the failures.

### DIFF
--- a/nbbuild/templates/projectized.xml
+++ b/nbbuild/templates/projectized.xml
@@ -86,7 +86,52 @@
         <property name="have-custom-javac-task" value="true" />
     </target>
 
-    <target name="build-init" depends="basic-init,jdk-8-check,set-buildnumber,-define-custom-javac-task">
+    <target name="-init-compile-bootclasspath-nb-vanilla" depends="basic-init,jdk-8-check,set-buildnumber,-define-custom-javac-task">
+        <property name="bootclasspath.prepend.nb" value="${nb_all}/java/libs.javacapi/external/nb-javac-15.0.0.2-api.jar${path.separator}${nb_all}/java/libs.javacimpl/external/nb-javac-15.0.0.2.jar" />
+        <property name="bootclasspath.prepend.vanilla" value="${nb_all}/nbbuild/external/vanilla-javac-api.jar${path.separator}${nb_all}/nbbuild/external/vanilla-javac-impl.jar" />
+    </target>
+    <target name="-init-bootclasspath-prepend-compile" depends="-init-compile-bootclasspath-nb-vanilla">
+        <condition property="bootclasspath.prepend" value="${bootclasspath.prepend.nb}">
+            <istrue value="${requires.nb.javac.impl}"/>
+        </condition>
+        <condition property="bootclasspath.prepend" value="${bootclasspath.prepend.vanilla}">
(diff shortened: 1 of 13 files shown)


sdedic (approved):

platform/o.n.bootstrap/src/org/netbeans/Util.java
```diff
@@ -170,6 +171,9 @@ public static boolean checkPackageDependency(Dependency dep, ClassLoader cl) thr
             } else {
                 pkg = Package.getPackage(packageName);
             }
+            if (pkg == null && sampleClass != null) {
+                pkg = sampleClass.getPackage();
```
JaroslavTulach, line 175: I think this is correct behavior. We load the `sampleClass` and we are interested in its package. Not sure why used to be done in such a complicated way by throwing the `sampleClass` away. Anyway, @jlahoda wanted to know why it is working in the IDE and it is broken in the test. The `getPackage()` method traverses the `parent` classloaders and here is their list from bottom most to the application classloader:
```
jdk.internal.loader.ClassLoaders$PlatformClassLoader
org.netbeans.junit.NbModuleSuite$JUnitLoader
java.net.URLClassLoader
org.netbeans.Module$StandardClassLoader[debugger.jpda]
```
the problem is probably caused by the `JUnitLoader` as it does its best to skip the application classloader (to force module classes to be loaded by module classloaders) and as 
```
l.loadClass("com.sun.jdi.VirtualMachineManager").getClassLoader()
$5 ==> jdk.internal.loader.ClassLoaders$AppClassLoader@5bc2b487
```
it is not found.

nbbuild/templates/projectized.xml
```diff
@@ -86,7 +86,46 @@
         <property name="have-custom-javac-task" value="true" />
     </target>
 
-    <target name="build-init" depends="basic-init,jdk-8-check,set-buildnumber,-define-custom-javac-task">
+    <target name="-init-bootclasspath-prepend" depends="basic-init,jdk-8-check,set-buildnumber,-define-custom-javac-task" unless="have-jdk-1.9">
```
jlahoda, line 89: It might be better to augment the conditions below that test for `test.use.jdk.javac` to also include `have-jdk1.9`, rather than skipping this completely, so that the `bootclasspath.prepend` is set for compilation.

platform/o.n.bootstrap/src/org/netbeans/Util.java
jlahoda: Thanks for the investigation!

jlahoda (approved):
Please see inline.

nbbuild/templates/projectized.xml
JaroslavTulach: I am not sure I understand, but I assume this has something to do with expected failure of [build on JDK-11](https://travis-ci.com/github/apache/netbeans/jobs/493533632) - `java.source.base` doesn't seen the nb-javac APIs during compilation.

JaroslavTulach: Addressed in eac2f44

PR #2814, opened 2021-03-18
Branded in "Trust all gradle" button.

sdedic:
This is an option that is available in NetBeans IDE, but not accessible in LSP client.


JaroslavTulach (approved):
This is a process oriented approval. I cannot imagine how the final behavior is going to look like just by seeing the three `.properties` lines.

sdedic:
It's true, that with the rest of fixes (open project on file open, trust when primed, ...) in place, it's really hard to see this message in a LSP client.

sdedic:
Retested, still seems to work. Merging.

PR #2816, opened 2021-03-18
[NETBEANS-5430] Implement ActionProvider.COMMAND_PRIME for gradle

sdedic:


extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
```diff
@@ -123,15 +124,42 @@ public ActionProviderImpl(Project project) {
...
-
+    
     @Override
     public void invokeAction(String command, Lookup context) throws IllegalArgumentException {
         if (COMMAND_DELETE.equals(command)) {
             DefaultProjectOperations.performDefaultDeleteOperation(project);
             return;
         }
+        if (ActionProvider.COMMAND_PRIME.equals(command)) {
+            // delegate to prooblem provider we know exists & is registered
```
JaroslavTulach, line 139: Typo `prooblem`

extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
```diff
@@ -123,15 +124,42 @@ public ActionProviderImpl(Project project) {
...
+            ActionProgress prg = ActionProgress.start(context);
+            LOG.log(Level.FINER, "Priming build starting for {0}", project);
+            if (prjImpl.isProjectPrimingRequired()) {
+                prjImpl.primeProject().
+                        thenAccept(gp -> {
+                            LOG.log(Level.FINER, "Priming build of {0} finished with status {1}, ", new Object[] { project, prjImpl.isProjectPrimingRequired() });
+                            prg.finished(prjImpl.isProjectPrimingRequired());
+                        }).
+                        exceptionally((e) -> { 
+                            LOG.log(Level.FINER, e, () -> String.format("Priming build errored: %s", project));
```
JaroslavTulach, line 150: The previous `LOG.log` message is using curly braces pattern. This one is using lambda callback. I'd welcome consistency and I prefer to use the curly braces in both places.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -286,7 +287,57 @@ public String toString() {
...
+            primingBuild = ret;
+        }
+        LOG.log(Level.FINER, "Submitting priming build runs for {0}: {1}", new Object[] { this, ret });
+        GRADLE_PRIMING_RP.submit(() -> {
+            try {
+                // this was explicitly invoked as project action, or problem resolution. Same level as
+                // Build project, so trust the project.
+                ProjectTrust.getDefault().trustProject(this, true);
+                GradleProject gradleProject = GradleProjectCache.loadProject(this, FULL_ONLINE, true, true);
+                LOG.log(Level.FINER, "Priming finished, reloading {0}: {1}", project);
```
JaroslavTulach, line 327: There are two `{}` references, but only one parameter passed in.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -286,7 +287,57 @@ public String toString() {
...
+                fireProjectReload(false);
+                ret.complete(gradleProject);
+            } catch (Throwable t) {
+                LOG.log(Level.FINER, t, () -> String.format("Priming errored for %s", project));
```
JaroslavTulach, line 331: Consider using curly braces pattern.

JaroslavTulach (approved):
It would be great to add an integration test. Nothing fancy, just a new basic Gradle project. Verify priming action exists, it can be invoked and result obtained. Possibly check the state of the problem before and after.

lkishalmi:
How this one relates to #2765 ?

lkishalmi:
I like that someone dig into the details of the Gradle project loading. Yes, unit test would be great. As I see this might actually work.
However I've just committed my work-in-progress code as of #2747 please take a look.

sdedic:
> How this one relates to #2765 ?

It does not.  
This action serves a different purpose. While it is desirable that e.g. Project is not reloaded again needlesly, or that is not marked as "having problems" then display an empty problem window (yes, happens, but I delayed investigation until after e.g. #2747 happens since it may change a lot) - this action allows external control of when the "priming" happens. IMHO there will be always a concept of 'first build'  in Gradle as it is there in Maven (or would be in Ivy), at least download necessary binary depdencies. Withoutt that step, many IDE functions are practically useless. This is a separate (reduced) action in Maven, distinct from Build and I believe it's may be special in Gradle too.

Having a separate action allows an external module, like LSP to change the workflow. In this case, our LSP client user base has different workflow habits, so we are trying to satisfy them. Using this action, LSP users can be happy (priming build can initiated e.g. when a project loads) and NetBeans users can be happy too (nothing is changed). With a possibility to add Project UI core setting "Run priming build after open".

JaroslavTulach:
> Having a separate action allows an external module, like LSP to change the workflow. 

Turns out many important aspects of project behavior are internal. This is not a problem in NetBeans IDE as each project can provide its own UI and communicate such internalities with the user on its own. However, when dealing with the projects only through general API of `org.netbeans.api.Project` or `org.netbeans.api.java.Source` there are certain limitations (revealed with with our work on VSNetBeans) that stand in the way. The `COMMAND_PRIME` is our attempt to fix one of such limitations.

> ...there will be always a concept of 'first build' in Gradle as it is there in Maven (or would be in Ivy), at least download necessary 
> binary dependencies. Without that step, many IDE functions are practically useless. 

Right. The Java Source support needs all the binaries on disk to work. This has been identified even before the Apache transition and externalized (made visible) via [ClassPath.Flag.INCOMPLETE](https://bits.netbeans.org/12.0/javadoc/org-netbeans-api-java-classpath/org/netbeans/api/java/classpath/ClassPath.Flag.html#INCOMPLETE) as introduced by @tzezula. Projects are supposed to communicate to Java Source that the state of the project isn't complete and it makes little sense to analyze it.

> This is a separate (reduced) action in Maven, distinct from Build and I believe it's may be special in Gradle too.

What can a UI do when the `ClassPath` is incomplete? It can start a build as we did in 12.2 - however it turns out that `COMMAND_BUILD` is too heavyweight - it also runs tests which may fail by itself. We need to trigger more lightweight action. As such Sváťa exposed (already existing Maven concept) of priming build as an action in #2795. Making this action work in Gradle is a natural step.

> With a possibility to add Project UI core setting "Run priming build after open".

I believe that by adding a _trust and prime_ checkbox to following dialog

obrazek

we may naturally address various security concerns raised by @emilianbold & others.

sdedic:
Filed for a week, approved for >= 72 hours. FInal notice - going to merge tomorrow.

lkishalmi:
+1

PR #603, opened 2018-06-25
[NETBEANS-980] Fix Home/End/Up/Down actions in line-wrapped editor

eirikbakke:
See https://issues.apache.org/jira/browse/NETBEANS-980 .

This pull request can be reviewed/merged independently of that for NETBEANS-977, which deals with the specific word breaking policy ( https://github.com/apache/incubator-netbeans/pull/598 ).

PR #598, opened 2018-06-22
[NETBEANS-977] Improve text layout in word wrap mode.

eirikbakke:
Improve text layout for the NetBeans editor's "word wrap" feature. This makes the word wrap behavior more in line with that of other editors, and makes wrapped text more readable. See detailed description and before/after screenshot (also attached here) at https://issues.apache.org/jira/browse/NETBEANS-977 .

wrappingdiff

PR #2819, opened 2021-03-20
Improving appearance and behavior of the prepended virtual text.

jlahoda:
With the prepended text (like inline parameter names in Java editor), there are several appearance and behavior issues, like:
-mark occurrences highlight will be painted under the prepended text, not only for the actual identifier, which I think does not look good.
-when navigating with left/right arrows around the prepended text, the cursor jumps weirdly.

The patch here is trying to solve that, but using only basic background highlights for the prepended text (which, to me, looks better when used; even though the implementation is not so nice), and cursor should now have an ability to stop before and after the prepended text.

Feedback/experience with this are welcome!

Thanks!

### DIFF
--- a/ide/editor.lib/src/org/netbeans/editor/BaseKit.java
+++ b/ide/editor.lib/src/org/netbeans/editor/BaseKit.java
@@ -2992,7 +2992,9 @@ public class BaseKit extends DefaultEditorKit {
                                 editorCaret.moveCarets(new CaretMoveHandler() {
                                     @Override
                                     public void moveCarets(CaretMoveContext context) {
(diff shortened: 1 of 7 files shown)


matthiasblaesing:
Changes worked for me as described, what I'm not sure is, whether the changes are worth the added complexity. While noting the latter, this is not a block.

eirikbakke:
Let me check that this does not interfere with the "word wrap" feature... (which was improved in https://github.com/apache/netbeans/pull/603 and https://github.com/apache/netbeans/pull/598 )

eirikbakke:
I don't see any immediate problems with word wrap. Would you mind posting a before/after screenshot? It would make it easier to understand what the patch does.

PR #2822, opened 2021-03-21
Offer to trust and prime the project when it is being opened.

JaroslavTulach:
Sváťa exposed existing Maven concept of priming build as an action in #2795. Support for such action is making its way throughout other project types (#2816) and I believe it is time to use the new API to improve user experience in NetBeans IDE! Let's add a _trust and prime_ checkbox to the following dialog:

obrazek

NetBeans pays quite a lot of attention to security of its users. NetBeans does its best to shield developers against attacks like [CVE-2020-11986](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11986). Clearly a user consent is needed to execute anything from the build. The IDE cannot execute random scripts when users are searching directories (via Favorites window) or opening random text files (via CLI or _File/Open_).

However users of other IDEs are surprised by questions that usually arise after a project is opened - it seems that other IDEs consider adding a project into a workspace or opening new window with a project an enough consent to trust the project.

I believe we can do the same in NetBeans. The _File/Open Project_ dialog/action is explicit enough consent to start trusting the project and perform the `COMMAND_PRIME` to download necessary libraries on disk. To stress that the trust is being granted I propose to add a checkbox (selected by default) into the project open dialog.

### DIFF
--- a/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
@@ -71,6 +71,7 @@ import static org.netbeans.modules.project.ui.Bundle.*;
 import org.netbeans.modules.project.ui.api.UnloadedProjectInformation;
 import org.netbeans.modules.project.ui.groups.Group;
 import org.netbeans.modules.project.uiapi.ProjectOpenedTrampoline;
+import org.netbeans.spi.project.ActionProvider;
 import org.netbeans.spi.project.ProjectContainerProvider;
 import org.netbeans.spi.project.SubprojectProvider;
 import org.netbeans.spi.project.ui.PrivilegedTemplates;
@@ -635,10 +636,10 @@ public final class OpenProjectList {
     }
     
     public void open(final Project[] projects, final boolean openSubprojects, final boolean asynchronously) {
-        open(projects, openSubprojects, asynchronously, null);
+        open(projects, false, openSubprojects, asynchronously, null);
     }
 
-    public void open(final Project[] projects, final boolean openSubprojects, final boolean asynchronously, final Project/*|null*/ mainProject) {
+    public void open(final Project[] projects, boolean prime, final boolean openSubprojects, final boolean asynchronously, final Project/*|null*/ mainProject) {
         if (projects.length == 0) {
             //nothing to do:
             return ;
@@ -664,7 +665,7 @@ public final class OpenProjectList {
                 @Override public void run() {
                     cancellation.t = Thread.currentThread();
                     try {
-                        open(projects, openSubprojects, handle, cancellation);
+                        open(projects, prime, openSubprojects, handle, cancellation);
                     } finally {
                         handle.finish();
                     }
@@ -674,7 +675,7 @@ public final class OpenProjectList {
                 }
             });
         } else {
-            open(projects, openSubprojects, null, null);
+            open(projects, prime, openSubprojects, null, null);
             if (mainProject != null && Arrays.asList(projects).contains(mainProject) && openProjects.contains(mainProject)) {
                 setMainProject(mainProject);
             }
@@ -691,7 +692,7 @@ public final class OpenProjectList {
         "# {0} - project display name", "OpenProjectList.finding_subprojects=Finding required projects of {0}",
         "# {0} - project path", "OpenProjectList.deleted_project={0} seems to have been deleted."
     })
-    public void open(Project[] projects, boolean openSubprojects, ProgressHandle handle, AtomicBoolean canceled) {
+    public void open(Project[] projects, boolean prime, boolean openSubprojects, ProgressHandle handle, AtomicBoolean canceled) {
         LOAD.waitFinished(0);
             
         List<Project> toHandle = new LinkedList<Project>();
@@ -705,6 +706,12 @@ public final class OpenProjectList {
                     toHandle.add(p2);
                 } else {
                     LOGGER.log(Level.WARNING, "Project in {0} disappeared", p.getProjectDirectory());
+                }
+                if (prime) {
+                    ActionProvider ap = p2.getLookup().lookup(ActionProvider.class);
+                    if (ap.isActionEnabled(ActionProvider.COMMAND_PRIME, p2.getLookup())) {
+                        ap.invokeAction(ActionProvider.COMMAND_PRIME, p2.getLookup());
+                    }
                 }
             } catch (IOException ex) {
                 LOGGER.log(Level.INFO, "Cannot convert " + p.getProjectDirectory(), ex);
--- a/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
@@ -71,6 +72,7 @@ import static org.netbeans.modules.project.ui.Bundle.*;
 import org.netbeans.modules.project.ui.api.UnloadedProjectInformation;
 import org.netbeans.modules.project.ui.groups.Group;
 import org.netbeans.modules.project.uiapi.ProjectOpenedTrampoline;
+import org.netbeans.spi.project.ActionProgress;
 import org.netbeans.spi.project.ActionProvider;
 import org.netbeans.spi.project.ProjectContainerProvider;
 import org.netbeans.spi.project.SubprojectProvider;
@@ -103,6 +105,8 @@ import org.openide.util.Parameters;
 import org.openide.util.RequestProcessor;
 import org.openide.util.Utilities;
 import org.openide.util.WeakListeners;
+import org.openide.util.lookup.Lookups;
+import org.openide.util.lookup.ProxyLookup;
 import org.openide.windows.WindowManager;
 
 /**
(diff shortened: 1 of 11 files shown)


sdedic:
Could you include some narrative tooltip on that checkbox ? The 'trust' concept may be quite alien for non-gradle people, as Maven users did not deal with the security that much (Maven is much more declarative -> safer)./

emilianbold:
I agree with not being overly paranoid and having that checkbox on by default seems sane.

It would be nice to present this better. A tooltip and/or rewarding the checkbox would also make sense. The whole concept of 'prime'-ining is basically foreign to most developers outside (low-level?) Gradle.

How about 'Trust the project's build script' and a tooltip along the lines of 'A project's build script might execute foreign code with the full permission of your user account'

neilcsmith-net:
This surely shouldn't be selected by default? At least until the user has had a chance to opt in to always trust. And agree needs description. Better UI would be great, but shouldn't reduce the protection.

emilianbold:
I think it should be selected by default but perhaps also accompanied by a global setting somewhere allowing it to be unselected by default by some users?

Maybe we could also be smart and pick this based on the project location. Ie. a project in the ~/Downloads folder should get be unchecked by default while a project in a folder we have some history with should be trusted.

Of course, you can imagine scenarios where this reduces overall protection, but from an UX standpoint being overly paranoid handicaps the IDE.

neilcsmith-net:
@emilianbold the choice to require explicit user consent was made by the PMC in response to your security report, and in the resulting CVE. Checked by default is not explicit consent. If it's overly paranoid it's because you made it so.

IMO this discussion should maybe move back to private@ to reconsider that and see if the balance is right.

emilianbold:
Way to shoot the messenger!

I showed you a proof of concept. How you implemented the fix and how you
present it to the user was all you.

Having this option now is a vast improvement vs the status quo before the
security issue.

I wouldn't have commented if I wouldn't have received a notification about
this. Please don't @ me anymore and you won't hear from me.

--emi

dum., 21 mar. 2021, 10:48 Neil C Smith ***@***.***> a scris:

JaroslavTulach:
> I think (the checkbox) should be selected by default 

Yes, that's what I believe in as well. 

>> Checked by default is not explicit consent.

I believe pressing the "Open Project" button is explicit enough.

> explicit user consent was made ... in response to (CVE) security report, and....

It is great to see even the CVE reporter agrees with _Open Project_ dialog with a checkbox being good enough.

> but perhaps also accompanied by a global setting somewhere allowing it to be unselected by default by some users?

Both the _"Trust"_ & _"Open required projects"_ settings are persisted for the next time. E.g. it is only about initial default. And yes, I can...

> tooltip on that checkbox
> _A project's build script might execute foreign code with the full permission of your user account_
> _Trust the project's build script_

...change the wording and add a tooltip. Great idea!

> ...I ... have received a notification

Guys, I am not sure what you have done to each other, but neither me nor my PR wants to be part of your personal dispute. I have CCed Emilian to see whether such _checkbox solution_ to his CVE is still perceived OK. It seems Emilian perceives it that way.

Thank you all for your reviews!

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
```diff
@@ -706,6 +707,12 @@ public void open(Project[] projects, boolean openSubprojects, ProgressHandle han
                 } else {
                     LOGGER.log(Level.WARNING, "Project in {0} disappeared", p.getProjectDirectory());
                 }
+                if (prime) {
+                    ActionProvider ap = p2.getLookup().lookup(ActionProvider.class);
+                    if (ap != null && ap.isActionEnabled(ActionProvider.COMMAND_PRIME, p2.getLookup())) {
+                        ap.invokeAction(ActionProvider.COMMAND_PRIME, p2.getLookup());
```
sdedic, line 713: The queries executed below query containment. It could be better to wait for PRIME action to finish, so e.g. gradle has a chance to load its full model and discover the project's content at all.

ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
```diff
@@ -174,6 +188,9 @@ public void actionPerformed( ActionEvent e ) {
         if ( e.getSource() == jCheckBoxSubprojects ) {
             OpenProjectListSettings.getInstance().setOpenSubprojects( jCheckBoxSubprojects.isSelected() );
         }
+        if (e.getSource() == jCheckBoxPrime) {
+            OpenProjectListSettings.getInstance().setTrustAndPrime(jCheckBoxPrime.isSelected());
```
sdedic, line 192: Is it OK that the settings here are remembered even if the user cancels the whole Project|Open at the end ? I know the 'Open Required Projects'  is working that way for ages, but remembering an action after Cancel is ... unexpected.

ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
```diff
@@ -342,6 +359,7 @@ private void setAccessoryEnablement( boolean enable, int numberOfProjects ) {
         jTextFieldProjectName.setEnabled( enable );
         jTextFieldProjectName.setForeground(/* i.e. L&F default */null);
         jCheckBoxSubprojects.setEnabled( enable );
+        jCheckBoxPrime.setEnabled(enable);
```
sdedic, line 362: I'd enable iff any of the selected project(s) provides such an action; e.g. Ant projects do not.

neilcsmith-net:
I have no idea what the personal issue is.  But the security report as reported and accepted asked to acquire explicit consent from the user on project load before executing any code.  I've added @matthiasblaesing as a reviewer, as he and Laszlo discussed most of how this should be implemented in the long discussions over this, as well as the CVE filing.  I don't like the default checked - maybe I've done too much on online consent where this wouldn't be valid - opt-in vs opt-out.  If they're both happy, then fine with me.

Personally, I think what's proposed would have been a sensible approach to the issue in the first place, assuming raised as a feature request and no security issue had been accepted / CVE created.

ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
JaroslavTulach: Poor design by @hrebejk! The values are read in another class, but stored here.

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
JaroslavTulach: Shouldn't load of a full model and set of trust happen as a result of `COMMAND_PRIME`? I believe it happens. As such I wanted to run the priming as soon as possible.

ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
JaroslavTulach: OK: commands.get(commandId)

matthiasblaesing:
> But the security report as reported and accepted asked to acquire explicit consent from the user on project load before executing any code. [...] I don't like the default checked - maybe I've done too much on online consent where this wouldn't be valid - opt-in vs opt-out. If they're both happy, then fine with me.

Yes - the security issue was explicitly, that the IDE executes foreign code without the users explicit consent and with an option, that default to "enabled", we IMHO would revert to that state. It would be interesting to know:

How do other IDEs handle this? I don't see how you could analyse gradle projects without executing them. If I'm not mistaken other ecosystems are even worse, for maven and gradle fetching the artifacts is a pretty save operation (https downloads), while for example npm runs builds as part of dependency installation (_not_ the project build).

So this might be better brought to dev@. Then in in the future we can reference that discussion and the decision from that and not get this again as a security issue.

> Personally, I think what's proposed would have been a sensible approach to the issue in the first place, assuming raised as a feature request and no security issue had been accepted / CVE created.

Yes - at that time I remember, that I was surprised, that that matter was brought to Apache Security and not to the Apache PMC and this gave it a bit of awkward spin. That might also be the explanation of the personal issues, that were raised.

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
sdedic: Understood, but the Prime action may fork the execution into its own RP; it does not necessarily complete synchronously within `invokeAction` scope.

jtulach:
Originally I was also fine with the resolution of [CVE-2020-11986](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11986) however then I got feedback from users of other IDEs and they were quite surprised with the scary warning NetBeans IDE shows before opening Gradle project.

> How do other IDEs handle this?

Further investigation showed that all the IDEs - Eclipse, Idea, VSCode - come with a concept of workspace. A set of "opened" projects. Adding a project/folder/etc. to the workspace is treated as a good enough consent. To quote myself:

> I believe we can do the same in NetBeans. The File/Open Project dialog/action is explicit enough consent to start trusting the project and perform the COMMAND_PRIME to download necessary libraries on disk. 

NetBeans shall not trust projects randomly found by searching the disk or opening individual files. However granting full trust, once the user invokes "File/Open Project...", is more than justified in my opinion.

neilcsmith-net:
> Adding a project/folder/etc. to the workspace is treated as a good enough consent to execute the build script ... granting full trust, once the user invokes "File/Open Project...", is more than justified in my opinion.

Unfortunately that is fundamentally the opposite to what we said by accepting, and responding to, the original security report -

> Opening a Gradle project with NetBeans seems to execute code from the project immediately.
>
> Users would normally expect that selecting some (build / run) action to execute code from a foreign project, not just the act of cloning and opening the project. Projects may often be opened just to read code.

So, are those other IDE's actually executing code on project open?  Because from what I recall of the conversation on this, I think there was recognition that NetBeans could be doing something different by addressing this concern.

I do like adding the control of this to the open project dialog though.  That's a big improvement.  Could we have off by default, remember the setting, and use notification line support to show a (friendlier) warning message whenever the checkbox is unticked?  Enough info that the user understands that decision.

JaroslavTulach:
> I do like adding the control of this to the open project dialog though. That's a big improvement. Could we have off by default, 

Done in 14b508b

ide/projectui/src/org/netbeans/modules/project/ui/OpenProjectList.java
JaroslavTulach: Is a772b3f OK as a fix?

sdedic (approved):

matthiasblaesing (approved):
Only gave this a quick look through, but looks sane to me.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
```diff
@@ -247,7 +247,7 @@ public boolean isActionEnabled(String action, Lookup context) throws IllegalArgu
             case ActionProvider.COMMAND_DEBUG_SINGLE:
                 return fo != null;
             default:
-                return true;
+                return false;
```
JaroslavTulach, line 250: This is not good change.

PR #2823, opened 2021-03-22
Change invokeLaters to Mutex.EVENT.

sdedic:
During prototyping I wanted to pass dynamic context through Lookup.executeWith() to some code deeply burried in EditorSupport; and I discovered a few places where replanning to EDT broke the Lookup-passing logic. Since I've targeted real UI with the prototype, I've caught a few places in WindowManager as well.

The added test in UiUtils was probably omitted from the contribution, but the test source content does not matter, just the class name is tested, so I've provided a dummy test data (the original test data was copied from Swing, I believe).

### DIFF
--- a/java/java.source/src/org/netbeans/api/java/source/UiUtils.java
+++ b/java/java.source/src/org/netbeans/api/java/source/UiUtils.java
@@ -34,7 +34,6 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Modifier;
 import javax.swing.Icon;
-import javax.swing.SwingUtilities;
 import javax.swing.text.StyledDocument;
 import org.netbeans.modules.java.source.pretty.VeryPretty;
 import org.netbeans.modules.java.source.save.DiffContext;
@@ -50,6 +49,7 @@ import org.openide.text.Line;
 import org.openide.text.Line.ShowOpenType;
 import org.openide.text.Line.ShowVisibilityType;
 import org.openide.text.NbDocument;
+import org.openide.util.Mutex;
 import org.openide.util.NbBundle;
 import org.openide.util.UserQuestionException;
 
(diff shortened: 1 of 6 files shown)


java/java.source/src/org/netbeans/api/java/source/UiUtils.java
```diff
@@ -377,27 +377,13 @@ private static boolean doOpen(FileObject fo, int offset) {
...
-        if (SwingUtilities.isEventDispatchThread()) {
-            l.show(ShowOpenType.OPEN, ShowVisibilityType.FOCUS, column);
-        } else {
-            SwingUtilities.invokeLater(new Runnable() {
-                public void run() {
-                    l.show(ShowOpenType.OPEN, ShowVisibilityType.FOCUS, column);
-                }
-            });
-        }
+        Mutex.EVENT.readAccess(() ->
```
JaroslavTulach, line 380: Simplification.

java/java.source/src/org/netbeans/api/java/source/UiUtils.java
```diff
@@ -377,27 +377,13 @@ private static boolean doOpen(FileObject fo, int offset) {
...
-        if (SwingUtilities.isEventDispatchThread()) {
-            oc.open();
-        } else {
-            SwingUtilities.invokeLater(new Runnable() {
-                public void run() {
-                    oc.open();
-                }
-            });
-        }
+        Mutex.EVENT.readAccess(oc::open);
```
JaroslavTulach, line 386: Simplification as well.

platform/core.windows/src/org/netbeans/core/windows/WindowManagerImpl.java
```diff
@@ -1534,8 +1520,8 @@ public void run() {
                     FloatingWindowTransparencyManager.getDefault().start();
                 }
             });
-
-            SwingUtilities.invokeLater(getExclusive());
+            
+            Mutex.EVENT.postReadRequest(getExclusive());
```
JaroslavTulach, line 1524: Yes, `postXYZRequest` is replacement for `invokeLater`.

JaroslavTulach (approved):
Great. `Mutex.EVENT` is more `Lookup`-friendly than `invokeLater`.

sdedic:
Wow, that was quick.

PR #2827, opened 2021-03-23
[NETBEANS-5485] cnd-rfs-1.0.zip compile fix

vieiro:
Fixes some compiler flags and a bug in the source code of cnd-rfs-1.0.zip (that is required to build cnd.remote).


vieiro:
Dismissed with new PR.

PR #2828, opened 2021-03-24
[NETBEANS-5482] Support for ExplicitProcessParameters in gradle/java projects

sdedic:
For the overview, please see [NETBEANS-5482](https://issues.apache.org/jira/browse/NETBEANS-5482). 

Notes:
I needed to declare the token names as '''stable API''' (@JaroslavTulach  insisted on that approach in corresponding Maven PR, as the token may ultimately appear in user-customized `action-mapping.xml`). As such I wanted to declare the tokens as Java constants, so they appear in signatures and are auto-checked. In addition, the Java symbol serves as a vehicle to include `@codesnippet` examples, and reference them from arch/apichanges docs. I am not overly sure with the placement and name of the `org.netbeans.modules.gradle.java.api.ProjectActions`, however.

There does not seem to be a way how to configure `JavaExec.jvmArgs` from the commandline. Luckily there's the Netbeans Tooling plugin, which can inject the info to the tasks. I used `jvmArgsProvider` callback so that `jvmArgs` contributed by other plugins are not replaced, rather they are appened to or overriden.

I've corrected a few formal bugs in documentation (missing apichanges/arch from `project.properties`, incorrect api id reference).


extide/gradle/arch.xml
```diff
@@ -150,6 +150,12 @@
          can be attached to that port, make the <code>netbeans-gradle-tooling</code>
          project debuggable in action.
      </api>
+     <api category="friend" group="property" name="runJvmArgs" type="export">
+         NetBeans Tooling plugin recognizes <code>runJvmArgs</code> project property, and injects
+         specified JVM arguments into JavaExec executing the user code. Use <code>-PrunJvmArgs=...</code>.
+         If passing more arguments, the whole -P must be properly quoted. Arguments that contain spaces must
```
JaroslavTulach, line 156: Example? I am not sure I can imagine how exactly one would do it, yet the wording seems scary enough.

java/gradle.java/arch.xml
```diff
@@ -98,6 +98,18 @@
 
 
  <answer id="exec-property">
+     <api category="stable" group="property" name="javaExec.jvmArgs" type="export">
+         {@code ${javaExec.jvmArgs}} token is replaced in action mappings for JVM arguments collected
+         from <a href="@org-netbeans-modules-extexecution-base@/org/netbeans/api/extexecution/base/ExplicitProcessParameters.html">ExplicitProcessParameters</a>
+         in action's context Lookup. See <a href="@TOP@/org/netbeans/modules/gradle/java/api/ProjectActions.html#TOKEN_JAVAEXEC_JVMARGS">ProjectActions.TOKEN_JAVAEXEC_JVMARGS</a>
+         for more details.
+     </api>
```
JaroslavTulach, line 106: OK.

java/gradle.java/src/org/netbeans/modules/gradle/java/api/ProjectActions.java
```diff
@@ -0,0 +1,52 @@
...
+package org.netbeans.modules.gradle.java.api;
+
+import org.netbeans.api.extexecution.base.ExplicitProcessParameters;
+
+/**
+ * Utilities and constants related to Gradle/Java project actions.
+ * @author sdedic
+ * @since 1.12
+ */
+public class ProjectActions {
```
JaroslavTulach, line 28: Missing private constructor.

java/gradle.java/test/unit/data/javasimple/src/main/java/test/App.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Collections;
+
+public class App {
+    public static void main(String[] args) {
+        for (int i = 0; i < args.length; i++) {
+            System.err.println("args." + i + "=" + args[i]);
+        }
+        String[] props = Collections.list(System.getProperties().propertyNames()).stream().
+            filter(n -> n.toString().startsWith("test.")).
```
JaroslavTulach, line 11: Are you sure the JDK doesn't (and will not) use any property prefixed with `test.`?

java/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/execute/JavaExecTokenProviderTest.java
```diff
@@ -0,0 +1,248 @@
...
+                launcherArg("-verbose").
+                build();
+        ExplicitProcessParameters params2 = ExplicitProcessParameters.builder().
+                args("Baggins").replaceArgs(false).
+                launcherArg("-Dcompanion=Sam").
+                build();
+        
+        ActionProvider actions = project.getLookup().lookup(ActionProvider.class);
+        if (actions == null) {
+            // fail, no action available
```
JaroslavTulach, line 200: assert actions != null : "No actions available";

JaroslavTulach (approved):
Looks consistent with the Maven support.

extide/gradle/arch.xml
sdedic: Hm. Only inlined example. In this case, I have no reasonable symbol to hook to with `@codesnippet` ;) 
But maybe just a wrong wording. OK - the plugin just captures `-PrunJvmArgs=...` property from `gradle` commandline; nothing more.

java/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/execute/JavaExecTokenProviderTest.java
sdedic: Fixed in 5d055fdd54

java/gradle.java/test/unit/data/javasimple/src/main/java/test/App.java
sdedic: Prefix changed in 5d055fdd54

java/gradle.java/src/org/netbeans/modules/gradle/java/api/ProjectActions.java
sdedic: private constructor + final - 5d055fdd54

sdedic:
The request will have been under review for 7 days shortly. Last call: if no objections are raised, will merge after additional 24 hours.

lkishalmi (approved):
Looks good at first sight.

PR #2829, opened 2021-03-24
[NETBEANS-5490] - General Availability for javadoc 16

pepness:
Update link to javadoc 16 now that is in general availability

Before:
https://download.java.net/java/early_access/jdk16/docs/api/

After:
https://docs.oracle.com/en/java/javase/16/docs/api/


sdedic (approved):
Looks sane, link is correct/worklng.

PR #2830, opened 2021-03-24
NETBEANS-5492 change test proxy connection url to apache

premek:
The redirect from http://netbeans.org was evaluated as test failure in General Options. I changed the target URL to https://netbeans.apache.org (which is the target of the redirect from netbeans.org)

The new URL is HTTPS which I hope should be fine or even better in case the proxy does a MIM attack as they sometimes do? (not sure)

I only tested this with No Proxy, System Proxy (which is also no proxy) and Manual Proxy Settings configured to invalid address.

I don't know what the `.sig` file is but I changed it there too.


oers:
This would be very helpful. Maybe change the testMethod in GeneralOptions too, to allow 301 - Moved Permanently as a successfull response (in case this happens again)

premek:
I'm not sure if allowing 301 - Moved Permanently would be correct. Could a misconfigured proxy use that in some case?

matthiasblaesing (changes requested):
@premek I was about to merge this, but then noticed, that the commit is broken. The author information is invalid (no fullname, no valid emailadress) - please fix.

Please also indicate, that you understand, that this change is donated to the ASF (see copyright header in the files you modified), if you don't agree, the change can't be integrated, the PR should be closed then without being integrated.

I'll restart the tests in the meantime, they are sometimes flaky.

oers:
would it be possible to merge this to the LTS branch, too?

premek:
Hi, yes, of course I understand and agree that this change is donated to the ASF.

Is it really necessary that the commit contain a full name and email? I'm using this github setting (using users.noreply.github.com email) trying to limit spam

matthiasblaesing:
Sorry, the ASF is very strict about copyright. It is already a stretch, that the Apache NetBeans project accepts commits without an ICLA (Individual Contributor License Agreement) to make contributing easier, not having authorship in git history would not do.

premek:
I force-pushed the same commit with my name and email, hope it is fine 

(It's unfortunate that to be able to do that you have to disable "Keep my email addresses private" option for your whole github account and all your other repositories)

matthiasblaesing:
Thank you.

For the github settings: the relevant part is the author information in the history - there is no need to open the email in the github settings. The thing to keep in mind: github is a git frontend, but the github GUI (Web, not sure about the desktop integration) is a major PITA.

matthiasblaesing (approved):

PR #2832, opened 2021-03-26
[NETBEANS-4619] [cnd] Adding some missing licenses

vieiro:
This PR adds some missing licenses to different files and updates `nbbuild/rat-exclusions.txt` excluding cnd templates and test files.


vieiro:
This also fixes a bug in one of the native files.

nbbuild/rat-exclusions.txt
```diff
@@ -146,6 +150,104 @@ ide/properties/src/org/netbeans/modules/properties/sample.properties.template
 java/selenium2.java/src/org/netbeans/modules/selenium2/java/SeleneseIT.java.template
 java/testng.ui/src/org/netbeans/modules/testng/ui/resources/*.template
 websvccommon/websvc.saas.api/src/org/netbeans/modules/websvc/saas/oauth/*.template
+# cnd template files
```
matthiasblaesing, line 153: Is it really not possible to include the ALv2 license in each of these files? We had other cases where the templates are processed and the template language allowed to have the licenses header and the licenses would not show up in the generated file (freemarker for example allows to place comments in files).

In anycase, I'd like to prefer to use the licenseinfo.xml facitily to document, that the files are actually ALv2 and not just plain ignored (the build system reads licenseinfo.xml and generates a report from that).

matthiasblaesing (commented):
In general good clean up. The fix I left an inline comment. For the C code
a) I would not have expected it in this PR
b) I see theoretical problems (consider the result when path already has a length of PATH_MAX characters, the resulting path will not what the code expects), but I think fixing the build error is worth it anyway.

vieiro:
Thanks for the review, @matthiasblaesing !
For CND we'll need many PRs, I'm afraid, and I'd like to **very** strict with all licensing details we'll be facing.
Let me try to investigate if those Makefile templates are/could be processed with Freemaker, let's see what should go into licensinfo.xml and let's also separate the two commits in different PRs.
So let's close this PR and do different PRs for them. I think we need this level of detail in CND. Apologies for the noise and thanks again for the review!

PR #2835, opened 2021-03-28
Opt-in to treat warnings as errors

JaroslavTulach:
I am trying to use some of the NetBeans libraries in GraalVM. Graal project is very strict when it comes to code styling issues. Formatting, warnings, Findbugs or Jackpot findings are treated as errors. I kinda got used to that. However it is not enough to fix the warnings once, it is important to compile with `-Werror` to prevent reintroduction of the warnings.

I think NetBeans would benefit from disallowing warnings as well. Commit 
70bf5cb  demonstrates how to go about it - add `-Werror` to your module `nbproject/project.properties`. Compilation with `ant clean build` is likely to fail. Fix the errors. One of the tricky problems is the usage of `@NonNull` & co. common annotations:
```
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NonNull.class): warning: Cannot find annotation method 'when()' in type 'Nonnull': class file for javax.annotation.Nonnull not found
warning: unknown enum constant When.ALWAYS
  reason: class file for javax.annotation.meta.When not found
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NullAllowed.class): warning: Cannot find annotation method 'when()' in type 'Nonnull'
warning: unknown enum constant When.MAYBE
warning: No processor claimed any of these annotations: org.netbeans.api.annotations.common.NullAllowed,javax.annotation.processing.SupportedAnnotationTypes,org.netbeans.api.annotations.common.NonNull
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NonNull.class): warning: Cannot find annotation method 'when()' in type 'Nonnull': class file for javax.annotation.Nonnull not found
warning: unknown enum constant When.ALWAYS
  reason: class file for javax.annotation.meta.When not found
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NullAllowed.class): warning: Cannot find annotation method 'when()' in type 'Nonnull'
warning: unknown enum constant When.MAYBE
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NonNull.class): warning: Cannot find annotation method 'when()' in type 'Nonnull': class file for javax.annotation.Nonnull not found
warning: unknown enum constant When.ALWAYS
  reason: class file for javax.annotation.meta.When not found
 org-netbeans-api-annotations-common.jar(org/netbeans/api/annotations/common/NullAllowed.class): warning: Cannot find annotation method 'when()' in type 'Nonnull'
warning: unknown enum constant When.MAYBE
```
which is caused by the way our build system works. This PR tries to address that as well and does that by removing the `jsr305.jar` in b906a98 and using IDE hints to create the missing classes, fields and methods (see f3cacd2). Exposing the `javax.annotation.meta` package to other modules then fixes the warnings in `api.intent` module.


JaroslavTulach:
Now, when I run `ant -f platform/api.annotations.common build` I get a [SigTest](http://github.com/jtulach/netbeans-apitest) failure:
```
Class javax.annotation.Nonnull
  "E2.8 - Removing default value from member of annotation type" : method public abstract !hasdefault javax.annotation.meta.When javax.annotation.Nonnull.when()
Class javax.annotation.meta.TypeQualifier
  "E1.2 - API type removed" : CLASS public abstract interface !annotation javax.annotation.meta.TypeQualifier
 anno 0 java.lang.annotation.Documented()
 anno 0 java.lang.annotation.Retention(java.lang.annotation.RetentionPolicy value=RUNTIME)
 anno 0 java.lang.annotation.Target(java.lang.annotation.ElementType[] value=[ANNOTATION_TYPE])
```
Looks like I haven't recreated the API properly. Let's fix that now!

JaroslavTulach:
Particularly tricky problem is the `-Xlint:-processor` option. The option isn't checking just `Retention.SOURCE` annotations (that have to have a processor to be useful), but all of kinds of them. Either we can disable the _annotation processor linting_ (as done in 70bf5cb  commit) or we need to mark the well known annotations processed by some dedicated processors to avoid warnings like:
```
warning: No processor claimed any of these annotations: javax.annotation.processing.SupportedAnnotationTypes
warning: No processor claimed any of these annotations: org.junit.Test,org.junit.Before
```
Commit 100db7f30d shows how such processor would look like.

sdedic:
Looks sane.

matthiasblaesing (approved):
Looks sane to me - I left two comments inline, but apart from this getting warnings down is a worthy goal.

platform/api.annotations.common/src/org/netbeans/api/annotations/common/proc/StaticResourceProcessor.java
```diff
@@ -35,12 +36,31 @@
...
+        all.add(CheckForNull.class.getCanonicalName());
+        all.add(CheckReturnValue.class.getCanonicalName());
+        all.add(NonNull.class.getCanonicalName());
+        all.add(NullAllowed.class.getCanonicalName());
+        all.add(NullUnknown.class.getCanonicalName());
+        all.add(StaticResource.class.getCanonicalName());
+        all.add(SuppressWarnings.class.getCanonicalName());
+
+        // other well-known Java platform annotations
+        all.add(SupportedAnnotationTypes.class.getCanonicalName());
```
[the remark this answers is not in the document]
neilcsmith-net: Yes, surely "Either we can disable the annotation processor linting.." is the right approach here? But what other warnings might be lost by doing that, and of those what might actually be errors? Curious a quick search found this - https://bugs.openjdk.java.net/browse/JDK-6999068

geertjanw:
Is this ready for 12.4?

neilcsmith-net:
IMO, not quite yet. I read as WIP approach to annotation warnings? As per my question (re. already suggested options), do we really want processor linting? If we really need an annotation processor to grab things that don't need grabbing, should it be a dedicated and documented one? Other than that, positive change though!

platform/libs.junit4/src/META-INF/services/javax.annotation.processing.Processor
```diff
@@ -14,4 +14,5 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-25EA2E8B0C338A877313BD4672D3FE056EA78F0D com.google.code.findbugs:jsr305:3.0.2
\ No newline at end of file
+#
+#org.netbeans.libs.junit4.NbJUnitProcessor
```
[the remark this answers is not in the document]
jtulach: Opps. Corrected in: 93699b9

jtulach:
> Is this ready for 12.4?

I guess it is ready. The remaining question is whether to keep 93699b9 in or rather use `-Xlint:proccesor`. I am fine with both alternatives, but I prefer to go with 93699b9 as there can be various warnings in annotation processing and having them as errors is useful. The [JDK bug](https://bugs.openjdk.java.net/browse/JDK-6999068) is unlikely to be fixed and moreover we want to compile on existing JDKs anyway. 

If 93699b9 is kept in, then: over time it may happen that we need to expand the list of annotations that deserve to be marked as processed. I see no problem with that. As people turn `-Werror` in certain modules, they should be ready to expand the annotation list (or to add `-Xlint:processor` - the decision can be done per each module).

neilcsmith-net:
Thanks for additional thoughts @JaroslavTulach Makes sense.

JaroslavTulach:
OK. There is an approval by Matthias. Unless objections are raised, I'll merge tomorrow together with 93699b9 - if that commit is too adventurous, speak up and I merge without it and replace it with `-Xlint:processor`.

JaroslavTulach:
Mac build failed due to network error. RAT check is failing for everyone. Let's integrate!

platform/api.annotations.common/build.xml
```diff
@@ -24,79 +24,4 @@
...
-        <property name="src.dir" location="build/generated"/>
-        <property name="src.common" location="src"/>
-
-    </target>
-
-    <target name="build-init" depends="findbugs-annotations-init, projectized.build-init"/>
-
-    <target name="compile" depends="findbugs-annotations, projectized-common.compile"/>
-
-    <target name="findbugs-annotations" if="api.annotations.common.findbugs" depends="build-init, projectized.findbugs-init">
```
phejl, line 40: It's been a long time so I don't remember exactly (and I'm not sure where this might have evolved) but this was present to change annotations to class retention and thus usable in FindBugs/SpotBugs check as it operates on class files.

JaroslavTulach: Welcome back Petr! Looks like the annotations now have `Retention.CLASS` by default: https://bits.netbeans.org/12.3/javadoc/org-netbeans-api-annotations-common/org/netbeans/api/annotations/common/package-summary.html

Otherwise `javac` wouldn't see them when `org-netbeans-api-annotation-commons.jar` is used as a library (which produces the nasty warnings). But probably I should have apply `Retention.CLASS` to `Nonnull`, `TypeQualifierNickname` and `TypeQualifier` as well...

phejl: Ok, then. Thanks @JaroslavTulach. I guess it would be safer to have the other annotations with CLASS retention as well.

PR #2838, opened 2021-03-29
Fixing build for older ant versions.

sdedic:
Master has failed to compile for me. It turned out that ant **1.10.1** I have been using does not support methods `stream()` and `isEmpty()` in that interface. ant **1.10.8** NetBeans itself is distributing is OK. But just removing the `@Override` may fix the build and allow earlier versions to be used.

Setting to WIP as the entire `commit-validation` failed for me once, but I wasn't able to verify whether it is some local problem or systematic problem with the method (i.e. isEmpty) actually not being called. Let's see what the CI says.


JaroslavTulach (approved):
Thank you for correcting my mistake. Breaking the build on older Ant versions wasn't my intention. I just wanted to [compile the code on JDK15](https://github.com/apache/netbeans/pull/2813/commits/29a20c9f81b5974e59d09c6608386918379fa67d) - compilation of the original code was complaining about a [default method being inherited from two super interfaces](http://wiki.apidesign.org/wiki/BinaryCompatibleDefaultMethods).

PR #2840, opened 2021-03-29
Application name is not set correctly on Linux with JDK 16+

matthiasblaesing:
JDK 16 by default denies reflective access to private members.
This prevents netbeans from setting the application name on X11 as
this requires breaking into XToolkit.

The issue is tracked as:
https://bugs.openjdk.java.net/browse/JDK-6528430


jlahoda (approved):
Looks good to me!

PR #2842, opened 2021-03-30
[NETBEANS-5511] Remove items via the iterator.

entlicher:
Explicit removal of items during iteration is causing `ConcurrentModificationException`.
We use `Iterator.remove()` instead.

### DIFF
--- a/platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/FeatureUpdateElementImpl.java
+++ b/platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/FeatureUpdateElementImpl.java
@@ -496,7 +496,8 @@ public class FeatureUpdateElementImpl extends UpdateElementImpl {
                     }
                     Module m = Utilities.toModule(cnb, null);
                     if (m != null && ! m.getProblems().isEmpty()) {
-                        dependenciesToModulesOrFeatures.remove(depSpec);
+                        it.remove();
+                        break;
                     }
                 }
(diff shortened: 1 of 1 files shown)


platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/FeatureUpdateElementImpl.java
```diff
@@ -496,7 +496,7 @@ private void initializeAgent() {
                     }
                     Module m = Utilities.toModule(cnb, null);
                     if (m != null && ! m.getProblems().isEmpty()) {
-                        dependenciesToModulesOrFeatures.remove(depSpec);
+                        it.remove();
```
matthiasblaesing, line 499: I ran against the same wall yesterday and came to the same conclusion. As this is a double nested loop, I think it would be good to either terminate the inner loop with a break or the continue the outer loop with a label. Else we could hit the remove a second time and enter undefined territory.

entlicher: You're right, we need to break here. Calling remove for the second time would throw `IllegalStateException`.

entlicher: break added.

matthiasblaesing (approved):
Looks sane to me.

PR #2843, opened 2021-03-30
Use JSON parser to understand java.mx.projects.

JaroslavTulach:
`java.mx.projects` module is using JavaScript engine to parse the `suite.py` files (after transforming them to JSON format). This is no longer working on JDKs that don't carry JavaScript implementation - e.g. JDK15+.

This PR replaces usage of JavaScript with existing NetBeans parsing library `json_simple`.

### DIFF
--- a/java/java.mx.project/nbproject/project.xml
+++ b/java/java.mx.project/nbproject/project.xml
@@ -60,11 +60,12 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
-                    <code-name-base>org.netbeans.api.scripting</code-name-base>
+                    <code-name-base>org.netbeans.libs.json_simple</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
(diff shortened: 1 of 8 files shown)


JaroslavTulach:
mx has a [special support](https://github.com/graalvm/mx/blob/945d88f1ea536db91aaa10d78304a78a2dc91542/mx_javacompliance.py#L58) for parsing `17-loom` version. NetBeans and VSNetBeans Code extension have to support such format as well otherwise `suite.py` files using such specification aren't going to be analysed correctly. Moreover a6eb9e9 adds robustness by falling back to `8+` compliance if the value cannot be parsed by currently known means.

CCing @gilles-duboscq and @dougxc

sdedic (approved):

JaroslavTulach:
Failure in [RAT check](https://travis-ci.com/github/apache/netbeans/jobs/495125885) is unrelated. Merging.

PR #2847, opened 2021-03-31
[NETBEANS-5524] Check for priming action before use.

sdedic:
I changed the Accessory panel so that it enables the checkbox when there's NO action provider at all. In reality that does not happen with well-behaved projects, as they have at least Build or Run actions. But Ergonomics-proxies have no actions on them (at least they appeared so).
Later, during open, the `getSupportedActions()` check is done again, to see whether the project really supports the action. Seemed to be working fine for PHP (and Maven, Ant, Gradle) on my clean setup.

### DIFF
--- a/ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
@@ -31,6 +31,7 @@ import java.net.URI;
 import java.text.Collator;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


ide/projectui/src/org/netbeans/modules/project/ui/ProjectChooserAccessory.java
```diff
@@ -346,7 +347,12 @@ private static int countPrimable(Iterable<Project> projects) {
         for (Project p : projects) {
             final Lookup lkp = p.getLookup();
             final ActionProvider ap = lkp.lookup(ActionProvider.class);
-            if (ap != null && ap.isActionEnabled(ActionProvider.COMMAND_PRIME, lkp)) {
+            if (ap == null) {
+                // a Project without any actions ? Most probably ergonomics-proxy, count it in!
+                cnt++;
```
JaroslavTulach, line 352: Wow, great idea!

JaroslavTulach (approved):

KacerCZ:
I tested this fix and opening PHP project works.
Thank you.

PR #2849, opened 2021-03-31
Update JavaFX to 16

matthiasblaesing:
This fixes at least one crasher bug, that makes JavaFX unsuitable for serious usage with prebuild web applications:

https://bugs.openjdk.java.net/browse/JDK-8242361

this causes segfaults if data-uris are used in stylesheets.


matthiasblaesing:
To test I created a minimal NetBeans module, that uses a WebView, Button, Label, Events work, Output works, WebView does not crash with an image, that is loaded from a data url.

PR #2855, opened 2021-04-02
Renormalize line-endings for profiler/lib.profiler/scripts/*.bat

matthiasblaesing:
The line endings are correctly set by git via .gitattributes


matthiasblaesing:
As a followup to #2700 this fixes the unnormalized line endings.

matthiasblaesing:
@geertjanw @neilcsmith-net when the tests run clean, I'll merge this. As this is not a content change I don't think this needs to wait.

neilcsmith-net:
Thanks @matthiasblaesing

PR #2853, opened 2021-04-01
Profiler lib native binaries: use ASF binaries with fix for NETBEANS-1428

lbruun:
Use ASF-created binaries for Windows X86, Windows X64, MacOS X64, Linux X86 and Linux X64. The ASF-binaries includes the fix for [NETBEANS-1428](https://issues.apache.org/jira/browse/NETBEANS-1428). For the rest of the platforms that were supported in the past (for example HP-UX, Solaris, Linux ARM) we keep these binaries in the distribution but they are no longer receiving bug fixes and they therefore _don't_ have the fix [NETBEANS-1428](https://issues.apache.org/jira/browse/NETBEANS-1428).

This PR continues (for now) to use the "external binaries" mechanism even if this mechanism was never intended for binaries originating from Apache NetBeans itself (duh!, they are not 'external'). In this light I probably didn't need to create the file `external/profiler-external-binaries-ASF.txt` but I did so for completeness.

PR #2856, opened 2021-04-02
Profiler lib native binaries: use ASF binaries with fix for NETBEANS-1428

matthiasblaesing:
This PR extends #2853 with a fix for the license file issue and merges master, so that the line endings problem is gone.

@geertjanw @neilcsmith-net I'll let github and travis do their work and do a local build in parallel. If all comes back green, I intend to merge - ok with you?


geertjanw:
OK with me and thank you!

PR #2859, opened 2021-04-04
[NETBEANS-5522] Minimal support for Gradle 7.0

lkishalmi:
This patch allows the IDE to read projects that require Gradle 7.0.
I've created another PR #2861, that would bump the tooling version used in NetBeans to 7.0 (actually it is 7.0-rc-2), though not sure if that could fit for 12.4, that is needed for the IDE to be run on JDK 16 and have the Gradle integration work.

This one is safe for 12.4

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
@@ -258,8 +258,10 @@ class NbProjectInfoBuilder {
                             model.info["sourceset_${sourceSet.name}_classpath_annotation"] = storeSet(sourceSet.compileClasspath.files)
                         }
                     }
-                    model.info["sourceset_${sourceSet.name}_configuration_compile"] = sourceSet.compileConfigurationName;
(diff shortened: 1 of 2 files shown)


extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
```diff
@@ -258,8 +258,10 @@ class NbProjectInfoBuilder {
                             model.info["sourceset_${sourceSet.name}_classpath_annotation"] = storeSet(sourceSet.compileClasspath.files)
                         }
                     }
-                    model.info["sourceset_${sourceSet.name}_configuration_compile"] = sourceSet.compileConfigurationName;
-                    model.info["sourceset_${sourceSet.name}_configuration_runtime"] = sourceSet.runtimeConfigurationName;
+                    beforeGradle('7.0') {
+                        model.info["sourceset_${sourceSet.name}_configuration_compile"] = sourceSet.compileConfigurationName;
```
sdedic, line 262: I've tried to find where this information is used in gradle project, but I've found no references to `GradleJavaSourceSet.getRuntimeConfigurationName()` ?
Is this information superseded by something or maybe should be the `getRuntimeConfigurationName()` annotated as `@deprecated` as they won't provide useful info for Gradle >= 7?

PR #2860, opened 2021-04-04
Return FileObject for non-existing JARs.

JaroslavTulach:
@dougxc is working on changes in `mx` and they interfere with the way `java.mx.projects` deal with references to libraries. While debugging I realized that, ideally, one could clone [graal repository](https://github.com/oracle/graal) and open `truffle` suite without `mx build` - e.g. compiling anything. The IDE should handle that.

Well, it should, but it doesn't - if the JAR in `mxbuild` directory isn't present - no URL, but rather `null` was returned as its location. Then it was impossible to locate sources that produce that JAR. Commit 3387fc6 changes that - it returns (generic) JAR location even if it doesn't exist.

The change is based on long time needed, but never really implemented API change that now allows to create `FileObject` instances for files that don't exist: a19c3e3 - say farewell to `Pair<FileObject,URI>` as used elsewhere - create `FileObject` instances even for relative locations that don't exist.

### DIFF
--- a/java/java.mx.project/nbproject/project.xml
+++ b/java/java.mx.project/nbproject/project.xml
@@ -158,7 +158,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>9.9</specification-version>
+                        <specification-version>9.23</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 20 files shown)


tzezula (approved):
Seems good to me.
There are places in the parsing.api and java.source which can be now updated to use FileObject instead of URI (not as a part of this PR).

PR #2861, opened 2021-04-04
[NETBEANS-5541] Upgrade Gradle Tooling to 7.0

lkishalmi:
Well, let's prepare for this. As it happened before the Gradle and NetBeans release come really close again.
This upgrade is required, if someone would like to use NetBeans on Java 16, even if the target project is using an older JDK.

### DIFF
--- a/java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
+++ b/java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
@@ -160,10 +160,26 @@ public final class GradleJavaSourceSet implements Serializable {
         return targetCompatibility.getOrDefault(type, getSourcesCompatibility(type));
     }
 
+    /**
+     * Returns the name of the configuration used by this SourceSet for compile.
+     * This method returns an <code>null</code> from Gradle 7.0 as the
+     * corresponding method has been removed in that version. 
+     * @return the name of the configuration or <code>null</code> if that's not available.
+     * @deprecated No replacement.
+     */
+    @Deprecated
     public String getRuntimeConfigurationName() {
         return runtimeConfigurationName;
     }
 
+    /**
+     * Returns the name of the configuration used by this SourceSet for runtime.
+     * This method returns an <code>null</code> from Gradle 7.0 as the
+     * corresponding method has been removed in that version. 
+     * @return the name of the configuration or <code>null</code> if that's not available.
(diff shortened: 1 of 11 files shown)


java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
```diff
@@ -160,10 +160,18 @@ public String getTargetCompatibility(SourceType type) {
         return targetCompatibility.getOrDefault(type, getSourcesCompatibility(type));
     }
 
+    /**
+     * Returns the name of the configuration used by this SourceSet for compile.
+     * This method returns an empty string from Gradle 7.0 as 
+     * @return the name of the configuration.
+     * @deprecated
```
lkishalmi, line 167: I need to finish these deprecation javadocs.

extide/libs.gradle/external/gradle-tooling-api-7.0-rc-2-license.txt
```diff
@@ -1,7 +1,7 @@
 Name: Gradle Wrapper
 Description: Gradle Tooling API
-Version: 6.7
-Files: gradle-tooling-api-6.7.jar
+Version: 7.0-rc-2
+Files: gradle-tooling-api-7.0-rc-2.jar
```
sdedic, line 4: Is it necessary to upgrade to RC2 for the 12.4 release ? Gradle policy document says:
```
 Release candidates are supported for as long as the release window is open, but they are not intended to be used long term for production use.
```
So we could end up with an unsupported (?) RC in our production / release build ?

lkishalmi: Well, that's a good question, and I'm happy that I'm not wearing the RM hat this time.
It is only necessary to upgrade if we expect the IDE running on jdk16 be able to work with Gradle projects.
We could expect that the final 7.0 Gradle come out during our beta/rc cycles (though there's no guarantee on that)
Or we can add a note to the release notes that this config is not supported.
Also with both options we can add a patch release when the final Gradle 7.0 would be available.

If I'd be the RM, I might risk taking up 7.0-rc-2 and I totally understand if someone would not take that one.

lkishalmi:
It seems Gradle 7.0 is out. I'm going to update this one to contain the GA release artifacts.

JaroslavTulach (approved):

sdedic (approved):

PR #2864, opened 2021-04-06
[NETBEANS-5540] Workaround for 12.4 beta: disable lazy-load by default.

sdedic:
This does not solve loading issues reported in [NETBEANS-5540](https://issues.apache.org/jira/browse/NETBEANS-5540), but disables the newer code for 12.4, while retaining the possibility to test it.


geertjanw:
Would this need a new Beta to be created to support VSNetBeans? (And a new vote thread etc?)

lkishalmi:
Thanks!

PR #2870, opened 2021-04-06
When LSP Server fails to start too many times, don't try start it again.

jlahoda:
Sometimes the LSP Servers fail soon after their process is started, and get re-started over and over, failing each time. This consumes resources unnecessarily. The idea in this patch is to try to start the server a few times, and if it fails, disable it (for the given IDE run, or until the corresponding provider restarts it using the restarter).

For cpplite, we reuse one server for multiple mime-types, but if the server fails, we continue to return the failing instance. We should rather try to start a new one (which is, of course, a subject to the limit on restarts as described above).


matthiasblaesing (approved):
Looks sane to me.

PR #2871, opened 2021-04-07
[NETBEANS-4663][CND] ResolveConflictExecutor in mercurial/subversion remote

vieiro:
Reusing/adapting ResolveConflictsExecutor.java from ide/subversion and ide/mercurial into dlight/subversion.remote and dlight/mercurial.remote, respectively. 
With this PR dlight/subversion.remote and dlight/mercurial.remote seem to compile and test properly, but we may need to review how well they work once CND runs in the IDE.


matthiasblaesing (approved):
No idea if this will work - just commenting from the "is this sound ok" perspective: I aggree. My gut feeling is, that once CND even runs, it will be painful at the start anyway.

vieiro:
@matthiasblaesing Fully agree. The plan with CND is to clear all licensing issues (in case we want to reuse in CPPLITE), then make it compile, then make it run and then test. There's a long road ahead, I'm afraid. Merging.

PR #2839, opened 2021-03-29
Update FlatLaf from 1.0 to 1.1.1 and added "FlatLaf" options

DevCharly:
This PR updates FlatLaf from 1.0 to 1.1.1 and adds a "FlatLaf" options panel under "Appearance" category.

FlatLaf 1.1 release notes:
https://github.com/JFormDesigner/FlatLaf/releases/tag/1.1

FlatLaf 1.1 comes with fantastic news for Windows 10 users/developers: native window decorations.
This enables dark frame/dialog title bar and embedded menu bar with all JREs while still having:
- native Window 10 border drop shadows
- native Window 10 resize functionality
- [Windows 10 snapping](https://support.microsoft.com/en-us/windows/snap-your-windows-885a9b1e-a983-a3b1-16cd-c531795e6241) functionality
- native Windows 10 system window menu when right-clicking on title bar or left-clicking on application icon

## Light

The menu bar and the search field are now embedded into the window title, which saves vertical space for code editing. 😉 
image

Yes, I know that the search field needs some improvements, but this is not part of this PR.

With option "Unified title bar background" enabled:
image

With option "Embedded menu bar" disabled and "Unified title bar background" enabled:
image

## Dark

Finally dark window title bars:
image

With option "Unified title bar background" enabled:
image

With option "Embedded menu bar" disabled and "Unified title bar background" enabled:
image

## New FlatLaf options panel

image

First 3 check boxes are only enabled on Windows 10.

Menu with option "Use underline menu selection" enabled:
image

## Disable

It is possible to disable FlatLaf window decorations or embedded menu bar on command line with following VM options:
~~~
-J-Dflatlaf.useWindowDecorations=false
-J-Dflatlaf.menuBarEmbedded=false
~~~

PR #2872, opened 2021-04-07
FlatLaf: update from 1.1.1 to 1.1.2 and fix/improve options

DevCharly:
This is a follow up to #2839, which unfortunately causes some problems in open editors when changing "Unified title bar" flag in FlatLaf Options: the line numbers ruler was hidden and sometimes I had the impression that some editing features no longer work. The problem was that `FlatLaf.updateUI()` was used to update all UI delegates. This PR no longer does this.

Changes:
- update from 1.1.1 to 1.1.2
- Options: option "Unified window title bar" is now enabled by default because this looks much better, especially in dark theme
- Options: disabling option "Window decorations" now works without restart
- Options: changing option "Unified window title bar" now works without `FlatLaf.updateUI()`, which fixes lost line numbers in editors
- Options: if option "Unified window title bar" is enabled, categories area gets same background as window title bar
- Options: moved categories area colors to FlatLaf properties files
  (this partially implements https://issues.apache.org/jira/browse/NETBEANS-3528 and removes all FlatLaf specific code from `OptionsPanel.java`, which was added in #1703)

Comparison of  "Unified window title bar" enabled/disabled:

image

image

image

image

### DIFF
--- a/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
+++ b/platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
@@ -124,12 +124,11 @@ public class OptionsPanel extends JPanel {
     private final boolean isMac = UIManager.getLookAndFeel ().getID ().equals ("Aqua");
     private final boolean isNimbus = UIManager.getLookAndFeel ().getID ().equals ("Nimbus");
     private final boolean isMetal = UIManager.getLookAndFeel() instanceof MetalLookAndFeel;
-    private final boolean isFlatLaf = UIManager.getLookAndFeel().getID().startsWith("FlatLaf");
     private final boolean isGTK = UIManager.getLookAndFeel ().getID ().equals ("GTK");
     private final Color selected = isMac ? new Color(221, 221, 221) : getSelectionBackground();
-    private final Color selectedB = isMac ? new Color(183, 183, 183) : (isFlatLaf ? selected : new Color (149, 106, 197));
+    private final Color selectedB = isMac ? new Color(183, 183, 183) : getUIColorOrDefault("nb.options.categories.selectionBorderColor", new Color (149, 106, 197));
     private final Color highlighted = isMac ? new Color(221, 221, 221) : getHighlightBackground();
-    private final Color highlightedB = isFlatLaf ? highlighted : new Color (152, 180, 226);
+    private final Color highlightedB = getUIColorOrDefault("nb.options.categories.highlightBorderColor", new Color (152, 180, 226));
     //private final Color iconViewBorder = new Color (127, 157, 185);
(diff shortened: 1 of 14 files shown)


lkishalmi:
@DevCharly please rebase this PR on release124 branch.

neilcsmith-net:
Please don't! We're merging to master until freeze, then we'll have a delivery branch as previously.

lkishalmi:
I'm sorry. You are right. Should have waited till my morning tea kicks in.

platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
```diff
@@ -394,7 +393,7 @@ public void focusLost(FocusEvent e) {
         showHint(true);
         
         pCategories = new JPanel (new BorderLayout ());
-        pCategories.setBorder (BorderFactory.createMatteBorder(0,0,1,0,isFlatLaf ? UIManager.getColor("Separator.foreground"): Color.lightGray)); //NOI18N
+        pCategories.setBorder (BorderFactory.createMatteBorder(0,0,1,0,getUIColorOrDefault("nb.options.categories.separatorColor", Color.lightGray))); //NOI18N
```
matthiasblaesing, line 396: This changes semantics - should this be three level fallback?

- if `nb.options.categories.separatorColor` is defined use that
- if `Separator.foreground` is defined use that
- else fall back to default

matthiasblaesing (commented):
I'm not sure about the new keys, but they seem to be inline with @lkishalmi suggestion. The only thing I'd like to see is add braces to the blocks of the if statements.

platform/options.api/src/org/netbeans/modules/options/OptionsPanel.java
neilcsmith-net: This looks like the same semantics to me. And very glad to see this change.

matthiasblaesing: I see a changed behavior here - if that is intend to change, I'm good with it.

neilcsmith-net: @matthiasblaesing unless you're proposing to keep the flatlaf check in there, what you suggested would return separator.foreground for all look and feels wouldn't it?

matthiasblaesing: @neilcsmith-net sorry, I overlooked the isFlatLaf check.

DevCharly:
> The only thing I'd like to see is add braces to the blocks of the if statements.

Done

matthiasblaesing (approved):
Looks sane to me.

PR #2876, opened 2021-04-10
[NETBEANS-5565] Disable running LSP servers for VCS diffs.

jlahoda:
Seems sensible to not run the LSP servers for diff windows. There does not seem to be a very clean way, so just looking at the directory names.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
@@ -162,7 +162,9 @@ public class LSPBindings {
         if (prj == null) {
             dir = file.getParent();
             File dirFile = FileUtil.toFile(dir);
-            if (dirFile.getName().startsWith("vcs-") && dirFile.getAbsolutePath().startsWith(System.getProperty("java.io.tmpdir"))) {
+            if (dirFile != null &&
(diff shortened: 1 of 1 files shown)


Chris2011:
So that means, that we can't have code completion in the working tree area where I can still change the code? It would be nice, if we can have this soon in NetBeans, unfortunately it is not part of 12.4.

ebarboni:
can we make this for 12.5 @jlahoda ?

Chris2011:
What is missing here? I think if there are no concerns, we can merge this. Also we should think about critical or blocker issues to prioritze them. Is it still time to get it into 12.5?

Chris2011:
Also I would like to get my question answered that I mentioned:

> So that means, that we can't have code completion in the working tree area where I can still change the code?

If it was not clear, I'm asking for the diff window. What happens with the code completion inside of such diff window, if the server is not running? I mean it is not about having the language server, for the TypeScript, it is just that the process where created again and again.

matthiasblaesing:
@Chris2011 12.5 was released about a day ago, so no, this definitely won't make it into 12.5.

Chris2011:
For me the code looks good, I also checked it months ago. If this is the way we go with just checking "vcs-" than it is ok. I just thought this is just a workaround. But I'm not into the API.

Chris2011:
Great thx :). Will have a look

PR #2877, opened 2021-04-11
Fixing newly created cpplite project, avoiding starting the C/C++ LSP server when the configuration is not readable.

jlahoda:


cpplite/cpplite.project/src/org/netbeans/modules/cpplite/project/ui/wizard/CPPLiteProjectWizardIterator.java
```diff
@@ -78,6 +80,7 @@ public Set instantiate() throws IOException {
         Preferences prefs = CPPLiteProject.getRootPreferences(projectDirectory);
         prefs.putBoolean(CPPLiteProject.KEY_IS_PROJECT, true);
         settings.getBuildConfig().save(getBuildPreferences(projectDirectory));
+        getRootPreferences(projectDirectory).put(KEY_COMPILE_COMMANDS, settings.getEditorConfigPath());
```
matthiasblaesing, line 83: I suggest to use the following, to be inline with the existing code in line 81:

```suggestion
        prefs.put(KEY_COMPILE_COMMANDS, settings.getEditorConfigPath())
```

cpplite/cpplite.editor/src/org/netbeans/modules/cpplite/editor/lsp/LanguageServerImpl.java
```diff
@@ -177,16 +183,13 @@ private static File getCompileCommandsDir(CProjectConfigurationProvider configPr
...
-                File commandsPath = new File(config.commandJsonPath);
-                if (commandsPath.canRead()) {
-                    try (InputStream in = new FileInputStream(commandsPath);
-                         OutputStream out = new FileOutputStream(tempFile)) {
-                        FileUtil.copy(in, out);
-                    } catch (IOException ex) {
-                        LOG.log(Level.WARNING, null, ex);
-                        return null;
-                    }
+            } else if (config.commandJsonPath != null && commandsPath.canRead()) {
```
matthiasblaesing, line 186: I would simplify to 

```suggestion
            } else if (commandsPath != null && commandsPath.canRead()) {
```

cpplite/cpplite.project/src/org/netbeans/modules/cpplite/project/ui/wizard/CPPLiteProjectWizardIterator.java
```diff
@@ -33,7 +33,9 @@
 import org.netbeans.api.templates.TemplateRegistration;
 import org.netbeans.modules.cpplite.project.BuildConfiguration;
 import org.netbeans.modules.cpplite.project.CPPLiteProject;
+import static org.netbeans.modules.cpplite.project.CPPLiteProject.KEY_COMPILE_COMMANDS;
 import static org.netbeans.modules.cpplite.project.CPPLiteProject.getBuildPreferences;
+import static org.netbeans.modules.cpplite.project.CPPLiteProject.getRootPreferences;
```
matthiasblaesing, line 38: Can be dropped with the change in line 83.

matthiasblaesing (commented):
I don't know much about the CPPLite code, but I suggest to get this into master, so that is does not bit rot. I left a few inline comments, that I think should be addressed though.

PR #2883, opened 2021-04-14
Reducing dependencies on Swing API when basic API exists.

sdedic:
When I was prototyping for LSP, I stumbled upon some uses of Progress API that were unnecessarily tied to Swing. So replacing usage of `api.progress.nb` calls with equivalent `api.progress` + reducing dependencies between modules.

### DIFF
--- a/extide/o.apache.tools.ant.module/nbproject/project.xml
+++ b/extide/o.apache.tools.ant.module/nbproject/project.xml
@@ -44,14 +44,6 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
-                    <code-name-base>org.netbeans.api.progress.nb</code-name-base>
-                    <build-prerequisite/>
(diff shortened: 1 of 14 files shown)


JaroslavTulach (approved):
Less [heavyweight modules](http://wiki.apidesign.org/wiki/HeavyWeight) doing the same as before are better than before!

eirikbakke:
After switching to NetBeans 18, I started to see a spurious "Checking for external changes" message when exiting the NetBeans IDE. I also see this message in my NetBeans Platform app, which should not be watching any files at all.

231017 Checking for external changes

Using the debugger, the process that's running seems to be the one in MenuWarmUpTask, which was last touched by this PR. Could there be some difference in the behavior of ProgressHandle.createHandle vs. ProgressHandleFactory.createHandle that causes the task to never register as complete?

PR #2885, opened 2021-04-14
Isolate userdirs used by different versions.

JaroslavTulach:
There are frequent user reports of _upgrading to new (development) version and nothing being working then_. 

Why that happens? One hypothesis suggests the problem is in sharing the same userdir between different versions. This PR avoids that by isolating the userdirs between different versions, by including the `.vsix` own version in the name of the _userdir_.

### DIFF
--- a/java/java.lsp.server/vscode/src/extension.ts
+++ b/java/java.lsp.server/vscode/src/extension.ts
@@ -359,11 +359,18 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
         }, time);
     };
 
+    const ownName = "asf.apache-netbeans-java";
+    const ownExtension = vscode.extensions.getExtension(ownName);
+    if (!ownExtension) {
+        throw `Cannot find ${ownName} extension!`;
+    }
+    const version = ownExtension.packageJSON.version;
     const beVerbose : boolean = workspace.getConfiguration('netbeans').get('verbose', false);
     let info = {
         clusters : findClusters(context.extensionPath),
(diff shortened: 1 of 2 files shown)


sdedic:
This can IMHO have some unexpected consequences, i.e. our famous issues with Gradle project trust: as soon as the user upgrades ASF NBJLS extension, his/her approvals will be gone. Now that particular question does not appear that often, but  still...
I can see considerable benefit when switching either from a stable TO a **development** version or maybe between development versions, less when switching between _released_ versions. It's also perfectly OK for the cachedir.

Could there be at least an (optional) setting where the user (config) dir is, so it could be possibly overriden ?

JaroslavTulach:
> considerable benefit when switching either from a stable TO a **development** version or maybe between development versions, less when switching between _released_ versions

Right. This feature is (hopefully) only needed for the development builds - at least the reports we have so far concern only dev builds. The question is: how to identify dev vs. release build?

MartinBalin:
Separating dev and release in VSCode can be hard. userdir is named just by extension inside Code/user/globalStorage. And naming the VSIX like asf.apache-netbeans-java-dev will be counterproductive as you will always have to uninstall release version when trying dev. ...

geertjanw:
Should this be merged, please go ahead, if so, @JaroslavTulach @sdedic.

ebarboni:
@JaroslavTulach can we merge this or should it be postponed?

JaroslavTulach:
There doesn't seem to be an agreement, that this is the right fix.

PR #2886, opened 2021-04-15
[NETBEANS-5370] Prevent generating abstract methods with PHPDoc return array type

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-5370

```php
<?php
namespace Netbeans5370;

abstract class TestAbstractClass
{
    /**
     * @return string[]
     */
    abstract public function testAbstract(): array;
}

// before
class TestClass1 extends TestAbstractClass {
    public function test(): string[] {
    }
}

// after
class TestClass1 extends TestAbstractClass {
    public function test(): array {
    }
}
```


tmysik (approved):

tmysik:
@junichi11 Could you please check the CI failures?

junichi11:
@tmysik I think there is no problem because it's Java area(9082.24 Test Java modules without nb-javac on Java 15). Just in case, I'll restart the job. Thank you!

PR #2888, opened 2021-04-15
Always display errors as an icon in bottom right corner

JaroslavTulach:
By default errors are displayed by blinking icon in bottom right corner in the production builds. Development builds (those with `-ea`) however display dialogs. That can be pretty annoying, especially when there is a lot of subsequent errors. I suggest to use the same mode as in production builds - e.g. to comment out the `assert use = true` check.


geertjanw (approved):
Good idea.

lkishalmi (approved):
Thank you for this one!
Living on the dev builds recently become a bumpy ride with those assertions coming.

geertjanw:
This does not seem controversial, so will merge this, or @JaroslavTulach, feel free to merge, when all checks pass.

matthiasblaesing (approved):
My ritual: Build netbeans from source, remove "-J-ea" from netbeans.conf and be done with it. Can live with this.

neilcsmith-net:
Sounds good!  One of my platform application build files has `-J-Dnetbeans.exception.report.min.level=99999` in it - I guess this must have annoyed me a lot at one point. :smile:

PR #2890, opened 2021-04-15
[NETBEANS-5557] [CND] mime-resolver-hex-based.xml

vieiro:
Adding a mime-resolver-hex-based.xml replacement to the cnd.utils module, that was missing from the 4th Oracle donation.


matthiasblaesing (approved):
I assume the idea here is, that the magic defines a byte array, that is expected at the start of the file and the mask allows to ignore parts of this. I did not check the signatures, but looks sane to me.

vieiro:
Exactly. Very much like the "file" utility works in *nix. This is required to compile, otherwise the MimeResolver annotation breaks the build. We will enhance as required once we're able to run. Merging.

PR #2895, opened 2021-04-16
[NETBEANS-5587] Support SNAPSHOT versions of dependencies.

sdedic:
I have changed the `gavsplit`, since versions provided in the artifact ID contain 4th component, the `{timestamp}-{sequenceno}` for Mavenrepo SNAPSHOT artifacts.

I changed the code to search from the start, as the 'official' syntax for artifact ID is `group`:`artifact`:`version`[:`classifier`][@`ext`] ... so looking for 3rd `:` delimiter seems to satisfy both this and `group`:`artifact`:`version`:`timestamp`-`sequence` which comes to `gavSplit` for Maven SNAPSHOTs.

### DIFF
--- a/extide/gradle/test/unit/src/org/netbeans/modules/gradle/GradleModuleFileCache21Test.java
+++ b/extide/gradle/test/unit/src/org/netbeans/modules/gradle/GradleModuleFileCache21Test.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


extide/gradle/test/unit/src/org/netbeans/modules/gradle/GradleModuleFileCache21Test.java
```diff
@@ -0,0 +1,60 @@
...
+        assertEquals(parts[0], "io.micronaut"); // NOI18N
+        assertEquals(parts[1], "micronaut-core"); // NOI18N
+        assertEquals(parts[2], "2.3.4"); // NOI18N
+    }
+
+    /**
+     * Checks GAV split with -SNAPSHOT and a maven-like timestamp/sequence as the
+     * snapshot unique id.
+     */
+    public void testGavSplitFixedSnapshotWithMavenTimestamp() throws Exception {
```
JaroslavTulach, line 47: Good!

sdedic:
@lkishalmi does the change looks sane ? Please review/comment.

lkishalmi (approved):
Nice one!
Being out of the Maven world it is really easy to forget that the bad practice of SNAPSHOT versions are still a thing there...

lkishalmi:
Sure. The change itself is "trivial" enough and it has tests as well.

geertjanw (approved):

PR #2897, opened 2021-04-16
NETBEANS-5590] Support for assisted download of sources/javadocs for dependencies

sdedic:
This is a follow-up to #2499 - the support developed for downloading javadoc/source is extended to:
- a project action
- a `SourceJavadocAttacherImplementation.Definer` plugin

This allows the user to click "Download" in the NetBeans IDE - and ask Gradle to download sources for the user. After (successful) download, the source re-opens instead of .class file. Since the source is downloaded to Gradle-managed area, from that point on its `SourceForBinaryQueryImplementation` should return the newly acquired source automatically.

During the course, I've made some little improvements:
- `SourceJavadocAttacher` now calls subsequent SPI implementations in the case a former fails to process the resource (the `AttachmentListener` reports a failure). This allows to bail out in case of some network error or something similar condition, giving a chance to the next Attacher
- Gradle 'reload' internal process is **sometimes** used to implement an action (prime, reload, download, ...). The loader now prints this info into the status bar to give the user better understanding of why the gradle is being executed, instead of simple "Reloading" message. This helps LSP client - its UI will display this informative string.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
@@ -126,6 +126,8 @@ public class ActionProviderImpl implements ActionProvider {
         }
         // add a fixed 'prime build' action
         actions.add(ActionProvider.COMMAND_PRIME);
+        actions.add(COMMAND_DL_SOURCES);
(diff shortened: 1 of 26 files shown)


lkishalmi (approved):
Generally it looks Ok to me, however checked the Gradle changes a bit more, those seems to be fine as well.

PR #2904, opened 2021-04-18
In cpplite project, when reading configuration, an empty string should not be considered a valid configuration.

jlahoda:
As reported here:
http://mail-archives.apache.org/mod_mbox/netbeans-users/202104.mbox/%3C56b6c75f-daca-1d28-1805-7e301639c2c2%40slipbits.com%3E

Running a cpplite project when no Run command was specified fails with an exception. When reading configuration, empty values should be ignored (as are missing values), so that the Run action is properly disabled.


ebresie:
Not sure if this is valid or not...If the prefs variable is not set, will the pref.get() portion before the null/empty check on the cmd potentially cause a null pointer/exception as well?

matthiasblaesing (approved):
I suggest to get this in - at worst it is harmless at best it fixes an issue.

For the comment from @ebresie `prefs` can't be null here. It is only ever called from the constructor of the `CPPLiteProject` and there `getBuildPreferences` is called, which takes care of creating the `Preferences` object.

matthiasblaesing:
Tested and found it to work as expected - merging.

PR #2905, opened 2021-04-18
Adding declarative Java hints to the LSP server.

jlahoda:
This should support some features (code completion, errors) for the Jackpot hints.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -1569,29 +1567,18 @@ private void runDiagnoticTasks(String uri) {
         //XXX: cancelling/deferring the tasks!
         diagnosticTasks.computeIfAbsent(uri, u -> {
             return BACKGROUND_TASKS.create(() -> {
-                computeDiags(u, (info, doc) -> {
-                    ErrorHintsProvider ehp = new ErrorHintsProvider();
-                    return ehp.computeErrors(info, doc, "text/x-java"); //TODO: mimetype?
-                }, "errors", false);
+                System.err.println("computing diags for: " + uri + ", errors");
```
JaroslavTulach, line 1570: Forgotten `println`.

ide/api.lsp/nbproject/project.xml
```diff
@@ -52,6 +52,15 @@
                         <specification-version>1.77</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.spi.editor.hints</code-name-base>
```
JaroslavTulach, line 56: This dependency is increasing [weight of the api.lsp module](http://wiki.apidesign.org/wiki/HeavyWeight). How much? I see five editor related libraries. Can such additional height be avoided?

I lived under the impression that the goal of LSP isn't to expose NetBeans Editor API, but rather abstract away from it. Bringing these libraries here via transitive dependencies seems to go against such goal.

@dbalek - right now there is one dependency on editor API in the `api.lsp` module - but hopefully it could be removed by replacing it with direct call to `(String)doc.getProperty("mimeType")`.

JaroslavTulach (commented):
The [increased weight of the module](http://wiki.apidesign.org/wiki/HeavyWeight) worries me a bit.

dbalek (approved):
Looks fine.

PR #2906, opened 2021-04-18
NETBEANS-4150 added a KeyringProvider using libsecret for new versions of Gnome

premek:
Thank you @matthiasblaesing for the libsecret bindings: https://github.com/matthiasblaesing/JNA-Demos/tree/master/LibSecret

I just put it together into GnomeLibSecretProvider.java

Gio.java, Glib.java & Libsecret.java are without a change from your github except I changed the license header with the one from other netbeans files. Is that correct? Or should it have the one you used?

The unit test is passing, I tested saving DB password and proxy password into it from the IDE. Below are some screenshots from how the saved passwords look like in seahorse app in gnome. The gibberish in the last one is part of the test data from the unit test. 
I also tested renaming libsecret in my system temporarily and the IDE used the Fallback KeyringProvider then (like before this change).

Screenshot from 2021-04-18 23-05-30
Screenshot from 2021-04-18 22-47-17
Screenshot from 2021-04-18 22-40-07

### DIFF
--- a/platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
+++ b/platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
@@ -1,0 +1,178 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.keyring.gnome.libsecret;
+
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import com.sun.jna.Structure;
+import com.sun.jna.ptr.PointerByReference;
+import java.io.ByteArrayOutputStream;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+import java.text.MessageFormat;
+import java.util.MissingResourceException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.netbeans.modules.keyring.gnome.libsecret.Glib.GError;
+import org.netbeans.modules.keyring.gnome.libsecret.LibSecret.SecretSchema;
+import org.netbeans.modules.keyring.gnome.libsecret.LibSecret.SecretSchemaAttribute;
+import org.netbeans.spi.keyring.KeyringProvider;
+import org.openide.util.NbBundle;
+import org.openide.util.lookup.ServiceProvider;
+
+@ServiceProvider(service = KeyringProvider.class, position = 90)
+public class GnomeLibSecretProvider implements KeyringProvider {
+
+    private static final Logger LOG = Logger.getLogger(GnomeLibSecretProvider.class.getName());
+    private static final String KEY = "key"; // NOI18N
+    private static final Charset CHARSET = Charset.forName(Native.getDefaultStringEncoding());
+
+    private final String appName;
+
+    private SecretSchema secretSchema = null;
+
+    public GnomeLibSecretProvider() {
+        appName = getAppName();
+    }
+
+    private SecretSchema getSchema() {
+        if (secretSchema != null) {
+            return secretSchema;
+        }
+
+        secretSchema = new SecretSchema();
+        secretSchema.name = appName;
+        secretSchema.flags = LibSecret.SECRET_SCHEMA_NONE;
+        secretSchema.attributes[0] = new SecretSchemaAttribute();
+        secretSchema.attributes[0].name = KEY;
+        secretSchema.attributes[0].type = LibSecret.SECRET_SCHEMA_ATTRIBUTE_STRING;
+        return secretSchema;
+    }
+
+    private String getAppName() {
+        try {
+            return MessageFormat.format(
+                    NbBundle.getBundle("org.netbeans.core.windows.view.ui.Bundle").getString("CTL_MainWindow_Title_No_Project"),
+                    "…");
+        } catch (MissingResourceException x) {
+            return "NetBeans"; // NOI18N
+        }
+    }
+
+    @Override
+    public boolean enabled() {
+        if (Boolean.getBoolean("netbeans.keyring.no.native")) {
+            LOG.fine("native keyring integration disabled");
+            return false;
+        }
+
+        try {
+            read("NoNeXiStEnT"); // NOI18N
+            return true;
+        } catch (RuntimeException e) {
+            LOG.log(Level.WARNING, null, e);
+            return false;
+        } catch (UnsatisfiedLinkError e) {
+            LOG.log(Level.FINE, null, e);
+            return false;
+        }
+    }
+
+    @Override
+    public char[] read(String key) {
+        PointerByReference gerrorBuffer = new PointerByReference();
+        SecretSchema schema = getSchema();
+
+        Pointer pointer = LibSecret.INSTANCE.secret_password_lookup_sync(schema, null, gerrorBuffer, KEY, key);
+
+        if (gerrorBuffer.getValue() != null) {
+            processError(gerrorBuffer);
+            return null;
+        }
+
+        if (pointer == null) {
+            return null;
+        }
+
+        return decode(readZeroTerminatedBytes(pointer));
+    }
+
+    @Override
+    public void save(String key, char[] password, String description) {
+        PointerByReference gerrorBuffer = new PointerByReference();
+        SecretSchema schema = getSchema();
+
+        String label = appName + " - " + (description != null ? description : key);
+        LibSecret.INSTANCE.secret_password_store_sync(schema, LibSecret.SECRET_COLLECTION_DEFAULT, label, encode(password), null, gerrorBuffer, KEY, key);
+
+        if (gerrorBuffer.getValue() != null) {
+            processError(gerrorBuffer);
+        }
+    }
+
+    @Override
+    public void delete(String key) {
+        PointerByReference gerrorBuffer = new PointerByReference();
+        SecretSchema schema = getSchema();
+
+        LibSecret.INSTANCE.secret_password_clear_sync(schema, null, gerrorBuffer, KEY, key);
+
+        if (gerrorBuffer.getValue() != null) {
+            processError(gerrorBuffer);
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.netbeans.modules.keyring.gnome.libsecret.Glib.GError;
+import org.netbeans.modules.keyring.gnome.libsecret.LibSecret.SecretSchema;
+import org.netbeans.modules.keyring.gnome.libsecret.LibSecret.SecretSchemaAttribute;
+import org.netbeans.spi.keyring.KeyringProvider;
+import org.openide.util.NbBundle;
+import org.openide.util.lookup.ServiceProvider;
+
+@ServiceProvider(service = KeyringProvider.class, position = 150)
```
premek, line 35: 150 is higher than (= after) the old Gnome provider but lower than (=before) the FallbackProvider

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,138 @@
...
+
+        schema = new SecretSchema();
+        schema.name = appName;
+        schema.flags = LibSecret.SECRET_SCHEMA_NONE;
+        schema.attributes[0] = new SecretSchemaAttribute();
+        schema.attributes[0].name = KEY;
+        schema.attributes[0].type = LibSecret.SECRET_SCHEMA_ATTRIBUTE_STRING;
+    }
+
+    private String getAppName() {
```
premek, line 55: copied from the old GnomeProvider

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,138 @@
...
+                    NbBundle.getBundle("org.netbeans.core.windows.view.ui.Bundle").getString("CTL_MainWindow_Title_No_Project"),
+                    "…");
+        } catch (MissingResourceException x) {
+            return "NetBeans"; // NOI18N
+        }
+    }
+
+    @Override
+    public boolean enabled() {
+        if (Boolean.getBoolean("netbeans.keyring.no.native")) {
```
premek, line 67: copied from the old GnomeProvider

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,138 @@
...
+        PointerByReference gerrorBuffer = new PointerByReference();
+
+        LibSecret.INSTANCE.secret_password_clear_sync(schema, null, gerrorBuffer, KEY, key);
+
+        if (gerrorBuffer.getValue() != null) {
+            processError(gerrorBuffer);
+        }
+    }
+
+    private void processError(PointerByReference gerrorBuffer) throws IllegalArgumentException {
```
premek, line 128: the old GnomeProvider also does not throw exceptions when the native code returns an error, just returns null in case of `read` and just logs the error in case of `read`, `save` and `delete`. 
It throws an exception when the native library is not available which is used in the `enabled` method.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final Logger LOG = Logger.getLogger(GnomeLibSecretProvider.class.getName());
+    private static final String KEY = "key"; // NOI18N
+
+    private final String appName;
+
+    public GnomeLibSecretProvider() {
+        appName = getAppName();
+    }
+
+    private SecretSchema getSchema() {
```
matthiasblaesing, line 47: I would make the Schema a final variable created in the constructor  -a single instance should be enough for us.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
matthiasblaesing: I have mixed feelings here. LibSecret can integrate with gnome _and_ kde keyring, so it might be an argument to raise it above both, if it is present.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public boolean enabled() {
+        if (Boolean.getBoolean("netbeans.keyring.no.native")) {
+            LOG.fine("native keyring integration disabled");
+            return false;
+        }
+
+        try {
+            read("NoNeXiStEnT"); // NOI18N
```
matthiasblaesing, line 75: I assume, this is needed to trigger library loading? That would at least be consistent with the caught `UnsatisfiedLinkError`

matthiasblaesing (commented):
Looks mostly sane to me - I see two big issues:

Normally it is advised to store passwords in char[] and not go through String, as Strings have the tendency to be retained by the JRE and thus stay longer and can't be explicitly be cleared. Here is an alternative implementation to read the password, assuming they are encoded as UTF-8:

```java
        if(result != null) {
            ByteArrayOutputStream baos = new ByteArrayOutputStream();
            for(int i = 0; i < 100000; i++) {
                byte curVal = result.getByte(i);
                if (curVal == 0) {
                    break;
                } else {
                    baos.write(i);
                }
            }
            return StandardCharsets.UTF_8.decode(ByteBuffer.wrap(baos.toByteArray())).array();
        } else {
            return null;
        }
```

The encoding can be done like this:

```java
        ByteBuffer encodedPasswordBuffer = StandardCharsets.UTF_8.encode(CharBuffer.wrap(password));
        byte[] encodedPassword = new byte[encodedPasswordBuffer.limit() + 1];
        encodedPasswordBuffer.get(encodedPassword, 0, encodedPasswordBuffer.limit());
```

The binding of `secret_password_store_sync` then needs to take a `byte[]` instead of String for the password.

For the record: I agree with the license change of the bindings, as the author.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
premek: I believe [libsecret](https://wiki.gnome.org/Projects/Libsecret) is a gnome project, the underlying d-bus interface should be the same on gnome and kde. 
My thinking here was to put it **after** the existing implementation so if the existing one works and is being used, nothing would change (and break)

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
premek: That is what I did initially but the way the KeyringProvider interface is being used is that an instance is created and then `enabled` is called on it and if the native library is not present, the `UnsatisfiedLinkError` is catched inside the `enabled` method and it returns false (not enabled).

If the constructor throws the exception (on systems where libsecret is not present) at least the current unit tests failed.

Should we make the field for the `schema` but not initialize it in the constructor but on the first use?

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
premek: This is based on how the current GnomeProvider worked. Attempting to read a secret would return the secret or null if it does not exist but it throws  UnsatisfiedLinkError if the native library is not present. This is used to detect if the Provider will be used or not (enabled)

premek:
I'm trying not to use Strings as you suggested but I'm struggling with passing the unit tests right now, I will try again and let you know

premek:
One more thing to consider regarding libsecret is this https://nvd.nist.gov/vuln/detail/CVE-2018-19358
When the user is logged in, the keyring is unlocked and other apps can read the passwords, but looks like it is being used like this anyway...

matthiasblaesing:
> One more thing to consider regarding libsecret is this https://nvd.nist.gov/vuln/detail/CVE-2018-19358
> When the user is logged in, the keyring is unlocked and other apps can read the passwords, but looks like it is being used like this anyway...

The CVE for me falls into the "works as intended" category. If your application runs on a level where it can acccess the whole dbus and thus runs as a local application, it can trivially modify local programms and thus you are already screwed.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
matthiasblaesing: Ok - I'm on gnome, once I install kwallet (because some dependency pulls it in), I'll see some random KDE tool buckering me, while libsecret will transparently solve this.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
matthiasblaesing: You can fill schema in the "enable" call - in the same `try` block as all further calls are required to happen only after `enable` was successfully called.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
premek: now I understand what you mean. I changed it to 90

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
premek: I changed it so the schema is only initialised on the first access

premek:
Thanks Matthias for the review, I updated the PR so it doesn't use Strings for passwords (and fixed the other things you mentioned). The related unit tests  are still passing on my machine

matthiasblaesing:
Thank you again - looks sane and worked for me.

Just for the record: Merging this assumes, that you agree that you donate the code.

platform/keyring.impl/src/org/netbeans/modules/keyring/gnome/libsecret/GnomeLibSecretProvider.java
```diff
@@ -0,0 +1,178 @@
...
+            return MessageFormat.format(
+                    NbBundle.getBundle("org.netbeans.core.windows.view.ui.Bundle").getString("CTL_MainWindow_Title_No_Project"),
+                    "…");
+        } catch (MissingResourceException x) {
+            return "NetBeans"; // NOI18N
+        }
+    }
+
+    @Override
+    public boolean enabled() {
```
dinix2008, line 80: Perhaps you need to add a check for the type of operating system so that MacProvider on MacOS with libsecret installed (for example, via brew) has higher priority.

PR #2907, opened 2021-04-20
NETBEANS-5444 hide Usage Statistics dialog and Option

premek:
As discussed in https://issues.apache.org/jira/browse/NETBEANS-5444 and https://github.com/apache/netbeans-website/pull/544 this change 
- stops showing Usage Statistics dialog (_Help us improve the NetBeans IDE by providing anonymous usage data_) to new users
- hides the option to enable this functionality in GeneralOptionsPanel (the first tab of Tools -> Options)

This doesn't remove any code that deals with enabling this functionality or the functionality itself as I'm not that familiar with the project and not confident for a bigger change like that. Is it OK to just hide it first?

Screenshot from 2021-04-20 21-20-31

### DIFF
--- a/platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java
+++ b/platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java
@@ -244,11 +244,11 @@ public class Installer extends ModuleInstall implements Runnable {
         logMetricsUploadFailed = prefs.getBoolean("metrics.upload.failed", false); // NOI18N
         corePref.addPreferenceChangeListener(new PrefChangeListener());
 
-        if (!Boolean.getBoolean("netbeans.full.hack") && !Boolean.getBoolean("netbeans.close")) {
-            usageStatisticsReminder();
-        }
-        
-        System.setProperty("nb.show.statistics.ui",USAGE_STATISTICS_ENABLED);
+//        if (!Boolean.getBoolean("netbeans.full.hack") && !Boolean.getBoolean("netbeans.close")) {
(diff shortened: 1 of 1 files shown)


platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java
```diff
@@ -244,11 +244,11 @@ private void restoredOnce(java.util.Queue<LogRecord> earlyRecords) {
         logMetricsUploadFailed = prefs.getBoolean("metrics.upload.failed", false); // NOI18N
         corePref.addPreferenceChangeListener(new PrefChangeListener());
 
-        if (!Boolean.getBoolean("netbeans.full.hack") && !Boolean.getBoolean("netbeans.close")) {
-            usageStatisticsReminder();
-        }
-        
-        System.setProperty("nb.show.statistics.ui",USAGE_STATISTICS_ENABLED);
+//        if (!Boolean.getBoolean("netbeans.full.hack") && !Boolean.getBoolean("netbeans.close")) {
+//            usageStatisticsReminder();
```
premek, line 248: This showed the "Help us improve..." dialog (on 2nd start of netbeans I think)

ebarboni:
Hi, added reviewer maybe this need to be included for 12.5 as current NetBeans just link to 404 page.
Code is commented not removed

mbien:
I think we can close this since this was indirectly solved by disabling the entire module responsible for usage statistics and the exception reporter. see #3410

Should be gone since NetBeans 13

feel free to reopen if this isn't the case.

PR #2909, opened 2021-04-21
LSP: Java code completion fixes.

dbalek:
Various fixes improving Java code completion over LSP.


JaroslavTulach (approved):
Code completion from annotation processors works better - both in NetBeans as well as VSCode.

jlahoda (approved):
Looks reasonable.

geertjanw (approved):

PR #2910, opened 2021-04-22
More consistency for virtual file objects

JaroslavTulach:
Jan discovered a surprising misbehavior in masterfs related to invalid child and its parents in e450db5. This PR expands the filesystem TCK and fixes masterfs implementation. Jan also suggested that:
```java
assert !ch.isValid()) : "Let's assume file object is not valid";

try (OutputStream os = ch.getOutputStream()) {
   os.write("Ahoj".getBytes("UTF-8"));
}
```
should work. True one can use the [javax.tools.FileObject](https://docs.oracle.com/javase/8/docs/api/javax/tools/FileObject.html) this way. As the `FileObject.getOutputStream()` method is `final`, it was easy to implement such behavior across all the `FileSystem` implementations.


neilcsmith-net:
If intended for 12.4 please change the base branch to delivery. Otherwise change the milestone. Thanks!

geertjanw (approved):

PR #2911, opened 2021-04-22
Preventing a NPE in LSPBindings, description.bindings can be null.

jlahoda:


geertjanw:
There is a failing test, but we're merging it anyway, please check that test.

geertjanw:
Here's the mail that's the reason for this fix, FYI: https://lists.apache.org/thread.html/r4190ef1ad15bbd8b1e880bdbb11116aef3339a6a0f0905ceec8973fc%40%3Cdev.netbeans.apache.org%3E

PR #2913, opened 2021-04-23
LspSession associates LSP and DAP servers, and check of gdb version.

entlicher:
It's unfortunate, that GDB bug https://sourceware.org/bugzilla/show_bug.cgi?id=26139 is affecting the newly introduced native image debugging. We need to warn users about that, since native image debugging is practically unusable with the buggy version. A change needs to be done in LSP as well, to be able to show the warning in VSCode.

I'm providing two commits.

1) Introducing `LspSession`, which provides LSP and DAP servers. The getters provide the servers while they are running, they are cleared when they finish. This is handled by `ConnectionSpec`, the launch of the servers was changed to return after the server is launched and the `ConnectionSpec` is waiting for the server finish explicitly, to be able to manage the servers in the `LspSession` instance.

2) We check the version of `gdb` and provide a warning when it's the buggy version. For that we need to start debugger in `lspServer.getSessionLookup()` and obtain the `DialogDisplayer` eagerly. That assures that we get the LSP version of `DialogDisplayer` when LSP is active and the default `DialogDisplayer` in NetBeans.

### DIFF
--- a/cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/ni/NIDebuggerProviderImpl.java
+++ b/cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/ni/NIDebuggerProviderImpl.java
@@ -21,6 +21,7 @@ package org.netbeans.modules.cpplite.debugger.ni;
 import java.io.File;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Semaphore;
 import java.util.function.Consumer;
 
 import org.netbeans.api.debugger.Breakpoint;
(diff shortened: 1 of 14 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspSession.java
```diff
@@ -0,0 +1,89 @@
...
+import org.netbeans.api.annotations.common.CheckForNull;
+import org.netbeans.modules.java.lsp.server.debugging.NbProtocolServer;
+import org.netbeans.modules.java.lsp.server.protocol.NbLspServer;
+
+/**
+ * A session that provides associated LSP and DAP servers.
+ *
+ * @author martin
+ */
+public final class LspSession {
```
sdedic, line 32: If this Session is really to be shared between all servers, shouldn't we move the 'session' Lookup from LanguageServerImpl here ?

entlicher: Well, the session lookup is based on the LSP server. We can not create it without LSP connection.
Thus, we can have a lookup that would delegate to `Lookup.getDefault()` and individual servers could add there its services.
If this is what you mean, then yes, that can be added.

entlicher: Done. `LspSession.getLookup()` is there.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/ni/NIDebuggerProviderImpl.java
```diff
@@ -119,9 +121,15 @@ public void finished() {
...
             });
+            started.release();
             debugger.execRun();
             return engineProcess.second();
         }, executionDescriptor, displayName).run();
+        // Wait for the debugger to actually start up.
+        // This is necessary to be able to safely call other methods on the NIDebuggerProvider.
+        try {
+            started.acquire();
+        } catch (InterruptedException ex) {}
```
sdedic, line 132: Nitpick: if thread's `interrupted` status happen to be true, `started` won't be acquired (waited for) but the code will continue. OK ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspSession.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    void setDapServer(NbProtocolServer dapServer) {
+        if (dapServer != null && this.dapServer != null) {
+            throw new IllegalStateException("DAP server is set already");
+        }
+        setServerLookup(dapServer, dapServices);
+        this.dapServer = dapServer;
+    }
+
+    private static void setServerLookup(ScheduledServer server, ProxyLookup.Controller lookupControler) {
```
sdedic, line 74: Q: if this is called first for DAP, then for LSP - will the `server.getServerLookup()` contain both DAP + LSP services ?

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/api/NIDebugRunner.java
```diff
@@ -68,18 +72,55 @@ public static NIDebugger start(File niFile, List<String> arguments, String debug
...
+            i = skipParanthesis(version, i);
+            int eol = version.indexOf("\\n", i);
+            if (eol > 0) {
+                String v = version.substring(i, eol).trim();
+                if (v.startsWith("8.") && !v.startsWith("8.0") || v.startsWith("9.")) {
+                    NotifyDescriptor descriptor = new NotifyDescriptor.Message(Bundle.MSG_GDBVersionBug(), NotifyDescriptor.WARNING_MESSAGE);
+                    displayer.notifyLater(descriptor);
+                }
+            }
+        }
```
sdedic, line 108: GDB version not identified => no message ... OK ?

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/ni/NIDebuggerProviderImpl.java
entlicher: It's true that it's safer to replace it with `acquireUninterruptibly`. Done, thanks.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspSession.java
entlicher: The order does not matter. But lookups of individual servers need to contain just their lookups. The joined lookup is returned from `LspSession.getLookup()`.

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/api/NIDebugRunner.java
entlicher: I belive it's O.K. When the version is not detected, then the versioning is different, or the vendor is different. We do not want to disturb users with false positives. We notify about the GDB bug only when we successfully detect the versions that are affected by that bug.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/LspSession.java
sdedic: Sorry, I was not reading well the diff (I thought there's just one controller).

sdedic (approved):

JaroslavTulach (approved):

geertjanw (approved):

PR #2914, opened 2021-04-23
[NETBEANS-5528] Fix NPE in CoverageReportTopComponent as summary may be null

ebarboni:
Refactor call to update stats
and check if null before try to update


matthiasblaesing:
Change looks straight forward, if it really should go into 12.4 the right target branch would be delivery, master is open for changes for 12.5.

ebarboni:
Oups sorry post poned to 12.5 do not want to stress RM for that.

matthiasblaesing:
Master is open of 12.5 so I suggest to get this in.

PR #2915, opened 2021-04-24
Build & Rebuild actions are disabled in mx projects!

JaroslavTulach:
A [mistake has been made](https://github.com/apache/netbeans/pull/2822/files#r619574208) when implementing #2822 - now build/rebuild actions are disabled for _mx projects_. This PR fixes that by returning `true` for all supported actions and `false` only otherwise.

When at it I also fixed long opened problem - to run/debug unit test one had to build first, otherwise the changes were not visible - now the build happens before running tests automatically. With 8c3b56d the VSNetBeans UI also knows when run/debug test action is over and can hide the debugging toolbar.

This is not a stopper for 12.4 (who cares about mx projects anyway?), but please consider this PR for inclusion in 12.4, if there is a chance.

### DIFF
--- a/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
+++ b/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
@@ -154,7 +154,7 @@ final class SuiteActionProvider implements ActionProvider {
         if (running != null) {
             ActionProgress progress = ActionProgress.start(context);
             running.handle((exitCode, error) -> {
-                progress.finished(exitCode == 0 && error == null);
+                progress.finished(error == null && exitCode == 0);
                 return null;
             });
         }
--- a/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
+++ b/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
@@ -34,6 +34,7 @@ import org.netbeans.api.extexecution.ExecutionDescriptor;
 import org.netbeans.api.extexecution.ExecutionService;
 import org.netbeans.api.extexecution.base.ProcessBuilder;
 import org.netbeans.api.extexecution.print.ConvertedLine;
+import org.netbeans.spi.project.ActionProgress;
 import org.netbeans.spi.project.ActionProvider;
 import org.netbeans.spi.project.SingleMethod;
 import org.openide.LifecycleManager;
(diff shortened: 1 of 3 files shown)


JaroslavTulach:
The tests, including the new [java.mx.project test](https://travis-ci.com/github/apache/netbeans/jobs/500861800#L10021) are all green. Clearly this change must be fixing something.

dbalek (approved):
Looks fine.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
```diff
@@ -139,27 +141,50 @@ public void invokeAction(String action, Lookup context) throws IllegalArgumentEx
...
+        }
     }
 
-    private boolean runMx(String taskName, String... args) {
+    private CompletionStage<Integer> ensureBuilt(FileObject fo) {
+        if (fo != null) {
+            SuiteSources.Group grp = prj.getSources().findGroup(fo);
+            if (grp != null) {
+                final String name = grp.getDisplayName();
+                return runMx(Bundle.MSG_BuildOnly(prj.getName(), name), "build", "--only", name); // NOI18N
```
sdedic, line 168: Should it be `.getDisplayName()` or `.getName()` for the `--only` argument ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
```diff
@@ -139,27 +141,50 @@ public void invokeAction(String action, Lookup context) throws IllegalArgumentEx
...
+            return runMx(taskName, args);
+        });
+    }
+
+    private CompletableFuture<Integer> runMx(String taskName, String... args) {
         final File suiteDir = FileUtil.toFile(prj.getProjectDirectory());
         if (!suiteDir.isDirectory()) {
             Toolkit.getDefaultToolkit().beep();
-            return true;
+            return CompletableFuture.completedFuture(-1);
```
sdedic, line 184: The calling method (`invokeAction`) checks for `runMx() != null`; this result could trigger `ActionProgress.start` - OK ?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
JaroslavTulach: `prj.getDisplayName()` doesn't exist. the `name` isn't a project (e.g. suite in mx terms) name, but rather a name of the source group - e.g. project in mx terms.

sdedic: Sorry, wrong wording. I wanted to ask: should it be
`..., "build", "--only", name);` or `..., "build", "--only", grp.getName());` ? Since `displayName` can be sometimes `displayName = name + "[" + rel + "]";`

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
JaroslavTulach: It would trigger start, but also `finish(false)`, right? At least that is consistent...

JaroslavTulach:
Waiting for approval from release co-ordinators to merge to `delivery` branch. Or Neil, Geertjan feel free to merge.

geertjanw (approved):

PR #2918, opened 2021-04-26
[NETBEANS-5495] [CND] Missing LongHashMap

vieiro:
Reusing LongHashMap from java/java.source.base and replacing functionalities in non-donated UTF class.


vieiro:
Let's close this until we see what's going on with the "NetBeans Remote FS".

PR #2920, opened 2021-04-28
[NETBEANS-2373] Better than nothing support for Gradle Distribution dirs

lkishalmi:
Gradle distribution plugin can be handy from time to time. This PR would add a SourceGroup in the project node for existing dist directories.
Unfortunately Gradle distributions are represented in Gradle as CopySpec-s which are enough to copy required files, but its API is developed for flexible composition, so it does not offer proper introspection. This plugin just assume that the default folder for distribution related sources is ```<project root>/src/<distribution name>/dist```

See: https://docs.gradle.org/current/userguide/distribution_plugin.html

### DIFF
--- a/extide/gradle.dists/build.xml
+++ b/extide/gradle.dists/build.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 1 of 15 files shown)


lkishalmi:
@JaroslavTulach, @sdedic  can you look on this? I miss something that prevents the IDE to load this module. So I was not even able to test/debug this one.

lkishalmi:
image

netbeansuser2019:
It is prefix with xy in case if dist is not in "main" so I suggest that should be done same way as source packages so that should be put to brackets after e.g. Distribution Files [xy].

PR #2922, opened 2021-04-28
12.4 release updates

MartinBalin:
Documentation and version updates for 12.4 Final build.


java/java.lsp.server/package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
```
dbalek, line 2: Why are you adding this file? Isn't it by mistake?

dbalek (approved):

FarmGeek4Life:
Can whoever performs the release build/installer creation re-compile NBI beforehand? https://issues.apache.org/jira/browse/NETBEANS-2523 (with fix https://github.com/apache/netbeans/pull/1844) solves issues with any enterprise that is using PowerShell to determine installed programs, but the issue is still present as of the 12.3 release.

PR #2924, opened 2021-04-29
[NETBEANS-5627] [NETBEANS-5629] Gradle early Lookup and Classpath initialization.

sdedic:
This is an attempt to solve [NETBEANS-5627](https://issues.apache.org/jira/browse/NETBEANS-5627) and [NETBEANS-5629](https://issues.apache.org/jira/browse/NETBEANS-5629)

**It is still work in progress - but complete enough for feedback**

Major highlights of the change:
- a fresh new (untrusted) project still loads as `FALLBACK` state. **But** the Fallback heuristic is immediately applied, so the project (typically) knows its (java) source groups.
- creation attempts to load `EVALUATED` quality. This will eventually load cached info, if available
- with LegacyProjectLoader loading from above EVALUATED the project creation will not invoke Gradle build (the same as now).
- Lookup is initialized with plugins from the start: Fallback loader guesses something, and there's also [NETBEANS-5628](https://issues.apache.org/jira/browse/NETBEANS-5628) that could further improve the initial experience.
- Lookups are in a defined order: Plugins first, the default Lookup first, Plugins next, `_any` (any plugin, fallbacks) last. Similar to Maven. (see [NETBEANS-5629](https://issues.apache.org/jira/browse/NETBEANS-5629)). This allows to add pre-process, make specific stuff with plugin-specific services and do a post-process.
- If plugin Lookups change, their relative order is maintained
- ClassPath starts to listen at the start of its existence
- there were quite a few potential races in both NbGradleProjectImpl and ClasspathProviderImpl. Attempted to synchronize, while firing events out outside of any locks.

In effect, a code that does
```
FileObject someSource = ...;
ClassPath cp = ClassPath.getClassPath(someSource);
```
starts working even on an **unopened** Gradle project, typically. In less typical setups one has to prime / build the project, so NetBeans create its caches.

There are things to improve still:
- apichanges: the Lookups need to be documented, but see the discussion in NETBEANS-5629. Also GradleProject has one more method to (re)load the project at the defined quality
- classpath queries **might** attempt to upgrade the project to FULL when it realizes the project is lesser quality. But I didn't want to do it synchronously in the query - maybe should be done asynchronosly, but that's still lacking some reload pooling
- an attempted reload could piggyback on an asynchronous **pending or scheduled** reload. Typically a reload could reuse stuff from reload-after-build or a pending upgrade-quality reload.
- I must write more tests.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
@@ -246,6 +247,37 @@ public final class NbGradleProjectImpl implements Project {
     
     /**
      * Obtains a project attempting at least the defined quality, without setting
+     * that quality level for subsequent loads. Same as {@link #projectWithQualityTask}
+     * but synchronous (runs in this thread).
+     * @param desc optional description for the loading process, can be {@code null}.
+     * @param aim aimed quality
+     * @param interactive true, if user messages/confirmations can be displayed
+     * @param force to force load even though the quality does not change.
+     * @return project instance
+     */
+    public GradleProject projectWithQuality(String desc, Quality aim, boolean interactive, boolean force) {
+       synchronized (this) {
+            GradleProject c = project;
+            if (c != null) {
+                if (c.getQuality().atLeast(aim)) {
+                    return c;
+                }
+                if (!force && attemptedQuality.atLeast(aim)) {
+                    return c;
+                }
+            }
+        }
+        try {
+            return loadOwnProject0(desc, false, interactive, aim, true, force).get();
+        } catch (InterruptedException | ExecutionException ex) {
+            // should not happen, the event dispatch + potential issues happen
+            // synchronously
+            return null;
+        }
+    }
+    
+    /**
+     * Obtains a project attempting at least the defined quality, without setting
      * that quality level for subsequent loads. Note that the returned project's quality
      * must be checked. If the currently loaded project declares the desired quality,
      * no load is performed.
@@ -258,27 +290,35 @@ public final class NbGradleProjectImpl implements Project {
      * </div>
      * @param desc optional description for the loading process, can be {@code null}.
      * @param aim aimed quality
+     * @param interactive true, if user messages/confirmations can be displayed
+     * @param force to force load even though the quality does not change.
      * @return project instance
      */
-    public GradleProject projectWithQuality(String desc, Quality aim, boolean interactive, boolean force) {
+    public CompletableFuture<GradleProject> projectWithQualityTask(String desc, Quality aim, boolean interactive, boolean force) {
         synchronized (this) {
             GradleProject c = project;
             if (c != null) {
                 if (c.getQuality().atLeast(aim)) {
-                    return c;
+                    return CompletableFuture.completedFuture(c);
                 }
                 if (!force && attemptedQuality.atLeast(aim)) {
-                    return c;
+                    return CompletableFuture.completedFuture(c);
                 }
             }
         }
-        try {
-            return loadOwnProject0(desc, false, interactive, aim, true).get();
-        } catch (InterruptedException | ExecutionException ex) {
-            // should not happen, the event dispatch + potential issues happen
-            // synchronously
-            return null;
-        }
+        CompletableFuture<GradleProject> toRet = new CompletableFuture<>();
+        RELOAD_RP.post(() -> 
+            loadOwnProject0(desc, false, interactive, aim, false, force)
+                .handle((p, e) -> {
+                   if (e != null) {
+                       toRet.complete(p);
+                   } else {
+                       toRet.completeExceptionally(e);
+                   }
+                   return null;
+                })
+        );
+        return toRet;
     }
 
     /**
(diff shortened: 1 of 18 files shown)


extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -223,52 +229,179 @@ void detachAllUpdater() {
...
-    void dumpProject() {
+    synchronized void dumpProject() {
         project = null;
+        attemptedQuality = null;
+        loadedProjectSerial = 0;
+        aimedQuality = FALLBACK;
     }
 
-    public Quality getAimedQuality() {
+    public final Quality getAimedQuality() {
```
lkishalmi, line 239: NbGradleProjectImpl class is already final.

java/gradle.java/src/org/netbeans/modules/gradle/java/classpath/ClassPathProviderImpl.java
```diff
@@ -70,36 +71,58 @@
...
+        NbGradleProject gp = NbGradleProject.get(project);
+        if (gp.isGradleProjectLoaded()) {
+            updateGroups();
+        }
+        NbGradleProject.addPropertyChangeListener(project, wPcl);
+    }
+    
+    private void updateResources(URI uri) {
+        if ((uri != null) && (uri.getPath() != null) && uri.getPath().endsWith(MODULE_INFO_JAVA)) {
+            GradleJavaProject gjp = GradleJavaProject.get(ClassPathProviderImpl.this.project);
```
lkishalmi, line 103: ClassPathProviderImpl.this.project -> project ?

lkishalmi (approved):
Really nice work! Thank you!
The changes seem to be reasonable, can't wait to test this out in my local build.
Mentioned only two stylistic thing, but those do not really matter.

victorwss:
Sorry to comment this here, but since you (@sdedic) are working in the issue and the mailing lists are a confusing mess that I'm unable to follow, let me comment here. Pardon me if you already know what I will tell.

I started to try out Netbeans 12.3 + Gradle 7 + Java 16 and the result was very bad. The machine was fresh and there was no Java nor Netbeans ever installed before. So I installed Java 16 and then just downloaded and installed Netbeans 12.3.

I got a small project in Gradle and tried to build it. My first result was a silent NullPointerExceptions somewhere inside Netbeans when it tried to access some array in order to build the project, with the output panel blank. If you are interested, I can backtrack to this point in order to reproduce it.

Anyway, trying to make it work, I manually downloaded Gradle 7.0 and setted it in the Options > Java > Gradle setting.

Then, when Netbeans tries to to the prime-loading the project with its internally downloaded Gradle 6.7 regardless of any setting that I to try to do to force the use of Gradle 7.0. This results in a Gradle instantly crashing with unsupported classfile format because Gradle 6.7 can't recognize Java 16.

To make it worse, when Netbeans fails in the prime-loading, it alerts that the project is unloadable due to errors and fallbacks to an unchangeable "-source 5" which makes the java files are full of spurious errors about uses of diamonds, lambdas, records and a lot of other stuffs.

The result is that so far, is that Netbeans 12.3 + Gradle 7 + Java 16 is pretty much an unusable combination. I lost some hours trying to make this work and after searching a lot on Google, I just landed here.

So, I just wanted to inform you about the experience in order to avoid that some of those cases are missed. And also, to tell you that this problem is surely seen in the wild and are biting some random people out there like me. Anyway, keep the good work!

sdedic:
@victorwss Thanks for your feedback. Sadly this particular PR is probably completely unrelated to gradle version + Java issues. 
This PR aims to improve NetBeans' tools cooperation with the gradle module; it never aimed to fix anything between NB gradle module and the gradle daemon itself.

If you can reproduce the defect(s), then great ! Please file JIRA issues for each one separately; if you manage to get a stacktrace of the error, please attach it to that issue. Include reproduction steps; ideally starting from a fresh install or project open. GitHub PRs discussions are not intended for reporting (general) bugs, or for product feedback, the primary purpose is to review and discuss changes proposed in the pull request.

lkishalmi:
@victorwss Use NetBeans 12.4 (rc1 is out rc2 is coming) that has Gadle 7.0 support.

sdedic:
@victorwss Incidentally, I had some internal (*wink*) report of the **same** issue that I needed to process. See [micronaut-starter#757](https://github.com/micronaut-projects/micronaut-starter/issues/757)

sdedic:
Added some tests, changed `toQuality` to return just `CompletionStage` plus improved `ClassPathProviderImpl` to make the project to initialize (if it was not yet).

@lkishalmi  please re-review

extide/gradle/src/org/netbeans/modules/gradle/api/NbGradleProject.java
```diff
@@ -206,6 +210,30 @@ public Quality getAimedQuality() {
...
+     * if the project is not trusted, its Gradle build will not be executed, so the returned quality can be {@link Quality#EVALUATED}.
+     * 
+     * @param reason reason for reload, may be {@code null}.
+     * @param q the desired quality of project information
+     * @param forceLoad force load even though the current info quality is sufficient.
+     * @return {@link CompletionStage} with the reloaded project. Use {@link CompletionStage#toCompletableFuture()}.{@link CompletableFuture#get get()} 
+     * to block waiting for the result.
+     * @since 2.11
+     */
+    public @NonNull CompletionStage<NbGradleProject> toQuality(@NullAllowed String reason, @NonNull Quality q, boolean forceLoad) {
```
JaroslavTulach, line 234: Great to see `CompletionStage` in out APIs! I'd say `CompletionStage` is the best after invention of sliced bread!

JaroslavTulach (approved):

extide/gradle/src/org/netbeans/modules/gradle/api/NbGradleProject.java
lkishalmi: Interesting way to move forward! Really waiting for this PR to go in mainline after the 12.4 API snapshot happens.

PR #2925, opened 2021-04-29
Sync delivery to release for 12.4 rc2

geertjanw:


geertjanw:
9247.23 Test java.hints batch2 without nb-javac on Java 14
9247.27 Test java.hints batch2 without nb-javac on Java 15

Can we go ahead despite these test failures?

jlahoda:
> Can we go ahead despite these test failures?

Seems those were random failures (they didn't repeat after the last merge), so OK for now, I think. Eventually, we should stabilize the tests, so they wouldn't be failing randomly.

PR #2927, opened 2021-04-29
[NETBEANS-5640] Fixing potential NPE on out of project Folder creation

lkishalmi:
Well I get a nasty NPE when try to create a folder out of a project. This might be triggered by another bug in the system as it is unlikely to have a sourcegroup without a project though in the context of the ```SimpleTargetChooserPanelGUI``` class having the project field as ```null``` is indeed allowed. 
This is a trivial fix to prevent an NPE when that happen.

Unfortunately I do not know how to reliably reproduce this event. On one of my IDE instances it happens all the time, the other instance it does not.


lkishalmi:
Oh. This happens with12.4-rc1

geertjanw:
OK.

geertjanw (approved):

PR #2929, opened 2021-04-30
[NETBEANS-5495] [CND] Missing LongHashMap and UTF in cnd/repository

vieiro:
Module cnd/repository is missing LongHashMap and UTF classes, that were not donated. 
LongHashMap in java/java.source.base can be used as a substitute.
The UTF class is replaced with DataInputStream and an existing method.


matthiasblaesing:
Looks sane to me.

vieiro:
Let's merge and move on.

PR #2934, opened 2021-05-04
[NETBEANS-5655] Fix build error caused by non-ASCII quotes in an arch.xml file

eirikbakke:
I was doing a build on Windows today, and got the following error:

```
BUILD FAILED
C:\Users\ebakke\ZRoot\nbsrc\incubator-netbeans\nbbuild\build.xml:395: The following error occurred while executing this line:
C:\Users\ebakke\ZRoot\nbsrc\incubator-netbeans\nbbuild\javadoctools\build.xml:48: The following error occurred while executing this line:
C:\Users\ebakke\ZRoot\nbsrc\incubator-netbeans\nbbuild\javadoctools\build.xml:149: javax.xml.transform.TransformerException: javax.xml.transform.TransformerException: com.sun.org.apache.xml.internal.utils.WrappedRuntimeException: Invalid byte 3 of 3-byte UTF-8 sequence.
```

Examining build.xml:149, it points to the file nbbuild/build/javadoc/alldatas.xml, which indeed appears to have a malformed UTF-8 character somewhere.

Doing some further investigation...

```
ebakke@EBLEN:/z/nbsrc/incubator-netbeans/nbbuild/build/javadoc$ diff alldatas1.xml alldatas2.xml
5139c5139
<       when running in “compile-on-save? mode
---
>       when running in “compile-on-save�? mode
```

Easy to fix--just use regular ASCII quotes instead of MS Word style "smart quotes".

### DIFF
--- a/nbbuild/javadoctools/build.xml
+++ b/nbbuild/javadoctools/build.xml
@@ -125,7 +125,7 @@
         <echo file="${export.interfaces}/footer.gen">
             &lt;/alldata&gt;
         </echo>
-        <concat destfile="${netbeans.javadoc.dir}/alldatas.xml">
+        <concat destfile="${netbeans.javadoc.dir}/alldatas.xml" encoding="UTF-8" outputencoding="UTF-8">
             <fileset dir="${export.interfaces}" includes="header.gen"/>
             <fileset file="${netbeans.javadoc.dir}/allclasses.xml"/>
             <fileset file="${netbeans.javadoc.dir}/modules.xml"/>


jglick:
Something else is wrong then. https://github.com/apache/netbeans/blob/4fd403dc36af972e7d75ec4e96b1f59a37a4e155/nbbuild/javadoctools/build.xml#L121 looks right. Presumably you are running Windows in a non-UTF-8 locale and somewhere it is corrupting text.

eirikbakke:
Windows does not use UTF-8 by default on any locale that I know of--it uses e.g. Windows-1252 on U.S. English versions.

In theory it would be nice if tools knew how to deal with character encodings correctly, but it's an area where bugs are extremely common. And bugs often show up only in certain configurations. Sticking to ASCII characters will always work, though.

matthiasblaesing:
Sorry I disagree - sticking to ASCII will keep the computer world in a broken world. Unicode is 20+ years and we still discuss whether valid unicode handling can be worked around by using ascii? In this case it is even worse - one of the mortal sins of XML processsing happend: Instead of using an xml processor, plain text processing was used, ignoring encodings and such.

Assuming all input files are UTF-8 (ascii is by construction UTF-8, UTF-8 is the dafult encoding for XML), you could see if the ant concat task with platform encoding breaks this:

https://github.com/apache/netbeans/blob/4fd403dc36af972e7d75ec4e96b1f59a37a4e155/nbbuild/javadoctools/build.xml#L128-L139

`concat` has the parameters `encoding` and `outputencoding`. I would set both to `UTF-8`, this should match the default behavior on linux. That will still be broken, but slightly less (assuming it even helps).

eirikbakke:
Feel free to fix the build system, but sticking to ASCII is robust to a wide range of bugs--including bugs we haven't discovered yet, or bugs which only occurs on certain developers' machines. For special characters, better to use XML entities. But Smart Quotes in particular are known to be time bombs (see https://dwheeler.com/essays/quotes-in-html.html ).

jglick:
So glad to have started a holy war! :grin:

vieiro:
The file `java/java.source.base/arch.xml` is a perfectly a valid XML document in UTF-8 encoding (see attached screenshot). This is so in Windows, Unix or any other operating system (when you use the proper tools to edit UTF-8 documents).

If the build system on Windows is not able to handle UTF-8 documents properly then I'd suggest modifying the build system following the instructions above. 

Otherwise I'd suggest closing this issue, as the commit does not solve any problem with the file.

imagen

eirikbakke:
Does anyone feel a strong need to keep the smart quotes in the XML file?

Meanwhile, I'll rerun my local build with @matthiasblaesing's suggestion and see if that fixes the problem.

matthiasblaesing:
> Does anyone feel a strong need to keep the smart quotes in the XML file?

I see no need to remove them, as they are not the cause of the problem.
 
> Meanwhile, I'll rerun my local build with @matthiasblaesing's suggestion and see if that fixes the problem.

Thank you. While not 100% correct (valid XML, that is not ASCII and not UTF-8 will still broken), it will improve the situation.

eirikbakke:
> I see no need to remove them, as they are not the cause of the problem.

OK, then I will ask--can we _please_ remove them? They already caused an hour of extra work on my side...

eirikbakke:
@matthiasblaesing's suggestion fixes the problem on Windows! The patch is updated to include both fixes, to minimize the chance of problems in the future.

vieiro:
Great news! This means people adding UTF-8 stuff to arch.xml files in the future won't have the problem! 
I'd suggest restoring the "arch.xml" file to its original state (with UTF-8 stuff) so that we can double-check any other encoding problems in the future.

eirikbakke:
Let's not tempt the UTF gods any further.

matthiasblaesing:
I won't block this, but I agree with @vieiro it would be better to keep the Unicode characters in. They acted as a canary - the change caught a real bug.

eirikbakke:
I think it's a bigger issue that could be addressed and discussed separately. To act as a canary, and not just a cause of intermittent bugs for some developers (who may or may not report the problem, or may even just give up), problems like these would need to be triggered as part of the automated Windows build on Travis.

In fact I'm curious why it doesn't trigger in the Travis Windows builds. I suspect someone explicitly set UTF-8 as the OS-level encoding on the Windows CI images. That might well be a good idea, except then the build does not reproduce problems that occurs for Windows developers who do not change the default encoding on their machines. One could argue that the Travis build should be explicitly set to Windows-1252, to match Windows developers' typical configuration. This would rightly generate some debate, though...

So I'd prefer to keep the patch as is for now, but I wouldn't be opposed to adding back a canary later if the Travis builds are adjusted so that these kinds of bugs actually cause failures.

(Also, by the way, still a big thank you to @jglick for the original unrelated patch that was coincidentally involved here. It's been solving a lot of Compile-on-Save problems in my project.)

matthiasblaesing:
@eirikbakke you don't want to follow the advise given, that is fine, but then just get this in, as the additional change does not break things.

eirikbakke:
Happy to do one more revision of this. What's your current preferred course of action, given the various pros/cons presented in this thread?

matthiasblaesing:
What I wanted to say with my previous comments, was that according to your test the change to `nbbuild/javadoctools/build.xml` is enough to fix the issue.
The changes to `java/java.source.base/arch.xml` are therefore not necessary and not having them, lets the file act as a canary for the issue. I have no strong feelings here. If you want to change it, it is ok, but I (and also @vieiro) would prefer to keep the file as before this PR.
So in any case, the fix to `nbbuild/javadoctools/build.xml` should go in soon, with or without the change to `java/java.source.base/arch.xml`.

eirikbakke:
Sure, will put the smart quotes back in.

matthiasblaesing (approved):
Looks good to me. I suggest to rebase this changeset onto current master, to get the test fix (https://github.com/apache/netbeans/commit/4da79d0c10b6359c42441124c35b4d3abfc1bbda) in and merge with clean test results.

eirikbakke:
Yup, just rebased to master, will wait until Travis passes here before merging...

PR #2937, opened 2021-05-04
[NETBEANS-5658] Convert remaining editor toolbar icons to SVG

eirikbakke:
Here are 14 more SVG icons, drawn by myself, covering the remaining ones in the editor toolbar, and a few more locations. The previous SVG icons were added in https://github.com/apache/netbeans/pull/2387 . I also adjusted the colors in two of the existing editor toolbar icons slightly, and identified a few more similar-looking icons that could use already-drawn ones.

These icons follow the style guide at https://issues.apache.org/jira/browse/NETBEANS-2617 , like the previously committed icons. The JIRA description for NETBEANS-2617 also describes the process for finding duplicate icons.

Whenever the existing bitmap icons (PNG or GIF files) have duplicates, the new SVG icons have been copied into each location, as before. There have been various previous discussions about finding a way to avoid this duplication, but this can be done separately.

An overview of the icons converted so far, and the next icons that should be prioritized, can be found here:

Here are the icons specifically affected by this PR:
Table of Affected Files

Editor toolbar before:
toolbar-before

Editor toolbar after:
toolbar-after


matthiasblaesing (approved):
Looks looks sane to me and I checked license headers.

eirikbakke:
Let me run this in my test build for a while before merging. I'm seeing a new intermittent deadlock on editor startup on my latest NetBeans build--I'd like to make sure it's not related to the SVG loader.

eirikbakke:
Thanks for the review!

eirikbakke:
Deadlock seems to be unrelated, thread dump via VisualVM includes no mention of the SVG loader. The deadlock seems to have involved the "Git.cache" thread, which was holding a lock in the Project SPI, which the Event Thread was also waiting for while loading a previously open editor on startup. Deleting my cache directory caused NetBeans to start properly again. If I see this problem again I'll look into it.

matthiasblaesing:
Should we get this in?

eirikbakke:
Sure, I'll merge--thank you! I just wanted the PR to sit for a little bit in case people had lingering objections to the visual style. (Better to hear these early than after hundreds of icons have been drawn...)

PR #2938, opened 2021-05-05
Run/Debug single .java files outside of project in VSCode.

JaroslavTulach:
Single `.java` files without a project that contain main method should be executable/debuggable when one clicks on _Run main_ or _Debug main_. Steps to reproduce:

* start `code` with an empty directory as a folder
* Create a Java file in there with a `main` method.
* Click _Run main_ and check if it is executed
* Place breakpoint into the `main` method and check it is hit

The JEP-330 functionality works on any JDK11+. Older JDKs compile the class first with `javac` and then run the class with `java`. Once [PR-3939](https://github.com/openjdk/jdk/pull/3939) is integrated (presumably in JDK17) local variables will be visible when debugging via JEP-330.


java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/DebugProcess.java
```diff
@@ -23,25 +23,44 @@
...
 import org.netbeans.api.java.platform.JavaPlatformManager;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
 
-/**
- *
- * @author Arunava Sinha
- */
-class DebugProcess {
+final class DebugProcess implements Callable<Process> {
```
sdedic, line 34: Nitpick: `LaunchProcess` could be more appropriate as it does debug / run.

sdedic (approved):

entlicher (approved):
Please update the year in apichanges. Changes look very good, thanks.

ide/extexecution/apichanges.xml
```diff
@@ -92,6 +92,19 @@ is the proper place.
 
     <changes>
 
+        <change>
+            <api name="extexecution_api"/>
+            <summary>postExecution callback with exit code value</summary>
+            <version major="1" minor="61"/>
+            <date day="10" month="5" year="2020"/>
```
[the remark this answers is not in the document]
JaroslavTulach: ;-) Fixed 80dd766

java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/DebugProcess.java
JaroslavTulach: done in  4f9aedf

JaroslavTulach:
Merging, assuming the most recent change of `apichanges.xml` isn't going to break the build.

ide/extexecution/src/org/netbeans/api/extexecution/ExecutionDescriptor.java
```diff
@@ -622,11 +623,34 @@ Runnable getPreExecution() {
...
+     * <code>this</code>.
+     *
+     * @param postExecution post execution callback that receives exit code of the
+     *    execution, <code>null</code> allowed
+     * @return new descriptor with configured post execution callback
+     * @since 1.61
+     */
+    @NonNull
+    @CheckReturnValue
+    public ExecutionDescriptor postExecution(@NullAllowed Consumer<Integer> postExecution) {
```
JaroslavTulach, line 648: Get exit code when the execution is over.

PR #2788, opened 2021-03-02
Snapshot of APIs as of 12.3.

neilcsmith-net:
Snapshot of APIs as of 12.3. Replacement of #2690 Not to be merged until 12.3 vc1 vote passes.

Created with following steps -

```bash
# checkout branch (or release tag)
git checkout release123
ant clean
ant build
ant gen-sigtests-release
git add -A
git stash
git checkout master
git checkout -b apis-nb123
git stash pop
git add -A
git commit -m "Snapshot of APIs as of 12.3"
# remove getPeer() calls
find . -name "*.sig" -exec sed -i '/java.awt.peer.ComponentPeer/{N;d;}' {} \;
# check git diff
git add -A
git commit --amend
```

PR #2881, opened 2021-04-13
Upgrading to SigTest 1.4 and using its release attribute

jtulach:
New version of SigTest [supports the release attribute](https://github.com/jtulach/netbeans-apitest/pull/3). That allows us to fix the JDK API to particular version. The PR sets it to `${javac.target}` for each module. To verify how this behaves try for example:
```bash
platform/openide.util.lookup$ ant clean build check-sigtest -Djavac.target=11  -Dsigtest.check.type=strictcheck

Missed Annotations
------------------

org.openide.util.Lookup$Template.init:  anno 0 java.lang.Deprecated()
org.openide.util.lookup.implspi.AbstractServiceProviderProcessor.register:      anno 0 java.lang.Deprecated()

Added Annotations
-----------------

java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
java.lang.Object.finalize:              anno 0 java.lang.Deprecated(boolean forRemoval=false, java.lang.String since="9")
[...]

PR #2940, opened 2021-05-06
Snapshot of APIs as of 12.4

geertjanw:
Followed these instructions: https://github.com/apache/netbeans/pull/2788 (except the getPeer search command didn't work for me, hope that having them included is not a big issue).

### DIFF
--- a/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
+++ b/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
@@ -1,5 +1,5 @@
 #Signature file v4.1
-#Version 2.83
+#Version 2.84
 
 CLSS public java.lang.Object
 cons public init()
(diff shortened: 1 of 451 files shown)


neilcsmith-net:
@geertjanw looks like my `sed` command might be Linux specific - https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux 😬 

The other option is to remove manually - that's what was done before - I got fed up of doing it.  Ideally it wouldn't be generated, but @jtulach might have a view on an approach here.

geertjanw:
Can we go ahead with this PR without that, i.e., is it OK as it is right now?

JaroslavTulach:
Since #2881 the removal of `getPeer()` shall not be necessary. Even if one is running on JDK9+, we compare against APIs of JDK8 thanks to the `release` attribute of `Sigtest` task.

JaroslavTulach (approved):

geertjanw:
Can this be merged now?

neilcsmith-net:
@geertjanw I would normally regenerate on the basis of the release, in case there were any late additions.  Although as this ended up being close to vote I think it might be fine (ideally we'd do first iteration for review before freeze).  If you're confident it's picked up everything then should be merged, and I think spec version increment is still pending following on from merging this?

ebarboni:
redone in #3024

PR #2942, opened 2021-05-09
[NETBEANS-5587] Fix regression in gavSplit

lkishalmi:
Well, this one fixes a regression detected by Scott Palmer.
I do not think this would be a major issue. Probably would happen on a few strange Gradle projects.
I put this against delivery and marked it for 12.4, so if the RM-s decide we need an rc3, then it would be nice to include this one.

@sdedic I've changed the argument order in assert equals as the first parameter is the expected value, the second one is the one that shall be tested.


lkishalmi:
The regression has been introduced in #2895

sdedic (approved):
Thanks !

geertjanw (approved):

PR #2947, opened 2021-05-11
When java.lang is not available, avoid throwing a FatalError, but rather try to report an error to the user and continue.

jlahoda:
This is when running without nb-javac. Will work properly only when running on a new-enough JDK.


ebarboni:
error on travis after many retry. postponing

arvindaprameya:
Approved, thank you Jan

PR #2848, opened 2021-03-31
Branding API to disable Maven index download

JaroslavTulach:
Users of VSNetBeans complain about the Maven index download (some users of NetBeans IDE, like me, complain as well) and the fact that there is no UI to disable it. To address their complains right now, before some final solution[1] is implemented, I propose to introduce branding API and use it in the `nbcode` suite to disable the download in VSNetBeans.

[1] Ideally I'd like the queries to Maven to be handled online by sending a query to https://search.maven.org/#search|ga|1|guice & co.

PR #2950, opened 2021-05-12
LSP: Find usages should search also in project dependecies.

dbalek:
`textDocument/implementation` and `textDocument/references` LSP request implementation should include project dependencies in its searches for usages.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -123,6 +123,7 @@ public static boolean isClientResponseThread(NbCodeLanguageClient client) {
     }
     
     public static NbLspServer launchServer(Pair<InputStream, OutputStream> io, LspSession session) {
+        System.setProperty("org.netbeans.modules.java.source.usages.BinaryAnalyser.fullIndex", "true");
```
JaroslavTulach, line 126: I am not sure what is this, but if we want to configure behavior of some other module, we should do it via branding API, not system property.

java/java.source.base/src/org/netbeans/api/java/source/SourceUtils.java
```diff
@@ -834,7 +835,17 @@ public ClasspathInfo getClasspathInfo() {
...
-    
+
+    /**
+     * Returns true if the given file is a class file
+     * @param file
+     * @return true if the given file is a class file
+     * @since 2.51
+     */
+    public static boolean isClassFile(@NonNull final FileObject file) {
+        return FileObjects.CLASS.equals(file.getExt()) || ClassParser.MIME_TYPE.equals(file.getMIMEType(ClassParser.MIME_TYPE));
```
JaroslavTulach, line 846: Possibly add a unit test.

JaroslavTulach (changes requested):
System properties are a good way to change the behavior of the system from a command line, when launching the application. But using them to change the behavior while the application is running isn't bulletproof - it always depends whether the code that sets the property runs sooner than the code that reads the property.

Please change the behavior to `fullIndex` using a branding API like in case of #2848

java/java.source.base/src/org/netbeans/api/java/source/SourceUtils.java
dbalek: Unit test added.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
dbalek: Solved by providing `BinaryAnalyser.Config` implemementation in VSNetBeans integration module.

java/java.lsp.server/nbcode/integration/nbproject/project.xml
```diff
@@ -59,6 +59,21 @@
                         <specification-version>1.6</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.java.source.base</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <implementation-version/>
```
JaroslavTulach, line 67: Ugly. Can we avoid implementation dependency in `nbcode`? Can `java.lsp.server` expose enough, so `nbcode` is free of implementation dependencies?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspAnalyserConfig.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.netbeans.modules.java.source.usages.BinaryAnalyser;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ *
+ * @author Dusan Balek
+ */
+@ServiceProvider(service = BinaryAnalyser.Config.class, position = 1000)
+public class LspAnalyserConfig extends BinaryAnalyser.Config {
```
JaroslavTulach, line 29: Other `LspXYZ` services are provided as `public abstract` by `java.lsp.server` and then subclassed in the `nbcode` and registered with the `@ServiceProvider` - exactly to avoid exposing internal details (lsp4j, etc.) to `nbcode` suite.

JaroslavTulach (changes requested):
Avoid introducing implementation dependencies into `nbcode` suite.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
JaroslavTulach: Functionally better, but the implementation dependency!

java/java.lsp.server/nbcode/integration/nbproject/project.xml
dbalek: Implementation dependency removed.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspAnalyserConfig.java
dbalek: Done.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspAnalyser.java
```diff
@@ -0,0 +1,80 @@
...
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.modules.java.lsp.server.ui.AnalyserConfig;
+import org.netbeans.modules.java.source.usages.BinaryAnalyser;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ *
+ * @author Dusan Balek
+ */
+@ServiceProvider(service = BinaryAnalyser.Config.class, position = 1000)
```
sdedic, line 30: Q: should be this service active (registered) during regular IDE operation ? If not, the `@ServiceProvider` should move to `nbcode`.

dbalek: This service is intended to be active (registered) during both regular IDE and LSP operation. It replaces the default `BinaryAnalyser.Config` implementation. Its functionality can be adjusted form `nbcode` using `LSPAnalyserConfig` without implementation dependency of `nbcode` suite on `Java Source Base`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspAnalyser.java
```diff
@@ -0,0 +1,80 @@
...
+public final class LspAnalyser extends BinaryAnalyser.Config {
+    private static final String PROP_USG_LVL = "org.netbeans.modules.java.source.usages.BinaryAnalyser.usages"; //NOI18N
+    private static final String PROP_ID_LVL = "org.netbeans.modules.java.source.usages.BinaryAnalyser.idents"; //NOI18N
```
JaroslavTulach, line 33: It's quite unfortunate these properties need to be copied into `java.lsp.server` module. I don't have a better advice yet, but I am thinking about it.

JaroslavTulach: The easiest solution would be to really use branding API as suggested by [comment](https://github.com/apache/netbeans/pull/2950#pullrequestreview-659995669).

dbalek: OK. Branding API used as suggested.

java/java.source.base/src/org/netbeans/modules/java/source/usages/BinaryAnalyser.java
```diff
@@ -284,9 +285,10 @@ public IdentLevel getIdentLevel() {
                 return idLvl;
             }
 
+            @NbBundle.Messages("USE_FULL_INDEX=false")
```
JaroslavTulach, line 288: Usually we document this API in `arch.xml` like in case of #2848.

JaroslavTulach (approved):
Please document the branding token in `arch.xml`.

java/java.source.base/src/org/netbeans/modules/java/source/usages/BinaryAnalyser.java
dbalek: `USE_FULL_INDEX` branding token documented as requested.

PR #2958, opened 2021-05-16
[NETBEANS-5683] - remove last remaining HashSet raw type warnings..

BradWalker:
This change will remove the last remaining warnings about raw type HashSet.

   [repeat] /home/bwalker/src/netbeans/platform/core.multiview/src/org/netbeans/core/multiview/MultiViewActionMap.java:124: warning: [rawtypes] found raw type: HashSet
   [repeat]         Set keys = new HashSet();
   [repeat]                        ^
   [repeat]   missing type arguments for generic class HashSet<E>
   [repeat]     E extends Object declared in class HashSet

At this point, we have removed all warnings related to this.. Hooray!!


mklaehn (approved):
Method internal or private method signatures.
Good work.

PR #2960, opened 2021-05-17
Easier way to customize templates.

JaroslavTulach:
Ever annoyed by _To change this license header, choose License Headers in Project Properties._ and _To change this template file, choose Tools | Templates_ let's replace that with a clickable hyperlink:
image
The `nbfs:` URL would be recognized as a hyperlink and a click would open the template file in the editor.


sdedic (approved):
Perfect, although I don't like that `nbfs:` sufacing right out to the user. Sadly, the Fold APIs do not support actionable folds (they already support content preview) - so [NETBEANS-5684](https://issues.apache.org/jira/browse/NETBEANS-5684).

JaroslavTulach:
> I don't like that `nbfs:` sufacing right out to the user. 

One of my earlier prototypes was using `config:/Templates/Classes/Class.java` - however then I decided to go public without introducing new protocol. That decision can change as this is really very visible, user facing feature.

junichi11:
URL is not highlighted.

#### Before:
netbeans-pr-2960-before


#### After:
netbeans-pr-2960-after

#### Exception:
```
java.lang.UnsupportedOperationException: Not supported yet.
	at org.netbeans.modules.editor.url.HighlightURLs$3.subSequence(HighlightURLs.java:217)
	at java.util.regex.Matcher.getSubSequence(Matcher.java:1294)
	at java.util.regex.Matcher.group(Matcher.java:541)
	at org.netbeans.modules.editor.url.Parser.recognizeURLsREBased(Parser.java:44)
	at org.netbeans.modules.editor.url.Parser.recognizeURLs(Parser.java:36)
	at org.netbeans.modules.editor.url.HighlightURLs.run(HighlightURLs.java:222)
Caused: org.openide.util.RequestProcessor$FastItem: task failed due to
```

Maybe `recognizeURLs()` is called many times, I suppose. Is the performance problem OK although I can be wrong?
(about using `URL_PATTERN.matcher(text)` instead)

JaroslavTulach:
With the most recent change I can open the `nbfs:` URLs even in VSCode:

obrazek

enterprise/micronaut/src/org/netbeans/modules/micronaut/hyperlink/HyperlinkImpl.java
```diff
@@ -0,0 +1,290 @@
+/*
```
dbalek, line 1: I do not like location of this file inside of `enterprise/micronaut` module. I would prefer leaving this file as a part of `ide/editor` module and adding its dependency on `ide/api.lsp` if necessary.

enterprise/micronaut/src/org/netbeans/modules/micronaut/hyperlink/HyperlinkImpl.java
```diff
@@ -0,0 +1,290 @@
...
+                } catch (IOException ex) {
+                    Exceptions.printStackTrace(ex);
+                }
+            }
+        }
+
+        return null;
+    }
+    
+    @MimeRegistration(mimeType = "text/x-java", service = HyperlinkLocationProvider.class)
```
dbalek, line 270: Try to use empty mime path to register this provider for all files.

dbalek (approved):
Very useful feature. Thanks.

JaroslavTulach:
There is a way to [registerTextDocumentContentProvider](https://code.visualstudio.com/api/references/vscode-api#workspace.registerTextDocumentContentProvider) - probably it should be used to handle the `nbfs` protocol and delay materialization of the file on disk until really needed - thanks for finding this out, @dbalek!

JaroslavTulach:
OK, @junichi11 - RegExp in Java is known to be quite slow:

> Test Result: times=2049 vs. 19441

As such I am keeping the old manually written parser: ce7a42a - using the old parser should also fix the problem with 

> URL is not highlighted.

it seems, it is highlighted now.

enterprise/micronaut/src/org/netbeans/modules/micronaut/hyperlink/HyperlinkImpl.java
JaroslavTulach: That didn't work, but hopefully we won't need this part at all.

enterprise/micronaut/src/org/netbeans/modules/micronaut/hyperlink/HyperlinkImpl.java
JaroslavTulach: Removed in ee9eb43

enterprise/micronaut/src/org/netbeans/modules/micronaut/hyperlink/Parser.java
```diff
@@ -0,0 +1,65 @@
+/*
```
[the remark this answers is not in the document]
JaroslavTulach: Done. `editor` now depends on `lsp.api`.

junichi11 (approved):
Perfect!
I've verified 
- there is no exception
- paths are highlighted
- jump to templates when I ([Ctrl]+)click paths

Thank you!  
nb-pr-2960

JaroslavTulach:
In addition to here-in provided work I can also offer [delegating hyperlink provider](https://github.com/apache/netbeans/commit/a07f2f8698b462b728cbf976bb5398d16e434f6b) that can bridge `lsp.api` and NetBeans editor. However review by @dbalek suggested to keep that change out of this PR. Including here for completeness. The code may be handy in the future.

PR #2961, opened 2021-05-17
Fix bug on using a platform dependent path separator

digiovinazzo:
The system property `cluster.path.final` is set when running unit tests, but when on Windows the last cluster in the list is separated using `:`. Fixed using the platform dependent system property `path.separator`.


sdedic (approved):
No review from more knowledgeable person for 3 weeks. Seems sane to me, let's get it in.

PR #2964, opened 2021-05-19
[NETBEANS-5693] Assure that debugger session finishes when native image application exits.

entlicher:
Assure that we notify the finish of the debug/run session through LSP.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
@@ -40,6 +40,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
 
 import org.netbeans.api.annotations.common.CheckForNull;
 import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.api.annotations.common.NullAllowed;
 import org.netbeans.api.debugger.DebuggerManager;
 import org.netbeans.api.debugger.DebuggerManagerAdapter;
 import org.netbeans.api.debugger.Session;
@@ -60,7 +61,6 @@ import org.netbeans.modules.java.lsp.server.progress.OperationContext;
 import org.netbeans.modules.java.lsp.server.progress.ProgressOperationEvent;
 import org.netbeans.modules.java.lsp.server.progress.ProgressOperationListener;
 import org.netbeans.modules.java.lsp.server.progress.TestProgressHandler;
-import org.netbeans.modules.java.lsp.server.protocol.NbLspServer;
 import org.netbeans.modules.java.nativeimage.debugger.api.NIDebugRunner;
 import org.netbeans.modules.nativeimage.api.debug.NIDebugger;
 import org.netbeans.spi.project.ActionProgress;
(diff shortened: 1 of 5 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
```diff
@@ -200,30 +201,43 @@ public void progressHandleCreated(ProgressOperationEvent e) {
...
                     });
                 });
             } else {
+                AtomicReference<Future<Integer>> run = new AtomicReference<>();
                 Lookups.executeWith(launchCtx, () -> {
-                    execNative(nativeImageFile, args, context, executionDescriptor, launchFuture);//, success);
+                    run.set(execNative(nativeImageFile, args, context, executionDescriptor, launchFuture));
+                });
+                new RequestProcessor().post(() -> {
+                    Integer ret;
```
JaroslavTulach, line 217: This looks like the old, ugly pattern. Since https://github.com/apache/netbeans/pull/2938/files#r635969684 once can get the `exitCode` (or `null`) synchronously as soon as execution is over.

JaroslavTulach (changes requested):
Obtain the `exitCode` synchronously.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
entlicher: That significantly simplified the code, thanks.
And it uncovered a bug in cpplite debugger - the GDB process did not finish after application exit. This is fixed, including a test, as well.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
```diff
@@ -198,16 +197,26 @@ public void progressHandleCreated(ProgressOperationEvent e) {
...
-                        startNativeDebug(nativeImageFile, args, miDebugger, context, executionDescriptor, launchFuture);
+                        startNativeDebug(nativeImageFile, args, miDebugger, context, ed, launchFuture, debugProgress);
                     });
                 });
             } else {
-                Lookups.executeWith(launchCtx, () -> {
-                    execNative(nativeImageFile, args, context, executionDescriptor, launchFuture);//, success);
+                ExecutionDescriptor ed = executionDescriptor.postExecution(exitCode -> {
+                    ioContext.stop();
+                    notifyFinished(context, exitCode == 0);
```
JaroslavTulach, line 216: The `exitCode` can be `null` in case there is an error launching the executable, etc. This would probably yield a `NPE`.

JaroslavTulach (approved):
Beware of potential NPE.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
entlicher: That's a catch! It's unexpected. This fact is not clear from the Javadoc. The annotation and the Javadoc says, that `postExecution` argument is allowed to be `null`. Not that the exit code passed to the Consumer can be `null`.
It'd be good to explicitly stress that.
Unfortunately, the functional features hurt the safeness of Java. It's impossible to put `@NullAllowed` on the `exitCode` parameter. I'd need to do it myself (which misses the point).

entlicher: Null check added.

JaroslavTulach: I've run into the problem myself as well. Then I checked the Javadoc and saw the `null` note. Then I have just fixed my code. Feel free to improve the javadoc or enhance the API.

PR #2965, opened 2021-05-19
[NETBEANS-5697] Various LAF/HiDPI improvements on Windows

eirikbakke:
This PR includes a few smaller improvements that were made to the Windows LAF while working on modernized tab components. Several of these relate to NetBeans' appearance when HiDPI scaling is active. Summary:

- Apply previous border tweaks for HiDPI screens in a few more places, notably in tooltips and the quick search text field, and in the editor toolbar. Improve JSpinner borders.
- Make the background color of tooltips consistent with that of other Windows 10 apps.
- Add some pixels of left margin to the editor toolbar.
- More improvements added in another commit, see the second pair of screenshots further below.

The screenshots below are from _before_ the fixes were applied.

Smaller fixes

- Also apply the workarounds from NETBEANS-3592 to JCheckboxMenuItem and friends. This fixes the following problem, where menu item checkmarks can become tiny or huge in certain situations (due to HiDPI-related bugs in Swing's Windows LAF, see https://bugs.openjdk.java.net/browse/JDK-8211715 ):

210517 Menu Checkbox Bug

### DIFF
--- a/ide/projectui/src/org/netbeans/modules/project/ui/actions/ActiveConfigAction.java
+++ b/ide/projectui/src/org/netbeans/modules/project/ui/actions/ActiveConfigAction.java
@@ -488,7 +488,6 @@ public class ActiveConfigAction extends CallableSystemAction implements LookupLi
         private Border defaultBorder = getBorder();
         
         ConfigCellRenderer() {
-            setOpaque(true);
         }
 
         public @Override Component getListCellRendererComponent(JList list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
@@ -517,10 +516,15 @@ public class ActiveConfigAction extends CallableSystemAction implements LookupLi
             if ( isSelected ) {
                 setBackground(list.getSelectionBackground());
                 setForeground(list.getSelectionForeground());             
+                setOpaque(true);
             }
             else {
                 setBackground(list.getBackground());
                 setForeground(list.getForeground());
+                /* Avoid painting a background that does not match the rest of the JComboBox, in
+                particular when we are painting the item in the toolbar box rather than in the popup
+                list. */
+                setOpaque(false);
             }
             
             return this;
--- a/platform/core.windows/src/org/netbeans/core/windows/view/ui/toolbars/ToolbarConfiguration.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/view/ui/toolbars/ToolbarConfiguration.java
@@ -593,12 +593,6 @@ public final class ToolbarConfiguration implements ToolbarPool.Configuration {
         mid = new Color(r, g, b);
     }
 
-    private static boolean isWindows8() {
-        String osName = System.getProperty ("os.name");
-        return osName.indexOf("Windows 8") >= 0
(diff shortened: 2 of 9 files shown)


DevCharly:
You could also remove bottom (and top?) separator of the main toolbar:

image

It is created here:

https://github.com/apache/netbeans/blob/dd5eb0d56e742e3a2e1c3566d64c0f20e78bee0b/platform/core.windows/src/org/netbeans/core/windows/view/ui/toolbars/ToolbarConfiguration.java#L621-L634

Looking at that code it actually seems to be a bug that the toolbar has top and bottom borders on Windows 10, because it does not have borders on Windows 8...

eirikbakke:
Good idea, will try it... (in fact, I think I already removed that border in my own NetBeans Platform application)

eirikbakke:
I added another commit with more improvements around the toolbar and menu bar area, including the proposed border change. Here is the latest look, with the newest changes annotated, at 150% HiDPI scaling:

More updates annotated

Here is the same window at 100% HiDPI scaling, without the annotations:

More updates

DevCharly:
that looks much better 👍 

I think you can remove method `ToolbarConfiguration.isWindows8()`. Seems to be unused now.

rasa-app:
IMHO, removing bottom border of toolbar completely, causes a kind of irregularity because toolbar icons messing with other icons, I think it's better to remove just shadows and keep a simple border or just like FlatLaf, make the toolbar background a little darker or lighter and  make a distinguish between toolbar and other parts.

eirikbakke:
Here's a mockup with a line under the toolbar, without adding spacing:

line

I don't think the line works very well here. It makes things look too cramped--alternatively it will require more margin to be added back around the toolbar, which consumes precious vertical space. Alternatively, different background colors don't look consistent; it's kind of like using too many fonts in a text document. Plus, whenever there's a change in the background color you need additional margin.

(On the topic of "change in background color", I also experimented with making the menu bar grey, or the toolbar white, to avoid the change between the menu bar and the toolbar. But the menu bar is always white on Windows 10, and a white toolbar looked odd.)

I do understand what you mean by "irregularity" with the space next to the tabs. But most of the time this will be filled up by tabs, so the typical appearance of the IDE won't have this problem.

eirikbakke:
I did a force push just to remove the now-unused ToolbarConfiguration.isWindows8() method, as earlier suggested. Other than that I think the discussion items were resolved here (judging by rasa-app's thumbs-up emoji :-).

I have used this patch in my working IDE and platform app for two months, so I think it should be safe to merge (once Travis passes again)...

matthiasblaesing:
This looks like a sane change, has positive reviews, does not introduce a revolution and seems to pretty much contained and is tested. I suggest to get this in now.

@eirikbakke minor comment for the future: I would add the issue marker to the commit messages themselves - its the individual commits, that are shown when running `git blame` (or the using the corresponding netbeans action) and it would make it easier to find the issue that motivated a change.

eirikbakke:
Thanks! @matthiasblaesing Sure, I will reference the issue in each individual commit message in the future.

PR #2966, opened 2021-05-19
[NETBEANS-5696] HiDPI-scalable "spinner" icon for window system tabs

eirikbakke:
The window system tabs component has a spinning "busy" icon which is displayed e.g. while the Navigator is initially loading. This issue's Pull Request replaces the current bitmap-based icon (busy_icon.png) with a scalable spinner graphic (drawn with Java2D code) that looks better on HiDPI/Retina screens. This benefits all LAFs. Screenshot attached below.

New Spinner Icon

The spinner graphic is made to look identical to the spinner that is shown on the MacOS/Aqua LAF when a JProgressBar is used with the JProgressBar.style=circular client property.


eirikbakke:
I've had this patch running in my working IDE for two months now, and I think it should be safe to merge.

PR #2967, opened 2021-05-19
[NETBEANS-5673] Modernize tab controls in the Windows LAF

eirikbakke:
This PR modernizes the look of the main NetBeans tabs components on the Windows LAF, and makes them look good on all the standard HiDPI scalings supported on Windows (100%, 125%, 150%, 175%, 200%, and 225%).

A before/after comparison is shown here, first at 100% scaling (regular, non-HiDPI screens). To review these images, be sure to view them at 100% scaling on your screen. (If you have a HiDPI screen yourself, you may need to adjust your browser's zoom level so that each pixel in the screenshot corresponds to one device pixel on your screen.)

Old vs  New Tab Style (1x scaling)

The new tabs adhere to the following design principles:
* Keep the tabs consistent with the style in native Windows 10 dialogs. This style is also used in the regular JTabbedPane, which is used in NetBeans e.g. in the "Output" pane.
* Keep the tab dimensions compact, as in the previous look. After trying various geometries, I ended up making the new tabs 2 pixels taller than the previous ones, but not more.
* Minimize visual complexity, while making it obvious where the user can click. Keep drawing the selected tab as an actual tab, with borders around it, but show a blue line highlight as seen in FlatLAF. Use the "flat" style from FlatLAF for unselected tabs. For the editor tab component, use the subtle separators that were introduced in FlatLAF between unselected tabs. This is also the style used in Chrome tabs. Make sure the "X" icon is not closer to the separator line than to the tab's caption; otherwise, the visual interaction between the X and the separator line can become confusing.

The new tabs also solve various problems that were previously seen when HiDPI scaling was active, especially for fractional scalings (e.g. 150%). Here are before/after screenshots on 150% HiDPI scaling:

Old vs  New Tab Style (1 5x scaling)

The new TabDisplayerUI classes are based on @DevCharly's previously added implementations for FlatLAF. These were already taking HiDPI scaling into account, and were very configurable. For now, I have added separate copies of these classes for the Windows LAF. I have split up the commits in this PR to make it clear which changes have been made to the "forked" copies. Some improvements, found in the second commit, could potentially be merged back into the FlatLAF classes if later desired. For example, one bug was fixed relating to the positioning of the "spinner" icon which is shown when the Navigator tab is being loaded:

Spinner position fix

To review this PR, it's best to look at each commit separately, since the first one just makes copies of the files from FlatLAF (with minimal necessary refactorings). The commits should probably be left unsquashed.

Some related improvements were split out into separate PRs, at https://github.com/apache/netbeans/pull/2965 and https://github.com/apache/netbeans/pull/2966 .

Finally, here are some screenshots of the entire NetBeans IDE with the old and the new tabs, this time at 200% HiDPI scaling.

### Before:
Full Screenshot Old Style Tabs (2x scaling)

### After:
Full Screenshot New Style Tabs (2x scaling)

### DIFF
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
@@ -90,7 +90,7 @@ final class Windows8VectorTabControlIcon extends VectorIcon {
     }
 
     private Windows8VectorTabControlIcon(int buttonId, int buttonState) {
-        super(14, 14);
+        super(14, buttonId == TabControlButton.ID_CLOSE_BUTTON ? 15 : 14);
         this.buttonId = buttonId;
         this.buttonState = buttonState;
     }
@@ -118,10 +118,14 @@ final class Windows8VectorTabControlIcon extends VectorIcon {
                 bgColor = closeColor != null ? closeColor : new Color(57, 100, 178, 255);
                 fgColor = Color.WHITE;
             } else if (buttonState == TabControlButton.STATE_ROLLOVER) {
-                bgColor = closeColor != null ? closeColor
-                        // Grey (via transparent black to work well on any background).
-                        : new Color(0, 0, 0, 70);
-                fgColor = Color.WHITE;
+                if (closeColor != null) {
+                    bgColor = closeColor;
+                    fgColor = Color.WHITE;
+                } else {
+                    /* Light blue, via transparency to work well on any background. In the grey
+                    toolbar, it comes out similar to the hover background used for toolbar buttons. */
+                    bgColor = new Color(0, 132, 247, 49);
+                }
             }
         }
         if (bgColor.getAlpha() > 0) {
@@ -142,7 +146,7 @@ final class Windows8VectorTabControlIcon extends VectorIcon {
                 strokeWidth *= 1.5f;
             }
             double marginX = 3.5 * scaling; // Don't round this one.
-            int topMarginY = round(3 * scaling);
+            int topMarginY = round(4 * scaling);
             int botMarginY = round(4 * scaling);
             // Flatten the top and bottom.
             g.clip(new Rectangle2D.Double(0, topMarginY, width, height - topMarginY - botMarginY));
--- a/platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/windows8/DPIUnscaledBorder.java
+++ b/platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/windows8/DPIUnscaledBorder.java
@@ -23,6 +23,7 @@ import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.Insets;
 import java.awt.geom.AffineTransform;
+import javax.swing.JTextField;
 import javax.swing.border.Border;
 import javax.swing.plaf.UIResource;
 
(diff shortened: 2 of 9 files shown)


platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
```diff
@@ -43,7 +43,7 @@
  * @author Eirik Bakke
  */
 @SuppressWarnings("serial")
-final class Windows8VectorTabControlIcon extends VectorIcon {
+public final class Windows8VectorTabControlIcon extends VectorIcon {
```
eirikbakke, line 46: No need to make this public... I'll get rid of the "public" here.

DevCharly:
Have tested this PR and it works and looks great 👍 

What I actually don't like is that the small control icons (minimize, scroll left/right, etc) are inverted when moving the mouse over them:

image

I think for a hover feedback this is too much. Would make it more "discreet" with lighter gray and not inverting icon.
Or use light blue as in toolbars or tabs:

image

DevCharly:
Not sure whether it is related to this PR, but just noticed following:

netbeans-win-docs-menu-issue

eirikbakke:
Hover colors are easy to tweak--I'll have a look. Any opinions on whether the "close tab" button should stay red, or just have a grey? The red color was copied from an earlier version of Chrome, but the latest chrome versions switched it to grey...

I'll check the dropdown bug too.

DevCharly (approved):
Looks good to me 👍

geertjanw:
Ready to merge?

eirikbakke:
Oh, sorry, let me check the outstanding issues today...

eirikbakke:
The popup switcher bug was unrelated to this PR, and will probably require separate work/review. I also found a more serious bug in the popup switcher, which could make the IDE unresponsive (also unrelated to this PR). I added the following JIRA issues for these:

[NETBEANS-5853 Repaint problems in tabcontrol ButtonPopupSwitcher](https://issues.apache.org/jira/browse/NETBEANS-5853)
[NETBEANS-5855 Closing unsaved editor tab from dropdown can put IDE in unusable state](https://issues.apache.org/jira/browse/NETBEANS-5855)

Will now tweak the hover color...

eirikbakke:
Here's the tweaked hover color...

New hover color

Will do a force push to update this.

eirikbakke:
I made the following changes in a force push:

```
diff --git a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
index 7d1465a..a7131f7 100644
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/Windows8VectorTabControlIcon.java
@@ -43,7 +43,7 @@ import javax.swing.Icon;
  * @author Eirik Bakke
  */
 @SuppressWarnings("serial")
-public final class Windows8VectorTabControlIcon extends VectorIcon {
+final class Windows8VectorTabControlIcon extends VectorIcon {
     private static final Map<Entry<Integer,Integer>,Icon> INSTANCES = populateInstances();
     private final int buttonId;
     private final int buttonState;
@@ -118,10 +118,14 @@ public final class Windows8VectorTabControlIcon extends VectorIcon {
                 bgColor = closeColor != null ? closeColor : new Color(57, 100, 178, 255);
                 fgColor = Color.WHITE;
             } else if (buttonState == TabControlButton.STATE_ROLLOVER) {
-                bgColor = closeColor != null ? closeColor
-                        // Grey (via transparent black to work well on any background).
-                        : new Color(0, 0, 0, 70);
-                fgColor = Color.WHITE;
+                if (closeColor != null) {
+                    bgColor = closeColor;
+                    fgColor = Color.WHITE;
+                } else {
+                    /* Light blue, via transparency to work well on any background. In the grey
+                    toolbar, it comes out similar to the hover background used for toolbar buttons. */
+                    bgColor = new Color(0, 132, 247, 49);
+                }
             }
         }
         if (bgColor.getAlpha() > 0) {
```

Once Travis tests pass, this should be safe to merge (without squash, as the 3 commits are meant to be separate). I've been running this patch in my own working IDE, and in my platform app, for two months now without issue.

PR #2968, opened 2021-05-20
[NETBEANS-5699] - Fixed test cases failing on JDK17

mishrasandeep:
https://issues.apache.org/jira/browse/NETBEANS-5699

1. com.sun.java.util.jar.pack.Attribute removed since JDK 14. Added new pass file for the same.
2. getNextVisualPositionFrom(int) method in javax.swing.plaf.basic.BasicTextUI class does not allow values greater than document length since JDK 9. Modified test method testModelToViewAtBoundaries for the same.


singh-akhilesh (approved):
Changes looks fine

Akshay-Gupta-Oracle (approved):
Looks correct.

PR #2969, opened 2021-05-20
Support for configurations in Gradle

sdedic:
The basic idea is:
- configuration can pass some **project properties** (-Px=y) to the gradle build (all tasks/actions)
- configuration can pass some additional **arguments** to the build (all tasks/actions)
- actions can have different mapping in a configuration

Currently this is used to enable Gradle's `--continuous` mode. Should the user choose that project configuration, `Run` starts gradle with `--continuous`, but `Debug` works as usual.

Still missing UI to manipulate / customize the configurations.

gradle

Note: this is based on the Maven configurations branch, which is going to be merged soon. Will rebase afterwards.


sdedic:
Summary of changes for reviewers:

- `GradleExecConfiguration` introduced. Final class that describes the configuration. Modifications possible from Gradle projects module only.
- `--continuous` flag moved to PARAM category (was UNSUPPORTED)
- `RunConfig` and `RunUtils` can accept `GradleExecConfiguration` as an additional parameter
- `action-mapping.dtd` now permits "profiles" similar to Maven implementation; so `GradleActionProvider`s can include configurations
- configurations (private, shared) are persisted using `AuxiliaryConfiguration` of the project (XML in the project dir for shared, XMLstring file attribute for private)
- action customizations for **default** configuration are placed in `gradle.properties`; for other configurations customizations are persisted as XMLs in the project directory (just like Maven)

`Java Gradle Projects` module now includes 'Continuous mode' configuration that adds `--continuous` to `run` and `run.single` actions.   The selected configuration is stored in project's AuxiliaryProperties, non-shared. The `--continuous` stuff is plugin-dependent; not all plugins support it, so let's have java covered at least.

lkishalmi:
Seems fine on the first sight. Need to check this in action. Had plans to implement this feature, so I'm glad I have not started to work on this one. In my mind I've expected less code to implement this, but that does not mean anything.

In order not to clash, my near future plans are: 
  * Support for Java Toolchain Detection (Gradle feature in 6.7+)
  * Better Test support
  * Rebase my Build NetBeans with Gradle branch on 12.4

extide/gradle/arch.xml
```diff
@@ -268,6 +268,13 @@ publishPlugins&lt;/args&gt;
 &lt;/actions&gt;
              </pre>
          </api>
+         <api category="devel" group="java.io.File" name="nb-actions-config.xml" type="export">
+             <p>
+                 An extension of <a href="#nb-actions-config.xml">nb-actions.config</a> is a per-Configuration action mapping.
```
JaroslavTulach, line 273: I convinced Laszslo in the past to not expose `nbactions*.xml` files and rather encode the actions into `build.properties` file. The introduciton of `nb-actions-config.xml` goes against that decision. It is OK, if you and Laszlo are OK with that. I just have to mention it.

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleExecAccessor.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Map;
+import org.netbeans.modules.gradle.api.execute.GradleExecConfiguration;
+import org.openide.util.Exceptions;
+import org.openide.util.NbBundle;
+
+/**
+ * Accessor to manipulate {@link GradleExecConfiguration} in ways not exposed in the API.
+ * @author sdedic
+ */
+public abstract class GradleExecAccessor {
```
JaroslavTulach, line 31: Accessor! Every good API needs an accessor!

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -2337,6 +2341,29 @@ private static boolean isArchiveFileImpl(final URL url, final boolean strict) {
...
+     * Hardcoded offending filename characters. Windows are more picky than Unixes; relying
+     * on just the current host OS will limit file portability when shared through VCS.
+     */
+    private static final Pattern ILLEGAL_FILENAME_CHARACTERS = Pattern.compile("[\\/:\"*?<>|]");
+    
+    /**
+     * Determines whether the string forms a valid filename (without a path component).
+     * @param fileName candidate string
+     * @return true, if the string can be used to name a file.
+     */
```
JaroslavTulach, line 2355: This is an API change and it comes with no versioning.

JaroslavTulach (approved):
Document, version the API change in `openide.filesystems`. Otherwise fine from my point of view.

extide/gradle/arch.xml
sdedic: This is a good topic for a discussion, I believe: at this point, the behaviour is +- consistent with Maven projects (which does not say anything about being valid). In a purist's view, the action mapping does not belong to the project shared metadata at all: 
- if an option (task, argument) is to be seen by the project itself, it should be put to the buildscript (pom) in a native way.
- if the project team agrees to share IDE-related metadata, it's best to have them file-separated, IMHO, from the project file(s) that are used by CI
- if the action mapping is not to be shared, it's best to put on `.gitignore` list; 

You're right that people frown upon `nbproject` and the like directories in VCS, but I see people routinely put their `.idea` directories to `.gitignore` lists. In a sense, its current impl that merges `gradle.properties` values that might affect CI with IDE-related setting may be also frowned upon. 

So I would stick to that pattern, and **document** to put 
```
**/nb-actions*.xml
**/nb-configuration*.xml
```
to `.gitignore` -- that would work for both Maven and Gradle.

@lkishalmi  -- what's your take on this ? I can +- easily change impl to persist all mappings to `gradle.properties` (using other prefixes); but that would eventually put there even customizations for private configurations (not sharable through `nb-configurations.xml`). I have no strong preference, as I can believe see benefits (or better consistency) of either solution, but slightly inclined to the separated files (although possibly in a subdir to limit the project pollution).

Re. this IDE-metadata as a general topic (should we create, should it be collected in a subdir, how to tackle private info - see [NETBEANS-5711](https://issues.apache.org/jira/browse/NETBEANS-5711): shouldn't it be debated on a mailing list instead ?

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
sdedic: Well. I was about task if, in your opinion, is this utility worth adding at all. But I've found ~9 occurrences of invalid filename characters (they differ, naturally ;)) 
In addition, I have found this interesting post - https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names ... really, you just can't make `PRN` folder under Windows :-D so I guess we could eventually fit more checks even in the `FileUtil.findFreeFileName` ...
So I'll keep the utility + [NETBEANS-5720](https://issues.apache.org/jira/browse/NETBEANS-5720), [NETBEANS-5721](https://issues.apache.org/jira/browse/NETBEANS-5721) in addition.

sdedic: Addressed in #b623dd4b129c

extide/gradle/arch.xml
lkishalmi: For me it is kind of indifferent. I see value in both approaches. My day job is as DevOps engineer, I usually can make either way accepted in source repositories.
I'd open a discussion on the dev (maybe even the users) mailing list(s).

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
lkishalmi: Also that would be better to be added in a different PR. This one is getting a bit bloated...

sdedic:
Hm, one more forgotten local file added (that was the reason for failing `extide` tests)

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
sdedic: @lkishalmi the initial impl of the utility is a separate commit here. The retrofit **will** be a separate PR (as well as the possible extension of the checks). I wonder :)) what happens when I place `CON` files to git on Linux and clone it to Windows ... ;)

entlicher:
The changes look fine to me.

entlicher (approved):

sdedic:
OK, 2 approvals one abstain, 72hours from last approval without further objections. The failing tests **do relate** to Gradle, but *do not relate to the current set o chages. Cause identified in platform - will make new PR for this.

Merging.

PR #2970, opened 2021-05-20
[NETBEANS-5698] Fix tabcontrol painting on MacOS retina displays (Aqua LAF)

eirikbakke:
This PR fixes scaling problems that occur in the window system tabs on the Aqua LAF on Retina displays (2x HiDPI scaling). All MacBooks and other Apple displays use retina scaling these days.

The fix is to use Graphics.fillRect instead of Graphics.drawLine; see the comment in the code. The patch adds a new "drawLine" utility method in the tab displayer UI classes for this purpose, and then does a plain search-and-replace of "g.drawLine(" with "drawLine(g, ", with no other changes.

Note that on MacOS, standard UI panel borders are two device pixels thick on Retina screens. So scaling everything up 2x on Retina displays is the right thing to do, and looks consistent with native MacOS apps.

The fixes are illustrated below, with retina scaling (2x) enabled. When viewing these images, be sure to view them at a zoom level that makes the individual pixels apparent (i.e. 100% scaling on non-HiDPI screens).

Summary

Tested on both Java 8 and Java 11. Full IDE screenshots included below.

### Before (opening in a separate tab is recommended):
Java 11 Tabs Old

### After:
Java 11 Tabs New


hectorespert (approved):

eirikbakke:
Merging after approval--thank you, @hectorespert !

PR #2971, opened 2021-05-21
LSP: Get project source roots, classpath, and packages commands added.

dbalek:
Adding the following LSP commands:
- `java.get.project.source.roots`
- `java.get.project.classpath`
- `java.get.project.packages`


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
```diff
@@ -136,6 +138,81 @@
...
+                            for (FileObject root : ClasspathInfo.create(sourceGroup.getRootFolder()).getClassPath(kind).getRoots()) {
+                                roots.add(root);
+                            }
+                        }
+                    }
+                    return roots.stream().map(fo -> Utils.toUri(fo)).collect(Collectors.toList());
+                });
+            }
+            case Server.JAVA_GET_PROJECT_PACKAGES: {
+                String uri = ((JsonPrimitive) params.getArguments().get(0)).getAsString();
```
sdedic, line 190: Nitpick: I have a vague feeling that all three code blocks could be factored out to a common method like
```
CompletableFuture<List<?>> collectFromGroups(String uri, String groupType, Function<SourceGroup, Object>)
```

sdedic (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
dbalek: Common code factored out to `CompletableFuture<List<FileObject>> getSourceRoots(String uri, String type)`.

PR #2972, opened 2021-05-21
Sync timestamps in test. Detailed logging added.

sdedic:
Call me dumb: a while after I merged [PR-2924](https://github.com/apache/netbeans/pull/2924), the [NbGradleProjectImplTest](https://github.com/apache/netbeans/compare/master...sdedic:gradle/loadtest-fix?expand=1#diff-72dfc9a2d87b3278ba23d2642ce186c38cbd15e9b7795ce23d9d1b9850eadb6b) started to occasionally fail. The cause is that while the test makes a copy of the project dir, it does not maintain file timestamps - so cache may get copied a little earlier than the build.gradle file.

In addition, I've discovered [this nice issue in JDK](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8177809)...

I have added some detailed logging during debugging, that could be worth to preserve for future diagnostics of the loading process

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
@@ -75,7 +75,7 @@ import org.openide.util.lookup.ProxyLookup;
  */
 public final class NbGradleProjectImpl implements Project {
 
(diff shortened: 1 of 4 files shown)


JaroslavTulach (approved):

lkishalmi (approved):
Looks good. Thanks!

PR #2977, opened 2021-05-24
[NETBEANS-5704] when generating --patch-module option, also use the generated source roots.

jlahoda:


java/maven/test/unit/src/org/netbeans/modules/maven/queries/UnitTestsCompilerOptionsQueryImplTest.java
```diff
@@ -59,13 +60,19 @@ public void testNoCompilerPluginSpecified() throws Exception {
...
+                                "package test;\n" +
+                                "public class Gen {}\n");
         FileObject testSource =
         TestFileUtils.writeFile(wd,
                                 "src/test/java/test/APITest.java",
                                 "package test;\n" +
                                 "public class APITest {}\n");
         assertEquals(Arrays.asList("--patch-module",
-                                   "test=" + FileUtil.toFile(wd.getFileObject("src/main/java")).getAbsolutePath()),
+                                   "test=" + FileUtil.toFile(wd.getFileObject("src/main/java")).toURI().getPath() +
```
matthiasblaesing, line 73: Are paths or URIs expected for patch-module? If paths are expected, this will break for example for spaces ("%20" vs " "). The documentation seems to be thin on this.

matthiasblaesing (commented):
I verified that the problem is reproducible with the sample provided in the referenced issue, verified, that the issue is fixed with this applied, but left a comment line regarding URI vs. file path.

java/maven/test/unit/src/org/netbeans/modules/maven/queries/UnitTestsCompilerOptionsQueryImplTest.java
jlahoda: Paths are expected. It is a good comment on escaping, but isn't URI.getPath() returning un-escaped paths? (I've tweaked the test to actually use a space with paths.)

matthiasblaesing: You are right, I had to reiterate the difference between java.net.URL and java.net.URI and I had to realize that an URI is used here.

matthiasblaesing (approved):

matthiasblaesing:
Lets get this in before it begins to bitrot.

PR #2979, opened 2021-05-28
[NETBEANS-5726] Fix JCheckboxMenuItem state for 'Show Editor Tooolbar'

eirikbakke:
Fix incorrect checkbox state shown by 'Show Editor Toolbar' and other boolean editor preference actions.

To reproduce the original bug:
1) Open a file for editing in the NetBeans editor.
2) Click View->"Show Only Editor"
3) Click View->"Show Editor Toolbar"
4) Click View->"Show Only Editor"
5) Click View->"Show Editor Toolbar"
6) There's now a checkmark next to View->"Show Editor Toolbar", even though there is no editor toolbar showing.

The problem was that the relevant actions were wrapped in WrapperAction, which cannot track changes in selection state. The state is initialized once and never subsequently updated. The fact that the checkbox state sometimes seems to track actual changes is merely an accident--it just toggles itself whenever the action is invoked from that particular JCheckboxMenuItem (as opposed to from e.g. a keyboard shortcut or from an external preference change).

### DIFF
--- a/ide/editor.actions/src/org/netbeans/modules/editor/actions/ToggleAction.java
+++ b/ide/editor.actions/src/org/netbeans/modules/editor/actions/ToggleAction.java
@@ -20,6 +20,7 @@
 package org.netbeans.modules.editor.actions;
 
 import java.awt.event.ActionEvent;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.swing.text.JTextComponent;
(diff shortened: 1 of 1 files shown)


ebarboni:
seems ok may you merge it @eirikbakke ?

eirikbakke:
Yes, the patch has worked without problem on my IDE for the last month. Making these actions non-lazy seems to be the only way to get them to behave correctly.

PR #2981, opened 2021-05-31
[NETBEANS-5736] Have clicking in "Projects" tree work along entire row

eirikbakke:
In the TreeView component, which is used for the "Projects" tab in the NetBeans IDE, the user should be able to click anywhere to the right of the node's icon to select said node. Drag-and-drop should work similarly.

This usability issue came up during user testing of a NetBeans Platform application. The user complained that files with a short filename were harder to click than files with long names. The user, a software engineer, pointed out that the equivalent tree component in IntelliJ allows the entire row to be clicked or dragged at to target a given node.

TreeView's behavior is derived from its superclass JTree. The OutlineView component does not have the same problem, as it derives from ETable and in turn JTable.

The fix required some gymnastics to trick JTree into believing the click was really within the bounds of the node's caption. The alternative (suggested by various StackOverflow posts) was to make the node actually wider, but this would be very hard to implement correctly.

Here's an illustration:
Clickfix

### DIFF
--- a/platform/openide.explorer/apichanges.xml
+++ b/platform/openide.explorer/apichanges.xml
@@ -26,6 +26,22 @@
 <apidef name="explorer">Explorer API</apidef>
 </apidefs>
 <changes>
+    <change id="setEntireRowClickable">
+        <api name="explorer"/>
+        <summary>Add the TreeView.setEntireRowClickable method.</summary>
+        <version major="6" minor="91"/>
+        <date day="13" month="3" year="2025"/>
+        <author login="ebakke"/>
+        <compatibility binary="compatible" source="compatible" deprecation="no" deletion="no" addition="yes"/>
+        <description>
+            The TreeView can now be optionally configured, via TreeView.setEntireRowClickable, so
+            that clicks in the empty space on the right-hand side of a node are interpreted as
+            clicks on said node.
+        </description>
+        <class package="org.openide.explorer.view" name="TreeView"/>
+        <issue number="NETBEANS-5736"/>
+    </change>
+
     <change id="ScrollOnExpand">
         <api name="explorer"/>
         <summary>Expose scrollOnExpand property on TreeView.</summary>
--- a/platform/openide.explorer/manifest.mf
+++ b/platform/openide.explorer/manifest.mf
@@ -2,5 +2,5 @@ Manifest-Version: 1.0
 OpenIDE-Module: org.openide.explorer
 OpenIDE-Module-Localizing-Bundle: org/openide/explorer/Bundle.properties
 AutoUpdate-Essential-Module: true
-OpenIDE-Module-Specification-Version: 6.90
+OpenIDE-Module-Specification-Version: 6.91
 
--- a/platform/openide.explorer/src/org/openide/explorer/view/TreeView.java
+++ b/platform/openide.explorer/src/org/openide/explorer/view/TreeView.java
@@ -155,6 +155,8 @@ public abstract class TreeView extends JScrollPane {
     /** Property indicating whether the default action is enabled. */
     transient boolean defaultActionEnabled;
 
+    transient boolean entireRowClickable;
+
     /** not null if popup menu enabled */
     transient PopupAdapter popupListener;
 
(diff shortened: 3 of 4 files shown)


neilcsmith-net:
Possibly -1 as proposed.  How do you provide access to the main Projects popup menu with this change?

eirikbakke:
That's a good question. You can still use the area to the left of nodes for this. Or the area under the last node--though if the content is taller than the scroll viewport, there won't be an area under the last area. So you'd need to collapse the nodes first to access this area.

Windows Explorer also has a special context menu which shows up if you right-click the empty area under the last node, but like here, it does become unavailable if the nodes fill the scroll viewport. Except there always seems to be a little bit of extra space added:

explorer

IntelliJ has the same problem--in their case, there is no way to bring up the node-neutral context menu if the scroll viewport is full.

Any suggestions?

neilcsmith-net:
> Any suggestions?

Right now, no.  Maybe the extra space.  But replacing one usability issue with another doesn't seem a good idea.  Maybe it's just my workflow, but I know I tend to use the node-neutral context menu a lot, and usually have enough projects open that the scroll viewport is full.  Also, which other views would be affected by this?

ebarboni:
Hi, maybe propose as an option. I perso prefers the current design.  Changing for one user seems a bit overkill.

eirikbakke:
Makes sense--it's kind of a breaking change, since previous uses of TreeView may assume that the root context menu is available more easily than it would be with the new behavior.

Is there interest in having this as an option for TreeView that can be enabled explicitly? I could update the patch to that effect.

sdedic:
FWIW, the Darcula L&F has the selection active on the entire line out of the box. Pls. test that this does not break with your changes. However the context right click only works on the node labels' area. 

I have also tried the Windows Explorer - they really have an usability issue: I managed to get just items from a single parent filling the entire tree viewport and there was NO space to click for the non-contex right-click menu.

Note that there was not an argument about the left-click (selection) behaviour, but just about the context menu: the usability issue to solve is that the area to right click may be **too narrow**. It's actually accessible always as the node icon (but not the expand handle!) works for right-click even for a "one space" name, so I wouldn't trade impossibility to right-click for the main menu for 'just' increased availability of the context menu.

What about a compromise ? A minimum width for the area active for a node right-click, in case the node's text itself is too narrow - expressed in 'eM's ?

eirikbakke:
Did you mean the external Darcula LAF plugin, or the new bundled FlatLAF Dark? I can test either. Now curious how they implemented full-row selection.

The thing with Windows Explorer and other apps that use the full-row behavior, is that the root-level context menu usually just provides shortcuts to actions that are available elsewhere (in the menu bar or in the context menu for individual nodes). So it's just an added convenience for cases where the user has nothing else to click, e.g. before creating a project. But in NetBeans, the "Projects" pane has a few actions which I'm not sure can be accessed in other ways, e.g. "Collapse All" or the "View Java Packages As" option. And if the root context menu has traditionally been easy to access, some people will get used to it over time.

> Note that there was not an argument about the left-click (selection) behaviour
OK, so perhaps this one should still select the whole row. There's also the behavior when dragging and dropping a node on another node. Here I think it's also good to have the whole row work--I don't think the root node is ever relevant here.

> A minimum width for the area active for a node right-click, in case the node's text itself is too narrow - expressed in 'eM's ?
I'll experiment with this, and see if there's an approach that feels natural.

sdedic:
> external Darcula LAF

Yes, the external Darcula (I have version 1.6 installed).

lkishalmi:
Well, I kind of like how this is working with the current implementation even with FlatLAF. FlatLAF paints the whole row as selected, but the row-popup context menu is only displayed if you right-click on the text. Otherwise the selection background clears and you get the project context popup.
It's not that consistent, maybe just the 10++ years wiring likes that in my brain.

eirikbakke:
I have revised this old PR to make it into an optional setting on TreeView instead, via the new API method TreeView.setEntireRowClickable, which is false by default.

The entireRowClickable behavior can be tested by setting `btv.setEntireRowClickable(true);` after `btv = new ProjectTreeView();` in ide/projectui/src/org/netbeans/modules/project/ui/ProjectTab.java.

neilcsmith-net:
I suggest opening a new PR for that, linking to the discussion here for context.  The title here, and most of the discussion, probably won't make sense linked from release notes if it's a new API rather than change to the project window.

I'm still inclined against using this in the IDE, and unsure about its benefit to the platform.  That all depends if the API you're adding ensures there's a way to access the context menu.

eirikbakke:
@neilcsmith-net OK, opened a new PR at https://github.com/apache/netbeans/pull/8329 where discussion can continue.

PR #2984, opened 2021-05-31
Completion items imported from Java delegate to ElementJavadoc to get the content.

sdedic:
Completion items that were taken from indexes or java classes had either no `ElementHandle` or some simple one, which did not allow to resolve them back. I added a Handle subclass that records java support's ElementHandle or at least FQNs more or less suficient to find an `ElementHandle` later (for creations not backed by Element from the java support).

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionHandler.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionHandler.java
@@ -22,12 +22,18 @@ import groovy.lang.MetaMethod;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.lang.model.element.Element;
 import javax.swing.text.Document;
 import javax.swing.text.JTextComponent;
 import org.codehaus.groovy.ast.ASTNode;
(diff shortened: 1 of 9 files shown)


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionHandler.java
```diff
@@ -53,12 +61,14 @@
...
 import org.netbeans.modules.groovy.editor.api.completion.util.CompletionContext;
+import org.netbeans.modules.groovy.editor.java.JavaElementHandle;
 import org.netbeans.modules.groovy.support.api.GroovySettings;
+import org.openide.util.Exceptions;
 import org.openide.util.NbBundle;
 import org.openide.util.Utilities;
 import org.openide.util.WeakListeners;
 
-public class CompletionHandler implements CodeCompletionHandler {
+public class CompletionHandler implements CodeCompletionHandler2 {
```
JaroslavTulach, line 71: `CodeCompletionHandler2` has been introduced in version of 2.43 of the csl.api. Please update dependency from 2.21 to 2.43.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionItem.java
```diff
@@ -910,30 +955,18 @@ public boolean equals(Object obj) {
         }
     }
 
-    public static class JavaFieldItem extends CompletionItem {
-
-        private final String className;
-        private final String name;
+    public static class JavaFieldItem extends JavaElementItem {
```
JaroslavTulach, line 958: API change.

JaroslavTulach (approved):
Update versioning. Please run `ant gen-sigtest` and check the changes are sane. It is unfortunate the implementation of some interface is itself in an API...

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/CompletionItem.java
sdedic: @JaroslavTulach  I'll better rework the change - it'd prefer not to introduce any API changes at all :-/ Those APIs are too open already. Sorry for the inconvenience - you'll be reviewing once more ...

sdedic:
Redone. Hopefully no API changes should be visible now.

JaroslavTulach (approved):

PR #2985, opened 2021-06-03
Avoid o(n^2) complexity when associating frames with threads in a heap dump.

JaroslavTulach:
Rather than always going through all the GC roots, let's build a mapping from Thread ID -> `ThreadObjectHprofGCRoot` during a single iteration. This may be a very useful fix when [project Loom](https://openjdk.java.net/projects/loom/) and its hundreds of threads become common. I've just tried to open a dump with 400000 threads and it didn't finish in five minutes! Then I gave up.

This PR is written by me and I have a right to integrate it to Apache NetBeans project. However, to keep difference between other `.hprof` reading libraries minimal, I want to provide a permission to use the patch with same license as NetBeans had when Oracle donated them to Apache.


sdedic:
Seems fine for me, but review from `lib.profiler` authors would be more meaningful :)

PR #2987, opened 2021-06-04
Debugger attach configurations ask for user input and configurations are split between Initial and Dynamic.

entlicher:
Currently, only the "Attach to Process" has a command associated, which lets the user to pick a process from a list.
Other attach configurations need to be edited with the desired values before they can be used.

To improve the usability, we create commands for individual attach attributes that ask the user for a value. This makes the attach configurations readily usable without any editing of the configurations themselves.

We also split the configurations between "Initial" and "Dynamic", as marking all of them as "Dynamic" caused unexpected behavior. Only the attach configurations are "Dynamic".

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/DebugConnector.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/DebugConnector.java
@@ -32,6 +32,12 @@ import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
 public final class DebugConnector {
 
     /**
+     * The identifier of the connector.
+     */
+    @NonNull
+    private String id;
+
+    /**
      * The display name identifier of the connector.
      */
     @NonNull
@@ -55,11 +61,28 @@ public final class DebugConnector {
     @NonNull
     private List<String> defaultValues;
 
-    public DebugConnector(String name, String type, List<String> arguments, List<String> defaultValues) {
+    /**
+     * Descriptions of debug connector arguments.
+     */
+    @NonNull
+    private List<String> descriptions;
+
+    public DebugConnector(String id, String name, String type, List<String> arguments, List<String> defaultValues, List<String> descriptions) {
+        this.id = id;
         this.name = name;
         this.type = type;
         this.arguments = arguments;
         this.defaultValues = defaultValues;
+        this.descriptions = descriptions;
+    }
+
+    @Pure
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
     }
 
     @Pure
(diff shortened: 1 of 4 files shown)


java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -705,13 +732,38 @@ class NetBeansConfigurationProvider implements vscode.DebugConfigurationProvider
...
+                                }).then((value) => {
+                                    if (value) {
+                                        commandValues.set(cmd, value);
+                                    }
+                                    return value;
+                                });
+                            }));
+                        }
+                    } else {
+                        debugConfig[ac.arguments[i]] = defaultValue;
```
sdedic, line 756: So if LSP server sends some dynamic default value, the client must implement prompting for that value (if desirable) ?

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -705,13 +732,38 @@ class NetBeansConfigurationProvider implements vscode.DebugConfigurationProvider
...
+                        if (!commandValues.has(cmd)) {
+                            commandValues.set(cmd, ac.defaultValues[i]);
+                            let description: string = ac.descriptions[i];
+                            context.subscriptions.push(commands.registerCommand(cmd, async (ctx) => {
+                                return vscode.window.showInputBox({
+                                    prompt: description,
+                                    value: commandValues.get(cmd),
+                                }).then((value) => {
+                                    if (value) {
+                                        commandValues.set(cmd, value);
```
sdedic, line 749: I can see that the `commandValues` storage receives user input. Is the InputBox displayed each time the `debugConfig` is launched (although pre-filled with the previous input) ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/DebugConnector.java
```diff
@@ -131,20 +166,19 @@ public boolean equals(Object obj) {
...
-            return false;
-        }
         return true;
     }
 
     @Pure
     @Override
     public String toString() {
         ToStringBuilder b = new ToStringBuilder(this);
+        b.add("name", id);
```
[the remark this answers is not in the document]
entlicher: A typo, thanks.

java/java.lsp.server/vscode/src/extension.ts
entlicher: Yes, it's up to the client to provide a way how user can customize the connector attributes. The server sends some defaults which it gets from JVM. The default for a hostName is the host name of the local machine, for instance. To connect somewhere else, you need to customize it on the client.
Instead of the current necessity to put the configuration into launch.json and edit it there, we provide commands that ask for the user input. The configurations usually can not work just with the defaults.

java/java.lsp.server/vscode/src/extension.ts
entlicher: Yes. If we would not ask again, the user would have no way to change the values.
If you attach often to some particular destination, you should put it into the launch.json, IMHO.
I tried to do the best with what we can do in VSCode, but suggestions are welcomed, of course.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/DebugConnector.java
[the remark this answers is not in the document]
entlicher: Corrected.

sdedic (approved):

PR #2988, opened 2021-06-05
[NETBEANS-4953] Fixing handling of WSL-based URIs.

jlahoda:
This is unfortunately a little bit involved. On Windows, one can have paths like `\\wsl$\Ubuntu-20.04\` (pointing to a Linux instance running in WSL). `java.nio.Path`s based on paths like this work normally, but NetBeans' FileObjects can't convert them to URIs. `java.nio.Path` will produce URIs like `file:////wsl$/Ubuntu-20.04/` for this path, and this patch fixes FileObjects to produce URIs like this.

The issue with these URIs is that `URI.normalize` will normalize them to `file:/wsl$/Ubuntu-20.04/` which no longer works. So this patch introduces `BaseUtilities.normalizeURI`, which should be used instead of `URI.normalize()`, and updates NetBeans to use the new method.


ebarboni:
works well

ebarboni:
I guess a a license file is missing for tools

ebarboni:
verify license is now ok.

ebarboni (approved):
locally tested works for me

PR #2989, opened 2021-06-05
nb-javac is not required any longer for compile on save

gitdode:
Since #2128, Compile on Save supports vanilla javac and does not require nb-javac any longer:

- removed the note "(requires nb-javac plugin)" and added "If nb-javac is not installed, vanilla javac is used." from/to the relevant occurrences
- removed the "Learn more about Compile on Save ..." link since the page it refers to does not exist any more (should create a new page and update the link instead?)


sdedic:
Yes, please - I think the limitations and workarounds described on the page are important, when the user faces problems. The archived page is for example http://web.archive.org/web/20210116190928/http://wiki.netbeans.org/FaqCompileOnSave
Should I create a page in Apache confluence with that content ?

gitdode:
That would be great, I also think the information on the page is important and helpful. If you could give me the link then I'll update and put back and test the "Learn more about Compile on Save ..." link.
Maybe we could update the page and mention vanilla javac as alternative to nb-javac and that it currently may have limitations like this one: [static initializer block not called from within NB](https://lists.apache.org/thread.html/r8deba8b7d20ed95eea6775048f962f9eddd240f8f6739c7ee2ac26fe%40%3Cusers.netbeans.apache.org%3E)

gitdode:
Sorry @sdedic , I obviously did something awfully wrong when trying to update the branch in my fork...
I'll be happy to make a new PR with only these changes:
- remove the note "(requires nb-javac plugin)" and add "If nb-javac is not installed, vanilla javac is used." from/to the relevant occurrences
- update the "Learn more about Compile on Save ..." link

PR #2990, opened 2021-06-08
Fixed NPEs in Truffle debugger when SourceSection is null.

entlicher:
Even though it's adviced that instrumentable Truffle Nodes provide `SourceSection`, it may happen that some Truffle language implementation does not provide it.

In the Truffle debugger implementation we need to count with `null` `SourceSection`.


JaroslavTulach (approved):

entlicher:
Removed an unnecessary parameters change of `org.netbeans.modules.debugger.jpda.truffle.source.SourcePosition` constructor, which changed friend API. No API changes are necessary.

PR #2992, opened 2021-06-12
Flatlaf 1.2 and public package

cahityilmaz:
Added FlatLaf packages to public package list. Is required if used SwingX addon in the NetBeans Platform Application.
Updated FlatLaf version to 1.2.


DevCharly:
Those packages should not be public.

This is no API and may change frequently,
which requires increasing module version frequently,
which causes problems for modules that depend on FlatLaf module
(e.g. upcoming JFormDesigner module).

Adding package `com.formdev.flatlaf.ui` to public package will probably break sigtest running with Java 8:
https://github.com/apache/netbeans/pull/1771#issuecomment-564956178

BTW FlatLaf JAR should be not checked into Git!

Isn't there a way to make packages accessible to specific other modules? (friends??)

matthiasblaesing:
> Isn't there a way to make packages accessible to specific other modules? (friends??)

Yes the friend mechanism might be the right approach, though it might become necessary to split the module as a module can either expose a public api or a friend api, but not both. The other option, if the two modules are hard dependencies for each other are implementation dependencies. An implementation dependency can be declared against every module - in that case the complete module can be accessed, but it will not be possible to combine the module with a different version.

The above should be taken with a grain of salt and validated :-)

cahityilmaz:
I need the com.formdev.flatlaf.ui package to public. Used the swingx components on the my project. Isn't there a way? I am currently using the nbm package that I produced.

PR #2993, opened 2021-06-13
[NETBEANS-5775] Removing dependency on a removed Log constructor.

jlahoda:


ebarboni:
looks good

arvindaprameya:
Approved, thank you for doing this Jan

PR #2994, opened 2021-06-13
NETBEANS-5775: NoSuchMethodError: 'void com.sun.tools.javac.util.Log.…

lhochet:
…<init>in .dump files when running NetBeans with JDK17+26

Since the code uses the same DEV_NULL for all the PrintWriters, replace the removed Log constructor taking 3 PrintWriters with the one that takes just one PrintWriter and use it for all outputs, that constructor exists from JDK 8 til JDK17+26/JDK18+1.


lhochet:
PR #2993 already fixes the issue

PR #2995, opened 2021-06-14
Add Run with configuration into code lenses.

entlicher:
Adding Run/Debug with project configurations for main methods into VSCode code lenses.
By default, only Run for Continuous Mode (in Gradle) and Micronaut's Dev mode (in Maven) is added.

The continuous mode/dev mode does not woth with the debugger, which is intentional. I'm introducing `ProjectConfigurationProvider.configurationAffectsAction()` to check whether an action is actually defined by a particular configuration.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
@@ -37,6 +37,7 @@ import java.net.URI;
 import java.net.URL;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
 import java.util.EnumSet;
@@ -51,6 +52,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.IntFunction;
 import java.util.logging.Level;
@@ -148,6 +150,7 @@ import org.netbeans.api.java.source.ui.ElementOpen;
 import org.netbeans.api.lexer.TokenSequence;
 import org.netbeans.api.lsp.Completion;
 import org.netbeans.api.lsp.HyperlinkLocation;
+import org.netbeans.api.project.FileOwnerQuery;
 import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectUtils;
 import org.netbeans.api.project.SourceGroup;
@@ -189,12 +192,17 @@ import org.netbeans.modules.refactoring.spi.Transaction;
 import org.netbeans.spi.editor.hints.ErrorDescription;
 import org.netbeans.spi.editor.hints.Fix;
 import org.netbeans.spi.lsp.ErrorProvider;
+import org.netbeans.spi.project.ActionProvider;
+import org.netbeans.spi.project.ProjectConfiguration;
+import org.netbeans.spi.project.ProjectConfigurationProvider;
 import org.openide.cookies.EditorCookie;
 import org.openide.filesystems.FileObject;
 import org.openide.text.NbDocument;
 import org.openide.text.PositionBounds;
 import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
+import org.openide.util.Pair;
 import org.openide.util.RequestProcessor;
 import org.openide.util.Union2;
 import org.openide.util.WeakSet;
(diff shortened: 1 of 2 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -1088,6 +1121,39 @@ public Void visitMethod(MethodTree tree, Void p) {
...
+            if (p != null) {
+                ProjectConfigurationProvider<ProjectConfiguration> provider = p.getLookup().lookup(ProjectConfigurationProvider.class);
+                List<Pair<String, String>> configDispNames = new ArrayList<>();
+                if (provider != null) {
+                    boolean skippedFirst = false;
+                    for (ProjectConfiguration c : provider.getConfigurations()) {
+                        if (skippedFirst) {
+                            String runConfig = null;
+                            String debugConfig = null;
+                            if (provider.configurationAffectsAction(c, ActionProvider.COMMAND_RUN_SINGLE)) {
```
sdedic, line 1136: I am not sure about this approach. If a Configuration does not define an action mapping, Maven and Gradle will fall back to the default (implicit) configuration's behaviour. This condition seems to disable the action if a configuration does not define a specialization for it. I think the semantic for LSP client diverges from the semantic seen by the IDE / headless LSP server.

sdedic: OK, one more thought. The code here seems to check if an action (run, debug) is **available** in a specific configuration. We have a similar issue with plain "Run" action in vscode (which means essentially "Debug" in vscode terminology). It was planned to display a warning message, something like "Debug is not supported in XXX config, running without debugger" - and for that I need a similar info.

What if, instead of `configurationAffectsAction` (or in addition to), `isActionEnabled(Configuration, String)` is introduced ? We can **enhance** Maven and Gradle XML format with `disable` flag for the actions (or autodetect when e.g. commandline or goal list is empty ?), and finally declare `debug` as disabled in the `mn:run` or `gradle --continuous` case. That would work for both code lens, the regular debug adapter, and would be consistent with the IDE behaviour.

@JaroslavTulach, @MartinBalin, @entlicher -- opinions ?

JaroslavTulach (commented):
I agree that the semantics of actions in configurations are getting more and more complicated.

Can't we stick to existing API? And just: if project's `ProjectConfigurationProvider != null` and `configurationsAffectAction(action)` and not `ActionProvider.isActionEnabled(action, singleton(configuration))` then don't show the action? 

Why? Because the action is _affected by configurations and currently disabled_ - e.g. not available in current configuration.

entlicher:
> I agree that the semantics of actions in configurations are getting more and more complicated.
> 
> Can't we stick to existing API? And just: if project's `ProjectConfigurationProvider != null` and `configurationsAffectAction(action)` and not `ActionProvider.isActionEnabled(action, singleton(configuration))` then don't show the action?
> 
> Why? Because the action is _affected by configurations and currently disabled_ - e.g. not available in current configuration.

This approach is not usable for me.
Because of this condition in gradle's `ConfigurableActionsProviderImpl.findActionProvider`:
```
if (m == null) {
    m = getActionData(GradleExecConfiguration.DEFAULT).getAction(action);
}
```
It defaults to the DEFAULT configuration. And this is exactly what I do not need. It makes sens to add only actions that differ in the configurations.
I must not include debug of continuous mode. And this would add it.
I'm not that experienced in projects, but I did not find an existing API that would suit the intended use-case.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
entlicher: There's no point in falling back. I intentionally do not want that.
The intention is to select unique Run/Debug actions from configurations. And this intention conveniently filters Debug in Continuous mode out. There's no point adding duplicit actions that would fall back somewhere into Code Lenses. Choosing simple Run/Debug is the fall-back already.

sdedic: My point was, that the project support(s) DO such fallback, and the API is defined so that configuration is just an override (the concept of disabling an action in a configuration is not present in the implementation, that's right). 

So the point is that this approach works now, in the current de-facto state, for these specific actions. But can break in the future - consider (some) Configuration that focuses on other technology aspect than `--continuous`, i.e. on build or package; if it won't defne `run.single`, Run would be disabled.

IMHO it would be better to somehow implement @JaroslavTulach's suggestion to allow plugins (maven, gradle) to indicate the action is to be disabled in a configuration - and propagate that from API through `isActionEnabled` query.

entlicher: O.K. it makes sense. It's better not to have the fallback to the default configuration in the action impl, but just in the projects infrastructure, I agree.

sdedic: OK, if you agree, I'll prepare a PR that allows to disable an action that for maven + gradle. 

Do you think we should disable `debug` for the 'continuous' mode in the IDE as well ? Right now running Debug in dev mode will just run the debugger, without 'continuous' parameter.

entlicher: Yes, thanks, I agree.
In the IDE I'm not sure. As the user does Debug explicitly, it's probably better to start debugger, but it'd be good to show a warning that we do not use the Continuous Mode configuration, I think.
In VSCode I didn't find a way how to switch to plain run when debug is started. We probably need to cancel debugger and auto-start Run.

entlicher:
I've removed projects changes and the API change. As soon as actions do not fall-back to the default configuration, this should work as intended.

JaroslavTulach (approved):

sdedic:
I believe this an be merged in parallel / after #3004.

sdedic (approved):

sdedic:
@entlicher  can this one be merged ?

MartinBalin:
Yes! Please...

entlicher:
Rebased and resolved conflicts.

PR #2996, opened 2021-06-14
Propagate createFromTemplate parameters to folder subobjects

JaroslavTulach:
To use `DataFolder.createFromTemplate` on project templates, like [the one in gradle](https://github.com/apache/netbeans/blob/master/java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/resources/layer.xml#L30), we need to propagate `packageBase`, `groupId`, etc. properties.

### DIFF
--- a/platform/openide.loaders/src/org/openide/loaders/DataFolder.java
+++ b/platform/openide.loaders/src/org/openide/loaders/DataFolder.java
@@ -835,10 +835,11 @@ public class DataFolder extends MultiDataObject implements DataObject.Container 
         DataFolder newFolder = (DataFolder)super.handleCreateFromTemplate (f, name);
         Enumeration<DataObject> en = children ();
 
+        Map<String, Object> params = CreateAction.getCallParameters(null);
         while (en.hasMoreElements ()) {
             try {
                 DataObject obj = en.nextElement ();
-                obj.createFromTemplate (newFolder);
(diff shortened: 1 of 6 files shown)


platform/openide.loaders/src/org/openide/loaders/DataFolder.java
```diff
@@ -835,10 +835,11 @@ protected DataObject handleCreateFromTemplate (
         DataFolder newFolder = (DataFolder)super.handleCreateFromTemplate (f, name);
         Enumeration<DataObject> en = children ();
 
+        Map<String, Object> params = CreateAction.getCallParameters(null);
```
sdedic, line 838: Good, this is consistent with `MultiDataObject.create` ... would it be appropriate to also call `FileBuilder.createFromTemplate` on the folder's FileObject ifself ? If no handler is interested (`CreateFromTemplateHandler.accept`) then creation can proceed recursively; otherwise, handlers can be defined for custom folder processing.

sdedic (approved):
Should work consistently with `MultiDataObject'.

platform/openide.loaders/src/org/openide/loaders/DataFolder.java
JaroslavTulach: As you wish: 76105b3

PR #2998, opened 2021-06-15
[NETBEANS-5784] Fixed defect: removal of deprecated constructor(Log.java) in JDK-17

singh-akhilesh:
https://issues.apache.org/jira/browse/NETBEANS-5784


ebarboni:
looks like #2993 but seems you edit test too :D

singh-akhilesh:
> looks like #2993 but seems you edit test too :D

Sorry, Not seen the PR is already present. It would be good if #2993 modify test case too.

ebarboni:
can we close this and focus on #2993 ?

singh-akhilesh:
> can we close this and focus on #2993 ?

Yes, Please do close. TIA

PR #2999, opened 2021-06-16
Instantiate Maven & Gradle project without Swing UI

JaroslavTulach:
I am supposed to implement an action to create a project in VSNetBeans with as minimal UI as possible. As such I am turning the project templates into `CreateTemplateHandler` that work with minimal configuration. I managed to do it for Gradle without any API changes - I just had to make the templates more robust.

TBD: Possibly apply similar changes to Maven. So far there is no `CreateTemplateHandler` yet.

There is a `VerifyProjectTemplatesTest` to instantiate selected project templates in the mode and verify whether they are buildable.


lkishalmi (approved):
Looks good to me.
Thanks form making the templates more robust as well!

extide/gradle/src/org/netbeans/modules/gradle/newproject/GradleProjectFromTemplateHandler.java
```diff
@@ -0,0 +1,93 @@
...
+    private static List<FileObject> sortByParentship(Collection<FileObject> fo) {
+        ArrayList<FileObject> files = new ArrayList<>(fo);
+        files.sort((a, b) -> {
+            if (a.equals(b)) {
+                return 0;
+            }
+            if (FileUtil.isParentOf(a, b)) {
+                return -1;
+            }
+            return 1;
```
sdedic, line 75: Are all the input files in strictly parent-child relationship ? If a, b are siblings, then compare(a,b) = compare(b, a) = 1 -> won't define a total ordering.

sdedic (approved):

extide/gradle/src/org/netbeans/modules/gradle/newproject/GradleProjectFromTemplateHandler.java
JaroslavTulach: Great catch: 1ffe9701c92e

extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
```diff
@@ -149,7 +149,7 @@ public void invokeAction(String command, Lookup context) throws IllegalArgumentE
                 prjImpl.primeProject().
                         thenAccept(gp -> {
                             LOG.log(Level.FINER, "Priming build of {0} finished with status {1}, ", new Object[] { project, prjImpl.isProjectPrimingRequired() });
-                            prg.finished(prjImpl.isProjectPrimingRequired());
+                            prg.finished(!prjImpl.isProjectPrimingRequired());
```
JaroslavTulach, line 152: @sdedic, would you agree this condition should be reverted?

sdedic: oh, absolutely; thanks.

JaroslavTulach:
The functionality seems to run fine locally. Once the checks are OK, I'll integrate.

sdedic:
Filed [NETBEANS-5847](https://issues.apache.org/jira/browse/NETBEANS-5847) for Maven projects.

PR #3001, opened 2021-06-16
Groovy Code Completion support available through LSP protocol.

sdedic:
Groovy support has very open APIs, so virtually anyone can derive from Groovy classes; that made a factory approach used in Java unsuitable - need to support potential alien implementations as well. At the same time I didn't want to expose any further APIs, although it was beneficial to add fields / getters to individual `CompletionItem` subclasses - so I have use package Accessor pattern to create/query the completion items. The accessor interface is hidden in a nonpublic package.

I moved most of the `CompletionHandler` implementation to a nonpublic package, `CompletionHandler` just delegates to it; gives a freedom to add methods for the module's internal consumption / testing.

The `CompletionCollector` implementation just translates original Groovy completion suggestions to the LSP protocol, using some extra information that items produced by `groovy.editor` module have, but alien items may have missed; such items could be eventually ignored on LSP.

The last bunch of changes is a compartor parallel to the one in CSL, which translates `Completion` instances back to textual form. As `Completion` carries less info than `CompletionProposal`, the missing parts are extracted from CompletionProposals - this way the original golden files can be used to verify the contents of LSP completion is the same as contents of editor native completion.

Update: some bugs were filed during impl. Should be fixed, but would typically change the CC for the IDE as well (= test data changes), so I wanted to separate them to separate PRs:
- [NETBEANS-5788](https://issues.apache.org/jira/browse/NETBEANS-5788)
- [NETBEANS-5787](https://issues.apache.org/jira/browse/NETBEANS-5787)

### DIFF
--- a/groovy/groovy.editor/manifest.mf
+++ b/groovy/groovy.editor/manifest.mf
@@ -3,4 +3,4 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.groovy.editor/3
 OpenIDE-Module-Layer: org/netbeans/modules/groovy/editor/resources/layer.xml
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/groovy/editor/Bundle.properties
-OpenIDE-Module-Specification-Version: 1.76
+OpenIDE-Module-Specification-Version: 1.77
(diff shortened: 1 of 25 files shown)


graemerocher:
Nice. Is there a build I can test this out somewhere?

JaroslavTulach (approved):

dbalek (approved):
Looks fine. Thanks.

PR #3002, opened 2021-06-16
make JVM settings in netbeans.conf easier to read and edit.

mbien:
small change which attempts to make netbeans.conf a bit easier to read and edit.

motivation:
the long list of module options is fairly uninteresting from the users point of view. The most likely usecase for editing netbeans.conf is to change the path to the JDK or to edit some JVM settings. This change extracts the module options and formats the remaining JVM flags so that they are easier to read and edit.


neilcsmith-net:
Does this work with the Windows launcher?!

mbien:
> 
> 
> Does this work with the Windows launcher?!

I just tested it and it doesn't seem to like the variable substitution. What is used to parse the conf file on windows?

neilcsmith-net:
Yes, I thought it might be problematic.  It's here I think - https://github.com/apache/netbeans/blob/master/nb/ide.launcher/windows/nblauncher.cpp#L461 and possibly replicated in the platform launcher.

mbien:
> 
> 
> Yes, I thought it might be problematic. It's here I think - https://github.com/apache/netbeans/blob/master/nb/ide.launcher/windows/nblauncher.cpp#L461 and possibly replicated in the platform launcher.

thanks for the pointers! i am going to take a look if there is a way to make all launchers happy. I am going to try to avoid having to build netbeans on windows since I have nothing set up there.

mbien:
the win launcher does not support any kind of variables, its only looking for hardcoded property names. I am going to close this.

PR #3003, opened 2021-06-17
Temporarily allow setting of SecurityManager.

jlahoda:
Future JDK versions may require setting system property `java.security.manager` to allow in order for `System.setSecurityManager` to be allowed. This patch sets that property for the unix launchers, incorporates Windows launchers that set the property, fixes the fall out (Equinox does not know how to handle `allow`, and neither do JDK older than JDK 12), and removes the sources for the Windows launchers that have been moved into a separate repository.


ebarboni:
Hi @jlahoda would it be possible for you to "move" the windows part of your PR on the 
https://github.com/apache/netbeans-native-launchers repo 
I take  your sentences for the readme.

jlahoda:
Please see for:
https://github.com/apache/netbeans-native-launchers/pull/2

for Windows launcher changes.

ebarboni:
maybe removing windows native code may also be good to avoid duplicate (and readme update somewhere) can do that in other PR maybe depends.

ebarboni:
looks ok for me. Why do we still need 8.2 launchers?

jlahoda:
> looks ok for me. Why do we still need 8.2 launchers?

These are still potentially used for the old pre-NB 7 platform applications.

ebarboni (approved):
looks good to me. Thanks for cleanup too.

ebarboni:
@jlahoda may I merge ? or should I do squash and merge

ebarboni:
had to insist :D, It would be great to have it for 12.5.

mbien:
@jlahoda @lahodaj do you remember why `-Djava.security.manager=allow` was set in the launchers instead of the .conf file?

lahodaj:
I think my goal was to reduce the risk of running without the option (as the `.conf` file can be more easily modified by the user).

PR #3004, opened 2021-06-18
Gradle Action Providers can disable an action.

sdedic:
With the introduction of `--continuous` mode in Gradle, it turned out that **Debug** style actions are confusing: I have defined them so they actually run in normal mode: consider a debugger, which would break out and restart the application on every Save to any project's file - that seems ... very very annoying (at least for me).

There's no "good" solution to that:
- we educate the user, that Debug is actually working differently (and users would complain that Debug does wrong thing in Continuous mode before they learn about the details). 
- we can disable Debug action for the configuration - which this PR does. The user can restore the action using NB GUI, if he/she chooses to live with the restrictions.


sdedic:
Good: I've checked this together with [PR-2995](https://github.com/apache/netbeans/pull/2995) and it seems to play well together. Code lenses only report 'Run' for Continuous mode to the LSP client.

sdedic:
I added a possibiltiy to disable an action (button **Disable**) and a way how the user can **Restore** the action:
- the Disable customized action will be removed, returning to the provided configuration's defaults (if that default is not 'disable' instruction), or
- the action def for the `default` configuration, if it exists.
This way the user can re-enable `Debug` actions and define whatever behaviour she/he likes in the continuous mode.

java/gradle.java/log
```diff
@@ -0,0 +1,297 @@
+Buildfile: /space/src/vscode/netbeans/java/gradle.java/build.xml
```
JaroslavTulach, line 1: Unnecessary.

extide/gradle/src/org/netbeans/modules/gradle/spi/actions/GradleActionsProvider.java
```diff
@@ -41,6 +41,11 @@
  * that must be present in the project in order for the {@link GradleActionsProvider} to activate. In that case, use 
  * {@link ProjectServiceProvider#projectType()} = {@code "org-netbeans-modules-gradle/Plugins/<plugin-id>"}.
  * <p>
+ * <b>From version 2.14</b> the {@code InputStream} returned by {@link #defaultActionMapConfig()} can request <b>disabling</b> an action.
```
JaroslavTulach, line 44: "can <b>disable</b> an action"

JaroslavTulach (approved):

java/gradle.java/log
sdedic: Removed / recommitted in f9a8ede so it does not appear in the history.

extide/gradle/src/org/netbeans/modules/gradle/spi/actions/GradleActionsProvider.java
sdedic: Corrected in f9a8ede

entlicher:
Works fine with https://github.com/apache/netbeans/pull/2995
Now we need something similar for Maven.

entlicher (approved):

extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
```diff
@@ -539,6 +541,39 @@ public static ReplaceTokenProvider simpleReplaceTokenProvider(final String token
...
+
+    /**
+     * Checks if the action is disabled. Use in preference to plain <code>actionMapping == null</code>
+     * as it also handles a mapping that forcefully deconfigures an action.
+     * 
+     * @param am action mapping to check
+     * @return true, if the action is <b>disabled</b>
+     * @since 2.14
+     */
+    public static boolean isActionDisabled(ActionMapping am) {
```
lkishalmi, line 570: Would be better as a static method in ActionMapping.isDisabled(ActionMapping)

extide/gradle/src/org/netbeans/modules/gradle/actions/DefaultActionMapping.java
```diff
@@ -145,5 +145,10 @@ public boolean equals(Object obj) {
         }
         return Objects.equals(this.withPlugins, other.withPlugins);
     }
+
+    public static DefaultActionMapping DISABLED = new DefaultActionMapping();
```
lkishalmi, line 149: Wouldn't it be better to have a static final member in ActionMapping interface as a specific instance?

lkishalmi (changes requested):

extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
sdedic: OK :) to be more fun, I had it that way (both isDisabled and DEFAULT) originally, then I've decided not to stretch the new Java features that much and refactored. No real preference from my side.

extide/gradle/src/org/netbeans/modules/gradle/actions/DefaultActionMapping.java
sdedic: Hm, originally I had it there - but then it's not that easy, since if the `ActionMapping` is to be returned from some Provider, it should probably hold that ActionMapping.getName().equals(the_id_passed_to_the_provider). A generic DISABLED with no `name` property at all should remain just an internal placeholder.
If something should be public, then a factory method `disabledAction(String name)` to have a correct instance floating around.

lkishalmi: Ok, reasoning approved, let it be. Do not make it public.

extide/gradle/src/org/netbeans/modules/gradle/api/execute/RunUtils.java
sdedic: Done in 688e72917b

lkishalmi (approved):
Looks good to me. Thank you!

sdedic:
OK, review done, will squash the review changes to the impl commits before merging.

neilcsmith-net:
@lkishalmi @sdedic could you check the ActionMapping.isDisabled(..) call with interaction with the navigator?  Debugging through from https://issues.apache.org/jira/projects/NETBEANS/issues/NETBEANS-5906 seems to come back with actions always disabled?

PR #3006, opened 2021-06-18
LSP server informs client about unsuccessful launch.

sdedic:
The launch request handler checks if the selected action is actually enabled in the project; if not, it will attempt to guess an appropriate alternative configuration and suggests it to the LSP client using a notification.


JaroslavTulach (approved):

dbalek (approved):
Nice.

entlicher:
:-D Github has blocked my comments till I pressed "Submit review" button _above_ the comments.
Thus here there are several days old comments.
The unused method is going to be removed in https://github.com/apache/netbeans/pull/3015

PR #3007, opened 2021-06-19
[NETBEANS-5799] : Add support for Pattern Matching for switch (Preview)

singh-akhilesh:
https://issues.apache.org/jira/browse/NETBEANS-5799

### DIFF
--- a/java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask117FeaturesTest.java
+++ b/java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask117FeaturesTest.java
@@ -1,0 +1,94 @@
+package org.netbeans.modules.java.completion;
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.lang.model.SourceVersion;
+import javax.swing.event.ChangeListener;
+import org.netbeans.junit.NbTestSuite;
+import org.netbeans.modules.java.source.parsing.JavacParser;
+import org.netbeans.spi.java.queries.CompilerOptionsQueryImplementation;
+import org.openide.filesystems.FileObject;
+import org.openide.util.lookup.ServiceProvider;
+/**
+ *
+ * @author arusinha
+ */
+public class JavaCompletionTask117FeaturesTest extends CompletionTestBase {
+
+    private static String SOURCE_LEVEL = "17"; //NOI18N
+
+    public JavaCompletionTask117FeaturesTest(String testName) {
+        super(testName);
+    }
+
+    public static NbTestSuite suite() {
+        NbTestSuite suite = new NbTestSuite();
+        try {
+            SourceVersion.valueOf("RELEASE_17"); //NOI18N
+            suite.addTestSuite(JavaCompletionTask117FeaturesTest.class);
+        } catch (IllegalArgumentException ex) {
+            //OK, no RELEASE_13, skip tests
+            suite.addTest(new JavaCompletionTask117FeaturesTest("noop")); //NOI18N
+        }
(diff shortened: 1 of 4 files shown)


java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask117FeaturesTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    static {
+        JavacParser.DISABLE_SOURCE_LEVEL_DOWNGRADE = true;
+    }
+    
+    private String getLatestSource(){
+        return SourceVersion.latest().name().substring(SourceVersion.latest().name().indexOf("_")+1);
+    }
+    private static final List<String> EXTRA_OPTIONS = new ArrayList<>();
```
jlahoda, line 92: Please move the EXTRA_OPTIONS into the TestCompilerOptionsQueryImplementation, as was done in: https://github.com/apache/netbeans/commit/8ec77c94d70e94af015ecccc2327626954c82237

This is so that other tests are not so much effected by the completion tests.

java/java.source.base/src/org/netbeans/api/java/source/TreeMaker.java
```diff
@@ -267,6 +267,18 @@ public CaseTree Case(List<? extends ExpressionTree> patterns, Tree body) {
...
     
+    /**
+     * Creates a new CaseTree for a rule case (case &lt;constants&gt; -> &lt;body&gt;).
+     *
+     * @param patterns the labels for this case statement.
+     * @param body the case's body
+     * @see com.sun.source.tree.CaseTree
+     * @since 2.39
+     */
+    public CaseTree CaseMultipleSwitchPatterns(List<? extends Tree> patterns, Tree body) {
```
jlahoda, line 278: Maybe just `CaseMultiplePatterns`? Just a suggestion.

Also - there should probably be the similar method taking statements instead of body.

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -2809,6 +2815,41 @@ private Boolean scanBindingPattern(Tree node, Void p) {
             return true;
         }
 
+        private Boolean scanGuardedPattern(Tree node, Void p) {
+            try {
+                Tree gpt = TreeShims.getGuardedPattern(node);
+                if (gpt != null) {
+                    Name name = TreeShims.getBinding(gpt);
```
jlahoda, line 2822: I am not sure if this works - the nested pattern in the guarded pattern can be parenthesized pattern. Probably should just call `scan` on `gpt`? (Which can't be null?)

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -2809,6 +2815,41 @@ private Boolean scanBindingPattern(Tree node, Void p) {
...
+                    Tree type = TreeShims.getBindingPatternType(gpt);
+                    scan(type, p);
+                    if (name != null) {
+                        removeWhiteSpace(IDENTIFIER);
```
jlahoda, line 2826: It is not quite clear what `remoteWhitespace` does - would `space();` work?

jlahoda (commented):
Overall, not bad. I've left some comments. A few more general comments:
-there should be more tests - for reformatting, for the hint, for `VeryPretty`. E.g. the reformatting tests will help avoid regressions on future versions of JDK.
-I would suggest to (now or later) look at `CasualDiff` for cases with patterns/guarded patterns/parenthesized patterns, that is likely to show up somewhere.

java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask117FeaturesTest.java
singh-akhilesh: Addressed review comment, moved EXTRA_OPTIONS to class TestCompilerOptionsQueryImplementation.

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
singh-akhilesh: space(); also worked here, but adding some extra black spaces after case labels. In some case wrapping extra blank line also doesn't worked. If you want, I'm fine to replace this with space();

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
singh-akhilesh: Addressed code review comment. Added code to handle  ---- case (String s) && s.<any> -> stmt;

lahodaj (approved):
Overall, looks good. One further comment: possibly instead of add `visitCase` to each relevant TreeScanner subclass, add it to `ErrorAwareTreeScanner`/`ErrorAwareTreePathScanner`.

singh-akhilesh:
> Overall, looks good. One further comment: possibly instead of add `visitCase` to each relevant TreeScanner subclass, add it to `ErrorAwareTreeScanner`/`ErrorAwareTreePathScanner`.

Addressed review comment- Removed visitCase method from individual classes and added to ErrorAwareTreeScanner/ErrorAwareTreePathScanner

PR #3010, opened 2021-06-22
Improving usability of New from Template in VSNetBeans.

JaroslavTulach:
One complain  was that it is not possible to cancel the creation of a template. Addressed by 270ec4e. 

Another complain suggested to avoid display _"Unrecognized file"_ obtained from a node short description  - addressed by ignoring it in 125b650.

Then there were complains about some templates producing empty files  - as such I wrote a test in c7de9d7 to go through all available templates and verify they lead to non-empty result. Few didn't and those are branded out in 3b3aec0. Moreover some Dušan's templates couldn't be instantiated without providing some variables - fixed in 6bd46cd.

### DIFF
--- a/java/java.lsp.server/nbcode/integration/test/unit/src/org/netbeans/modules/nbcode/integration/VerifySimpleTemplatesTest.java
+++ b/java/java.lsp.server/nbcode/integration/test/unit/src/org/netbeans/modules/nbcode/integration/VerifySimpleTemplatesTest.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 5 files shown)


java/java.lsp.server/nbcode/integration/test/unit/src/org/netbeans/modules/nbcode/integration/VerifySimpleTemplatesTest.java
```diff
@@ -0,0 +1,128 @@
...
+                final DataObject generated = t.createFromTemplate(scratch, "Test" + ++cnt);
+                final FileObject pf = generated.getPrimaryFile();
+                final Editable edit = generated.getLookup().lookup(Editable.class);
+                if (edit != null) {
+                    LOG.log(Level.WARNING, "Editing {0}", pf.getPath());
+                    edit.edit();
+                } else {
+                    Openable open = generated.getLookup().lookup(Openable.class);
+                    if (open != null) {
+                        LOG.log(Level.WARNING, "Opening {0}", pf.getPath());
```
sdedic, line 90: Is this `WARNING` level message necessary in this 'successful' path - or is it meant to trace the progress in the test job logs ?

sdedic (approved):

java/java.lsp.server/nbcode/integration/test/unit/src/org/netbeans/modules/nbcode/integration/VerifySimpleTemplatesTest.java
JaroslavTulach: This kind of integration IDE tests is better to be verbose, I think.

PR #3011, opened 2021-06-22
Small LSP fixes

sdedic:
Two fixes.
1. If the project is not known, thanks to @JaroslavTulach  null check the call proceeds through a series of async calls to finally return an empty array; this change does the same in a short path. I think that with no project, the original code does not wait on any Future from the backlog of projects being opened, so the change should have no impact on the sequence of operations.

2. The commit a2446ef introduced `HintsInvoker` invocation during execution of `codeAction` LSP request; although the request can return an unfinished `CompletableFuture`, which allows the client to issue additional requests through LSP protocol, the impl fully computed the code actions, then returned a `completedFuture`. The change only forks the heavy computation to background, and completes the future at the end of the task.


dbalek (approved):

sdedic:
Detected conflict with `master`; rebased & adjusted `TextDocumentServiceImpl`.

sdedic:
No further response from reviewer; merging.

PR #3012, opened 2021-06-22
[NETBEANS-5803] Handle Maven projects with https name space.

entlicher:
The Maven project support handles only `http` namespace. It does not count with the secure protocol.
As a result, Micronaut projects, for instance, can not be customized in NetBeans.

This PR improves the error dialog with a message describing the reason why POM wasn't parsed
and adds `https` name space.

### DIFF
--- a/apisupport/maven.apisupport/nbproject/project.xml
+++ b/apisupport/maven.apisupport/nbproject/project.xml
@@ -180,7 +180,7 @@
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>1</release-version>
-                        <specification-version>1.34</specification-version>
+                        <specification-version>1.57</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenPackageModifierImpl.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenPackageModifierImpl.java
@@ -175,7 +175,7 @@ public class MavenPackageModifierImpl implements PackageModifierImplementation {
                         }
                         if (export) {
                             for (String exp : packagesToExport) {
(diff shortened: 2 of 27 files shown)


neilcsmith-net:
Isn't that because it's an invalid namespace?

sdedic:
Strictly speaking the `http://maven.apache.org/POM/4.0.0` is an identifier although in a form of URI. The referenced `https://maven.apache.org/xsd/maven-4.0.0.xsd` really declares **http://** URI as its `targetNamespace`:
```
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" xmlns="http://maven.apache.org/POM/4.0.0" targetNamespace="http://maven.apache.org/POM/4.0.0">
```
so I'd say it's a lucky coincidence that a POM that uses `https://` namespace URI actually validates.

neilcsmith-net:
Also found this, so we're not alone - https://youtrack.jetbrains.com/issue/IDEA-270998  This should be punted back to Micronaut?

matthiasblaesing:
Yes - a namespace does not need to relate to a concrete document. Its the job of the resolver to fetch the right document. NetBeans carries multiple catalogs, that resolve namespaces to actual documents and the built-in resolvers resolve to local documents bundled with netbeans (have a look at the `enterprise/j2ee.dd` and `enterprise/j2ee.ddloaders` modules, which carry the XSDs for the JavaEE deployment descriptors).

A good part of the XSDs has a targetNamespace of:

http://xmlns.jcp.org/xml/ns/javaee

It is pure luck, that at that location a list of descriptors is placed, that can be used by humans to download the files.

If the micronaut applications really target the `https...` namespace, they target a different namespace, then maven does itself. I would consider it a bug in micronaut.

entlicher:
Thanks for your comments and pardon my ignorance.
I've created https://github.com/micronaut-projects/micronaut-core/issues/5642

entlicher:
I'll create a PR with the first commit, at least. Otherwise this looks like a weird NetBeans bug.

PR #3013, opened 2021-06-22
Provide an error message when POM status is invalid.

entlicher:
If there is a Maven project with wrong pom.xml, just a generic error dialog is displayed: https://issues.apache.org/jira/browse/NETBEANS-5803

This PR adds the specific error message to the dialog.

### DIFF
--- a/ide/xml.xam/src/org/netbeans/modules/xml/xam/AbstractModel.java
+++ b/ide/xml.xam/src/org/netbeans/modules/xml/xam/AbstractModel.java
@@ -60,6 +60,7 @@ public abstract class AbstractModel<T extends Component<T>>
     private PropertyChangeSupport pcs;
     protected ModelUndoableEditSupport ues;
     private State status;
+    private String statusMessage;
     private boolean inSync;
     private boolean inUndoRedo;
     private EventListenerList componentListeners;
@@ -179,12 +180,22 @@ public abstract class AbstractModel<T extends Component<T>>
         return status;
     }
     
+    @Override
+    public String getStatusMessage() {
+        return statusMessage;
+    }
+    
     protected void setState(State s) {
+        setState(s, null);
+    }
+
+    protected void setState(State s, String message) {
         if (s == status) {
             return;
         }
         State old = status;
         status = s;
+        statusMessage = message;
         PropertyChangeEvent event =
                 new PropertyChangeEvent(this, STATE_PROPERTY, old, status);
         if (isIntransaction()) {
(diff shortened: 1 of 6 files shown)


ide/xml.xam/src/org/netbeans/modules/xml/xam/AbstractModel.java
```diff
@@ -179,12 +180,22 @@ public State getState() {
...
+    }
+
+    protected void setState(State s, String message) {
         if (s == status) {
             return;
         }
         State old = status;
         status = s;
+        statusMessage = message;
         PropertyChangeEvent event =
```
matthiasblaesing, line 199: `statusMessage`is a different property, so I would expect both `PropertyChangeEvents` to be fired.

ide/xml.xam/src/org/netbeans/modules/xml/xam/AbstractModel.java
```diff
@@ -179,12 +180,22 @@ public State getState() {
...
+    @Override
+    public String getStatusMessage() {
+        return statusMessage;
+    }
+    
     protected void setState(State s) {
+        setState(s, null);
+    }
+
+    protected void setState(State s, String message) {
```
matthiasblaesing, line 192: The `if` construct in line 193 (new) looks wrong - the state could stay the same, but the message could change.

java/maven/nbproject/project.xml
```diff
@@ -369,7 +369,7 @@
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>1</release-version>
-                        <specification-version>1.4</specification-version>
+                        <specification-version>1.45</specification-version>
```
matthiasblaesing, line 372: Looks like a large jump, should this be `1.5`?

matthiasblaesing (commented):
Thank you - I left a few minor inline comments.

java/maven/nbproject/project.xml
sdedic: No - see [change in XAM in this PR](https://github.com/apache/netbeans/pull/3013/files#diff-d4cbe9616caee5c9e8a0b7766f54bf932349c61e0bcc223082bbebfa3c117ae2R24)

ide/xml.xam/src/org/netbeans/modules/xml/xam/AbstractModel.java
```diff
@@ -179,12 +180,22 @@ public State getState() {
...
         if (s == status) {
```
sdedic, line 193: `statusMessage` does not behave as a property; can be only updated during a state change ... which can be reasonable. Pls. document the behaviour, `status` name looks like property updatable even during a single state.

java/maven/nbproject/project.xml
matthiasblaesing: Thank you - makes sense now.

ebarboni:
pending conversation moving to 12.6

neilcsmith-net:
What's the current status of this PR?

neilcsmith-net:
@entlicher has conflicts and unresolved conversations, so removing from all milestones and marking as stale.  Please update or close as appropriate.

PR #3015, opened 2021-06-23
Removing unused findEnabledTarget() method.

entlicher:
Github decided to ignore my comment to https://github.com/apache/netbeans/pull/3006

This is how I see that PR:
Screenshot from 2021-06-23 16-33-49

Apparently, Github ignores my comments and does not show it to others, maybe just because I did not start a "Review". Apparently, comments have no value without a review and are blocked from others.


sdedic (approved):
@entlicher I've added the missing `return` you wrote about in the original PR. Please review as well :)

PR #3016, opened 2021-06-24
Support for disabling actions in Maven (used by Micronaut)

sdedic:
A follow-up to 3004, similar support to **disable project action** in a configuration is now implemented for Maven. The action customizer allows a simple way to disable an action, or to revert the 'disabled' instruction to a default configuration's action mapping.

Micronaut impl was updated to disable Debugs in 'dev mode' (`mn:run`).

### DIFF
--- a/enterprise/micronaut/nbproject/project.xml
+++ b/enterprise/micronaut/nbproject/project.xml
@@ -56,7 +56,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.17</specification-version>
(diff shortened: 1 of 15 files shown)


sdedic:
(sorry for all the mis-clicks)

JaroslavTulach (approved):

dbalek (approved):

neilcsmith-net:
@sdedic could you take a look at whether this PR is the cause for the disabled global actions mentioned in this thread https://lists.apache.org/thread/l5yrztopsb7z8odt8jkdyb6omn3tb0w7  ?? Thanks!

PR #3021, opened 2021-06-28
Make sure Java: New Project... is always active

JaroslavTulach:
Declare `oncommand:java.workspace.newproject` as an activation point for the extension.


dbalek (approved):
Fine.

PR #3022, opened 2021-06-28
LSP: Hover, go to declaration, and code lens implementation for Groovy

dbalek:
* LSP: Groovy Spock test methods finder implemented
* LSP: Hover and go to declaration implementated for Groovy

### DIFF
--- a/groovy/libs.groovy/nbproject/project.xml
+++ b/groovy/libs.groovy/nbproject/project.xml
@@ -44,7 +44,6 @@
                 <package>org.codehaus.groovy.control</package>
                 <package>org.codehaus.groovy.control.customizers</package>
                 <package>org.codehaus.groovy.control.customizers.builder</package>
-                <package>org.codehaus.groovy.control.io</package>
                 <package>org.codehaus.groovy.control.messages</package>
                 <package>org.codehaus.groovy.reflection</package>
                 <package>org.codehaus.groovy.runtime</package>
--- a/groovy/libs.groovy/nbproject/project.xml
+++ b/groovy/libs.groovy/nbproject/project.xml
@@ -44,6 +44,7 @@
                 <package>org.codehaus.groovy.control</package>
                 <package>org.codehaus.groovy.control.customizers</package>
                 <package>org.codehaus.groovy.control.customizers.builder</package>
+                <package>org.codehaus.groovy.control.io</package>
                 <package>org.codehaus.groovy.control.messages</package>
                 <package>org.codehaus.groovy.reflection</package>
                 <package>org.codehaus.groovy.runtime</package>
--- a/ide/gsf.testrunner.ui/apichanges.xml
+++ b/ide/gsf.testrunner.ui/apichanges.xml
@@ -64,6 +64,19 @@
         </description>
         <class package="org.netbeans.modules.gsf.testrunner.ui.api" name="TestMethodController"/>
     </change>
+    <change id="ComputeTestMethodsAndTestMethodGetTestClassPosition">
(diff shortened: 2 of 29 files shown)


ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/spi/ComputeTestMethods.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public void cancel();
+
+    /**
+     * Factory for creating {@link ComputeTestMethods} instances.
+     * It should be registered in the {@link MimeLookup}.
+     *
+     * @since 1.25
+     */
+    public interface Factory {
```
JaroslavTulach, line 56: Interface, a factory & co. seems a bit too verbose. I would prefer interface with a single method `computeTestMethods` and some parameter:
```java
interface ComputeTestMethods {
  public void computeTestMethods(Info info);
  public final class Info {
     private Info()
     public Parser.Result getResult();
     public boolean addMethod(TestMethod m); // if returns false, consider the request cancelled
  }
}
```
with this approach the implementors need to implement just a single interface with a single method. Only one instance needs to be created. Evolution plan: add new methods into the `Info` class.

groovy/libs.groovy/nbproject/project.xml
```diff
@@ -44,6 +44,7 @@
                 <package>org.codehaus.groovy.control</package>
                 <package>org.codehaus.groovy.control.customizers</package>
                 <package>org.codehaus.groovy.control.customizers.builder</package>
+                <package>org.codehaus.groovy.control.io</package>
```
JaroslavTulach, line 47: This change should be associated with some versioning. Increment spec version of the `libs.groovy` module; make sure those who depend on the `org.codehaus.groovy.control.io` package require the new version.

ide/gsf.testrunner.ui/apichanges.xml
```diff
@@ -64,6 +64,19 @@
...
     </change>
+    <change id="ComputeTestMethodsAndTestMethodGetTestClassPosition">
+        <api name="CommonTestrunnerUIAPI"/>
+        <summary>Added ComputeTestMethods SPI and getTestClassPosition() to TestMethod</summary>
+        <version major="1" minor="25"/>
+        <date day="26" month="6" year="2021"/>
+        <author login="dbalek"/>
+        <compatibility addition="yes"/>
+        <description>
+            Added SPI to provide a list of {@link TestMethod}s found in a parsed {@link Source}s.
```
JaroslavTulach, line 75: `{@link}` is unlikely to work. Use `<a href="@TOP@/pkg/TestMethod.html">`

JaroslavTulach (approved):
Consider removing the `Factory` interface. Otherwise fine.

groovy/libs.groovy/nbproject/project.xml
dbalek: Removed. Already part of https://github.com/apache/netbeans/pull/3027

ide/gsf.testrunner.ui/apichanges.xml
dbalek: Fixed.

ide/gsf.testrunner.ui/src/org/netbeans/modules/gsf/testrunner/ui/spi/ComputeTestMethods.java
dbalek: `Factory` interface removed.

PR #3025, opened 2021-06-29
Completion provider for VS Code's launch.json.

entlicher:
The server side can provide completion for run/debug configurations that are put into launch.json file.
The launch.json is parsed on the client and depending on the position, the server is asked to provide either:
- whole configurations, or
- additional attributes into an existing configuration, or
- value of an attribute in a configuration

We're adding a dependency on `jsonc-parser`, a single parsing by the client simplifies completion providers. It's added into a new file `launchConfigurations.ts` not to add more clutter into the main `extension.ts`. The client also handles the necessary commas before/after the inserted text.

A new command `java.project.configuration.completion` is called to retrieved the completion from the server.
To add a completion it's enough to implement and register `LaunchConfigurationCompletion`.
We provide two implementations for now:
- `AttachConfigurationCompletion` to complete debugger attach configurations and
- `ProjectConfigurationCompletion` to complete project configurations

The attach logic was redesigned a bit, mainly because the name of the configuration can be edited now and thus we can not use it as the identifier of the attach connector. We need to infer the correct connector from the configuration attributes.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
@@ -18,6 +18,7 @@
  */
 package org.netbeans.modules.java.lsp.server;
 
+import com.google.gson.stream.JsonWriter;
 import com.sun.source.tree.CompilationUnitTree;
 import com.sun.source.tree.LineMap;
 import com.sun.source.tree.Tree;
@@ -27,6 +28,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.StringWriter;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.util.Properties;
@@ -206,4 +208,57 @@ public class Utils {
     private static File getCacheDir() {
         return Places.getCacheSubfile("java-server");
     }
+
+    private static final char[] SNIPPET_ESCAPE_CHARS = new char[] { '\\', '$', '}' };
+    /**
+     * Escape special characters in a completion snippet. Characters '$' and '}'
+     * are escaped via backslash.
+     */
+    public static String escapeCompletionSnippetSpecialChars(String text) {
+        if (text.isEmpty()) {
+            return text;
+        }
+        for (char c : SNIPPET_ESCAPE_CHARS) {
+            StringBuilder replaced = null;
+            int lastPos = 0;
+            int i = 0;
+            while ((i = text.indexOf(c, i)) >= 0) {
+                if (replaced == null) {
+                    replaced = new StringBuilder(text.length() + 5); // Text length + some escapes
+                }
+                replaced.append(text.substring(lastPos, i));
+                replaced.append('\\');
+                lastPos = i;
+                i += 1;
+            }
+            if (replaced != null) {
+                replaced.append(text.substring(lastPos, text.length()));
+                text = replaced.toString();
+            }
+            replaced = null;
+        }
+        return text;
+    }
+
+    /**
+     * Encode a String value to a valid JSON value. Enclose into quotes explicitly when needed.
+     */
+    public static String encode2JSON(String value) {
+        if (value.isEmpty()) {
+            return value;
+        }
+        StringWriter sw = new StringWriter();
+        try (JsonWriter w = new JsonWriter(sw)) {
+            w.beginArray();
+            w.value(value);
+            w.endArray();
+            w.flush();
+        } catch (IOException ex) {
+            Exceptions.printStackTrace(ex);
+        }
+        String encoded = sw.toString();
+        // We have ["value"], remove the array and quotes
+        return encoded.substring(2, encoded.length() - 2);
+    }
+
 }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.java.lsp.server.debugging.attach;
+
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.InsertTextFormat;
+import org.netbeans.api.project.Project;
+import org.netbeans.modules.java.lsp.server.Utils;
+import org.netbeans.modules.java.lsp.server.protocol.LaunchConfigurationCompletion;
+import org.openide.util.Exceptions;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ * Completion of debugger attach configurations.
+ *
+ * @author Martin Entlicher
+ */
+@ServiceProvider(service = LaunchConfigurationCompletion.class, position = 200)
+public class AttachConfigurationCompletion implements LaunchConfigurationCompletion {
+
+    @Override
+    public CompletableFuture<List<CompletionItem>> configurations(Supplier<CompletableFuture<Project>> projectSupplier) {
+        return CompletableFuture.supplyAsync(() -> {
+            return createCompletion(AttachConfigurations.get());
+        }, AttachConfigurations.RP);
+    }
+
+    @Override
+    public CompletableFuture<List<CompletionItem>> attributes(Supplier<CompletableFuture<Project>> projectSupplier, Map<String, Object> currentAttributes) {
+        return CompletableFuture.supplyAsync(() -> {
+            return createAttributesCompletion(AttachConfigurations.get(), currentAttributes);
+        }, AttachConfigurations.RP);
+    }
+
+    @Override
+    public CompletableFuture<List<CompletionItem>> attributeValues(Supplier<CompletableFuture<Project>> projectSupplier, Map<String, Object> currentAttributes, String attribute) {
+        return CompletableFuture.completedFuture(Collections.emptyList());
+    }
+
+    private static List<CompletionItem> createCompletion(AttachConfigurations attachConfigurations) {
+        return attachConfigurations.getConfigurations().stream().map(configAttrs -> createCompletion(configAttrs)).collect(Collectors.toList());
+    }
+
+    private static CompletionItem createCompletion(ConfigurationAttributes configAttrs) {
+        CompletionItem ci = new CompletionItem("Java 8+: " + configAttrs.getName());    // NOI18N
+        StringWriter sw = new StringWriter();
+        try (JsonWriter w = new JsonWriter(sw)) {
+            w.setIndent("\t");                                              // NOI18N
+            w.beginObject();
+            w.name("name").jsonValue("\"${1:" + Utils.escapeCompletionSnippetSpecialChars(Utils.encode2JSON(configAttrs.getName())) + "}\""); // NOI18N
+            w.name("type").value(AttachConfigurations.CONFIG_TYPE);         // NOI18N
+            w.name("request").value(AttachConfigurations.CONFIG_REQUEST);   // NOI18N
+            int locationIndex = 2;
+            for (Map.Entry<String, ConfigurationAttribute> entry : configAttrs.getAttributes().entrySet()) {
+                ConfigurationAttribute ca = entry.getValue();
+                if (ca.isMustSpecify()) {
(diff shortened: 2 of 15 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
```diff
@@ -206,4 +206,35 @@ public static synchronized FileObject fromUri(String uri) throws MalformedURLExc
     private static File getCacheDir() {
         return Places.getCacheSubfile("java-server");
     }
+
+    private static final char[] SNIPPET_ESCAPE_CHARS = new char[] { '\\', '$', '}' };
```
sdedic, line 210: OK, `\` must come first :)

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
```diff
@@ -0,0 +1,176 @@
...
+            }
+        }
+        if (completionItems != null) {
+            return completionItems;
+        } else {
+            return Collections.emptyList();
+        }
+    }
+
+    /*
```
sdedic, line 159: Leftover ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
```diff
@@ -0,0 +1,176 @@
...
+        ConfigurationAttributes currentConfiguration = attachConfigurations.findConfiguration(currentAttributes);
+        if (currentConfiguration != null) {
+            Map<String, ConfigurationAttribute> attributes = currentConfiguration.getAttributes();
+            for (Map.Entry<String, ConfigurationAttribute> entry : attributes.entrySet()) {
+                String attrName = entry.getKey();
+                if (!currentAttributes.containsKey(attrName)) {
+                    StringWriter sw = new StringWriter();
+                    try (JsonWriter w = new JsonWriter(sw)) {
+                        w.beginObject();
+                        w.name(attrName).value(entry.getValue().getDefaultValue());
```
sdedic, line 134: Is escaping unnecessary here ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
```diff
@@ -0,0 +1,122 @@
...
+            CompletionItem ci = new CompletionItem("Java 8+: " + launchName);   // NOI18N
+            StringWriter sw = new StringWriter();
+            try (JsonWriter w = new JsonWriter(sw)) {
+                w.setIndent("\t");                                          // NOI18N
+                w.beginObject();
+                w.name("name").value(launchName);                           // NOI18N
+                w.name("type").value(CONFIG_TYPE);                          // NOI18N
+                w.name("request").value("launch");                          // NOI18N
+                w.name("mainClass").value("${file}");                       // NOI18N
+                w.name("launchConfiguration").value(configDisplayName);     // NOI18N
```
sdedic, line 92: I wonder if we shouldn't finally add nonlocalizable IDs to the configuration API.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
```diff
@@ -0,0 +1,122 @@
...
+        List<CompletionItem> completionItems = new ArrayList<>(configurations.size() - 1);
+        boolean skipFirst = true;
+        for (ProjectConfiguration c : configurations) {
+            if (skipFirst) {
+                skipFirst = false;
+                continue;
+            }
+            String configDisplayName = c.getDisplayName();
+            String launchName = Bundle.LBL_LaunchJavaConfig(configDisplayName);
+            CompletionItem ci = new CompletionItem("Java 8+: " + launchName);   // NOI18N
```
sdedic, line 83: Constant or localizable String

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
```diff
@@ -0,0 +1,122 @@
...
+        } else {
+            return CompletableFuture.completedFuture(Collections.emptyList());
+        }
+    }
+
+    @NbBundle.Messages({"# {0} - Configuration name", "LBL_LaunchJavaConfig=Launch Java: {0}",
+                        "# {0} - Configuration name", "LBL_LaunchJavaConfig_desc=Launch a Java 8+ application using {0}."})
+    private static List<CompletionItem> createConfigurationsCompletion(Project p) {
+        ProjectConfigurationProvider<ProjectConfiguration> provider = p.getLookup().lookup(ProjectConfigurationProvider.class);
+        Collection<ProjectConfiguration> configurations = provider.getConfigurations();
```
sdedic, line 73: Check for `null`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
```diff
@@ -0,0 +1,122 @@
...
+            ci.setInsertText(sw.toString());
+            ci.setDocumentation(Bundle.LBL_LaunchJavaConfig_desc(configDisplayName));
+            completionItems.add(ci);
+        }
+        return completionItems;
+    }
+
+    private List<CompletionItem> createLaunchConfigCompletion(Project p) {
+        ProjectConfigurationProvider<ProjectConfiguration> provider = p.getLookup().lookup(ProjectConfigurationProvider.class);
+        Collection<ProjectConfiguration> configurations = provider.getConfigurations();
```
sdedic, line 107: Check for `null`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
```diff
@@ -0,0 +1,122 @@
...
+        List<CompletionItem> completionItems = new ArrayList<>(configurations.size() - 1);
+        boolean skipFirst = true;
+        for (ProjectConfiguration c : configurations) {
+            if (skipFirst) {
+                skipFirst = false;
+                continue;
+            }
+            String configDisplayName = c.getDisplayName();
+            CompletionItem ci = new CompletionItem(configDisplayName);
+            ci.setInsertText("\"" + configDisplayName + "\"");
```
sdedic, line 117: The presets do not use weird chars as quotes, but for the future, better escape possible `"` when creating a String literal.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
```diff
@@ -241,6 +247,45 @@
             case Server.JAVA_FIND_DEBUG_PROCESS_TO_ATTACH: {
                 return AttachConfigurations.findProcessAttachTo(client);
             }
+            case Server.JAVA_PROJECT_CONFIGURATION_COMPLETION: {
```
sdedic, line 250: Maybe document the supported formats of the request in a comment.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LaunchConfigurationCompletion.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @author Martin Entlicher
+ */
+public interface LaunchConfigurationCompletion {
+
+    /**
+     * Provide configurations of Run/Debug actions.
+     *
+     * @param projectSupplier Supplier of the relevant project
+     * @return a list of completion items
```
sdedic, line 42: I'd document that the impl cannot return `null` even in case of unsupported projects, LSP code relies on that; or improve LSP code to handle possible `null`s.

java/java.lsp.server/vscode/src/launchConfigurations.ts
```diff
@@ -0,0 +1,147 @@
...
+                            // Get all configurations:
+                            completionItems = commands.executeCommand('java.project.configuration.completion', uri);
+                        } else {
+                            let node: jsoncp.Node = currentNode;
+                            if (currentNode.type == 'property' && currentNode.parent) {
+                                let propName = currentNode.children?.[0]?.value;
+                                node = currentNode.parent;
+                                let attributesMap = getAttributes(node);
+                                // Get possible values of property 'propName':
+                                completionItems = commands.executeCommand('java.project.configuration.completion', uri, attributesMap, propName);
```
sdedic, line 48: OK, if `propName` is undef ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
```diff
@@ -206,4 +206,35 @@ public static synchronized FileObject fromUri(String uri) throws MalformedURLExc
...
+    /**
+     * Escape special characters in a completion snippet. Characters '$' and '}'
+     * are escaped via backslash.
+     */
+    public static String escapeCompletionSnippetSpecialChars(String text) {
+        if (text.isEmpty()) {
```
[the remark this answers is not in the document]
entlicher: I'm not calling it with `null` anywhere AFAIK.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
entlicher: Yes, thanks!

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
entlicher: Yes because `value()` does the escaping itself. `jsonValue()` does not do the escaping.
We need to apply escaping ourselfs and use the `jsonValue()` only when we use `${}` in the snippets so that we do not get it escaped for the snippet apply logic to work.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: I didn't seen that in other extensions. Did you?

sdedic: Sorry - that was for our NetBeans Project API. ProjectConfiguration API interface does not export its ID, but maven + gradle internally have one, to serve as persistent handle; nonlocalizable (while the displayName is localized).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: We do not localize "Java 8+" anywhere AFAIK. It's the identification name of our integration. I have seen that name of other configurations always start with the same name. Like "Chrome: ...", "Edge: ...", "Node.js: ..." And for us it's "Java 8+: ..."
It makes no sense to me to localize the prefix. But, maybe we should address the right-to-left orientation in some languages.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LaunchConfigurationCompletion.java
entlicher: Right. `a list of completion items` means that it needs to return a `List` and thus not `null`, as I understand it.
I'll make that explicit.

java/java.lsp.server/vscode/src/launchConfigurations.ts
entlicher: I don't think it can be undefined. A property ought to have a name.
If this can ever happen, I'd not call anything and return simply and empty CompletionList.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: I don't think this is a problem. The user definitely does not want to have two configurations localized to the identical String. And when they are different, we don't have a problem.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: I do not see how the other launch configurations look like in right-to-left languages. The whole UI component would need to be right-to-left and that's beyond our control. Issue https://github.com/microsoft/vscode/issues/11770 is still open. I'd rather keep it like it is.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/attach/AttachConfigurationCompletion.java
entlicher: Removed.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: A check added.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
entlicher: Comments added.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LaunchConfigurationCompletion.java
entlicher: Javadoc improved.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectConfigurationCompletion.java
entlicher: Check for null added.

sdedic (approved):

PR #3029, opened 2021-07-01
Source/javadoc attacher must report download actions as enabled.

sdedic:
As part of work on 'Continuous' gradle mode, the infrastructure started to test actions for `isEnabled()` before actually invoking them. That broke source+javadoc attacher introduced in 12.4, that downloads sources/javadocs for the user - the java code never reported the actions as 'enabled', but carried them out anyway. And to be even better, the attacher relied on ActionProgress to report `finished` which never happened, since gradle's `ActionProviderImpl` returned silently without reporting anything if the client invoked a disabled action.
As a result, the Attacher blocked, the user needs to cancel it from the UI (or will run indefinitely if run headless).


dbalek (approved):
Looks fine

JaroslavTulach (approved):

lkishalmi (approved):

PR #3030, opened 2021-07-01
Avoid NPEs when Finder returns DeclarationLocation.NONE or no FObject.

sdedic:
2nd attempt to fix NPEs thrown to LSP client when hovering over source code. The DeclarationFinder sometimes returns DeclarationLocation.NONE (i.e. erroneous annotation in Groovy source), which gets reported through the CompletableFutures and fails later as it contains `null` FileObject.


dbalek (approved):
Thanks.

PR #3032, opened 2021-07-01
Ability to add breakpoints into Groovy sources.

entlicher:
Groovy is added as a debugger language and breakpoints language. We handle the creation of groovy line breakpoints.


JaroslavTulach (approved):

dbalek (approved):
Thanks.

PR #3020, opened 2021-06-27
[NETBEANS-5816] ${header.dir} may contain spaces

nigjo:

PR #3033, opened 2021-07-01
[NETBEANS-5816] ${header.dir} may contain spaces

nigjo:


matthiasblaesing:
Review was done in #3020. The automatic checks like it also. Thank you.

PR #3034, opened 2021-07-01
Full ClassNode loads in indexer

sdedic:
The groovy parser currently loads only stubs for types, which leads to issues like inability to resolve fields or methods; this was done for performance reasons, but as the ClassNode remain in caches, the crippled data are used during opened editor source parsing and report strange errors, such as a symbol is not an annotation, of that a field / method does not exist etc.

I would somewhat prefer usability over performance, and try to improve performance by further pull request throughout the next release.

As more complete ClassNodes are loaded, richer typeinfo can be used in generic handling code - I needed to fix some tests since the golden files listed methods like `add(T)` even for explicitly bound typeargs, e.g. for `List<String>`.


dbalek (approved):
Looks fine.

PR #3036, opened 2021-07-02
Prefer .class execution over .java if the class already exists

JaroslavTulach:
Bug report: Create single `Hello.java` in an empty folder, with its `main`... method. Run and debug using JDK8. `Hello.class` file appears in the same folder. Change JDK to JDK11. Debug it. Following output prints to Debug console and it does not debug:
```
Listening on 51317 
User program running 
LineBreakpoint Hello.java : 3 successfully submitted.
error: class found on application class path: Hello
```

This PR fixes the problem by recompiling the `.class` if it already exists and continues executing using `java -cp ...` just like on JDK8 even on newer JDKs.

### DIFF
--- a/java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/LaunchProcess.java
+++ b/java/java.api.common/src/org/netbeans/modules/java/api/common/singlesourcefile/LaunchProcess.java
@@ -53,7 +53,7 @@ final class LaunchProcess implements Callable<Process> {
 
     private Process setupProcess(String port) throws InterruptedException {
         try {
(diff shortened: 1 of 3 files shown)


sdedic:
This compil-or-run logic is in the IDE, but is not applied form commandline, so surprises can happen.
Is there a feasible way to print a warning to the execution output windows ? I mean the .class file lef tin the folder may on purpose, or just a garbage so we can point that to the user.

sdedic (approved):

JaroslavTulach:
> I mean the .class file lef tin the folder may on purpose, or just a garbage

This PR treats such `.class` file as garbage.

PR #3037, opened 2021-07-02
LSP: spock test discovery improvements

dbalek:
Spock test methods found based on "org.spockframework.runtime.model.FeatureMetadata" annotation presence.

### DIFF
--- a/groovy/groovy.support/src/org/netbeans/modules/groovy/support/actions/GroovyComputeTestMethods.java
+++ b/groovy/groovy.support/src/org/netbeans/modules/groovy/support/actions/GroovyComputeTestMethods.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.swing.text.Position;
+import org.codehaus.groovy.ast.AnnotationNode;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.MethodNode;
 import org.codehaus.groovy.ast.ModuleNode;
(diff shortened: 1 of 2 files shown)


groovy/groovy.support/src/org/netbeans/modules/groovy/support/actions/GroovyComputeTestMethods.java
```diff
@@ -53,19 +54,24 @@
...
-                        int startOffset = getOffset(text, startLine, startColumn);
-                        int endOffset = getOffset(text, endLine, endColumn);
-                        result.add(new TestMethodController.TestMethod(classNode.getName(),
-                                new SimplePosition(classOffset),
-                                new SingleMethod(parserResult.getSnapshot().getSource().getFileObject(), methodNode.getName()),
-                                new SimplePosition(startOffset),
-                                new SimplePosition(startOffset),
-                                new SimplePosition(endOffset)));
+                    for (AnnotationNode annotation : methodNode.getAnnotations()) {
+                        if ("org.spockframework.runtime.model.FeatureMetadata".equals(annotation.getClassNode().getName())) {
```
sdedic, line 58: Q: does this mean that previously some methods (without FeatureMetadata annotation) could be to the test results although they weren't eligible actually ?

dbalek: Yes, previously all methods of classes extending `spock.lang.Specification` were erroneously marked as Spock test methods (including e.g. `setup()`, `cleanup()`, ...)

sdedic (approved):

PR #3038, opened 2021-07-03
Jackpot rule files migration

mbien:
Added OptionsExport config to the java refactoring module so that NetBeans can migrate custom "inspect and transform" rules to new config folders automatically.

### DIFF
--- a/java/refactoring.java/src/org/netbeans/modules/refactoring/java/resources/Bundle.properties
+++ b/java/refactoring.java/src/org/netbeans/modules/refactoring/java/resources/Bundle.properties
@@ -17,3 +17,5 @@
 
 OpenIDE-Module-Name=Java Refactoring
 OpenIDE-Module-Display-Category=Java
+
+RefactoringRules.Options.Export.displayName=Refactoring Rules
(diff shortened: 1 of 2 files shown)


JaroslavTulach:
Looks like a nice usability improvement. Was the system even usable when your _custom rules_ got reset with every update to new version of NetBeans? Could you write documentation/blog post describing how you use the _advanced rule files_? I am sure @geertjanw would publish it somewhere!

mbien:
>Looks like a nice usability improvement. Was the system even usable when your custom rules got reset with every update to new version of NetBeans?

I used to copy some configs manually when I updated NetBeans, rule files were just one item on the list, I also maintain my own etc/netbeans.conf, pointing to a different JDK location and with slightly different JVM flags.
(tried to improve the situation there too but I didn't anticipate the windows launcher https://github.com/apache/netbeans/pull/3002)

>Could you write documentation/blog post describing how you use the advanced rule files?

i wrote one a decade ago, quite dated by now :)
https://mbien.dev/blog/entry/you_have_won_the_jackpot

If I come up with a interesting use case I might write another one. Its just that I am often too lazy to write anything, esp when I also have the option to code instead.

PR #3040, opened 2021-07-07
Fixed source+javadoc Gradle download

sdedic:
There was a typo in PR #3029 -- an [inverted condition](https://github.com/apache/netbeans/pull/3029/files#diff-5c103f8068d80ea76751550a19a5c963ccf80879987162a74edba320ee218eefR306).  The actual [typo is fixed here](https://github.com/apache/netbeans/pull/3040/files#diff-5c103f8068d80ea76751550a19a5c963ccf80879987162a74edba320ee218eefR306) - one liner change. There was yet another [subtle bug (still unnoticed)](https://github.com/apache/netbeans/pull/3040/files#diff-c49a8b6e435ec670b8a00303efd7cee16fa59f7ed89ecde40a3b5766b20814acR177): if plugin did not attach anything the API still reported success through the listener.

The rest of this PR are tests and test infrastructure: for the source download to be reliably initiated, I've needed to run Gradle in tests with a different gradle user home, so the gradle cache is not polluted by previously downloaded artifacts. So the rest of changes basically allows the test to change gradle user home for executed gradle tasks and ProjectConnection commands - but is not exposed through APIs, it's only accessible to test using accessor classes.

I felt like I should provide a test to demonstrate the bug is finally gone...

### DIFF
--- a/java/gradle.java/test/unit/src/org/netbeans/modules/gradle/queries/GradleSourceAttacherImplTest.java
+++ b/java/gradle.java/test/unit/src/org/netbeans/modules/gradle/queries/GradleSourceAttacherImplTest.java
@@ -1,0 +1,176 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.gradle.queries;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Locale;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import static junit.framework.TestCase.assertNotNull;
+import org.junit.Assume;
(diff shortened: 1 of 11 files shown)


java/gradle.java/test/unit/src/org/netbeans/modules/java/j2seplatform/api/Bundle_da.properties
```diff
@@ -0,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
lkishalmi, line 1: da = yes in slavic languages, it's a minor, but can't we name it Bundle_yes then? At first I thought that this is for Danish locale.

lkishalmi (approved):
Just a minor thing which would be nice to change, but otherwise it can go!

Thanks for the accessor and the GradleHomeProvider!

sdedic:
@lkishalmi "da" was used elsewhere ;) and selected because "da" (Russian) = "yes" (english) && is a valid locale code (so you can actually do `new Locale("DA")` without an exception.

lkishalmi:
Oh functional cheating! I'm fine with that.

java/gradle.java/test/unit/src/org/netbeans/modules/gradle/queries/GradleSourceAttacherImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+                result.complete(false);
+            }
+        }
+        
+        // need to get the classpath root inside the archive.
+        URL archiveRoot = FileUtil.getArchiveRoot(utilJar.toURI().toURL());
+        L l = new L();
+        SourceJavadocAttacher.attachSources(archiveRoot, l);
+        // wait at most 20 seconds for a download.
+        boolean r = l.result.get(20, TimeUnit.SECONDS);
```
jlahoda, line 173: I suspect this might lead to random failures, if the artifacts cannot be downloaded quickly enough. Possibly not too problematic, but thought I'd point that out.
Maybe it would be possible to setup a repository with dummy jars on the local disk and the do something like:
```
    maven {
        url "file://<path>"
    }
```
instead of `mavenCentral()` in `build.gradle`?

jlahoda (approved):
I am not an expert in this area - looks reasonable (except maybe the downloading from Maven central), but I don't think that means much.

java/gradle.java/test/unit/src/org/netbeans/modules/gradle/queries/GradleSourceAttacherImplTest.java
sdedic: Hm ! Certainly that would improve the stability, especially in the case of a clean gradle user home.

PR #3041, opened 2021-07-07
profiler BUILDINFO.txt shouldn't be versionned

ebarboni:


lbruun:
Forgive me, I have possibly forgotten everything about how this build works, but at the moment I cannot really see how `BUILDINFO.txt` would ever end up in that location?  From what I remember `BUILDINFO.txt` is only created when the build runs in GitHub Actions. 
(it is created in https://github.com/apache/netbeans/blob/master/.github/workflows/native-binary-build-lib.profiler.yml)  

Sure, if you take the ZIP with the build artifacts and explode into that dir then you'll have make sure not to commit it. Is that what you mean?

ebarboni:
@lbruun zip seems deflated with BUILDINFO.txt in release folder. I didn't notice lib folder were already ignored. 
After an ant execution BUILDINFO.txt is the only file that is detected changed by git. So we should ignore it suppose.

I would like to maybe set another repo to build profiler native in a more votable way like https://github.com/apache/netbeans-native-launchers but I'm not sure how to do that for multiple platform at once yet.

lbruun:
Apologies, I forgot that when the Ant build script runs it puts the downloaded files into `release` folder. 

With that your change is for sure valid, Go ahead and merge.

And yes, it is a good idea to move into own repo. It should be pretty straight-forward to do for this particular build, however you must

1. Remember to bring the GitHub Actions build script with you (https://github.com/apache/netbeans/blob/master/.github/workflows/native-binary-build-lib.profiler.yml)
1. Change (1) in the step(s) which generate the JNI headers. See the comments where it says "the sole purpose".
1. As a result there will be some changes needed to paths and what not in both the yaml file and the build-native script files.

I'll be happy to assist.

ebarboni:
Maybe will need a more mavenized way to have source and binaries a easy to generate. I'm not aware of all native in the code base, will see later.

PR #3042, opened 2021-07-08
Optionally disable groovy support in NBJLS

sdedic:
The groovy parser (or better: type resolution) is terribly slow, so it quite undermines usability of Groovy + Spock sources in the project. This PR makes it possible to disable indexing - with all the negative effects on the quality of code completion when the source is finally parsed - so that LSP client can disable it.
NetBeans IDE user can eventually disable it too using a system property; but the feature should remain undocumented and officially unsupported, since I (we) hope the parser's performace could be improved during the next release cycle.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
@@ -236,4 +236,25 @@ public final class GroovyUtils {
         }
         return sb.toString();
     }
+    
+    /**
+     * True, if the indexing is enabled. Depends system property or {@link #setIndexingEnabled(boolean)}.
+     * 
+     */
+    private static volatile boolean indexingEnabled = Boolean.valueOf(System.getProperty("org.netbeans.modules.groovy.editor.api.indexingEnabled", "true"));
+    
+    /**
+     * Disables completely Groovy indexing. Temporary options only for 12.5 release, will be hopefully
+     * removed after Groovy performance improves. Currently used reflectively from java.lsp.server module only.
+     * DO NOT expose as an API.
+     * @param enabled 
+     */
+    public static void setIndexingEnabled(boolean enabled) {
+        indexingEnabled = enabled;
+    }
+    
+    public static boolean isIndexingEnabled() {
+        return indexingEnabled;
+    }
+
 }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientCapabilities.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientCapabilities.java
@@ -55,6 +55,11 @@ public final class NbCodeClientCapabilities {
      * </ul>
      */
     private Boolean testResultsSupport;
+    
+    /**
+     * Asks for groovy support. Temporary option, will be removed.
+     */
+    private Boolean wantsGroovySupport;
 
     public ClientCapabilities getClientCapabilities() {
         return clientCaps;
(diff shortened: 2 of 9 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientCapabilities.java
```diff
@@ -55,6 +55,11 @@
      * </ul>
      */
     private Boolean testResultsSupport;
+    
+    /**
+     * Akss for groovy support. Temporary option, will be removed.
```
jlahoda, line 60: typo:
```suggestion
     * Asks for groovy support. Temporary option, will be removed.
```

jlahoda (approved):
Looks reasonable.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
```diff
@@ -236,4 +236,25 @@ private static String stripPackageFromTypeParams(String params) {
         }
         return sb.toString();
     }
+    
+    /**
+     * True, if the indexing is enabled. Depends system property or {@link #setIndexingEnabled(boolean)}.
+     * 
+     */
+    private static volatile boolean indexingEnabled = Boolean.valueOf(System.getProperty("org.netbeans.modules.groovy.editor.api.indexingEnabled", "true"));
```
JaroslavTulach, line 244: This property should be documented in _arch.xml_ file as `group="systemproperty"`.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
```diff
@@ -236,4 +236,25 @@ private static String stripPackageFromTypeParams(String params) {
...
+    
+    /**
+     * Disables completely Groovy indexing. Temporary options only for 12.5 release, will be hopefully
+     * removed after Groovy performance improves. Currently used reflectively from java.lsp.server module only.
+     * DO NOT expose as an API.
+     * @param enabled 
+     */
```
JaroslavTulach, line 251: Missing `@since`?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -829,4 +831,22 @@ public void logMessage(MessageParams message) {
...
+    /**
+     * Hacky way to enable or disable Groovy support. Since it is hack, it will disable Groovy for the whole NBJLS, not just a specific client / project. Should
+     * be revisited after NetBeans 12.5, after Groovy parsing improves
+     * @param caps 
+     */
+    private static void hackConfigureGroovySupport(NbCodeClientCapabilities caps) {
+        boolean b = caps.wantsGroovySupport();
+        try {
+            Class clazz = Lookup.getDefault().lookup(ClassLoader.class).loadClass("org.netbeans.modules.groovy.editor.api.GroovyIndexer");
+            Method m = clazz.getDeclaredMethod("setIndexingEnabled", Boolean.TYPE);
```
JaroslavTulach, line 845: I see. There's a section in _arch.xml_ called `<answer id="exec-reflection">` where groovy could _export_ this reflective API.

java/java.lsp.server/vscode/package.json
```diff
@@ -84,6 +84,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Enable Run/Debug test in editor"
+				},
+				"netbeans.groovySupport.enabled": {
```
JaroslavTulach, line 88: Does changing the value of this option automatically restarts the JVM or does it have to be restarted manually?

JaroslavTulach (commented):
> the feature should remain undocumented and officially unsupported

The feature should have been documented as _private_ API. That's how NetBeans project declares a _feature is unsupported_. E.g. not by not documenting it, but documenting it is a _private_ contract.

java/java.lsp.server/vscode/package.json
sdedic: Not yet.  Should we ever fix it / document it, if we believe we can improve things during 12.5 release ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
sdedic: Will add, but it isn't a public/friend accssible class anyway.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/utils/GroovyUtils.java
sdedic: I considered the 'solution' a hack, so I didn't plan to document it  in arch document; otherwise I'd make a proper friend API call for it. After all, this hack should improve / change during 12.5 release - should we document it in the API/arch docs ?

PR #3043, opened 2021-07-09
[NETBEANS-5878] Added a menu entry to copy fully qualified class names

asbachb:
This PR add a menu to Code Editor menu and Projects menu to copy fully qualified class names.

As I don't think it's worth to create a dedicated module for it I placed it in Java Refactoring module as it seemed for me to best fitting. As I'm not very familiar with NetBeans platform development there might be a need for taking a little bit deeper look into.

20210709_00h19m24s_grim
20210709_00h19m31s_grim

see https://issues.apache.org/jira/browse/NETBEANS-5878

### DIFF
--- a/java/java.source/src/org/netbeans/modules/java/ui/CopyFullyQualifiedClassNameAction.java
+++ b/java/java.source/src/org/netbeans/modules/java/ui/CopyFullyQualifiedClassNameAction.java
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-package org.netbeans.modules.refactoring.java.ui;
+package org.netbeans.modules.java.ui;
 
 import java.awt.Toolkit;
 import java.awt.datatransfer.ClipboardOwner;
--- a/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
+++ b/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.modules.refactoring.java.ui;
+
+import java.awt.Toolkit;
+import java.awt.datatransfer.StringSelection;
+import org.netbeans.api.java.classpath.ClassPath;
+import org.openide.awt.ActionID;
+import org.openide.awt.ActionReference;
+import org.openide.awt.ActionReferences;
+import org.openide.awt.ActionRegistration;
+import org.openide.filesystems.FileObject;
+import org.openide.nodes.Node;
+import org.openide.util.HelpCtx;
+import org.openide.util.NbBundle;
+import org.openide.util.actions.NodeAction;
+
+/**
+ * Action to copy fully qualified name of selected classes.
+ * 
+ * @author Benjamin Asbach
+ */
+@ActionID(id = "org.netbeans.modules.refactoring.java.java.ui.CopyFullyQualifiedClassNameAction", category = "Edit")
+@ActionRegistration(displayName = "Copy Fully Qualified Class Name", lazy = false)
+@ActionReferences( value = {
+    @ActionReference(path = "Editors/text/x-java/Popup", position = 3500),
+    @ActionReference(path = "Loaders/text/x-java/Actions", position = 610)
+})
+public class CopyFullyQualifiedClassNameAction extends NodeAction {
+
+    private final String name;
+    
+    public CopyFullyQualifiedClassNameAction() {
+        this.name = NbBundle.getMessage(this.getClass(), "LBL_CopyFullyQualifiedClassNameAction");
+    }
+    
+    @Override
+    protected void performAction(Node[] activatedNodes) {
+        for (Node activatedNode : activatedNodes) {
+            FileObject javaSource = activatedNode.getLookup().lookup(FileObject.class);
+            if (javaSource == null) {
+                continue;
+            }
+
+            ClassPath classPath = ClassPath.getClassPath(javaSource, ClassPath.SOURCE);
+            if (classPath == null) {
+                continue;
+            }
+            String packagePath = classPath.getResourceName(javaSource.getParent());
+            String packageName = packagePath.replace("/", ".");
+
+            String fullyQualifiedClassName;
+            if (isDefaultPackage(packageName)) {
+                fullyQualifiedClassName = javaSource.getName();
+            } else {
+                fullyQualifiedClassName = packageName + "." + javaSource.getName();
+            }
+
+            setClipboardContent(fullyQualifiedClassName);
+        }
+    }
+
+    private boolean isDefaultPackage(String packageName) {
+        return "".equals(packageName);
+    }
+
+    private void setClipboardContent(String content) {
+        Toolkit.getDefaultToolkit().getSystemClipboard().setContents(new StringSelection(content), null);
+    }
+
+    @Override
+    protected boolean enable(Node[] activatedNodes) {
+        return containsAnyFileObject(activatedNodes);
+    }
+
+    private boolean containsAnyFileObject(Node[] activatedNodes) {
+        for (Node activatedNod : activatedNodes) {
+            FileObject fileObject = activatedNod.getLookup().lookup(FileObject.class);
+            if (fileObject != null) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public HelpCtx getHelpCtx() {
+        return HelpCtx.DEFAULT_HELP;
+    }
+
+}
--- a/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
+++ b/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
@@ -22,6 +22,8 @@ package org.netbeans.modules.refactoring.java.ui;
 import java.awt.Toolkit;
 import java.awt.datatransfer.ClipboardOwner;
 import java.awt.datatransfer.StringSelection;
+import java.util.ArrayList;
+import java.util.List;
 import org.netbeans.api.java.classpath.ClassPath;
 import org.openide.awt.ActionID;
 import org.openide.awt.ActionReference;
@@ -56,6 +58,8 @@ public class CopyFullyQualifiedClassNameAction extends NodeAction {
     
     @Override
     protected void performAction(Node[] activatedNodes) {
+        List<String> fullyQualifiedClassNames = new ArrayList<>(activatedNodes.length);
+
         for (Node activatedNode : activatedNodes) {
             FileObject javaSource = activatedNode.getLookup().lookup(FileObject.class);
             if (javaSource == null) {
@@ -76,8 +80,10 @@ public class CopyFullyQualifiedClassNameAction extends NodeAction {
                 fullyQualifiedClassName = packageName + "." + javaSource.getName();
             }
 
-            setClipboardContent(fullyQualifiedClassName);
+            fullyQualifiedClassNames.add(fullyQualifiedClassName);
         }
+
+        setClipboardContent(String.join(System.lineSeparator(), fullyQualifiedClassNames));
     }
 
     private boolean isDefaultPackage(String packageName) {
--- a/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
+++ b/java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
@@ -91,8 +91,8 @@ public class CopyFullyQualifiedClassNameAction extends NodeAction {
     }
 
     private boolean containsAnyFileObject(Node[] activatedNodes) {
-        for (Node activatedNod : activatedNodes) {
-            FileObject fileObject = activatedNod.getLookup().lookup(FileObject.class);
+        for (Node activatedNode : activatedNodes) {
+            FileObject fileObject = activatedNode.getLookup().lookup(FileObject.class);
             if (fileObject != null) {
                 return true;
             }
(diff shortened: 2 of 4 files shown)


eirikbakke:
What is the use case for this feature?

It might be better to make this work with the regular Copy action--i.e. if you copy a node that happens to be a class, and try to paste it into a text editor, it will paste the fully qualified class name. Similarly, copying a node that is a file (but not a Java class) might paste the file path. (Adding new actions for very workflow-specific features makes menus very long...)

geertjanw:
I recommend discussing your approach to solving an issue on the mailing list before implementing it.

matthiasblaesing:
This plugin basicly implements this: https://github.com/Chris2011/nb-copy-fqn

geertjanw:
If we think this is the right way to implement this, let’s merge, after taking a look at the code, but isn’t @eirikbakke’s comment valid that this should be in the standard Copy action?

asbachb:
@eirikbakke 

> What is the use case for this feature?

Everytime I want to share the exact location of a class I'll need a fully qualified class name. Like Spring xml configuration, referencing classes exact location for Jira Tickets, ...

> It might be better to make this work with the regular Copy action--i.e. if you copy a node that happens to be a class, and try to paste it into a text editor, it will paste the fully qualified class name.

That might be an alternative. Tbh I oriented the implementation to the plugin @matthiasblaesing and eclipse. Only downside I'd see it would a little bit slower to access with a mouse than with a context menu in navigator and code editor.

> Adding new actions for very workflow-specific features makes menus very long...

I wonder if it really a workflow specific feature. I agree that before adding items to these kind of menus that should be part of a discussion.

@geertjanw 

> I recommend discussing your approach to solving an issue on the mailing list before implementing it.

Yeah you might be right. Acutally I wanted to get some handy on the platform in preparation of trying to implement `mvnd`.

@matthiasblaesing 

> This plugin basicly implements this: https://github.com/Chris2011/nb-copy-fqn

You're right. It's a reimplementation since the plugin does not work with recent NetBeans version and with quite none dependencies.

Saljack:
I completely agree this function is pretty helpful. I used mentioned plugin too. I need it for adding ignored classes for Spotbugs etc. I think that copy on a node is not a good idea because it is not straightforward. What if you want to paste only classname, filename or path and not fully qualified name?

ebresie:
Is there a JIRA ticket for this?

asbachb:
@ebresie not yet. If it's necessary I can open a ticket.

eirikbakke:
> I think that copy on a node is not a good idea because it is not straightforward. What if you want to paste only classname, filename or path and not fully qualified name?

Then you can just edit the pasted string with the keyboard. Remember--this is just a shortcut. And we don't want to pollute the menu with dozens of options... "Copy", "Copy fully qualified name", "Copy Path" etc...

In any case, copying either the path or the fully qualified name would be strictly an improvement over the current behavior, which is to paste nothing (when pasting into a text editor).

asbachb:
Actually I don't like the idea hiding it behind the copy node functionality as I don't see that should be the default behaviour and so it would a hidden functionality to the user.

The only reason which speaks adding it to node copy would that actually there's no text copy action for that.

I think it's a functionality which should be visible accessible.

I agree that we should check by additional menu entries for context menu. But on the other side than there should be an additional discussion of current items need to be there as well and what's the criteria for an item to be there.

matthiasblaesing:
I tend to agree, that the new meu entry is sensible. I did not check this PR yet, but if copying the FQN of methods and field is not yet implemented, it would make sense to implement it also later.

IMHO arguing that "another place is possible" is not a good argument, many of the current entries in the menu could be argued identically (the whole refactoring submenu). "I have no use for it" is also not a good argument, as this is also true for existing entries, as I don't remember using the CodeFolds menu.

In the end this is a single menu entry and even if more Copy actions are added, we can trivially introduce a submenu (see refactoring). This was deemed important enough that someone bothered to create a plugin for this, pulled it through the time and now another developer looked into this and suggests to pull it into the core codebase.

eirikbakke:
> Actually I don't like the idea hiding it behind the copy node functionality as I don't see that should be the default behaviour and so it would a hidden functionality to the user.

That's a good point.

> This was deemed important enough that someone bothered to create a plugin for this, pulled it through the time and now another developer looked into this and suggests to pull it into the core codebase.

If multiple people find the feature worthy of a separate menu item, then that sounds fair enough to me :-)

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,114 @@
...
+import org.openide.util.NbBundle;
+import org.openide.util.actions.NodeAction;
+
+/**
+ * Action to copy fully qualified name of selected classes.
+ * 
+ * @author Benjamin Asbach
+ */
+@ActionID(id = "org.netbeans.modules.refactoring.java.java.ui.CopyFullyQualifiedClassNameAction", category = "Edit")
+@ActionRegistration(displayName = "Copy Fully Qualified Class Name", lazy = false)
```
eirikbakke, line 41: Is the lazy=false needed here? Isn't there a way to add a non-lazy action that is enabled only when a FileObject is in the lookup?

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,114 @@
...
+            String packageName = packagePath.replace("/", ".");
+
+            String fullyQualifiedClassName;
+            if (isDefaultPackage(packageName)) {
+                fullyQualifiedClassName = javaSource.getName();
+            } else {
+                fullyQualifiedClassName = packageName + "." + javaSource.getName();
+            }
+
+            setClipboardContent(fullyQualifiedClassName);
```
eirikbakke, line 76: Better to make this action enabled with single selections only. Or concatenate the class names with newlines between them if there is more than one, so that the multiple selection case gives a sensible result. But avoid calling setClipboardContent more than once; clipboard operations are sometimes a bit fragile, and can interact in odd ways with external applications.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,114 @@
...
+        Toolkit.getDefaultToolkit().getSystemClipboard().setContents(new StringSelection(content), null);
+    }
+
+    @Override
+    protected boolean enable(Node[] activatedNodes) {
+        return containsAnyFileObject(activatedNodes);
+    }
+
+    private boolean containsAnyFileObject(Node[] activatedNodes) {
+        for (Node activatedNod : activatedNodes) {
```
eirikbakke, line 94: Typo: activatedNod should be activatedNode.

eirikbakke (changes requested):

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
asbachb: I'm unsure I looked at some other actions. I'm not 100% certain what's all happening with `lazy` `true` vs `false`. Changed it to `true` and tested it which worked.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
asbachb: Added the ability to select multiple nodes and setting clipboard just once.

eirikbakke:
Thanks for revising! I tested the latest patch and it works, including with multiple selection and with both the default package and other packages.

One issue for brief discussion: the name "Copy Fully Qualified Class Name" is very long, making the context menu significantly wider (being the widest item in the menu) on Linux and Windows 10 (but probably not on MacOS). Thought it's probably the most descriptive name. I'd be inclined to keep it as it is, if no one has a better idea.

Last, before we can squash-and-merge, I think the author of the PR needs to have a proper email address publicly visible on github, which then becomes part of NetBeans repository's public history.

eirikbakke:
(Oh and do we require a JIRA ticket to be created for every PR these days? I'm not sure...)

BradWalker:
On Sun, Jul 25, 2021 at 3:57 PM Eirik Bakke ***@***.***>
wrote:

> (Oh and do we require a JIRA ticket to be created for every PR these days?
> I'm not sure...)
>
I believe we do.. It helps to keep things "clean"..

-brad w.

asbachb:
> (Oh and do we require a JIRA ticket to be created for every PR these days? I'm not sure...)

Initially I checked the merged PRs with contains some which does not reference an issue. But I've no problem in opening an issue for that.

asbachb:
@eirikbakke 

> I think the author of the PR needs to have a proper email address publicly visible on github, which then becomes part of NetBeans repository's public history.

Added an E-Mail to my public profile.

eirikbakke (approved):

asbachb:
@eirikbakke can I squash + force push?

sdedic:
FWIW, I think that having 'copy qualified name' on file node may not be the best idea. After all, project explorer view is file-oriented, not class-oriented. As someone noted, (simple) name, qualified name or file paths are valid text/plain `Transferables` for a file node. FWIW, the basic Copy should copy a filename or project-relative path to the file into `text/plain` targets, but that's out of scope here.

Also note that there may be *multiple* qualified name(s) as the file can contain nested classes, for example. Or even worse, multiple toplevel classes, which is not widely used, but valid. In this case, copy FQN on file node is not sufficient, or may be misleading (copies the 1st or public toplevel class FQN only). 

The mentioned (obsolete) plugin integrates with the Navigator (which allows to copy FQNs of individual classes based on navigator node selection) and Java Editor (which allows the same by inspecting caret position -> ElementHandle at that position). With a shortcut attached to the navigator / editor action, this could be better approach: Goto Type -> open file (caret positions) -> shortcut-copy-fqn (fqn of found class copied) -> CTRL-F4 (previous editor selects) -> paste.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.modules.refactoring.java.ui;
```
sdedic, line 20: The class does not use any `java.refactoring` concepts; shouldn't it be rather in some `org.netbeans.modules.java.sourceui` or even lower (i.e. next to JavaDataObject) ?

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,123 @@
...
+    
+    public CopyFullyQualifiedClassNameAction() {
+        this.name = NbBundle.getMessage(this.getClass(), "LBL_CopyFullyQualifiedClassNameAction");
+    }
+    
+    @Override
+    protected void performAction(Node[] activatedNodes) {
+        List<String> fullyQualifiedClassNames = new ArrayList<>(activatedNodes.length);
+
+        for (Node activatedNode : activatedNodes) {
```
sdedic, line 63: The action attaches to Java editor, however if the activated node represents an editor, it does not work with caret position; so in a file with inner classes, just the top-level class' FQN is copied, although the caret may point on an inner class.

asbachb: Yes. You're right. Something I did not pay attention to. I wonder if implementing the caret processing is that easy. Maybe we should remove that context entry from the code view for now and just stick with that project view context menu. As providing this functionality only on file class level might be confusing as well.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
asbachb: Actually I struggled with the location. I already thought about putting it next to `JavaDataObject`, but it seems to be misplaced for some kind of a ui component.

Going to move to `org.netbeans.modules.java.ui` which seems to be a better place than now.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/ui/CopyFullyQualifiedClassNameAction.java
asbachb: Removed that popup menu for the editor for now as it's not that easy to implement.

asbachb:
> I think that having 'copy qualified name' on file node may not be the best idea. After all, project explorer view is file-oriented, not class-oriented. As someone noted, (simple) name, qualified name or file paths are valid text/plain `Transferables` for a file node. FWIW, the basic Copy should copy a filename or project-relative path to the file into `text/plain` targets, but that's out of scope here.
> 
> Also note that there may be _multiple_ qualified name(s) as the file can contain nested classes, for example. Or even worse, multiple toplevel classes, which is not widely used, but valid. In this case, copy FQN on file node is not sufficient, or may be misleading (copies the 1st or public toplevel class FQN only).
> 
> The mentioned (obsolete) plugin integrates with the Navigator (which allows to copy FQNs of individual classes based on navigator node selection) and Java Editor (which allows the same by inspecting caret position -> ElementHandle at that position). With a shortcut attached to the navigator / editor action, this could be better approach: Goto Type -> open file (caret positions) -> shortcut-copy-fqn (fqn of found class copied) -> CTRL-F4 (previous editor selects) -> paste.

I slightly agree that the position in projects view might not be optimal. On the other hand I wonder if the other menu entries apply to a file only context. e.g. `Find usages` looks for class usage not for the file usage.

I fully agree that - in case a menu entry is available in editor view - it should be caret position aware as I also see the end user expectation to be that way. So I removed that entry for now.

java/java.source/src/org/netbeans/modules/java/ui/CopyFullyQualifiedClassNameAction.java
```diff
@@ -0,0 +1,122 @@
...
+ * Action to copy fully qualified name of selected classes.
+ * 
+ * @author Benjamin Asbach
+ */
+@ActionID(id = "org.netbeans.modules.refactoring.java.java.ui.CopyFullyQualifiedClassNameAction", category = "Edit")
+@ActionRegistration(displayName = "Copy Fully Qualified Class Name", lazy = true)
+@ActionReferences( value = {
+    @ActionReference(path = "Loaders/text/x-java/Actions", position = 610)
+})
+public class CopyFullyQualifiedClassNameAction extends NodeAction {
```
jtulach, line 48: It is probably not necessary to subclass `NodeAction` - implement [context aware action](https://bits.netbeans.org/12.0/javadoc/org-openide-awt/org/openide/awt/Actions.html#context-java.lang.Class-boolean-boolean-org.openide.util.ContextAwareAction-java.lang.String-java.lang.String-java.lang.String-boolean-) and check for `FileObject`:

asbachb:
Thanks for all of the efforts. I guess this needs some more work. I'm closing the PR for now as I'm currently not going to work on it.

PR #3044, opened 2021-07-09
Report progress from Gradle's CreateFromTemplateHandler

sdedic:
@JaroslavTulach started with a patch for `BaseGradleWizardIterator` that just attaches a `ProgressHandle` to the `TemplateOperation` created; but it turned out that in LSP the actual creation needs to be done asynchronously, otherwise the message reception thread won't be able to receive `workDoneProgressToken` from the client before the creation completes.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
@@ -49,6 +49,7 @@ import org.openide.nodes.Node;
 import org.openide.util.Exceptions;
 import org.openide.util.NbBundle;
 import org.openide.util.Pair;
+import org.openide.util.RequestProcessor;
 import org.openide.util.UserCancelException;
 import org.openide.util.Utilities;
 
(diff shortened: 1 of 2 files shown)


lkishalmi (approved):
Looks good to me!

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
```diff
@@ -98,7 +105,7 @@ private LspTemplateUI() {
     private CompletableFuture<Object> templateUI(DataFolder templates, NbCodeLanguageClient client, ExecuteCommandParams params) {
         CompletionStage<DataObject> findTemplate = findTemplate(templates, client);
         CompletionStage<Pair<DataFolder, String>> findTargetFolderAndName = findTargetAndName(findTemplate, client, params);
-        return findTargetFolderAndName.thenCombine(findTemplate, (targetAndName, source) -> {
+        return findTargetFolderAndName.thenCombineAsync(findTemplate, (targetAndName, source) -> {
```
JaroslavTulach, line 108: I see.

JaroslavTulach (commented):
Thanks a lot for the fix. The progress is now really visible!

PR #3047, opened 2021-07-13
Visually differentiate lines with DWARF information during native image debugging.

entlicher:
We retrieve source location information and variable symbols from the native debuggee and compare it with method bodies and static fields.
When we find a code location that is not represented in the native debuggee, we use use the introduced text decorators to shadow the unused code.

### DIFF
--- a/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/Location.java
+++ b/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/Location.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.nativeimage.api;
+
+/**
+ * A location in the debuggee program.
+ *
+ * @since 0.2
+ */
+public final class Location {
+
+    private final long pc;
+    private final int line;
+    private final int column;
+
+    private Location(long pc, int line, int column) {
+        this.pc = pc;
+        this.line = line;
+        this.column = column;
+    }
+
+    /**
+     * Get the program counter position.
+     *
+     * @since 0.2
+     */
+    public long getPC() {
+        return pc;
+    }
+
+    /**
+     * Get 1-based line number. Returns 0 when the line is not defined.
+     *
+     * @since 0.2
+     */
+    public int getLine() {
+        return line;
+    }
+
+    /**
+     * Get 1-based column number. Returns 0 when the column is not defined.
+     *
+     * @since 0.2
+     */
+    public int getColumn() {
+        return column;
+    }
+
+    /**
+     * Creates a builder to build a new {@link Location}.
+     *
+     * @since 0.2
+     */
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    @Override
+    public String toString() {
+        return "Location{" + "pc=" + pc + ", line=" + line + ", column=" + column + '}';
+    }
+
+    /**
+     * Location's builder.
+     *
+     * @since 0.2
+     */
+    public static final class Builder {
+
+        private long pc;
+        private int line;
+        private int column;
+
+        Builder() {}
+
+        /**
+         * Set a program counter location in the native binary.
+         *
+         * @since 0.2
+         */
+        public void pc(long pc) {
+            this.pc = pc;
+        }
+
+        /**
+         * Set an 1-based line. The line is treated as unknown when 0.
+         *
+         * @since 0.2
+         */
+        public void line(int line) {
+            if (line < 0) {
+                throw new IllegalArgumentException("Line must not be negative");
+            }
+            this.line = line;
+        }
+
+        /**
+         * Set an 1-based column. The column is treated as unknown when 0.
+         *
+         * @since 0.2
+         */
+        public void column(int column) {
+            if (column < 0) {
+                throw new IllegalArgumentException("Column must not be negative");
+            }
+            this.column = column;
+        }
+
+        /**
+         * Build the {@link Location} object.
+         *
+         * @since 0.2
+         */
+        public Location build() {
+            if (column > 0 && line == 0) {
+                throw new IllegalStateException("Column can not be defined without a line.");
+            }
+            if (line == 0 && pc == 0) {
+                throw new IllegalStateException("No location information is defined.");
+            }
+            return new Location(pc, line, column);
+        }
+    }
+}
--- a/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/SourceInfo.java
+++ b/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/SourceInfo.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.nativeimage.api;
+
+import java.util.Objects;
+
+/**
+ * A source file information in the debuggee program.
+ *
+ * @since 0.2
+ */
+public final class SourceInfo {
+
+    private final String fileName;
+    private final String fullName;
+
+    private SourceInfo(String fileName, String fullName) {
+        this.fileName = fileName;
+        this.fullName = fullName;
+    }
+
+    /**
+     * Get the file name, or <code>null</code> when unknown. May return a relative path.
+     *
+     * @since 0.2
+     */
+    public String getFileName() {
+        return fileName;
+    }
+
+    /**
+     * Get the full file name, or <code>null</code> when unknown. Returns an absolute path, if any.
+     *
+     * @since 0.2
+     */
+    public String getFullName() {
+        return fullName;
+    }
+
+    /**
+     * Creates a builder to build a new {@link SourceInfo}.
+     *
+     * @since 0.2
+     */
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 7;
+        hash = 17 * hash + Objects.hashCode(this.fileName);
+        hash = 17 * hash + Objects.hashCode(this.fullName);
+        return hash;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final SourceInfo other = (SourceInfo) obj;
+        if (!Objects.equals(this.fileName, other.fileName)) {
+            return false;
+        }
+        if (!Objects.equals(this.fullName, other.fullName)) {
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public String toString() {
+        return "SourceInfo{" + "fileName=" + fileName + ", fullName=" + fullName + '}';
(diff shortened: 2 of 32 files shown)


ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/Location.java
```diff
@@ -0,0 +1,140 @@
...
+        this.line = line;
+        this.column = column;
+    }
+
+    /**
+     * Get the program counter position.
+     *
+     * @since 0.2
+     */
+    public long getPc() {
```
JaroslavTulach, line 43: `getPC`?

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/Symbol.java
```diff
@@ -0,0 +1,131 @@
...
+    public String toString() {
+        return "Symbol{" + "name=" + name + ", type=" + type + ", description=" + description + '}';
+    }
+
+    /**
+     * Symbol's builder.
+     *
+     * @since 0.2
+     */
+    public static final class Builder {
```
JaroslavTulach, line 84: Usage of builders seems fine.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/spi/debug/NIDebuggerProvider.java
```diff
@@ -120,4 +124,24 @@
...
+
+    /**
+     * Provide a list of locations for a given file path.
+     *
+     * @param filePath a file path
+     * @return list of locations, or <code>null</code> when there's no location
+     *         information about such file.
+     * @since 0.2
+     */
+    default List<Location> listLocations(String filePath) {
```
JaroslavTulach, line 136: Consider adding subinterface `SymbolProvider extends Provider`

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
```diff
@@ -165,6 +170,49 @@ public String getVersion() {
...
+     *
+     * @param name a name pattern
+     * @param includeNondebug include also symbols from the symbol table
+     * @param maxResults maximum number of results
+     * @return list of source information and their symbols, or <code>null</code>
+     *         when there are no matching symbols.
+     * @since 0.2
+     */
+    @CheckForNull
+    public List<Pair<SourceInfo, List<Symbol>>> listFunctions(String name, boolean includeNondebug, int maxResults) {
```
JaroslavTulach, line 197: What's the difference between `name` and `filePath` at line 176?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/files/OpenedDocuments.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import javax.swing.text.Document;
+
+/**
+ * Gathers documents opened by the LSP client.
+ *
+ * @author Martin Entlicher
+ */
+public final class OpenedDocuments {
```
JaroslavTulach, line 35: There is an implicit `public` constructor and I don't think it should be there.

JaroslavTulach (commented):
Too huge to review in one day. At least for me.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
entlicher: `name` is the name of the symbol. `filePath` is the path to the file the symbol is declared in.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/files/OpenedDocuments.java
entlicher: I'd need to move it to the `protocol` package then. I'm not sure it really belongs there.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/spi/debug/NIDebuggerProvider.java
entlicher: There's just one implementation and one module that uses it. I'd rather keep the single provider till it grows more...

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
```diff
@@ -165,6 +170,49 @@ public String getVersion() {
...
+     *
+     * @param name a name pattern
+     * @param includeNondebug include also symbols from the symbol table
+     * @param maxResults maximum number of results
+     * @return list of source information and their symbols, or <code>null</code>
+     *         when there are no matching symbols.
+     * @since 0.2
+     */
+    @CheckForNull
+    public List<Pair<SourceInfo, List<Symbol>>> listVariables(String name, boolean includeNondebug, int maxResults) {
```
JaroslavTulach, line 212: `List<Pair<X,List<Y>>>` might rather be `Map<X,List<Y>>` - at least that's more traditional way to represent such structure. One can still use `map.entrySet()` and iterate the elements pair by pair (or rather `Entry` by `Entry`).

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/SourceInfo.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    @Override
+    public String toString() {
+        return "SourceInfo{" + "fileName=" + fileName + ", fullName=" + fullName + '}';
+    }
+
```
JaroslavTulach, line 67: If `SourceInfo` is used in `Map`, then it needs proper `equals` and `hashCode`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/ni/NILocationVisualizer.java
```diff
@@ -0,0 +1,414 @@
...
+import org.openide.util.Lookup;
+import org.openide.util.Pair;
+
+/**
+ *
+ * @author Martin Entlicher
+ */
+public final class NILocationVisualizer implements Consumer<String> {
+
+    private final File niFileSources;
```
JaroslavTulach, line 76: Shouldn't the NetBeans code rather work with `FileObject`s?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientWrapper.java
```diff
@@ -144,4 +144,19 @@ public void semanticHighlighting(SemanticHighlightingParams params) {
     public void notifyProgress(ProgressParams params) {
         remote.notifyProgress(params);
     }
+
+    @Override
+    public CompletableFuture<String> createTextEditorDecoration(DecorationRenderOptions params) {
```
JaroslavTulach, line 149: I am not sure I understand the purpose of these new messages. They look standard (e.g. `window/createTextEditorDecoration`)...

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/OverviewRulerLane.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private static final Map<Integer, OverviewRulerLane> lookup = new HashMap<>();
+
+    static {
+        for (OverviewRulerLane value : OverviewRulerLane.values()) {
+            lookup.put(value.getIntValue(), value);
+        }
+    }
+
+    public static OverviewRulerLane get(Integer intValue) {
```
JaroslavTulach, line 53: Use `int` instead of `Integer`. Simple `switch` would be more effective then a `lookup` hash map.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -824,6 +832,24 @@ public void showMessage(MessageParams messageParams) {
             return x;
         }
 
+        @Override
+        public CompletableFuture<String> createTextEditorDecoration(DecorationRenderOptions params) {
+            logWarning(params);
```
JaroslavTulach, line 837: Shouldn't the methods be abstract? It is strange to see such dummy code in production.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -235,7 +236,6 @@
     /**
      * Documents actually opened by the client.
      */
-    private final Map<String, Document> openedDocuments = new ConcurrentHashMap<>();
```
JaroslavTulach, line 238: `OpenDocuments` class is a refactored version of the map then, right?

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -1604,6 +1604,21 @@ public void showMessage(MessageParams params) {
             public void logMessage(MessageParams arg0) {
                 throw new UnsupportedOperationException("Not supported yet.");
             }
+
+            @Override
+            public CompletableFuture<String> createTextEditorDecoration(DecorationRenderOptions params) {
+                throw new UnsupportedOperationException("Not supported yet.");
```
JaroslavTulach, line 1610: `TestProgressHandlerTest` is using `fail()` while the other tests are using `UnsupportedOperationException` - I guess I like `fail()` more.

JaroslavTulach (approved):
Fine for 12.5 after few more tweaks.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/ni/NILocationVisualizer.java
entlicher: The CPPLite debugger uses java.io.File. The FileObject abstraction is not necessary when we expect local files that can be launched via gdb/lldb.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
entlicher: Good point.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientWrapper.java
entlicher: Yes, they are extensions for VSCode.

entlicher: Decorations are not part of the standard LSP. We need to extend it.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/OverviewRulerLane.java
entlicher: The switch would duplicate the numbers. But it'd save memory and a bit of performance, yes.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
entlicher: I've used the same pattern as there is in `showMessageRequest()`. This class can not be abstract, it's the stub LSP client, which is used only when no LSP client is found (no impl of `NbCodeLanguageClient` is registered).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/NbCodeClientWrapper.java
JaroslavTulach: Then it is surprising we put the messages into the `window/` namespace. But I see that we have done that even before with `window/showInputBox` & co.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
entlicher: Yes, it's a wrapped map with added ability to listen for changes (consume the added content).

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
entlicher: O.K.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
entlicher: Done.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/SourceInfo.java
entlicher: Yes, done.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/ni/NILocationVisualizer.java
entlicher: It can be changed in the future when the abstraction is needed.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/OverviewRulerLane.java
entlicher: Done. I wrote a simple test that verifies that the switch handles all cases correctly.

PR #3050, opened 2021-07-14
LSP: Run all tests action fixed in Gradle projects.

dbalek:
Regression in "run all tests" action fixed.


JaroslavTulach (approved):
Integrated into 12.4.301 release as cd6bfbc

PR #3054, opened 2021-07-15
[NETBEANS-5832] Fixing compilation of static initializer for vanilla indexing.

jlahoda:


java/java.source.base/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorker.java
```diff
@@ -712,7 +712,7 @@ public Void visitClass(ClassTree node, Void p) {
                             csym.members_field.remove(member);
                         }
                     }
-                    if (errorClass || def.hasTag(JCTree.Tag.ERRONEOUS) || def.hasTag(JCTree.Tag.BLOCK)) {
```
eirikbakke, line 715: From the logs I see you wrote the original  JCTree.Tag.BLOCK condition here, so I'm sure the adjustment makes sense.

eirikbakke (approved):
Brilliant! Confirmed to fix both https://issues.apache.org/jira/browse/NETBEANS-5832 and https://issues.apache.org/jira/browse/NETBEANS-5757 with vanilla javac and CoS enabled. I tried without the patch first, confirmed the bug with the examples in the JIRA issues, and then applied the patch and saw the bug disappear. Thank you for this!

matthiasblaesing:
Looks sane, has positive review (with verification, that it fixes reported problems) and should have minimal impact. Lets get this in.

Dongocthach:
have you checked that?  print   doesn`t work.

eirikbakke:
@Dongocthach Can you be more specific? What did you try, on which version of NetBeans?

Dongocthach:
Sorry for my bad english skill.
I am not sure that,I have commented on right place or not.

The issue is at nebean 12.4." print "doesn't work.
file attached
[image: loinetbean.jpg]

Vào Th 3, 24 thg 8, 2021 vào lúc 23:03 Eirik Bakke <
***@***.***> đã viết:

PR #3055, opened 2021-07-15
[NETBEANS-5859] Upgraded Jacoco to 0.8.7 to support JDK 17 coverage

lkishalmi:


matthiasblaesing (approved):
Only eyeballed this, but looks sane to me. The failing unittests are the expected bunch, but I restartet the runs.

PR #3057, opened 2021-07-16
allow more dot separated digit for version scheme

ebarboni:
Quick fix for allowing version like 12.4.301 as for Apache NetBeans we were set with Float

test is on  nb/o.n.upgrader/test/unit/src/org/netbeans/upgrade/AutoUpgradeTest.java.

copied to Autopudate. 
Reorder modifier around

### DIFF
--- a/nb/autoupdate.pluginimporter/src/org/netbeans/modules/autoupdate/pluginimporter/Installer.java
+++ b/nb/autoupdate.pluginimporter/src/org/netbeans/modules/autoupdate/pluginimporter/Installer.java
@@ -19,6 +19,7 @@
 package org.netbeans.modules.autoupdate.pluginimporter;
 
 import java.io.File;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
Float, eek! :smile:  Couldn't we reuse something like [SpecifcationVersion](https://github.com/apache/netbeans/blob/master/platform/openide.modules/src/org/openide/modules/SpecificationVersion.java) for this?

On the other aspect, I just noticed `12.4.301` in the previous versions list while building from master.  Do we really want that in there at all?  Could it cause any issue as it is?  Maybe just filter the list by those that parse as float first?

ebarboni:
Using Specification Version looks better :D. This should allow to target previous version of NetBeans maybe for import settings. Not sure the better stragegy ignoring or generalizing.

neilcsmith-net (approved):
Looks good to me if it works.  Longer term we could probably simplify further by sorting all folders by name where lower than running version - rather than having to pass in versions to check in build.

PR #3024, opened 2021-06-29
Snapshot of APIs as of 12.4

ebarboni:
replace #2940

```
# checkout branch (or release tag)
git checkout release124
ant clean
ant build
ant gen-sigtests-release
git add -A
git stash
git checkout master
git checkout -b apis-nb124
git stash pop
git add -A
git commit -m "Snapshot of APIs as of 12.4"
# remove getPeer() calls
find . -name "*.sig" -exec sed -i '/java.awt.peer.ComponentPeer/{N;d;}' {} \;
# check git diff
git add -A
git commit --amend
```

Had hard time to put code markup. I used the 1.75 version for groovy. Maybe was incremented on master

PR #3059, opened 2021-07-16
Increment spec versions for 12.5

neilcsmith-net:
As far as I can tell, spec version increment is still pending for 12.5.  Was waiting on the delayed #3024 although the reasons for doing those in order may no longer be relevant.


JaroslavTulach (approved):

neilcsmith-net:
Thanks @JaroslavTulach

@ebarboni merging now so it's in by freeze, and before anyone decides they need to increment a spec version and cause a merge conflict. :smile:

PR #3061, opened 2021-07-18
Use j.u.Properties instead of o.o.u.EditableProperties to reduce CPU usage

matthiasblaesing:
It was observed, that with a big project, massive time was spend in 
org.openide.util.EditableProperties.addItem(). Given, that none of the
benefits of EditableProperties take hold:

- the properties are used as pure backing store, without the expectation
  of a user to modify them
- the order of the entries is irrelevant
- the file is not put into VCS (it is in the cache path)
- special line break handling is not relevant

the easiest work around is to drop usage of EditableProperties and
use plain Properties.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CustomIndexerImpl.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CustomIndexerImpl.java
@@ -21,7 +21,9 @@ package org.netbeans.modules.lsp.client.bindings;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Collection;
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@jlahoda could you please have a look at this? The project I observed the massive times in showed a property list with > 70.000 entries and the sampling profiler from visual vm led me directly to the addItem method of EditableProperties.

jlahoda:
@matthiasblaesing, sorry for trouble. Using `Properties` looks fine. It would also probably be enough to use `EditableProperties(false)` to avoid sorting (not sure why I used `true` anyway - probably no particular good reason).

jlahoda (approved):

matthiasblaesing:
@jlahoda thank you for the review.

PR #3063, opened 2021-07-18
[NETBEANS-5864] Sightly better support of GTK with a dark theme

lkishalmi:
This PR just makes NetBeans more aware of GTK used with a dark theme. The implementation is trivial, does not support GTK theme changes happening when NetBeans is running. It's a better than nothing approach.


matthiasblaesing (approved):
Code change looks sane and I gave it a spin with "Yaru-dark" - looked ok to me (maybe even better than the light variant).

For the future it might be worth pairing the dark theme with the "Flatlaf - Dark" editor theme.

PR #3064, opened 2021-07-19
Fix java.editor's Java formatting tests

jlahoda:
These probably never passed for the current formatting implementation. Formatting needs to be able to parse the input, but many of the samples are not really parseable. This patch attempts to fix that, and also tries to create the setup the Document so that the formatter has enough information to perform the parse.


ebarboni:
seems ok.

PR #3068, opened 2021-07-20
Force bytecode parsing. Cache resource misses.

sdedic:
Poor performance of the Groovy parser was the reason #/3042 allows to disable Groovy indexing. I've discovered some hot points in Groovy parsing; [attaching profiling data to this PR for reference](https://github.com/apache/netbeans/files/6849268/sampler-10ms.zip)

Groovy parser tries to call `getResource` on parsing ClassLoader, which in turn delegates to `URLClassLoader` that goes through all JARs included in the compilation. While URLClassLoader may be quite fast, adding a **negative** cache for misses would avoid querying 20+ JAR implementations for a resource that is already known to be missing.

Next the parser for whatever reason tries to **load a Class to JVM** which is both slow and **dangerous** as class initializers may be run. This feature can be disabled in parser's options.

In effect on my pet test project (micronaut `http-client`), indexing time for `main/test/groovy` went down from ~160 secs to ~20-30secs. No current groovy test seems to fail.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/compiler/CompilationUnit.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/compiler/CompilationUnit.java
@@ -39,6 +39,7 @@ import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.GenericsType;
 import org.codehaus.groovy.ast.MixinNode;
 import org.codehaus.groovy.control.ClassNodeResolver.LookupResult;
+import org.codehaus.groovy.control.ClassNodeResolver;
 import org.codehaus.groovy.control.CompilerConfiguration;
 import org.netbeans.api.annotations.common.NonNull;
 import org.netbeans.api.java.source.ClasspathInfo;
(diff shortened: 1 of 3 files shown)


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/compiler/CompilationUnit.java
```diff
@@ -75,6 +79,13 @@ public CompilationUnit(GroovyParser parser, CompilerConfiguration configuration,
                 },
                 security, this.configuration, cpInfo, classNodeCache);
     }
+    
+    private static class DebugResolver extends ClassNodeResolver {
```
JaroslavTulach, line 83: Looks like an unused class.

JaroslavTulach (approved):

ebarboni:
Looks ok minus one removal to be on 12.5 if you want

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/compiler/CompilationUnit.java
sdedic: Removed in d834942c32

dbalek (approved):

sdedic:
Squashed before merging.

sdedic:
I would be somewhat careful with including this change in 12.5: while the negative cache is (in theory harmless), the PR disables loading classes through classloaders that SHOULD be OK as well, but might surface in unexpected ways. Especially I wouldn't like to break editing in Gradle scripts (which also use Groovy parsing).

@jtulach, @dbalek - opinions ?

JaroslavTulach:
There is at least a month till release 12.5. Plenty of time to fix problems, if they are reported.

sdedic:
Good. Travis failures seem unrelated to the PR. Merging.

PR #3071, opened 2021-07-22
Proper JDK & location of the binary classes in mx projects

JaroslavTulach:
Using proper JDK based on `compliance` of each source root. Looks like all classes are in
```
ls graal/compiler/mxbuild/*/bin -d
```
e.g. the previous usage of subDir was incorrect.

### DIFF
--- a/java/java.mx.project/src/org/netbeans/modules/java/mx/project/Compliance.java
+++ b/java/java.mx.project/src/org/netbeans/modules/java/mx/project/Compliance.java
@@ -99,12 +99,23 @@ public final class Compliance {
         }
     }
 
+    boolean matches(Compliance check) {
+        if (min > check.min) {
+            return false;
+        }
+        if (max < check.max) {
+            return false;
+        }
+        return true;
+    }
+
     String getSourceLevel() {
         if (min == 8) {
             return "1.8";
         }
         return "" + min;
     }
+
     
     public interface Provider {
         public Compliance getCompliance();
--- a/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
+++ b/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
@@ -18,16 +18,11 @@
  */
 package org.netbeans.modules.java.mx.project;
 
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.List;
 import org.netbeans.api.java.classpath.ClassPath;
 import org.netbeans.api.java.classpath.GlobalPathRegistry;
 import org.netbeans.api.java.classpath.JavaClassPathConstants;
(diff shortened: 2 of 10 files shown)


ebarboni:
for me would be ok to merge.

ebarboni (approved):

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
```diff
@@ -76,7 +61,7 @@ public ClassPath findClassPath(FileObject file, String type) {
 
     @Override
     public void projectOpened() {
-        GlobalPathRegistry.getDefault().register(ClassPath.BOOT, new ClassPath[] {bootCP});
+        // GlobalPathRegistry.getDefault().register(ClassPath.BOOT, new ClassPath[] {bootCP});
```
tzezula, line 64: Why the BOOT cp is not registered and unregistered?

java/java.mx.project/src/org/netbeans/modules/java/mx/project/Compliance.java
```diff
@@ -99,12 +99,23 @@ public String toString() {
         }
     }
 
+    boolean matches(Compliance check) {
+        if (min > check.min) {
```
sdedic, line 103: `>` or `>=` ?

JaroslavTulach: `return false` if outside of range, but you are right, there should be a test: 1ce26972e9f0

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
JaroslavTulach: Now I would have to collect all bootpaths from all the used platforms.

tzezula: I think it can work. The boot cp should be indexed because it's discovered in the `RepositoryUpdater#findDependencies(...)`. The `PathRegistry#getRootPathIds()` will not know the boot cp roots but `PathRegistry#getRootPathIds()` is not important for binary roots.
Just check that the boot cp is indexed (you see the platform types from the correct platform in Go To Type dialog).

tzezula (approved):
Check that platform is indexed. I think it is but better to check.
Otherwise seems OK.

java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteClassPathProvider.java
JaroslavTulach: Done in 3c0dfd659abeabadbd5f33b511f5a5151cfe6512

PR #3077, opened 2021-07-28
[NETBEANS-2443] Fix-Can't start profiler windows

moacirrf:
https://issues.apache.org/jira/browse/NETBEANS-2443

Fix start of profiler on windows the bug was caused by wrong command  generated, so i compare both commands:
image

After the change 
image

I try don't mess thing so much, because i don't understand why so much lines are generated, but a tested on Linux and windows.
Run, Debug e Profiler.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
@@ -413,8 +413,6 @@ public class MavenCommandLineExecutor extends AbstractMavenExecutor {
         //#164234
         //if maven.bat file is in space containing path, we need to quote with simple quotes.
         String quote = "\"";
-        // the command line parameters with space in them need to be quoted and escaped to arrive
-        // correctly to the java runtime on windows
         for (Map.Entry<? extends String,? extends String> entry : config.getProperties().entrySet()) {
             String k = entry.getKey();
             // filter out env vars AND internal properties.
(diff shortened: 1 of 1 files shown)


java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
```diff
@@ -424,9 +421,8 @@ public void run() {
             }
             //skip envs, these get filled in later.
             //#228901 since u21 we need to use cmd /c to execute on windows, quotes get escaped and when there is space in value, the value gets wrapped in quotes.
-            String value = (Utilities.isWindows() ? entry.getValue().replace(quote, escaped) : entry.getValue().replace(quote, "'"));
+            String value = (Utilities.isWindows() ? entry.getValue().replace(quote, "") : entry.getValue().replace(quote, "'"));
             if (Utilities.isWindows() && value.endsWith("\"")) {
-                //#201132 property cannot end with 2 double quotes, add a space to the end after our quote to prevent the state
```
ebarboni, line 429: may no need to remove this comment

sdedic:
The reasons are more deeper: I fear this patch will fix the profiler, but will break other code that may use e.g. system properties that contains spaces or other 'illegal' chars. As noted in [NETBEANS-5732](https://issues.apache.org/jira/browse/NETBEANS-5732), the culprit is the profiler integration which quotes the output of StartupExtender, when it should not.

java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
moacirrf: humm..I will put again this comment

moacirrf:
> The reasons are more deeper: I fear this patch will fix the profiler, but will break other code that may use e.g. system properties that contains spaces or other 'illegal' chars. As noted in [NETBEANS-5732](https://issues.apache.org/jira/browse/NETBEANS-5732), the culprit is the profiler integration which quotes the output of StartupExtender, when it should not.

Yes, is possible...i will test with netbeans on path with spaces.
Thanks

PR #3078, opened 2021-07-28
[NETBEANS-5841] prevent NPE in LSP client using a Python Language Server

MKuettner:
This prevents a NPE when using Python Language Server: https://github.com/palantir/python-language-server
Otherwise frequently a NPE will be thrown.


ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/MarkOccurrences.java
```diff
@@ -101,6 +101,9 @@ private OffsetsBag computeHighlights(Document doc, int caretPos) {
         try {
             List<? extends DocumentHighlight> highlights =
                     server.getTextDocumentService().documentHighlight(new DocumentHighlightParams(new TextDocumentIdentifier(uri), Utils.createPosition(doc, caretPos))).get();
+            if(highlights == null) {
```
junichi11, line 104: Please format.

junichi11: Ah, you should create a branch when you create a PR. (e.g. `netbeans-5841`) See also: https://netbeans.apache.org/participate/submit-pr.html

jlahoda (approved):
Looks good to me. Adding the space between `if` and `(` would be nice, but not absolutely necessary IMO.

MKuettner:
Is there any more action needed from my side?

junichi11:
@MKuettner 
I don't think we need to format existing code. However, I think that we should format new code as much as possible. I don't think it is good that we add unnecessary commits. 
So, could you please fix them?

junichi11:
@MKuettner Well, As I wrote above, I don't think you need to format existing code. Because it is hard for reviewers to read diff. (It would be better to format only your changes.) I think you should create a branch then re-create a new PR.

CC: @ebarboni

MKuettner:
@junichi11 I just applied the original Netbeans formatter to the whole file, since there were a lot of these formatting "issues". From my point of view this file is small enougth for reviewers to be able to read the diff. 
And if there are formatting guidelines so why not apply them to existing code?

But anyway...

To clarify next steps for me:
- i will close this PR
- create a new branch (e.g. called "netbeans-5841")
- make a single commit including the fix (using correct format)
- create a new PR
?

junichi11:
@MKuettner Yes, in this case, I think revewers can read it. However, I suggested it because some people don't like that if it has huge changes...

Yes, your steps look good to me :) Please see also above URL. Thank you for your time and your work.

ebarboni:
just reruning the travis failing part. Will squash and merge. I guess overformatting is ok because classes is small.

junichi11:
@ebarboni Leave it to you. Thank you!

PR #3079, opened 2021-07-28
Filtering previous release 

ebarboni:


ebarboni:
https://github.com/apache/netbeans-jenkins-lib/pull/31 related

ebarboni:
should be ported to former release branches

neilcsmith-net:
:+1: Thanks, was just looking at similar.  Bit annoying if it's going to require porting to every release branch, though.  On the other hand, there's not much else we can do to filter out the unwanted release docs I guess?  May be useful should we want to stop publishing some docs after a time?  Maybe should drop the list of all releases in top bar?  Is that the only thing that requires old releases to need updating for changes?

ebarboni:
backporting is not fun but I did it a few time. I perso like the up bar because it tell students that their are not up to date and allow to navigate the api diff.
But we may have other issues like jquery to be changed, or browser deprecating old html. 

I'm not sure we can stop publishing doc but, we should ensure that their is no NetBeans IDE connection to thoose page otherwise lots of 404

neilcsmith-net:
Good points.  Keep it simple, at least for now.

Longer term, maybe we bring the template from elsewhere (like release info) - number of branches to backport to is only going to increase! :smile:

neilcsmith-net (approved):

PR #3080, opened 2021-07-28
Attach to a running Native Image.

entlicher:
We have a launch of native image working. But attach to a running process of native image was missing.
This PR introduced the attach to both NetBeans and LSP/VSCode.
A list of running processes is provided, sorted from the newest to the oldest. In NetBeans it's added to the attach dialog, in VSCode we're introducing `nativeImage.attachDebugger.pickProcess` command.

### DIFF
--- a/cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
+++ b/cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
@@ -22,6 +22,8 @@ package org.netbeans.modules.cpplite.debugger;
 import java.io.File;
 import java.util.List;
 
+import org.netbeans.api.annotations.common.NullAllowed;
+
 /**
  *
  * @author lahvac
@@ -30,11 +32,14 @@ public final class CPPLiteDebuggerConfig {
 
     private final List<String> executable;
     private final File directory;
+    @NullAllowed
+    private final Long processId;
     private final String debugger;
 
-    public CPPLiteDebuggerConfig(List<String> executable, File directory, String debugger) {
+    public CPPLiteDebuggerConfig(List<String> executable, File directory, @NullAllowed Long processId, String debugger) {
         this.executable = executable;
         this.directory = directory;
+        this.processId = processId;
         this.debugger = debugger;
     }
 
@@ -42,12 +47,39 @@ public final class CPPLiteDebuggerConfig {
         return (!executable.isEmpty()) ? executable.get(0) : "<empty>";
     }
 
+    /**
+     * Get the executable with arguments.
+     */
     public List<String> getExecutable() {
         return executable;
     }
 
+    /**
+     * Get the directory in which the executable command is to be launched.
+     */
     public File getDirectory() {
         return directory;
+    }
+
+    /**
+     * Check if debugger should attach to a running process. Get the process ID
+     * from {@link #getAttachProcessId()}.
+     */
+    public boolean isAttach() {
+        return processId != null;
+    }
+
+    /**
+     * Get the process ID to attach to.
+     *
+     * @return the process ID
+     * @throws IllegalStateException when {@link #isAttach()} is false.
+     */
+    public long getAttachProcessId() {
+        if (processId == null) {
+            throw new IllegalStateException("No process to attach to.");
+        }
+        return processId;
     }
 
     public String getDebugger() {
--- a/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
+++ b/ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
@@ -114,6 +114,20 @@ public final class NIDebugger {
     }
 
     /**
+     * Attach to a process and create a debugging session. Call this typically after breakpoints are added.
+     *
+     * @param executablePath path to an executable representing the native image
+     * @param processId a process to attach to
+     * @param debugger the native debugger command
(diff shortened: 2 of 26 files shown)


cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
```diff
@@ -21,6 +21,7 @@
 
 import java.io.File;
 import java.util.List;
+import java.util.Optional;
```
JaroslavTulach, line 24: `Optional`? I don't like that class much.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
```diff
@@ -50,6 +53,14 @@ public File getDirectory() {
         return directory;
     }
 
+    public boolean isAttach() {
+        return processId.isPresent();
+    }
+
+    public long getProcessId() {
+        return processId.get();
```
JaroslavTulach, line 61: What happens when the `Optional` isn't set? A `NullPointerException`?

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/api/ProcessInfo.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Information about a running process.
+ *
+ * @since 0.3
+ */
+public final class ProcessInfo {
```
JaroslavTulach, line 30: This might better belong into `extexecution` API. It is a generic functionality used to talk to executed processes.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
entlicher: I can change it to `Long` with `null` allowed, not a problem.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
entlicher: There is `isAttach()` to test it. If it's false and you still ask for the process ID, it fails.
I can replace it with `Long getProcessId()`, which may return `null`. It'd be more straightforward.

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/api/ProcessInfo.java
entlicher: Well, thinking about this more, I'd remove it from the API and duplicate the classes instead. This will get obsolete when we abandon JDK 8. There is a `ProcessHandle` API since JDK 9.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
entlicher: I've removed the use of `Optional`.

cpplite/cpplite.debugger/src/org/netbeans/modules/cpplite/debugger/CPPLiteDebuggerConfig.java
entlicher: It's documented and checked explicitly.

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/api/ProcessInfo.java
entlicher: The class is duplicated instead of creating an API.
Duplication is documented in both classes and a note about future replacement with `ProcessHandle` is added.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
```diff
@@ -113,6 +113,20 @@ public void removeBreakpoint(Object id) {
...
+     * Attach to a process and create a debugging session. Call this typically after breakpoints are added.
+     *
+     * @param executablePath path to an executable representing the native image
+     * @param processId a process to attach to
+     * @param debugger the native debugger command
+     * @param startedEngine the corresponding DebuggerEngine is passed to this consumer
+     * @return future that completes on the execution finish
+     * @since 0.4
+     */
+    public CompletableFuture<Void> attach(String executablePath, long processId, String debugger, Consumer<DebuggerEngine> startedEngine) {
```
JaroslavTulach, line 126: Wouldn't it be better to return `CompletionStage<DebuggerEngine>` and remove the 4th parameter? Probably not, as the `Consumer<DebuggerEngine>` is called way sooner than the `CompletableFuture<Void>` is set, rigth?

JaroslavTulach (approved):

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
entlicher: Right. The consumer may be called even before the `attach` method returns, depending on the actual threads scheduling.

PR #3084, opened 2021-08-02
New Project wizard for Gradle should use gradle init

JaroslavTulach:
The New Java Project wizard uses an unusual Gradle template with an older deprecated plugins syntax and jacoco included for some reason.

To make this feature maintainable we should instead use the existing `gradle init` functionality:
```
gradle init --type java-application --test-framework junit-jupiter --dsl groovy --package com.example --project-name example
```
My take on that. Let's extend the API of `TemplateOperation` to be able to invoke `gradle init`. Then use it in Java application and Java library project. When at it also change the default to create Gradle wrapper when no option is specified. 

Done dd9ca25  - Document and version API changes
Done da8f615 - Disable many fields in the wizard panel?
Resolving version update conflict in 634593b

How do you feel about this @lkishalmi?


lkishalmi:
As of my feelings: go for it!
As of jacoco. That was just a one liner, to enable code-coverage on projects out of the box.

PR #3085, opened 2021-08-02
[Netbeans-5734] Quick fix to make NB 12.4+ run Glassfish 6.1 on JDK 11+

matthiasblaesing:
This PR is a cleaned up and enhanced version of #3056. It was also rebased onto the release125 branch.

This is primary a bug fix, so should be considered for 12.5.

Closes: #3056

### DIFF
--- a/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/StartTask.java
+++ b/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/StartTask.java
@@ -62,6 +62,7 @@ import org.netbeans.modules.glassfish.spi.GlassfishModule.ServerState;
 import org.openide.execution.NbProcessDescriptor;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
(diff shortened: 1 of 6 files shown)


matthiasblaesing:
@juneau001 

For a discussion please see the referenced original PR. This is the consolidated version, I'd appreciate opinions as I currently don't use glassfish myself and so working from past experiences.

neilcsmith-net:
👍 Should be based on delivery branch though (which I guess won't make a difference yet).

matthiasblaesing:
@neilcsmith-net thanks for the check - something felt wrong, but I could not pin-point it. At this point it is indeed uncritical as the base commit is the same for all candidates, so I switched target branch in the PR settings.

juneau001 (commented):
Loaded PR into local development environment.  Compiled and built without issues.  Configured GlassFish 6.1 with JDK 11 and successfully deployed Jakarta EE application using JDK 11.

ebarboni:
green light I will merge, and resync (I hope to success in this step) to make a "beta" with all 12.5 intended content :D

PR #3089, opened 2021-08-05
[NETBEANS-5533] Remove the URL of the old wiki

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-5533

- The old wiki is available no longer (http://wiki.netbeans.org/HowToConfigureXDebug)


tmysik (approved):
@junichi11 Was the information in that page useful even today? I mean, should we create a similar page?

tmysik:
@ebarboni @junichi11 

The patch is OK, feel free to merge. Thanks.

junichi11:
@tmysik Thank you for your review :)

> Was the information in that page useful even today? I mean, should we create a similar page?

Maybe, the information was so old... So, I added the following URL before: https://xdebug.org/docs

https://github.com/apache/netbeans/blob/6456f473f48f5d6daea9bc247198bccaa9017aa9/php/php.dbgp/src/org/netbeans/modules/php/dbgp/ConnectionErrMessage.java#L242

Instead of the old wiki, I suppose the following page is helpful for users: https://xdebug.org/docs/install
Thanks.

ebarboni:
ok mergin

PR #3094, opened 2021-08-06
Stop PHP from invoking autocompletion for all keystrokes inside HTML

neilcsmith-net:
Stop PHP from invoking autocompletion for all keystrokes inside embedded HTML.

Changing from QueryType.ALL_COMPLETION to QueryType.NONE for embedded HTML tokens stops the PHP code completion overriding the HTML one.

This appears to fix at least some of the issues reported in https://lists.apache.org/thread.html/r6324903231edcb4ea322fc5060fe7a0620d1be6a413d33556df70468%40%3Cusers.netbeans.apache.org%3E

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletion.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/completion/PHPCodeCompletion.java
@@ -2166,7 +2166,7 @@ public class PHPCodeCompletion implements CodeCompletionHandler2 {
             Token t = ts.token();
             if (t != null) {
                 if (t.id() == PHPTokenId.T_INLINE_HTML) {
-                    return QueryType.ALL_COMPLETION;
+                    return QueryType.NONE;
                 } else {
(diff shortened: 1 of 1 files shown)


geertjanw (approved):

junichi11 (approved):
Looks good to me. Thank you!

Related to https://bz.apache.org/netbeans/show_bug.cgi?id=233890

junichi11:
@tmysik Could you please have a look at this?
@ebarboni Could you please wait for Tomas?

tmysik (approved):
I am not sure (I am not editor guy) but I believe that the original motivation was to have the PHP code completion in HTML. Those days, mixing these 2 languages was very common (this is not true anymore, which is a good thing).

But, as I wrote, I can be wrong :)

neilcsmith-net:
Thanks @junichi11 @tmysik 

This change should keep the auto completion popup showing according to the HTML settings as far as I can see. NONE is neutral. From that bugzilla link it looks like it was changed to stop it force closing the popup, but by triggering force open instead? Those enum values could be better named and/or documented!

PR #3096, opened 2021-08-08
Exposing "Frontend UI Pages" to users of Java Frontend Application

jtulach:
When a Gradle project for Java Frontend Application is created (like in #3095), it doesn't navigate user to `pages` directory with `index.html` - finding that directory is quite tricky and hurts experience of those who try to use this support. The PR fixes that by creating a dedicated node, if the `src/main/webapp/pages` directory is found.

obrazek

Laszlo, can you please check if I am using the Gradle project extension mechanism properly? It does work, but maybe there are better ways to achieve the same.

Sváťo, maybe the plugin could register actions and associate Run with `bck2brwsrShow -t` without the need to specify that in the `gradle.properties`? However I haven't found an example to do it - `micronaut` project doesn't seem to specify that, but I think @sdedic has created an API that should work for Gradle projects as well.

If the current PR is fine from @lkishalmi's perspective, I'd like to get it to 12.5 - the node is very handy. If it is not, we can wait for 12.6.

### DIFF
--- a/java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaPagesNodeFactory.java
+++ b/java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaPagesNodeFactory.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.modules.gradle.htmlui;
+
+import java.awt.Image;
+import org.netbeans.modules.gradle.api.NbGradleProject;
+import org.netbeans.modules.gradle.spi.nodes.AbstractGradleNodeList;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.Collections;
+import java.util.List;
+import javax.swing.event.ChangeListener;
+import org.netbeans.api.project.Project;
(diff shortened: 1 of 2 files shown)


java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaPagesNodeFactory.java
```diff
@@ -0,0 +1,110 @@
...
+                public String getDisplayName() {
+                    return Bundle.MSG_HtmlJavaPages();
+                }
+            };
+            return fn;
+        }
+
+        @Override
+        public void propertyChange(PropertyChangeEvent evt) {
+            if (NbGradleProject.PROP_PROJECT_INFO.equals(evt.getPropertyName())) {
```
lkishalmi, line 94: This would invoke change if the Gradle project files would be changed. Maybe the intent was to refresh on PROP_RESOURCES changes. Though in order that one to work you need to register a WatchedResourceProvider for "src/main/webapp/pages". In that case evt.getNewValue() will hold the URI of the changed resource.

lkishalmi (approved):
Well, it can work. I hope not too many web applications are out there with pages directory in it. In that case the UI might get a bit confusing. I provided some hint, otherwise I'm good with this change.

java/gradle.htmlui/src/org/netbeans/modules/gradle/htmlui/HtmlJavaPagesNodeFactory.java
JaroslavTulach: I admit I had no intent, I just copied [WebPagesNodeFactory](https://github.com/apache/netbeans/blob/master/enterprise/gradle.javaee/src/org/netbeans/modules/gradle/javaee/nodes/WebPagesNodeFactory.java#L72). Dump, complete (e.g. not even **selective**) [cluelessness](http://wiki.apidesign.org/wiki/Cluelessness)...

lkishalmi: I guess  you are not the only one [doing that](https://github.com/apache/netbeans/blob/0719659fdb6d42a5d2314dfd0436a50c6ef81c77/enterprise/maven.j2ee/src/org/netbeans/modules/maven/j2ee/ui/nodes/WebPagesNodeFactory.java#L108)...

sdedic:
Re. `-t`: no, Micronaut does not do that, as `-t` is supported by Gradle itself (there's a exec profile that adds `--continuous` by default). You can however register a declarative action provider ([see example action provider registration](https://github.com/apache/netbeans/blob/master/extide/gradle/test/unit/src/META-INF/generated-layer.xml#L24)) and can override the default action if a specific plugin is present in Gradle build.

JaroslavTulach:
Eric, can this change be merged into `delivery` for 12.5?

jen801 (approved):

PR #3098, opened 2021-08-09
[NETBEANS-2443] [NETBEANS-5732] Support profiler or other StartupExtenders that use spaces in args

sdedic:
An alternative to #3077 - an attempt to fix flaky`StartupExtender` specification. The main issue is that while `StartupExtender` API provides `List` of arguments, the API did not make it explicit that no quoting should be used - that the returned list items contain the verbatim conents that should be passed to the VM.  In general quoting should be done **always by the launcher**: it can put individual arguments right into the `ProcessBuilder` or somehow quote/concatenate into a commandline string.

I've added a flag that allows a `StartupExtender` to declare it does/does not quoting. `getRawArguments` (aded API method) will then contain literal parameter value regardless of the extender policy. For compatibity, the `getArguments` method is unchanged.

I have adapted maven and gradle. Maven has more tweaks as I found multiple issues with space or quote handling in Maven / Windows.

This should fix [NETBEANS-2443](https://issues.apache.org/jira/browse/NETBEANS-2443) and [NETBEANS-5732](https://issues.apache.org/jira/browse/NETBEANS-5732)

Unit tests added.


sdedic:
Eh ... messed this up with delivery / master target. Will create a new PR.

PR #3101, opened 2021-08-10
Fixes wrong detection message when selecting a GF6.2 location.

nicolaken:
When adding a new server, after selecting "Glassfish Server" and pressing next, when selecting an install location, the GUI searches for a valid instance by comparing all possible instances (versions) with the one found in the location specified.

```
org.netbeans.modules.glassfish.common.wizardsServerWizardIterator::isValidInstall
org.netbeans.modules.glassfish.common.ServerDetails::isInstalledInDirectory
```

If the Glassfish version is 6.2, it shows it has found 6.1, but then goes on with the 6 version, as the 6.1 instance says it's 6.

`org.netbeans.modules.glassfish.common.ServerDetails::getVersion`

Now 6.1 says it's 610 and 6.0 says it's 600 as other versions do.

Fixes small bug in Jakarta EE 9 GlassFish 6 Support (#2902).

### DIFF
--- a/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/ServerDetails.java
+++ b/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/ServerDetails.java
@@ -209,7 +209,7 @@ public enum ServerDetails {
      */
     GLASSFISH_SERVER_6(NbBundle.getMessage(ServerDetails.class, "STR_6_SERVER_NAME", new Object[]{}), // NOI18N
         "deployer:gfv6ee9", // NOI18N
-        6,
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
Base branch for this should be `delivery` not `release125` - can be edited above.

matthiasblaesing (approved):
Looks sane to me. @juneau001 could you please also have a look?

juneau001:
Built and tested successfully.  Thanks for the PR!

matthiasblaesing:
@juneau001 thanks for the check
@nicolaken thank you for the fix

Lets get this into delivery.

PR #3102, opened 2021-08-10
[NETBEANS-5910] Fix illegible heap view text with dark LAFs

neilcsmith-net:
This fixes an issue with the heap view text being ugly and illegible on dark LAFs such as FlatLaf Dark. Not a major issue, but really stands out in the UI.

Unfortunately, #2965 added a hardcoded, non-optional highlight colour to the text. I considered making the value truly optional (ie. don't render the highlight at all if no key), but deriving the colour from the background colour should work OK.

### Before

NETBEANS-5910-before

### After

NETBEANS-5910-after

### DIFF
--- a/platform/openide.actions/src/org/openide/actions/HeapView.java
+++ b/platform/openide.actions/src/org/openide/actions/HeapView.java
@@ -54,7 +54,7 @@ import org.openide.util.NbPreferences;
  * <li> nb.heapview.background - Color of widget background
  * <li> nb.heapview.foreground - Color of text
  * <li> nb.heapview.chart - Color of area chart
- * <li> nb.heapview.background - Color of outline around the text, to provide a contrast against
+ * <li> nb.heapview.highlight - Color of outline around the text, to provide a contrast against
  *                               the chart (may have a non-opaque alpha value)
  * </ul>
  * @author sky, radim, peter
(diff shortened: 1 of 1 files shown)


platform/openide.actions/src/org/openide/actions/HeapView.java
```diff
@@ -54,7 +54,7 @@
  * <li> nb.heapview.background - Color of widget background
  * <li> nb.heapview.foreground - Color of text
  * <li> nb.heapview.chart - Color of area chart
- * <li> nb.heapview.background - Color of outline around the text, to provide a contrast against
+ * <li> nb.heapview.highlight - Color of outline around the text, to provide a contrast against
```
eirikbakke, line 57: Yup, that was a typo.

platform/openide.actions/src/org/openide/actions/HeapView.java
```diff
@@ -112,17 +112,20 @@
...
-
         c = UIManager.getColor("nb.heapview.background"); //NOI18N
         if (null == c) {
             c = new Color(0xCEDBE6);
         }
         BACKGROUND_COLOR = c;
+
+        c = UIManager.getColor("nb.heapview.highlight"); //NOI18N
+        if (null == c) {
+            c = new Color(BACKGROUND_COLOR.getRed(),
```
eirikbakke, line 123: Much better. Not sure what I was thinking with the white-based default...

eirikbakke (approved):

KacerCZ:
Thanks for the fix.
I tested it with FlatLaF Light and Dark, Metal, Dark Metal, GTK + and Dark Nimbus and it works.

DevCharly (approved):

PR #3103, opened 2021-08-11
Fixing TreeUtilities.pathFor in the presence of the synthetic 'value=' in annotations.

jlahoda:
The synthetic assignment as ending position `-1`, which breaks `pathFor`. It is not quite clear if the `-1` is reasonable, but we can workaround for NetBeans, and think of the correct end position separately.


matthiasblaesing:
Manually run with JDK 8 the unittests that fail on travis (java.hints) come back clean. I suggest to get this in now.

arvindaprameya:
Approved, thank you Jan for doing this.

PR #3105, opened 2021-08-12
NETBEANS-5906: treat custom actions as enabled

sdedic:
See [NETBEANS-5906](https://issues.apache.org/jira/browse/NETBEANS-5906). The main issue was that **enabled** is defined by **GradleActionProviders**, but the Navigator creates its own custom action (that do not belong to any proivider, naturally). 
Similarly user-defined custom actions are not managed by any provider. So this omission is likely to break any user-defined custom actions, too: Since they are reported as disabled, they won't be processed.

During implementation I've discovered [NETBEANS-5918](https://issues.apache.org/jira/browse/NETBEANS-5918): funny issue caused by a weak listener that is not referenced from anywhere and is GCed. So the cache does not react properly when a **gradle.properties** or a action mapping XML files are created in the project dir.

Finally [NETBEANS-5919](https://issues.apache.org/jira/browse/NETBEANS-5919) - it's an inconsistency, since Maven project support reports custom actions as supported...

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
@@ -121,11 +121,7 @@ public class ActionProviderImpl implements ActionProvider {
 
     @Override
     public String[] getSupportedActions() {
-        List<? extends GradleActionsProvider> providers = ActionToTaskUtils.actionProviders(project);
-        Set<String> actions = new HashSet<>();
-        for (GradleActionsProvider provider : providers) {
-            actions.addAll(provider.getSupportedActions());
-        }
+        Set<String> actions = new HashSet<>(ActionToTaskUtils.getAllSupportedActions(project));
         // add a fixed 'prime build' action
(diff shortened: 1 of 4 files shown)


neilcsmith-net:
Thanks for looking into root cause of this.  Base branch should be `delivery`?!

sdedic:
@neilcsmith-net Thanks. Rebased.

neilcsmith-net (approved):
Checked out and built changes, and looks good for fixing NETBEANS-5906. And custom actions work fine - haven't checked what was happening before though.

As an aside, the custom actions UI seems clunkier than the Maven one - hard to easily add an action?

Thanks!

chrismaster:
With PR#3105 run task from Navigator executes the task.
From Projects a right click on the project name with the Gradle icon [Run Gradle] -> [Tasks] has no action.

lkishalmi:
Actually I'm working on the same issue. It seems the truth is somewhere in a combined solution, so I merge this one now and rebase my work.

lkishalmi (approved):
Fair enough.

PR #3106, opened 2021-08-14
Java code completion should show effectivelly final variables for use from anonymous classes.

jlahoda:
With vanilla javac, effectivelly final variables are not shown in the (Java) code completion in anonymous classes. This is because the variables from the Scope do not have the effectivelly final flag set. The proposal is to keep the flag when creating Scope.


arvindaprameya:
Approved, thank you for doing this.

dbalek (approved):

PR #3108, opened 2021-08-17
[NETBEANS-5740] Not taking write lock inside the parsing lock.

jlahoda:


neilcsmith-net:
Thanks @jlahoda for looking at this.  I didn't know whether best to push the call to `TestMethodController.setTestMethods` on to the EDT or do this.

Rebased on delivery, requested review from @ebarboni and tagged with 12.5.

PR #3110, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
Does travis_retry really solve the build issues? According to the data in paper [An empirical study of the long duration of continuous integration builds](https://dl.acm.org/doi/10.1007/s10664-019-09695-9), travis_retry can only solve 3% of the build failures. And it may cause unstable build and increase build time.

According to [Build times out because no output was received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received), we should carefully use travis_wait, as it may make the build unstable and extend the build time.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,7 +108,7 @@ matrix:
             - ant $OPTS clean
             - ant $OPTS build
           script:
-            - travis_retry hide-logs.sh ant $OPTS -f platform/api.htmlui test
+            - hide-logs.sh ant $OPTS -f platform/api.htmlui test
             - hide-logs.sh ant $OPTS -f platform/api.intent test
             - hide-logs.sh ant $OPTS -f platform/api.io test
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@YunLemon the changes were added because there were problems. Do you have prove, that your changes don't make the situation worse?

YunLemon:
Can the root causes of these problems be solved? 
`travis_retry` and  `travis_wait` means the result of the command is uncertain. However, the build process has to be deterministic. Flaky behavior, e.g., tests that sometimes fail, should be avoided, because this causes maintenance issues. Addressing this issue through retries after failures might not only hide the underlying problem, but makes issues also harder to debug when they only occur sometimes.
And a lot of time will be wasted if the command fails multiple times.

PR #3113, opened 2021-08-18
[NETBEANS-2360] Auto-detect HiDPI and anti-aliasing settings on Linux

eirikbakke:
This commit modifies the NetBeans IDE launcher scripts to automatically handle two display-related conditions on Linux:
1) Automatically enable the -Dawt.useSystemAAFontSettings system parameter if necessary to enable anti-aliased text. Veteran NetBeans users have known about this setting for a long time, but after a lot of research, I believe I have found the specific condition in which it is necessary, namely on the K desktop environment.
2) Automatically detect cases where the user has enabled 2x HiDPI scaling, via the Xft.dpi Xresource, and set the GDK_SCALE environment variable accordingly so that OpenJDK will handle this accordingly. This seems to work on both KDE and Gnome. HiDPI scaling works best on FlatLAF, but seemed to work tolerably on the default GTK LAF as well. (Perhaps in the future we should consider making FlatLAF the default LAF on Linux.)

See the in-code comments for more details. (There are two copies of the new code: one in the IDE launcher script, and one in the launcher script for NetBeans Platform applications.)

It would be great if multiple Linux people could test NetBeans with the modified launcher scripts, on both Gnome and KDE (and any other desktop environments that exist out there).


eirikbakke:
(Tried to add @timboudreau as a suggested reviewer here, too, but his name does not show up in the dropdown.)

eirikbakke:
I also considered using the sun.java2d.uiScale system property instead of the GDK_SCALE environment variable. But looking at the JDK source code, neither of these two support fractional scalings, e.g. 1.5x. The sun.java2d.uiScale value is always rounded to an int, even though it's initially parsed as a double, and it appears as an undocumented debugging property only. Experiments confirm that fractional scalings are not supported on Linux with either property. So keep using GDK_SCALE, and handle the 2x case only.

eirikbakke:
I added another condition to the script to try to handle the DPI scaling case described by Michal Hlavac in https://issues.apache.org/jira/browse/NETBEANS-2360 , on openSUSE Tumbleweed.

DevCharly (approved):
Looks good to me.

Have not tried to run the modified scripts (assuming they work 😉 ), but executed single commands (`xrdb -query` and `xdpyinfo`) on various Linux distros (Ubuntu, Kubuntu and Fedora) and various scaling factors and checked the output.

BTW since Java 17, it does no longer try to detect any scaling factor on Linux:
https://github.com/openjdk/jdk/blob/master/src/java.desktop/unix/native/common/awt/systemscale/systemScale.c
A lot of code was removed for https://bugs.openjdk.java.net/browse/JDK-8239894

Regarding FlatLaf: in case that fractional scaling is used (which is supported by KDE), or "Large Text" on Gnome, FlatLaf detects this and uses a larger font and scales its components. This scaling works only if FlatLaf method `UIScale.scale()` is used, which is not done in NB (except in FlatLaf module).

So it is definitely a good idea to set `GDK_SCALE` for 200% scaling.

eirikbakke:
@DevCharly Thanks! The patch won't handle every case, but I think it should be an improvement that won't cause new problems. So I'll merge it if no one else drops by and objects within a day or two.

PR #3116, opened 2021-08-18
[NETBEANS-5931] Consolidate duplicated logic to set rendering hints

eirikbakke:
There is a lot of duplicated logic around the NetBeans codebase relating to the setting of rendering hints on Graphics2D objects. This PR is an attempt to consolidate this logic into a single utility method (org.openide.awt.GraphicsUtils.configureRenderingHints(Graphics) in the org.openide.awt module).

This PR touches a lot of files throughout the codebase, and so should be tested in a development IDE for a while before reaching a release. I will start using this patch in my working IDE. Meanwhile I'm marking this PR as "work-in-progress".

The PR for this issue will also fix a minor issue where non-editable text values in the property sheet ended up not being anti-aliased. See attached screenshot. With the new utility method, this becomes a one-line fix.

NoAntialiasing

I put the new utility method in the org.openide.awt module, though alternatively it could be put in org.openide.util.ui . Here are the dependencies between these modules:

dependencies

### DIFF
--- a/ide/csl.api/src/org/netbeans/modules/csl/editor/completion/MethodParamsTipPaintComponent.java
+++ b/ide/csl.api/src/org/netbeans/modules/csl/editor/completion/MethodParamsTipPaintComponent.java
@@ -21,9 +21,9 @@ package org.netbeans.modules.csl.editor.completion;
 
 import java.awt.*;
 import java.util.List;
-import java.util.Map;
 import javax.swing.*;
 import javax.swing.text.JTextComponent;
+import org.openide.awt.GraphicsUtils;
 
 /**
  *
@@ -51,20 +51,8 @@ public class MethodParamsTipPaintComponent extends JToolTip {
     }
     
     public @Override void paint(Graphics g) {
-        Object value = (Map) (Toolkit.getDefaultToolkit().getDesktopProperty("awt.font.desktophints")); //NOI18N
-        Map renderingHints = (value instanceof Map) ? (java.util.Map) value : null;
-        if (renderingHints != null && g instanceof Graphics2D) {
-            Graphics2D g2d = (Graphics2D) g;
-            RenderingHints oldHints = g2d.getRenderingHints();
-            g2d.setRenderingHints(renderingHints);
-            try {
-                super.paint(g2d);
-            } finally {
-                g2d.setRenderingHints(oldHints);
-            }
-        } else {
-            super.paint(g);
-        }
+        GraphicsUtils.configureDefaultRenderingHints(g);
+        super.paint(g);
     }
 
     public @Override void paintComponent(Graphics g) {
--- a/platform/api.visual/src/org/netbeans/api/visual/widget/SceneComponent.java
+++ b/platform/api.visual/src/org/netbeans/api/visual/widget/SceneComponent.java
@@ -28,8 +28,8 @@ import java.awt.dnd.*;
 import java.awt.event.*;
 import java.awt.geom.AffineTransform;
 import java.util.List;
-import java.util.Map;
 import org.netbeans.modules.visual.laf.DefaultLookFeel;
+import org.openide.awt.GraphicsUtils;
 import org.openide.util.NbBundle;
 
 /**
@@ -101,12 +101,7 @@ final class SceneComponent extends JComponent implements Accessible, MouseListen
 //        System.out.println ("CLIP: " + g.getClipBounds ());
 //        long s = System.currentTimeMillis ();
         Graphics2D gr = (Graphics2D) g;
-
-        Object props = Toolkit.getDefaultToolkit ().getDesktopProperty ("awt.font.desktophints"); // NOI18N
-        if (props instanceof Map)
-            gr.addRenderingHints ((Map) props);
-        gr.setRenderingHint (RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
-        gr.setRenderingHint (RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+        GraphicsUtils.configureDefaultRenderingHints(gr);
         scene.setGraphics (gr);
 
         AffineTransform previousTransform = gr.getTransform ();
--- a/platform/o.n.core/src/org/netbeans/beaninfo/editors/FontEditor.java
+++ b/platform/o.n.core/src/org/netbeans/beaninfo/editors/FontEditor.java
@@ -30,6 +30,7 @@ import org.netbeans.core.UIExceptions;
 import org.openide.DialogDisplayer;
 
 import org.openide.NotifyDescriptor;
+import org.openide.awt.GraphicsUtils;
 import org.openide.awt.Mnemonics;
 import org.openide.explorer.propertysheet.editors.XMLPropertyEditor;
 import org.openide.util.Exceptions;
@@ -42,11 +43,6 @@ import org.openide.util.Utilities;
 * @author Ian Formanek
 */
 public class FontEditor implements PropertyEditor, XMLPropertyEditor {
-
-    static final boolean antialias = Boolean.getBoolean("nb.cellrenderer.antialiasing") // NOI18N
-         ||Boolean.getBoolean("swing.aatext") // NOI18N
-         ||(isGTK() && gtkShouldAntialias()) // NOI18N
-         || isAqua();
 
     // static .....................................................................................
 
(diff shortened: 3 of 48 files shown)


ide/csl.api/src/org/netbeans/modules/csl/editor/completion/MethodParamsTipPaintComponent.java
```diff
@@ -51,20 +51,8 @@ public MethodParamsTipPaintComponent(List<List<String>> params, int idx, JTextCo
...
-            try {
-                super.paint(g2d);
-            } finally {
-                g2d.setRenderingHints(oldHints);
-            }
-        } else {
-            super.paint(g);
-        }
+        GraphicsUtils.configureRenderingHints(g);
+        super.paint(g);
```
neilcsmith-net, line 55: This isn't the same behaviour.  Should it be?  What are the implications of not being?

Another method could be `GraphicsUtils.withRenderingHints(Graphics g, Consumer<Graphics> paint)` that sets and resets around the paint task?

neilcsmith-net (commented):
Good idea, but don't know if the changes in behaviour are?

ide/csl.api/src/org/netbeans/modules/csl/editor/completion/MethodParamsTipPaintComponent.java
eirikbakke: Yes, here is one example of a change in behavior. My inclination here is that resetting the hints is unnecessary, and that it would be worth simplifying the code. In the worst case, some other painting code ends up using the leftover hints--but they are the correct hints that should be used throughout the IDE in any case. Any code that wishes to use explicitly different hints had better set their own hints.

eirikbakke: The other example of a change in behavior is code that previously just took hints from awt.font.desktophints, but which now ends up doing the whole "isGTK() && gtkShouldAntialias()" etc. condition that is found in some existing places. Arguably doing the same in every place is better.

eirikbakke: I also added the "put(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)" hint, which enables anti-aliasing for non-text things such as diagonal lines and curves. Some existing sites did this already (presumably the ones that actually needed it).

platform/api.visual/src/org/netbeans/api/visual/widget/SceneComponent.java
```diff
@@ -101,12 +101,7 @@ public void paint (Graphics g) {
 //        System.out.println ("CLIP: " + g.getClipBounds ());
 //        long s = System.currentTimeMillis ();
         Graphics2D gr = (Graphics2D) g;
-
-        Object props = Toolkit.getDefaultToolkit ().getDesktopProperty ("awt.font.desktophints"); // NOI18N
-        if (props instanceof Map)
-            gr.addRenderingHints ((Map) props);
-        gr.setRenderingHint (RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
-        gr.setRenderingHint (RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
```
eirikbakke, line 109: Here's another case with a slight change in behavior from before. Previously, standard text antialiasing would be applied, but now, subpixel antaliasing might be applied on Windows if it's enabled in the rest of the IDE. I think that's good.

platform/o.n.core/src/org/netbeans/beaninfo/editors/FontEditor.java
```diff
@@ -140,10 +136,7 @@ public void paintValue (Graphics g, Rectangle rectangle) {
     }
 
     private void paintText (Graphics g, Rectangle rectangle, String text) {
-        if( antialias && g instanceof Graphics2D ) {
-            ((Graphics2D)g).setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
-            ((Graphics2D)g).setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
-        }
+        GraphicsUtils.configureRenderingHints(g);
```
eirikbakke, line 139: Here's another spot where the new behavior might end up enabling subpixel antialiasing, vs. the old standard antialiasing. That's good.

platform/o.n.swing.plaf/src/org/netbeans/swing/plaf/util/UIUtils.java
```diff
@@ -25,21 +25,12 @@
...
-import java.util.Map;
 
 /** XP color scheme installer.
  *
  * @author  Dafe Simonek
  */
 public final class UIUtils {
-    private static HashMap<RenderingHints.Key, Object> hintsMap = null;
-    private static final boolean noAntialias =
-        Boolean.getBoolean("nb.no.antialias"); //NOI18N
```
eirikbakke, line 38: This property will no longer be respected. I think that's fine; it doesn't seem official, or needed anymore.

platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/ColorUtil.java
```diff
@@ -38,8 +39,6 @@
     private static Map<RenderingHints.Key, Object> hintsMap = null;
     private static final boolean noGpCache = Boolean.getBoolean(
             "netbeans.winsys.nogpcache");  //NOI18N
-    private static final boolean noAntialias = 
-        Boolean.getBoolean("nb.no.antialias"); //NOI18N
```
eirikbakke, line 42: This property will no longer be respected. I think that's fine; it doesn't seem official, or needed anymore.

platform/openide.awt/src/org/openide/awt/GraphicsUtils.java
```diff
@@ -0,0 +1,91 @@
...
+public final class GraphicsUtils {
+    private GraphicsUtils() {
+    }
+
+    private static final boolean antialias =
+        // System property to automatically turn on antialiasing for html strings
+        Boolean.getBoolean("nb.cellrenderer.antialiasing") // NOI18N
+         ||Boolean.getBoolean("swing.aatext") // NOI18N
+         ||(isGTK() && gtkShouldAntialias()) // NOI18N
+         || isAqua();
```
eirikbakke, line 39: I don't think nb.cellrenderer.antialasing or swing.aatext, or the isAqua() conditions are relevant or needed anymore. But I've left them in.

platform/openide.awt/src/org/openide/awt/GraphicsUtils.java
```diff
@@ -0,0 +1,91 @@
...
+        return "Aqua".equals(UIManager.getLookAndFeel().getID());
+    }
+
+    private static boolean isGTK () {
+        return "GTK".equals(UIManager.getLookAndFeel().getID());
+    }
+
+    private static final boolean gtkShouldAntialias() {
+        if (gtkAA == null) {
+            Object o = Toolkit.getDefaultToolkit().getDesktopProperty("gnome.Xft/Antialias"); //NOI18N
```
eirikbakke, line 53: Nowadays, I think this condition is already incorporated by JDK, and put into the awt.font.desktophints property (since https://bugs.openjdk.java.net/browse/JDK-5030990 ). But it doesn't hurt to leave it.

eirikbakke:
OK, I have reviewed the whole diff again myself, and everything seems to be as intended. Will remove the "work-in-progress" tag once I have tested the patch in my working IDE for a while.

eirikbakke:
I have now run this patch on a while on my Windows IDE (Windows LAF), and also tested it on MacOS (Aqua LAF) and Linux (on both the GTK+ and FlatLAF Dark LAFs). Text rendering looks good/unchanged.

Removing the "work-in-progress" label from this PR; I think it should be safe to merge at this point.

matthiasblaesing:
I eyeballed this and to me this looks like a good cleanup. The amount of duplicate code and differing implementations was interesting, but I think the consolidation makes it easier to reason about the graphics state and unify hacks into a single place.

I had this running on my IDE at home for a few days and prepared a test build to use at work. I hope I see nothing special :-)

eirikbakke:
Great, thank you for testing!

matthiasblaesing:
The cycle is short, so I suggest to work on this now. I now ran a few weeks with this PR merged and did not see problems. I have a small followup change, that uses this and improves the look of the unittest bar and would like to get this done :-).

From my POV is needed here:
- rebase this onto current master (there are conflicts and IMHO rebasing gives cleaner results than merging master)
- document the new method in `apichanges.xml`. For the name might it make sense to name the method "configureDefaultRenderingHints"? And an implementation note that describes what that means?

eirikbakke:
Great, I'll rebase and apply the suggested changes...

eirikbakke:
I force-pushed a revised version of these two commits:
* Rebased to master to fix merge conflict.
* Added apichanges.xml entry.
* Renamed the new method to configureDefaultRenderingHints.
* Added Javadoc to the new configureDefaultRenderingHints.

Now just waiting for the build and tests to finish...

matthiasblaesing (approved):
To me this looks sane.

eirikbakke:
Thanks--will merge! The remaining travis errors seem to be unrelated (all in the "Test groovy modules" series).

PR #2570, opened 2020-11-30
[NETBEANS:4447]:Fixed NullPointerException while computing switch exp…

Mohan-Sarilla:
Fixed NullPointerException while computing switch expression conversion and formatting issue

PR #3118, opened 2021-08-19
By default build release-like version of .vsix

JaroslavTulach:
The `.vsix` file as published on [marketplace](https://marketplace.visualstudio.com/items?itemName=ASF.apache-netbeans-java) has traditionally been assembled together with `nbjavac` library for better cross-JDK editing support and ability to run the extension on JDK8 without any additional downloads. Time to make it the default. That's what  1d74cad does.

In addition to that I am opening access to default release catalog - originally removed in a panic by ac183a7904d1c53d7e89df18e36bcc8188005b19 - turned out that properly building the `.vsix` file with proper metarelease information and also fa85b5178041cba0ab65d70ec2e6a6340a04ef5e is enough to prevent the former problems.

Moreover having the update center enabled allows one to install additional modules not packaged in `.vsix` by default:
```bash
java/java.lsp.server/vscode$ npm run nbcode -- --modules --install .*debugger.*

> apache-netbeans-java@12.4.0 nbcode /home/devel/NetBeansProjects/netbeans2/java/java.lsp.server/vscode
> node ./out/nbcode.js "--modules" "--install" ".*debugger.*"

Installing org.netbeans.modules.debugger.jpda.ant@1.51

Installing org.netbeans.modules.debugger.jpda.heapwalk@1.44
Installing org.netbeans.modules.debugger.jpda.visual@1.30
Installing org.netbeans.modules.web.javascript.debugger@1.28
Installing org.netbeans.modules.ant.debugger@1.50.0.2
Installing org.netbeans.modules.debugger.jpda.jsui@1.13
Installing org.netbeans.modules.debugger.jpda.js@1.23
Installing org.netbeans.modules.debugger.jpda.kit@1.28

modules=21

Establishing a connection ...
Establishing a connection ...
```
which greatly enhances the flexibility of our VSCode support.

### DIFF
--- a/java/java.lsp.server/build.xml
+++ b/java/java.lsp.server/build.xml
@@ -41,11 +41,8 @@
         <chmod file="${lsp.build.dir}/bin/nbcode" perm="u+x" />
         <chmod file="${lsp.build.dir}/platform/lib/nbexec" perm="u+x"/>
         <chmod file="${lsp.build.dir}/java/maven/bin/mvn" perm="u+x" />
-        <delete file="${lsp.build.dir}/lib/nb-javac-9-api.jar" />
-        <delete file="${lsp.build.dir}/lib/nb-javac-9-impl.jar" />
-        <delete file="${lsp.build.dir}/lib/org-netbeans-modules-java-source-nbjavac.jar" />
 
-        <property name="3rdparty.modules" value="dont-install-anything"/>
+        <property name="3rdparty.modules" value=".*nbjavac.*"/>
         <autoupdate todir="${lsp.build.dir}/extra" updatecenter="file://${nb_all}/nb/updatecenters/build/classes/org/netbeans/modules/updatecenters/resources/3rdparty-catalog.xml">
             <modules includes="${3rdparty.modules}"/>
         </autoupdate>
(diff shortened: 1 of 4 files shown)


entlicher:
That's nice.
Installation of additional modules is nice for power users, the corresponding functionality might be limited by the existing LSP support. But if LSP is made extensible via NB modules, anyone could write a VSCode extension that would install some NB modules to extend the VSCode functionality.

entlicher (approved):

sdedic (approved):
At last the users won't need to download necessary binaries themselves.

neilcsmith-net (approved):
Looks good to me!

Now we're past legal questions on nb-javac, be good to move to bundling by default with the IDE too for 12.6.  Which might later lead to revising this?

One question for both would be - what's the mechanism for disabling nb-javac when bundled, possibly for testing, but also when release target is higher (eg. 17-EA) than supported by nb-javac?

JaroslavTulach:
> Now we're past legal questions on nb-javac, be good to move to bundling by default with the IDE too for 12.6. Which might later lead to revising this?

Hopefully the operation is idempotent: if the nb-javac modules are already in, nothing happens. Of course, if the catalog isn't produced during the build, this piece of Ant script has to be deleted as well.

> One question for both would be - what's the mechanism for disabling nb-javac when bundled, possibly for testing, but also when release target is higher (eg. 17-EA) than supported by nb-javac?

I usually use

```bash
$ find netbeans | grep nbjavac | xargs rm
```
but I agree, it is not very end-user friendly. If nb-javac is visible in "Tools/Plugins", users of NetBeans IDE could potentially disable it.

JaroslavTulach:
> That's nice.
> ... anyone could write a VSCode extension that would install some NB modules to extend the VSCode functionality.

Yes, this was always the goal! Anyone can write a VSCode extension and extend the functionality. To quote #2570 

> The launcher attempts to support extensibility. nbcode.ts scans all activated extensions and collects clusters from nbcode subdirectory. As such it is possible for various VSCode extensions to contribute modules into the same backend and enhance its functionality.

PR #3119, opened 2021-08-19
Restart frame action implemented in the DAP.

entlicher:


ide/spi.debugger.ui/src/org/netbeans/spi/debugger/ui/DebuggingView.java
```diff
@@ -508,6 +508,36 @@ default String getSourceMimeType() {
...
+
+        /**
+         * Pop all frames up to and including this frame off the stack.
+         *
+         * @throws UnsupportedOperationException thrown when popping of stack frames is not supported.
+         * @throws PopException when the pop frame operation fails.
+         * @since 2.70
+         */
+        default void popOff() throws UnsupportedOperationException, PopException {
+            throw new UnsupportedOperationException("The frame pop is not supported.");
```
JaroslavTulach, line 520: I would consider throwing just `PopException("Not supported")`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
```diff
@@ -349,6 +352,32 @@ public NbProtocolServer(DebugAdapterContext context) {
...
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        NbFrame stackFrame = (NbFrame) context.getThreadsProvider().getThreadObjects().getObject(args.getFrameId());
+        String popError = null;
+        if (stackFrame != null) {
+            try {
+                stackFrame.getDVFrame().popOff();
+                ActionsManager am = DebuggerManager.getDebuggerManager().getCurrentEngine().getActionsManager();
+                am.doAction("stepInto");
+                future.complete(null);
+            } catch (UnsupportedOperationException ex) {
```
JaroslavTulach, line 369: If only `PopException` is thrown, this catch can be removed.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbProtocolServer.java
```diff
@@ -349,6 +352,32 @@ public NbProtocolServer(DebugAdapterContext context) {
...
+                popError = Bundle.MSG_FrameRestartUnsupported();
+            } catch (DebuggingView.PopException ex) {
+                popError = Bundle.MSG_FrameRestartFailed(ex.getLocalizedMessage());
```
JaroslavTulach, line 372: Please note that `PopException` is final and doesn't override `getLocalizedMessage()` - e.g. the value is always going to be same as `getMessage()`.

JaroslavTulach (approved):

PR #3123, opened 2021-08-24
LSP: Move refactoring added.

dbalek:
- Move refactoring added
- CodeActions rewitten to Parsing API

### DIFF
--- a/ide/refactoring.api/src/org/netbeans/modules/refactoring/plugins/FileMovePlugin.java
+++ b/ide/refactoring.api/src/org/netbeans/modules/refactoring/plugins/FileMovePlugin.java
@@ -80,7 +80,7 @@ public class FileMovePlugin implements RefactoringPlugin {
     public void cancelRequest() {
     }
     
-    private class MoveFile extends SimpleRefactoringElementImplementation {
+    public class MoveFile extends SimpleRefactoringElementImplementation {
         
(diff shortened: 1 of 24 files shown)


ide/refactoring.api/src/org/netbeans/modules/refactoring/plugins/FileMovePlugin.java
```diff
@@ -80,7 +80,7 @@ public Problem checkParameters() {
     public void cancelRequest() {
     }
     
-    private class MoveFile extends SimpleRefactoringElementImplementation {
+    public class MoveFile extends SimpleRefactoringElementImplementation {
```
JaroslavTulach, line 83: Is this new API class supposed to be subclassable? As it seems it is...

JaroslavTulach: There should be an API change note, I assume. Btw. is everything referenced - e.g. `SimpleRefactoringElementImplementation`, `RefactoringElementsBag` already public? Btw. I don't see a reason why this class was made public - searching for `MoveFile` yields no results...

JaroslavTulach (changes requested):
Justify, document making `MoveFile` class public, please.

PR #3124, opened 2021-08-25
Merging Groovy bugfixes from delivery to master

sdedic:
I have some development (more bugfixes) ready to form a PR, but I need to build on top of changes that are currently only in `delivery`. This PR brings those commits to `master`, so I can build on top of master ...

If the `delivery` is bulk-merged in in the meantime by RM, this PR becomes obsolete.


neilcsmith-net:
Delivery was bulk merged a few hours ago (not that I'm meant to be RM'ing) Please check what you're expecting is in there.

PR #3125, opened 2021-08-25
Use static type checking to attribute AST

sdedic:
The basic idea is to use Groovy compiler's `StaticTypeCheckingVisitor` to attribute the AST. The visitor leaves evaluated types in node's metadata ('StaticTypesMarker.INFERRED_TYPE`), which is more precise than our own type evaluator (naturally).
Even more important is `StaticTypesMarker.DIRECT_METHOD_CALL_TARGET` that points to the method (`MethodNode`) to be called from the expression: out code completion attemoted to evaluate the types itself, but did not even count with subtyping (only with exactly same types) ! Not to mention things like boxing, unboxing, closures etc.

The static visitor however posts some error message. Some of them are meant for a real static compilation - report usage of disallowed features - I've filtered them out. Other ones are valid (i.e. method suitable for arg types not found), but would fail the compilation process. So I have collected them aside, making the compiler think that everything is OK and let the compilation fail after last phase if there are some errors. 

---
Code completion did not check access properly, offered sometimes even private stuff - should be fixed for properties, fields and methods. There are some special cases, too. `GString` is autocoerced to `String` when necessary. So for `GString` expressions,. bot String and GString methods are valid for completion. For this I had to fix up method and field signatures, so completion from two type-unrelated classes can merge without duplicities. Another such case is usage of class name literal - it acts both as a class qualifier, so static class members should be offered by the completion. At the same time, the class name is a shorthand for `ClassName.class`, so all `java.lang.Class<T>` properties / methods are valid completion items too - again merging.

### DIFF
--- a/groovy/groovy.editor/test/unit/src/org/netbeans/modules/groovy/editor/api/completion/MethodCCTest.java
+++ b/groovy/groovy.editor/test/unit/src/org/netbeans/modules/groovy/editor/api/completion/MethodCCTest.java
@@ -157,5 +157,83 @@ public class MethodCCTest extends GroovyCCTestBase {
     public void testMethods4() throws Exception {
         checkCompletion(BASE + "Methods4.groovy", "iface.meth^", false);
     }
+    
+    /**
+     * Checks method completion on direct reference to a variable
+     */
+    public void testMethods5_chain1() throws Exception {
+        checkCompletion(BASE + "Methods5.groovy", " someFile.getP^arentFile()", true);
+    }
+
+    /**
+     * Checks method completion on 2nd reference in the chain, filtering for "m"
+     */
+    public void testMethods5_chain2() throws Exception {
+        checkCompletion(BASE + "Methods5.groovy", " someFile.getParentFile().mk^dirs()", true);
+    }
+
+    /**
+     * Checks method completion on 2nd reference in the chain, without any filter (so with
+     * fields as well)
+     */
+    public void testMethods5_chain2a() throws Exception {
+        checkCompletion(BASE + "Methods5.groovy", " someFile.getParentFile().^mkdirs()", true);
+    }
+
+    /**
+     * Checks completion in even higher in the hierarchy.
+     */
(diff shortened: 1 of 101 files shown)


sdedic:
* `GroovyIndexer` did not record modifiers. Groovy does not use them, in fact, when calling Groovy classes (see [GROOVY-3010](https://issues.apache.org/jira/browse/GROOVY-3010)) but they can be useful to have in the index.
* Meta-methods were weaved into normal class methods, now they are sorted at the end
* When merging, the `j.l.Class` items should be sorted after the type's own ones.
* Static modifier was not checked on members
* Access modifiers were not checked on Java members
* Added a possibility to enable access-aware completion (in the future), the fact that the IDE suggests to access fields directly is nothing but ugly ...

groovy/groovy.editor/test/unit/src/org/netbeans/modules/groovy/editor/api/completion/MethodCCTest.java
```diff
@@ -157,5 +157,83 @@ public void testCompletionNoPrefixString2() throws Exception {
...
+     */
+    public void testMethods5_chain1() throws Exception {
+        checkCompletion(BASE + "Methods5.groovy", " someFile.getP^arentFile()", true);
+    }
+
+    /**
+     * Checks method completion on 2nd reference in the chain, filtering for "m"
+     */
+    public void testMethods5_chain2() throws Exception {
+        checkCompletion(BASE + "Methods5.groovy", " someFile.getParentFile().mk^dirs()", true);
```
JaroslavTulach, line 172: This shows that the chained calls work well with code completion, right?

JaroslavTulach (approved):
I've checked the tests and some of them seem to move the code completion further. Looks like an improvement.

sdedic:
I had to disable error reporting from static analysis - too much noise. Still can be enabled by system property. Also updated test golden files for various JDK flavours.

JaroslavTulach (approved):
Tests are there and they seem to improve something.

PR #3129, opened 2021-08-27
Eating VSCode own Dog Food

JaroslavTulach:
Support for editing so called _API support_ projects has been intentionally left out of VSNetBeans. It is archaic and out of interest of the new generation of developers. However we, the NetBeans oldtimers, do care about editing/compiling/debugging Apache NetBeans sources.

To balance the need of these two target groups I propose a simple way to build own customized VSNetBeans extension with _API support_ modules in. Just build the extension as you are used to and then execute:

```bash
java/java.lsp.server/vscode$ npm run apisupport
```
Then you can open any Apache NetBeans project and work with it without any limitations:

```bash
java/java.lsp.server/vscode$ code --extensionDevelopmentPath=`pwd` ../../platform/autoupdate.cli/
```

Should the project still not be recognized, consider removing previous debris. I had to use 
```bash
java/java.lsp.server/vscode$ rm -rf ~/.config/Code/User/globalStorage/asf.apache-netbeans-java/ nbcode/
```
few times.

### DIFF
--- a/platform/autoupdate.cli/src/org/netbeans/modules/autoupdate/cli/ModuleOptions.java
+++ b/platform/autoupdate.cli/src/org/netbeans/modules/autoupdate/cli/ModuleOptions.java
@@ -23,7 +23,6 @@ import java.io.PrintStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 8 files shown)


dbalek (approved):
Thanks.

platform/autoupdate.cli/src/org/netbeans/modules/autoupdate/cli/ModuleOptions.java
```diff
@@ -395,7 +381,11 @@ private void install(final Env env, String... pattern) throws CommandException {
         }
         final InstallSupport support = operate.getSupport();
         if (support == null) {
-            env.getOutputStream().println(Bundle.MSG_InstallNoMatch(Arrays.asList(pats)));
+            if (!found) {
+                sb.insert(0, MSG_CannotFindAnyPattern(cnm));
+                throw new CommandException(55, sb.toString());
```
sdedic, line 386: This exitcode should be documented somehow.

sdedic (approved):

PR #3130, opened 2021-08-27
Keep old plain for loops in NetBeans sources

JaroslavTulach:
I am all for _clean code_ and keeping our sources without warnings and not applied hints. However, the default set of hints includes "convert to functional operations" and I find that one particularly annoying. I am quite fine using _OPFL - old plain for loops_ most of the time.

As such I propose to suppress the hint a bit. Make it visible only on affected line, for example. I have used the NetBeans IDE UI that allows to disable the hint and following change appeared on my disk. I assume, if we integrate it, the "convert to functional operations" hint is going to be suppressed (a bit).

Various IDEs offer summaries of errors/warnings in the code and I'd be really thankful if only important hints were included in such summaries.

### DIFF
--- a/nbbuild/misc/hints-settings.xml
+++ b/nbbuild/misc/hints-settings.xml
@@ -100,7 +100,10 @@
             <attribute name="enabled" value="true"/>
         </node>
         <node name="org.netbeans.modules.java.hints.bugs.BroadCatchBlock.broadCatch"/>
-        <node name="org.netbeans.modules.java.hints.jdk.mapreduce.ForLoopToFunctionalHint"/>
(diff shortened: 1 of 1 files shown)


matthiasblaesing (approved):
Looks sane to me. Thank you.

neilcsmith-net:
> Make it visible only on affected line, for example.

+1 This is one of a bunch of recent hints that I'd prefer to see only on the affected line - IDE is getting noisier.

JaroslavTulach:
Thanks to advice by @jlahoda I found a way to let the "Use Funcitonal Operation" hint be visible when one is on the `for` cycle line, but not reported as a global warning. As a result one can still use the hint, when putting the cursor on the right line, but one is not annoyed by aggressive suggestions to use too functional coding style.

PR #3136, opened 2021-08-28
[NETBEANS-5914] Call notifyAll + avoid delivering changes under a lock

JaroslavTulach:
Visual inspection of `LazyLookupProviders` code suggests an inconsistency: `Object.wait()` waits on value of `LOCK[0]` and such value can be provided without  setting calling `Object.notifyAll()`. This should be an easy fix for [this deadlock](https://issues.apache.org/jira/secure/attachment/13032558/13032558_12.5-beta2-threaddump-1629986898821.tdump).

The rest of the [NETBEANS-5914](https://issues.apache.org/jira/browse/NETBEANS-5914) issue is going to be exorcized by 9227546

### DIFF
--- a/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookup.java
+++ b/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookup.java
@@ -18,9 +18,12 @@
  */
 package org.netbeans.modules.projectapi;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Executor;
 import java.util.logging.Level;
 import org.openide.filesystems.FileObject;
 import org.openide.util.Exceptions;
@@ -29,7 +32,6 @@ import org.openide.util.lookup.Lookups;
 import org.openide.util.lookup.ProxyLookup;
 
 final class LazyLookup extends ProxyLookup {
-    
     private final Map<String, Object> attrs;
     private final Lookup lkp;
     private Collection<String> serviceNames;
@@ -39,12 +41,28 @@ final class LazyLookup extends ProxyLookup {
         this.attrs = attrs;
         this.lkp = lkp;
         this.serviceNames = Arrays.asList(((String) attrs.get("service")).split(",")); // NOI18N
-    } // NOI18N
+    }
 
     @Override
     protected void beforeLookup(Template<?> template) {
+        class NotifyLater implements Executor {
+            private List<Runnable> pending = new ArrayList<>();
+            @Override
+            public void execute(Runnable command) {
+                pending.add(command);
+            }
+            
+            public void deliverPending() {
+                List<Runnable> notify = pending;
+                pending = null;
+                for (Runnable r : notify) {
+                    r.run();
+                }
+            }
+        }
         LazyLookupProviders.safeToLoad(this.lkp);
         Class<?> service = template.getType();
+        NotifyLater later = null;
         synchronized (LOCK) {
             for (;;) {
                 if (serviceNames == null || !serviceNames.contains(service.getName())) {
@@ -57,10 +75,6 @@ final class LazyLookup extends ProxyLookup {
                     return;
                 }
                 try {
-                    /* deadlock - probably wait without a notify
-                    https://issues.apache.org/jira/secure/attachment/13032558/13032558_12.5-beta2-threaddump-1629986898821.tdump
-                    always make sure LOCK.notifyAll is called when changing the LOCK[0] value
-                     */
                     LOCK.wait();
                 } catch (InterruptedException ex) {
                     LazyLookupProviders.LOG.log(Level.INFO, null, ex);
@@ -75,9 +89,10 @@ final class LazyLookup extends ProxyLookup {
                 // JRE #6456938: Class.cast currently throws an exception without details.
                 throw new ClassCastException("Instance of " + instance.getClass() + " unassignable to " + service);
             }
-            setLookups(Lookups.singleton(instance));
+            setLookups(later = new NotifyLater(), Lookups.singleton(instance));
             synchronized (LOCK) {
                 serviceNames = null;
+                LOCK.notifyAll();
             }
         } catch (Exception x) {
             Exceptions.attachMessage(x, "while loading from " + attrs);
@@ -87,13 +102,20 @@ final class LazyLookup extends ProxyLookup {
                 LOCK[0] = null;
                 LOCK.notifyAll();
             }
+            if (later != null) {
+                later.deliverPending();
+            }
         }
     }
 
+    boolean isInitializing() {
+        return LOCK[0] != null;
+    }
+    
     @Override
     @SuppressWarnings(value = "element-type-mismatch")
     public String toString() {
         return "LazyLookupProviders.LookupProvider[service=" + attrs.get("service") + ", class=" + attrs.get("class") + ", orig=" + attrs.get(FileObject.class) + "]";
     }
-    
+
 }
--- a/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookup.java
+++ b/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookup.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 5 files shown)


neilcsmith-net:
In theory was the call to `notifyAll()` at line 107 assumed to cover all bases?  While this is probably a good change, is not the call to `setLookups()` a problem here as well as it's effectively holding a lock and firing listeners?  Should they both be passing an executor, possibly a single one, to fire listeners async?

eg. one of the stack traces has this - 

```
        at java.lang.Object.wait(java.base@11.0.11/Object.java:328)
        at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:84)
        - waiting to re-lock in wait() <0x0000000099880000> (a [Ljava.lang.Thread;)
        at org.openide.util.lookup.ProxyLookup.beforeLookup(ProxyLookup.java:268)
        ....
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
        at org.netbeans.spi.project.support.DelegatingLookupImpl.doDelegate(DelegatingLookupImpl.java:168)
        - locked <0x0000000099880458> (a java.util.ArrayList)
        at org.netbeans.spi.project.support.DelegatingLookupImpl.resultChanged(DelegatingLookupImpl.java:84)
        at org.openide.util.lookup.ProxyLookup$1Notify.run(ProxyLookup.java:239)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:248)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
        at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:97)
```

JaroslavTulach:
>  is not the call to setLookups() a problem here as well

Possibly. _Git.cache_ thread holds the 0x00000000cf989738 lock:
```
	at org.netbeans.spi.project.support.DelegatingLookupImpl.doDelegate(DelegatingLookupImpl.java:168)
	- locked <0x00000000cf989738> (a java.util.ArrayList)
```
and many other threads are waiting for it.

neilcsmith-net:
Yes, there's the one in `DelegatingLookupImpl`, and the one in `LazyLookupProviders` earlier in the above stack that is (effectively) under the lock here.  Seems to be a deadlock between those two?  I was reviewing this and trying to understand why it was hanging before the later call to `notifyAll()`.

https://github.com/apache/netbeans/blob/master/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookupProviders.java#L97

JaroslavTulach:
9227546 comes with a test and a fix to make sure the `delegates` lock isn't held when notifying changes.

sdedic (approved):
Great !

neilcsmith-net:
Doesn't this need the same change to take the notifications outside of the lock in `LazyLookupProviders`?  Looking at the stack traces provided with NETBEANS-5914, all the threads with the `DelegatingLookupImpl` lock are stuck in `collectFires()` - that won't change with this will it?  It's before the async notification - https://github.com/apache/netbeans/blob/master/platform/openide.util.lookup/src/org/openide/util/lookup/ProxyLookup.java#L228  They are all in a listener call from a `LazyLookupProvider` under lock though.

JaroslavTulach:
I don't think there is a problem in `org.openide.util.lookup` package. Feel free to write a test to prove I am wrong.

neilcsmith-net:
> I don't think there is a problem in org.openide.util.lookup package. Feel free to write a test to prove I am wrong.

Neither do I.  I think the problem is in `org.netbeans.modules.projectapi.LazyLookupProviders` and that [call to setLookups](https://github.com/apache/netbeans/blob/master/ide/projectapi/src/org/netbeans/modules/projectapi/LazyLookupProviders.java#L97) that is done holding the conditional lock.

I'll have a think about a test for that scenario if you need one, but it's what the stack trace seems to suggest ? ...

> Btw. I don't see any dump where lock are stuck in collectFires(). Please point me where you see something like that.

I mean the thread with the lock from DelegatingLookupImpl is stuck in collectFires() rather than in calling through to its listeners.  eg. all the stack traces show the following similar (trimmed) stack, with resultChanged() being called in DelegatingLookupImpl fired under the lock from one LazyLookupProviders proxy, taking the lock in DelegatingLookupImpl, then waiting on the lock in another LazyLookupProviders proxy during a call to collectFires().

```
        at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:84)
        - waiting to re-lock in wait() <0x0000000601a8dca0> (a [Ljava.lang.Thread;)
        at org.openide.util.lookup.ProxyLookup.beforeLookup(ProxyLookup.java:268)
        .....
        at org.openide.util.lookup.ProxyLookup$R.collectFires(ProxyLookup.java:645)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:228)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
        at org.netbeans.spi.project.support.DelegatingLookupImpl.doDelegate(DelegatingLookupImpl.java:168)
        - locked <0x0000000601a8ef08> (a java.util.ArrayList)
        at org.netbeans.spi.project.support.DelegatingLookupImpl.resultChanged(DelegatingLookupImpl.java:84)
        at org.openide.util.lookup.ProxyLookup$1Notify.run(ProxyLookup.java:239)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:248)
        at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
        at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:97)
```

Various of the other threads waiting on the lock in DelegatingLookupImpl are also being called under lock from a LazyLookupProvider proxy, presumably one of which is the one the above stack is waiting for.  I'm not sure there's an easy way to ascertain which from a stack trace with this sort of conditional lock?  I'd not even noticed it until I looked at your first commit here.

jtulach:
Hello Neil, this is 3rd or 4th version of my reply. Sometimes I agree, sometimes I don't...

This PR clearly fixes [something](https://github.com/apache/netbeans/pull/3136/commits/9227546960199ecec44755a03cce6f1d59ade69d) (as shown by the test). Whether that is enough to address the problem of [this deadlock](https://issues.apache.org/jira/secure/attachment/13032558/13032558_12.5-beta2-threaddump-1629986898821.tdump):
```
	at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:84)
	- locked <0x00000000cf989400> (a [Ljava.lang.Thread;)
	at org.openide.util.lookup.ProxyLookup.beforeLookup(ProxyLookup.java:268)
``` 
remains to be seen. There is:
```
"org.netbeans.modules.project.ui.actions.LookupSensitiveAction" #51 daemon prio=1 os_prio=-2 cpu=46.88ms elapsed=78.28s tid=0x0000000031c145b0 nid=0x1f3c waiting for monitor entry  [0x0000000047bce000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.netbeans.spi.project.support.DelegatingLookupImpl.doDelegate(DelegatingLookupImpl.java:100)
	- waiting to lock <0x00000000cf989738> (a java.util.ArrayList)
	at org.netbeans.spi.project.support.DelegatingLookupImpl.resultChanged(DelegatingLookupImpl.java:84)
	at org.openide.util.lookup.ProxyLookup$1Notify.run(ProxyLookup.java:239)
	at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:248)
	at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
	at org.netbeans.modules.projectapi.LazyLookupProviders$1Prov$1.beforeLookup(LazyLookupProviders.java:97)
```
which would at the end notify the other waiters in `LazyLookupProviders$1Prov$1`. Right now it is blocked, but with changes in this PR it might become free. If not, then the next thing to tackle is
```
	at org.openide.util.lookup.ProxyLookup.setLookups(ProxyLookup.java:184)
	at org.netbeans.spi.project.support.DelegatingLookupImpl.doDelegate(DelegatingLookupImpl.java:168)
	- locked <0x00000000cf989738> (a java.util.ArrayList)
	at org.netbeans.spi.project.support.DelegatingLookupImpl.resultChanged(DelegatingLookupImpl.java:84)
```
e.g. eliminate call to `setLookups` under the lock `0x00000000cf989738`.

JaroslavTulach:
Or better. Let's deliver events only when `LOCK[0] == null`: 5a44f56

neilcsmith-net (approved):
> Or better. Let's deliver events only when LOCK[0] == null: 5a44f56

Perfect! Thanks for bearing with me on this.

ebarboni:
Ok will merge

PR #3137, opened 2021-08-28
Add missing --add-opens to fix warnings on recent JDKs

matthiasblaesing:
Covered packages:

- java.desktop/javax.swing.plaf.synth
- java.desktop/com.sun.java.swing.plaf.gtk
- java.desktop/sun.awt.shell
- java.desktop/sun.awt.im
- java.base/java.nio


matthiasblaesing:
Warnings were observed running NetBeans 12.5 beta 2 on JDK 15. This prepares for JDK-16 were access will be denied by default.

I created this PR against delivery as we might get a beta 3 (NETBEANS-5914 is still marked as blocker), then I think this is appropriate. If there is no need for another beta, this can be retargetted to master.

neilcsmith-net:
Did you catch where these were triggering warnings?  The plaf ones I remember seeing in the past, but not the others.  I, and I assume quite a few others, have been running 12.4 and 12.5 betas on JDK 16 for months.  Do we have anything important being denied here that might actually be a cause of significant issues?

I think NETBEANS-5914 will end up being bumped to 12.6 unless we have more reports - a quick fix might be riskier - so probably won't force a beta3.

ebarboni:
beta 3 will be created so I will merge this

matthiasblaesing:
Thank you!

PR #3138, opened 2021-08-30
LSP: Extract Superclass and Extract Interface refactorings added.

dbalek:
Extract Superclass and Extract Interface refactorings available via LSP.


JaroslavTulach (approved):
I don't see any architectonic issues.

PR #3140, opened 2021-08-31
Improve wait condition for the test: both Lookup and configurations must change.

sdedic:
Just trying to execute `MavenActionsTest` test in CI environment, as it fails on CI and not locally. Trying to find out what's wrong.


sdedic:
The test is failing in other PRs. I'd recommend to merge this & remove output when the bug is fixed.

dbalek (approved):

sdedic:
I have some suspicions ... let's add some more messages :)

PR #3141, opened 2021-09-01
Include URLs in Java code templates for easier customization

JaroslavTulach:
While working on #2960 I forgot to modify Java code templates. That's annoying as these templates are exactly something that needs customization:

image

Clicking the URL in editor opens the template.

### DIFF
--- a/java/java.hints/test/unit/src/org/netbeans/modules/java/hints/suggestions/ImplementMethodsTest.java
+++ b/java/java.hints/test/unit/src/org/netbeans/modules/java/hints/suggestions/ImplementMethodsTest.java
@@ -44,7 +44,7 @@ public class ImplementMethodsTest {
                 .assertOutput("package test;\n" +
                               "public abstract class Test implements Runnable {\n" +
                               "    public void run() {\n" +
-                              "        throw new UnsupportedOperationException(\"Not supported yet.\"); //To change body of generated methods, choose Tools | Templates.\n" +
+                              "        throw new UnsupportedOperationException(\"Not supported yet.\"); // Generated from nbfs://nbhost/SystemFileSystem/Templates/Classes/Code/GeneratedMethodBody\n" +
                               "    }\n" +
                               "}\n");
     }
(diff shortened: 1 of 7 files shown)


junichi11 (approved):
Great!

sdedic:
Could it be possible to define a standard variable, like ${open_url} for the template expansion ?

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/suggestions/ImplementMethodsTest.java
```diff
@@ -44,7 +44,7 @@ public void testSimple1() throws Exception {
                 .assertOutput("package test;\n" +
                               "public abstract class Test implements Runnable {\n" +
                               "    public void run() {\n" +
-                              "        throw new UnsupportedOperationException(\"Not supported yet.\"); //To change body of generated methods, choose Tools | Templates.\n" +
+                              "        throw new UnsupportedOperationException(\"Not supported yet.\"); // Generated from nbfs://nbhost/SystemFileSystem/Templates/Classes/Code/GeneratedMethodBody\n" +
```
JaroslavTulach, line 47: There doesn't seem to be any way to treat the comment as `.*` without specifying the exact content, right @jlahoda?

JaroslavTulach:
> Could it be possible to define a standard variable, like ${open_url} for the template expansion ?

There is `${.data_model["org.openide.filesystems.FileObject"].getPath()` in https://github.com/apache/netbeans/pull/2960/files#diff-498697d41fe5fc687f18b0e779cb675567c29ef83e224dac048a58b43a9a9ceeR3 - I am not sure if that is better, however.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/suggestions/ImplementMethodsTest.java
JaroslavTulach: Let's assume there is no better way and all tests had to be modified in ba90ee1

PR #3145, opened 2021-09-02
Gradle Support for Payara Micro

lkishalmi:
Well this PR would be a series of commits which at the end would enable the Gradle Payara Plugin be detected and alter some IDE actions. Theoretically it would be very easy if the reload support action shall not be provided, though I think it would be nice to do that as well. So here is where the problem begins. The Payara Reload action is an always enabled action that does some Maven goal execution and touches a reload file. However the implementation works. It is not nice to have an action in the tool bar and menu which is always enabled and when executed in the wrong context it just updates the status line with an error message.

So the first commit here would be to refactor the implementation to be more bend in the SPI of the Maven project and Project UI modules.


enterprise/payara.micro/src/org/netbeans/modules/fish/payara/micro/project/Actions.java
```diff
@@ -0,0 +1,89 @@
...
+            lazy = false
+    )
+    @ActionReferences({
+        @ActionReference(path = "Menu/BuildProject", position = 55),
+        @ActionReference(path = "Toolbars/Build", position = 325),
+        @ActionReference(path = "Projects/org-netbeans-modules-maven/Actions", position = 1020),
+        @ActionReference(path = "Shortcuts", name = "DS-A")
+    })
+    @NbBundle.Messages("CTL_ReloadAppAction=Reload")
+    public static Action reloadAction() {
```
lkishalmi, line 61: This makes the Reload action a real context sensitive project action.

enterprise/payara.micro/src/org/netbeans/modules/fish/payara/micro/project/MicroActionConvertor.java
```diff
@@ -0,0 +1,68 @@
...
+            switch (action) {
+                case COMMAND_RUN:
+                case COMMAND_DEBUG:
+                case COMMAND_PROFILE:
+                case RUN_SINGLE_ACTION:
+                case DEBUG_SINGLE_ACTION:
+                case PROFILE_SINGLE_ACTION:
+                    convertedAction = "micro." + action; //NOI18N
+                    break;
+                case COMMAND_MICRO_RELOAD:
```
lkishalmi, line 61: This is a pseudo action string for the reload action, converted according to the CoS flag.

enterprise/payara.micro/src/org/netbeans/modules/fish/payara/micro/project/MicroActionsProvider.java
```diff
@@ -65,78 +36,40 @@
         service = MavenActionsProvider.class,
         projectType = MAVEN_WAR_PROJECT_TYPE
 )
-public class MicroActionsProvider implements MavenActionsProvider {
+public final class MicroActionsProvider extends AbstractMavenActionsProvider {
 
     @StaticResource
     private static final String ACTION_MAPPINGS = "org/netbeans/modules/fish/payara/micro/project/resources/action-mapping.xml";
-        
-    private final AbstractMavenActionsProvider actionsProvider = new AbstractMavenActionsProvider() {
```
lkishalmi, line 73: Not needed as we provide unique actions in our mapping file. The MicroActionConvertor would convert standard actions to micro actions when they are supported by the project.

java/maven/src/org/netbeans/modules/maven/spi/actions/ActionConvertor.java
```diff
@@ -22,11 +22,23 @@
 import org.openide.util.Lookup;
 
 /**
+ * Register an {@code ActionConverter} in a Maven Project Lookup to convert
```
lkishalmi, line 25: Added missing JavaDoc.

enterprise/payara.micro/src/org/netbeans/modules/fish/payara/micro/project/resources/action-mapping.xml
```diff
@@ -21,7 +21,7 @@
 -->
```
lkishalmi, line 21: Made the action names specific to this plugin. Some actions are dash, some are dot separated, later on it will be standardized on dots, that's how most of the actions defined in the IDE.

lkishalmi (commented):
Adding some comments that would help the review. Some tests would be welcome, as the Maven application created by the wizard of this plugin does not start. I'll try with the wizard on the web later on.

jGauravGupta:
Thanks @lkishalmi for the PR, I will review it soon.

enterprise/payara.micro.gradle/src/org/netbeans/modules/fish/payara/micro/gradle/action-mapping.xml
```diff
@@ -0,0 +1,37 @@
...
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<!DOCTYPE actions SYSTEM "action-mapping.dtd">
+<actions>
+    <apply-for plugins="war,fish.payara.micro-gradle-plugin">
+        <action name="run">
+            <args>-DpayaraMicro.debug=true microStop microStart</args>
```
jGauravGupta, line 26: `payaraMicro.debug` flag should be removed from the `run` task.

jGauravGupta:
`MicroExecutionChecker.BUILD_ACTIONS` & `MicroExecutionChecker.RUN_ACTIONS` names are mismatching with `enterprise\payara.micro\src\org\netbeans\modules\fish\payara\micro\project\resources\action-mapping.xml`.

enterprise/payara.micro/src/org/netbeans/modules/fish/payara/micro/project/MicroActionsProvider.java
```diff
@@ -65,78 +36,40 @@
...
     }
-
+    
     @Override
-    public NetbeansActionMapping getMappingForAction(String actionName, Project project) {
-        MicroApplication microApplication = MicroApplication.getInstance(project);
-        if (microApplication != null) {
-            return actionsProvider.getMappingForAction(actionName, project);
+    public boolean isActionEnable(String action, Project project, Lookup lookup) {
+        if ("micro-explode".equals(action)) {
```
jGauravGupta, line 52: `Constants.EXPLODE_ACTION` constant can be used here

lkishalmi:
Yes. This one is a bit old. I have to refresh this for NB17

PR #3147, opened 2021-09-02
Fixing the configure hint sub-fix to work more properly for per-project hint settings.

jlahoda:
When selecting "Configure ... Hint" in the editor, when per project hints settings are in effect, the project settings should be open rather than the global settings.


JaroslavTulach (approved):

matthiasblaesing (approved):
Looks sane and a quick local test worked correctly. Any reason not to merge this?

PR #3148, opened 2021-09-02
Library to process Heap dumps without accessing java.io.File

JaroslavTulach:
Certain environments are special and do not provide direct access to the filesystem. As such I cannot use `java.io.File`, `java.io.FileInputStream` & co. Yet processing `.hprof` files may be needed even in such environments. Let's provide an API to load an `.hprof` file content from a `ByteBuffer`. The other issue is `CachedDirectory` - it cannot be used either, as it is directly accessing the filesystem. As far as I can say the current code seems to be ready to work without `CachedDirectory`, this PR just provides a way to request cache-less mode via a new API method.

### DIFF
--- a/profiler/lib.profiler/test/unit/src/org/netbeans/lib/profiler/heap/HeapFromBufferTest.java
+++ b/profiler/lib.profiler/test/unit/src/org/netbeans/lib/profiler/heap/HeapFromBufferTest.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


sdedic (approved):

profiler/lib.profiler/test/unit/src/org/netbeans/lib/profiler/heap/HeapFromBufferTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
```
matthiasblaesing, line 24: This should be the ALv2 header

matthiasblaesing (commented):
Looks ok to me. I agree with @sdedic unifiy the initialization logic in `HprofHeap` would be nice and the license header of the new test needs to be fixed - Travis is rightfully complaining :-)

PR #3150, opened 2021-09-02
Proper handling of LSP logging breakpoints.

entlicher:
The logging text is implemented for Truffle breakpoints, some corrections are applied to Java breakpoints and expression evaluation in the logging text.

### DIFF
--- a/java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/access/TruffleAccess.java
+++ b/java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/access/TruffleAccess.java
@@ -38,6 +38,7 @@ import java.beans.PropertyVetoException;
 import java.io.InvalidObjectException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


JaroslavTulach (commented):
I have little clue what is going on or what this change is fix. Do we have a description of the faulty behavior?

entlicher:
When examining the properties that can be associated with breakpoints via DAP, I've found some issues:
- logging breakpoints should not break the execution according to the spec.
- log message can contain {<expression>} to log expression values
- Truffle breakpoints (for guest languages) did not have an ability to log messages

The above issues are resolved by this PR:
- The LSP's `NbBreakpoint` instructs breakpoints not to suspend when there's a log message. It also translates patterns of `{...}` to `{=...}` which assure that JPDA (and now also Truffle) breakpoints evaluate the expressions.
- `TruffleLineBreakpoint` has new suspend and printText properties. 
- There is a new `TruffleBreakpointsRegistry`, which associates `TruffleLineBreakpoint` (`JSLineBreakpoint` in general, which can be set in JS files) with `com.oracle.truffle.api.debug.Breakpoint` in the debuggee (represented by `ObjectReference`).
- When Truffle breakpoint is hit, we create the new `HitBreakpointInfo` and use the new `TruffleBreakpointOutput` to print out log message or condition exception for Truffle breakpoints.
- There's also a small change in JPDA's `DebuggerOutput` that assures that names of guest languages sources are printed out instead of the infrastructure's `JPDATruffleAccessor` class.

dbalek (approved):

JaroslavTulach (approved):

PR #3151, opened 2021-09-03
Snapcraft Upgrade for core18

lkishalmi:
The main theme is to upgrade the Snapcraft package generation to core18 as the "coreless" package builds are no longer supported having Ubuntu 16.04 has reached it's end of life. The new build method requires a build in container with multipass or LXD. That is actually safer and a good step towards more reliable build. It also means that the Snapcraft descriptors are no longer able to access the code in their project parent directories (relative ../../ path references do not work any more), so they need to rely on Git checkout.

As the Snap project are no longer bound to the source code it might make sense to store them in an other repo. Opinions?

Also unfortunately Apache build grid does not support neither multipass nor LXD builds at the moment. Though I have filed [INFRA-22200](https://issues.apache.org/jira/browse/INFRA-22200) for that, and it seems feasible to have a few build nodes with LXD in the future.

There is also a change in the default home and cache directories. The default home would be automatically copied over to its new location by Snapcraft whenever a new release happens, and it will use a clean cache directory as well. That would mean seamles upgrade from version to version. It would make also possible to safely revert to a previous version if needed.

There is a small improvement on GitBranchHash which is able to get at least the hash of a detached HEAD, as that would be the scenario when building a release Snap out of a git tag.

It would be nice to have this in 12.5.


ebarboni:
@lkishalmi I wanted to start voting later this week on beta3. Is this a blocker for 12.5 or can we make it for 12.6 ?

neilcsmith-net:
> As the Snap project are no longer bound to the source code it might make sense to store them in an other repo. Opinions?

I'm in favour of this.  Maybe in NetBeans Tools repo?  I'm currently working on another packaging submission which is currently intended to be a PR there.  Or we start a repo for extra packaging infrastructure?

Anyway, be good to look at alternatives to delaying 12.5 any further IMO.

lkishalmi:
@ebarboni I had hoped that there would be an RC before the vote. But it would be fine to vote without this one.

neilcsmith-net:
@lkishalmi why?! We're close to running a month late already.  A discussion to move to dev@, but both my beta2 email and Eric's beta3 stated the intention to have them as vote - beta3 only happened due to couple of criticals. Issues in naming, process and/or communication need resolving here?

lkishalmi:
@neilcsmith-net I'm fine not having an RC. It was just unusual. I was quite remote from NB in the last few months, just starting to approach it again.

ebarboni:
np but that should be done prior to  next release ?

geertjanw:
We have been working for a few releases now that the last beta becomes the voting candidate. Unless there's a reason to change that, let's continue, and start voting once we have a voting candidate, hopefully soon.

lkishalmi:
I'm going to create a separate PR-s, to patch GitBranchHash, remove the Snap package creation from the netbeans repository and add the Snap package creation stuff to the netbeans-tools repository.

PR #3152, opened 2021-09-03
fixes reponse code check for http status code 301 moved permanently;

joestr:
If you are testing the connection in the settings it fails with `Connection failed (null)` despite setting the option to _No Proxy_

The method responsible for only checks for `HTTP OK (200)` and `HTTP MOVED TEMPORARILY (302)` despite http://netbeans.org (`ProxySettings.HTTP_CONNECTION_TEST_URL`) responding back `HTTP MOVED PERMANENTLY 301`.

image

### DIFF
--- a/platform/core.ui/src/org/netbeans/core/ui/options/general/GeneralOptionsModel.java
+++ b/platform/core.ui/src/org/netbeans/core/ui/options/general/GeneralOptionsModel.java
@@ -308,8 +308,11 @@ class GeneralOptionsModel {
         httpConnection.setConnectTimeout(5000);
         httpConnection.connect();
 
-        if (httpConnection.getResponseCode() == HttpURLConnection.HTTP_OK || 
-                httpConnection.getResponseCode() == HttpURLConnection.HTTP_MOVED_TEMP) {
+        int responseCode = httpConnection.getResponseCode();
+
+        if (responseCode == HttpURLConnection.HTTP_OK || 
+            responseCode == HttpURLConnection.HTTP_MOVED_TEMP ||
+            responseCode == HttpURLConnection.HTTP_MOVED_PERM) {
             result = true;
         }
 


junichi11:
@joestr Could you please create a new issue in [the JIRA](https://issues.apache.org/jira/projects/NETBEANS/issues) and read the following?
https://netbeans.apache.org/participate/submit-pr.html

NOTE: current target branch is `master` (we have to change it to `delivery` if we would like to merge it for 12.5)

neilcsmith-net:
@lkishalmi @junichi11 not sure this is critical enough to force a further beta and delay in 12.5?!

@joestr thanks for the fix. However, would it be better to check for a range of responses here?  Is anything > 0 and < 400 OK here?  And possibly changing the domain to check to apache.org?

joestr:
Maybe changing the URL to an always accesible, never changing host like http://captive.apple.com?

joestr:
@neilcsmith-net Checking for a `HTTP OK (200)` response should suffice.

matthiasblaesing:
Setting up a host that returns a http Status 200 for a given URL should be trivial and thus it is unnessary to use a third party for this. The currently configured url for connection checking is:

https://github.com/apache/netbeans/blob/3b964b0079f7b4726fd447f89cb71b56899df953/platform/o.n.core/src/org/netbeans/core/ProxySettings.java#L65

which returns a valid 200:

```
matthias@enterprise:~$ curl -sSkv -o /dev/null https://netbeans.apache.org 2>&1 | grep "^<"
< HTTP/2 200 
< server: Apache
< last-modified: Sun, 30 May 2021 17:02:24 GMT
< etag: "303c-5c38f12ab8493"
< access-control-allow-origin: *
< content-type: text/html
< via: 1.1 varnish, 1.1 varnish
< accept-ranges: bytes
< date: Sat, 04 Sep 2021 10:08:10 GMT
< age: 3730
< x-served-by: cache-hel6831-HEL, cache-muc13921-MUC
< x-cache: HIT, HIT
< x-cache-hits: 1, 1
< x-timer: S1630750091.887080,VS0,VE0
< vary: Accept-Encoding
< content-length: 12348
< 
matthias@enterprise:
```

The URL change was issued about 5 months ago by: https://github.com/apache/netbeans/commit/026738caa88157117c9ac7828cc68cef7c23f5e1. For the usecase "Apache NetBeans *" I don't see a reason to change anything here. 

Testing with master showed, that connection checking works and I had a look with wireshark and at least judging from the SNI header of the TLS handshake the right host is addressed. Is this a fix, that is not needed?

neilcsmith-net:
@matthiasblaesing good point!  I only looked at the changes here briefly - more concerned about it being marked up for 12.5.

@joestr please check with the latest 12.5 beta and open an issue if you have problems there.

joestr:
It looks like this fix seems unnecessary if the URL has been changed already.

PR #3153, opened 2021-09-04
LSP Client Improvements (Foldmanager, Performance)

matthiasblaesing:
In the last few weeks I noticed two issues in the LSP Client code. One is located in the way background tasks
are queue. From my analysis identical background tasks can be schedules multiple times and executed quickly
ater one another, causing performance issues. The second is error reporting from the fold manager complaining
about being provided with identical folds multiple times.

Please refer to the individual commit messages for more information.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
@@ -432,14 +432,14 @@ public class LSPBindings {
     }
 
     private static void scheduleBackgroundTask(RequestProcessor.Task req) {
-        WORKER.post(req, DELAY);
+        req.schedule(DELAY);
     }
 
     public static synchronized void rescheduleBackgroundTask(FileObject file, BackgroundTask task) {
         RequestProcessor.Task req = backgroundTasksMapFor(file).get(task);
 
         if (req != null) {
-            WORKER.post(req, DELAY);
+            scheduleBackgroundTask(req);
         }
     }
 
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImpl.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImpl.java
@@ -20,7 +20,10 @@ package org.netbeans.modules.lsp.client.bindings;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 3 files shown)


ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
```diff
@@ -432,14 +447,19 @@ public void runOnBackground(Runnable r) {
     }
 
     private static void scheduleBackgroundTask(RequestProcessor.Task req) {
-        WORKER.post(req, DELAY);
+        synchronized(WORKER_SCHEDULED) {
+            if(! WORKER_SCHEDULED.containsKey(req)) {
+                WORKER_SCHEDULED.put(req, null);
+                WORKER.post(req, DELAY);
```
jlahoda, line 453: Hmmm, I apparently forgot how this works - I somehow managed to think `post` will defer the task when the same task is posted again, but that is wrong. But I don't think there is a need to keep track of the tasks manually - the correct idiom is (I hope I am right this time!) `req.schedule(DELAY);`. Should take care of the re-scheduling/moving to a new time. Should work similarly to the map (I hope!) but much simpler.

Sorry for the trouble.

ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
```diff
@@ -400,9 +403,17 @@ public InitializeResult getInitResult() {
         return initResult;
     }
 
-    @SuppressWarnings("AccessingNonPublicFieldOfAnotherObject")
+    @SuppressWarnings({"AccessingNonPublicFieldOfAnotherObject", "NestedSynchronizedStatement"})
```
jlahoda, line 406: Seems like a mistake in the hint, will try to take a look.

jlahoda (commented):
I think this is a correct direction, but `Task.schedule` should do the same, but should be easier. Sorry I messed this up.

ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
matthiasblaesing: You mean the hint is wrong because the synchronisation is inside a lambda, that is not executed in the containing method? My first though was, that the hint is right, but after reading twice, you are right.

With the next push, this will be gone again, as as your proposed solution is much easier and does not need explicit synchronisation.

ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
matthiasblaesing: I think the behavior of my proposed solution and the your suggestion is slightly different. 

From my observation (and my reading of the documentation of RequestProcessor), with RequestProcessor.Task#schedule, if an event occurs inside the 500ms timeframe, the execution of the background update will be postponed again for 500ms, if the the next event happens inside this 500ms frame, updates will be postponed again and so on. So the view is only updated, once the user stops changing things.

My solution would update after 500ms and again 500ms later.

I observed the behavior of the java editor and at least the navigator windows follows the principle described for RequestProcessor.Task#schedule and I never had a problem with that. 500ms seem to be short enough to be a good balance between wait time and resource usage.

I updated this PR accordingly.

matthiasblaesing:
If noone objects, I'll merge this by the end of the week.

PR #3155, opened 2021-09-05
Switch typescript and cpplite editors to multiviews

matthiasblaesing:
I noticed that the "History" tab was missing for typescript editor. At times I use it to access the history
of a single file. I think it is appropriate to use MultiViews for the typescript editor. As the typescript
and cpplite editors are from the same heritage, I adjusted them too.


cpplite/cpplite.editor/src/org/netbeans/modules/cpplite/editor/file/CDataObject.java
```diff
@@ -112,14 +116,27 @@
 @GrammarRegistration(grammar="resources/c.tmLanguage.json", mimeType=MIMETypes.C)
 public class CDataObject extends MultiDataObject {
 
+    static final String ICON = "org/netbeans/modules/cpplite/editor/file/resources/CSrcIcon.gif";
```
junichi11, line 119: Maybe, we can add `@StaticResource`.
Other files as well.

matthiasblaesing: Valid point - updated this accordingly.

matthiasblaesing:
If noone objects, I'll merge this by the end of the week.

junichi11 (approved):
Great! Looks good to me. 
(Nitpick: It would be nice if you could add screenshots for before and after.)

matthiasblaesing:
Screenshot after:

Screenshot

The highlighted area is the result of adding multiviews. The editor is the the base view, that was present before, while the history tab get added by this changeset.

junichi11:
Perfect! Thank you for it.

PR #3158, opened 2021-09-07
Performance optimizations to speed up remote debugging.

entlicher:
Remote debugging over slow connections (about 180ms to 200ms per single JDI call that involves JDWP communication over the globe) has a significant impact on debugger experience.

A speed up is achieved by removal of unused functionality, by caching and by parallelization of requests.

### DIFF
--- a/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/DeadlockDetectorImpl.java
+++ b/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/DeadlockDetectorImpl.java
@@ -34,15 +34,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+
 import org.netbeans.api.debugger.jpda.CallStackFrame;
 import org.netbeans.api.debugger.jpda.DeadlockDetector;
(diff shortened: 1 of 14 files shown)


dbalek (approved):

matthiasblaesing:
@entlicher this breaks the visualization of the debugging state in the IDE (and maybe other things not directly observable). I bisected the visual issue shown below to this commit.

Consider this simple programm:

### Problem

```java
public class Main {
        System.out.println("Line 1");
        System.out.println("Line 2");
        System.out.println("Line 3");
        System.out.println("Line 4");
    }   
}
```

I have a line break point on the line with "Line 2" in it. I exect the IDE to highlight that line when the break point is hit. I then expect the highlighting to move when I single step through the execution. This does not work anymore. With this merged:

Breakpoint hit:
image

One step further:
image


### How it should look

Breakpoint hit:
image

One step further:
image


Please have another look at this.

matthiasblaesing:
Control of break points when breakpoints on multiple threads are hit also does not seem to work anymore.

matthiasblaesing:
I now used current master, reverted this commit and the result is that single stepping works again.

PR #3159, opened 2021-09-09
Concentrating I/O access into JavaIoFile private class

JaroslavTulach:
In the quest to [make the Heap library work without file access](https://github.com/apache/netbeans/pull/3148) I am refactoring the code logic to be `java.io.File` and `java.io.RandomAccessFile` free. The commit a69dca3 creates two package private classes `File` and `RandomAccessFile` (naming is chosen to minimize the size of the diff) and moves all the I/O code into them. The rest of the code stays untouched as much as possible.

There are other two problems:
*  lib.profiler is calling into `SwingUtilities.invokeLater` - eliminated by ab8b1e5 - the idea is to keep compatibility of `HeapProgress`, but unless somebody (e.g. a Swing UI of profiler) calls it, the class is not going to initialize
*  lib.profiler is using `ex.printStackTrace()` - 1998eb2  solves it the same way as I/O access - e.g. by moving the calls into `Systems` package private class

### DIFF
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
@@ -43,14 +43,18 @@ enum Progress {
         }
     }
 
-    synchronized static void notifyUpdates(Handle h, int type) {
+    private synchronized static void notifyUpdates(Handle h, Type type) {
         for (Listener onChange : listeners) {
             switch (type) {
-                case 1: onChange.started(h); break;
-                case 2: onChange.progress(h); break;
+                case STARTED: onChange.started(h); break;
+                case PROGRESS: onChange.progress(h); break;
                 default: onChange.finished(h);
             }
         }
+    }
+
+    private enum Type {
+        STARTED, PROGRESS, FINISHED;
     }
 
     static interface Listener {
@@ -67,7 +71,7 @@ enum Progress {
 
         private Handle(Progress type) {
             this.type = type;
-            notifyUpdates(this, 1);
+            notifyUpdates(this, Type.STARTED);
         }
 
         void progress(long value, long endValue) {
@@ -83,14 +87,14 @@ enum Progress {
 
         @Override
         public void close() {
-            notifyUpdates(this, 2);
+            notifyUpdates(this, Type.FINISHED);
         }
 
         private void doProgress(long value, long startOffset, long endOffset) {
             this.value = value;
             this.endOffset = endOffset;
             this.startOffset = startOffset;
-            notifyUpdates(this, 1);
+            notifyUpdates(this, Type.PROGRESS);
         }
 
         long getValue() {
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.lib.profiler.heap;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+enum Progress {
+    COMPUTE_INSTANCES,
+    COMPUTE_REFERENCES,
+    FILL_HEAP_TAG_BOUNDS,
+    COMPUTE_GC_ROOTS;
+
+    Handle start() {
+        return new Handle(this);
+    }
+
+    private static List<Listener> listeners = Collections.emptyList();
+    synchronized static void register(Listener onChange) {
+        if (listeners.isEmpty()) {
+            listeners = Collections.singletonList(onChange);
+        } else {
+            List<Listener> copy = new ArrayList<>(listeners);
+            copy.add(onChange);
+            listeners = copy;
+        }
+    }
+
+    synchronized static void notifyUpdates(Handle h, int type) {
+        for (Listener onChange : listeners) {
+            switch (type) {
+                case 1: onChange.started(h); break;
+                case 2: onChange.progress(h); break;
+                default: onChange.finished(h);
+            }
+        }
+    }
+
+    static interface Listener {
+        void started(Handle h);
+        void progress(Handle h);
+        void finished(Handle h);
+    }
+
+    static final class Handle implements AutoCloseable {
+        final Progress type;
+        private long value;
+        private long startOffset;
+        private long endOffset;
+
+        private Handle(Progress type) {
+            this.type = type;
+            notifyUpdates(this, 1);
+        }
+
+        void progress(long value, long endValue) {
+            progress(value, 0, value, endValue);
+        }
+
+        void progress(long counter, long startOffset, long value, long endOffset) {
+            // keep this method short so that it can be inlined
+            if (counter % 100000 == 0) {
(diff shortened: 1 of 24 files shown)


sdedic (approved):
Curious what will be the next step...

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
```diff
@@ -0,0 +1,114 @@
...
+        @Override
+        public void close() {
+            notifyUpdates(this, 2);
+        }
+
+        private void doProgress(long value, long startOffset, long endOffset) {
+            this.value = value;
+            this.endOffset = endOffset;
+            this.startOffset = startOffset;
+            notifyUpdates(this, 1);
```
matthiasblaesing, line 99: I think reading the code in `notifyUpdates` this should be `notifyUpdates(this, 2)`

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
```diff
@@ -0,0 +1,114 @@
...
+        void progress(long counter, long startOffset, long value, long endOffset) {
+            // keep this method short so that it can be inlined
+            if (counter % 100000 == 0) {
+                doProgress(value, startOffset, endOffset);
+            }
+        }
+
+        @Override
+        public void close() {
+            notifyUpdates(this, 2);
```
matthiasblaesing, line 92: This should from my reading of `notifyUpdates` either be `notifyUpdates(this, 0)` or `notifyUpdates(this, 3)`

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
```diff
@@ -0,0 +1,114 @@
...
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
```
matthiasblaesing, line 24: This header needs to be the default ALv2 header

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,46 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.lib.profiler.heap;
+
+final class Systems {
+
+    //~ Static fields/initializers -----------------------------------------------------------------------------------------------
+    static final boolean DEBUG = false;
```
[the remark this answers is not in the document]
JaroslavTulach: That's not the same when the code gets compiled... `static final boolean` is treated as a constant by the C1/C2/Graal compilers and as such the whole  `if (DEBUG) { ... }` is completely left out of the compiled code (including the check). Using `LOG.isLoggable(...)` isn't constant and the check needs to be done all the time.

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,46 @@
...
+
+    static boolean isLinux() {
+        String osName = System.getProperty("os.name"); // NOI18N
```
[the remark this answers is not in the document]
JaroslavTulach: I am using the _heap library_ outside of NetBeans. Right now it is completely standalone. Adding dependency on `BaseUtil` would increase the [weight of this module](http://wiki.apidesign.org/wiki/HeavyWeight).

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,46 @@
...
+    static void printStackTrace(Throwable t) {
+        t.printStackTrace();
+    }
+
+    static void printStackTrace(String msg, Throwable t) {
+        System.err.println(msg);
+        t.printStackTrace(System.err);
+    }
+
+    static void debug(String msg) {
```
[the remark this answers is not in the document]
JaroslavTulach: Using `Logger` is good idea. It also [increases the weight](http://wiki.apidesign.org/wiki/HeavyWeight) as `java.util.logging` isn't part of `java.base` module, but be it: 938dcf0

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private Systems() {
+    }
+    
+    static void printStackTrace(Throwable t) {
+        t.printStackTrace();
+    }
+
+    static void printStackTrace(String msg, Throwable t) {
+        System.err.println(msg);
```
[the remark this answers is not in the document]
JaroslavTulach: Done in 938dcf0

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,46 @@
...
+    static boolean isLinux() {
+        String osName = System.getProperty("os.name"); // NOI18N
+        return osName.endsWith("Linux"); // NOI18N
+    }
+
+    private Systems() {
+    }
+    
+    static void printStackTrace(Throwable t) {
+        t.printStackTrace();
```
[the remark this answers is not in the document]
JaroslavTulach: Right. Done in 938dcf0

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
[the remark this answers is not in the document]
JaroslavTulach: In 938dcf0 I decided to keep the `DEBUG` final field, but to initialize it from `Logger.isLoggable` value. That will keep the compiled code the same and allow the logging be customized by [typical log configuration](http://bits.netbeans.org/dev/javadoc/org-openide-util/org/openide/util/doc-files/logging.html) methods.

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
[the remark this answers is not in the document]
JaroslavTulach: I rather keep the simple property check.

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
JaroslavTulach: Opps, thank you. Better to not be lazy: 01c307f

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
JaroslavTulach: Done in 01c307f

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Progress.java
JaroslavTulach: Of course. Updated.

JaroslavTulach:
> Curious what will be the next step...

a3162ad  is your answer. Full abstraction over `java.io.File` and `java.io.RandomAccessFile` with a pluggable `Factory` to specify when wiring `CacheDirectory` and `HprofHeap` together.

An additional package private class can re-implement `File.Factory` and fully replace the I/O access with something different without changing the existing logic of the heap dump processing.

profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/Systems.java
```diff
@@ -0,0 +1,47 @@
...
+ * under the License.
+ */
+package org.netbeans.lib.profiler.heap;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class Systems {
+    private static final Logger LOG = Logger.getLogger(System.class.getPackage().getName());
+    static final boolean DEBUG = LOG.isLoggable(Level.FINE);
```
matthiasblaesing, line 26: Configurable logging at runtime is one of the key benefits of the logging system to me. Yes there is overhead in the check at runtime, but so is overhead in virtual methods, the VM concept in itself, the interpreter, GC and so on.

I'm not religious here, but reading such code feels strange to me.

JaroslavTulach: Let's keep the spirit of the code for now. If needed `DEBUG` can be replaced with `LOG::isLoggable` later.

PR #3162, opened 2021-09-11
Remove JARs fom repository to simplify release review

matthiasblaesing:
- Mock jars in maven project used for unittests only contain a pom.properties
  file and are generated at test runtime
- empty.jar used in autoupdate services unittest is generated with other
  signed jars at test build time
- dummy jar files for tests of versioned jar suffix matcher of glassfish
  and payara tooling are generated at test runtime


jtulach (approved):

lkishalmi:
+1
Throw rocks at me, but I like this! 
Thank you!

matthiasblaesing:
I pushed another set of cleanups. Given that pack200 compression will go away and thus the NBMs become unreviewable and other code here was zipped "just because", I don't see this as a bad changeset. None the less, PR #3190 shows, that the cases are not clear and it is not clear, what "we want", so settings this as "do not merge".

To be clear: This is ready to merge, but I don't see consensus whether it should be.

PR #3163, opened 2021-09-12
Return a copy instead of cached array

sdedic:
During some implementation I've mindlessly modified the result of `ClassPath.getRoots()` call ... and then spent more than 5 hours by hunting down the bug since it wasn't originally obvious, but "sometimes" the parser stopped to resolving symbols against the classpath. "sometimes" = when the added (debugging) code run and modified the ClassPath caches for all IDE.
I think making the code fool-proof is worth the array-copy overhead...

### DIFF
--- a/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
+++ b/ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
@@ -241,7 +241,7 @@ public final class ClassPath {
         long current;
         synchronized (this) {
             if (rootsCache != null) {
-                return this.rootsCache;
+                return rootsCache.clone();
             }
             current = this.invalidRoots;
         }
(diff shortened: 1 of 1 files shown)


ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
```diff
@@ -241,7 +241,7 @@ public ClassPathImplementation getClassPathImpl(ClassPath cp) {
         long current;
         synchronized (this) {
             if (rootsCache != null) {
-                return this.rootsCache;
+                return Arrays.copyOf(rootsCache, rootsCache.length);
```
JaroslavTulach, line 244: I'd use `this.rootsCache.clone()`.

JaroslavTulach (approved):

ide/api.java.classpath/src/org/netbeans/api/java/classpath/ClassPath.java
sdedic: Good point; recommitted as 3fc65de. Thanks.

PR #3164, opened 2021-09-12
Add Collection/Map.size() > 0 case to isEmpty() refactoring.

mbien:
SizeEqualsZero perf inspection currently only triggers on size() == 0 and size() != 0. Added the case for size() > 0 which I sometimes encounter in the wild.
 + size() > 0 case is handled like the size() != 0 case in the UI (same check box)
 + updated the warning/fix/desc text accordingly and made it a little bit more dynamic


mbien:
the failed jdk8 + nbjavac build seems to be unrelated as far as i see.
https://app.travis-ci.com/github/apache/netbeans/jobs/536851479

jlahoda (approved):
Looks good, thanks!

PR #3165, opened 2021-09-12
Performance: groovy loads classes using FileSystems

sdedic:
This PR adds some performance counters and hacks that allow individual Groovy compiler's phases and visitors to be measured during parsing. Statistics are printed at the end of each parsing task and in case of indexing, summary and a per-file averages are printed after root's indexing completes.
The performance logs are enabled by setting `org.netbeans.modules.groovy.editor.compiler.PerfData` loglevel to `FINER` or more.

Next, the PR changes resource loading for Groovy:
- the resource is loaded using NB `ClassPath` (which uses FileSystems) rather than through `ClassPath.createClassLoader()`, which seems to be very slow.
- java classes are loaded from java indexer caches: this speeds up loading referenced classes from the project itself

As a result, the Groovy parsing is **about 5-10 times faster** (!), but there are still some Java parser invocations, so the overall indexer speedup is about 2-3x.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndexer.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndexer.java
@@ -18,7 +18,6 @@
  */
 package org.netbeans.modules.groovy.editor.api;
 
-import groovy.transform.PackageScope;
 import groovyjarjarasm.asm.Opcodes;
 import java.io.IOException;
 import java.net.URL;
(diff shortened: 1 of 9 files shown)


JaroslavTulach (approved):
Please make sure the new APIs are documented at https://bits.netbeans.org/dev/javadoc/properties.html - that probably means to include the groovy module in `nbbuild/build.properties` in the development javadoc section.

Consider documenting the loggers as (a private) API as well.

Update versions of the modules & dependencies to avoid ClassNotFoundErrors.

dbalek (approved):
Looks fine.

PR #3166, opened 2021-09-13
Updated Collection.toArray() refactoring

mbien:
The performance inspection currently advises to refactor `col.toArray(new T[0])` into `col.toArray(new T[col.size])` which is exactly the wrong way around for modern Java[1].

Updated it to refactor to `T[]::new` on JDK 11+ and `new T[0]` for older JDKs (determined by the project src lvl).

[1] https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
@@ -164,25 +164,34 @@ public class Tiny {
         final String literal = ctx.getInfo().getText().substring(start, end);
 
         Fix f = new JavaFix(ctx.getInfo(), toSearch) {
-@Override protected String getText() {
-return NbBundle.getMessage(Tiny.class, "FIX_LengthOneStringIndexOf");
-}
-@Override protected void performRewrite(TransformationContext ctx) {
-WorkingCopy wc = ctx.getWorkingCopy();
-TreePath tp = ctx.getPath();
-String content;
+            @Override
+            protected String getText() {
+                return NbBundle.getMessage(Tiny.class, "FIX_LengthOneStringIndexOf");
+            }
 
-if ("'".equals(data)) content = "\\'";
-else if ("\"".equals(data)) content = "\"";
-else {
-content = literal;
-if (content.length() > 0 && content.charAt(0) == '"') content = content.substring(1);
-if (content.length() > 0 && content.charAt(content.length() - 1) == '"') content = content.substring(0, content.length() - 1);
-}
+            @Override
+            protected void performRewrite(TransformationContext ctx) {
+                WorkingCopy wc = ctx.getWorkingCopy();
+                TreePath tp = ctx.getPath();
+                String content;
 
-wc.rewrite(tp.getLeaf(), wc.getTreeMaker().Identifier("'" + content + "'"));
-}
-}.toEditorFix();
+                if ("'".equals(data)) {
+                    content = "\\'";
+                } else if ("\"".equals(data)) {
+                    content = "\"";
+                } else {
+                    content = literal;
+                    if (content.length() > 0 && content.charAt(0) == '"') {
+                        content = content.substring(1);
+                    }
+                    if (content.length() > 0 && content.charAt(content.length() - 1) == '"') {
+                        content = content.substring(0, content.length() - 1);
+                    }
+                }
+
+                wc.rewrite(tp.getLeaf(), wc.getTreeMaker().Identifier("'" + content + "'"));
+            }
+        }.toEditorFix();
         
         String displayName = NbBundle.getMessage(Tiny.class, "ERR_LengthOneStringIndexOf", literal);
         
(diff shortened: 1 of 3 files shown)


mbien:
travis failures seem unrelated on this one too

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
```diff
@@ -337,11 +353,24 @@ public static ErrorDescription collectionsToArray(HintContext ctx) {
...
-            String fixDisplayName = NbBundle.getMessage(Tiny.class, "FIX_Tiny_collectionsToArray");
 
-            fixes = new Fix[] {
-                JavaFixUtilities.rewriteFix(ctx, fixDisplayName, ctx.getPath(), "$collection.toArray(new $clazz[$collection.size()])")
-            };
+            SourceVersion version = ctx.getInfo().getSourceVersion();
+            TreePath type = ctx.getVariables().get("$clazz");
+            String typeName = type.getLeaf().toString();
+            
+            if (Integer.parseInt(version.name().substring(version.name().indexOf('_')+1)) >= 11) {
```
jlahoda, line 361: Use of `SourceVersion`/`CompilationInfo.getSourceVersion()` is correct, but `SourceVersion` is `Comparable`, so can be compared without integer tricks. Please see method `unnecessaryTempFromString` in this class for an example.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
```diff
@@ -314,15 +323,22 @@ private static ErrorDescription enumHint(HintContext ctx, String baseName, Strin
...
-          enabled=false,
+          enabled=true,
           suppressWarnings="CollectionsToArray")
-    @TriggerPattern(value = "$collection.toArray(new $clazz[0])",
-                    constraints=@ConstraintVariableType(variable="$collection",
-                                                        type="java.util.Collection"))
+    @TriggerPatterns({
+        @TriggerPattern(value = "$collection.toArray(new $clazz[$collection.size()])",
+                        constraints = @ConstraintVariableType(variable="$collection", type="java.util.Collection")),
+        @TriggerPattern(value = "$collection.toArray(new $clazz[0])",
```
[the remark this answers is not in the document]
mbien: the thought process was as follows:
`list.toArray(String[]::new)` is cleaner and less hacky than 
`list.toArray(new String[0])` which looks like sub optimal code for some (although both have identical performance on modern JVMs - see updated PR text). IDEs also advised against it for a long time. Someone might add col.size() again not knowing the details and we are where we started.

The same inspection can be used to refactor from array-with-zero to array-with-new if the source lvl allows it, which should be best practice for 11+.

on 8 it refactors `new String[list.size()]` to `new String[0]`,
on 11+ it refactors both `new String[0]` and `new String[list.size()]` to `list.toArray(String[]::new)`.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: can i compare against 11 if NB is running on <11 without nb-javac? SourceVersion is part of the JDK if i see this correctly - the enum would not be there.

mbien: unnecessaryTempFromString compares against 5 which doesn't have the same problem since 8 is minimum required JDK version for NB.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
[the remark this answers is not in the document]
jlahoda: Right, sorry. The check is fine (we could split the checks into two method, but then the hint would need to be defined in a class not a method, so we would need to reorganize things here, and there's probably not a good reason for that).

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
jlahoda: Right again, but I think I'd still somewhat prefer to compare SourceVersions than play integer tricks. I'd try to create a static final field, initialize to SourceVersion.RELEASE_11 if it exists (`SourceVersion.valueOf("RELEASE_11")`), or null. Hopefully easier to find when we can depend on the new API.

Thanks.

mbien: I implemented it here locally, but i gotta say it still looks very hacky. Esp the 
`MODEL_VERSION_11 != null && MODEL_VERSION_11.compareTo(version) >= 0
The main issue i see is that it will spread, other hints would have to keep doing the same thing, there is no nice way to make utility method out of this, to keep the ugly part in one place.

I am glad that `Runtime.Version` uses ints to represent the feature version since this shows that enums are a bad idea for things like that (many other projects ran into the same trap).

what we actually want to write is
```
version.supports(11)
```
or even better
```
version.feature() >= 11
```
Thats it, no time traveling needed to figure out potential future enums.

what would you say to a reusable utility like:
 ```
    public static boolean supports(SourceVersion model, int featureVersionToCheck) {
        return model.ordinal() >= featureVersionToCheck;
    }
```
This should be forward compatible for a long time, until JDK hopefully adds `version.feature()` to make it compatible with their own Runtime.version().feature() before breaking the enum ordering.

matthiasblaesing:
Lets get this in - it was reviewed and review comments were addressed to my understanding.

PR #3168, opened 2021-09-15
Update NetBeans icon to match macOS 11 look and feel

oyarzun:
Update NetBeans macOS icon according to https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon

Screen Shot 2021-09-15 at 9 49 32 AM

Screen Shot 2021-09-15 at 9 44 46 AM


matthiasblaesing:
For the record: the icns file is in no way mac os specific. I have it integrated for example in gnome-shell. Given that everybody seems to want to have dark desktops, I'm surprised, that an icon, that works on both is put onto a light background and that is the future?

mcdonnell-john:
Is this something that should be raised on the mailing lists? To get the communities feedback on?

I have no issues approving this, I do however prefer the current icon style over the Apple style but ultimately not pushed either way.

oyarzun:
> For the record: the icns file is in no way mac os specific. I have it integrated for example in gnome-shell.
@matthiasblaesing  I do not believe this is correct, I think the icns file is only used on macOS. 

https://github.com/apache/netbeans/search?q=netbeans.icns

The gnome desktop entry uses the png

```
[Desktop Entry]
Encoding=UTF-8
Name=Apache NetBeans IDE 12.5
Comment=The Smarter Way to Code
Exec=/bin/sh "/home/oyarzun/netbeans-12.5/netbeans/bin/netbeans"
Icon=/home/oyarzun/netbeans-12.5/netbeans/nb/netbeans.png
Categories=Application;Development;Java;IDE
Version=1.0
Type=Application
Terminal=0
```

@mcdonnell-john I'll send an email to the mailing list for additional feedback.

oyarzun:
nb1
nb2
nb3
nb4

swpalmer:
Something like the last image, with a less drastic gradient, the dark side on the bottom, and a slight drop shadow for the NB logo.

oyarzun:
nb-ds

swpalmer:
That’s the one, fits in with the general look and feel of icons on macOS.  Beautiful!

mcdonnell-john:
That last one is the best one imo @oyarzun...  Still not overly fussed on changing the icon personally, but I do like that one better then the previous ones you've done :)

eirikbakke:
I think the logo has been squashed slightly in the proposed icons above; be sure to maintain the aspect ratio when resizing the art. Here's the logo on the NetBeans website overlaid 50% with the icon:

overlay_nbicon

oyarzun:
@eirikbakke I think you are correct. Here is an updated svg.
apache-netbeans-macos-ds2 copy

mbien:
the more this PR progresses the larger the icons get :)

eirikbakke:
Thanks! Will we still get the nice drop shadow as was in https://user-images.githubusercontent.com/2364606/133636825-ef6804ee-d4a5-4941-9821-5ce9151c2bb7.png ?

Could we see how the latest version looks in the Dock together with other standard MacOS apps?

oyarzun:
@eirikbakke added back the drop shadow.

apache-netbeans-macos-ds2 copy

Screen Shot 2022-06-30 at 5 15 44 PM

As seen in the screenshot of the dock, when NetBeans is minimized it uses the original icon since this is coming from the `MainWindow`. https://github.com/apache/netbeans/blob/f56685947798457737d8272b76b173577506b6d5/platform/core.windows/src/org/netbeans/core/windows/view/ui/MainWindow.java#L506-L523

eirikbakke:
Perhaps size the logo down a little bit, so that there's a bit of margin to the edge of the rounded square. See the other icons in the doc:

logosize

Ideally we'd get the same icon showing in minimized state. What happens if f.setIconImages is commented out in MainWindow, and the application is run from an application bundle? Does the main window and any dialog boxes then end up with the icon specified in -Xdoc:icon?

(We could add some MacOS specific behavior in MainWindow, probably enabled by an environment variable that's passed in from the same script that sets -Xdock:icon.)

mhalachev:
I agree with @eirikbakke that scaling down the logo would enhance its appearance, as it currently extends to the top and bottom of the icon. Additionally, maintaining a white background color would ensure consistency with both the splash screen and the about dialog, as in both cases the logo is presented on a white background with no effects applied.

Perhaps we could consider using the small side element as a reference for scaling? It might help maintain proportion and balance across the design:

scale

Here's how it would appear in the dock, alongside other icons:

dock

neilcsmith-net:
The installer changes here are effectively dead.  If this change goes in to the IDE, will need a matching PR to update https://github.com/apache/netbeans-nbpackage/tree/master/src/main/resources/org/apache/netbeans/nbpackage for the macOS installer.

mhalachev:
Thanks. I will prepare an appropriate PR for this with just one more tiny adjustment: aligning the logo to the icon grid, as both the inner and outer visual components fit perfectly.

icon-grid

PS: I'll provide an update once the icon set is ready, as it requires some adjustments for the smaller sizes. In the meantime, any other comments and suggestions are welcome.

mhalachev:
PR in is now available: https://github.com/apache/netbeans-nbpackage/pull/51

oyarzun:
Closing in favor of PR #7132 using @mhalachev's icon

PR #3171, opened 2021-09-16
Jackpot hint reload on save + run action registration + small cleanup

mbien:
NetBeans has some very well hidden features :) for example, I didn't know that you can run hint files - if you can find the aciton.

This PR
 * registers run file action on the hint file right click menu
 * reloads the inspection automatically when a hint file is saved (main reason why i started this effort)
 * hintfile editor shows now only errors as errors, rest is mapped to warning (if its enabled)
 * java 8 language level updates to the two packages I looked at (separate commits), e.g diamonds, lambdas try-with resource etc
   * contains trivial improvements, e.g refactoring to Deque interface, descending iterator instead of reversing collections etc
   * refactored while testing jackpot rules of course :)
 
feedback needed for UpdateGlobalHintsOnSaveTask (6d5ac68). This task has to run after the file is saved but the OnSaveTask happens before IO. The task would have to be enqueued somewhere, I just didn't know where (i am a bit rusty nb platform wise). Its a pure coincidence that enqueuing it to the EDT (which has to be done anyway) creates the right timing so that it works.


java/java.hints.declarative/src/org/netbeans/modules/java/hints/declarative/idebinding/UpdateGlobalHintsOnSaveTask.java
```diff
@@ -0,0 +1,73 @@
...
+                // todo: this reloads *all* global hints, which isn't optimal
+                SwingUtilities.invokeLater(() -> RulesManager.getInstance().reload());
+            }
+
+            @Override
+            public void runLocked(Runnable task) {
+                // the file isn't actually saved yet, so we delay the task a bit
+                // it seems to be working fine even without a delay due to the timing of the EDT
+//                CompletableFuture.runAsync(task, delay); // 9+
+                scheduler.schedule(task, 200, TimeUnit.MILLISECONDS);
```
matthiasblaesing, line 59: If a `RequestProcessor` is used, `RequestProcessor#post` can be used, which can then be used to cancel the task. Please Remove the commented code.

matthiasblaesing:
I left two inline comments, but how can this be tested? I build this PR branch and opened the `java.hints.declarative` module and opened one of the unittest `.hint' files and got:

```
	at org.netbeans.modules.java.hints.declarative.idebinding.HintsTask.computeErrors(HintsTask.java:102)
	at org.netbeans.modules.java.hints.declarative.idebinding.HintsTask.run(HintsTask.java:84)
	at org.netbeans.modules.parsing.impl.TaskProcessor.callParserResultTask(TaskProcessor.java:561)
[catch] at org.netbeans.modules.parsing.impl.TaskProcessor$RequestPerformer.run(TaskProcessor.java:786)
	at org.openide.util.lookup.Lookups.executeWith(Lookups.java:279)
	at org.netbeans.modules.parsing.impl.TaskProcessor$RequestPerformer.execute(TaskProcessor.java:702)
	at org.netbeans.modules.parsing.impl.TaskProcessor$CompilationJob.run(TaskProcessor.java:663)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

java/java.hints.declarative/src/org/netbeans/modules/java/hints/declarative/idebinding/UpdateGlobalHintsOnSaveTask.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Reloads the global hint if the file changed after a save action.
+ * @author mbien
+ */
+@MimeRegistration(mimeType=DeclarativeHintTokenId.MIME_TYPE, service=OnSaveTask.Factory.class, position = 3000)
+public final class UpdateGlobalHintsOnSaveTask implements Factory {
+
+//    private static final Executor delay = CompletableFuture.delayedExecutor(200, TimeUnit.MILLISECONDS); // 9+
+    private static final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(0);
```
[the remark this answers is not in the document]
mbien: thanks for taking a look mathias. I was sure there would be something better I could submit the task to - I will change it to RequestProcessor as suggested.

mbien:
> I left two inline comments, but how can this be tested? I build this PR branch and opened the `java.hints.declarative` module and opened one of the unittest `.hint' files and got:
> ...

i have some hint files here: https://github.com/mbien/jackpot-inspections 
README has a guide how to use hints.

I haven't encountered the NPE before. I take a look. On first glance it looks like my changes should not be the cause of it. All you did was to open a hint file?

java/java.hints.declarative/src/org/netbeans/modules/java/hints/declarative/idebinding/UpdateGlobalHintsOnSaveTask.java
mbien: done

mbien:
sorry had to force push because i rebased yesterday since the master was green for a short time. Only change is in UpdateGlobalHintsOnSaveTask.java

matthiasblaesing (approved):
Looks sane to me. @jlahoda could you please also have a look at this?

PR #3172, opened 2021-09-17
Apis nb125 (never merge)

ebarboni:
This is the snapshot for 12.5 from an attempt without getpeer removal. just to compare


ebarboni:
closing

PR #3175, opened 2021-09-17
Avoid ClassLoader resource lookups

sdedic:
I've missed yet another resource lookup through ClassLoader.  In the case of `NbClassNodeResolver`, alll resources were loaded by the parent (= `ClassLoaderSupport` created by `ClassPath`) of the GroovyclassLoader (= `ParsingClassLoader`) so the resource lookup through ClassPath's classloader is completely unnecessary. in addition, loadByClassLoading was set to false, so I've deleted the relevant part of the code.

i also completely disabled error failures in compilation phases; Groovy tends to bail out on trivial errors (like imported symbol not exists) and will skip later more interesting compilation phases which attribute the tree.


dbalek (approved):

sdedic:
Just unrelated `SdkSuiteTest` has failed; merging.

PR #3177, opened 2021-09-20
[NETBEANS-6023] - cleanup SQLParser.java warning errors

BradWalker:
I wanted to explore the SQL capabilities of Netbeans. As a result, I cleaned up some of the
warnings in SQLParser.jj and friends.

   [repeat] /home/bwalker/src/netbeans/ide/db.sql.visualeditor/src/org/netbeans/modules/db/sql/visualeditor/parser/SQLParser.java:75: warning: [rawtypes] found raw type: ArrayList
   [repeat] SelectNode SQLSelect() throws ParseException {ArrayList selctItemList=new ArrayList();
   [repeat]                                               ^
   [repeat]   missing type arguments for generic class ArrayList<E>


BradWalker:
> This should be save as the module is either used from within the tree or only from age old visualweb modules. The changes do not affect API, so get it in :-)

Hey @matthiasblaesing , thanks very much!

PR #3178, opened 2021-09-20
Update JNA and JNA platform to 5.9.0

matthiasblaesing:
sigtest data was regenerated and checked. The changes to the interfaces
in the platform are ok:

- additions to the interfaces just expose additional native methods
- incompatible changes are fixes for functions that were broken hard
  before the change and thus are either already broken or were not
  used


matthiasblaesing:
I intent to merge this shortly, please speak up if you want to object.

PR #3179, opened 2021-09-20
Update JavaFX to 17

matthiasblaesing:


mbien:
would be an opportunity to set the language level to 8 for this module :)

matthiasblaesing:
As main tests come back green and only groovy tests _all_ fail (and not only in this PR), I assume the errors are unrelated. I intent to merge this shortly. If anyone wants to object, please do so.

@mbien you are right, source level 8 is required for multi catch.

ebarboni (approved):
looks good to me

PR #3182, opened 2021-09-22
Do not persist breakpoints and watches when debugging through LSP server.

entlicher:
Debugger has it's own `Properties` API for settings persistence.
We use it in LSP to override the default for breakpoints and watches persistence.

### DIFF
--- a/ide/spi.debugger.ui/arch.xml
+++ b/ide/spi.debugger.ui/arch.xml
@@ -110,6 +110,15 @@ Registrations in standard Lookup:
 <li><api name="org.netbeans.spi.debugger.ui.BreakpointType.ContextAware.createService" group="lookup" type="import" category="official" url="@org-netbeans-api-debugger@"/> Loads all breakpoint type providers.</li>
 <li><api name="org.netbeans.modules.debugger.ui.registry.ColumnModelContextAware.createService" group="lookup" type="import" category="official" url="@org-netbeans-api-debugger@"/> Loads all column models registered for specific view (according to the path).</li>
 </ul>
+Persistent properties stored through org.netbeans.api.debugger.Properties:
+<ul>
+<li><api name="debugger" group="property" type="export" category="official" url="@org-netbeans-api-debugger@"/>The main namespace for debugger properties.</li>
+<li><api name="debugger.persistence" group="property" type="import" category="official" url="@org-netbeans-api-debugger@"/>The main namespace for debugger persistence settings. Used for configuration of individual services to determine what is to be stored persistently.</li>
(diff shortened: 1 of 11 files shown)


JaroslavTulach:
I am not familiar with the debugger `Properties` API - branding API (originally suggested to fix this problem) would look differently.

Anyway this is an API and I recommend to:
- version it
- document it in arch.xml
- document a change in apichanges
- write a test

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/LspDebuggerOptions.java
```diff
@@ -0,0 +1,52 @@
...
+    public String[] getSupportedPropertyNames() {
+        return new String[] {
+            PERSISTENCE_BREAKPOINTS,
+            PERSISTENCE_WATCHES,
+        };
+    }
+
+    @Override
+    public Object getDefaultPropertyValue(String propertyName) {
+        if (propertyName.startsWith("debugger.persistence")) {      // NOI18N
```
singh-akhilesh, line 45: Small suggestion: if else could be replaced by ternary in a single line --  return propertyName.startsWith("debugger.persistence") ? false : null;

entlicher: It's a matter of taste.

entlicher:
@JaroslavTulach I've added the suggested API description and test.

ide/spi.debugger.ui/arch.xml
```diff
@@ -110,6 +110,15 @@ Registrations in standard Lookup:
 <li><api name="org.netbeans.spi.debugger.ui.BreakpointType.ContextAware.createService" group="lookup" type="import" category="official" url="@org-netbeans-api-debugger@"/> Loads all breakpoint type providers.</li>
 <li><api name="org.netbeans.modules.debugger.ui.registry.ColumnModelContextAware.createService" group="lookup" type="import" category="official" url="@org-netbeans-api-debugger@"/> Loads all column models registered for specific view (according to the path).</li>
 </ul>
+Persistent properties stored through org.netbeans.api.debugger.Properties:
+<ul>
+<li><api name="debugger" group="property" type="export" category="official" url="@org-netbeans-api-debugger@"/>The main namespace for debugger properties.</li>
```
JaroslavTulach, line 115: Great!

JaroslavTulach (approved):

PR #3184, opened 2021-09-22
Must reach ELEMENTS_RESOLVED before working with java els

sdedic:
Even after recent delivery merge, I've found some more places where I should call .toPhase().


JaroslavTulach (approved):
I approve anything you wish.

dbalek (approved):

PR #3186, opened 2021-09-23
LSP: Test results displayed using the new TestExplorer API.

dbalek:
The new VSCode's API used to display test results.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
```diff
@@ -508,7 +508,7 @@ public void finished(boolean success) {
 
         if (provider == null) {
             command = debug ? mainSource ? ActionProvider.COMMAND_DEBUG
-                                         : ActionProvider.COMMAND_DEBUG // DEBUG_TEST is missing?
+                                         : ActionProvider.COMMAND_TEST //TODO: COMMAND_DEBUG_TEST is missing?
```
JaroslavTulach, line 511: This change isn't very satisfying, right? Instead of being able to debug something, we run tests with debugging disabled.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TestSuiteInfo.java
```diff
@@ -108,23 +109,23 @@ public void setFile(final String file) {
     }
 
     /**
-     * The line within the specified file where the suite definition starts (if known).
+     * The range within the specified file where the suite definition is located (if known).
      */
     @Pure
-    public Integer getLine() {
-        return line;
+    public Range getRange() {
```
JaroslavTulach, line 115: Is this class a base for a JSON message transferred between LSP backend and LSP client? If so, than this classifies as an API change. How are we going to document/version this kind of API?

java/java.lsp.server/vscode/package.json
```diff
@@ -4,7 +4,7 @@
 	"description": "Apache NetBeans Language Server Extension for Visual Studio Code",
 	"author": "Apache NetBeans",
 	"license": "Apache 2.0",
-	"version": "12.4.0",
+	"version": "12.5.0",
```
JaroslavTulach, line 7: Please put hire `0.1.0` otherwise we are going to have changes with every local `ant build-vscode-ext` - don't you see `0.1.0` automatically being written into this location with every build?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TestSuiteInfo.java
dbalek: The API has not been marked stable or official yet. It is an extension to the LSP used by our VSCode plugin (sources of the plugin are part of the same module) as a client and it is currently work under development. Agree, the API should be documented once finalized.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchDelegate.java
dbalek: The specific combination of the method parameters should lead to running of all project tests under a debugger. However, since the `COMMAND_DEBUG_TEST` action is no available yet (it should be implemented in a separate PR) it it IMHO better to run all tests without debugger than to run the project under debugger.

java/java.lsp.server/vscode/package.json
dbalek: Done

JaroslavTulach (approved):

PR #3187, opened 2021-09-23
ant clean fix for javacc java file

ebarboni:
Was compiling on a very very old clone and java file generated from javacc were old 3.2 and not deleted.
Should be more rebuildable


matthiasblaesing (approved):
Looks good to me.

BradWalker:
> Looks good to me.

I agree.. I was just looking at "why this is" as a result of some of my cleanup work..

PR #3190, opened 2021-09-24
[NETBEANS-6044] Maven looks for annotation processor classpath in compiler plugin's configuration.

sdedic:
The reason for [NETBEANS-6044](https://issues.apache.org/jira/browse/NETBEANS-6044) is that Maven returned just compilation classpath for the "classpath/processor". In fact, the `maven-compiler-plugin` can specify path to look for annotation processors and if it does, only that path should be used (see [annotationProcessorPaths](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html)).

This PR causes Maven projects to return a separate Classpath, which is either `classpath/compile`, if the plugin does not configure any annotation processor paths, or that explicit setting. Since the plugin can be configured differently for different goals, test sources may get its own annotation processor path - this is also covered.
The impl should adapt the classpath if the POM file changes (i.e. the `annotationProcessorPaths` element is removed / added / changed) and the processor classpath reports broken artifact references.

As part of the process, I've added a new API helper method to `PluginPropertyUtils` since getting artifact path + dependencies is not that trivial as it seemed to initially.


jtulach (approved):
Amazing tests!

matthiasblaesing:
Could you please have a look at this PR: https://github.com/apache/netbeans/pull/3162 it is in a state where all NBM and JAR files are removed from repository and created at build time. Could this approach also work for you here:

https://github.com/apache/netbeans/blob/cb4f276304e3b08c58074fe7a51e7eda212fd193/platform/autoupdate.services/build.xml#L55-L79

The idea: The contents of the JAR/NBM is stored in exploded form and the NBM/JAR is build at test build test time. This seems to work nicely for the `autoupdate.services` module. There the problem is even deeper as nested structures are needed. The nested structures are simulated by ".jardir" directories inside the parent NBM directories.

If that is not acceptable, I think we need to have a discussion whether or not the fuzz about jar files is ok or not.

matthiasblaesing:
Something is strange here, I ran the unittests and got failures for the new tests. These are the stack traces for the failures:

```
	at org.netbeans.modules.maven.classpath.ClassPathProviderImplTest.testBrokenAnnotationPath(ClassPathProviderImplTest.java:472)
	at java.lang.Thread.run(Thread.java:748)
	at org.netbeans.modules.maven.classpath.ClassPathProviderImplTest.testExplicitAnnotationClasspath(ClassPathProviderImplTest.java:440)
	at java.lang.Thread.run(Thread.java:748)
	at org.netbeans.modules.maven.classpath.ClassPathProviderImplTest.testDifferentAnnotationClasspathForTests(ClassPathProviderImplTest.java:498)
	at java.lang.Thread.run(Thread.java:748)
	at org.netbeans.modules.maven.classpath.ClassPathProviderImplTest.testDefaultAnnotationClasspath(ClassPathProviderImplTest.java:410)
	at java.lang.Thread.run(Thread.java:748)
	at org.netbeans.modules.maven.classpath.ClassPathProviderImplTest.testChangeProcessorsFromImplicitToExplicit(ClassPathProviderImplTest.java:546)
	at java.lang.Thread.run(Thread.java:748)
```

I tested with JDK 8 and JDK 11 and it ssems travis agrees to (I ignore the groovy tests, as they all seem to fail since a few weeks).

sdedic:
@matthiasblaesing thanks, will do a clean checkout and investigate.

matthiasblaesing:
With the updated push the build failures are gone. Thank you.

This is a suggestion how to create the repo structure at runtime and thus not require JARs in the repository: https://github.com/matthiasblaesing/netbeans/commit/487e8cd3540f07db80f4a4a04edae71c4345b58b. In the sample commit only the repo folder is removed, but the test source code apart from `test-app` could be removed too.

sdedic:
@matthiasblaesing I have quite a preference for maintainability: java code that generates xml code that represents maven repository ... is really awkward. So let's just zip the content at build time, if JARs are that bad. The metadata in `repo` were generated by a real maven (`mvn install`).

I wonder to what length with hard-to-maintain code the anti-jar effort will go: what about a PDF test data ;) for example ?

sdedic:
Great. Instead of "just" exploding an archive build by an external tool (mvn install), one now need to edit/append licenses manually. All this to satisfy formalism of 'not having binaries'. But still better than to replicate "necessary subset" (which can change) of that tool's logic in a testsuite, which is harder to read and track changes that the exploded archive.

We **should** bring the topic to the mailing list, to get a consensus on where the 'binary' boundary is effective.

neilcsmith-net:
I'm with @sdedic - I don't like special casing JARs for the sake of it. And if we are exploding them, far prefer exceptions for completely generated code.  Not sure the license header even makes sense on such files.  +1 to mailing list topic!

matthiasblaesing:
There are tons of generated code in the unittests - from my POV having explicit test data is more an exception than the norm.

Another aspect: JARs can become unmaintainable. The NBMs (essentially JARs under a different name) in the autoupdate.service module are pack200 compressed. There is a realistic chance, that in a few months many people will not be able to work with these files and thus can't check them. If the files are generated at build time, we at least get the error early.

I think the exploded JAR and/or the generated at runtime approach both work.

Two final statements:
- I did not ask for work I was not willing to do!
- I do not insist on this, if no one wants the work, fine, then I'll delete the branch and remove #3162. BUT! Then make the useless discussion stop.

ebarboni:
I agree this should be discussed, I've seen openwhisk done such a removal and others like maven/archiva still having jar for their tests (some voluntary wrong for testing purpose).
My personal logic is if we should'nt have binary in sources why do should we keep them.

sdedic:
Well, I see the effort as partially valid and - more importantly - I want to get the **functional** changes in this PR into `master`. As for JARs, if it is the _contents_ of them what's the subject of the test, then keeping the exploded content in Git is good - changes are well tracked. That's why I've adapted this PR to https://github.com/apache/netbeans/pull/3162

Note: I had to change `MANIFEST.MF` line endings to LF, which is required by `nbbuild/travis/check_line_endings.sh` ... but that's wrong for MANIFEST.MF format - it must use CRLF. This is a price for keeping jar exploded....  the MANIFEST parsing is ikely broken; but it works for these specific tests.

If anyone objects against the current PR state (the exploded Jars), please say so - I'll separate the PR into two to get the functional part to the `master` soon.

JaroslavTulach (approved):

PR #3191, opened 2021-09-25
Prevent NullPointerException in CPP LanguageServerImpl

matthiasblaesing:
There are various scenarios where the start routine for the language
server could result in a null value. If on that value Pair#second is
called, this results in a direct NullPointerException being raised.

The value needs to be checked first before the #second invocation is
done.


jlahoda (approved):
Looks good, thanks!

PR #3192, opened 2021-09-26
Better implementation of PSR-4 Namespace rules

aleferri:
Consider src to be the base directory (PSR-4 definition) and the "surplus" leading segments as parts of the namespace prefix.
Implementation is rough. For it to be better, i would need to add a configuration options in the project to let the user set the namespace prefix; fetch the configuration in this Checker and remove the appropriate leading part of the namespace. Unfortunately i don't know how to do that.


junichi11:
Can you submit it to the JIRA and add an example project there?

aleferri:
Did create the issue as improvements
https://issues.apache.org/jira/browse/NETBEANS-6180

junichi11:
@aleferri I'm sorry I'm late... Maybe, I overlooked the notification.

`src` is hardcoded. So, I think we should get the source directory.
Unit tests are missing.

php/php.editor/src/org/netbeans/modules/php/editor/verification/PSR4Hint.java
```diff
@@ -102,6 +102,9 @@ public void visit(NamespaceDeclaration node) {
                         if (CancelSupport.getDefault().isCancelled()) {
                             return;
                         }
+                        if (currentDir.getName().equals("src")) { // PSR-4 require corrispondence to the directory structure only for the contiguos subnamespaces
```
tmysik, line 105: Typos in the comment.

tmysik:
Agreed, some unit test(s) would be fine. Thanks.

aleferri:
@junichi11 How do i get the environment inside the Hint? My original idea was to search for the top folder, but i couldn't figure how to get objects outside of the current file.

junichi11:
I would the the source directory like the following:

```java
PhpModule phpModule = PhpModule.Factory.forFileObject(fileObject);
if (phpModule != null) {
    FileObject sourceDirectory = phpModule.getSourceDirectory();
}
```

junichi11:
BTW, you should create a PR with not the master branch but the new branch.

tmysik:
@junichi11 Good catch.

@aleferri Please, create a branch in your clone, containing the number of the ticket and ideally also a few words, from the ticket title, typically (e.g. `1234-fix-npe`).

Thanks.

aleferri:
Ok, i am going to open a new PR then

aleferri:
Sorry to ask, but how i can run the tests? -Dcluster.config=php build netbeans, but the tests didn't run.

matthiasblaesing:
Either from the UI Kontextmenu of the project or on CLI: `ant -f php/php.editor/build.xml test`.

aleferri:
Tests goes in timeout

matthiasblaesing:
@aleferri which tests timeout? Your local run? Which test.

With flaky tests I regularly run the tests before my change, note which fail and run after. I consider my changes "sane" if the same set of tests succeeds and fails (and yes in perfect world this would not be necessary, but...)

neilcsmith-net:
@junichi11 @tmysik marked stale and removed from NB16 milestone for now. Please follow up as you see fit - re-add, re-review or close.

tmysik:
It seems to me that (all?) the comments are still valid.

aleferri:
The implementation doesn't work, as it simply disable the warning even when the thing is wrong. I did that test by building the whole thing and opening my test project and also a should-fail one (aka i did a manual test).
Unfortunately i am out of ideas on how to fix it and automated tests still don't run (not even one of them, all of them go in timeout).

neilcsmith-net:
I hadn't noticed how small the actual proposed changes were either - busy going through closing or marking stale PRs with milestones that hadn't been completed.  Closing for discussion elsewhere then.

junichi11:
Please submit it as a new PR with a new branch. Thanks!

PR #3193, opened 2021-09-26
[NETBEANS-6065] Make Gradle Project test close connections to Gradle Daemon

lkishalmi:


extide/gradle/src/org/netbeans/modules/gradle/GradleProjectConnection.java
```diff
@@ -111,6 +111,10 @@ public synchronized void close() {
         compatConn = null;
     }
 
+    boolean hasConnection() {
```
eirikbakke, line 114: If this class is meant to be thread-safe, which seems to be the case, then this method should be synchronized. Though it probably doesn't matter for the curent use, which is just for the single test class.

extide/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectImplTest.java
```diff
@@ -49,6 +52,25 @@ public NbGradleProjectImplTest(String name) {
...
+    
+    @After
+    @Override
+    public void tearDown() throws Exception {
+        ProjectConnection pconn = prj.getLookup().lookup(ProjectConnection.class);
+        if (pconn instanceof GradleProjectConnection) {
+            GradleProjectConnection gpconn = (GradleProjectConnection) pconn;
+            gpconn.close();
+        }
+        super.tearDown();
```
eirikbakke, line 72: Might be good to set prj = null for symmetry with setUp. But shouldn't matter.

extide/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectImplTest.java
```diff
@@ -220,18 +244,18 @@ public void testIncreaseAimedQualityChangesProject() throws Exception {
...
         assertTrue(ngp.getQuality().worseThan(NbGradleProject.Quality.EVALUATED));
-        
+        assertHasNoConnection(prj);
         NbGradleProjectImpl prjImpl = prj.getLookup().lookup(NbGradleProjectImpl.class);
         GradleProject curProject = prjImpl.getGradleProject();
         prjImpl.setAimedQuality(Quality.EVALUATED);
         GradleProject newProject = prjImpl.getGradleProject();
         assertTrue(newProject.getQuality().worseThan(NbGradleProject.Quality.EVALUATED));
         assertSame(newProject, curProject);
+        assertHasNoConnection(prj);
```
eirikbakke, line 258: What kinds of actions can cause the project to have a connection? I assume there is at least one test that causes hasConnection to become true, since it was necessary to add logic to close the connection in tearDown? If the latter is true, does this mean we are now depending on the various test methods to execute in a specific order, so that the connection is only made later?

eirikbakke (commented):
This PR contains changes to testing code only, so should be low-risk.

My main question: Is it ever expected that hasConnection will turn true at some point during NbGradleProjectImplTest? If so, is it valid to call assertHasNoConnection(prj) from various methods, since the prj object is now shared between test methods?

extide/gradle/test/unit/src/org/netbeans/modules/gradle/NbGradleProjectImplTest.java
lkishalmi: Well certain operation shall not create any connections. These are just checkpoints, to fail the test if that would ever happen.

sdedic:
@lkishalmi great !

PR #3194, opened 2021-09-26
NETBEANS-5722: Added `com.sun.java.swing.plaf.motif` to exported modules

asbachb:
The module is not exported by OpenJDK. Without the manual export NetBeans is unable to invoke `MotifFileChooserUI.createUI(JComponent)` and falls back to `MetalFileChooser` which is also unable to create as the Look and Feel for Metal was not initialized.


matthiasblaesing:
Looks sane to me.

PR #3196, opened 2021-09-27
LSP: various refactor bugfixes

dbalek:
* GR-33924: Code completion partially broken in method bodies in latest builds.
* GR-32319: Hover failed: ClassCastException: FilterDocument cannot be cast to LineDocument.
* GR-33740: Using .class on a type completes with null on the end.
* GR-31747: JUnit tests that fail to initialise lead to false success state.
* GR-33741: Quick fix for checked exception should not add java.util.logging.
* GR-33804: Null related message in Extract Superclass. 
* GR-33785: Try/catch block with resources.
* GR-33778: Can't rename constructor.
* GR-33807: Generate toString method default template.
* GR-33789: VSCode is accepting moving inner class inside itself.

### DIFF
--- a/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
+++ b/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
@@ -148,7 +148,6 @@ import org.netbeans.modules.java.hints.infrastructure.JavaErrorProvider;
 import org.netbeans.modules.java.source.BootClassPathUtil;
 import org.netbeans.modules.parsing.impl.indexing.implspi.CacheFolderProvider;
 import org.netbeans.spi.java.classpath.ClassPathProvider;
-import org.netbeans.spi.java.classpath.PathResourceImplementation;
 import org.netbeans.spi.java.classpath.support.ClassPathSupport;
 import org.netbeans.spi.java.queries.AnnotationProcessingQueryImplementation;
 import org.netbeans.spi.lsp.ErrorProvider;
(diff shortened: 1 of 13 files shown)


java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -2212,7 +2209,7 @@ public void logMessage(MessageParams arg0) {
         assertEquals("\n" +
                      "    @Override\n" +
                      "    public void run() {\n" +
-                     "        throw new UnsupportedOperationException(\"Not supported yet.\"); //To change body of generated methods, choose Tools | Templates.\n" +
+                     "        throw new UnsupportedOperationException(\"Not supported yet.\"); // Generated from nbfs://nbhost/SystemFileSystem/Templates/Classes/Code/GeneratedMethodBody\n" +
```
JaroslavTulach, line 2212: How comes this hasn't shown up before?

JaroslavTulach (approved):

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
dbalek: Good question. I would expect the `ServerTest` to be falling since https://github.com/apache/netbeans/commit/3db24dc9aea3c76aed71efb129e9af2e7b345c8e.

java/java.lsp.server/vscode/src/testAdapter.ts
```diff
@@ -52,6 +53,7 @@ export class NbTestAdapter {
     async run(request: TestRunRequest, cancellation: CancellationToken): Promise<void> {
         cancellation.onCancellationRequested(() => this.cancel());
         this.currentRun = this.testController.createTestRun(request);
+        this.itemsToRun = new Set();
```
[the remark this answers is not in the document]
dbalek: This change is related to GR-31747. If there is an exception thrown form a test suite initialization, the new VSCode explorer marks the corresponding suite node as erroneous, however individual test nodes below the suite nodes display the state from the tests last run. That could lead to false success state. It is IMHO better to marked these test as 'skipped' (they were not executed due to exception in the suite initialization).

PR #3197, opened 2021-09-29
Fix test failures in Groovy.editor

sdedic:
Merge of #3175 has introduced bugs - test failures. It turned out that in that PR merge process, I didn't check properly Travis :-/
So this PR:
* only suppresses errors after parsing phase. There is an 'error recovery' that handles errors thrown from the parsing. This bug failed some tests.
* updates `BookmarkController.groovy.indexed` golden file: originally, the compiler failed on non-existent imported symbol and did not proceed to phase 6-7 where accessor methods are generated
* used `IndexingTask` marker API to indicate indexing is in progress; groovy.editor will not run static analysis that may produce additional errors: this is the original behaviour.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/parser/GroovyParser.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/parser/GroovyParser.java
@@ -528,6 +528,9 @@ public class GroovyParser extends Parser {
         @Override
         public void gotoPhase(int phase) throws CompilationFailedException {
             super.gotoPhase(phase);
+            if (phase > Phases.CONVERSION) {
+                ((NbGroovyErrorCollector)errorCollector).setDisableErrors(true);
+            }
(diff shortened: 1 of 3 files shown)


ebarboni:
😍 thanks in advance, was looking for it but no idea what to do.

JaroslavTulach (approved):

ebarboni:
seems a last test is failing. 

Testcase: testIndex1(org.netbeans.modules.groovy.editor.api.GroovyIndexerTest):	FAILED
    [junit] Content does not match between 'testfiles/BookmarkController.groovy' and testfiles/BookmarkController.groovy.indexed'

sdedic:
OK ... I have such a mess on local from all the groovy experiments  that I'll do a clean checkout ;)

jlahoda:
+1, thanks for doing this!

ebarboni:
Green light, will merge later today if nobody beat me.

PR #3198, opened 2021-09-29
NETBEANS-5971 support Maven Wrapper (mvnw) in projects.

neilcsmith-net:
Search for existence of mvnw(.cmd) in project basedir or a parent project basedir and prefer this over embedded / specified Maven if present.  Added global setting and UI mirroring Gradle setting for this.

Trying to keep this as a light touch implementation for now.  No UI to configure wrapper yet. Longer term should probably consider replacing the built-in experimental support for downloading Maven versions with UI to use the wrapper?

Work in progress - not tested on Windows yet!


neilcsmith-net:
OK, I think this is good for review.  Have tested locally with single and multi module projects, using Maven Enforcer to verify correct Maven version used.  Not sure if there's a useful way to add tests for this feature?

sdedic (approved):
Excellent !

PR #3199, opened 2021-09-30
NETBEANS-6080: NoSuchMethodError thrown while applying CreateClassFix.

dbalek:
`ErrorType` constructor signature in vanilla javac differ from the corresponding signature in nb-javac. Temporary fix until fixed in nb-javac.


jlahoda (approved):
Looks OK.

JaroslavTulach:
Switching to automatically generated [JDK17+35](https://github.com/JaroslavTulach/nb-javac/tree/jtulach/UseJdk17) would also fix this signature error. Having 1:1 signatures with latest (e.g. JDK17) javac would in general be preferable.

PR #3201, opened 2021-09-30
Enable antialiasing for test progressbar

matthiasblaesing:
Before (Only the progressbar is changed, see the text "Test passed: 100,00%"):

Bildschirmfoto von 2021-09-30 19-40-22

After:

Bildschirmfoto von 2021-09-30 19-44-32


eirikbakke (approved):
Looks good! There's another issue with this component and HiDPI screens (https://issues.apache.org/jira/browse/NETBEANS-4198). I can look into that one after you merge this PR.

eirikbakke:
I opened another PR at https://github.com/apache/netbeans/pull/3203 that also handles the HiDPI issue, by using the same kind of text outline painting that was added to HeapView a while back, in place of the ImageBuffer based method.

matthiasblaesing:
Closing in favor of #3203

PR #3202, opened 2021-09-30
Selected range is ignored for CSS reformatting

matthiasblaesing:
Without this change the hole file is always formatted and a selected
range is ignored.


matthiasblaesing:
The test failure in the gradle.java testcase makes no sense to be connected to this changeset, so will merge this shortly.

ebarboni:
Hi, this looks good

PR #3203, opened 2021-09-30
[NETBEANS-4198] Fix anti-aliasing and HiDPI scaling in the Test Results progress bar.

eirikbakke:
A follow-up on https://github.com/apache/netbeans/pull/3201 which also fixes HiDPI scaling issues in the "Test Results" progress bar.

Before/after screenshots:

Test Results painting
:


matthiasblaesing (approved):
Looks sane to me. Thank you.

eirikbakke:
The only remaining test seems unrelated (gradle stuff)... merging. Thanks for digging up the location of this bug!

PR #3204, opened 2021-10-01
Performance optimizations to speed up remote debugging.

entlicher:
This contains the https://github.com/apache/netbeans/pull/3158 + corrections.
The original PR was reverted, thus here it is again in the first commit + corrections in the second commit.
The corrections are:
- corrected parallel threads filtering, where the original `threads.set(i, thread);` caused `IndexOutOfBoundsException`
- moved the VSCode-specific layer location out of IDE scope.


matthiasblaesing:
Not a review, but given that I noticed the problem in the first iteration, I gave it a spin. I ran a simple single stepping example and also tested parallel threads. Breakpoints were correctly hit, the break point location was shown in the editor and switching between threads worked as it should. The things I noticed to be broken in the previous iteration look fixed. Thank you.

entlicher:
Thanks Matthias for your verification of the corrected change.

dbalek (approved):
Looks fine

matthiasblaesing:
Debugging is broken again - another corner case I suppose. Take this trivial code:

```java
import java.util.ArrayList;
import java.util.List;

public class ThreadingTest {

	List<Thread> ts = new ArrayList<>();

	for(int i = 0; i < 10; i++) {
	    ts.add(new T());
	}
	
	for(Thread t: ts) {
	    t.start();
	}
    }

    static class T extends Thread {

	@Override
	public void run() {
	    System.out.println("L1");
	    System.out.println("L2");
	    System.out.println("L3");
	    System.out.println("L4");
	}

    }
}
```

Place a beakpoint on line 24 (the one, that outputs "L3"). Configure the breakpoint to suspend all threads:

Bildschirmfoto von 2021-10-11 20-55-56

 and run it with "Debug file". The debugging control does not work anymore:

Bildschirmfoto von 2021-10-11 20-56-41

At the time of the screenshot I would have expected line 24 to be highlighted, the debugging tab showing three threads (the first started, the cleaner and the main thread). I would expect all threads to be suspended. I would expect the "Playbutton" to become response. I would expect the stop button to stop the debugging session and let the programm run to completion.

Nothing of this works. In conclusion: Debugging with "Suspend all threads" is broken.

How shall we go on? Shall I directly create a revert commit or can this be fixed quickly?

PR #3207, opened 2021-10-03
Netbeans 5956: Adding default java trust store to xml retriever

asbachb:
The current implementation does not include the default java trust store to validate https connections. This results into a popup asking the user to if he trusts the certificate.

This change adds the default java trust store to the validation logic. In addition some code cleanup were performed.


ide/xml.retriever/src/org/netbeans/modules/xml/retriever/impl/SecureURLResourceRetriever.java
```diff
@@ -119,45 +139,48 @@ public void checkServerTrusted(X509Certificate[] certs, String authType)
...
-        } catch (java.security.GeneralSecurityException e) {
+            con.setHostnameVerifier(this::acceptAllHosts);
+        } catch (GeneralSecurityException e) {
             ErrorManager.getDefault().notify(e);
         }
     }
-    
+
+    private boolean acceptAllHosts(String host, SSLSession sslSession) {
+        return true;
```
matthiasblaesing, line 180: This looks dangerous (and yes I know, that you did not write the code). If I read the new code correctly, it will try to verify the certificate with the system truststore. If that works, we are good. If not the user is prompted and if accepts, the certifcate is stored to be accepted list and user is not asked again.

With the certificate verified, it needs to be tied to the hostname - this is a different issue and can't be done on the TrustManager level. Am I right that with the current implementation just accepts any certificate <-> hostname combination?

asbachb: Actuall I wonder if it can be removed completely. I suspect that the validation should be done by the default SSL/TLS processing.

matthiasblaesing: Depends on which certificates you want to support. There are:

1. trusted certificates with the CN or SAN holding the right hostname/IP
2. trusted certificates used with the wrong hostname
3. untrusted certificates, whose CN or SAN entries match the hostname
4. untrusted certificates, whose CN or SAN entries do not match the hostname

Without the hostname checker configured, the system only accepts cases 1+ 3, with the Hostname checker accepting all certificates also allows cases 2+4 to work (though is less save).

Slightly not directly related to this, maybe it is time, to create a platform module, that adds certificate handling to the IDE. I created this Plugin in the past:

https://plugins.netbeans.apache.org/catalogue/?id=20
https://github.com/nb-ldap-explorer/nb-ldap-explorer/tree/master/ssl_certificate_exception

https://github.com/nb-ldap-explorer/nb-ldap-explorer/blob/master/ssl_certificate_exception/src/main/java/com/google/code/nb_ldap_explorer/ssl_certificate_exception/TrustManagerImpl.java
https://github.com/nb-ldap-explorer/nb-ldap-explorer/blob/master/ssl_certificate_exception/src/main/java/com/google/code/nb_ldap_explorer/ssl_certificate_exception/HostnameVerifierImpl.java

asbachb: @matthiasblaesing  Maybe it makes more sense to open up a up a follow up bug to change this behavior?

asbachb:
@matthiasblaesing Surely I agree. Happy to contribute to my favourite IDE.

PR #3208, opened 2021-10-04
[jackpot] Hints SPI code cleanup / warning removal

mbien:
compiler warning fixes and some low hanging fruit refactorings. Only a handful of warnings remain for this module.

Its going to be easier to review the commits one by one. I left a few lambdas out which would create unnecessarily large diffs for little gain.

pinging @BradWalker to make sure we don't work on the same module :)


mbien:
travis has a few builds failing. hard to tell if it has anything to do with this PR.

matthiasblaesing:
Only unittests for java.mx.project error and I saw that also happen in other cases. This has two positive reviews and I also did not see problems. Lets get this in.

Thank you.

PR #3209, opened 2021-10-04
Allow plugins to exclude files from being indexed

matthiasblaesing:
This commit adds a query, that allows plugins to prevent files from
being indexed completely or by certain indexers.

One motivator for this is the JS indexer in combination with huge
node_modules directories. In the case of angular/typescript projects the
JS indexer does not help with development, as code assistence is
provided by the typescript LSP server. But the JS indexer is ran anyway
and created huge scan times and big indices. With these new hooks a
plugin could limit the indexing when an angular project is detected.

### DIFF
--- a/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/FilteringIterable.java
+++ b/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/FilteringIterable.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.parsing.impl.indexing;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+
+/**
+ * Iterable wrapping another iterable, yielding an interator, that only offers
+ * values, for this the supplied filter returns a true value.
+ */
+class FilteringIterable<T> implements Iterable<T> {
+    private final Iterable<? extends T> delegate;
+    private final Function<T,Boolean> filter;
+
+    public FilteringIterable(Iterable<? extends T> delegate, Function<T,Boolean> filter) {
+        this.delegate = delegate;
+        this.filter = filter;
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return new FilteringIterator(delegate.iterator(), filter);
+    }
+
+    private static class FilteringIterator<T> implements Iterator<T> {
+        private final Iterator<? extends T> delegate;
+        private final Function<T,Boolean> filter;
+
+        public FilteringIterator(Iterator<? extends T> delegate, Function<T,Boolean> filter) {
+            this.delegate = delegate;
+            this.filter = filter;
+        }
+
+        private T next;
+
+        @Override
+        public boolean hasNext() {
+            fillNextIfPossible();
+            return next != null;
+        }
+
+        @Override
+        public T next() {
+            fillNextIfPossible();
+            if(next == null) {
+                throw new NoSuchElementException();
+            }
+            T value = next;
+            next = null;
+            return value;
+        }
+
+        private void fillNextIfPossible() {
+            if (next == null) {
+                while (delegate.hasNext()) {
+                    T nextCandiate = delegate.next();
+                    if(filter.apply(nextCandiate)) {
+                        next = nextCandiate;
+                        break;
+                    }
+                }
+            }
+        }
+    }
+}
+
--- a/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexabilityQuery.java
+++ b/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexabilityQuery.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.modules.parsing.impl.indexing;
+
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
+import org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation;
+import org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation.IndexabilityQueryContext;
+import org.openide.filesystems.FileObject;
+import org.openide.util.Exceptions;
+import org.openide.util.Lookup;
+import org.openide.util.LookupEvent;
+import org.openide.util.LookupListener;
+
+/**
+ * Determine whether files should be skipped by the netbeans indexing infrastructure.
+ *
+ * @see org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation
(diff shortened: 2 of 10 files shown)


matthiasblaesing:
With this applied the plugin can be as simple as:

```java
package eu.doppelhelix.dev.netbeans.indexability;

import java.net.URL;
import java.util.HashSet;
import java.util.Set;
import javax.swing.event.ChangeListener;
import org.openide.util.lookup.ServiceProvider;
import org.openide.util.lookup.ServiceProviders;
import org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation;

@ServiceProviders({
    @ServiceProvider(service=org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation.class),
    @ServiceProvider(service=NodeModulesExcluder.class)
})
public class NodeModulesExcluder implements IndexabilityQueryImplementation {
       
    private static final Set<String> BLOCKED_INDEXERS = new HashSet<>();
    static {
        BLOCKED_INDEXERS.add("js");
        BLOCKED_INDEXERS.add("angular");
        BLOCKED_INDEXERS.add("requirejs");
        BLOCKED_INDEXERS.add("knockoutjs");
        BLOCKED_INDEXERS.add("TLIndexer");
        BLOCKED_INDEXERS.add("tests");
    }
            
    @Override
    public boolean preventIndexing(String indexerId, String indexerClassName, URL indexable, URL rootUrl) {
        return indexable.getPath().contains("/node_modules/") && BLOCKED_INDEXERS.contains(indexerId);
    }

    @Override
    public void addChangeListener(ChangeListener l) {
    }

    @Override
    public void removeChangeListener(ChangeListener l) {
    }
}
```

matthiasblaesing:
@JaroslavTulach it would be good if you could have a look at this from an API pespective and maybe you know someone (or are that one yourself), that could check if the implementation makes sense from the indexing infrastructure perspective.

@Chris2011 you might want to have a look at this, as this picks up an idea you planted in my head. You suggested, that the node_modules folder should be added to the "Ingnored Files Pattern" and thus be excluded from the IDE via the `VisibilityQuery`. The issue with that is, that that is a sharp weapon. I tried it for a time, but noticed, that I had to look more often into the `node_modules` folder, than I thought. The difference between the `VisibilityQuery` and the `IndexabilityQuery` is, that the first completely hides the file from the IDE, while the latter is focused on the indexing.

Chris2011:
@matthiasblaesing oh thats great, because of that functionality, that NetBeans hides the folders from the views, I created a plugin which brings back the files/folders as nodes to have a look into the node_modules as you already mentioned. So that means, that this will skip the stuff from indexing, but folder will be visible in the views, right? As we have it for gitignore. Files are greyed out a bit, still visible but ignored by git.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/FilteringIterable.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+
+/**
+ * Iterable wrapping another iterable, yielding an interator, that only offers
+ * values, for this the supplied filter returns a true value.
+ */
+public class  FilteringIterable<T> implements Iterable<T> {
```
JaroslavTulach, line 29: Factory method is better than (subclassable!) class. A candidate for `NbCollections` extension, I guess.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/FilteringIterable.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    @Override
+    public Iterator<T> iterator() {
+        return new FilteringIterator(delegate.iterator(), filter);
+    }
+
+}
+
+class FilteringIterator<T> implements Iterator<T> {
```
JaroslavTulach, line 45: Using inner class would be more typical NetBeans coding style, I guess.

ide/parsing.indexing/src/org/netbeans/modules/parsing/spi/indexing/IndexabilityQueryImplementation.java
```diff
@@ -0,0 +1,82 @@
...
+ * {@link org.openide.filesystems} with respect to threading semantics.
+ * </p>
+ *
+ * @see org.netbeans.api.queries.IndexabilityQuery
+ * @since org.netbeans.modules.parsing.indexing 9.22.0
+ */
+public interface IndexabilityQueryImplementation {
+
+    /**
+     * Determine if the supplied file object should be indexed.
```
JaroslavTulach, line 43: I assume the result must be stable for a `FileObject` until a change event is fired. What will happen when a module providing this implementation is uninstalled or new version (with different behavior) is installed?

ide/parsing.indexing/src/org/netbeans/modules/parsing/spi/indexing/IndexabilityQueryImplementation.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param indexerName name of the indexer that will be invoked on the
+     * indexable
+     * @param factoryClassName class name of the factory for the indexer that
+     * will be invoked on the indexable
+     * @param indexable the file that is about to be indexed
+     * @param rootUrl the root of the index process, can be {@code null}
+     * @return
+     */
+    default boolean preventIndexing(String indexerName, String factoryClassName, URL indexable, URL rootUrl) {
```
JaroslavTulach, line 65: Factory class name of indexers becomes an API! Weird.

ide/parsing.indexing/src/org/netbeans/modules/parsing/spi/indexing/IndexabilityQueryImplementation.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public interface IndexabilityQueryImplementation {
+
+    /**
+     * Determine if the supplied file object should be indexed.
+     * 
+     * @param fo file object to be indexed
+     * @return true if indexing should be prevented
+     */
+    default boolean preventIndexing(FileObject fo) {
```
JaroslavTulach, line 48: Is the "duality" of two `default` methods (which [are usually bad](http://wiki.apidesign.org/wiki/DefaultMethods) in an API as they increase fuzziness) necessary? Since [ability to return non-existing files](https://bits.netbeans.org/12.5/javadoc/org-openide-filesystems/apichanges.html#getFileObjectNonExisting) we may not need the `URI` variant at all.

ide/parsing.indexing/src/org/netbeans/modules/parsing/spi/indexing/IndexabilityQueryImplementation.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Determine if the supplied indexable should be indexed using the given
+     * indexer (identified either by the name of the indexer or its factory
+     * class).
+     *
+     * @param indexerName name of the indexer that will be invoked on the
+     * indexable
+     * @param factoryClassName class name of the factory for the indexer that
+     * will be invoked on the indexable
+     * @param indexable the file that is about to be indexed
```
JaroslavTulach, line 61: I suggest to get ready for [evolution](http://wiki.apidesign.org/wiki/Evolution) of this method. Instead of this multi-parameter callback method, consider creating an [open space](http://wiki.apidesign.org/wiki/APIvsSPI) `public final class` parameter.

JaroslavTulach (approved):
I am suggesting few cosmetic changes to the proposed API. My major concern is proper _invalidation_ of the indexed material when a module is uninstalled/upgraded.  I cannot judge whether this filtering approach is/isn't proper from the parsing API internals. I've added Tomáš, Dušan and Sváťa to comment on that.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexabilityQuery.java
```diff
@@ -0,0 +1,160 @@
...
+import org.openide.util.Lookup;
+import org.openide.util.LookupEvent;
+import org.openide.util.LookupListener;
+
+/**
+ * Determine whether files should be skipped by the netbeans indexing infrastructure.
+ * 
+ * @see org.netbeans.modules.parsing.spi.indexing.IndexabilityQueryImplementation
+ */
+public final class IndexabilityQuery {
```
sdedic, line 41: The main entry method, `preventIndexing` is package-private; maybe the whole class should be as well ?

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexabilityQuery.java
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+    }
+
+    @SuppressWarnings("ReturnOfCollectionOrArrayField")
+    private synchronized List<IndexabilityQueryImplementation> getIqiInstances() {
+        if (cachedIqiInstances == null) {
+            iqiResult = Lookup.getDefault().lookupResult(IndexabilityQueryImplementation.class);
+            iqiResult.addLookupListener(resultListener);
+            setupChangeListeners(null, new ArrayList<>(iqiResult.allInstances()));
```
sdedic, line 120: I'd call the `allInstances` outside of `synchronize` mutex, just for case.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/IndexabilityQuery.java
```diff
@@ -0,0 +1,160 @@
...
+        public void resultChanged(LookupEvent ev) {
+            setupChangeListeners(cachedIqiInstances, new ArrayList<>(iqiResult.allInstances()));
+            fireChange(new ChangeEvent(this));
+        }
+    }
+
+    private class IqiChangedListener implements ChangeListener {
+        @Override
+        public void stateChanged(ChangeEvent e) {
+            fireChange(e);
```
sdedic, line 157: The refired event carries the original `IndexabilityQueryImplementation` as the `getSource()` -- is it intended ?

ide/parsing.indexing/src/org/netbeans/modules/parsing/spi/indexing/IndexabilityQueryImplementation.java
sdedic: I was about to suggest to pass the `CustomIndexerFactory` instance instead, but that might leak the ability to create Indexers :) to other parties.

matthiasblaesing:
I intent to do another self review this week/weekend and if I'm happy with the result I intent to merge. If anyone wants to object, this would be a good time.

PR #3216, opened 2021-10-06
[NETBEANS-6107] Allow to use Gradle 7.2 distributions on JDK17.

sdedic:
If a project (i.e. a project created by Micronaut 3.0.3 wizard) specifies (e.g. `gradle/wrapper/gradle-wrapper.properties`) that Gradle 7.2 is to be used, and NB runs on JDK17, the IDE falls back to gradle 7.0 that fails to load system classes bcs of incompatible major class version.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
@@ -96,6 +96,7 @@ public final class GradleDistributionManager {
         GradleVersion.version("6.3"), // JDK-14
         GradleVersion.version("6.7"), // JDK-15
         GradleVersion.version("7.0"), // JDK-16
+        GradleVersion.version("7.2"), // JDK-17
     };
     private static final int JAVA_VERSION;
 


lkishalmi:
Using introspection build actions in Gradle, has no opportunity to set the JAVA_HOME it will always use the Java version on which the IDE runs. Well the compatibility check still have to work on.
Gradle plans to support Java 17 in 7.3 officially. It shall be out in early October. Beside of this change we need to increase the bundled gradle-tooling library as well, whenever 7.3 comes out.
So let's wait for 7.3.

sdedic:
At this moment 7.2 "somehow" handles JDK17 - when using  7.0 (current state) a Gradle 7.2-based project will not even load metadata and appears totally broken in the IDE. The user experience is significantly degraded - and without a reason, 7.2 gradle support allows at least the IDE to load something for simpler projects.

There's even no 7.3 RC yet. Sure we should note in release notes that JDK17 Gradle support is experimental and make an update once Gradle officially supports JDK17 in its 7.3 release.

lkishalmi:
Well, the change is pretty harmless, let's have it, as it may help.
Just checked, unfortunately even the final 7.2 tooling-api is unavailable on the Gradle release maven repo, the latest is the 7.2-rc1.
I'd keep an eye on that one. The nightlies are 7.4 already though.

PR #3221, opened 2021-10-07
[NETBEANS-6081] - Cleaning enterprise module part 1/8:

pepness:
I did some cleaning to the `enterprise` module with the `Inspect & Transform` tool. I divided the cleaning in 8 parts/commits. This is the first PR.

Cleaning enterprise module part 1:

- Remove unnecessary boxing
- Remove unnecessary unboxing
- Remove boxing of already boxed value
- Remove creating new Boolean

I did full builds and fix the errors that pop out.

NetBeans Testing:
- Full build done (many times)
- Verify successful execution of unit tests for all modules inside `enterprise` and compare the output to the same unit test from master
- Successfully work with my projects


mbien:
you are right there is a missing rule for
```
int five = Integer.valueOf("5");
=>
int five = Integer.parseInt("5");
```

there only a rule which triggers on
```
int five = Integer.valueOf("5").intValue()
=>
int five = Integer.parseInt("5");
```

so if the autoboxing refactoring was applied first, the perf refactoring won't trigger. Probably something I could fix.

mbien:
started to update some autoboxing rules https://github.com/apache/netbeans/pull/3240

PR #3223, opened 2021-10-08
Handle dots in class name by treating the prefix as a package structure

JaroslavTulach:
A currious user of VSCode wrote:

_If you create a class with a package name an invalid class is created. The package should be created and the file placed in the correct sub directory according the the package and named correctly._

The VSCode UI is pretty minimalistic and rather than enhancing it, I suggest to improve `JavaDataObject.createFromTemplate` to understand `.` and create the necessary structure itself.


JaroslavTulach:
One more note: __file the name should be run through Java character validation, since you don't want folks using symbols and weird unsupported characters in the name. __ resolved in 3df666b. Plus I did 4fb57d2 - something I always wanted to do as I never remember the proper format for the parameter comment.

PR #3224, opened 2021-10-08
LSP: Various Test Explorer fixes.

dbalek:
- Support for displaying `@Nested` junit tests added.
- GR-29769: New Test Explorer support doesn't show output from Standard out.
- GR-34232: Test Explorer out of sync when adding tests.
- GR-34105: Run all tests from Test Explorer UI never finishes.


JaroslavTulach:
`org.netbeans.modules.gradle.VerifyGradleProjectTemplatesTest` failure in [this run](https://app.travis-ci.com/github/apache/netbeans/jobs/542457414) is probably my error and can be ignored.

PR #3225, opened 2021-10-08
Enable Groovy by default in vscode and NBJLS.

sdedic:
Groovy indexing was disabled because of its slowness, but that seems to be fixed, so let's retain the ability to disable the support if the user sees it as harmful, but enable by default.


sdedic:
`JavaExecTokenProviderTest` is failing, but unrelated to this PR. Will fix separately.

PR #3227, opened 2021-10-08
[NETBEANS-6111] auto completion for "record" is now prefix aware.

mbien:
fixes https://issues.apache.org/jira/browse/NETBEANS-6111

auto completion for "record" didn't take already typed prefixes into account.

### DIFF
--- a/java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask114FeaturesTest.java
+++ b/java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask114FeaturesTest.java
@@ -33,7 +33,7 @@ import org.openide.util.lookup.ServiceProvider;
  */
 public class JavaCompletionTask114FeaturesTest extends CompletionTestBase {
 
-    private static String SOURCE_LEVEL = "14"; //NOI18N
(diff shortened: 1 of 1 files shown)


jlahoda:
Looks good, but it would be nice to have a test (e.g. `java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTask114FeaturesTest.java`).

mbien:
@jlahoda done, updated the test + I moved the two .pass files around since i believe they are meant to be in the folder for 14 and not 1.8.

Pinging @matthiasblaesing, maybe we can get this fix for a (fairly annoying) bug in too.

jlahoda:
Thanks for the test, but please return the golden files to the 1.8 folder - the golden files are based on the runtime platform, not on the source level (as e.g. the content of `java.lang` differs across JDK versions). It is a bit confusing to run JDK 14+ language feature tests on JDK 8, but without that, we would have no coverage of these features on 8-13. (The tests runs with nb-javac on JDK 8, so the javac can understand records, just the runtime has a different content of things like `java.lang`.)

mbien:
@jlahoda roger, the two .pass files are back in the 1.8 folder

PR #3228, opened 2021-10-09
fixed class modifier auto completion for sealed classes.

mbien:
minor bugfix:
`contains(String)` was called on a `Set` containing `Modifier` enums - which is always false.

Although we can't use the two sealed enums directly since they are JDK 17+, they can be still queried via a little utility method.

Added auto completion tests for sealed class declaration.

### DIFF
--- a/harness/nbjunit/src/org/netbeans/junit/NbTestCase.java
+++ b/harness/nbjunit/src/org/netbeans/junit/NbTestCase.java
@@ -739,7 +739,7 @@ public abstract class NbTestCase extends TestCase implements NbTest {
         } else {
             try {
                 if (diffImpl.diff(test, pass, diffFile)) {
-                    throw new AssertionFileFailedError(message, null == diffFile ? "" : diffFile.getAbsolutePath());
+                    throw new AssertionFileFailedError(message+"\n diff: "+diffFile, null == diffFile ? "" : diffFile.getAbsolutePath());
                 }
             } catch (IOException e) {
                 fail("exception in assertFile : " + e.getMessage());
@@ -760,7 +760,7 @@ public abstract class NbTestCase extends TestCase implements NbTest {
      * already initialized, when passed in this assertFile function.
      */
     static public void assertFile(File test, File pass, File diff, Diff externalDiff) {
-        assertFile(null, test, pass, diff, externalDiff);
+        assertFile("Difference between " + test + " and " + pass, test, pass, diff, externalDiff);
     }
     /**
      * Asserts that two files are the same, it compares two files and stores possible differences
@@ -789,7 +789,7 @@ public abstract class NbTestCase extends TestCase implements NbTest {
      * by the '.diff'.
      */
     static public void assertFile(File test, File pass, File diff) {
-        assertFile(null, test, pass, diff, null);
+        assertFile("Difference between " + test + " and " + pass, test, pass, diff, null);
     }
     /**
      * Asserts that two files are the same, it just compares two files and doesn't produce any additional output.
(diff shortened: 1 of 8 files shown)


mbien:
@Akshay-Gupta-Oracle @matthiasblaesing @jlahoda anyone interested to review this?

Had time to add some tests now. i give this a 75% chance that I did put the golden files into the wrong folder (again).

harness/nbjunit/src/org/netbeans/junit/NbTestCase.java
```diff
@@ -739,7 +739,7 @@ static public void assertFile(String message, File test, File pass, File diff, D
         } else {
             try {
                 if (diffImpl.diff(test, pass, diffFile)) {
-                    throw new AssertionFileFailedError(message, null == diffFile ? "" : diffFile.getAbsolutePath());
+                    throw new AssertionFileFailedError(message+"\n diff: "+diffFile, null == diffFile ? "" : diffFile.getAbsolutePath());
```
matthiasblaesing, line 742: How large are the files? I think the `diffFile` should lenght limited.

java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
```diff
@@ -4880,6 +4886,14 @@ private void addClassModifiers(Env env, Set<Modifier> modifiers) {
         }
     }
 
+    private static boolean contains(Set<Modifier> modifiers, String modifier) {
+        try {
+            return modifiers.contains(Modifier.valueOf(modifier));
+        } catch (IllegalArgumentException ex) {
```
matthiasblaesing, line 4892: How performance critical is this? I read somewhere, that exception handling is not the fast path in the JVM, so if this is a problem we'd need to buffer that.

matthiasblaesing (commented):
I left a single inline comment - apart from that, this looks sane to me. It would be great if the other reviewers could also have a look. I'll restart the tests.

harness/nbjunit/src/org/netbeans/junit/NbTestCase.java
mbien: it only prints the fileneme. so that you can quickly cat in bash to see what the problem is.

java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
mbien: it shouldn't be a problem there since its called only a handful of times per completion. But i do agree in principle, exceptions should be avoided if possible. 

Additionally: it won't throw with recent javacs, be it auto-javac or JDK's. If java 17 javac is standard we can remove all the workarounds and use the enum directly.

harness/nbjunit/src/org/netbeans/junit/NbTestCase.java
matthiasblaesing: A should have been obvious - thx.

java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
matthiasblaesing: Thanks for confirming.

matthiasblaesing (approved):
Ran tests locally and worked as expected. @mbien if in one week no second review comes in, please ping me and I'll merge.

mbien:
@matthiasblaesing thanks. will do. Would be nice if @jlahoda could take a look. Because i have the suspicion that the golden files still have to be in the 1.8 folder but i am not 100% certain. It depends on whether NB 13 having JDK 17 compatible nb-javac installed by default or not I guess.

jlahoda:
> @matthiasblaesing thanks. will do. Would be nice if @jlahoda could take a look. Because i have the suspicion that the golden files still have to be in the 1.8 folder but i am not 100% certain. It depends on whether NB 13 having JDK 17 compatible nb-javac installed by default or not I guess.

Yes, I think it would be preferrable to have the in the 1.8 folder. The folders represent the runtime platform, and are unrelated to the source level supported by the Java parser. (If the selected Java parser does not support the source level, the test will be skipped anyway, but I think we are heading to a place where this won't happen normally.) I realize this is fairly confusing.

Otherwise looks good!

Thanks!

mbien:
@jlahoda thanks jan! Moved them to the 1.8 folder.

jlahoda:
Looks good.

PR #3240, opened 2021-10-12
[jackpot] Updates to autoboxing inspections

mbien:
* repurposed a dusty perf rule to suggest `Boxed.valueOf() -> Boxed.parseType()` refactoring where applicable
  * added reverse case
  * added rules for overloaded method variants
* extracted some functionality into a new rule which has the lone purpose to get rid of deprecated boxed primitive construction
* some rules overlapped with other inspections, removed those to keep things simple.
  * this might mean however that some inspections have to be applied multiple times, since one fix could trigger another

example:
```java
    private static void a(Integer i) {}
    private static void b(String foo, int i) {}
    private static void c(String foo, double a, long i) {}

    public static void test() {
        
        a(Integer.valueOf(""));                             // do nothing
        b("abcd", Integer.valueOf("1"));                    // => parseInt()
        b("abcd", Integer.valueOf("1").intValue());         // => remove .intValue() => parseInt()
        c("asdf", Float.valueOf("1"), Long.valueOf("1"));   // => parseFloat() and => parseLong()
        
        
        ArrayList<Object> arrayList = new ArrayList<>(Integer.valueOf("6")); // => parseInt()
        arrayList.get(Integer.valueOf("1"));                                 // => parseInt()
        
        
        float prim1 = Float.parseFloat("5");        // do nothing
        prim1 = Float.valueOf("6");                 // => parseFloat()
        prim1 = Float.valueOf("6").floatValue();    // => remove .floatValue() => parseFloat()
        prim1 = Float.parseFloat("6");              // do nothing
        
        Float boxed1 = Float.parseFloat("5");       // => valueOf()
        boxed1 = Float.valueOf("6");                // do nothing
        boxed1 = Float.valueOf("6").floatValue();   // => remove .floatValue()
        boxed1 = Float.parseFloat("6");             // => valueOf()

        byte b = new Byte("5").byteValue();     // => remove .byteValue() => valueOf() => parseByte()

        // short form and AIC works too
        Executors.newFixedThreadPool(1).submit(() -> {
            return Integer.parseInt("1");         // => valueOf()
        });
        ....
    }
```

todo/future
* ~tests~
* getDestinationType() covers hopefully the most common cases but this might not be enough
  * potentially useful utility method if made accessible from hint files (but that is for later)
* might be cleaner to put all boxing-related rules into the same package and display those in the same category in the UI too, they are currently all over the place (potential future task)

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
@@ -22,12 +22,15 @@ package org.netbeans.modules.java.hints.perf;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.NewArrayTree;
+import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.ParameterizedTypeTree;
 import com.sun.source.tree.ParenthesizedTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.Tree.Kind;
 import com.sun.source.util.TreePath;
+import com.sun.source.util.Trees;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -39,6 +42,7 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.ExecutableType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
@@ -61,6 +65,10 @@ import org.netbeans.spi.java.hints.ErrorDescriptionFactory;
 import org.netbeans.spi.java.hints.JavaFix.TransformationContext;
 import org.netbeans.spi.java.hints.JavaFixUtilities;
 import org.openide.util.NbBundle;
+
+import static com.sun.source.tree.Tree.Kind.METHOD_INVOCATION;
+import static com.sun.source.tree.Tree.Kind.NEW_CLASS;
+import static javax.lang.model.type.TypeKind.EXECUTABLE;
 
 /**
  *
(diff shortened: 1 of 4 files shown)


matthiasblaesing:
@jlahoda could you please give this a look?

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
```diff
@@ -462,36 +485,94 @@ public static ErrorDescription redundantToString(HintContext ctx) {
...
-            return null;
+
+        String type;
+        String method;
+
+        // determine if the destination is primitive
+        TypeMirror destType = getDestinationType(ctx, ctx.getPath());
+        TypeMirror srcType = ctx.getInfo().getTrees().getTypeMirror(ctx.getPath());
+
+        if (destType instanceof PrimitiveType && !(srcType instanceof PrimitiveType)) {
```
jlahoda, line 496: Generally, the javac APIs discourage `instanceof` and encourage use of `.getKind() == <type>` (or probably `.getKind().isPrimitive()` in this case), even though that obviously lacks the implicit null check. The reason is that one implementation class may implement many of these interfaces (applies to all `javax.lang.model.**` and `com.sun.source.**` that have kinds). As a real-world example, the `com.sun.tools.javac.code.Type.ClassType` implements both `DeclaredType` and `ErrorType`, so doing `instanceof DeclaredType` or `instanceof ErrorType` would not work properly. The `.getKind()` returns the correct kind for the given type.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
```diff
@@ -462,36 +485,94 @@ public static ErrorDescription redundantToString(HintContext ctx) {
...
         }
-        String[] arr = PARSE_METHODS.get(resType.getKind());
-        if (arr == null) {
-            return null; // just in case
+
+        Fix fix = JavaFixUtilities.rewriteFix(ctx, Bundle.FIX_UnnecessaryTempFromString1(type, method), ctx.getPath(), type + "." + method + "($v)"); // NOI18N
+        return ErrorDescriptionFactory.forTree(ctx, ctx.getPath(), Bundle.TEXT_UnnecessaryTempFromString(), fix); // NOI18N
+    }
+
+    private static TypeMirror getDestinationType(HintContext ctx, TreePath path) {
```
jlahoda, line 516: There is `java/java.hints/src/org/netbeans/modules/java/hints/errors/CreateElementUtilities.java#resolveType` which presumably does mostly what is needed here. There are some tricks played there, but I hope that could work.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
```diff
@@ -462,36 +485,94 @@ public static ErrorDescription redundantToString(HintContext ctx) {
...
+        Types types = ctx.getInfo().getTypes();
+        Trees trees = ctx.getInfo().getTrees();
+
+        if (parentLeaf instanceof MethodInvocationTree) {
+
+            MethodInvocationTree met = (MethodInvocationTree) parentLeaf;
+
+            int index = met.getArguments().indexOf(path.getLeaf());
+
+            TypeMirror method = trees.getElement(new TreePath(path, met.getMethodSelect())).asType();
```
jlahoda, line 530: As a general note - doing `Element.asType()` will return the type of the generic declaration (i.e. a type with the generic type parameters, like e.g. `(E)boolean` for `List.add`). What is likely needed here is the actual types used for the method invocation, with generic types (if any) resolved (e.g. `(Integer)boolean` for `List<Integer> l; l.add(...)`). `trees.getTypeMirror(<method-select>)` should return that, AFAIK.

(Note there's a slight inconsistency/difference between the `getElement` and `getTypeMirror` methods:
-if the method select is passed into the methods, they will return the `ExecutableElement` and `ExecutableType` for the invocation
-if the method invocation is passed into the methods, `getElement` will still return the `ExecutableElement`, but `getTypeMirror` will return the actual return type (with all generics resolved as needed)
)

jlahoda (commented):
Overall, seems fine. I left a few general comments on the APIs and a pointer to a possible current impl of `getDestinationType`.

My main comment to this patch as such is that I think it is preferred to not have hints/fixes that immediately lead to another hint/fix. I think I would prefer if the changes would be done in one step, unless it is very difficult to do so.

Thanks for working on this!

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: good to know -> switched everything to getKind()

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: awesome! I tried to find a utility like this but must have overlooked it. Replacement was easy.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: I tried getTypeMirror first actually and it worked for methods, but I could not figure out how to make the NewClassTree work too. So instead of using two ways of handling very similar branches of an if, i decided to solve both the same.

The current version I just committed uses getTypeMirror for the method branch and Element.asType for the class branch. I would need some help to make both use getTypeMirror (i am missing the getMethodSelect() equivalent for constructors) - since i could not figure out how.

mbien: the new code:
https://github.com/apache/netbeans/blob/96f3355749c8500a53cae56a98abe2288887521a/java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java#L521-L531

mbien:
> My main comment to this patch as such is that I think it is preferred to not have hints/fixes that immediately lead to another hint/fix. I think I would prefer if the changes would be done in one step, unless it is very difficult to do so.

I thought about this a lot while writing jackpot scripts in past (I uploaded some general purpose ones here: https://github.com/mbien/jackpot-inspections). At first I tried to maintain "shortcuts" which attempted to refactor everything in one step, but quickly saw that this would let the rule count explode. A shortcut is also only a single path from a specific bad code pattern to clean code, there are more creative ways of bad code than there can be shortcut rules - finding them all is a bit hopeless.

Additionally, experience shows from refactoring multiple projects from old java to new java that rules interact in fairly unpredictable ways. I always apply the same set of rules twice, since the second time will usually find something new if the project is large enough - its not really possible to account for that in a shortcut rule.

(Maybe its worth investigating if there would be a way to automatically merge rules. If a rule triggers, it could check if the refactoring would trigger another one.)

But back to this concrete PR. The problem here was that the rules were all over the place. There are autoboxing rules in Tiny and also several in the jdk package.  There were gaps, some overlapped and some were not correct (as noticed in https://github.com/apache/netbeans/pull/3221). Its probably worth to move everything into one package and see after that if some steps could be left out somehow by merging rules.

> 
> Thanks for working on this!

no problem

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: @jlahoda you think it is OK to leave the NEW_CLASS if-branch as is? I still don't know how to get a handle on a constructor, with getTypeMirror I would only get the class type, not the constructor type with its parameters.

mbien:
my mistake. I forgot to press the re-request review button two weeks ago.

java/java.hints/src/org/netbeans/modules/java/hints/perf/Tiny.java
mbien: @jlahoda ping :)

mbien:
@jlahoda are there any blockers left before integration? This PR getting a bit old and I am not quite sure what is left to do here. We got about a week before feature freeze.

PR #3242, opened 2021-10-13
[NETBEANS-6086] YAML Parser Changes (fix: OOME)

lkishalmi:
Working as DevOps, I need to have a reliable ans fast YAML engine under my hand.
The current one in NetBeans is famously slow and in 12.4 and 12.5 the error tolerant parser has a memory leak which is not that hard to encounter with.

So I took a deep breath. I've replaced the parser engine to snakeyaml-engine which is more recent and actively maintained, and support YAML 1.2. I was almost able to make all the previous test pass when using the SnakeYaml's own composer and the Node API. The ones which failed were mostly alias issues. 7 failed out of the original 98. 

Hitting a dead end, I've implemented a custom composer, which does not resolve aliases on the fly. That required a major rework on the existing code. But finally all the tests are passing with some minor, but necessary, changes on the test data. 

I plan to add parser auto-recovery back for better experience. It would be based on splitting the YAML source around the errors to parsable fragments. In my tests it is more or less working, though some edge cases can lead to infinite parsing, though I have not encountered with memory leak even when the infinite parsing happens.

So this PR is for review and give a hopefully reliable YAML experience.

### DIFF
--- a/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlParser.java
+++ b/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlParser.java
@@ -38,7 +38,6 @@ import org.netbeans.modules.parsing.api.Task;
 import org.netbeans.modules.parsing.spi.ParseException;
 import org.netbeans.modules.parsing.spi.SourceModificationEvent;
 import org.openide.util.NbBundle;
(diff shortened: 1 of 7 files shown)


ppisl:
This good change. I would vote for it.

lkishalmi:
Closing in favor of #3245

PR #3244, opened 2021-10-13
[NETBEANS-6127] Do not parallelize notifyToBeResumedNoFire(), because it runs under a lock.

entlicher:
When all threads are suspended as a result of an event (e.g. a breakpoint that suspends all threads), we notify all instances of `JPDAThreadImpl` about the suspension (via `JPDADebuggerImpl.notifyToBeResumedAllNoFire()` method) under debugger write lock.
To update its state, the `JPDAThreadImpl` acquires debugger read lock. As long as it's run synchronously, there's no issue as the read lock can be acquired under the write lock. But when we parallelized the calls to individual threads to improve performance on slow connection, we suddenly tried to acquire the debugger read lock on a new thread. With the original thread still holding the write lock. This has led to a deadlock.

The solution is to notify the threads synchronously again.


matthiasblaesing:
Thank you - the fix seems to work. I tested various combinations (conditional break point, suspend all threads, suspend no thread). It seems the behavior matches the original behavior. I also briefly tested in a real application. I'll prepare a nightly version with my configuration and will see the next few day how the system  behaves. It improves the situation, so if @JaroslavTulach agrees, IMHO this should go in.

PR #3247, opened 2021-10-14
[NETBEANS-6130] - Fix wrong Groovy version in ant library

pepness:
Wrong version number in the library.

It is 2.5.11
NETBEANS-6130

Should be 3.0.8


svenreimers:
Would be nice to add a latest 2.5 as well as a 3.x library for convenience and compatibility?

lkishalmi:
Trivial change, let's have it in!

PR #3254, opened 2021-10-19
[NETBEANS-3391] Support PSR-12 valid formatting rules

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-3391

- Add new wrapping options for parentheses of "for/foreach" and Method Call Args, and Method Params
- Wrap After '(' and Wrap ')': Wrap when there is a newline within ()
- Keep ')' and '{' on the same line: Keep them when the Braces option
for Method Declaration is "New Line" and wrap ()

e.g.
```php
//for ($index = 0;
//$index < count($array);
//$index++) {
//    echo $index;
//}
for (
        $index = 0;
        $index < count($array);
        $index++
) {
    echo $index;
}

//test($param1,
//        $param2,
// $param3);
test(
        $param1,
        $param2,
        $param3
);

//function test(
//$param1,
//$param2,
//$param3,): void {
//
//}
function test(
        $param1,
        $param2,
        $param3,
): void {

}

// Keep ')' and '{' on the same line
// the Braces option for Method Declaration is "New Line"
//function test($param1, int $param2): void{
//}
function test($param1, int $param2): void
{

}

// Keep ')' and '{' on the same line
// the Braces option for Method Declaration is "New Line"
//function test(int $param1,
//        Test &$param2,
//): void{
//}
function test(
        int $param1,
        Test &$param2,
): void {

}

```
- Note: It doesn't work correctly with new options yet when the wrapping
option of Method Parameters/Method Call Arguments/For is "If Long"

#### Screenshot

netbeans-3391-formatting-wrapping-options


junichi11:
@tmysik Could you please have a look at this although I think it's impossible to review all? Sorry for the huge changes.

Feature freeze: https://lists.apache.org/thread.html/r3201513afcc326391b344a33a32475e7e61b661f4342281cbe93acfe%40%3Cdev.netbeans.apache.org%3E
cc: @neilcsmith-net

junichi11:
Will add a screenshot of the option later.

tmysik:
@junichi11 

Let me know if I can merge it (when? now?).

Since we have **many** tests here, I would expect that it should be fine.

Thanks a lot for your great work!

PR #3257, opened 2021-10-20
[NETBEANS-6087] Use the refactoring feature for private members of trait instead of the instant renamer

junichi11:
- https://issues.apache.org/jira/browse/NETBEANS-6087
- Classes can also use private members of traits. If the declaration of the member is in another file, only member names in the same file are renamed using the instant renamer. So, we have to use the refactoring feature instead of it.

#### Screenshots

netbeans-6087-1

netbeans-6087-2


tmysik:
@junichi11 

There are some CI failures - do you want to review them? Or can I merge this PR? Thanks.

PR #3262, opened 2021-10-22
[NETBEANS-5847] Maven project template uses template for project structure.

sdedic:
The PR improves how the 'native project templates' for maven are handled. 

The native templates, created by NB code (instead of `archetype:create` run) were generated entirely. The Layer template was just a stub whose only function was to invoke the wizard. The `api.templates` API lacks the ability to construct a subtree populated with files; this PR fixes that - the default processing will copy out folder-based template. Naturally if a specific `CreateFromTemplateHandler` chooses a different implementation, it will take precedence. As part of the changes, I've introduced possibility to interpolate template parameters in filenames - see the provided test resources.

There are a few helper methods for CreateFromTemplateHandler implementors, which do stuff whose details are IMHO better not to export from the `api.templates` module or have the CFTH implementors replicate.

There's also [NETBEANS-5847](https://issues.apache.org/jira/browse/NETBEANS-5847)  that notes that the maven application Wizard cannot be run from a headless mode (provided all necessary parameters are in). This PR moves much of WizardIterator's logic into `CreateFromTemplateHandler` implementation, and the Wizard just populates the `FileBuilder` parameter map.


java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/layer.xml
```diff
@@ -39,7 +39,7 @@
             </folder>
             <file name="Import_hidden"/>
             <folder name="Maven2">
-                <file name="JavaApplication_hidden"/>
+                <file name="maven.archetype_hidden"/>
```
JaroslavTulach, line 42: Why was existing `JavaApplication` file renamed? Or where the `maven.archetype` file/folder comes from.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/package-info.java
```diff
@@ -21,14 +21,6 @@
     @TemplateRegistration(folder = "Classes/Code", position = 100, content = "resources/GeneratedMethodBody.template", scriptEngine = "freemarker", displayName = "#GeneratedMethodBody", category = "java-code-snippet"),
     @TemplateRegistration(folder = "Classes/Code", position = 200, content = "resources/OverriddenMethodBody.template", scriptEngine = "freemarker", displayName = "#OverriddenMethodBody", category = "java-code-snippet"),
     @TemplateRegistration(folder = "Classes/Code", position = 300, content = "resources/LambdaBody.template", scriptEngine = "freemarker", displayName = "#LambdaBody", category = "java-code-snippet"),
-    @TemplateRegistration(
```
JaroslavTulach, line 24: Less differences between VSNetBeans and NetBeans is good.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
```diff
@@ -138,6 +138,7 @@ private LspTemplateUI() {
                 final DataFolder target = targetAndName.first();
                 final String name = targetAndName.second();
                 Map<String,String> prjParams = new HashMap<>();
+                // PENDING: this is a mavenism; 
                 prjParams.put("version", "1.0-SNAPSHOT"); // NOI18N
```
JaroslavTulach, line 142: Please modify existing templates (as exposed by VSNetBeans) to use `1.0-SNAPSHOT` as default and remove this attribute.

java/maven/src/org/netbeans/modules/maven/layer.xml
```diff
@@ -33,6 +33,22 @@
...
                 <attr name="displayName" bundlevalue="org.netbeans.modules.maven.newproject.Bundle#Templates/Project/Maven2"/>
                 <attr name="position" intvalue="100"/>
+
+                <!-- Simple Java project structure -->
+                <folder name="JavaApplication2">
+                    <folder name="src">
+                        <folder name="main">
+                            <folder name="java">
+                                <folder name="${packagePath}">
+                                    <file name="${mainClassName}.java" url="nbresloc:/org/netbeans/modules/maven/resources/App.java.template">
```
JaroslavTulach, line 43: Using new substitution mechanism for names is OK, but the syntax! Let's make the syntax pluggable! Sváťo, what do you think about:

If there is an `<attr name="javax.script.ScriptEngine" stringvalue="freemarker"/>` attribute on a file or folder, apply the specified engine (e.g. `freemarker`) to the name of the file/folder as well.

java/maven/src/org/netbeans/modules/maven/newproject/TemplateUtils.java
```diff
@@ -0,0 +1,268 @@
...
+import org.netbeans.spi.project.ActionProvider;
+import org.netbeans.spi.project.ui.support.CommonProjectActions;
+import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
+import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
+import org.openide.util.lookup.Lookups;
+import org.openide.util.lookup.ProxyLookup;
+
+public final class TemplateUtils {
```
JaroslavTulach, line 47: Can this class be package private?

java/maven/src/org/netbeans/modules/maven/newproject/idenative/IDENativeTemplateHandler.java
```diff
@@ -0,0 +1,181 @@
...
+import org.openide.filesystems.FileUtil;
+import org.openide.util.BaseUtilities;
+import org.openide.util.NbBundle;
+
+/**
+ * Allows to create the 'ide native' projects in the headless mode. Is not registered as an interceptor, but rather
+ * listed explicitly for a template it handles.
+ * @author sdedic
+ */
+public class IDENativeTemplateHandler extends CreateFromTemplateHandler {
```
JaroslavTulach, line 48: Can we use some other name than IDE native? There is no native code and I find the name confusing.

java/maven/src/org/netbeans/modules/maven/newproject/idenative/SimpleJavaNativeMWI.java
```diff
@@ -19,103 +19,29 @@
...
-import org.netbeans.modules.maven.options.MavenSettings;
 import org.openide.util.NbBundle.Messages;
 
 /**
  *
  * @author mkleint
  */
 @TemplateRegistration(
-    id = "JavaApplication",
+    id = "JavaApplication2/",
```
JaroslavTulach, line 32: Why is this template being renamed?

java/maven/test/unit/src/org/netbeans/modules/maven/newproject/simplewizard/SimpleWizardTemplateHandlerTest.java
```diff
@@ -0,0 +1,174 @@
...
+        whereTo = FileUtil.createFolder(FileUtil.toFileObject(getWorkDir()), "whereTo");
+        
+        FileObject template = FileUtil.getConfigFile("Templates/Project/Maven2/JavaApplication2");
+        FileBuilder bld = new FileBuilder(template, whereTo);
+        bld.name("TestProject");
+        bld.defaultMode(FileBuilder.Mode.COPY);
+
+        bld.param("artifactId", "netbeans-test-project");
+        bld.param("groupId", "org.netbeans.modules.maven");
+        // leave out version
```
JaroslavTulach, line 56: Looks like the version can be left out.

platform/api.templates/src/org/netbeans/api/templates/CreateDescriptor.java
```diff
@@ -181,4 +187,17 @@ public boolean hasFreeExtension() {
...
+    void setLookup(Lookup contextLookup) {
+        this.contextLookup = contextLookup;
+    }
+
+    /**
+     * Lookup that may contain additional services or instructions for
+     * the template creation.
+     * @since 1.23
+     */
+    public Lookup getLookup() {
```
JaroslavTulach, line 200: The class could now `implements Lookup.Provider`.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
```diff
@@ -59,4 +64,171 @@
...
 
+    /**
+     * Replaces ${param} tokens in a String. The parameter reference has syntax "${" paramName [":" defaultValue ] "}".
+     * Parameters are replaced recursively. "${" can be escaped by \, so "\${" will not act as parameter reference start.
+     * @param expression string template
+     * @param parameters parameter values
+     * @return evaluated string.
+     * @since 1.23
+     */
+    public static String mapParameters(String expression, Map<String, ?> parameters) {
```
JaroslavTulach, line 75: There is `org.openide.util.MapFormat` - if the suggestion to use `freemarker` isn't going to fly - wouldn't it be better to reuse the existing `MapFormat` or expand it?

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
```diff
@@ -59,4 +64,171 @@
...
+     * The {@code "template"} attribute is never copied.
+     * 
+     * @param h the Handler responsible for file creation, usually the calling handler.
+     * @param from the original file
+     * @param to the target file
+     * @throws IOException if an I/O error occurs.
+     */
+    public static void copyAttributesFromTemplate(CreateFromTemplateHandler h, FileObject from, FileObject to) throws IOException {
+        // copy attributes; some attributes are filtered by FileSystems API already
+        FileUtil.copyAttributes(from, to);
```
JaroslavTulach, line 178: Consider expanding the `copyAttributes` method with a `Predicate<String> attributeName` rather than copying everything and then cleaning it up.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
```diff
@@ -59,4 +64,171 @@
...
+     * since it has Lookup, parameters etc set up. They will be passed on to individual created files. The `contentsParent` may
+     * be {@code null}, in which case the {@link CreateDescriptor#getTemplate()} folder will be used.
+     * @param origDescriptor original descriptor to get parameters / lookup from
+     * @param contentsParent template folder whose contents should be copied
+     * @param target target folder
+     * @return created files
+     * @throws IOException 
+     * @since 1.23
+     */
+    protected static List<FileObject> defaultCopyContents(CreateDescriptor origDescriptor, FileObject contentsParent, FileObject target) throws IOException {
```
JaroslavTulach, line 202: Strange method. `protected static` is a strange combination of modifiers - if the class is subclassable, there is nothing "protected" in this method. If the class isn't subclassable, making a method `protected` makes no sense.

platform/api.templates/src/org/netbeans/api/templates/FileBuilder.java
```diff
@@ -109,6 +135,20 @@
     public FileBuilder(@NonNull FileObject template, @NonNull FileObject target) {
         descriptor = new CreateDescriptor(template, target);
     }
+
+    /**
+     * Use specific Lookup for creation. This allows to pass in additional
+     * parameters or services. A GUI-aware caller may, for example, add
+     * a ProgressHandle or a project ActionProgress interface to the lookup,
```
JaroslavTulach, line 142: OK.

platform/api.templates/src/org/netbeans/api/templates/TemplateRegistration.java
```diff
@@ -156,4 +156,10 @@
      * @since 1.4
      */
     String[] techIds() default {};
+
+    /**
+     * Specifies the default handler to instantiate the template. 
+     * @since 1.23
+     */
+    Class<? extends CreateFromTemplateHandler> createHandlerClass() default CreateFromTemplateHandler.class;
```
JaroslavTulach, line 164: Nice.

JaroslavTulach (approved):
Please consider few API related changes. Overall the change looks fine.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/layer.xml
sdedic: Hm. I thought that `maven.archetype` was a way how to allow LSP to create a project. This is now replaced by the `JavaApplication2` template.

Re. rename: that's a question how to handle such situation: the original template is a **file**. I need to provide a folder structure, so:
* the template can be no longer a **file**, but must turn to a folder. Existing installations or extensions that may have contributed to that file in the ConfigFS would break (impossible to layer-merge file + folder).
* a new template will be introduced, which I did (JavaApplication2).
* a "contents here" attribute will be introduced. So the project itself can continue as **file** template, but its contents will be elsewhere. I even made code support for it, but at the end decided not to use it in the template.

Please share your thoughts.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
sdedic: Yes, you can always do a trampoline from some subclass and made it public to whatever code you like. It was meant as a helper primarily for CreateFromTemplateHandler imlementations (with limited use elsewhere), but can be `public` as well.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
sdedic: Reuse: I don't think its's possible 'as is' as I'd like to keep the default - which gives some troubles if the default itself contains delimiters (see this impl). Also the replacement in MapFormat is not recursive - that could be also solved by the caller by calling MapFormat repeatedly, but could be added.

If you think it's worth moving to MapFormat, I'll try; but I'd like to retain the `${...}` syntax, which is consistent with Ant, Freemarker or other interpolation engines.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/layer.xml
JaroslavTulach: Isn't the `maven.archetype` registered only in `java/java.lsp.server/nbcode` suite? If so, it doesn't have to be hidden.

Changing `file` to `folder` has some drawback, your solution is fine. Consider calling the folder `JavaApp` to avoid the `2`.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
JaroslavTulach: I still like the [use of the templating engine itself](https://github.com/apache/netbeans/pull/3262#discussion_r740114513) more than our own substitution mechanism...

java/maven/src/org/netbeans/modules/maven/layer.xml
sdedic: well, respecting the ScriptEngine could be helpful in some more complicated cases, asi the expression can reference context objects for example; but in simple ones like this ...

JaroslavTulach: Yes, it is an overkill.

On the other hand, the templating engine must be around and it provides familiar syntax. E.g. one would be using the same syntax for the file name as well as for the file content.

ide/projectui/test/unit/src/org/netbeans/modules/project/ui/test-layer.xml
```diff
@@ -0,0 +1,38 @@
...
+
+-->
+<!DOCTYPE filesystem PUBLIC "-//NetBeans//DTD Filesystem 1.2//EN" "http://www.netbeans.org/dtds/filesystem-1_2.dtd">
+<filesystem>
+    <folder name="Templates">
+        <folder name="Test">
+            <folder name="FolderTemplate">
+                <folder name="${packagePath!&quot;foo-bar&quot;}">
+                    <attr name="javax.script.ScriptEngine" stringvalue="freemarker"/>
+                    <file name='${mainClassName!"App"}.java' url="nbres:/org/netbeans/modules/project/ui/AppTemplate.java.template">
```
JaroslavTulach, line 29: It is not that bad!

ide/projectui/test/unit/src/org/netbeans/modules/project/ui/test-layer.xml
```diff
@@ -0,0 +1,38 @@
...
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<!DOCTYPE filesystem PUBLIC "-//NetBeans//DTD Filesystem 1.2//EN" "http://www.netbeans.org/dtds/filesystem-1_2.dtd">
+<filesystem>
+    <folder name="Templates">
+        <folder name="Test">
+            <folder name="FolderTemplate">
+                <folder name="${packagePath!&quot;foo-bar&quot;}">
```
JaroslavTulach, line 27: Ugly. The variant with `'.."..."..'` is nicer.

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -937,6 +938,23 @@ public static FileObject toFileObject(File file) {
     */
     public static void copyAttributes(FileObject source, FileObject dest)
     throws IOException {
+        copyAttributes(source, dest, null);
+    }
+    
+    /** Copies attributes from one file to another.
+    * Note: several special attributes will not be copied, as they should
+    * semantically be transient. These include attributes used by the
```
JaroslavTulach, line 946: I suggest to define field or function to let people access:
```java
public static final BiFunction<String, Object, Object> DEFAULT_ATTR_TRANSFORMER;
```
then they can either:
* use it as the 3rd parameter to get behavior of the two args method
* delegate to it from own transformer as a last resort
* skip it by not delegating from own transformer

I never really liked the hard-coded list of attributes, but they are useful. Exposing them as `BiFunction` keeps the concept of important attributes on, but puts the full control into the hands of `attrTransformer` writer.

java/maven/src/org/netbeans/modules/maven/layer.xml
```diff
@@ -35,7 +35,7 @@
                 <attr name="position" intvalue="100"/>
 
                 <!-- Simple Java project structure -->
-                <folder name="JavaApplication2">
+                <folder name="JavaApp">
```
JaroslavTulach, line 38: OK.

platform/api.templates/src/org/netbeans/api/templates/CreateFromTemplateHandler.java
```diff
@@ -59,4 +63,169 @@
...
     ) throws IOException;
 
+    /**
+     * Replaces ${param} tokens in a String. The parameter reference has syntax "${" paramName [":" defaultValue ] "}".
+     * Parameters are replaced recursively. "${" can be escaped by \, so "\${" will not act as parameter reference start.
+     * @param expression string template
+     * @param parameters parameter values
+     * @return evaluated string.
+     */
+    static String mapParameters(String expression, Map<String, ?> parameters) {
```
JaroslavTulach, line 73: I was hoping this _in-house mapping_ can now be removed when we can use the Freemarker based one (with `'...'`) easily. But I can live with both variants. Up to you.

JaroslavTulach (approved):
Great!

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
sdedic: Good to mention it now. It requires a semantic change in javadoc which would be hard to do after introduction; thanks.

sdedic: Addressed in ffc97b6

sdedic:
Finally all checks passed. We **NEED** to stabilize the testusites.

PR #3265, opened 2021-10-23
enable encoding compatibility mode to avoid problems on JDK18 due to JEP 400.

mbien:
JDK 18 makes UTF-8 the default. ~~For consistency reasons we should probably set `-Dfile.encoding=UTF-8` now on all JDKs to catch problems early, otherwise NB could behave differently dependent on which JDK it is launched on.~~ (Too late for that)

if issues are encountered we could temporarily set `-Dfile.encoding=COMPAT` till they are resolved.

[1] https://openjdk.java.net/jeps/400
[2] https://inside.java/2021/10/04/the-default-charset-jep400/

### DIFF
--- a/nb/ide.launcher/netbeans.conf
+++ b/nb/ide.launcher/netbeans.conf
@@ -63,7 +63,7 @@ netbeans_default_cachedir="${DEFAULT_CACHEDIR_ROOT}/@@metabuild.RawVersion@@"
 # (see: https://issues.apache.org/jira/browse/NETBEANS-1344)
 #
 
-netbeans_default_options="-J-XX:+UseStringDeduplication -J-Xss2m @@metabuild.logcli@@ -J-Dapple.laf.useScreenMenuBar=true -J-Dapple.awt.graphics.UseQuartz=true -J-Dsun.java2d.noddraw=true -J-Dsun.java2d.dpiaware=true -J-Dsun.zip.disableMemoryMapping=true -J-Dplugin.manager.check.updates=false -J-Dnetbeans.extbrowser.manual_chrome_plugin_install=yes @@metabuild.jms-flags@@ -J-XX:+IgnoreUnrecognizedVMOptions"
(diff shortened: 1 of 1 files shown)


mbien:
I set this to draft because testing with the encoding flag on JDK 17 or without the flag on 18 is blocked anyway.
https://github.com/apache/netbeans/pull/3278/ is a start, but making the tests actually work will be a larger task.

mbien:
the first issues start to appear on JDK 18+win. To buy us more time we should consider setting `-J-Dfile.encoding=COMPAT` as described in the original PR text.

Mid-term we should test without having the flag set and correct the encodings on the problematic streams until everything can run under the new default without compat mode.

setting NB14 milestone && RFR since it will have support for JDK 18

matthiasblaesing:
Please don't! JNA for example has a library out, that already uses the new flag. Once we go down the "compatibility" route, it will be difficult to pull back as the layer above will rely on it. I'll have a look at terminal code, but it won't be ready now.

mbien:
@matthiasblaesing I am all for fixing the actual issues. But fact is that we can't even test on JDK 17 yet (the PR with pipeline which builds on JDK 17 is awaiting review, i got no reply on the dev list), it all has to be one manually. 

NB has 840 sub projects the last time i counted, I would be very surprised if we would be able to find all encoding issues for NB 14 and fix them in time (I am not even running NB on windows so it doesn't affect me). This PR is old and nothing really happened since then - nobody fixed encoding issues with UTF-8 set on windows (I an not complaining, just saying as it is).

Compat mode buys us time and we can remove it once we are confident everything works. We can even set a timeframe. Lets remove it for NB15 or 16.

matthiasblaesing:
I explained in https://github.com/apache/netbeans/pull/3265#issuecomment-950195690 why the workaround will create a mess. So again. Please wait!

mbien:
> I explained in [#3265 (comment)](https://github.com/apache/netbeans/pull/3265#issuecomment-950195690) why the workaround will create a mess. So again. Please wait!

I am not merging without approval, don't worry.

mbien:
closing in favor of #3836

PR #3270, opened 2021-10-26
Shade whole methods which are not included in the native image.

entlicher:
Lines without the executable locations are shaded during native image debugging.
When a whole method was missing in the native image, its content was correctly shaded, but not the method header. This change makes sure that the method header itself (the method tree, not just the method body) is shaded as well.


dbalek (approved):
Looks fine.

PR #3266, opened 2021-10-25
Support for JEP 413: Snippet  java docs

singh-akhilesh:
Support for JEP 413 https://openjdk.java.net/jeps/413 java doc Internal Snippet.
https://issues.apache.org/jira/browse/NETBEANS-6480

PR #3274, opened 2021-10-26
Introduce SourceVersion utility.

mbien:
Each module has currently its own way of comparing feature versions with a SourceVersion enum which may or may not exist on the given JDK. This PR introduces a small utility in `java.source.base` which puts the comparison code in one place and hopefully makes it easier to find and change it in future (if needed). Since it doesn't rely on exceptions, just on simple integer comparison, it should be also much faster as a bonus.

e.g see discussion https://github.com/apache/netbeans/pull/3166 or changes made to `java.source.base` in this PR.

This PR also refactors `java.source.base` and `java.source` to use SourceVersion.
Future PRs might refactor more modules and/or the Jackpot API in a way so that enums don't have to be used in hint files which would greatly improve compatibility between JDK versions.

There is also a OpenJDK CSR in review related to this (thanks to Joe Darcy)
https://bugs.openjdk.java.net/browse/JDK-8275888
(update: changes made it into JDK 18 https://github.com/openjdk/jdk/pull/5973)

note to self: rebase ontop https://github.com/apache/netbeans/pull/3266 once its integrated, there might be other conflicting PRs pending.

### DIFF
--- a/java/java.source.base/src/org/netbeans/api/java/source/TreeUtilities.java
+++ b/java/java.source.base/src/org/netbeans/api/java/source/TreeUtilities.java
@@ -938,17 +938,7 @@ public final class TreeUtilities {
         }
     }
 
-    private static boolean VARIABLE_CAN_OWN_VARIABLES;
-    static {
-        boolean result;
-        try {
-            SourceVersion.valueOf("RELEASE_12");
-            result = true;
-        } catch (IllegalArgumentException ex) {
-            result = false;
-        }
-        VARIABLE_CAN_OWN_VARIABLES = result;
-    }
+    private static boolean VARIABLE_CAN_OWN_VARIABLES = SourceVersions.supports(12);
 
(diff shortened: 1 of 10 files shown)


lkishalmi:
Is it a new API that can be used by other modules?
If so please document this in the apichanges.xml

mbien:
> Is it a new API that can be used by other modules? If so please document this in the apichanges.xml

Yes, I keep this as draft until some conflicting PRs are merged so that i can rebase. But any feedback outside apichanges.xml and the fact that the package is not exported would be appreciated (e.g about the API or where to put it etc).

mbien:
@jlahoda what is your opinion on the utility? Is the API ok? Right place?

mbien:
with the integration of auto-nb-javac based on javac 17, this became significantly less useful since client code can now use the enums without having to worry that they might not be there at runtime. The API would make version checks a bit more concise but most of the ugly init code could be removed right now even without the introduction of a new API.

PR #3275, opened 2021-10-27
Filter jvm-options based on VM name

jGauravGupta:
Payara Platform 5.2021.8 support HotSwap Agent via DCEVM which is activated by jvm-options in domain.xml.

With the introduction of the HotSwap feature, JVM Options in domain.xml need to be filtered out based on the JVM name.

For e.g: In the following snippet Azul is the vendor name and Dynamic Code Evolution is the VM name both JVM options should be activated based on the Vendor or VM name.

````
<jvm-options>[Azul-1.8.0u222|1.8.0u260]-XX:+UseOpenJSSE</jvm-options>
<jvm-options>[9|]-Djdk.attach.allowAttachSelf=true</jvm-options>
<jvm-options>[Dynamic Code Evolution-11.0.10|]-XX:HotswapAgent=core</jvm-options>
<jvm-options>[Dynamic Code Evolution-11.0.10|]-Xlog:redefine+class*=info</jvm-options>
````

### DIFF
--- a/enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/data/JDKVersion.java
+++ b/enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/data/JDKVersion.java
@@ -47,6 +47,11 @@ public final class JDKVersion {
      */
     private final Optional<String> vendor;
 
+    /**
+     * JDK vm
+     */
+    private final Optional<String> vm;
+
     private final static int MAJOR_INDEX = 0;
     private final static int MINOR_INDEX = 1;
     private final static int SUBMINOR_INDEX = 2;
@@ -58,21 +63,23 @@ public final class JDKVersion {
 
     private static final Short DEFAULT_VALUE = 0;
 
-    private JDKVersion(String version, String vendor) {
+    private JDKVersion(String version, String vendor, String vm) {
         short[] versions = parseVersions(version);
         this.major = versions[MAJOR_INDEX];
         this.minor = Optional.ofNullable(versions[MINOR_INDEX]);
         this.subminor = Optional.ofNullable(versions[SUBMINOR_INDEX]);
         this.update = Optional.ofNullable(versions[UPDATE_INDEX]);
         this.vendor = Optional.ofNullable(vendor);
+        this.vm = Optional.ofNullable(vm);
     }
 
-    JDKVersion(Short major, Optional<Short> minor, Optional<Short> subminor, Optional<Short> update, Optional<String> vendor) {
+    JDKVersion(Short major, Optional<Short> minor, Optional<Short> subminor, Optional<Short> update, Optional<String> vendor, Optional<String> vm) {
         this.major = major;
         this.minor = minor;
         this.subminor = subminor;
         this.update = update;
         this.vendor = vendor;
+        this.vm = vm;
     }
 
     /**
@@ -112,12 +119,21 @@ public final class JDKVersion {
     }
 
     /**
-     * Get JDK Vendor.
+     * Get JDK Vendor name.
      *
      * @return JDK vendor.
      */
     public Optional<String> getVendor() {
         return vendor;
+    }
+
+    /**
+     * Get JDK VM name.
+     *
+     * @return JDK vm.
+     */
+    public Optional<String> getVM() {
+        return vm;
     }
 
     public boolean gt(JDKVersion version) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/data/JDKVersion.java [hunk 1]
```diff
@@ -241,15 +257,15 @@ public String toString() {
...
         if (version != null && version.matches(VERSION_MATCHER)) {
-            return new JDKVersion(version, null);
+            return new JDKVersion(version, null, null);
         } else {
             return null;
         }
     }
 
-    public static JDKVersion toValue(String version, String vendor) {
+    public static JDKVersion toValue(String version, String vendor, String vm) {
```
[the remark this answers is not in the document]
jGauravGupta: `JDKVersion ` is not used by other modules.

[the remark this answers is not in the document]
neilcsmith-net: @jGauravGupta well it was exposed to other modules, and is therefore caught by the sigtests. There is still a failure in JvmConfigReader. IMO the original signatures need adding back to match the old generated sigtests - that's what they're there for.  However, as it's friend-only exposure at the moment, there's an argument it can evolve.

enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/server/parser/JvmConfigReader.java
```diff
@@ -201,7 +201,7 @@ public boolean isMonitoringEnabled() {
     public static class JvmOption {
 
         public final String option;
-        public final Optional<String> vendor;
+        public final Optional<String> vendorOrVM;
```
neilcsmith-net, line 204: This is also failing sigtests -

```
Class org.netbeans.modules.payara.tooling.server.parser.JvmConfigReader$JvmOption

  "E1.2 - API type removed" : field public final java.util.Optional org.netbeans.modules.payara.tooling.server.parser.JvmConfigReader$JvmOption.vendor

  warn: "W5.6 - Adding fields" : field public final java.util.Optional org.netbeans.modules.payara.tooling.server.parser.JvmConfigReader$JvmOption.vendorOrVM
```

enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/data/JDKVersion.java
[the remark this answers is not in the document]
JaroslavTulach: Keep the original method and let it call the new one.

enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/server/parser/JvmConfigReader.java
JaroslavTulach: Keep the original field name. You may change its runtime meaning (that of course isn't going to be detected by signature tests, but only by failures in the runtime), or you may introduce new field and possibly deprecate the old one.

Lesson to learn: don't expose classes that aren't needed in your API. Don't expose fields, but rather getters. Chapter 5: [Don't Expose More Than You Want](http://wiki.apidesign.org/wiki/Do_Not_Expose_More_Than_You_Want).

enterprise/payara.tooling/test/unit/src/org/netbeans/modules/payara/tooling/data/JDKVersionTest.java
```diff
@@ -40,23 +40,23 @@
     public void parseJDKVersion() {
         Map<String, JDKVersion> jdkVersions = new HashMap<>();
         jdkVersions.put("1.8",
-                new JDKVersion((short) 1, Optional.of((short) 8), Optional.of((short) 0), Optional.of((short) 0), Optional.empty()));
+                new JDKVersion((short) 1, Optional.of((short) 8), Optional.of((short) 0), Optional.of((short) 0), Optional.empty(), Optional.empty()));
```
JaroslavTulach, line 43: Unrelated: I don't like `Optional` and these lines show why: wouldn't it be simpler, to just use `null` instead of `Optional.empty()` and `(short)0` instead of `Optional.of(...)`?

JaroslavTulach (changes requested):
Keep the original methods and field names.

enterprise/payara.tooling/test/unit/src/org/netbeans/modules/payara/tooling/data/JDKVersionTest.java
neilcsmith-net: Well, I like `Optional` when it's used as intended. I agree with you here!

neilcsmith-net:
Thanks @jGauravGupta

Sigtests passing. Looks good. Would prefer it squashed, but can merge as is.

@JaroslavTulach can you check changes meet your request, then I'll merge in.  Thanks!

neilcsmith-net (approved):

enterprise/payara.tooling/test/unit/src/org/netbeans/modules/payara/tooling/data/JDKVersionTest.java
jGauravGupta: This change will require refactoring many functions and due to signature issues deprecated functions need to be added, I will look into this in separate PR.

neilcsmith-net:
Well, looks good to me, and needs to merge now if it's going to be in 12.6, so merging ...

PR #3278, opened 2021-10-27
Build Java cluster on JDK17

mbien:
Changes required to build the java cluster on JDK17 (15 is the latest version NetBeans can be built with right now). I omitted the other clusters to keep the diff smaller.

This won't make the tests work - that would be a larger issue.

note:
this does not include the Gradle version upgrade required to build on JDK17, since 7.3 it is still in the rc phase.
see ~~https://github.com/apache/netbeans/pull/3322~~ https://github.com/apache/netbeans/pull/3326 for the gradle update

### DIFF
--- a/extide/o.apache.tools.ant.module/build.xml
+++ b/extide/o.apache.tools.ant.module/build.xml
@@ -35,7 +35,7 @@
                 <path path="${src-bridge.cp}"/>
             </classpath>
         </depend>
-        <javac srcdir="src-bridge" destdir="build/bridge-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6" target="1.6" includeantruntime="false">
+        <javac srcdir="src-bridge" destdir="build/bridge-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" release="8" includeantruntime="false">
             <classpath>
                 <path path="${src-bridge.cp}"/>
             </classpath>
--- a/harness/apisupport.harness/nbproject/project.properties
+++ b/harness/apisupport.harness/nbproject/project.properties
@@ -16,7 +16,7 @@
 # under the License.
 
 javac.compilerargs=-Xlint -Xlint:-serial
-javac.source=1.6
+javac.source=1.8
 jnlp.cp=\
     ${o.n.bootstrap.dir}/lib/boot.jar:\
     ${openide.modules.dir}/lib/org-openide-modules.jar:\
--- a/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
@@ -66,21 +66,30 @@ public class CustomJavac extends Javac {
 
     @Override
     public void execute() throws BuildException {
-        String tgr = getTarget();
-        if (tgr.matches("\\d+")) {
-            tgr = "1." + tgr;
-        }
-        if (!isBootclasspathOptionUsed()) {
-            setRelease(tgr.substring(2));
-        }
-        String src = getSource();
-        if (src.matches("\\d+")) {
-            src = "1." + src;
-        }
-        if (!JavaEnvUtils.isAtLeastJavaVersion(src)) {
-            log("Cannot handle -source " + src + " from this VM; forking " + maybeFork, Project.MSG_WARN);
-            super.setFork(true);
-            super.setExecutable(maybeFork);
+        String release = getRelease();
+        if (release == null || release.isEmpty()) {
+            String tgr = getTarget();
+            if (tgr.matches("\\d+")) {
+                tgr = "1." + tgr;
+            }
+            if (!isBootclasspathOptionUsed()) {
+                setRelease(tgr.substring(2));
+            }
+            String src = getSource();
+            if (src.matches("\\d+")) {
+                src = "1." + src;
+            }
+            if (!JavaEnvUtils.isAtLeastJavaVersion(src)) {
+                log("Cannot handle -source " + src + " from this VM; forking " + maybeFork, Project.MSG_WARN);
+                super.setFork(true);
+                super.setExecutable(maybeFork);
+            }
+        } else {
+            if (!JavaEnvUtils.isAtLeastJavaVersion(release)) {
+                log("Cannot handle -release " + release + " from this VM; forking " + maybeFork, Project.MSG_WARN);
+                super.setFork(true);
+                super.setExecutable(maybeFork);
+            }
         }
         generatedClassesDir = new File(getDestdir().getParentFile(), getDestdir().getName() + "-generated");
         if (!usingExplicitIncludes) {
@@ -205,7 +214,7 @@ public class CustomJavac extends Javac {
         File d = getDestdir();
         if (!d.isDirectory()) {
             return;
-}
+        }
         FileSet classes = new FileSet();
         classes.setDir(d);
         classes.setIncludes("**/*$*.class");
--- a/webcommon/libs.jstestdriver/build.xml
+++ b/webcommon/libs.jstestdriver/build.xml
@@ -25,7 +25,7 @@
 
     <target name="ext-compile" depends="build-init">
         <mkdir dir="build/extclasses"/>
-        <javac srcdir="extsrc" destdir="build/extclasses" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6" target="1.8">
+        <javac srcdir="extsrc" destdir="build/extclasses" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" release="8">
             <classpath>
                 <path path="${module.classpath}"/>
                 <path location="${cluster}/${module.jar}"/>
--- a/apisupport/apisupport.ant/test/unit/src/org/netbeans/modules/apisupport/project/queries/SourceLevelQueryImplTest.java
+++ b/apisupport/apisupport.ant/test/unit/src/org/netbeans/modules/apisupport/project/queries/SourceLevelQueryImplTest.java
@@ -37,7 +37,7 @@ public class SourceLevelQueryImplTest extends TestBase {
         String path = "java/junit/src/org/netbeans/modules/junit/api/JUnitSettings.java";
         FileObject f = nbRoot().getFileObject(path);
         assertNotNull("found " + path, f);
-        assertEquals("1.6 used for an average module", "1.6", SourceLevelQuery.getSourceLevel(f));
+        assertEquals("1.8 used for an average module", "1.8", SourceLevelQuery.getSourceLevel(f));
     }
     
 }
--- a/harness/apisupport.harness/build.xml
+++ b/harness/apisupport.harness/build.xml
@@ -50,7 +50,7 @@
 
     <target name="compile-jnlp-launcher" depends="init,compile">
         <mkdir dir="${build.dir}/jnlp-launcher-classes"/>
-        <nb-javac srcdir="jnlp-src" destdir="${build.dir}/jnlp-launcher-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6" target="1.6" includeantruntime="false">
+        <nb-javac srcdir="jnlp-src" destdir="${build.dir}/jnlp-launcher-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" release="8" includeantruntime="false">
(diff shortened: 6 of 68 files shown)


lkishalmi:
I'd close this one in favor of #3322

mbien:
@lkishalmi please reconsider, this PR is not about gradle

lkishalmi:
My bad, have not checked the list of changes.

mbien:
@lkishalmi would you like to review this? This would make the java cluster buildable together with the recently merged gradle PR.

mbien:
@matthiasblaesing @neilcsmith-net would be great to get this in. Any reviewer volunteers? easy review

`ant -Dcluster.config=basic` should finally build after this.

harness/apisupport.harness/nbproject/project.properties
```diff
@@ -16,7 +16,8 @@
 # under the License.
 
 javac.compilerargs=-Xlint -Xlint:-serial
-javac.source=1.6
+javac.source=1.8
+javac.target=1.8
```
matthiasblaesing, line 20: You don't need to set `javac.target` - @JaroslavTulach ensured this in https://github.com/apache/netbeans/commit/4661771. The build system ensures, that `javac.target` is set to the value of `javac.source` is the former is not set.

mbien: ok. but it doesn't hurt right? does everything use the same build xml which has this functionality? having source and target at different values can cause trouble if its missed somewhere. (ultimately we should aim to use --release in distant future)

i can remove it if you prefer it gone.

matthiasblaesing: You can see the mechanic in the `build.xml`:

https://github.com/apache/netbeans/blob/61795d673369f587d2ae2f50254b442a407b64c0/ide/css.editor/build.xml#L22-L25

The build looks simple, but truth is, that it just inherits its power from:

-  `nbbuild/templates/projectized.xml` (direct)
-  `nbbuild/templates/common.xml` (transitive)
-  `nbbuild/default.xml`
-  `nbbuild/default-properties.xml`
-  `nbbuild/jdk.xml`
-  `nbbuild/binaries-default-properties.xml`

The `javac.source` and `javac.target` default values are set by the `basic-init` target in `nbbuild/templates/common.xml`. 

From the default target `build` it is reached via:

- `build`
- `netbeans
- `-netbeans`
- `init`
- `basic-init`

For using `--release`, this was also tackled by @JaroslavTulach (I think it is in the same PR) - see here:

https://github.com/apache/netbeans/blob/96720b9d39122e514fe21536a831247e5dc17b10/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java#L68-L91

This works only for normal module builds, special cases, where the ant `javac` task is called directly, adjustments need to be made there.

NetBeans build is a complex beast, but it some very nice things in it, at the price that its complex.

For the `javac.target`, yes I think it would be good to not add them, else next time, someone will argue, that the two should be adjusted independently from each other.

mbien: thanks for the info. will remove it.

mbien: done, removed `javac.target` from all properties I touched, even if it was there before.

I kept the changes to build xmls since they all had target specified.

mbien: it doesn't build. looks like i have to add a few targets again.

mbien: ok, it builds again on JDK 17.

please note that this branch is fairly old and i don't want to abuse the CI too much just to rebase and push again. So to get the JDK 17 gradle changes do a
`
git pull --rebase git@github.com:apache/netbeans.git master`

after you checked out this branch.

extide/o.apache.tools.ant.module/build.xml
```diff
@@ -35,7 +35,7 @@
                 <path path="${src-bridge.cp}"/>
             </classpath>
         </depend>
-        <javac srcdir="src-bridge" destdir="build/bridge-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6" target="1.6" includeantruntime="false">
+        <javac srcdir="src-bridge" destdir="build/bridge-classes" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.8" target="1.8" includeantruntime="false">
```
JaroslavTulach, line 38: Consider adding `release="8"` attribute or using `nb-javac` which is supposed to do that automatically.

mbien: 8 doesn't know `--release`. So i would have to do both, switch to nb-javac and replace both attributes with the release attribute i suppose.

Could we leave this as is until we move to JDK 11+? After that everything will be easier. I am not sure if its worth to put a lot of energy into this, since its only the basic cluster - i expect a lot more changes. It doesn't look like JDK will drop support for source/target any time soon and it does currently allow NB to be built on JDK 17.

Once NB dropped 8 support it would be a fairly simple change. Remove the build support logic Matthias mentioned and switch everything to `javac.release`.

neilcsmith-net: @mbien the Ant compiler task ignores `release` on JDK 8.

mbien: @neilcsmith-net interesting! didn't know that. 

plus: "When set and running on JDK 9+ the source and target attributes as well as the bootclasspath will be ignored. Since Ant 1.9.8."

which means we can set all three to the same value? I would be ok with this.

neilcsmith-net: @mbien yes, as far as I know. Even an argument that only release is necessary, as the only JDK that's going to ignore it will default to source and target being 8 anyway.  But being specific is probably better!

mbien: pushed update:
 - `release` requires the new versioning scheme on modern JDKs -> so i have to use 8 instead of 1.8
 - nb-javac throws NPEs when only release is set -> added source and target there too
 - javac now uses only release on all xml files i touched for this PR
 - exception: debugger.jpda.ant can't use release without it causing compiler errors -> kept source/target
 - rebased ontop of latest master, no need to pull gradle changes anymore

JaroslavTulach: > set all three to the same value

Looks like Ant's approach is better than Maven's one (fail on JDK8 when release is used) when shielding developers from JDK differences.

> nb-javac throws NPEs when only release is set -> added source and target there too

You can fix the task in `nbbuild/antsrc`...

JaroslavTulach (approved):
I like usage of `release="8"`!

extide/o.apache.tools.ant.module/build.xml
mbien: > You can fix the task in nbbuild/antsrc...

yes i do want to take a look later. We shouldn't have a configuration divergence between the two javac tasks if its avoidable.

mbien: added 1d773e0

webcommon/libs.jstestdriver/build.xml
```diff
@@ -25,7 +25,7 @@
 
     <target name="ext-compile" depends="build-init">
         <mkdir dir="build/extclasses"/>
-        <javac srcdir="extsrc" destdir="build/extclasses" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" source="1.6" target="1.8">
+        <javac srcdir="extsrc" destdir="build/extclasses" deprecation="${build.compiler.deprecation}" debug="${build.compiler.debug}" release="8">
```
matthiasblaesing, line 28: If I read the ant source correctly, this is not correct on JDK 8. You need to set `source`, `target` and `release` to get the desired behavior:

- on JDK 8, `release` will be ignored, as it is invalid for the compiler
- on JDK 9+, `source` and `target` are ignored

mbien: in theory yes. but this would be only a problem if we would compile on jdk 7 or lower too. 8 would have src and target at 8 by default anyway. Everything else can use release already. @neilcsmith-net pointed this out which i didn't consider either initially.

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
```diff
@@ -66,21 +66,30 @@ public void setIncludes(String includes) {
 
     @Override
     public void execute() throws BuildException {
-        String tgr = getTarget();
```
matthiasblaesing, line 69: Why is this necessary?  The original implementation moved target to release - why not keep that behavior?

mbien: it still does that if release isn't set, scroll down.

matthiasblaesing: No - why setting release? It is not necessary. The original change ensured, that target is transferred to release, unless the bootclasspath is set.

mbien: https://github.com/apache/netbeans/pull/3278#discussion_r760538319

mbien: + https://github.com/apache/netbeans/pull/3278#discussion_r760775688

matthiasblaesing: Not an answer

mbien: this change ensures that javac and nb-javac do not diverge configuration wise, as mentioned in the linked conversation.

mbien: what is the combination of attributes/JDKs which should be supported but isn't after that change?

matthiasblaesing:
Please check the failing unittests - given the error message:

```
    [junit] INFO: println: /home/travis/build/apache/netbeans/nbbuild/netbeans/harness/common.xml:207: Compile failed; see the compiler error output for details. listener: Hyperlink[file:/home/travis/build/apache/netbeans/nbbuild/netbeans/harness/common.xml:207:-1:-1:-1]
    [junit] Dec 02, 2021 8:37:52 PM org.netbeans.modules.apisupport.project.InputOutputProviderImpl$OW write
    [junit] INFO: BUILD FAILED (total time: 1 second)
```

This might indeed a problem.

mbien:
i see many of those messages on a build which doesn't fail. E.g taken from latest master build:
https://app.travis-ci.com/github/apache/netbeans/jobs/550425157#L9197-L9199

its difficult to figure out what is failing in a build log full with "build failed" unfortunately

matthiasblaesing:
Thank you - travis and github actions are happy now. Would you please squash the commits into a single one?

mbien:
done. Maybe @JaroslavTulach would want to take another look since he approved before I changed the ant task.

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
```diff
@@ -66,21 +66,30 @@ public void setIncludes(String includes) {
...
-        String src = getSource();
-        if (src.matches("\\d+")) {
-            src = "1." + src;
-        }
-        if (!JavaEnvUtils.isAtLeastJavaVersion(src)) {
-            log("Cannot handle -source " + src + " from this VM; forking " + maybeFork, Project.MSG_WARN);
-            super.setFork(true);
-            super.setExecutable(maybeFork);
+        String release = getRelease();
+        if (release == null || release.isEmpty()) {
```
JaroslavTulach, line 70: Setting the `release` only when it is not set yet seems OK to me.

JaroslavTulach (approved):

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
mbien: thanks for taking a look jaroslav!

matthiasblaesing:
Ok - travis is happy, lets get this in.

JaroslavTulach:
FYI: [Jenkins VSCode build #739](https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/739/) failed and continues to fail. Coincidence?

matthiasblaesing:
#739 fails, but #742 was green. Given, that the test where the timeout occurs is marked @RandomlyFails, a broken tests sounds like a reasonable explanation.

JaroslavTulach:
A coincidence, then. Sorry for bothering.

PR #3284, opened 2021-10-28
[jackpot] added generics aware rule to IteratorToFor inspection

mbien:
IteratorToFor.whileIdiom() only triggered on code without generics. Added another rule for the generic case.

All other rules in this inspection did take generics into account already, this one had it missing.


matthiasblaesing:
Change looks good - what do you think about adding a test for this case to IteratorForTest.java?

mbien:
Sure. Added a test for the generics aware rule.

matthiasblaesing:
The only unittest failing is testing a removed module - it is unclear why a hybrid of the commits is picked up (module removed, but test reference still present), but apart from that it looks sane.

PR #3160, opened 2021-09-10
ConvertTextBlockToString warning should be a hint/suggestion

mbien:
Text Blocks are in java since 15. Converted the inspection into a hint so that it doesn't show up as warning and moved it to the suggestion category.

PR #3287, opened 2021-10-30
CheckRegexp hint causes all lines with regexp shown with a warning

matthiasblaesing:
The severity of the CheckRegexp of VERIFIER causes all lines in source
files containing a regexp to be marked as warning. This is invalid, as
the regexp are perfectly ok.

The severity is reduced to HINT. That way the hint is readily available
when the regexp is under the cursor, but does not clutter the whole
file.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/jdk/CheckRegex.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/jdk/CheckRegex.java
@@ -38,6 +38,7 @@ import javax.swing.SwingUtilities;
 import org.netbeans.api.annotations.common.CheckForNull;
 import org.netbeans.api.java.source.CompilationInfo;
 import org.netbeans.spi.editor.hints.ErrorDescription;
+import org.netbeans.spi.editor.hints.Severity;
 import org.netbeans.spi.java.hints.ConstraintVariableType;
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
@mishrasandeep please have a look at this change

neilcsmith-net:
@matthiasblaesing +1 and happy to merge this one into delivery if you want (it's bugging me too!). There's a whole bunch of hints that should have their level reduced - the editor is getting far too noisy IMO.

mishrasandeep:
> 
> 
> @mishrasandeep please have a look at this change
Looks Good.

jkovalsky (approved):
I tested the change and it works fine. Approved. Thanks Matthias.

matthiasblaesing:
Thank you all for review. I have rebased this onto delivery, but I seem to have messed up the java.hints tests. I'll look into it first and if I get the tests green (or only see the known flaky tests fail), will merge.

mbien:
@matthiasblaesing you have to swap out `:verifier:` with `:hint:` in the test. Example is here: https://github.com/apache/netbeans/pull/3160

thanks for fixing this, I wanted to do that too but completely forgot.

matthiasblaesing:
@mbien thanks for the hint - I should have known, that there was a test, that verified this.

PR #3288, opened 2021-10-30
[NETBEANS-5734] Add details for Glassfish 6.2.1 server and improve version detection

matthiasblaesing:
With Glassfish 6.2.1 support for JDK 17 was added, so meta data for
that version was added.

The version detection was improved in the following way:

If an exact match is found, that entry is used. Else the list of
server definitions is iterated from highest to lowest. The highest
definition, where the version number is lower than or equal to the
target version is used to configure the server.

As an example: If you try to configure Glassfish 6.2.2, it will not
find an exact match. Instead 6.2.1 is matched. Prior to this change,
the configuration for 6.0 would have been chosen.

### DIFF
--- a/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/Bundle.properties
+++ b/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/Bundle.properties
@@ -160,6 +160,7 @@ STR_501_SERVER_NAME=GlassFish Server 5.0.1
 STR_510_SERVER_NAME=GlassFish Server 5.1.0
 STR_6_SERVER_NAME=GlassFish Server 6
 STR_610_SERVER_NAME=GlassFish Server 6.1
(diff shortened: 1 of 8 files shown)


matthiasblaesing:
@nicolaken @juneau001 could you please have a look at this? I based this on delivery and would consider it a bugfix, as currently glassfish 6.2.2 is detected as 6, which is broken.

juneau001 (approved):
These changes look sane and they will be a great improvement for those working with the most recent versions of GlassFish.

neilcsmith-net:
OK, merging - just the usual Travis suspect red. Please confirm the fix when RC2 is out and close the JIRA ticket, thanks.

nicolaken:
Yes, very nice, works well.
Thank you!

The more I look at this part of the code the more I wonder "what were they thinking" when structuring, as having to make so many changes in so many files for adding a version is not, ahem, optimal.

PR #3289, opened 2021-10-31
[NETBEANS-6115] UTF-8 input is broken in Maven project.

mbien:
Chars were written as bytes to a stream, ignoring the encoding.
By wrapping the stream in a writer, chars can now be encoded properly.

Chose to not specify the encoding so that the default can be used or
overwritten by -Dfile.encoding=UTF-8.

JDK 18+ uses UTF-8 as default, it will be probably best practice
to set the encoding to UTF-8 for most JDK 17 projects too, while using
the new 'native.encoding' property for edge cases.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
@@ -24,12 +24,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.io.Reader;
+import java.io.Writer;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


mbien:
windows build failure seems to be unrelated. Would be nice if someone could restart it.

lkishalmi:
Well, Travis looks, good. I've restarted the one failing job.

lkishalmi:
It seems to be trivial enough to be in 12.6. @mbien rebase this PR on delivery branch, so the RMs can vet this change to be included or not.

mbien:
going to have to force push since it doesn't rebase cleanly without also merging all the other commits

neilcsmith-net (approved):
Giving Travis one more go, but looks sane to me. Let's get it in 12.6.

Yes, doing spec versions earlier does make switching base branch from master to delivery a little more work.

PR #3290, opened 2021-10-31
[NETBEANS-6114] Restore automatic constructor completion functionality.

mbien:
After successful classtype completion, a new auto-completion should
appear for constructor selection, without having to press ctrl+space again.

Regression since 12.3.

took me a while to find this. Had to git bisect to find a9871c5085106f1ffd6c1c7ec594774185cf69d6 which introduced this small regression.

It seems to work fine, I keep this as draft until I did a bit more manual testing.

example:

```java
new ArrayL
```
press ctrl+space, enter
```java

new ArrayList<>
```
should show a completion window for the constructors without having to press ctrl+space again, since the completion is not finished.

### DIFF
--- a/ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
+++ b/ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
@@ -226,6 +226,10 @@ public final class CodeTemplateInsertHandler implements TextRegionManagerListene
         }
         // Insert the template into document
         insertTemplate();
+
+        if (masterParameters.stream().noneMatch(param -> param.isEditable())) {
+            SwingUtilities.invokeLater(Completion.get()::showCompletion);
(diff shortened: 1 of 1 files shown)


mbien:
everything seems to be working fine. I went through the reproducer steps of the issue (https://issues.apache.org/jira/browse/NETBEANS-4646) whose fix caused the regression and it worked too as far as I can see.

@neilcsmith-net if this is a potential candidate for 12.6 I can rebase if needed.

neilcsmith-net:
Yes, please rebase on `delivery`, good candidate for a fix IMO. Asking @lkishalmi and @arsadykov to have a look here.  Adding `do not merge` label just until it's rebased and reviewed.

Curiously, when I first checked the bug described I happened to use a line with `var` and it seems to work OK there?!

lkishalmi (approved):
Looks sane as far as I can tell.

mbien:
@neilcsmith-net rebased.

thanks for review everyone

PR #3291, opened 2021-11-01
Handle generated sources during native image debugging.

entlicher:
The native image debugger needs to be aware of project's sources to resolve the paths correctly.
We use `JavaProjectConstants.SOURCES_TYPE_JAVA` already, but that does not include generated sources. We need to add `JavaProjectConstants.SOURCES_TYPE_GENERATED` to be able to recognize generated sources.

### DIFF
--- a/java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/displayer/JavaFrameDisplayer.java
+++ b/java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/displayer/JavaFrameDisplayer.java
@@ -34,6 +34,7 @@ import org.netbeans.api.java.queries.SourceForBinaryQuery;
 import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectUtils;
 import org.netbeans.api.project.SourceGroup;
+import org.netbeans.api.project.Sources;
 import org.netbeans.modules.nativeimage.api.debug.NIFrame;
 import org.netbeans.modules.nativeimage.spi.debug.filters.FrameDisplayer;
 import org.netbeans.spi.java.classpath.PathResourceImplementation;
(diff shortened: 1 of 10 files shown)


java/maven/src/org/netbeans/modules/maven/classpath/MavenSourcesImpl.java
```diff
@@ -78,7 +78,7 @@
 public class MavenSourcesImpl implements Sources, SourceGroupModifierImplementation, OtherSourcesExclude {
     public static final String TYPE_OTHER = "Resources"; //NOI18N
     public static final String TYPE_TEST_OTHER = "TestResources"; //NOI18N
-    public static final String TYPE_GEN_SOURCES = "GeneratedSources"; //NOI18N
```
sdedic, line 81: This may lead to a possible incompatibility, if someone looks for "GeneratedSources". Would it be possible to handle both type strings ?

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/displayer/JavaFrameDisplayer.java
```diff
@@ -161,35 +162,44 @@ private static ClassPath findSourcePath(Project project) {
...
+            if (ecp != null) {
+                FileObject[] binaryRoots = ecp.getRoots();
+                for (FileObject fo : binaryRoots) {
+                    if (addedBinaryRoots.contains(fo)) {
+                        continue;
+                    }
+                    addedBinaryRoots.add(fo);
+                    FileObject[] roots = SourceForBinaryQuery.findSourceRoots(fo.toURL()).getRoots();
+                    for (FileObject fr : roots) {
+                        if (!preferredRoots.contains(fr)) {
```
sdedic, line 193: It seems that `preferredSourceRoots` is just an optimization for `contains` -- what if the `allSourceRoots` would be a `LinkedHashMap` ?

java/java.project/src/org/netbeans/api/java/project/JavaProjectConstants.java
```diff
@@ -46,6 +46,13 @@ private JavaProjectConstants() {}
      */
     public static final String SOURCES_TYPE_MODULES = "modules"; // NOI18N
 
+    /**
+     * Generated root sources type.
+     * @see org.netbeans.api.project.Sources
+     * @since 1.88
+     */
+    public static final String SOURCES_TYPE_GENERATED = "generated"; // NOI18N
```
sdedic, line 54: Note that other languages may also generate sources; e.g. in Gradle build, one can easily have `build/generated-sources/groovy/...` or kotlin. There's sadly not an is-a formal hierarchy among source types, although Gradle/Groovy support provides one (retursn Groovy source root for "java" query) -  but maybe we could declare "java-resources" from the start here ?

@JaroslavTulach  ?

JaroslavTulach (commented):
> We need to add JavaProjectConstants.SOURCES_TYPE_GENERATED to be able to recognize generated sources.

Can you please describe this use-case and how it shall be realized in the proposed API? Feel free to add it to Javadoc or `arch.xml`. What shall one do when searching for generated Java sources? What shall one do when searching for generated JavaScript/PHP sources?

matthiasblaesing:
The target milestone was moved twice, there is a merge conflict and unanswered comments, so removing NB15 milestone and not setting a new one.

PR #3292, opened 2021-11-01
[NETBEANS-6068] Add back SOURCES_TYPE_JAVA into GroovySources

neilcsmith-net:
[NETBEANS-6068] Add back SOURCES_TYPE_JAVA into GroovySources::getGroovySourceGroups

This change added as part of [NETBEANS-4975](https://issues.apache.org/jira/browse/NETBEANS-4975) seems to be a cause of some issues reported in [NETBEANS-6068](https://issues.apache.org/jira/browse/NETBEANS-6068).

@lkishalmi could you take a look at this, which reverts a part of #2507 I'm not sure of the reasoning for this change, and I don't use Groovy much, but as it was raised as a potential blocker for 12.6 I had a look.  This seems to at least to be the cause of some exceptions mentioned in NETBEANS-6068

### DIFF
--- a/groovy/groovy.support/src/org/netbeans/modules/groovy/support/api/GroovySources.java
+++ b/groovy/groovy.support/src/org/netbeans/modules/groovy/support/api/GroovySources.java
@@ -22,6 +22,7 @@ package org.netbeans.modules.groovy.support.api;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.netbeans.api.java.project.JavaProjectConstants;
 import org.netbeans.api.project.SourceGroup;
 import org.netbeans.api.project.Sources;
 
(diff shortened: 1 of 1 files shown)


lkishalmi (approved):
Fine by me. This might cause some quirks on Gradle Groovy projects (or gradle project that have groovy sources as well), but certainly would not cause a function loss.
I have to check how shall we properly support Ant and Maven projects.

neilcsmith-net:
What quirks exactly? I guess I don't understand what this change was trying to do, and I don't know this area well enough. And what exactly needs proper support in Ant and Maven?  I noticed while debugging that Maven hits a bunch of Maven-specific Groovy support, while Gradle hits some Ant bits?

lkishalmi:
What i can think of is Creating a Groovy class in the wizard would display java folders as source root. Maybe duplicate entries in source root selection. Nothing big, I hope.
In Gradle and Maven src/java/main and src/groovy/main can co-exists. It is ok to create a java file in the groovy folders, but it is not to create a groovy file in the java folders.

neilcsmith-net:
OK, makes sense. But as the wizard refuses to display at all in certain circumstances* without reverting this, I'll merge and add back for next rc, and maybe we can figure out what causes the problem for 13?

* the source roots passed in is completely empty without this so an exception is thrown.

PR #3293, opened 2021-11-01
[NETBEANS-5846] Minimal support of java-platfom Gradle projects.

lkishalmi:
Slightly more than just disabling the warnings on java-platform Gradle projects.
Did some testing on the weekend including the test why api and some other configurations are not showing up in the UI.
Those are the configurations the users often use. It turned out that Gradle marks those configurations as non-resolvabe (canBeResolved property is false).
So I made some adjustments to let them displayed.
Before:
image

After:
image

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
@@ -29,7 +29,8 @@ import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.ResolvedDependency
 import org.gradle.api.artifacts.ResolveException
 import org.gradle.api.artifacts.FileCollectionDependency
-import org.gradle.api.artifacts.ExternalModuleDependency
+import org.gradle.api.artifacts.ModuleDependency
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.component.ModuleComponentSelector
 import org.gradle.api.artifacts.component.ProjectComponentSelector
 import org.gradle.api.artifacts.result.ArtifactResult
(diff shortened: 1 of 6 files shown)


neilcsmith-net:
Sorry, missed this was opened against delivery as missing a milestone.  Intended for 12.6?  I've only given it a glance.  If you think it needs to go in RC2, please feel free to merge before tomorrow.

extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
```diff
@@ -341,10 +374,18 @@ class NbProjectInfoBuilder {
...
             def unresolvedIds = []
             def projectNames = []
             long time_inspect_conf = System.currentTimeMillis()
+            
+            it.dependencies.withType(ModuleDependency) {
+                def group = it.group != null ? it.group : '';
+                def name = it.name
+                def version = it.version != null ? ':' + it.version : ''
+                def id = group + ':' + name + version;
+                componentIds.add(id)
```
sdedic, line 387: is it OK to pass on malformed component IDs, e.g. no name or no group ?

extide/gradle/src/org/netbeans/modules/gradle/nodes/ConfigurationsNode.java
```diff
@@ -134,8 +134,12 @@ public Image getOpenedIcon(int type) {
...
     @Override
     public String getShortDescription() {
-        GradleProject gp = project.getGradleProject();
-        return gp.getBaseProject().isResolved() ? Bundle.HINT_ConfigurationsNode() : Bundle.HINT_ConfigurationsNodeUnresolved();
+        return !needsResolve() ? Bundle.HINT_ConfigurationsNode() : Bundle.HINT_ConfigurationsNodeUnresolved();
+    }
+
+    private boolean needsResolve() {
+        GradleBaseProject gbp = GradleBaseProject.get(project);
+        return !gbp.isResolved() && !gbp.hasPlugins("java-platform"); //NOI18N
```
sdedic, line 142: Hm, this means that if a configuration is NOT resolved , but the project does use `java-platform` plugin - the project  as a whole ceases needing to be resolved ?

neilcsmith-net:
I'm going to sync and trigger RC2 now, so given review in progress, please don't merge now.  We'll likely have an RC3 so we can merge for that once review process complete. Thanks!

neilcsmith-net:
@sdedic @lkishalmi what's the status of review here now?  Thanks!

lkishalmi:
Let's have this on early 13.0 instead. I'm going to rebase this on master.

extide/gradle/netbeans-gradle-tooling/src/main/groovy/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.groovy
lkishalmi: No group, no version names are valid.

extide/gradle/src/org/netbeans/modules/gradle/nodes/ConfigurationsNode.java
lkishalmi: Well it is about the warning badge and the tooltip. We put warning the configuration is not resolved in general.
However java-platform project can't be resolved by definition. So we do not place warning badge and unresolved config hint on the node.

lkishalmi:
I'm merging this to the master, let's se how this works out

PR #3294, opened 2021-11-02
git show history should show results without having to press search

mbien:
It was limited to files, now it works on folders/projects too.


mbien:
@neilcsmith-net same here. another small one.

neilcsmith-net:
Will depend if we need a third RC.

PR #3295, opened 2021-11-02
Splitting workflows into individual files

JaroslavTulach:
It always drives me mad, when one of the GitHub Action tests fails. There is currently only one button to re-run all of them. However I'd rather re-run just the failed one. If we split the GitHub actions into individual files, can I be able to restart only one of them? Looks like we will...


ebarboni:
Like the idea very much. No side issue ?

neilcsmith-net (approved):
Like @ebarboni I also like this idea very much. Although GH Actions not quite as annoying as Travis .. yet. Definitely a good idea if we move more stuff over and/or start revising what runs on what triggers.

mbien:
this might not be needed if we wait a little: https://github.com/github/roadmap/issues/271

PR #3297, opened 2021-11-02
[NETBEANS-5962] Remove default bgcolor for multiline Strings

neilcsmith-net:
[NETBEANS-5962](https://issues.apache.org/jira/browse/NETBEANS-5962) Remove default bgcolor for multiline Strings - inherited and illegible on various editor themes. This was reported against FlatLaf Dark but it affects others too. This is a quick fix for 12.6 - need to consider how to display indents and make sure they work across all editor themes, in particular handling dark ones.

cc/ @arusinha @jlahoda

### DIFF
--- a/java/java.editor/src/org/netbeans/modules/java/editor/resources/fontsColors.xml
+++ b/java/java.editor/src/org/netbeans/modules/java/editor/resources/fontsColors.xml
@@ -90,7 +90,7 @@
     <fontcolor name="mod-protected" />
     <fontcolor name="mod-public" />
     <fontcolor name="mod-keyword" default="keyword" />
-    <fontcolor name="mod-unindented-text-block" bgColor="EEDDDD" />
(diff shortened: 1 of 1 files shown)


jlahoda:
Thanks, I think this is fine for now. I was thinking of improving the visualization - maybe we should try a rectangle around the real content of the text block. Or (at least) a line at the left edge of the text.

neilcsmith-net:
> Or (at least) a line at the left edge of the text.

Yes, my thoughts too. In fact, think left edge possibly better than rectangle.  I couldn't see a quick fix to enable that?  Tried with prefix text and a pipe (`|`) - looked terrible :smile:

The one background colour that might work is to inherit from `highlight-caret-row` - that should always be legible, and not have the strange flash as you move across the text?

mbien:
thanks for fixing this. it was blinding me in some files :) the jshell window has a similar problem btw

PR #3299, opened 2021-11-03
[NETBEANS-5209] Document switcher popup not grouping by project on first use.

mbien:
issue:
1) tools -> options -> appearance -> doc tabs -> check sort opened docs by project
2) open some files, open the tab switcher popup on the upper right corner of the editor
3) you will notice first open does not group by projects as requested
4) open it again... should work now

fix for document switcher popup not grouping by project on first use:
- file owner query is now a blocking opperation with timeout
- under normal conditions the query returns a result right away
- if it doesn't for whatever reason it won't block the EDT due to the timeout
  - fallback is the regular tab list
- it fixes the issue as side effect + avoids repaint flickering of the original impl

### DIFF
--- a/ide/core.multitabs.project/nbproject/project.properties
+++ b/ide/core.multitabs.project/nbproject/project.properties
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-javac.source=1.6
+javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial
 nbm.needs.restart=true
 spec.version.base=1.25.0
--- a/ide/core.multitabs.project/src/org/netbeans/core/multitabs/project/ProjectSupportImpl.java
+++ b/ide/core.multitabs.project/src/org/netbeans/core/multitabs/project/ProjectSupportImpl.java
@@ -44,7 +44,7 @@ import org.openide.windows.TopComponent;
 @ServiceProvider(service = ProjectSupport.class)
 public class ProjectSupportImpl extends ProjectSupport {
 
-    private static final Map<FileObject, Project> file2project = new WeakHashMap<FileObject, Project>(50);
+    private static final Map<FileObject, Project> file2project = new WeakHashMap<>(50);
     private static final RequestProcessor RP = new RequestProcessor("TabProjectBridge"); //NOI18N
     private static final ChangeSupport changeSupport = new ChangeSupport(RP);
     private static PropertyChangeListener projectsListener;
@@ -62,12 +62,7 @@ public class ProjectSupportImpl extends ProjectSupport {
         synchronized( changeSupport ) {
             changeSupport.addChangeListener(l);
             if( null == projectsListener ) {
-                projectsListener = new PropertyChangeListener() {
-                    @Override
-                    public void propertyChange(PropertyChangeEvent evt) {
-                        changeSupport.fireChange();
-                    }
-                };
+                projectsListener = (PropertyChangeEvent evt) -> changeSupport.fireChange();
                 OpenProjects.getDefault().addPropertyChangeListener( projectsListener );
             }
         }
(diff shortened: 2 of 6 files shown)


KacerCZ:
@mbien I filed this issue some time ago as https://issues.apache.org/jira/browse/NETBEANS-5209

KacerCZ:
@mbien I tested this fix with PHP project - it fixes first use after loading IDE.
Newly opened files are still sorted under "no project" and switcher has to be closed and opened again.

mbien:
@KacerCZ ah interesting. Going to take a look at this case tomorrow. I might just add another commit which removes the async code. I could probably test it by putting a project on a usb drive or something similar. 

Thanks for linking the issue to the PR.

mbien:
ca3c620  has the async logic removed. Updated PR text.

KacerCZ:
I'm happy to confirm that after last commit the issue seems to be fixed.

mbien:
rebased onto delivery, since I saw there will be a RC 3

neilcsmith-net:
@mbien I'm not sure about integrating this in to the 12.6 release, but also not investigated the changes in detail? Does this move the project resolution via FileOwnerQuery on to the EDT?

mbien:
> @mbien I'm not sure about integrating this in to the 12.6 release, but also not investigated the changes in detail. Does this move the project resolution via FileOwnerQuery on to the EDT?

yes. but at this point everything should be already in the cache. I had about 100 projects open and opened one file per project and the switcher appeared basically right away after IDE start.

If you look at the doc of FileOwnerQueryImplementation:
https://github.com/apache/netbeans/blob/c084119009d2e0f736f225d706bc1827af283501/ide/projectapi/src/org/netbeans/spi/project/FileOwnerQueryImplementation.java#L27-L29


If it is still determined that it still needs to be off EDT i would probably try to decouple it a few layers higher, e.g https://github.com/apache/netbeans/blob/cf6c1a8d4ec0e2faca0b6143cf8cef928bc00d94/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java#L455

to avoid getting barraged by property change events fired from inner loops.

neilcsmith-net:
Yes, I read the docs and looked at a couple of the implementations. I'm not sure if there might be edge cases that could lock up the EDT or not here. I presume it was made async for a reason. I also agree that moving up a few layers would make more sense.

In terms of 12.6, going to request review - I'm not making the call on whether to merge, but it seems a slightly risky one to me at this stage when we're hoping this will be the final release.

mbien:
no need to rush anything. I rebased most of my bugfix PRs just in case someone wants to integrate them while i am not available. Lets get 12.6 released ;)

neilcsmith-net:
OK, punting this to NB13 as no-one has reviewed.

netbeansuser2019:
Great, finally you pick NETBEANS-3752 up.

KacerCZ:
I was using NetBeans with this fix daily for one week without any problems.

mbien:
changed the impl as previously discussed to make it less risky. (decoupled from EDT on a higher level, one task for all tabs with timeout)

mbien:
@matthiasblaesing @neilcsmith-net @timboudreau any opinions on this. The code which unblocks the EDT is here

https://github.com/apache/netbeans/commit/92619c994a009165c7c37da441a81203b34701b6#diff-793eded4bbc5a3b7a16c8d300f53f467eb7538137d6d7338a36c433cf11473daR80

The task is usually done in <10ms. Fallback would be no groups in the popup on time out - but this does not happen under normal operation. This is purely there to ensure that the EDT can in fact not be blocked indefinitely, without having to review the entire implementation behind the project lookup logic for this minor fix (there is a lot of caching going on there).

neilcsmith-net:
@mbien I say, feel free to merge it so we get it in 13-rc1.  I would normally be inclined to trigger back to the EDT via invokeLater with something like this.  But here, this should be fine, and I suspect with any open tabs the information is already calculated?

PR #3300, opened 2021-11-03
[jackpot] for-loop to function should be an inline hint instead of a warning

mbien:
changed the default to hint at current line.


mbien:
@neilcsmith-net I based it on delivery right away. No pressure though :)

neilcsmith-net:
Will depend if we need a third RC.

neilcsmith-net:
Retrying a few failing tests. One of the fails not a usual suspect, but in hints.  Don't think it's related, but holding fire on merging to see if it passes.

PR #3301, opened 2021-11-04
An attempt to fix unstable test for java.mx.projects

dbalek:
Invalidate existing Sources after reindexing.


JaroslavTulach (approved):
Thanks for fixing a randomly failing test! It is an important activity as #3298 discussion shows.

PR #3302, opened 2021-11-04
[NETBEANS-6171] Interpret ClassCast as "interface unsupported"

sdedic:
According to [GraalSDK documentation](https://docs.oracle.com/en/graalvm/enterprise/20/sdk/org/graalvm/polyglot/Value.html#as-java.lang.Class-), the `as` method may throw ClassCastException if the passed object does not fit into the supported cases; the [Invocable.getInterface](https://docs.oracle.com/javase/7/docs/api/javax/script/Invocable.html#getInterface(java.lang.Object,%20java.lang.Class)) however returns `null` in such case.


JaroslavTulach (approved):

neilcsmith-net:
@sdedic if it's open against master, it's not going in 12.6.

sdedic:
Rebased & retargetted against `delivery`

PR #3303, opened 2021-11-04
Consider the reported trustlevel of the KeyStoreProvider when querying KeyStores

matthiasblaesing:
Utilities#getKeyStore needs to check the supplied trustlevel against
the trustlevel reported by the KeyStoreProvider. Else all keystores
are returned for all levels and thus all KeyStores are reported on
level TRUST.

### DIFF
--- a/platform/autoupdate.services/build.xml
+++ b/platform/autoupdate.services/build.xml
@@ -84,14 +84,18 @@
         <delete file="${validation-store}" />
         <genkey
             keystore="${validation-store}"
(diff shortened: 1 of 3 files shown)


mbien:
LGTM

https://github.com/apache/netbeans/pull/3278 has another place where keyalg needs to be set which i found while trying to make the java cluster build on jdk17.

matthiasblaesing:
@mbien thank you for the review.

PR #3307, opened 2021-11-07
[NETBEANS-5783] Make sure that getDisplayName() is non null in Gradle Projects.

lkishalmi:
Tried to reproduce the behavior with tests, but was not able to. I've closed some possible paths making sure that the Gradle Project does not return ```null``` as DisplayName or Name.


neilcsmith-net (approved):
Thanks. Let's see if it removes some of those problems.

PR #3311, opened 2021-11-09
Correct the parsing of String values in native image.

entlicher:
In GraalVM native image debugging, parsing of String from the native image data is corrected for JDK 9+


neilcsmith-net (approved):
OK, I'll merge once @JaroslavTulach has reviewed. RC3 will be a few days late waiting on Gradle full release anyway.

JaroslavTulach (approved):
I am adding my process review only approval. Otherwise I don't understand the code and I don't see a test, so I can't confirm the code actually improves.

entlicher:
> Otherwise I don't understand the code and I don't see a test, so I can't confirm the code actually improves.
Debugging of GraalVM EE images are supported now, CE images will be supported after the debugging info in CE is complete. Then we'll have tests with GraalVM CE.

PR #3313, opened 2021-11-10
[NETBEANS-6107] Use Gradle 7.3 Tooling API

lkishalmi:
I hope I've replaced everything. Bit late for me today.


mbien:
is the wrapper handled separately? (would unblock #3278)
/extide/gradle/netbeans-gradle-tooling/gradle/wrapper/gradle-wrapper.properties

neilcsmith-net:
Thanks @lkishalmi Rerunning a few Travis jobs - more than the usual suspect failed, but probably not an issue.

Curious at @mbien question - not looked in depth at the difference, or why they're out of sync - compile time only?

neilcsmith-net:
OK, Travis is as happy as it ever is!

lkishalmi:
The wrapper in the build is compile time only thing. We need to update that when we are compiling on JDK17. That's for 13.0

neilcsmith-net (approved):
Thanks. That's what I thought. Definitely something for NB13. Merging this and will trigger RC3 in next day or so. Please close NETBEANS-6107 once that's out and working.

mbien:
> The wrapper in the build is compile time only thing. We need to update that when we are compiling on JDK17. That's for 13.0

can we do that early in the NB 13 dev cycle? I looked into it myself but someone would have to upload the library to that lib hosting service the NB is using - otherwise i would have done it in the PR linked above.

Getting NB to build on JDK 17 will be the easy part. The tests have the potential to be "interesting", since netbeans hooks into an osgi lib which is so old that it isn't even officially supported on JDK 8. The fact that the tests were working up to JDK 15 is a little miracle.

CI can't test on anything beyond JDK 11 which isn't EOL right now.

neilcsmith-net:
@mbien agreed, but please move that discussion on to dev@ And ideally no more uploading to that lib hosting service!

PR #2959, opened 2021-05-17
Dynamically switching from all-file processing to single-file processing of multi file JavaSources.

jlahoda:
This is an attempt to improve the current facility to run processing of multiple Java files in a single javac instance by trying the processing, and falling back to file-by-file processing if it fails to fit into memory.

PR #3316, opened 2021-11-12
[NETBEANS-6196] Package code completion doesn't work in empty groovy …

ppisl:
This PR allow to complete package keyword in empty groovy file. Also after the package keyword the code completion offers only the packages.

Netbeans Issue: https://issues.apache.org/jira/browse/NETBEANS-6196


JaroslavTulach (approved):
Great contribution Petr!

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/KeywordCompletion.java
```diff
@@ -98,6 +110,27 @@ public boolean complete(Map<Object, CompletionProposal> proposals, CompletionCon
         return true;
     }
 
+    boolean isFirstStatement(final CompletionContext request) {
+        TokenSequence<GroovyTokenId> ts = LexUtilities.getGroovyTokenSequence(request.doc, 1);
+
+        if (ts != null) {
+            ts.move(request.lexOffset);
+            if (ts.movePrevious()) {
+                while (ts.isValid() && ts.movePrevious() && ts.offset() >= 0) {
```
sdedic, line 119: Nitpick: IMHO ts.offset() >= 0 is redundant on all places -- if `movePrevious` had failed bcs there's no previous token, it would return `false`; and the 1st token is always at offset 0.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/KeywordCompletion.java
```diff
@@ -98,6 +110,27 @@ public boolean complete(Map<Object, CompletionProposal> proposals, CompletionCon
...
+                    Token<GroovyTokenId> t = ts.token();
+                    if (!(t.id() == GroovyTokenId.NLS || t.id() == GroovyTokenId.WHITESPACE
```
sdedic, line 121: This could be moved off to some utilities a a `Predicate<Token<GroovyTokenId>>` as I fear that whitespace-traversing code is on many places and does not take all the WS/comment tokens into account (see for example the package name back traversal code in this PR).

sdedic (approved):

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/KeywordCompletion.java
ppisl: I wanted to be consistent with the rest of the code. Else this should be changed in more places, not only in my patch.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/KeywordCompletion.java
ppisl: This is good idea. Do you want to do it in this PR?

sdedic: perhaps yes, but a possible retrofit of other code (that should use the helper instead of custom WS token checks) should be perhaps done separately. Your call.

sdedic:
Seems that Travis failures relate to 77cd083d9a (#2959) rather than this PR.

PR #3317, opened 2021-11-12
LSP: Organize imports action added.

dbalek:
- Organize imports action added
- Basic options added
- Organize imports on save added

### DIFF
--- a/java/java.lsp.server/vscode/package.json
+++ b/java/java.lsp.server/vscode/package.json
@@ -108,6 +108,34 @@
 					"type": "integer",
 					"default": 100,
 					"description": "Timeout (in milliseconds) for loading Javadoc in code completion (-1 for unlimited)"
+				},
+				"netbeans.java.onSave.organizeImports": {
+					"type": "boolean",
+					"default": true,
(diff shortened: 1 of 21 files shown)


java/java.lsp.server/vscode/package.json
```diff
@@ -108,6 +108,34 @@
...
+					"description": "Groups of import statements (specified by their package prefixes) and their sorting order. Import statements within a group are ordered alphabetically",
+					"default": [
+						"java",
+						"javax",
+						"org",
+						"com",
+						""
+					]
+				},
+				"netbeans.java.imports.countForUsingStarImport": {
```
JaroslavTulach, line 128: Do we need an option for wildcard imports at all? It is a bad practice to `import ....*` and I'd just not advertise it in settings at all.

JaroslavTulach (approved):

java/java.lsp.server/vscode/package.json
dbalek: Options exported to be on par with RH ext.

PR #3319, opened 2021-11-13
[NETBEANS-5783] Testcase checking Gradle ProjectInfo never return null

lkishalmi:
I've accidentally missed these test from  #3307


matthiasblaesing:
Looks sane to me - lets get this in.

PR #3320, opened 2021-11-13
[NETBEANS-6004] Give info when the IDE's Java is not compatible with Gradle

lkishalmi:
image


lkishalmi:
Also let be a bit more strict on Gradle with Java 17 and require 7.3 instead 7.2 (which Java 17 somewhat working)

sdedic (approved):

PR #3322, opened 2021-11-17
Use Gradle distribution service and public repo for their binaries.

lkishalmi:
Well, this one became a bit more complex than originally I'd liked.

Let's see what the CI and the reviewers say for this.


matthiasblaesing:
For the download mechanics I don't think using a maven based approach here does not work sanely. Maven central has one main benefit: It is a one stop solution with a big mirror network. Gradle decided not to benefit from that and roll their own distribution.

In the end it is just another server we have to download from, so why pretend otherwise? We will have to face this in other contexts and will have do discuss how many/which servers are acceptable for this. The implementation for external files from arbitrary URLs can be found here:

https://github.com/matthiasblaesing/netbeans/commit/eea97fed303aef87dfc11402cc7921d6b57cc056

And based on that this is the gradle work:

https://github.com/matthiasblaesing/netbeans/commit/0043892b0fd49facf0bb4b471877f624e808e918

This way the download mechanic is still in one place and license and notice handling are still centraliced.

mbien:
haven't tested it yet but i do like that everything is now using the same version

lkishalmi:
@matthiasblaesing I've checked your commit regaring the Gradle upgrade. If you make a PR on that, I'd happily suppot that one over this.

As of why Gradle did not go with Maven central, I can understand them. They probably needs their download stats, and maybe other things as well. The binaries are distributed using Amazon Cloudfront, so it has also benefits a wide range of "mirrors".

lkishalmi:
Closed in favor of #3326

PR #3326, opened 2021-11-18
Use Gradle distribution service and public repo for their binaries.

matthiasblaesing:


matthiasblaesing:
This is an alternative to #3322 which adds the ability to load arbitrary URLs as external binaries. On that ability the gradle tooling is build with binaries from their download cite.

I was observed, that loading an older gradle project fails:

image

The log was empty (why is a warning not logged?!) apart from this:

```
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming EVALUATED for Unloaded Gradle Project: GradleFiles[projectDir=/home/matthias/src/jbake, rootDir=/home/matthias/src/jbake]
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming FULL for Gradle: jbake[FALLBACK]
INFO [org.netbeans.modules.gradle.execute.GradleDistributionProviderImpl]: Gradle Distribution for Gradle: jbake[FALLBACK] is GradleDistribution{gradleUserHome=/home/matthias/.gradle, distributionDir=/home/matthias/.gradle/wrapper/dists/gradle-6.3-bin/8tpu6egwsccjzp10c1jckl0rx/gradle-6.3, distributionURI=https://services.gradle.org/distributions/gradle-6.3-bin.zip, version=Gradle 6.3}
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming EVALUATED for Unloaded Gradle Project: GradleFiles[projectDir=/home/matthias/src/jbake/jbake-core, rootDir=/home/matthias/src/jbake]
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming EVALUATED for Unloaded Gradle Project: GradleFiles[projectDir=/home/matthias/src/jbake/jbake-dist, rootDir=/home/matthias/src/jbake]
INFO [org.netbeans.modules.parsing.impl.indexing.RepositoryUpdater]: Resolving dependencies took: 6 ms
INFO [org.netbeans.modules.parsing.impl.indexing.RepositoryUpdater]: Complete indexing of 0 binary roots took: 1 ms
INFO [org.netbeans.modules.parsing.impl.indexing.RepositoryUpdater]: Complete indexing of 0 source roots took: 0 ms (New or modified files: 0, Deleted files: 0) [Adding listeners took: 0 ms]
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming FULL_ONLINE for Gradle: jbake[FALLBACK]
```

@lkishalmi can you give me a hint where to look?

lkishalmi:
Could it be that you are seeing an issue explained in #3320 ? (Running the IDE on a JVM which is not supported by the target Gradle version?)

matthiasblaesing:
I managed to get a full stack trace from the error and I think this is the core part:

> [exec] Caused by: java.lang.AbstractMethodError: Receiver class org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder does not define or inherit an implementation of the resolved method 'abstract java.lang.Object getProperty(java.lang.String)' of interface groovy.lang.GroovyObject.

Fulltrace:  [stacktrace.txt](https://github.com/apache/netbeans/files/7572297/stacktrace.txt)

I see a structural change here:

https://docs.groovy-lang.org/docs/groovy-3.0.9/html/api/groovy/lang/GroovyObject.html
https://docs.groovy-lang.org/docs/groovy-2.5.15/html/api/groovy/lang/GroovyObject.html

In groovy 3 the methods are declared as default methods. My gut feeling is, that our groovy code is built against groovy 3, but is executed with a groovy 2 runtime.

matthiasblaesing:
From my perspective it comes down to this problem: How can we build with gradle-7.3 against the API of groovy 2.X.

Or an alternative Question: Could we implement the `netbeans-gradle-tooling` in java?

lkishalmi:
Probably I need to do a Java port. Though theoretically it is not impossible to modify the build script to use Groovy 2.x instead of the one provided by Gradle.

matthiasblaesing:
I'll also have a look - seems to be doable.

lkishalmi:
Started to work on the Java migration. Making good progress.

matthiasblaesing:
I did the groovy -> java translation (including the most recent change to `NbProjectInfoBuilder`) and rebased the PR onto master.

lkishalmi:
OMG. What could happen while I've been sleeping. The migration looks like mine, so I won't finish the remaining ~200 lines of that.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -0,0 +1,643 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.codehaus.groovy.runtime.InvokerHelper;
```
lkishalmi, line 37: As I see this is the only Groovy dependency in this source. Is it safe? Wouldn't it be better to implement the Java reflection code here for an extra couple of lines?

matthiasblaesing: I jumped into the `getProperty` method in the helper and it contains special handling code to deal with the specialities of groovy - I doubt, that java reflection is just some lines. My gut feeling is, that the groovy code base is pretty stable. The issue we saw came from building against 3 and expecting it to work on 2.

matthiasblaesing:
I found a translation error, where property keys were not correctly set - I pushed an according update.

matthiasblaesing:
Point taken - I pushed an update.  Could you please have a look at these Unittests:

org.netbeans.modules.gradle.VerifyGradleProjectTemplatesTest.testGradleProjectTemplates()
org.netbeans.modules.gradle.java.classpath.ClassPathProviderImplTest.testCompilePreTrusted()
org.netbeans.modules.gradle.queries.GradleSourceAttacherImplTest.testSourceAttachActionInvokesGradle()

They fail locally and at least the testSourceAttachActionInvokesGradle also fails on travis, the rest I'm not sure.

matthiasblaesing:
I think I finally spotted the problem in the dependecy resolution/collection code. Lets see what travis has to say about this.

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ignored-overlaps
```diff
@@ -65,24 +65,49 @@ enterprise/web.core.syntax/external/struts-taglib-1.3.10.jar enterprise/web.stru
...
+extide/gradle/external/gradle-7.3-bin.zip java/maven/external/aether-api-1.13.1.jar
+extide/gradle/external/gradle-7.3-bin.zip java/maven.embedder/external/apache-maven-3.6.3-bin.zip
+extide/gradle/external/gradle-7.3-bin.zip platform/api.annotations.common/external/jsr305-3.0.2.jar
+extide/gradle/external/gradle-7.3-bin.zip platform/libs.junit4/external/hamcrest-core-1.3.jar
+extide/gradle/external/gradle-7.3-bin.zip platform/o.apache.commons.codec/external/commons-codec-1.13.jar
+extide/gradle/external/gradle-7.3-bin.zip platform/o.apache.commons.io/external/commons-io-2.6.jar
+extide/gradle/external/gradle-7.3-bin.zip ide/libs.antlr4.runtime/external/antlr4-runtime-4.7.2.jar
+extide/gradle/external/gradle-7.3-bin.zip platform/libs.junit4/external/junit-4.13.2.jar
+extide/gradle/external/gradle-7.3-bin.zip platform/libs.testng/external/jcommander-1.78.jar
+extide/gradle/external/gradle-7.3-bin.zip ide/o.eclipse.jgit/external/org.eclipse.jgit-5.7.0.202003110725-r.jar
```
lkishalmi, line 89: Just one question do we really have extide/libs.gradle/external/gradle-7.3-bin.zip on the tree as well as extide/gradle/external/gradle-7.3-bin.zip

lkishalmi (approved):
Great work @matthiasblaesing !

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/ignored-overlaps
matthiasblaesing: You are right - sometime between I tried to extract the tooling from `gradle-7.3-bin.jar`, but that failed miserably, as the embedded tooling is not identical to the independently runnable jar. So it is downloaded again.

I also checked the other `ignored-overlaps` for gradle again and could reduce the list considerably.

With this cleanup, I squashed the changes together and think now it is ready to be merged. When Travis agrees I'll merge.

Thanks for the check and patience :-)

lkishalmi:
GitHub and Travis are happy. I've merged this one to master as I'm about to do some refactoring on Gradle splitting the UI and non-ui parts.

matthiasblaesing:
Thx

PR #3329, opened 2021-11-22
Fixed some lost-in-translation issues with Gradle Tooling

lkishalmi:
Well, these are just some quick catches. I've missed these during the review.


matthiasblaesing (commented):
Looks sane to me and sorry for the mess I caused :+1:

matthiasblaesing (approved):

lkishalmi:
No problem, at least I could add something to the effort!
I do not know why I would not have converted that stuff to Java earlier. Could have spared a few hours maybe days debugging Groovy.

PR #3332, opened 2021-11-25
Switch to FlatLaf Light as default look and feel for IDE

neilcsmith-net:
Switch to FlatLaf Light as default look and feel, and editor profile, for the IDE.  This is the simplest approach using nb/ide.branding, so keeps the existing defaults for the platform, and the laf used for the upgrader dialog.

Related to discussion at https://lists.apache.org/thread/9v3tlz8klqs93c1z6s0kvg03tr8w6kdx

### DIFF
--- a/nb/ide.branding/build.xml
+++ b/nb/ide.branding/build.xml
@@ -110,5 +110,11 @@ Contributor(s): Vincent Brabant, Maxym Mykhalchuk
       <branding name="nb"/>
     </locjar>
 
+    <locjar warnMissingDir="true"
+      basedir="o.n.swing.plaf/src"
+      jarfile="${cluster}/modules/org-netbeans-swing-plaf.jar">
+      <branding name="nb"/>
+    </locjar>
+
   </target>
 </project>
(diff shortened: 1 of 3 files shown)


lkishalmi (approved):
A really elegant solution, indeed!

eirikbakke:
Did this approach successfully avoid the problems from the "The Options ruban remains white with DarkLaf" email thread, including on the very first startup with a clean user directory?

(In my own NetBeans Platform app, I noticed that simply setting "--laf com.formdev.flatlaf.FlatLightLaf" in app.conf did not work, due to the FlatLAF module not being loaded early enough in the startup sequence.)

neilcsmith-net:
@eirikbakke try it! It's what review is for. :wink: 

The problem in that thread, and probably in your own platform application, is a missing dependency on the non-API `org.netbeans.swing.laf.flatlaf`. That module is autoload so doesn't kick in without one. We have that in `core.kit`. Having that module as autoload is a bug in my opinion. One perhaps to fix in this PR.

geertjanw (approved):

geertjanw:
Great.

eirikbakke:
Sorry, I didn't get a chance to test this patch yet--but I'm sure it works fine!

(I'll apply it on my NetBeans Platform app eventually, though, and report problems if there are any.)

lkishalmi:
Looks Great!

image

The picture is taken from a new install of a freshly cut netbeans-dev Snap

PR #3334, opened 2021-11-25
Small improvement on GitBranchHash which is able to get at least the hash of a detached HEAD

lkishalmi:
Fix from an unmerged commit #3151.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/GitBranchHash.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/GitBranchHash.java
@@ -79,19 +79,25 @@ public class GitBranchHash extends Task {
             if (headroot != null && Files.size(headroot) > 0l) {
                 List<String> lines = Files.readAllLines(headroot);
                 String line = lines.get(0);
-                String revLink = line.substring(line.indexOf(':')+1).trim();
-                branch = revLink.substring(revLink.lastIndexOf('/')+1).trim();
-                Path revPath = headroot.getParent().resolve(revLink);
-                if(Files.isRegularFile(revPath) && Files.size(revPath) > 0l) {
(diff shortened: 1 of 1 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/GitBranchHash.java
```diff
@@ -79,31 +79,40 @@ public void execute() throws BuildException {
...
                 String line = lines.get(0);
-                String revLink = line.substring(line.indexOf(':')+1).trim();
-                branch = revLink.substring(revLink.lastIndexOf('/')+1).trim();
-                Path revPath = headroot.getParent().resolve(revLink);
-                if(Files.isRegularFile(revPath) && Files.size(revPath) > 0l) {
-                    List<String> revlines = Files.readAllLines(revPath);
-                    String revline = revlines.get(0);
-                    if(revline.length()>=12){
-                        hash = revline.trim();
+                if (!line.contains(":") && (line.length() == 40)) {
```
neilcsmith-net, line 82: Not sure if this should be `>= 40` and also trimmed? I have also no idea if `>= 12` was just copying from HgID or necessary. :shrug:

neilcsmith-net (approved):
Looks good. Always meant to go back and look at this.  Ideally we'd pick up the branch from the hash - at least all release hashes are only on one branch - or maybe get the branch from the json file if any matching hash. Would make building from tags work correctly.

nbbuild/antsrc/org/netbeans/nbbuild/GitBranchHash.java
lkishalmi: I do not know that either. Just made some adjustments to work with a detached head.

PR #3335, opened 2021-11-26
Fixed typo: runSingle presence depends on runClassName property.

sdedic:
This PR fixes typo introduced in #3326 that broke `runSingle` task, which is used by "run" or "debug" actions on a single file (a class with a psv main method).


matthiasblaesing (approved):
Looks good, sorry for the mess and thanks for fixing.

lkishalmi (approved):

lkishalmi:
Thanks!

PR #3336, opened 2021-11-26
[NETBEANS-6235] Build of vscode extension fails

ppisl:
This is simple fix for the situation, when newer version of npm is used during building VSCode extension. It filters out .packe-lock.json from the files that needs a license.


JaroslavTulach (approved):

sdedic:
OK, fails just on `SecureURLResourceRetrieverTest`, which fails quite independently on `master`, too.

PR #3338, opened 2021-11-26
Remove jruby related YAML parser libs

lkishalmi:
Removed old YAML support library modules ```libs.jvyamlb``` and ```libs.bytelist```. As far as I've checked they are not referenced in our repository any more. Old JRuby support, if that still work, might need these though. Also it is possible to get them as nbm-s from archived pre 13 release.


matthiasblaesing (approved):
I think this is a valid move. Distributing libraries, because they might be used can't be our premise.

PR #3339, opened 2021-11-27
Deactivate broken xmlretriever test (client cert expired and a new one is not yet issued)

matthiasblaesing:
As @sdedic and @JaroslavTulach noticed the unittest fails and can't be fixed at this point in time:

https://github.com/chromium/badssl.com/issues/482


matthiasblaesing:
Travis is green, just not getting updated.

PR #3340, opened 2021-11-28
[NETBEANS-5599] PHP 8.1 Support: Never type

junichi11:
https://issues.apache.org/jira/browse/NETBEANS-5599

RFC: https://wiki.php.net/rfc/noreturn_type

- Convert from UnusableTypesUnhandledError to UnusableTypesHintError

(Part 1)
- Add `hasNeverType()` to `PhpVersion`
- Fix the lexer
- Mark as an error if something is returned when the return type is `never`
- Mark as an error if `never` is used at incorrect places (e.g. private never $propertyType)
- Add unit tests for the lexer and hints

(Part 2)
- Fix the code completion
- Add and fix unit tests for the code completion

#### Screenshot
nb-php81-never-type


junichi11:
CI failed: see https://github.com/apache/netbeans/pull/3339

junichi11:
@tmysik Could you please have a look at this? If there is no problem, let's merge it. Thanks.

tmysik (approved):
Looks good to me, many unit tests. Thanks for your great work!

junichi11:
Thank you for your review :)

PR #3343, opened 2021-11-29
[NETBEANS-6241] - Fixed switchToRuleSwitch Hint for BindingPattern and GuardedPattern

mishrasandeep:
https://issues.apache.org/jira/browse/NETBEANS-6241
https://github.com/apache/netbeans/issues/3966


java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
```diff
@@ -1966,8 +1967,23 @@ protected int diffBindingPattern(JCBindingPattern oldT, JCBindingPattern newT, i
 
     protected int diffCase(JCCase oldT, JCCase newT, int[] bounds) {
         int localPointer = bounds[0];
-        List<JCExpression> oldPatterns = getCasePatterns(oldT);
-        List<JCExpression> newPatterns = getCasePatterns(newT);
+        List<? extends JCTree> oldPatterns;
+        List<? extends JCTree> newPatterns;
+        
+        if(TreeShims.isJDKVersionRelease17_Or_Above() &&
```
lahodaj, line 1973: This should always be true currently, right? I mean, NetBeans no longer supports javac API older that 17 (in NetBeans 13) or 18 (in the mainline).

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
```diff
@@ -3196,7 +3198,7 @@ public static void performRewriteRuleSwitch(JavaFix.TransformationContext ctx, T
         for (Iterator<? extends CaseTree> it = cases.iterator(); it.hasNext();) {
             CaseTree ct = it.next();
             TreePath casePath = new TreePath(tp, ct);
-            patterns.addAll(ct.getLabels());
+            patterns.addAll(ct.getLabels());
```
lahodaj, line 3201: Nit: unnecessary change.

lahodaj (commented):
I think it would be good to use this opportunity to cleanup the code. I believe NB is only supporting javac API from JDK 17 or JDK 18, depending on the NB version - so there should be no need to use reflection to access the JDK 17 javac API, and no need to check if the API is available.

java/java.source.base/src/org/netbeans/modules/java/source/save/CasualDiff.java
mishrasandeep: Thanks for the input. have adressed the comments and removed reflection access wherever possible.

arvindaprameya:
Approved, thanks for completing this.

java/java.hints/src/org/netbeans/modules/java/hints/errors/Utilities.java
mishrasandeep: Thanks for the input. Have made the changes.

lahodaj: Seems the change is still there - the import above seems unnecessary, and this line seems to have some unnecessary trailing whitespaces.

lahodaj (approved):

singh-akhilesh (approved):
LGTM

PR #3344, opened 2021-11-29
[NETBEANS-6245] Supply MySQL connector location by default. Upgrade MySQL/J library.

sdedic:
`LICENSE` file was updated from the one found inside the `mysql-connector-8.0.27.jar` file (in META-INF/)

### DIFF
--- a/ide/db.mysql/external/binaries-list
+++ b/ide/db.mysql/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-53DA6AFDC5F7B45CAAA5BC3627028B2041A36CEE mysql:mysql-connector-java:8.0.17
+F1DA9F10A3DE6348725A413304AAB6D0AA04F923 mysql:mysql-connector-java:8.0.27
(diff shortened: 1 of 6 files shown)


matthiasblaesing:
To my knowledge Connector/J is distributed under GPLv2-only, so we can't ship it. I tested with my local version and when I try to create a new mysql connection, I'm prompted to point NetBeans to the JAR, so that it can be used.

The JAR itself was added to the downloads to make unittests possible (https://github.com/apache/netbeans/commit/aec700e82092af3851fd9d8883eadfcd942ee035) and was marked to be `compile-time`:

https://github.com/apache/netbeans/blob/c73b134cf520288a4ee5101885fb4039253ec496/ide/db.mysql/external/mysql-connector-java-8.0.17-license.txt#L4

sdedic:
Oops. Did not check properly the licensing stuff. 
OK, closing this one, I'd guess - thanks, @matthiasblaesing

JaroslavTulach:
Yup, I believe dual licensing (GPL vs. commercial license) was always the core of MySQL business. Looks like the new owner continues to keep it.

PR #3347, opened 2021-11-30
Snapshot of APIs as of 12.6

neilcsmith-net:
Snapshot of APIs as of 12.6. We should have produced this earlier - had slipped my mind, anyway.

Had to build and produce these on JDK 8. Doing with JDK 11 is failing with an error.  Same for other people?  We'll need to fix that by NB13 - haven't looked at what is failing - some reference to JDK classes IIRC.

Also had to resolve following conflicts in merging - ignored the removed modules, left javacapi as in master.

```
deleted by us:   ide/libs.bytelist/nbproject/org-netbeans-libs-bytelist.sig
deleted by us:   ide/libs.jvyamlb/nbproject/org-netbeans-libs-jvyamlb.sig
both modified:   java/libs.javacapi/nbproject/org-netbeans-libs-javacapi.sig
```

Built in the usual way

```
# checkout branch (or release tag)
git checkout release126
ant clean
ant build
ant gen-sigtests-release
git add -A
git stash
git checkout master
git checkout -b apis-nb126
git stash pop
git add -A
git commit -m "Snapshot of APIs as of 12.6"
# remove getPeer() calls
find . -name "*.sig" -exec sed -i '/java.awt.peer.ComponentPeer/{N;d;}' {} \;
# check git diff
git add -A
git commit --amend
```

### DIFF
--- a/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
+++ b/apisupport/apisupport.ant/nbproject/org-netbeans-modules-apisupport-ant.sig
@@ -1,5 +1,5 @@
 #Signature file v4.1
-#Version 2.85
+#Version 2.86
 
 CLSS public java.lang.Object
 cons public init()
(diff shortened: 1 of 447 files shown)


JaroslavTulach (approved):
Looks sane at first glance.

neilcsmith-net:
Thanks @JaroslavTulach Merging. Let's get it in before we get any more conflicts. Any thoughts on the JDK 11 problem?

PR #3348, opened 2021-12-01
Dispose/disconnect tree view after client stop.

sdedic:
When a JDK is switched, and the lang client is restarted, the old TreeViewService is still active as well as any handlers it hooked to the vscode. This PR adds unregistration/disposal of the service on Lang client stop.


dbalek (approved):

sdedic:
Failures unrelated. Merging early as the bug impacts master's usability of the vscode ext.

PR #3349, opened 2021-12-01
Rich HTML UI for (now only for) VSCode Refactorings

JaroslavTulach:
This PR contains the prototyping work we (Jirka Sedláček, Dušan Bálek and me) have done to make sure Rich UI for _Change Parameters Refactoring_ can be productized.  All changes were squashed and integrated by #3357 and #3361. This PR is now closed, but holds history of individual commits and few changes showing how to use and test the HTML UI in classical NetBeans IDE.

By default VSCode UI follows principles of asceticism and is very simplistic and straightforward. It is possible to use `WebView` for more complex UIs, but:
- it's always preferable to not do that
- coding HTML in Java might not seem easy

Anyway we really need it for the refactoring and we made it. Here is the result:

obrazek

My goal is to get this into January version of NetBeans and its VSCode extension. That requires fixes in [HTML/Java API](http://github.com/apache/netbeans-html4j), some API fixes in NetBeans itself and then implementation of the refactoring. Those changes were done by #3357 and #3361.

The remaining (unmerged) part of this PR demonstrates how to use the HTML UI in NetBeans. It adds two actions into refactoring menu:

obrazek

The first action opens dialog with the same _Change Method Parameters Refactoring_ UI as provided to VSNetBeans by #3361 in a dialog - e.g. it shows sharing of the HTML UI is possible among the Swing and Electron based flavors of the IDE. Second _Open HTML Hello World!_ action opens a top component with wizard generated sample

obrazek

showing input line and a button that opens a dialog. I used this example to test that interaction between HTML, Swing and JavaFX works unchanged when refactoring the code in #3357

obrazek

Cons: I haven't yet convinced JavaFX WebView to load `codicons.css` and `codeicons.ttf` as provided by #3361 - as such the HTML UI in NetBeans isn't as slick. Further enhancements into the JavaFX webview integration are needed.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ChangeMethodParametersRefactoringAction.java
```diff
@@ -0,0 +1,84 @@
...
+                            System.err.println("e: " + e);
+                            if (e.getKind() == ElementKind.METHOD) {
+                                method = (ExecutableElement) e;
+                                break METHOD;
+                            }
+                        }
+                    }
+                    if (method != null) {
+                        ElementHandle<ExecutableElement> handle = ElementHandle.create(method);
+                        ChangeMethodParameterUI[] model = {null};
```
JaroslavTulach, line 72: This action displays the same VSCode UI in NetBeans IDE! The CSS isn't optimized (e.g. no screenshot to share), but it does work (if the JDK has JavaFX libraries available).

While the current effort is only targeted towards VSCode, ideally we can reuse the UIs on both sides in a future - in NetBeans IDE as well as VSCode. To do so following has to happen:
- the HTML has to be (almost) without CSS
- CSS is applied by the "chrome" and needs to be compatible with current L&F
- modules using the UI in NetBeans IDE need a "presenter"
- currently we use JavaFX web view as a presenter - e.g. it is necessary to install JavaFX libraries
- we could show the UI in an external browser - so called "browser" presenter.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ui/ChangeMethodParameters.html
```diff
@@ -0,0 +1,59 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="Content-Security-Policy" content="default-src http: 'unsafe-inline' 'unsafe-eval'">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="refactoring.css">
+    <link rel="stylesheet" href="codicon.css">
```
JaroslavTulach, line 9: These CSS shall ideally be provided dynamically by the "chrome".

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -675,6 +676,53 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
         window.showErrorMessage('Error initializing ' + reason);
     });
 
+    async function showHtmlPage(params : HtmlPageParams): Promise<string> {
```
JaroslavTulach, line 679: When there is a need to show the UI, the JVM opens a port (unsecured right now) with a simplistic HTTP server ands sends a LSP request to show a page where it passes the URL to show. The request is handled in this VSCode extension and opens a `WebView` that connects to the HTTP port and makes the page "live" and controllable from a Java code.

platform/api.htmlui/src/org/netbeans/modules/htmlui/HTMLDialogBase.java
```diff
@@ -0,0 +1,56 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.htmlui;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public abstract class HTMLDialogBase {
```
JaroslavTulach, line 25: The HTMLUI API module no longer requires JavaFX - it can work with different displayers - like the one provided by VSCode. To do that I had to abstract the `HTMLDialogBase` common functionality and provide two implementations - one for Swing (and JavaFX) and one for VSCode.

platform/api.htmlui/src/org/netbeans/spi/htmlui/HtmlViewer.java
```diff
@@ -0,0 +1,35 @@
...
+import java.net.URL;
+import java.util.concurrent.Callable;
+import java.util.function.Consumer;
+
+/**
+ *
+ * @param <HtmlView>
+ * @since 1.21
+ */
+public interface HtmlViewer<HtmlView> {
```
JaroslavTulach, line 30: The new SPI between HTML UI API and the VSCode module to do the "viewing" differently.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/HttpServer.java
```diff
@@ -0,0 +1,62 @@
...
+ * under the License.
+ */
+package org.netbeans.modules.java.lsp.server.htmlui;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.Writer;
+import org.netbeans.html.boot.spi.Fn;
+
+abstract class HttpServer<Request, Response, WebSocket, Runner> {
```
JaroslavTulach, line 26: This package is a 1:1 copy of code from HTML/Java APIs. It'd be better to use the API than copy, however, until https://github.com/apache/netbeans-html4j/pull/26 gets integrated, we don't have much other options than to copy.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ui/codicon.css
```diff
@@ -0,0 +1,485 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
matthiasblaesing, line 4: I assume, that the ttf file is under the same license?! Please add an entry for the `codicon.ttf` to `license-info.xml`. The same is true for other font and graphics files (I see `newHTML.png`).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
```diff
@@ -0,0 +1,150 @@
...
+        whenReady.run();
+    }
+
+    @Override
+    public void load(View view, ClassLoader loader, URL pageUrl, Callable<Object> initialize, String[] techIds) {
+        view.load(pageUrl, initialize);
+    }
+
+    @JavaScriptBody(args = {}, body = "\n"
+        + "const vscode = acquireVsCodeApi();\n" // this method can be called only once per WebView existance
```
matthiasblaesing, line 51: Is this really LSP or is this an API hardcoded against VSCode?

java/libs.javacapi/external/binaries-list
```diff
@@ -15,3 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 B195FD4F74CA4BA8BF5985E9103B96C6F0D9079D com.dukescript.nbjavac:nb-javac:jdk-17+35:api
+E748E597EACD349D65869A26FCB99A8B4DF887C2 com.dukescript.nbjavac:nb-javac:jdk-17+35
```
matthiasblaesing, line 18: Was this PR not done on a recent master checkout? This change is already in master?!

nb/updatecenters/extras/nbjavac.impl/build/no-license.txt
```diff
@@ -0,0 +1 @@
+[NO LICENSE SPECIFIED]
```
matthiasblaesing, line 1: This confirms, that this reintroduces changes, that were already reviewed and in this case is broken. Please drop all changes relating to the nbjavac changes from this and rebase the remaining parts onto master. That should ensure, that no regressions are pulled in.

platform/generic/manifest.mf
```diff
@@ -0,0 +1 @@
+OpenIDE-Module: generic
```
matthiasblaesing, line 1: The name is not good.

java/libs.javacapi/external/binaries-list
JaroslavTulach: No, this PR has been worked on for a long time. However Dušan merged latest master in and now this diff is outdated.

nb/updatecenters/extras/nbjavac.impl/build/no-license.txt
JaroslavTulach: Done by 276330e

platform/generic/manifest.mf
JaroslavTulach: Renamed to `org.netbeans.html.presenters.spi`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
JaroslavTulach: The `java.lsp.server` is using some extensions to the standard LSP protocol. The _rich HTML UI_ is one such extension. There is a hand-shake negotiation to find out whether the client supports these extensions. @sdedic is an expert in this area. Sváťo, please guide me to use these _capabilities_ concept properly.

matthiasblaesing: What I see is, that refactoring is in no way specific to a Java LSP and looking at:

https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction
https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#codeActionKind

I think this is a known problem and the protocol is beginning to build up support for it. I'm in no way an export on LSP, but I doubt, that it is sensisble, if such a generic functionality (refactorings) is reimplemented by each provider and even worse by each client.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ChangeMethodParametersRefactoring.java
```diff
@@ -0,0 +1,363 @@
...
+        }
+        ChangeMethodParameterUI model = new ChangeMethodParameterUI();
+        model.withName(method.getSimpleName().toString())
+                .withReturnType(Utilities.getTypeName(ci, method.getReturnType(), true).toString())
+                .withSelectedModifier(mod)
+                .withParameters(params)
+                .assignData(client, file, TreePathHandle.from(handle, ClasspathInfo.create(file)));
+        model.applyBindings();
+        return (id) -> {
+            if ("accept".equals(id)) {
```
JaroslavTulach, line 211: Here you can validate the content of the dialog, @dbalek.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
sdedic: @JaroslavTulach  I am not sure that I understand the question: The LSP client declares its capabilities - see https://github.com/apache/netbeans/blob/master/java/java.lsp.server/vscode/src/extension.ts#L628 ; in this case, the client should probably sent something like `"org.netbeans.html.HtmlViewerCapable" : true` which could be checked by the LSP server.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
```diff
@@ -0,0 +1,741 @@
...
+            int from = 8080;
+            int to = 65535;
+            int port = config.getPort();
+            if (port != -1) {
+                from = to = port;
+            }
+            server.init(from, to);
+
+            String prefix = "/" + new UUID(config.random.nextLong(), config.random.nextLong()).toString();
+
```
JaroslavTulach, line 203: https://github.com/apache/netbeans-html4j/pull/26#issuecomment-981762288 questions the security when opening a port. The 24cb4da fix is supposed to make sure the port is only available locally. Moreover e91bdcb uses random UUID to prefix the server pages and in case of misconnect, it closes the server. 

That shall mitigate the basic problem. We can treat the rest as a bug.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
JaroslavTulach: Done in 7875578. Introducing `showHtmlPageSupport` capability. If it is missing, no dialog is shown and only a warning message is shown to the user.

platform/api.htmlui/src/org/netbeans/spi/htmlui/HtmlViewer.java
JaroslavTulach: Integrated by  #3357

platform/api.htmlui/src/org/netbeans/modules/htmlui/HTMLDialogBase.java
JaroslavTulach: No longer true. The most recent changes in  #3357 use just a single implementation for all `HTMLViewerSpi`s.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
JaroslavTulach: a65a7de synchronizes the `SimpleServer` class with https://github.com/apache/netbeans-html4j/pull/26/commits/9bc1429686ea37037bba33d9d3a8123ae9ff1148

JaroslavTulach:
Opps, I wanted to create a new PR, but instead I integrated 9e27e191c829978a9dbdcdaeee774d4208412348 into `master`. I made a mistake and forget to create new branch before push. I am sorry.

We don't have a way to remove commit from `master` (as far as I know). Rather than rolling back, I suggest to keep it and possibly polish it in the final PR I am about to create now. Sorry for the mess.

neilcsmith-net:
> Opps, I wanted to create a new PR, but instead I integrated [9e27e19](https://github.com/apache/netbeans/commit/9e27e191c829978a9dbdcdaeee774d4208412348) into `master`. I made a mistake and forget to create new branch before push. I am sorry.

Aside - it happens! :smile: I've long been in the habit of using `git remote set-url --push upstream DISABLED`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ChangeMethodParametersRefactoringAction.java
neilcsmith-net: +1 to using for VSCode. -1 to using in NetBeans IDE.  Not unless we get at least a presenter we can ship with, is more stable, and integrates properly with the L&F.

JaroslavTulach:
With integration of  #3357 and #3361 the goal of this PR has been achieved.

PR #3353, opened 2021-12-03
LSP + VSCode extension: Run/Debug via project actions.

entlicher:
For consistency reasons we execute project Run/Debug actions instead of file Run/Debug in VSCode. File-sensitive actions are provided as code lenses. Via project's Run/Debug actions users get consistent behavior.

Currently there was `mainClass` property set to `${file}` in VSCode's `launch.json`. We are changing `mainClass` property to be optional, it's not set in newly created `launch.json` and we're using `file` property set to `${file}` to get the context. `mainClass` allows to set a specific file as a context to Run/Debug. When not set, we use `file` to determine the current project to run Run/Debug on.

We need to adjust Maven projects to be able to Run/Debug without `mainClass` set. The project should know it's entry point. Currently, when you create a new Maven project from template, it asks you for the main class before Run/Debug. This is an unexpected behavior for the user, as the newly created project should know its main class. This is why we set `exec.mainClass` in Maven now.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
@@ -62,7 +62,7 @@ public final class NbLaunchRequestHandler {
         // validation
         List<String> modulePaths = (List<String>) launchArguments.getOrDefault("modulePaths", Collections.emptyList());
         List<String> classPaths = (List<String>) launchArguments.getOrDefault("classPaths", Collections.emptyList());
-        if (!isNative && (StringUtils.isBlank((String)launchArguments.get("mainClass"))
+        if (!isNative && (StringUtils.isBlank((String)launchArguments.get("mainClass")) && StringUtils.isBlank((String)launchArguments.get("file"))
                           || modulePaths.isEmpty() && classPaths.isEmpty())) {
             ErrorUtilities.completeExceptionally(resultFuture,
                 "Failed to launch debuggee VM. Missing mainClass or modulePaths/classPaths options in launch configuration.",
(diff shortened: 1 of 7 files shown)


JaroslavTulach (commented):
I see no description and I don't understand the goal of the changes. Shall I _disassemble_ your changes and try to assign them a purpose?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
```diff
@@ -93,23 +93,18 @@
...
-                    try {
-                        URI uri = new URI(filePath);
-                        ioFile = Utilities.toFile(uri);
-                    } catch (URISyntaxException ex) {
-                        // Not a valid file
-                    }
-                }
-            }
-            file = ioFile != null ? FileUtil.toFileObject(ioFile) : null;
+            file = getFileObject(filePath);
```
sdedic, line 107: I didn't run the code, but ... if `mainClass` is defined, then `filePath` now contains main class' FQN - so it can't be converted to FileObject ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
```diff
@@ -93,23 +93,18 @@
 
         activeLaunchHandler.preLaunch(launchArguments, context);
 
-        String filePath = (String)launchArguments.get("mainClass");
+        String filePath = (String)launchArguments.get("file");
+        String mainFilePath = (String)launchArguments.get("mainClass");
```
sdedic, line 97: is the `mainClass` an URI actually ?

entlicher: Yes.

sdedic (approved):

entlicher:
> I see no description and I don't understand the goal of the changes. Shall I _disassemble_ your changes and try to assign them a purpose?

entlicher:
I've unintentionally closed the PR when I wanted to remove an unintentional comment.

entlicher:
I've added the description.

PR #3354, opened 2021-12-03
LSP: attempt to speed up open type action.

dbalek:
Computation of a type's `Location` is expensive. Especially for the types form library jars when their corresponding source jars have to be downloaded first. This PR prevents computation of the final `Location' for every type item displayed in the selection list; the final `Location` is computed lazily for the selected type item only.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
@@ -26,7 +26,11 @@ import com.sun.source.util.TreePath;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
```diff
@@ -261,7 +267,50 @@
...
+                    }).thenAccept(loc -> {
+                        if (loc != null) {
+                            ShowDocumentParams sdp = new ShowDocumentParams(Utils.toUri(loc.getFileObject()));
+                            Position position = Utils.createPosition(loc.getFileObject(), loc.getStartOffset());
+                            sdp.setSelection(new Range(position, position));
+                            client.showDocument(sdp).thenAccept(res -> {
+                                if (res.isSuccess()) {
+                                    result.complete(null);
+                                } else {
+                                    result.completeExceptionally(new IllegalStateException("Cannot open source for: " + typeHandle.getQualifiedName()));
```
sdedic, line 302: The exception msg  is going to be printed to the client, maybe L10n it.

sdedic (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
dbalek: A majority of error messages within the language server implementation is not localized.

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -237,6 +237,10 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
     context.subscriptions.push(vscode.debug.registerDebugAdapterDescriptorFactory('java8+', debugDescriptionFactory));
     context.subscriptions.push(vscode.debug.registerDebugAdapterDescriptorFactory('nativeimage', debugDescriptionFactory));
 
+    // register content provider
+    let sourceForContentProvider = new NetBeansSourceForContentProvider();
+    context.subscriptions.push(vscode.workspace.registerTextDocumentContentProvider('sourceFor', sourceForContentProvider));
```
JaroslavTulach, line 242: Using [virtual documents](https://code.visualstudio.com/api/extension-guides/virtual-documents) that's clever.

PR #3356, opened 2021-12-07
mx now defaults to building under the jdk prefix

timfel:
setting MX_OUTPUT_ROOT_INCLUDES_CONFIG=false when running mx avoids
this. however, the generated sources are placed not directly under mxbuild if a
subDir is specified, so the subDir needs to be considered as well in the
binPrefix

/cc @jtulach


timfel:
[Test ide modules | openjdk8](https://app.travis-ci.com/github/apache/netbeans/jobs/551401023) says "BUILD SUCCESSFUL" but is marked as error?
[10694.13 Test Java modules with nb-javac on Java 8 | openjdk8](https://app.travis-ci.com/github/apache/netbeans/jobs/551401026) also says "BUILD SUCCESSFUL" but is marked as error?
[10694.24 Test Java modules without nb-javac on Java 15 | openjdk8](https://app.travis-ci.com/github/apache/netbeans/jobs/551401037) has timed out, but I don't see that it tests anything against mx projects, so I don't know how this could be caused by me?

PR #3357, opened 2021-12-07
Updating HTML/Java API to 1.7.3 version

JaroslavTulach:
[HTML/Java API](https://bits.netbeans.org/html+java/1.7.1/) has recently released its [1.7.3 version](https://lists.apache.org/thread/h6hzsm60ygodv5sz1c5q8xvjctwlqzy6) which (among other) gives proper OSGi bundle names to its modules. Former `generic` module is now `org.netbeans.html.presenters.spi`. Time to upgrade.

When at it, I am also splitting the `api.htmlui` module to two parts. The goal is to move the JavaFX WebView based implementation into its own module and allow the API to be used without it. That's all built around `HTMLViewerSpi` interface. 40f460e adds a _fallback_ implementation of the SPI interface that warns the user gracefully.

Also, to satisfy old @dbalek request, I am introducing `OnSubmit` callback that, allows validation of the state of the dialog before it is closed.

The commit c182fc33e409021950cdeba60bede1fe492a6348 modifies `nbjunit` module. I tried to use `MockServices` in TestNG without depending on JUnit4 and was getting linkage errors. Replacing the exceptions with `java.lang.AssertionError` fixes the problems.

This PR is the first in a line of others. The ultimate goal is to get _Rich HTML UI for (now only for) VSCode Refactorings_ as outlined in #3349 into NetBeans and its VSCode extension.


platform/o.n.html.presenters.spi/manifest.mf
```diff
@@ -0,0 +1 @@
+OpenIDE-Module: generic
```
matthiasblaesing, line 1: Missing rename?

JaroslavTulach: Fixed in 1b111c539faafa58665, thanks.

The missing rename went unnoticed as this is an _OSGi module_. It's build script contains `<target name="jar"/>` and as such the JAR is never built, but rather copied according to instructions in properties: `release.external/generic-1.7.3.jar=modules/org-netbeans-html-presenters-spi.jar`

dbalek (approved):
Proposed API looks fine to me.

lkishalmi (approved):
Well, skimming through, I've not found anything wrong with this.

neilcsmith-net (approved):
As with @lkishalmi I have only skimmed through. Seems a good idea. I would happily see the back of all HTML dialogs in the IDE though!

PR #3363, opened 2021-12-10
Added more actions to LSP Projects View

sdedic:
project-actions

The following actions were enabled:
- PROJECTS (view caption):
   - inline:  new project, collapse all 
   - context: build workspace, clean workspace, run/debug workspace
- PROJECT (node):
   - inline:  new from template, run/debug project
   - context: clean project, build project, test project
- PACKAGE (node):
  - inline:  new from template
  - context: delete package
CLASS (node):
  - context: delete class?

### DIFF
--- a/java/java.lsp.server/nbcode/integration/src/META-INF/services/org.openide.explorer.ExtendedDelete
+++ b/java/java.lsp.server/nbcode/integration/src/META-INF/services/org.openide.explorer.ExtendedDelete
@@ -15,12 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# Use fully qualified names, one class per line. Syntax is:
-#   qualified.name=value    : value will be used instead of qualified name.
-#   [prefix.]qualified.name : the prefix will be stripped
-#   [prefix=subst]          : the prefix in all classes will be replaced by subst, possibly empty
-#   * as the first character: watch for changes in this interface, fire changes.
-# Names will be converted to "class:<name>" tokens.
-
-org.netbeans.api.actions.Openable
-[org.netbeans.api.]project.Project
+# Suppress refactoring-aware delete as its UI is not suitable for LSP yet
+#-org.netbeans.modules.refactoring.spi.impl.SafeDeleteAction
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.netbeans.api.project.Project;
 import org.netbeans.modules.java.lsp.server.Utils;
 import static org.netbeans.modules.java.lsp.server.explorer.NodeLookupContextValues.nodeLookup;
 import org.openide.filesystems.FileObject;
@@ -150,8 +151,13 @@ public class DefaultDecorationsImpl implements TreeDataProvider.Factory {
                 }
             }
             if (n.canDestroy()) {
-                d.addContextValues(CTXVALUE_CAP_DELETE);
-                set = true;
+                // TODO Hack: exclude projects from delete capability. The TreeItemData probably needs to support
+                // exclusion... Project delete UI is not suitable for LSP at the moment
+                Project p = n.getLookup().lookup(Project.class);
+                if (p == null || f == null || !p.getProjectDirectory().equals(f)) {
+                    d.addContextValues(CTXVALUE_CAP_DELETE);
+                    set = true;
+                }
             }
             if (n.canRename()) {
                 d.addContextValues(CTXVALUE_CAP_RENAME);
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
@@ -34,6 +34,7 @@ import org.netbeans.modules.java.lsp.server.Utils;
 import static org.netbeans.modules.java.lsp.server.explorer.NodeLookupContextValues.nodeLookup;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
+import org.openide.loaders.DataFolder;
 import org.openide.nodes.AbstractNode;
 import org.openide.nodes.Children;
 import org.openide.nodes.Node;
@@ -49,6 +50,10 @@ public class DefaultDecorationsImpl implements TreeDataProvider.Factory {
 
     public static final String EXPLORER_ROOT = "Explorers"; // NOI18N
     public static final String COOKIES_EXT = "contextValues"; // NOI18N
+    
+    public static final String CTXVALUE_FOLDER = "is:folder"; // NOI18N
+    public static final String CTXVALUE_CAP_RENAME = "cap:rename"; // NOI18N
+    public static final String CTXVALUE_CAP_DELETE = "cap:delete"; // NOI18N
 
     private static final Logger LOG = Logger.getLogger(DefaultDecorationsImpl.class.getName());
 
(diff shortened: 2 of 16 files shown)


java/java.lsp.server/nbcode/integration/src/META-INF/services/org.openide.explorer.ExtendedDelete
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Suppress refactoring-aware delete as its UI is not suitable for LSP yet
+#-org.netbeans.modules.refactoring.spi.impl.SafeDeleteAction
```
JaroslavTulach, line 19: Good. The `ExtendedDelete` is really a horrible pattern. I am not proud for allowing it to get into the platform.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
```diff
@@ -110,23 +116,52 @@ public TreeItemData createDecorations(Node n, boolean expanded) {
             }
 
             FileObject f = n.getLookup().lookup(FileObject.class);
+            boolean nodeChecked = false;
+            if (f == null) {
+                DataFolder df = n.getLookup().lookup(DataFolder.class);
+                if (df != null) {
+                    f = df.getPrimaryFile();
+                    // Workaround for possible bug in data folder: DataFolder itself and the Fileobject for the folder do not
+                    // have a Node in their Lookups.
```
JaroslavTulach, line 125: Is it Folder node or Package node? I believe @entlicher (when working on version control system integration?) tried hard to not treat Package node as regular `FileObject`, but rather a "flat" folder.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
```diff
@@ -273,9 +286,48 @@ private static String singleSelection(List<QuickPickItem> selectedGroups) {
         }
         return selectedGroups.get(0).getUserData().toString();
     }
+    
+    private static final String ATTR_SOURCE_GROUP = "new.sourceGroup";  // NOI18N
+    private static final String ATTR_SOURCE_HINT = "new.sourceHint"; // NOI18N
```
JaroslavTulach, line 291: We should start documenting these [magic strings](http://wiki.apidesign.org/wiki/MagicalStrings) somewhere. Especially when we start spreading them outside of the `java.lsp.server` module.

Even if we don't, it should be `<arch stability="private"..../>` as that is an API someone can depend on (but shouldn't).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectActionParams.java
```diff
@@ -0,0 +1,72 @@
...
+ * under the License.
+ */
+package org.netbeans.modules.java.lsp.server.protocol;
+
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.eclipse.xtext.xbase.lib.Pure;
+import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
+
+/**
+ * Describes 1st command parameter for 'java.project.action' generic command.
```
JaroslavTulach, line 26: That's probably not the ultimate description how things shall work, right? I guess one needs to dig the functionality from elsewhere...

java/java.lsp.server/vscode/package.json
```diff
@@ -372,15 +372,22 @@
...
-				"category": "Java"
+				"category": "Java",
+				"icon": "$(new-file)"
 			},
 			{
 				"command": "java.workspace.newproject",
 				"title": "New Project...",
-				"category": "Java"
+				"category": "Java",
+				"icon" : "$(new-folder)"
```
JaroslavTulach, line 390: What does `${new-folder}` mean?

java/java.lsp.server/vscode/package.json
```diff
@@ -464,22 +504,76 @@
...
 				},
 				{
 					"command": "nbls:Database:netbeans.db.explorer.action.Disconnect",
-					"when": "nbJavaLSReady && viewItem =~ /class:ddl.DBConnection/"
+					"when": "viewItem =~ /class:ddl.DBConnection/"
 				},
 				{
 					"command": "nbls:Edit:org.openide.actions.DeleteAction",
-					"when": "nbJavaLSReady && viewItem =~ /class:ddl.DBConnection/"
+					"when": "viewItem =~ /cap:delete/"
```
JaroslavTulach, line 549: What does `=~ /cap:delete/` mean?

JaroslavTulach (commented):
I left few questions. I guess I am lost in the highlevel architecture.

java/java.lsp.server/vscode/package.json
sdedic: a reference to VS embedded icon. Feel free to suggest a (visually) better one.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
sdedic: The bug was found on the `package` node, which indeed does not have any FileObject (directly) in its Lookup. But it has a DataFolder in it ... and the `DataFolder` has its `primaryFile` property. During debugging it turned out (IIRC) that `DataFolder.getLookup().lookup(Node.class) == null` as well as `primaryFile.getLookup().lookup(Node.class) == null`, which is different from file-based DataObjects.

java/java.lsp.server/vscode/package.json
sdedic: I guess this is another 'private API' that needs a documentation. 

We had defined `class:something` to represent services on the node; I've "defined" `cap:` as prefix for capabilities (canRemove, canDestroy (well, delete), canRename -- possibly we'll find more) - neither of these is represented by a type, so different prefix to avoid possible clash. And yet `is:folder` -- I could turn that one into `class:DataFolder` (although it not exactly true, but could be better than yet another magic selector).

java/java.lsp.server/vscode/package.json
```diff
@@ -407,6 +414,39 @@
 			{
 				"command": "nbls:Edit:org.openide.actions.DeleteAction",
 				"title": "Delete"
+			},
+			{
+				"command": "workbench.action.debug.run",
+				"title": "Run Without Debugging",
+				"icon": "$(run)"
+			},
```
dbalek, line 422: Do we want all these new commands to be visible in the Command Palette? If yes, consider adding the `category` attribute, if no, create corresponding items (containing "when": "false") under `menus/commandPalette`.

dbalek (approved):

java/java.lsp.server/vscode/package.json
sdedic: Actually I think that actions like java.project.debug could be in the command palette ... they are able to work with editor URI to derive the target project, so I'll add categories to them. Not sure if I should use `Java`, or rather invent a `Project` ...

sdedic: Categories assigned in c3c3eb783c

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
sdedic: Arch document created and magic strings described in d56b619f25.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/ProjectActionParams.java
sdedic: Javadoc fixed in 11856ce30f

java/java.lsp.server/vscode/package.json
sdedic: Covered in d56b619

sdedic:
Rebased on latest `master` + version bump.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
JaroslavTulach: If it is a bug, can it be fixed?

JaroslavTulach (approved):
Having an `arch.xml` document describing the layer API and marking as `private` is all I can ask for at this stage, I guess. Thank you, Sváťo.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
sdedic: It should be - but I need to prioritize given my limited capacity: tracked as [NETBEANS-6296
(https://issues.apache.org/jira/browse/NETBEANS-6294)

PR #3364, opened 2021-12-11
Merge the code lens actions with the launch config properties and adapt Gradle to apply env.

entlicher:
Code lens actions are run via `runDebug` in `extension.ts` and they use a special launch configuration that did not inherit the configuration from `launch.json`. This is why any additional properties set in `launch.json` were not applied. We merge the launch configuration in `mergeWithLaunchConfig()`.

Gradle project run/debug now applies environment variables set via `runEnvironment` project's property.

`NetBeansRunSinglePlugin` can be simplified now, the set of the environment variables is removed and we set `System.in` and `runWorkingDir` for all Java executions.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
@@ -34,6 +34,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.StreamCorruptedException;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
@@ -63,6 +64,7 @@ import org.netbeans.spi.project.ui.support.BuildExecutionSupport;
 import org.openide.awt.StatusDisplayer;
 import org.openide.execution.ExecutorTask;
 import org.openide.filesystems.FileUtil;
+import org.openide.util.BaseUtilities;
 import org.openide.util.Lookup;
 import org.openide.util.NbBundle;
 import org.openide.util.NbBundle.Messages;
@@ -81,6 +83,7 @@ public final class GradleDaemonExecutor extends AbstractGradleExecutor {
 
     private CancellationTokenSource cancelTokenSource;
     private static final Logger LOGGER = Logger.getLogger(GradleDaemonExecutor.class.getName());
+    private static final String JAVA_HOME = "JAVA_HOME";    // NOI18N
 
     private final ProgressHandle handle;
     private InputStream inStream;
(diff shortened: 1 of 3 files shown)


extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
```diff
@@ -383,6 +411,88 @@ private static String gradleExecutable() {
         return Utilities.isWindows() ? "bin\\gradle.bat" : "bin/gradle"; //NOI18N
     }
 
+    private static String[] unescapeParameters(String s) {
+        final int NULL = 0x0;
```
sdedic, line 415: I was trying to find a difference between this and `BaseUtilities.parseParameters` -- if it is there, please put a comment in the source, so maintainer(s) are aware about the important differences.

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
```diff
@@ -282,6 +276,40 @@ public void statusChanged(org.gradle.tooling.events.ProgressEvent event) {
...
+                javaHome = platformProvider.getJavaHome().getCanonicalPath();
+            } catch (IOException ex) {
+                io.getErr().println(Bundle.NO_PLATFORM(ex.getMessage()));
+                gradleTask.finish(1);
+                return false;
+            }
+        }
+        if (javaHome != null || runEnvironment != null) {
+            Map<String, String> envs = new HashMap<>(System.getenv());
+            if (javaHome != null) {
```
sdedic, line 293: I'd maybe put the `javaHome` last, so it overrides values in the environment acquired from `runEnvironment` project property.
@lkishalmi what's your opinion ?

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -356,6 +356,27 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
             await commands.executeCommand(selected.userData.command.command, ...(selected.userData.command.arguments || []));
         }
     }));
+    const mergeWithLaunchConfig = (dconfig : vscode.DebugConfiguration) => {
```
sdedic, line 359: Excellent ! This will perfectly work with the project view's actions (run project, debug project).

sdedic (approved):

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
entlicher: Ah, thanks! I had a private copy elsewhere and forgot that it;s in BaseUtilities. I'll call it from there...

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
entlicher: I wasn't sure about the preferred order.

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
entlicher: Done.

extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java
entlicher: To be transparent, I've added a print of a warning: `Warning: JAVA_HOME environment variable is replaced with the current Java platform path <javaHome set to platformProvider.getJavaHome().getCanonicalPath()>`.

PR #3365, opened 2021-12-12
LSP: Allow to specify default DB connection

sdedic:
default-db-connection

This PR required an API change in `ConnectionManager` in the DB module - but IMHO quite beneficial, since the IDE code now can also use the 'preferred' connection for whatever features not originating from the DB Connection node. For example SQL embedded in Java / other code files may also benefit from the added API.

The LSP client may query the default connection and also set it - the value will be persisted by the NBLS.

### DIFF
--- a/ide/db/apichanges.xml
+++ b/ide/db/apichanges.xml
@@ -85,6 +85,18 @@ is the proper place.
     <changes>
         <change>
             <api name="database_explorer_api"/>
+            <summary>Concept of preferred connection introduced</summary>
+            <version major="1" minor="83"/>
+            <author login="sdedic"/>
+            <compatibility addition="yes"/>
+            <description>
+                Preferred connection can be defined to be used throughout the IDE as the default
+                or fallback one.
+            </description>
+            <class package="org.netbeans.api.db.explorer" name="ConnectionManager"/>
+        </change>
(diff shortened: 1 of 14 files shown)


matthiasblaesing:
There is already a concept in the netbeans codebase, that provides a capability similar to this. Please have a look at the PHP module `org.netbeans.modules.php.editor` in the `org.netbeans.modules.php.editor.sql`  package. The support implemented there provides a DB connection on a per project basis, which sounds like a good candiate to be generalized instead of introducing a single global connection.

sdedic:
> connection on a per project basis, which sounds like a good candiate to be generalized instead of introducing a single global connection.

Hm, I thought about project-specific setting as the next step, having the (traditional) hierarchy of Project > global setting. After all, the data sources (from db explorer) are global, so it makes sense. I'll try to invent / promote concept from PHP directly into `db` module -- filed as [NETBEANS-6288](https://issues.apache.org/jira/browse/NETBEANS-6288)

Thanks for the insight.

dbalek (approved):
Looks fine.

sdedic:
Spurious error in commit-validation / MacOS X + also random error in java testsuite unrelated. Merging.

PR #3366, opened 2021-12-13
Avoid spurious exception in log. Warn if perf counters are enabled on incompatible JDK

sdedic:
Groovy editor's internal performance counters override `final` groovy compiler's member field's value using reflection, which does not work on JDK17. These performance counters are likely to become obsolete so the easiest way how to avoid mostly harmless notification in the `message.log` is to log it only if the performance dumps are enabled.

// cc: @ppisl


ppisl:
I don't have nothing against this. But I don't have rights to approve it :).

JaroslavTulach (approved):

sdedic:
something in java module testsuite failed, but this is extide - unrelated.

PR #3367, opened 2021-12-13
LSP: Provide completion for SQL queries for repository annotations.

dbalek:
Completion for SQL queries for repository annotations should be available via LSP.

### DIFF
--- a/java/j2ee.persistence/nbproject/project.xml
+++ b/java/j2ee.persistence/nbproject/project.xml
@@ -53,6 +53,15 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.api.lsp</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>1</release-version>
+                        <specification-version>1.7</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
                     <code-name-base>org.netbeans.api.progress</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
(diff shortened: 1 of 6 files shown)


java/j2ee.persistence/nbproject/project.xml
```diff
@@ -52,6 +52,15 @@
                         <specification-version>1.0</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.api.lsp</code-name-base>
```
JaroslavTulach, line 56: Adding such lightweight dependency is OK.

java/java.lsp.server/nbcode/nbproject/platform.properties
```diff
@@ -325,12 +311,10 @@ disabled.modules=\
     org.netbeans.modules.xml.catalog.ui,\
     org.netbeans.modules.xml.jaxb,\
     org.netbeans.modules.xml.jaxb.api,\
-    org.netbeans.modules.xml.multiview,\
```
JaroslavTulach, line 328: A bit too many new dependencies. Could some of them be eliminated by restructuring the modules? I am sure you don't really need `xml.multiview` in VSCode. Why do we need `schema2beans`? `eclipselink`?

I understand that currently the needed modules (`*.persistance`?) have such dependencies. What it would take to eliminate some of them?

JaroslavTulach (approved):
I am concerned with the amount of new modules going into VSNetBeans extension. There should be a plan to make the extension slimmer again...

java/java.lsp.server/nbcode/nbproject/platform.properties
dbalek: The problem is that the `j2ee.persistence` module is huge and its split into  `j2ee.persistence.base` and  `j2ee.persistence` would require a substantial amount of time. We should IMHO plan some NBLS module cleanup for the next release.

sdedic: I filed [NETBEANS-6295](https://issues.apache.org/jira/browse/NETBEANS-6295) ... the LSP is getting bloated. We should also invent a way how the project- db- etc supports can plug into the basic LSP protocol server so that we don't increase the dependency nexus. So filed also [NETBEANS-6296](https://issues.apache.org/jira/browse/NETBEANS-6296)

sdedic (approved):

PR #3368, opened 2021-12-13
[NETBEANS-6291] Navigation to fields sometimes doesn't work in Groovy

ppisl:
This fix doesn't cover all cases, but at least in the simple cases it works.


sdedic (approved):
Looks fine.

sdedic:
Commit validation on Mac is a transient failure, merging.

PR #3369, opened 2021-12-14
Single file Groovy Run and Debug in VSCode.

entlicher:
1. When we get no ClassPath in `GroovyBreakpointFactory.java`, we suppose the file name for single-file execution.
2. Prevent from NPE when project is null in `WorkspaceServiceImpl.java`
3. `findProjectConfigurations` need to return the desired value and not wrap it in the Future.


entlicher:
I've added a check that assures that COMMAND_DEBUG_TEST_SINGLE is called with the context file, till https://issues.apache.org/jira/browse/NETBEANS-6297 is fixed.

jtulach (approved):

PR #3372, opened 2021-12-15
Remove obsolete permgen flags.

mbien:
Java 7 was the last release with a permgen, Java 8 replaced it with metaspace. Java 17 does not start if PermSize is set.

-> lets remove it, brings NB one small step closer to be testable on JDK 17


matthiasblaesing (commented):
Looks sane to me - lets see if the unittests succeed after restart.

matthiasblaesing:
Travis and github are both happy and JDK 8 must die, so lets get this in.

PR #3373, opened 2021-12-15
Change nb.sass.libsass to TRUE, to fix broken SASS

t-oster:
With default NetBeans modern sass implementations cannot be used, because it defaults to the unmaintained ruby-sass implementation, which supports the command line switch **--cache-location**.

There is a System property to switch that but it defaults to ruby sass. This PR just changes the default value, so most users can just use modern sass and the minority using ruby-sass can set the value to false expliciteley.

There is still an issue with the sourcemaps option, but until #1234 is sorted out, this PR will make Sass usable again for most users, which is IMHO priority.

### DIFF
--- a/ide/css.prep/src/org/netbeans/modules/css/prep/sass/SassCli.java
+++ b/ide/css.prep/src/org/netbeans/modules/css/prep/sass/SassCli.java
@@ -51,9 +51,10 @@ public abstract class SassCli {
 
     // #247890
     /**
-     * System property to be set to "true" if <tt>libsass</tt> should be used.
+     * System property to be set to "false" if <tt>libsass</tt> should be not be used
+     * for legacy RubySass implementation (https://github.com/apache/netbeans/pull/1234)
      */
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@t-oster not all unittests are rock solid, so some of just fail randomly, I restarted the failing tests and they succeeded now.

As already pointed out in #1234 this is not the final solution - but it is at least a bit better.

Closes: #1234

t-oster:
Thanks for the quick reply and merging. Yes, I know this is not the final solution. We should either support different implementations by dropdown or autodetection or just remove all command line flags, which are not present in every commonly used implementation (e.g. remove the source-maps checkbox) and leave it to the user to provide the correct command line.

However just flipping that bit makes it at least possible to use for the vast majority right now without modifying startup scripts, so IMHO this is better.

Maybe we should not close #1234 or at least create a follow-up issue so that somebody can pick up the issue with the different parameters for source maps.

t-oster:
(sorry, I forgot #1234 is just a PR and your issues are in bugzilla, so ignore my last comment)

PR #3375, opened 2021-12-16
NETBEANS-6307: check global artifact cache, ignore per-project cached  structure if not consistent.

sdedic:
For detailed explanation, see [bug description](https://issues.apache.org/jira/browse/NETBEANS-6305). The fix is actually a workaround - the cached project structure/information is validated against global artifact cache. The cached info is rejected if the global artifact cache does not contain items that would have been present/updated if the current IDE instance had loaded the project fully and created the caches (both global and project).

A test is attached that demonstrates the issue, if [the validation call](https://github.com/apache/netbeans/pull/3375/files#diff-e0b2de3f6a585632d92c8419e75f246740dcb1f90a52b33e4a58147423b69bdbL45) is removed.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleArtifactStore.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleArtifactStore.java
@@ -28,12 +28,18 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.nio.file.Files;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.swing.event.ChangeListener;
+import org.netbeans.modules.gradle.GradleModuleFileCache21;
 import org.netbeans.modules.gradle.GradleProject;
 import org.openide.modules.OnStart;
 import org.openide.modules.Places;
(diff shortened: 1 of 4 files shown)


sdedic:
72hours elapsed; reviewers, please comment / suggest improvements. Otherwise I'll merge in the change on 21th Dec.

extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleArtifactStore.java
```diff
@@ -159,7 +171,54 @@ void verify() {
             store();
         }
     }
-
+    
+    /**
+     * Checks that all dependencies the project thinks should be in the global cache
+     * are actually in the global cache. If the global artifact cache does not contain
+     * an entry from a resolved depency in the project cache then many random failures can
```
JaroslavTulach, line 178: typo: dependency

extide/gradle/test/unit/src/org/netbeans/modules/gradle/loaders/DiskCacheProjectLoaderTest.java
```diff
@@ -0,0 +1,169 @@
...
+                        addTest(DiskCacheProjectLoaderTest.class).
+                        addTest("testLoadCachedProject").
+                        reuseUserDir(false).
+                        gui(false).
+                        enableModules(".*gradle.*").
+                        reuseUserDir(false);
+        
+        Test run2 = NbModuleSuite.create(conf);
+        ts.addTest(run1);
+        ts.addTest(run2);
```
JaroslavTulach, line 81: A test with a restart of the IDE! Amazing.

JaroslavTulach (approved):
Great work creating the test. That must mean something has been fixed!

entlicher (approved):

sdedic:
Transient errors went off, merging.

PR #3376, opened 2021-12-16
Really convert IDs to numbers to use with Map. Re-fetch items that have been marked as changed.

sdedic:
This PR fixes two bugs. First - call me an idiot, but I used just syntax-sugar conversion instead of `Number(x)` to convert item IDs (which are numeric) to numbers. I did not notice it during var inspection in the debugger (just '9' vs 9) ... therefore items were not matched and updated properly, the items were never found in the `treeData` Map.

Second is a race condition: sometimes, especially with lazy-initialized Nodes in NBLS, a query for Children started to initialize the Children, and so a certain `nodes/nodeChanged` could be received **before** the vscode even received the first `nodes/info` response for the node; the node change was then simply ignored. That results in dangling nodes, nodes like 'Please wait...' which are in the tree indefinitely etc.

The fix is to track those pending changes, and delay `Promise<TreeItem>` completion after a new `nodes/info` response arrives.

### DIFF
--- a/java/java.lsp.server/vscode/src/explorer.ts
+++ b/java/java.lsp.server/vscode/src/explorer.ts
@@ -92,6 +92,7 @@ class VisualizerProvider extends vscode.Disposable implements CustomizableTreeDa
   private root: Visualizer;
   private treeData : Map<number, Visualizer> = new Map();
   private decorators : TreeItemDecorator<Visualizer>[] = [];
+  private pendingRefresh : Set<number> = new Set();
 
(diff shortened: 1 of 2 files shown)


sdedic:
I am stil hitting an issue with vscode -- when multiple change - getTreeItems are called, the vscode infrastructure errors out. That does not happen when "normal" children re-fetches, like when a new file is created.

JaroslavTulach (approved):

PR #3377, opened 2021-12-17
Rewrite lambda to an inner class to prevent from Gradle warning about execution optimizations.

entlicher:
When a Micronaut Gradle app is run, following message is printed:
```
> Task :run
Execution optimizations have been disabled for task ':run' to ensure correctness due to the following reasons:
  - Property 'jvmArgumentProviders.$0' was implemented by the Java lambda 'org.netbeans.modules.gradle.tooling.NetBeansRunSinglePlugin$$Lambda$1154/0x00000008409ea840'. Reason: Using Java lambdas is not supported as task inputs. Please refer to https://docs.gradle.org/7.3.1/userguide/validation_problems.html#implementation_unknown for more details about this problem.
```
While this is not a severe issue, it's distracting and the user has a little chance to understand what this means.
We need to replace the lambda with an anonymous class in `NetBeansRunSinglePlugin` to get rid of the message.


mbien:
please add a short comment so that nobody converts it to a lambda again in future code cleanups.

lkishalmi (approved):
Thank you for this one!

PR #3379, opened 2021-12-17
[NETBEANS-6312] Assure that TokenHierarchy is called with Document's read lock.

entlicher:
The `TokenHierarchy.isActive()` is not called with the associated Document's read lock, as can be seen from the stack trace in the associated issue.
This change adds read lock to `SyntaxHighlighting`, which makes `TokenHierarchy.isActive()` to be called with the read lock.

### DIFF
--- a/ide/editor.util/src/org/netbeans/lib/editor/util/swing/DocumentUtilities.java
+++ b/ide/editor.util/src/org/netbeans/lib/editor/util/swing/DocumentUtilities.java
@@ -472,24 +472,21 @@ public final class DocumentUtilities {
      */
     public static boolean isReadLocked(Document doc) {
         if (checkAbstractDoc(doc)) {
-            if (isWriteLocked(doc))
+            if (isWriteLocked(doc)) {
                 return true;
-            if (numReadersField == null) {
-                Field f = null;
+            }
+            Field f = numReadersField;
+            if (f == null) {
                 try {
                     f = AbstractDocument.class.getDeclaredField("numReaders"); // NOI18N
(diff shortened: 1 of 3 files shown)


ide/editor.util/src/org/netbeans/lib/editor/util/swing/DocumentUtilities.java
```diff
@@ -65,9 +65,9 @@
     
     private static final Object TYPING_MODIFICATION_KEY = new Object();
     
-    private static Field numReadersField;
+    private static volatile Field numReadersField;
```
sdedic, line 68: BTW is this necessary, strictly speaking ? The `nuimReadersField` may be eventually initialized several times (a thread that was already started at the time of the previous assignment could see `null`) - but the Field instance should point to the same data (?).  But bcs of number of invocations (and other heavy locking on Document) I guess it makes no difference between volatile/nonvolatile.

entlicher: I agree that the synchronization is wrong. Looking at it, I'll adjust it a bit...

entlicher: I've removed the wrong synchronization, it's up to the volatile flag only now. It can initialize multiple times, but after that it's just one volatile access.

sdedic (approved):
Thanks!

sdedic:
All clear, merging.

PR #3380, opened 2021-12-18
[NETBEANS-6315] Escape HTML entities in PHPStan report

KacerCZ:
https://issues.apache.org/jira/browse/NETBEANS-6315

- Escapes HTML entities in message from PHPStan report

Before - message on right side of Inspector does not show `<string>`:
netbeans-phpstan-message-before

After - message on right side is complete:
netbeans-phpstan-message-after


KacerCZ:
@junichi11 Please check if this PR is OK. I'm not sure about module version change.

KacerCZ:
@junichi11 Thank you.

PR #3382, opened 2021-12-20
Delay change events for node after the node's pending queries are resolved

sdedic:
This PR is rather intricate. It turned out that `vscode` is probably confused by aggressive change events fired from the NBLS. The PR attempts to buffer changes for nodes that are being just queried by vscode: if such a change is received, it is put aside. Once queries for the node (children, info) are resolved, a vscode node change will be fired for the node - which will probably result in an additional `getTreeItem` or `getChildren` call. 

With this queue in place, nodes seem to refresh properly.

I added some logging (now disabled by a constant in the code) - but logs to the Output channel of the NBLS extension: if the LSP trace is enabled at the same time, the LSP message traces are interleaved with the tree processing diagnostics. It's easier to find out what and when is happening.

### DIFF
--- a/java/java.lsp.server/vscode/src/explorer.ts
+++ b/java/java.lsp.server/vscode/src/explorer.ts
@@ -291,7 +291,6 @@ class VisualizerProvider extends vscode.Disposable implements CustomizableTreeDa
   delayedFire : Set<Visualizer> = new Set<Visualizer>();
 
   async fetchItem(n : number) : Promise<Visualizer> {
-    let refreshVis = this.treeData.get(n);
     let d = await this.client.sendRequest(NodeInfoRequest.info, { nodeId : n });
     let v = new Visualizer(d, this.ts.imageUri(d));
     if (d.command) {
--- a/java/java.lsp.server/vscode/src/explorer.ts
+++ b/java/java.lsp.server/vscode/src/explorer.ts
@@ -4,14 +4,19 @@ import {  LanguageClient } from 'vscode-languageclient/node';
 import { NbLanguageClient } from './extension';
 import { NodeChangedParams, NodeInfoNotification, NodeInfoRequest } from './protocol';
 
+const doLog : boolean = false;
 export class TreeViewService extends vscode.Disposable {  
+  
   private handler : vscode.Disposable | undefined;
   private client : NbLanguageClient;
   private trees : Map<string, vscode.TreeView<Visualizer>> = new Map();
   private images : Map<number, vscode.Uri> = new Map();
   private providers : Map<number, VisualizerProvider> = new Map();
-  constructor (c : NbLanguageClient, disposeFunc : () => void) {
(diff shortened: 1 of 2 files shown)


JaroslavTulach (approved):
It is my understanding that this change is needed. Hence my approval.

However I am not able to get the _architectural idea_ of the change. All I see is that the code gets complex. We should think about writing some unit tests and possibly ensuring some invariants that we want to hold.

java/java.lsp.server/vscode/src/explorer.ts
```diff
@@ -92,11 +97,11 @@ class VisualizerProvider extends vscode.Disposable implements CustomizableTreeDa
   private root: Visualizer;
   private treeData : Map<number, Visualizer> = new Map();
   private decorators : TreeItemDecorator<Visualizer>[] = [];
-  private pendingRefresh : Set<number> = new Set();
-
+  
   constructor(
     private client: LanguageClient,
     private ts : TreeViewService,
+    private log : vscode.OutputChannel,
```
entlicher, line 104: `this.log = log` is missing in the constructor body.

java/java.lsp.server/vscode/src/explorer.ts
```diff
@@ -154,42 +169,131 @@ class VisualizerProvider extends vscode.Disposable implements CustomizableTreeDa
...
       }
     }
-
     return f(element.copy());
   }
 
+  delayedFire : Set<Visualizer> = new Set<Visualizer>();
+
   async fetchItem(n : number) : Promise<Visualizer> {
+    let refreshVis = this.treeData.get(n);
```
entlicher, line 294: `refreshVis` is not used anywhere.

entlicher: If refreshVis exists, it's not necessary to create it below, is that right?

entlicher (changes requested):

java/java.lsp.server/vscode/src/explorer.ts
sdedic: Right it is a leftover. Updates of the existing `Visualizer` instances are done in `fetchItem` callers.

java/java.lsp.server/vscode/src/explorer.ts
sdedic: I thought I've used [Parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) in the constructor with an automatic assignment to the member.

entlicher: Aha, O.K. then it's fine. Thanks for explanation.

entlicher (approved):

java/java.lsp.server/vscode/src/explorer.ts
entlicher: Thanks.

PR #3384, opened 2021-12-20
Disable client-side projects and nodejs support.

sdedic:
The NBLS distribution contains a broken client-side project support: some modules are there, some are not, like HTML parser & co. Given the feature is not really used (vscode as the primary LSP client uses a different langserver for JS, Node) and it is broken anyway, I'd prefer to remove it form the NBLS distribution.

We can later add it back, but with all dependencies so that it really works. Now the client-side project errors out with an exception:
```
java.lang.IllegalArgumentException: No parser for mime type: text/html
        at org.netbeans.modules.parsing.nb.DataObjectEnvFactory.findMimeParser(DataObjectEnvFactory.java:78)
        at org.netbeans.modules.parsing.api.ParserManager.findParser(ParserManager.java:374)
        at org.netbeans.modules.parsing.api.ParserManager.parseWhenScanFinished(ParserManager.java:339)
        at org.netbeans.modules.web.clientproject.remotefiles.RemoteFiles.update(RemoteFiles.java:77)
        at org.netbeans.modules.web.clientproject.remotefiles.RemoteFiles.getRemoteFiles(RemoteFiles.java:112)
        at org.netbeans.modules.web.clientproject.remotefiles.RemoteFilesNodeFactoryImpl$RemoteFilesNodeListImpl.keys(RemoteFilesNodeFactoryImpl.java:64)
        at org.netbeans.spi.project.ui.support.NodeFactorySupport$DelegateChildren$1.run(NodeFactorySupport.java:177)
```

### DIFF
--- a/java/java.lsp.server/nbcode/nbproject/platform.properties
+++ b/java/java.lsp.server/nbcode/nbproject/platform.properties
@@ -127,6 +127,7 @@ disabled.modules=\
     org.netbeans.modules.debugger.jpda.js,\
     org.netbeans.modules.debugger.jpda.jsui,\
     org.netbeans.modules.debugger.jpda.kit,\
+    org.netbeans.modules.debugger.jpda.trufflenode,\
     org.netbeans.modules.debugger.jpda.visual,\
     org.netbeans.modules.derby,\
     org.netbeans.modules.dlight.nativeexecution.nb,\
@@ -204,6 +205,7 @@ disabled.modules=\
     org.netbeans.modules.javascript.gulp,\
     org.netbeans.modules.javascript.jstestdriver,\
     org.netbeans.modules.javascript.karma,\
+    org.netbeans.modules.javascript.nodejs,\
     org.netbeans.modules.javascript.v8debug.ui,\
     org.netbeans.modules.javascript2.debug.ui,\
     org.netbeans.modules.javascript2.extdoc,\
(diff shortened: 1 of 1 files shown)


java/java.lsp.server/nbcode/nbproject/platform.properties
```diff
@@ -127,6 +127,7 @@ disabled.modules=\
     org.netbeans.modules.debugger.jpda.js,\
     org.netbeans.modules.debugger.jpda.jsui,\
     org.netbeans.modules.debugger.jpda.kit,\
+    org.netbeans.modules.debugger.jpda.trufflenode,\
```
JaroslavTulach, line 130: I am not really sure about this one. Will https://github.com/graalvm/graal-js-archetype continue to work/be debugable?

JaroslavTulach (approved):
However it is clear we don't want HTML project.

java/java.lsp.server/nbcode/nbproject/platform.properties
sdedic: @entlicher should know...

entlicher (approved):

java/java.lsp.server/nbcode/nbproject/platform.properties
entlicher: The `jpda.trufflenode` binds to the node.js support in NetBeans. As it does not make sense to use an old node.js support from NetBeans in VSCode, it's fine to remove it.
We would need to invent a different approach in VSCode how to execute debugger for node.js projects via Truffle on GraalVM to connect node.js & Java debugging.

PR #3385, opened 2021-12-22
[NETBEANS-6328] avoid IllegalArgumentException in FoldManagerImpl

MKuettner:
Extracts FoldInfo list computation into a new method in order to be testable in unit tests. 
Adds some unit tests.
Introduces Utils.getEndCharacter(..) that gets last character index of given line.
Changes FoldInfo calculation in case of FoldingRanges endCharacter is NULL.

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImpl.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImpl.java
@@ -113,36 +113,11 @@ public class FoldManagerImpl implements FoldManager, BackgroundTask {
         EditorCookie ec = file.getLookup().lookup(EditorCookie.class);
         Document doc = ec != null ? ec.getDocument() : null;
         if (doc == null) {
-            return ;
+            return;
         }
-        Set<FoldingRangeInfo2> foldingRangesSeen = new HashSet<>();
+
         List<FoldingRange> ranges = computeRanges(bindings, file);
-        List<FoldInfo> infos = new ArrayList<>();
-        if (ranges != null) {
-            for (FoldingRange r : ranges) {
-                int start = Utils.getOffset(doc, new Position(r.getStartLine(), r.getStartCharacter() != null ? r.getStartCharacter() : 0));
-                int end;
-                if (r.getEndCharacter() == null) {
-                    end = Utils.getOffset(doc, new Position(r.getEndLine() + 1, 0)) - 1;
-                } else {
-                    end = Utils.getOffset(doc, new Position(r.getEndLine(), r.getEndCharacter()));
-                }
-                // Map the fold range type to netbeans as far as possible
-                FoldType foldType;
-                if("comment".equals(r.getKind())) {
-                    foldType = FoldType.COMMENT;
-                } else if ("imports".equals(r.getKind())) {
-                    foldType = FoldType.IMPORT;
-                } else {
-                    foldType = FoldType.CODE_BLOCK;
-                }
-                FoldingRangeInfo2 fri2 = new FoldingRangeInfo2(start, end, foldType);
-                if(! foldingRangesSeen.contains(fri2)) {
-                    infos.add(FoldInfo.range(start, end, foldType));
-                    foldingRangesSeen.add(fri2);
-                }
-            }
-        }
+        List<FoldInfo> infos = computeInfos(doc, ranges);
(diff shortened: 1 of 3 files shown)


ide/lsp.client/test/unit/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImplTest.java
```diff
@@ -0,0 +1,150 @@
...
+    @Test
+    public void computeFoldInfosWithoutStartEndCharacterTest() throws Exception {
+        Document doc = createDocument(EXAMPLE_SIMPLE);
+
+        List<FoldingRange> ranges = new ArrayList<>();
+        ranges.add(new FoldingRange(0, 2));
+
+        List<FoldInfo> infos = FoldManagerImpl.computeInfos(doc, ranges);
+        assertEquals(1, infos.size());
+        assertEquals(0, infos.get(0).getStart());
```
matthiasblaesing, line 68: Is this correct? My reading of the documentation indicates, that this should be 6 (length of first line):

    /**
	 * The zero-based character offset from where the folded range starts. If
	 * not defined, defaults to the length of the start line.
	 */
	startCharacter?: uinteger;

If I read that correctly to get a fold, that starts at the beginning of the line, you need to explicitly supply the `startCharacter`

MKuettner: This is the original behavior of FoldManagerImpl. I just added some unit tests before bug fixing to make sure everything is fine afterwards.
From documentation point of view i agree with you. But in running IDE everything seems to work fine. 
I manually tested a lot of scenarios using a Python language server.

matthiasblaesing (approved):
I had a look at this and it looks sane to me. For the start character however I had another look at it and found, that our implementation is indeed not ideal, so I adjusted it:

https://github.com/matthiasblaesing/netbeans/commit/9ad251f671d0f6fd3a6f9d2d3fde28eb80dbc7f3

It changes the display from (ignore colors, just the structural change of the folds is relevant):

image

to:

image

It would be good, if you could test my change with the python LSP and either pull it into this PR or indicate, that this PR should be merged independently, then I'll merge this and add my own PR.

Thank you!

ebresie:
Not directly related but are you working on a LSP Python implementation at the moment?

ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/FoldManagerImpl.java
```diff
@@ -113,49 +113,56 @@ public void run(LSPBindings bindings, FileObject file) {
...
+            });
+        });
+    }
+    
+    static List<FoldInfo> computeInfos(Document doc, List<FoldingRange> ranges) {
+        Set<FoldingRangeInfo2> foldingRangesSeen = new HashSet<>();
         List<FoldInfo> infos = new ArrayList<>();
         if (ranges != null) {
             for (FoldingRange r : ranges) {
                 int start = Utils.getOffset(doc, new Position(r.getStartLine(), r.getStartCharacter() != null ? r.getStartCharacter() : 0));
```
sdedic, line 139: This is probably the place @matthiasblaesing  noted in the earlier comment: if `startCharacter` is nto defined, then, according to the documentation, it should equal to the length of the start line == the end of that line.

Please either correct the expression, or make a comment that the code diverges from the spec and why exactly is so.

MKuettner:
@matthiasblaesing:  I tested your changes with the Python LSP and it seems to works fine (it looks much better than before).
It would be nice if you could create an own PR and merge it independently. Thanks!

MKuettner:
@ebresie: I'm currently integrating an existing [Python LSP](https://github.com/palantir/python-language-server) into a Netbeans RCP. This RCP acts as a frontend for a Python scripting framework where parts of the Python code are generated from a data source and other parts can be scripted by the user. 

Except some smaller issues this LSP works quite OK.

matthiasblaesing:
@MKuettner thanks for the check, will merge. Of course the same remark as in the last merges applies: I assume, that you read the license headers on top of the changed/added files and agree that the changes are donated to the ASF.
@sdedic I'll create a PR with the fix for the wrong start offset shortly after this is merged (https://github.com/matthiasblaesing/netbeans/commit/9ad251f671d0f6fd3a6f9d2d3fde28eb80dbc7f3)

PR #3386, opened 2021-12-26
Replacing SecurityManager with a tracking agent

jlahoda:
This is a prototype of an attempt to replace the deprecated SecurityManager with a custom agent-based hooks.

What functionality is kept:
-preventing inadvertent System.exit
-tracking file I/O (probably the most complex part, might be missing some hooks)
-checking for deprecated system property access
-checking calls to setAccessible to open Unsafe (this didn't work on JDK 8+, but can be fixed)
-checking/preventing calls to System.setSecurityManager. Possibly no longer that critical, as it is not possible to disable the hooks using this method.
-tracking new Window creation (used by internal execution)

What functionality is not kept:
-`checkConnect` for connecting over network. This didn't seem to be actually used, as the original code was:
```
        try {
            checkPermission(allPermission);
            return;
        } catch (SecurityException e) {
        }
//proceed with the actual check
```
But `checkPermission` seems to always(?) pass, so the check was never performed
-`checkLogger`, which was disabled anyway
-tweaks related to installation of a custom Swing clipboard - these appear to exists for JDK-4818143, which is marked as fixed, so presumably not needed anymore.

When looking at the patch, `TrackingAgent` is the agent class that does various transformations to install hooks, `TrackingHooks` are the callback, `TopSecurityManager` (`o.n.bootstrap`), `FileChangedManager` (`masterfs`), `SecMan` (`core.execution`) are the actual uses of the hooks.


mbien:
this is pretty cool!

nbbuild/templates/common.xml
```diff
@@ -532,6 +532,11 @@
                 <isset property="test-have-jdk-1.9"/>
             </or>
         </condition>
+        <condition property="test.tracking.agent.dir" value="${nbplatform.active.dir}/platform/core">
+            <available file="${nbplatform.active.dir}/platform/core" type="dir" />
+        </condition>
+        <property name="test.tracking.agent.dir" location="${o.n.bootstrap.dir}/core/" />
+        <property name="test.tracking.agent.args" value="-javaagent:${test.tracking.agent.dir}/tracking-agent.jar -Xbootclasspath/a:${test.tracking.agent.dir}/tracking-hooks.jar"/>
```
JaroslavTulach, line 539: Having to manually modify `-Xbootclasspath` is annoying. Can't agent modify classpath programatically? There are methods like [appendToBootstrap...](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#appendToBootstrapClassLoaderSearch-java.util.jar.JarFile-).

JaroslavTulach (commented):
Magnificent work.

I'd prefer if the code was packaged as NetBeans module for ease of tooling as well as packaging and deploying. With proper Javadoc it could even be used as a great standalone library.

Inviting @sdedic as a reviewer. I'd say Sváťa is long time NetBeans expert when it comes to bytecode patching.

platform/o.n.bootstrap/nbproject/org-netbeans-bootstrap.sig
```diff
@@ -558,27 +499,17 @@ meth public abstract void flushCaches(java.io.DataOutputStream) throws java.io.I
 
 CLSS public org.netbeans.TopSecurityManager
```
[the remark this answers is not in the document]
jlahoda: Given this is a friend API, I don't see such a strong need to keep compatibility, esp. in an aspect that will cease to work at some point.

platform/core.execution/src/org/netbeans/core/execution/Install.java
```diff
@@ -88,7 +89,7 @@
     private static final Logger LOG = Logger.getLogger(Install.class.getName());
     
     public @Override void run() {
-        TopSecurityManager.register(SecMan.DEFAULT);
+        TrackingHooks.register(SecMan.DEFAULT, 100, TrackingHooks.HOOK_EXIT, TrackingHooks.HOOK_NEW_AWT_WINDOW);
```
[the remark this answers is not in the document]
jlahoda: Seems fairly complex, as there are uses that actively use the pre-existing register/unregister mechanism in `TopSecurityManager`:
https://github.com/apache/netbeans/blob/ebb8086be371e8b3ec502a10cf9674c1bfb49dca/platform/core.execution/src/org/netbeans/core/execution/Install.java#L91
https://github.com/apache/netbeans/blob/ebb8086be371e8b3ec502a10cf9674c1bfb49dca/platform/core.execution/src/org/netbeans/core/execution/Install.java#L100

It is not clear to me how to replace that with a Lookup-based registration. But if you can provide a patch to resolve all existing uses of `TopSecurityManager.register/unregister`, then we can switch to a Lookup-based registration.

platform/o.n.bootstrap/netbeans-tracking-agent/tracking-agent/src/org/netbeans/agent/TrackingAgent.java
```diff
@@ -0,0 +1,1099 @@
...
+                        
+                        List<Integer> newConstantPoolEntries = new ArrayList<>();
+
+                        for (String cpEntry : me.get().constantPool) {
+                            newConstantPoolEntries.add(cpLastEntry[0]++);
+                            byte[] data = decodeData(cpEntry, newConstantPoolEntries);
+                            additionalConstantPool.write(data);
+                        }
+
+                        int maxStack = readShort(classfileBuffer, pp); pp += 2;
```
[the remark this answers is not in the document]
jlahoda: On one hand, it would be easier to use ASM. On the other hand, ASM cannot handle new classfiles - we could workaround that by downgrading the versions of the classfiles sent to ASM; it would also mean to place ASM into the agent path, or doing some trampoline from the agent to a normal module. I guess I'd try with the handcoded bytecode patcher, and we'll see what are the experiences with that.

jlahoda:
I've updated the patch to reflect most of the comments (tracking agent and hooks are now modules, SPI is separate from the callback, etc.) I've put a comment on conversation where I think it is difficult to implement the requested change.

jlahoda:
I wonder if there are any further comments on this. It might be reasonable to try to avoid too much dependency on the SecurityManager.

matthiasblaesing:
Given that we are early in the NB16 cycle, now would be a good time to merge this if ready.

platform/o.n.core/src/org/netbeans/core/NbLifecycleManager.java
```diff
@@ -67,9 +67,6 @@ public static void advancePolicy() {
         if (!Boolean.getBoolean("TopSecurityManager.disable")) {
             // set security manager
             TopSecurityManager.install();
-            if (CLIOptions.isGui()) {
-                TopSecurityManager.makeSwingUseSpecialClipboard(Lookup.getDefault().lookup(ExClipboard.class));
```
sdedic, line 71: What was this supposed to do ?

jlahoda: I believe this was part of a workaround for JDK-4818143. That bug is marked as fixed in JDK 6. So, it seems that keeping the workaround is probably not useful anymore (and, in case the problem would not be actually fixed, this is a way to find out).

neilcsmith-net: Doesn't that also stop the system clipboard being an instance of ExClipboard?  If so, seem a bunch of places that expect it to be one.

.travis.yml
```diff
@@ -152,6 +152,7 @@ matrix:
             - travis_retry hide-logs.sh ant $OPTS -f platform/masterfs test
             - hide-logs.sh ant $OPTS -f platform/masterfs.linux test
             - hide-logs.sh ant $OPTS -f platform/netbinox test  -Dtest.config=stableBTD
+            - hide-logs.sh ant $OPTS -f platform/o.n.agent test
```
mbien, line 155: this job moved to github. This could be added to `platform-modules-test1` or `platform-modules-test2`
https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml#L727

test1 is only JDK 8. test2 has some JDK 11 tests too.

mbien:
all green. excellent!

mbien:
[JEP 451: Prepare to Disallow the Dynamic Loading of Agents](https://openjdk.org/jeps/451)
will require the `-XX:+EnableDynamicAgentLoading` flag (sooner or later, will print warnings first).

It might be better to load the agent using JVM flags on startup (instead of dynamically), since there are probably good reasons JEPs like that exist.

neilcsmith-net:
@mbien isn't it already doing this - https://github.com/apache/netbeans/pull/3386/files#diff-d4b605d5fdbc82d1afa51dc1d6c3ce44516cf46c5e2ac0c79a98118bdcefb896R195

Or are there other uses in there?  I assume we'll need a Windows launcher update too?  Wish that could be a script too!

PR #3342, opened 2021-11-28
Fixing pattern matching for binding patterns

jlahoda:

PR #3388, opened 2021-12-26
Clean merge of "Fixing the configure hint sub-fix to work more properly for per-project hint settings."

matthiasblaesing:
The specification version was bumped and thus documentation (api  changes and javadoc) need to be updated.

Work was done and reviewed in #3342


matthiasblaesing:
Small correction, work was done and reviewed in #3147

PR #3391, opened 2021-12-27
Updated Subversion support to 1.14.0

lkishalmi:
Well, let's see if we could just update to SubVersion 1.14 on JavaHL.
Also made the SvnClientAdapterFactory so it would be easier to re-implement the SvnKit adapter, whoever would do that.

I'd also like to test if the current (1.14.0) JavaHL would also work with version 1.13 (Ubuntu 20.04) and 1.9 (Ubuntu 18.04) as well.

### DIFF
--- a/ide/libs.svnClientAdapter.javahl/external/adapter-javahl-1.14.0-license.txt
+++ b/ide/libs.svnClientAdapter.javahl/external/adapter-javahl-1.14.0-license.txt
@@ -1,6 +1,6 @@
 Name: SVNClientAdapter Library
 Description: Integration library for subversion client
-Version: 1.10.12
+Version: 1.14.0
 Origin: Subclipse
 License: Apache-2.0
 URL: http://subclipse.tigris.org/svnClientAdapter.html
@@ -206,3 +206,4 @@ URL: http://subclipse.tigris.org/svnClientAdapter.html
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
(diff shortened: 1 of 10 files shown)


lkishalmi:
The bad thing is, I've needed to upload the libraries to OSUL, as the Maven packages on GitHub requires authorization even if they are public.

matthiasblaesing:
> The bad thing is, I've needed to upload the libraries to OSUL, as the Maven packages on GitHub requires authorization even if they are public.

We have options to download from other sources - one such example is:

https://github.com/apache/netbeans/blob/6cf2bda0ea4c1937c7708539038a6a83dced8b2b/extide/gradle/external/binaries-list#L18

I will not say that this is optimal, but we don't necessarily need maven central as a source, just a stable download from a source we expect to last (maven central is proven, the rest still has to prove, whether they'll provide the same stability).

lkishalmi:
Well, I've meant to use that kind of download method. Though I doubt it supports github authentication out of the box. Yes, that authentication is still required for downloading otherwise public artifacts. I did not feel the energy to add a support for that one. It would also require anyone who would build the source to have a GutHub account. (Maybe we can use a never expiring service key that we could pack in our code, but I have not elaborated that way at all.)

matthiasblaesing:
Thanks - another service with good intentions and a broken implementation. One should think, that making a file downloadable with https at a fixed location should be a no brainer. Adding support for such a broken scheme would IMHO be an error, who knows how long github will keep this broken scheme around...

lkishalmi:
Just some reading on that topic: https://github.community/t/download-from-github-package-registry-without-authentication/14407

neilcsmith-net:
Still marked WIP so punting to NB14.

neilcsmith-net:
OK, thanks @lkishalmi - just made it! :wink:

PR #3393, opened 2021-12-30
Allow searching for multi-line Strings.

mbien:
The search dialog (`ctrl+shift+F or ctrl+shift+H `) didn't support searching for multi-line Strings since it used the wrong matcher for the task when a multi-line String was provided.

Updated the UI too. Its now resizable and the combo box editors are now multi-line capable.

Dialog looks identical on first open:
default

but when the window is resized vertically:
resized

### DIFF
--- a/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
+++ b/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
@@ -26,6 +26,7 @@ import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.ItemSelectable;
+import java.awt.KeyboardFocusManager;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.ItemEvent;
@@ -34,6 +35,7 @@ import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.swing.*;
@@ -213,7 +215,9 @@ final class BasicSearchForm extends JPanel implements ChangeListener,
     private void initComponents(final boolean searchAndReplace) {
 
         lblTextToFind = new JLabel();
-        cboxTextToFind = ComponentUtils.adjustComboForSearchPattern(new JComboBox<String>());
+        JComboBox<String> box = new JComboBox<>();
+        box.setEditor(new MultiLineComboBoxEditor());
+        cboxTextToFind = ComponentUtils.adjustComboForSearchPattern(box);
         lblTextToFind.setLabelFor(cboxTextToFind.getComponent());
         btnTestTextToFind = new JButton();
         lblTextToFindHint = new JLabel();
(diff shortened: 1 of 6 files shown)


matthiasblaesing:
I had a look how `canBeMultilinePattern` is used and I think it would be easier and more consistent to modify the regexp `MULTILINE_REGEXP_PATTERN` in `TextRegexpUtil`. The util method is used to determine if a single line in the file is enough for examination or if the whole file needs to be matched.

You can already search for multiline patterns, if you use regexp. Enter `\nTEST\n` into the search box and switch to match to "Regular Expression" and you will find only places where `TEST` is the only content in a line.

mbien:
> I had a look how `canBeMultilinePattern` is used and I think it would be easier and more consistent to modify the regexp `MULTILINE_REGEXP_PATTERN` in `TextRegexpUtil`. The util method is used to determine if a single line in the file is enough for examination or if the whole file needs to be matched.

yeah I looked at that regexp too. Its just that the method actually expects a regexp (see doc and also the actual pattern) and checks if the regexp would want to match multiline. If it would also check if the regexps itself would be multiline that would make it a weird frankenstein method - so i decided against it.

> 
> You can already search for multiline patterns, if you use regexp. Enter `\nTEST\n` into the search box and switch to match to "Regular Expression" and you will find only places where `TEST` is the only content in a line.

True, but it would have made things more complicated since i wanted to replace a string which contained regexp control characters. Turning on regexps would require escaping.

```java
public class ${name} {
    public static void main( String[] args )
    {
        System.out.println( "Hello World!" );
    }
}

```
thought might as well fix it instead of using the bash.

mbien:
the tests for canBeMultilinePattern:
https://github.com/apache/netbeans/blob/c084119009d2e0f736f225d706bc1827af283501/platform/api.search/test/unit/src/org/netbeans/modules/search/TextRegexpUtilTest.java#L304-L316

matthiasblaesing:
> > I had a look how `canBeMultilinePattern` is used and I think it would be easier and more consistent to modify the regexp `MULTILINE_REGEXP_PATTERN` in `TextRegexpUtil`. The util method is used to determine if a single line in the file is enough for examination or if the whole file needs to be matched.
> 
> yeah I looked at that regexp too. Its just that the method actually expects a regexp (see doc and also the actual pattern) and checks if the regexp would want to match multiline. If it would also check if the regexps itself would be multiline that would make it a weird frankenstein method - so i decided against it.

If you rename the method in your mind to `needsMultilineContext` it is in no way frankenstein, it does exactly what it should. A regexp containing a "real" newline character will also fail, so why not fix it there?

Without fixing it in this one method, you need to also fix `FastMatcher`. 

> [Tests]

The tests are not exhaustive, the cases:

`test\ntest`
`test\rtest`
`test\r\ntest`

are missing. You demonstrated, that they can be entered by copying and adding them causes test failures.

mbien:
ok I will rename the method to let it do both and add test cases. The ML matcher appears to be able to match actual multi line regexes so it can cover both use cases - I didn't expect it to be able to do that. Updating the method makes sense in that case.

matthiasblaesing (approved):
Looks sane to me. Thank you.

mbien:
@matthiasblaesing updated the UI too.

matthiasblaesing:
Looks sane to me. I have two minimal nitpick in the test window - this seems to be to small by default (see the cut of Cancel button):

image

mbien:
right, will fix that. I think this even happens without the changes.

mbien:
gave it 50px more. I am not sure why the buttons didn't resize the pane in the first place, but I also don't know much about the GroupLayout manager, beside that its used in form designer tools.

matthiasblaesing:
Thank you - looks good to me (literally).

PR #3394, opened 2021-12-30
Formatted the templates to be inline with the default java formatting.

mbien:
The recently added new maven based java project templates use a formatting which is different to NetBeans' default formatting settings. I think they should be consistent to whatever the default formatter is configured with.

(had to fix multi-line search and replace first before i could find the templates :) see https://github.com/apache/netbeans/pull/3393)


mbien:
@sdedic any reason I should not merge this? (e.g are there more new file/new project templates with non-default formatting I overlooked?)

sdedic:
Nope, sorry

sdedic (approved):

PR #3395, opened 2022-01-01
[jackpot] Add sourceVersion(int) to rule file utils and deprecate enum variant.

mbien:
- version enums cause problems if the enum value can not be resolved
 - shorter
 
With nb-javac 17 installed by default, having a version check API like https://github.com/apache/netbeans/pull/3274 became less of a priority. However the same issue still persists for rule files. Loading a rule file with a JDK 18 version check on nb-javac 17 would throw an exception and no hints are shown. Comparing versions without the enum would avoid this issue entirely.


mbien:
@jlahoda @JaroslavTulach what do you think about this?

jlahoda (approved):
Seems reasonable.

PR #3396, opened 2022-01-01
nb-javac checks cleanup and dialog removal.

mbien:
post nb-javac 17 on-by-default refactoring
- removed "get nb-javac" dialog and nb-javac checks
- removed sealed class workaround added in #3228
- trees cleanup after #3283 
  - removes ByteBuddy interceptors and reflection code
- some SourceVersion related cleanup

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
@@ -21,7 +21,6 @@ package org.netbeans.modules.java.source;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.security.ProtectionDomain;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.lang.model.SourceVersion;
@@ -35,33 +34,24 @@ import org.openide.modules.OnStart;
  */
 public class NoJavacHelper {
 
-    private static final AtomicReference<Boolean> hasWorkingJavac = new AtomicReference<>();
-    public static boolean hasWorkingJavac() {
-        Boolean res = hasWorkingJavac.get();
-        if (res != null) {
-            return res;
-        }
+    private static final boolean hasWorkingJavac;
+
+    static {
+        boolean res;
         try {
             SourceVersion.valueOf("RELEASE_17");
             res = true;
         } catch (IllegalArgumentException ex) {
-            //OK
             res = false;
         }
-        hasWorkingJavac.compareAndSet(null, res);
-        return hasWorkingJavac.get();
+        hasWorkingJavac = res;
     }
 
-    public static boolean hasNbJavac() {
-        try {
-            Class.forName("com.sun.tools.javac.comp.Repair");
-            return true;
-        } catch (ClassNotFoundException ex) {
-            //OK
-            return false;
-        }
+    public static boolean hasWorkingJavac() {
+        return hasWorkingJavac;
     }
 
+    // safety net if someone manages to start NB on JDK 8 with nb-javac uninstalled
     @OnStart
     public static class FixClasses implements Runnable {
 
@@ -106,7 +96,7 @@ public class NoJavacHelper {
                 Class targetClass = Class.forName(injectToClass);  //NOI18N
 
                 Method defineClass = unsafeClass.getDeclaredMethod("defineClass", String.class, byte[].class, int.class, int.class, ClassLoader.class, ProtectionDomain.class);  //NOI18N
-                defineClass.invoke(unsafe, fqn, classData, 0, classData.length, targetClass.getClassLoader(), targetClass.getProtectionDomain());  //NOI18N 
+                defineClass.invoke(unsafe, fqn, classData, 0, classData.length, targetClass.getClassLoader(), targetClass.getProtectionDomain());  //NOI18N
             } catch (Throwable t) {
                 //ignore...
                 Logger.getLogger(NoJavacHelper.class.getName()).log(Level.WARNING, null, t);
--- a/java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
@@ -20,7 +20,6 @@
 package org.netbeans.modules.java.source.indexing;
 
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -59,6 +58,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import javax.lang.model.SourceVersion;
 
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ModuleElement;
@@ -82,7 +82,6 @@ import org.netbeans.api.java.source.ClasspathInfo;
 import org.netbeans.api.java.source.ElementHandle;
 import org.netbeans.api.project.FileOwnerQuery;
 import org.netbeans.api.project.Project;
-import org.netbeans.modules.classfile.ClassFile;
 import org.netbeans.modules.java.source.ElementHandleAccessor;
 import org.netbeans.modules.java.source.JavaSourceTaskFactoryManager;
 import org.netbeans.modules.java.source.ModuleNames;
@@ -115,7 +114,6 @@ import org.netbeans.spi.java.classpath.support.ClassPathSupport;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
 import org.openide.filesystems.URLMapper;
-import org.openide.modules.InstalledFileLocator;
 import org.openide.util.Exceptions;
 //import org.openide.util.NbBundle;
 import org.openide.util.Pair;
@@ -1395,15 +1393,7 @@ public class JavaCustomIndexer extends CustomIndexer {
     private static Pair<Object,Method> heapDumper;
 
     private static String computeJavacVersion() {
-        if (NoJavacHelper.hasNbJavac()) {
-            File nbJavac = InstalledFileLocator.getDefault().locate("modules/ext/nb-javac-impl.jar", "org.netbeans.modules.nbjavac.impl", false);
-            if (nbJavac != null) {
-                return String.valueOf(nbJavac.lastModified());
-            }
-            return "-1";
-        } else {
-            return System.getProperty("java.vm.version", "unknown");
-        }
+        return SourceVersion.latest().toString();
     }
 
     private static class FilterOutJDK7AndLaterWarnings implements Comparable<Diagnostic<? extends JavaFileObject>> {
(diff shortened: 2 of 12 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
```diff
@@ -18,99 +18,34 @@
...
-                Class targetClass = Class.forName(injectToClass);  //NOI18N
-
-                Method defineClass = unsafeClass.getDeclaredMethod("defineClass", String.class, byte[].class, int.class, int.class, ClassLoader.class, ProtectionDomain.class);  //NOI18N
-                defineClass.invoke(unsafe, fqn, classData, 0, classData.length, targetClass.getClassLoader(), targetClass.getProtectionDomain());  //NOI18N 
-            } catch (Throwable t) {
-                //ignore...
-                Logger.getLogger(NoJavacHelper.class.getName()).log(Level.WARNING, null, t);
-            }
-        }
+    @Deprecated
```
jlahoda, line 46: I think I'd simply remove this method. This is not an API, we can remove whatever we want. (Maybe just increment the implementation version in `java.source.base/manifest.mf`, so that old and new modules won't mix.)

java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
```diff
@@ -18,99 +18,34 @@
...
-        } catch (ClassNotFoundException ex) {
-            //OK
-            return false;
-        }
+    public static boolean hasWorkingJavac() {
+        return hasWorkingJavac;
     }
 
-    @OnStart
-    public static class FixClasses implements Runnable {
```
jlahoda, line 66: I am a bit on a fence with the removal of the `FixClasses`. The sole purpose of this was to prevent wild linkage exceptions while running on JDK 8 without nb-javac. This is obviously a much smaller issue now that (nb-)javac is a default, but the exceptions are likely to make the IDE hard to use. So I can see an argument for both removing the code and keeping it.

java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
```diff
@@ -1395,15 +1395,7 @@ private static File dumpHeap(@NonNull final String path) {
...
-        if (NoJavacHelper.hasNbJavac()) {
-            File nbJavac = InstalledFileLocator.getDefault().locate("modules/ext/nb-javac-impl.jar", "org.netbeans.modules.nbjavac.impl", false);
-            if (nbJavac != null) {
-                return String.valueOf(nbJavac.lastModified());
-            }
-            return "-1";
-        } else {
-            return System.getProperty("java.vm.version", "unknown");
-        }
+        return System.getProperty("java.vm.version", "unknown");
```
jlahoda, line 1398: This would mean complete re-index when running on a different JDK, even with nb-javac. That is probably too much. I guess I'd suggest to either simply return a fixed value (e.g. `return "nb-javac";`) regardless what the underlying version is, or check the version of nb-javac if installed, e.g.:
```
        String version = System.getProperty("java.vm.version", "unknown");

        for (ModuleInfo mi : Lookup.getDefault().lookupAll(ModuleInfo.class)) {
            if ("org.netbeans.libs.nbjavacapi".equals(mi.getCodeNameBase())) {
                version = mi.getSpecificationVersion().toString();
            }
        }

        return version;
```

(assuming it is able to find the wrapper module.)

java/java.source.base/test/unit/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorkerTest.java
```diff
@@ -1958,13 +1958,13 @@ protected void tearDown() throws Exception {
     }
 
     public static Test suite() {
-        if (NoJavacHelper.hasNbJavac()) {
-            return new VanillaCompileWorkerTest("noop");
-        } else {
+//        if (NoJavacHelper.hasNbJavac()) {
```
jlahoda, line 1961: Nit: simply delete the non-nb-javac lines.

jlahoda (approved):
Great, this will be helpful! I've left a few comments inline.

mbien:
@jlahoda thats too early for a review - i hacked this quickly together to get rid of the startup dialog :) But thanks I will follow your suggestions of course. More commits will follow which remove workarounds I added in past PRs for example. (travis is inactive?)

java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
mbien: roger, lets keep it.

java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
mbien: will do.

java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
mbien: could we use SourceVersion.latest() there? This would make it nb-javac independent.

mbien: nevermind, i believe your suggestion is better. However using the "java.version" property as fallback would make it compatible with the module spec version. It uses the same "17.0.1" format.

java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
```diff
@@ -1395,15 +1393,12 @@ private static File dumpHeap(@NonNull final String path) {
...
+        for (ModuleInfo mi : Lookup.getDefault().lookupAll(ModuleInfo.class)) {
+            if ("org.netbeans.libs.nbjavacapi".equals(mi.getCodeNameBase())) {
+                return mi.getSpecificationVersion().toString();
             }
-            return "-1";
-        } else {
-            return System.getProperty("java.vm.version", "unknown");
         }
+        return System.getProperty("java.version", "unknown");
     }
```
mbien, line 1402: please note that the nb-javac module spec version is `feature.interim.update` (17.0.1), which is the same as the `java.version` property.

This would have the effect of triggering a re-index if a nb-javac update is delivered via the plugin portal.

a potential alternative would be simply returning `SourceVersion.latest()` which is feature version only (would also handle nb-javac and javac the same).

mbien:
@jlahoda incorporated your review remarks into the first commit + added 3 more commits.

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1452,6 +1452,7 @@ protected JCExpression checkExprStat(JCExpression t) {
             return super.checkExprStat(t);
         }
 
+/*  TODO: unused? called a method which is only available from 8-11.
         protected JCCase switchBlockStatementGroupImpl() throws Throwable {
```
mbien, line 1456: i commented it out, but I think we can remove this, right?

mbien:
travis jobs 13 and 16 are failing.

JaroslavTulach (approved):

java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
mbien: the module lookup code caused module loading failures in the tests. I am going to return `SourceVersion` temporarily and see if this alone is enough to fix travis (it works locally).

mbien: tests are green without the module lookup.

java/java.source.base/src/org/netbeans/modules/java/source/indexing/JavaCustomIndexer.java
```diff
@@ -1395,15 +1394,12 @@ private static File dumpHeap(@NonNull final String path) {
...
-            return "-1";
-        } else {
-            return System.getProperty("java.vm.version", "unknown");
-        }
+//        for (ModuleInfo mi : Lookup.getDefault().lookupAll(ModuleInfo.class)) {
+//            if ("org.netbeans.libs.nbjavacapi".equals(mi.getCodeNameBase())) {
+//                return mi.getSpecificationVersion().toString();
+//            }
+//        }
+        return SourceVersion.latest().toString();
```
jlahoda, line 1402: I like this, good idea. The commented out code could be simply removed.

java/java.source.base/test/unit/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorkerTest.java
```diff
@@ -1958,13 +1958,7 @@ protected void tearDown() throws Exception {
...
 
     public static Test suite() {
-        if (NoJavacHelper.hasNbJavac()) {
-            return new VanillaCompileWorkerTest("noop");
-        } else {
-//            return new VanillaCompileWorkerTest("testAnonymousClasses");
-//            return new VanillaCompileWorkerTest("testPreserveValidInitializers");
-            return new NbTestSuite(VanillaCompileWorkerTest.class);
-        }
+        return new VanillaCompileWorkerTest("noop");
```
jlahoda, line 1961: Sorry, but the other branch should be preserved:
```
return new NbTestSuite(VanillaCompileWorkerTest.class);
```

Or simply remove the `suite` method, as this should be the default.

The intent here was to not run the test with the old nb-javac, but the new one is basiclly a javac from the JDK, so the test should run.

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
jlahoda: Yes, I think so.

jlahoda (approved):

java/java.source.base/test/unit/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorkerTest.java
mbien: done (it confused me a bit since both javac versions are now similar, so there would be plausible reasons to run vanilla tests but also nb-javac tests :))

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
mbien: All tests green with the requested changes integrated. 

Lets try to remove the ByteBuddy interceptor which delegates to one of the two methods.

mbien: interceptor removal is squashed with the trees commit 748afc8

mbien:
everything green now. Had to brute force through job 13 though

```
org.netbeans.modules.java.lsp.server.protocol.ServerTest
    failed: testCancelProgressHandle

restart job
org.netbeans.modules.gradle.java.classpath.ClassPathProviderImplTest
    failed: testCompilePreTrusted

restart job
org.netbeans.modules.java.lsp.server.protocol.ServerTest
    failed: testCancelProgressHandle

restart job
green
```

mbien:
@jlahoda @JaroslavTulach @matthiasblaesing ok to merge?

JaroslavTulach:
> everything green now. Had to brute force through job 13 though
> 
> ```
> org.netbeans.modules.java.lsp.server.protocol.ServerTest
>     failed: testCancelProgressHandle
> ```

Sorry for that. Feel free to mark the test as `@RandomlyFails`. The maintainers of the `ServerTest` shall suffer, not people working on the rest of NetBeans!

jlahoda (approved):
Looks good to me.

PR #3397, opened 2022-01-03
Follow LSP specification for null value for startCharacter in FoldingRange

matthiasblaesing:
The LSP specification specifies for the startCharacter:

The zero-based character offset from where the folded range starts. If
not defined, defaults to the length of the start line.

In contrast to the specification the first character was choosen as the
start of the fold, hiding for example the function name.


matthiasblaesing:
This is a followup to #3385. It was already tested by the author of that PR and will be merged, once unittests come back green.

PR #3398, opened 2022-01-03
LSP: Never remove imports on save when the code isn't bug free.

dbalek:
Never remove imports on save when the code isn't bug free.


JaroslavTulach (approved):
I'd like to see a test, of course, but the change is so simple, it must be correct.

sdedic (approved):

PR #3399, opened 2022-01-03
[cnd] small-1.1: Moving some cnd modules to contrib

vieiro:
Moving some modules from "cnd" into "contrib".
- The build will still fail as some module dependencies have to be modified (that will be done in future commits).
- The "cnd" cluster has been modified in "cluster.properties".
- The "contrib" cluster has NOT been modified in "cluster.properties".

This is the first committ in the cnd/small-1 stage of the "cnd-small" proposal https://cwiki.apache.org/confluence/display/NETBEANS/cnd+integration+proposal%3A+cnd-small


matthiasblaesing (approved):
Looks sane to me (only renames and removals from the CND cluster config).

PR #3400, opened 2022-01-03
Be able to resolve project problems in VSCode extension.

entlicher:
When a project problem is detected, there is no way to resolve it from within VSCode extension.
This PR introduces `java.project.resolveProjectProblems` command, which attempts to resolve the project problems and when they are not resolved, it displays the resolution error.
`JavaErrorProvider` displays a dull `error(0): null` message, which we change to `error(1): <line:column range>`
The gradle project needs to be adapted to provide the resolution error message.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
@@ -492,7 +492,7 @@ public final class WorkspaceServiceImpl implements WorkspaceService, LanguageCli
                 return (CompletableFuture<Object>) (CompletableFuture<?>) joinedFuture;
             }
             case Server.JAVA_PROJECT_RESOLVE_PROJECT_PROBLEMS: {
-                CompletableFuture<Object> result = new CompletableFuture<>();
+                final CompletableFuture<Object> result = new CompletableFuture<>();
                 List<Object> arguments = params.getArguments();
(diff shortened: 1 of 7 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
```diff
@@ -485,6 +491,61 @@
...
+                    Project project = FileOwnerQuery.getOwner(file);
+                    if (project != null) {
+                        ProjectProblemsProvider ppp = project.getLookup().lookup(ProjectProblemsProvider.class);
+                        if (ppp != null) {
+                            Collection<? extends ProjectProblemsProvider.ProjectProblem> problems = ppp.getProblems();
+                            if (!problems.isEmpty()) {
+                                List<Pair<ProjectProblemsProvider.ProjectProblem, Future<ProjectProblemsProvider.Result>>> resolvers = new LinkedList<>();
+                                for (ProjectProblemsProvider.ProjectProblem problem : ppp.getProblems()) {
+                                    if (problem.isResolvable()) {
+                                        resolvers.add(Pair.of(problem, problem.resolve()));
```
sdedic, line 515: is the `.resolve()` called in the LSP message thread ? If so, should be called in a RequestProcessor, since it may eventually trigger a project build.

entlicher: Done.

sdedic (approved):

PR #3401, opened 2022-01-03
Add binaries.repos property to downloadbinaries call in projectized.xml.

neilcsmith-net:
Add binaries.repos property to downloadbinaries call in projectized.xml. This was accidentally left out in a #2710 Amongst other things, this helps with testing libraries only in local .m2 cache.

Tiny fix cherry picked from something else I was working on which I'm rewriting.


matthiasblaesing (approved):
Looks sensible to me.

neilcsmith-net:
Thanks @matthiasblaesing

PR #3403, opened 2022-01-03
print simple junit report on travis job failure.

mbien:
This should hopefully print a list of failed tests if a job fails and make reading the travis log about 20% less painful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ addons:
       - xvfb
       - openjfx
       - libgfortran3
+      - libxml2-utils
 
 install:
   - export PATH="$PATH:$TRAVIS_BUILD_DIR/nbbuild/travis"
@@ -34,6 +35,10 @@ install:
 cache:
   directories:
   - $HOME/.hgexternalcache
+
+git:
+  quiet: true
+  depth: 3
 
 matrix:
     include:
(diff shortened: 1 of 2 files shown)


mbien:
oh no. only commit-validation outputs TEST-*.xml files?

matthiasblaesing:
I think you are looking at the wrong location. The reports should be in <base>/<cluster>/<module>/build/test/unit/results

sdedic:
Excellent idea !
Would it be also possible to output stdout/err in the failure case ?

mbien:
> Excellent idea ! Would it be also possible to output stdout/err in the failure case ?

sure, logs would be in the xml too. Its just that the more is printed the less useful it becomes IMO. I often have a hard time actually figuring out what failed just by looking at the travis log. This little report would list the failed testcases so that I can run them locally without even having to go through the log.

But the optimal situation would be to upload it somewhere + present a report using a plugin. I had that set up for github actions once (in my fork).

mbien:
```
junit report / failed tests:

org.openide.awt.ContextActionNonAWTTest

    failed: testContextDelegate

end of report
```
it works :)

sdedic:
/me like it. Do you want to enhance it further or merge :) ?

mbien:
lets merge it. We can improve it later if there is demand.

PR #3404, opened 2022-01-03
Fix cluster splitting of RAT errors

matthiasblaesing:
Since the module directories were moved from the toplevel into
per-cluster folders, alls RAT violations were reported in the
"not cluster" cluster.


vieiro (approved):
Works as expected. See https://github.com/vieiro/netbeans/tree/fix_rat_report_test for an example: dlight cluster licensing errors are now properly reported in "dlight.xml" rat report file.

matthiasblaesing:
Thank you for the check @vieiro!

PR #3405, opened 2022-01-04
Adding command for JPA Entity Classes from DB

jhorvath:
Adding command for creating JPA Entity Classes from DB to java.lsp.server

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.java.lsp.server.db;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.CodeActionParams;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.MessageType;
+import org.netbeans.api.db.explorer.ConnectionManager;
+import org.netbeans.api.db.explorer.DatabaseConnection;
+import org.netbeans.api.db.explorer.DatabaseException;
+import org.netbeans.api.java.project.JavaProjectConstants;
+import org.netbeans.api.progress.aggregate.BasicAggregateProgressFactory;
+import org.netbeans.api.progress.aggregate.ProgressContributor;
+import org.netbeans.api.project.Project;
+import org.netbeans.api.project.ProjectUtils;
+import org.netbeans.api.project.SourceGroup;
+
+import org.netbeans.modules.j2ee.persistence.api.entity.generator.EntitiesFromDBGenerator;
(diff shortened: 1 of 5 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+@NbBundle.Messages({
+    "MSG_NoProject=No Project Open",
+    "MSG_NoDbConn=No DB Connection",
+    "MSG_NoSourceRoot=No source root found",
+    "MSG_SelectTables=Select Database Tables",
+    "MSG_EnterPackageName=Enter package name"
+})
+@ServiceProvider(service = CodeActionsProvider.class)
+public class DBEntityFromTables extends CodeActionsProvider {
```
dbalek, line 69: I do not like reusing `CodeActionsProvider` for registering simple Command providers. Consider creating a new `CommandProvider` interface/abstract class for such purpose. The original `CodeActionsProvider` should extend/implement it.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final String COMMAND_ENTITY_FROM_TABLES = "db.entity.from.tables"; //NOI18N
+
+    @Override
+    public CompletableFuture<Object> processCommand(NbCodeLanguageClient client, String command, List<Object> arguments) {
+        if (!COMMAND_ENTITY_FROM_TABLES.equals(command)) {
+            return null;
+        }
+        Lookup.getDefault().lookup(LspServerState.class).openedProjects().thenAccept((projects) -> {
+            if (projects.length > 0) {
+                createEntityClassesInProject(client, projects[0]);
```
sdedic, line 80: Question: if the command is invoked with an active editor focused: shouldn't we create the classes in the editor's project instead in the 1st opened one ? The order of projects in `openedProjects` may be rather random.
Of course provided that the LSP command arguments contain some URI etc to begin with.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
```diff
@@ -0,0 +1,139 @@
...
+                                    try {
+                                        generator.generate(pc);
+                                    } catch (SQLException | IOException ex) {
+                                        client.showMessage(new MessageParams(MessageType.Error, ex.getMessage()));
+                                    }
+                                });
+                        return;
+                    });
+        } catch (SQLException | IllegalArgumentException | DatabaseException ex) {
+            System.out.println(ex.getMessage());
```
sdedic, line 124: make this ` client.showMessage` instead, or use `NotifyDescriptor` + `DialogDisplayer.getDefault().notify()` that will be automagically remoted to the client.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
```diff
@@ -0,0 +1,139 @@
...
+                return null;
+            }
+            ConnectionManager.getDefault().connect(connection);
+            SourceGroup[] sr = ProjectUtils.getSources(prj).getSourceGroups(JavaProjectConstants.SOURCES_TYPE_JAVA);
+            if (sr.length < 1) {
+                client.showMessage(new MessageParams(MessageType.Error, Bundle.MSG_NoSourceRoot()));
+                return null;
+            }
+            Connection conn = connection.getJDBCConnection();
+            ResultSet rs = conn.getMetaData().getTables(conn.getCatalog(), conn.getSchema(), "%", new String[]{"TABLE", "VIEW"}); //NOI18N
```
sdedic, line 103: This call eventually waits for the data from the JDBC connection; I'd suggest to run the DB query in a RequesProcessor so the LSP communication is not blocked. The behaviour depends on the DB location / network conditions.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBEntityFromTables.java
```diff
@@ -0,0 +1,139 @@
...
+            }
+            return client.showQuickPick(new ShowQuickPickParams(Bundle.MSG_SelectTables(), true, dbItems))
+                    .thenApply(items -> items.stream().map(item -> item.getLabel()).collect(Collectors.toList()))
+                    .thenAccept(tables -> {
+                        client.showInputBox(new ShowInputBoxParams(Bundle.MSG_EnterPackageName(), "")) //NOI18N
+                                .thenAccept(packageName -> {
+                                    EntitiesFromDBGenerator generator = new EntitiesFromDBGenerator(tables, true, packageName, sr[0], connection, prj, null);
+                                    ProgressContributor pc = BasicAggregateProgressFactory.createProgressContributor("entity"); //NOI18N
+                                    try {
+                                        generator.generate(pc);
```
sdedic, line 116: Also potentially run inside LSP event loop (input box response is sent as a protocol response to the server). Please check and if so, run asynchronously. The `ProgressContributor` indicates that the code called runs for some time and reports progress - that won't work when LSP connection is blocked.

sdedic (changes requested):

PR #3408, opened 2022-01-04
[NETBEANS-6361] Code completion for class fields doesn't work.

ppisl:
This fixing resolving types of fields.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/inference/TypeInferenceVisitor.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/inference/TypeInferenceVisitor.java
@@ -99,6 +99,8 @@ public class TypeInferenceVisitor extends TypeVisitor {
         if (sameVariableName(leaf, node)) {
             if (node.hasInitialExpression()){
                 guessedType = deriveExpressonType(node.getInitialExpression());
+            } else if (node.getType() != null) {
+                    guessedType = node.getType();
             }
         }
     }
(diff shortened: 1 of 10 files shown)


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/inference/TypeInferenceVisitor.java
```diff
@@ -99,6 +99,8 @@ public void visitField(FieldNode node) {
         if (sameVariableName(leaf, node)) {
             if (node.hasInitialExpression()){
                 guessedType = deriveExpressonType(node.getInitialExpression());
+            } else if (node.getType() != null) {
+                    guessedType = node.getType();
```
sdedic, line 103: Q: would `GroovyUtils.findInferredType()` in addition to getType() bring some benefits here ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/occurrences/TypeVisitor.java
```diff
@@ -139,6 +139,11 @@ public void collect() {
                         visitClosureListExpression((ClosureListExpression) scope);
                         return;
                     }
+                } else if (scope instanceof ClassNode) {
+                    ClassNode classNode = (ClassNode) scope;
+                    for (FieldNode fieldNode: classNode.getFields()) {
+                        visitField(fieldNode);
```
sdedic, line 145: just fields ?

sdedic (approved):

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/inference/TypeInferenceVisitor.java
ppisl: I don't think so. In the time the metadata are null in this case.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/occurrences/TypeVisitor.java
ppisl: For this moment yes.

PR #3410, opened 2022-01-05
disabled uihandler modules (exception+statistics reporter).

mbien:
disabling the module does two things:
 - removes the "Help us improve NetBeans" dialog which appears on second start (and counterpart in general options)
 - and the exception reporter functionality (window -> ide tools -> exception reporter)

non of this worked anymore anyway as far as I can tell.

todo: travis might fail if it has a test config for it, we will see.

### DIFF
--- a/nb/uihandler.exceptionreporter/nbproject/project.properties
+++ b/nb/uihandler.exceptionreporter/nbproject/project.properties
@@ -15,3 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
+javac.source=1.8
+is.autoload=true
--- a/ide/utilities/src/org/netbeans/modules/utilities/Bundle.properties
+++ b/ide/utilities/src/org/netbeans/modules/utilities/Bundle.properties
@@ -26,3 +26,5 @@ which helps you open a file in a previously unmounted filesystem.
 
 Loaders/application/pdf/Factories/org-netbeans-modules-pdf-PDFDataLoader.instance=PDF Documents
 Loaders/text/url/Factories/org-netbeans-modules-url-URLDataLoader.instance=URL Files
+
+CTL_ReportIssueAction=&Report Issue
--- a/ide/utilities/src/org/netbeans/modules/utilities/ReportNBIssueAction.java
+++ b/ide/utilities/src/org/netbeans/modules/utilities/ReportNBIssueAction.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.netbeans.modules.bugzilla.exceptionreporter;
+package org.netbeans.modules.utilities;
 
 import org.openide.util.actions.SystemAction;
 import org.openide.util.HelpCtx;
@@ -35,7 +35,7 @@ import org.openide.util.NbBundle;
  * 
  * @author Tomas Stupka
  */
-@ActionID(id = "org.netbeans.modules.bugzilla.exceptionreporter.ReportNBIssueAction", category = "Help")
+@ActionID(id = "org.netbeans.modules.utilities.ReportNBIssueAction", category = "Help")
 @ActionRegistration(lazy = false, displayName = "#CTL_ReportIssueAction")
 @ActionReference(path = "Menu/Help", position = 450)
 public class ReportNBIssueAction extends SystemAction {
--- a/nb/bugzilla.exceptionreporter/nbproject/project.properties
+++ b/nb/bugzilla.exceptionreporter/nbproject/project.properties
@@ -14,6 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-is.eager=true
-javac.source=1.6
+javac.source=1.8
+is.autoload=true
 javac.compilerargs=-Xlint -Xlint:-serial
--- a/nb/bugzilla.exceptionreporter/src/org/netbeans/modules/bugzilla/exceptionreporter/Bundle.properties
+++ b/nb/bugzilla.exceptionreporter/src/org/netbeans/modules/bugzilla/exceptionreporter/Bundle.properties
@@ -19,5 +19,3 @@ OpenIDE-Module-Display-Category=Team
 OpenIDE-Module-Name=Bugzilla-Exception Reporter Bridge
 
 OpenIDE-Module-Short-Description=Providess exception reporter access to the netbeans.org bugzilla
-
-CTL_ReportIssueAction=&Report Issue
--- a/nb/ide.branding.kit/nbproject/project.xml
+++ b/nb/ide.branding.kit/nbproject/project.xml
@@ -51,12 +51,6 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
-                    <code-name-base>org.netbeans.modules.uihandler.exceptionreporter</code-name-base>
-                    <run-dependency>
-                        <specification-version>1.0</specification-version>
-                    </run-dependency>
-                </dependency>
-                <dependency>
                     <code-name-base>org.netbeans.modules.welcome</code-name-base>
                     <run-dependency>
                         <release-version>1</release-version>
--- a/platform/lib.uihandler/nbproject/project.properties
+++ b/platform/lib.uihandler/nbproject/project.properties
@@ -14,9 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-is.autoload=true
 javac.compilerargs=-Xlint:unchecked
-javac.source=1.6
+javac.source=1.8
+is.autoload=true
 javadoc.arch=${basedir}/arch.xml
 javadoc.apichanges=${basedir}/apichanges.xml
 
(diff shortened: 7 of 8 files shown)


nbbuild/cluster.properties
```diff
@@ -299,8 +298,7 @@ nb.cluster.platform=\
         spi.actions,\
         spi.quicksearch,\
         templates,\
-        templatesui,\
-        uihandler
```
JaroslavTulach, line 303: Both [lib.uihandler](https://bits.netbeans.org/12.6/javadoc/org-netbeans-lib-uihandler/org/netbeans/lib/uihandler/package-summary.html) as well as [uihandler](https://bits.netbeans.org/12.6/javadoc/org-netbeans-modules-uihandler/overview-summary.html) are an API of 12.6.

We should keep them around as autoload modules, in case some plugin or NetBeans application continues to use them. Disabled autoloads have close to no overhead on the runtime performance of the system and given NetBeans project commitment to compatibility (or is it mine?) I'd rather kept them around.

JaroslavTulach (changes requested):
Thanks for cleaning up the UI. The dialog is currently useless and needlessly scary. On the other hand, I'd prefer to keep the library modules in platform cluster around as our commitment for 3rd parties using it.

Off-topic: there was some discussion about using the UI Handler infrastructure to gain information about performance and UI responsiveness of VSNetBeans among my peers. Should that ever conclude, we'd get the serverside running again and start to use these libraries again.

neilcsmith-net:
Thanks @mbien for taking a look at this.  Hadn't got around to following up yet, although would like to get done for 13.  Agree with @JaroslavTulach on the need to keep in on an API perspective.  Think I saw a way to disable the dialog, but will check again.

OT: VSNetBeans using would require ASF server side or a VSNetBeans release that isn't from ASF I guess.  A discussion for dev@ anyway.

nbbuild/cluster.properties
mbien: i had that thought too.

If I revert the changes in the property files, set the modules to `autoload=false` in manifest but keep the dependency change in ide.branding.kit/nbproject/project.xml. I think we can be backwards compatible while having stats and exception reporting not loaded by default. Would that be ok?

mbien: update: it still loads without autoload

is there a way to debug which module is responsible for loading a particular other module?

mbien:
> Thanks @mbien for taking a look at this. Hadn't got around to following up yet, although would like to get done for 13. Agree with @JaroslavTulach on the need to keep in on an API perspective. Think I saw a way to disable the dialog, but will check again.

https://github.com/apache/netbeans/blob/c084119009d2e0f736f225d706bc1827af283501/platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java#L247-L251

that would be those lines (I tried doing that first actually). First is for the dialog. last line is for the options UI. Doesn't solve the exception reporter window though.

Not loading the module at all would be cleaner IMO.

neilcsmith-net:
@mbien actually think I was possibly looking at branding to preset these settings -

https://github.com/apache/netbeans/blob/master/platform/uihandler/src/org/netbeans/modules/uihandler/Installer.java#L338-L339

Agree that not loading would be good.  Actually possibly both, so that it still doesn't show if anything does trigger the load?

nbbuild/cluster.properties
JaroslavTulach: I usually put conditional breakpoint into `org.netbeans.Module.setEnabled(...)` to see who's enabling the module.

mbien: i can't explain why its getting loaded on startup.
```java
    void setEnabled(boolean enabled) {
        /* #13647: actually can happen if loading of bootstrap modules is rolled back: */
        if (isFixed() && ! enabled) throw new IllegalStateException("Cannot disable a fixed module: " + this); // NOI18N
        if (!this.isEnabled()) {
            Dependency[] deps = this.data.getDependencies();
            if(deps != null) {
                for (Dependency dependency : deps) {
                    if (dependency.getName().contains("uihandler") || dependency.getName().contains("exceptionreporter")) {
                        System.out.println(this.data.getCodeName()+" (autoload: "+isAutoload()+", eager: "+isEager()+")");
                        System.out.println("depends on: "+dependency);
                        System.out.println();
                    }
                }
            }
        }
        this.enabled = enabled;
    }
```
results in:
```
org.netbeans.modules.uihandler (autoload: false, eager: false)
depends on: module org.netbeans.lib.uihandler > 1.37

org.netbeans.modules.uihandler.exceptionreporter (autoload: false, eager: false)
depends on: module org.netbeans.modules.uihandler > 2.0

org.netbeans.modules.bugzilla.exceptionreporter (autoload: false, eager: true)
depends on: module org.netbeans.lib.uihandler > 1.37
```
nothing depends on the uihandler modules which is also autoloading.

JaroslavTulach: `eager: true` - means to enable when possible - e.g. when all dependencies are satisfied. Is the `org.netbeans.modules.bugzilla.exceptionreporter` enabled in runtime? Than it enables also its autoload dependencies. Probably make the module `!eager && autoload`.

neilcsmith-net: That module (bugzilla.exceptionreporter) also provides the action in the Help menu to open the issues reporting page on the website I think. Can we remove the dependency instead?

mbien: question: I see xmls like:
```xml
<module name="org.netbeans.modules.uihandler">
    <param name="autoload">false</param>
    <param name="eager">false</param>
    <param name="enabled">true</param>      <------- 
    <param name="jar">modules/org-netbeans-modules-uihandler.jar</param>
    <param name="reloadable">false</param>
</module>
```
(netbeans/platform/config/Modules)
what in the build is influencing if its "enabled" or not? Some don't have that param. I suppose that is what you mean by "enabled by the runtime"?

@neilcsmith-net that action could be potentially moved into a different module

mbien: answer:
`is.enabled=false`

that was easy :)

neilcsmith-net: @mbien absolutely, although not sure that module does much else now. Either way, please don't lose the action within the context of these changes.

mbien: modules are now disabled and set to autoload + report issue action moved to "user utilities" (ide/utilities). This is the same module which is also providing the "docs and support" link and the kb shortcuts in the help menu.

mbien: @neilcsmith-net we could probably remove the `org.netbeans.modules.bugzilla.exceptionreporter` module and `NBBugzillaUtils` in the bugzilla module. Unless you are planning to migrate back to bugzilla in about a year :)

But even the Kenai integration is still there... probably better to leave that to other PRs.

mbien:
> Agree that not loading would be good. Actually possibly both, so that it still doesn't show if anything does trigger the load?

Sure why not, but i have to say now that it is disabled its very low on my "things to improve" list :)

nb/uihandler.exceptionreporter/nbproject/project.properties
```diff
@@ -15,3 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
+javac.source=1.8
+is.autoload=true
+is.enabled=false
```
JaroslavTulach, line 20: Does it make sense to specify `is.enabled` status for `is.autoload` module? Removing line with `is.enabled=...` shall have no effect, I assume.

JaroslavTulach (approved):
Leaving the modules around is better for compatibility of the platform. Thanks.

nb/uihandler.exceptionreporter/nbproject/project.properties
mbien: i actually completely misunderstood what autoload does :) If you are saying that setting autoload=true and not setting anything else will keep the modules off, i can give it another try tomorrow.

JaroslavTulach: The old explanation by @jglick is at:
- https://bits.netbeans.org/11.1/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html
- just search for "autoload"

Surprisingly it is not included in the latest Javadoc:
- https://bits.netbeans.org/12.6/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html
- I guess our Javadoc system broke somehow...

> setting autoload=true and not setting anything else will keep the modules off

Check `nbbuild/netbeans` and `platform/config/Modules/*.xml` with/without `is.enabled` setting. Possibly the `*xml` files stay unchanged. Then they would (obviously) behave the same in the runtime.

jglick: https://bits.netbeans.org/11.1/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html#enablement specifically

mbien: @jglick do friend dependencies cause a module to load? trying to disable some findbugs modules atm.

jglick: Yes, IIRC (and it has been a couple decades!) the only thing special about a “friend” is which packages it can access via class loader. No relation to enablement logic.

PR #3411, opened 2022-01-05
Remove unnecessary log entry from javadoc build

jhorvath:
There is too many occurrences of `[arch] Resolved to embedded DTD` in javadoc build log. Information value of this line is low, so removing to make log more concise and readable.


sdedic (approved):

sdedic:
I'd ignore the failed Win build -- infrastructure timeout.

PR #3414, opened 2022-01-05
Support for image IDs, LSP client can redefine images/icons

sdedic:
This PR allows LSP clients to "rebrand" or alter icons supplied byt he NBLS server for exported nodes. It's not exactly branding, since the LSP client may have its own ideas or icon schemes that do not fit to the NB Node implementations - which shall not be affected at all. The vscode client, in addition, uses fontified icons (bcs they scale), which more or less prevent branding - except maybe if I'd extract the font glyphs (possible license issues) and converted them to SVG branding resources.

Instead, the URL of the image's resource will be extracted from Node's icon and transmitted. As it acts like an unique ID (not exactly, see notes below), the client may replace it with its own icon. Icons that are the same on NB side (NB design) may be matched and replaced based on `contextValue` or `name` exported node's metadata (which is actually more powerful than just branding). Branding can be still used, if appropriate.

Note that certain prominent icons **do not come from URL resources**, but from `UIManager.getDefaults()` - the exact shape is defined by LaF. So these icons either do not have URLs at all, or their URLs differ between LaFs, which is unfortunate if the LaF-specific URL was used an ID.

The 1st commit in this PR changes openide `ImageUtilities`, it's probably the most important one from this PR.
* Images and icons produced by `ImageUtilities` maintain not only `url` (see `Image.getProperty` API) but also `uri`. As `url` property must be typed as URL, and non-location IDs cannot be represented as URLs (no URLStreamHandler registered, and the underlying resource bits are really not known) - another property is invented. Since the URI syntax is more flexible, even badges or filters can be represented as URIs (see at the end) in the future.

* During implementation it turned out that if `ToolkitImage` is wrapped in a delegating Image instance weird things happen: the SurfaceManager is not initialized properly so an attempt to `graphics.drawImage` will fail with `IllegalArgumentException`. So super-hacky `originalImage` or `originalIcon` backdoor is invented instead - so the anticipated delegating implementation may pass the surface-compatible real image as appropriate.

This part of impl is not finished: IMHO some convenience accessor methods for the URL (already was there) and URI should be added to `ImageUtilities`, the URI of regular resource-based icons should **not** be affected by branding or localization (it is now) and eventual badges should be somehow accessible as `URI[]` and also the possibly applied filter od disabled icon filter...
That's why I did not (yet) documented the added properties in `arch.xml`. I plan to do so after the implementation proves worthy hopefuly before NB13 finalizes.

The [UIDefaultsIconMetadata](https://github.com/apache/netbeans/pull/3414/files#diff-f39410866a776e64c3daf59786391b7fccf8d07adab3e9b89f61b88a9cd8a38eR50) is a NBLS startup hack, that replaces UIDefault images for delegates that provide the metadata (URL, URI). These are then preserved during `ImageUtilities.mergeImages` so a combined image inherits the base ID from the base image. 

Finally the URI is transmitted to the LSP client in a `ImageDescriptor` structure that will grow in the future (badge and filter URIs).

The vscode client reads icon replacement definitions from `package.json` of  all vscode extensions. It matches the transmitted image based on base URI and possibly `contextValue` items and also `name` as it is sometimes used as programmatic ID (in nodes that represent containers). If a match is found, icon data is ignored and vscode's `ThemeIcon` or a local icon (provided by an extension) is used.
This way the vscode can alter icons to match its visual style and ensure consistency if an icon was defined already in the vscode world. Other LSP clients can do just the same.

Note that the tree LSP messages need some optimization. For example, the server should only transmit image URI/descriptor - and the client should call back for URL's contents (image data) if it wants to display the original image. This will be solved in a future PRs.

### DIFF
--- a/java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
+++ b/java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
@@ -1,0 +1,359 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.nbcode.integration;
+
+import java.awt.Component;
+import java.awt.EventQueue;
+import java.awt.Graphics;
+import java.awt.HeadlessException;
+import java.awt.Image;
+import java.awt.MediaTracker;
+import java.awt.Toolkit;
+import java.awt.Transparency;
+import java.awt.image.BufferedImage;
+import java.awt.image.ColorModel;
+import java.awt.image.ImageObserver;
+import java.awt.image.WritableRaster;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Enumeration;
+import java.util.Hashtable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.swing.Icon;
+import javax.swing.ImageIcon;
+import javax.swing.UIDefaults;
+import javax.swing.UIDefaults.ActiveValue;
+import javax.swing.UIDefaults.LazyValue;
+import javax.swing.UIManager;
+import org.openide.modules.OnStart;
+import org.openide.util.ImageUtilities;
+
+/**
+ * Goes through UIDefaults and replaces Icons and Images with variants that
+ * return image's URL from the 'url' property.
+ * @author sdedic
+ */
+@OnStart
+public class UIDefaultsIconMetadata implements Runnable {
+    private static final Logger LOG = Logger.getLogger(UIDefaultsIconMetadata.class.getName());
+    
+    // for diagnostic purposes only
+    public static final String PROP_ORIGINAL_ICON = "originalIcon"; // NOI18N
+    // for diagnostic purposes only
+    public static final String PROP_ORIGINAL_IMAGE = "originalImage"; // NOI18N
+    
+    private static final String URN_DEFAULTS_PREFIX = "uidefaults:"; // NOI18N
+    
+    private static Object wrapImage(String key, Image image, Icon icon, Object... meta) {
+        Object o = image.getProperty(ImageUtilities.PROPERTY_ID, null); // NOI18N
+        if (o instanceof URL) {
+            return image;
+        }
+        
+        o = image.getProperty(ImageUtilities.PROPERTY_ID, null);
+        if (o instanceof String || o instanceof URI) {
+            return image;
+        } else {
+            
+            Hashtable props = new Hashtable();
+            for (int i = 0; i < meta.length; i+= 2) {
+                props.put(meta[i], meta[i+1]);
+            }
+            props.put(ImageUtilities.PROPERTY_ID, URN_DEFAULTS_PREFIX + key);
+            return createNew(image, icon, props);
+        }
+    }
+    
+    @Override
+    public void run() {
+        EventQueue.invokeLater(() -> {
+            // force LaF initialization before the UIDefaults are read/patched.
+            UIManager.get("force.laf.initialization");
+            replaceIconsAndImages(UIManager.getDefaults());
+        });
+    }
+    
+    static void replaceIconsAndImages(UIDefaults defs) {
+        refreshUIDefaults(defs);
+        defs.addPropertyChangeListener(new PropertyChangeListener() {
+            @Override
+            public void propertyChange(PropertyChangeEvent evt) {
+                if (null == evt.getPropertyName() || "UIDefaults".equals(evt.getPropertyName())) { // NOI18N
+                    LOG.log(Level.INFO, "Refreshing all UIDefaults");
+                    refreshUIDefaults(defs);
+                } else {
+                    String k = evt.getPropertyName();
+                    Object o = evt.getNewValue();
+                    if (o == null || o instanceof W) {
+                        return;
+                    }
+                    if (o instanceof LazyValue) {
+                        LOG.log(Level.INFO, "Patched lazy value {0}", k);
+                        defs.put(k, new LazyWrapper(k, (LazyValue)o));
+                    } else if (o instanceof ActiveWrapper) {
+                        LOG.log(Level.INFO, "Patched active value {0}", k);
+                        defs.put(k, new ActiveWrapper(k, (ActiveWrapper)o));
+                    } else if (o instanceof Icon || o instanceof Image) {
+                        LOG.log(Level.INFO, "Patched Icon/Image value {0}", k);
+                        defs.put(k, new SimpleWrapper(k, o));
+                    }
+                }
+            }
+        });
+    }
+    
+    public static void refreshUIDefaults(UIDefaults defs) {
+        for (Enumeration en = defs.keys(); en.hasMoreElements(); ) {
+            Object ok = en.nextElement();
+            
+            // Note: this actually instantiates the object, if it implements LazyValue.
+            Object o = defs.get(ok);
+            if (o instanceof W) {
+                continue;
+            }
+            if (o instanceof Icon || o instanceof Image) {
+                LOG.log(Level.INFO, "Replaced {0} with wrapped", ok);
+                defs.put(ok, new SimpleWrapper(ok.toString(), o));
+            }
+        }
+    }
+    
+    private static Object wrapInstance(Object ok, Object o) {
+        if (o instanceof W) {
+            return o;
+        }
+        try {
+            if (o instanceof ImageIcon) {
+                ImageIcon ii = (ImageIcon)o;
+                Object wrapper = wrapImage(ok.toString(), ii.getImage(), ii,
+                        PROP_ORIGINAL_IMAGE, ii.getImage(), PROP_ORIGINAL_ICON, ii);
+                if (wrapper != ii.getImage()) {
+                    return (Icon)wrapper;
+                } else {
+                    return o;
+                }
+            }
+
+            if (o instanceof Icon) {
+                final Icon ico = (Icon)o;
+                Image converted = ImageUtilities.icon2Image(ico);
+                Object wrapper = wrapImage(ok.toString(), converted, ico, PROP_ORIGINAL_ICON, ico);
+                if (wrapper != converted) {
+                    return (Icon)wrapper;
+                } else {
+                    return o;
+                }
+            }
+
+            if (o instanceof Image) {
+                Object wrapper = wrapImage(ok.toString(), (Image)o, null, PROP_ORIGINAL_IMAGE, o);
+                return wrapper;
+            }
+        } catch (RuntimeException ex) {
+            // setting to FINE as some icons cannot be image-converted (ClassCastEx during paintIcon).
+            LOG.log(Level.FINE, "Error wrapping default {0}: ", ok.toString());
+            LOG.log(Level.FINE, "Exception thrown", ex);
+        }
+        return o;
+    }
+    
+    private static void replace(UIDefaults defs, Object key, String type, Object replacement) {
+        LOG.log(Level.INFO, "Replaced {0}", key);
+        defs.put(key, replacement);
+    }
+    
+    public static final class MetadataImageIcon extends ImageIcon {
+        private final Icon delegateIcon;
+
+        public MetadataImageIcon(Icon delegateIcon, Image original) {
+            super(original);
+            this.delegateIcon = delegateIcon;
+        }
+
+        /**
+         * Get an {@link Icon} instance representing a scalable version of this {@code Image}.
+         *
+         * @return may be null
+         */
+        public Icon getDelegateIcon() {
+            return delegateIcon;
+        }
+
+        @Override
+        public int getIconHeight() {
+            return delegateIcon.getIconHeight();
+        }
+
+        @Override
+        public int getIconWidth() {
+            return delegateIcon.getIconWidth();
+        }
+
+        @Override
+        public void paintIcon(Component c, Graphics g, int x, int y) {
+            delegateIcon.paintIcon(c, g, x, y);
+        }
+    }
+    
+    /**
+     * Delegating Image which adds metadata to the original image's {@link Image#getProperty}.
+     * @author sdedic
+     */
+    public static class MetadataImage extends BufferedImage implements W {
+        private final Image original;
+
+        public MetadataImage(Image original, ColorModel cm,
+                          WritableRaster raster,
+                          boolean isRasterPremultiplied,
+                          Hashtable<?,?> properties) {
+            super(cm, raster, isRasterPremultiplied, properties);
+            this.original = original;
+        }
+
+        @Override
+        public Object getProperty(String name, ImageObserver observer) {
+            Object o = super.getProperty(name, observer);
+            return o != UndefinedProperty ? o : original.getProperty(name, observer);
+        }
+    }
+    
+    public static Object createNew(Image image, Icon icon, Hashtable props) {
+        ensureLoaded(image);
+        boolean bitmask = (image instanceof Transparency) && ((Transparency) image).getTransparency() != Transparency.TRANSLUCENT;
+        ColorModel model = colorModel(bitmask ? Transparency.BITMASK : Transparency.TRANSLUCENT);
+        int w = Math.max(1, image.getWidth(null));
+        int h = Math.max(1, image.getHeight(null));
+        if (icon == null && image instanceof Icon) {
+            icon = (Icon)image;
+        }
+        WritableRaster raster = model.createCompatibleWritableRaster(w, h);
+        BufferedImage newImage = new MetadataImage(image, model, raster, bitmask, props);
+
+        java.awt.Graphics g = newImage.createGraphics();
+        g.drawImage(image, 0, 0, null);
+        g.dispose();
+        if (icon != null) {
+            return new MetadataImageIcon(icon, newImage);
+        } else {
+            return newImage;
+        }
+    }
+    
+    private interface W {}
+    
+    private static final class LazyWrapper implements LazyValue, W {
+        private final String key;
+        private final LazyValue original;
+
+        public LazyWrapper(String key, LazyValue original) {
+            this.key = key;
+            this.original = original;
+        }
+        
+        @Override
+        public Object createValue(UIDefaults table) {
+            LOG.log(Level.INFO, "Resolved lazy wrapped for {0}", key);
+            return wrapInstance(key, original.createValue(table));
+        }
+    }
+    
+    private static final class ActiveWrapper implements ActiveValue, W {
+        private final String key;
+        private final ActiveValue original;
+
+        public ActiveWrapper(String key, ActiveValue original) {
+            this.key = key;
+            this.original = original;
+        }
+
+        @Override
+        public Object createValue(UIDefaults table) {
+            LOG.log(Level.INFO, "Resolved active wrapped for {0}", key);
+            return wrapInstance(key, original.createValue(table));
+        }
+    }
+    
+    private static final class SimpleWrapper implements LazyValue, W {
+        private final String key;
+        private final Object original;
+
+        public SimpleWrapper(String key, Object original) {
+            this.key = key;
+            this.original = original;
+        }
+        
+        @Override
+        public Object createValue(UIDefaults table) {
+            LOG.log(Level.INFO, "Resolved wrapped for {0}", key);
+            return wrapInstance(key, original);
+        }
+    }
+
+    // ---------- Copied from ImageUtilities - needed to render the image --------
+    
+    private static final Component component = new Component() {};
+
+    private static final MediaTracker tracker = new MediaTracker(component);
+
+    private static int mediaTrackerID;
+
+    private static void ensureLoaded(Image image) {
+        if (
+            (Toolkit.getDefaultToolkit().checkImage(image, -1, -1, null) &
+                (ImageObserver.ALLBITS | ImageObserver.FRAMEBITS)) != 0
+        ) {
+            return;
+        }
+
+        synchronized (tracker) {
+            int id = ++mediaTrackerID;
+            tracker.addImage(image, id);
+
+            try {
+                tracker.waitForID(id, 0);
+            } catch (InterruptedException e) {
+                System.out.println("INTERRUPTED while loading Image");
+            }
+
+            // #262804 assertation disabled because of error, when using ImageFilter
+            // assert (tracker.statusID(id, false) == MediaTracker.COMPLETE) : "Image loaded";
+            tracker.removeImage(image, id);
+        }
+    }
+
+    static private ColorModel colorModel(int transparency) {
+        ColorModel model;
+        try {
+            model = java.awt.GraphicsEnvironment.getLocalGraphicsEnvironment()
+                .getDefaultScreenDevice().getDefaultConfiguration()
+                .getColorModel(transparency);
+        }
+        catch(ArrayIndexOutOfBoundsException | HeadlessException aioobE) {
+            //#226279
+            model = ColorModel.getRGBdefault();
+        }
+        return model;
+    }
+
+    // ---------- End copy
+}
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
@@ -56,6 +56,7 @@ public class DefaultDecorationsImpl implements TreeDataProvider.Factory {
     public static final String EXPLORER_ROOT = "Explorers"; // NOI18N
     public static final String COOKIES_EXT = "contextValues"; // NOI18N
     
+    public static final String CTXVALUE_FILE = "is:file"; // NOI18N
     public static final String CTXVALUE_FOLDER = "is:folder"; // NOI18N
     public static final String CTXVALUE_PROJECT = "is:project"; // NOI18N
     public static final String CTXVALUE_PROJECT_ROOT = "is:projectRoot"; // NOI18N
@@ -166,7 +167,8 @@ public class DefaultDecorationsImpl implements TreeDataProvider.Factory {
                                 }
                                 folder = true;
                                 d.addContextValues(CTXVALUE_FOLDER);
-                            } else {
+                            } else if ((f.isData() || f.isValid()) && !f.isVirtual()) {
+                                d.addContextValues(CTXVALUE_FILE);
                                 // PENDING: this could be moved to the VSNetbeans module ?
                                 d.setCommand("vscode.open"); // NOI18N
                             }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
@@ -43,7 +43,6 @@ public class TreeItem {
     public String description;
     // ?: string | Uri | {dark: string | Uri, light: string | Uri} | ThemeIcon
     public URI iconUri;
-    public int iconIndex;
     // id for the tree item that has to be unique across tree.
     // The id is used to preserve the selection and expansion state of the tree item.
     public int id;
@@ -56,6 +55,27 @@ public class TreeItem {
     public String resourceUri;
     // ?: string | MarkdownString | undefined
     public Object tooltip;
+    
+    // NBLS-specific items, to be processed by the client:
+    // Metadata for the icon.
+    public IconDescriptor iconDescriptor;
+    // The icon's index
+    public int iconIndex;
+    
+    /**
+     * Metadata that describe an icon origin or contents.
+     */
+    public static class IconDescriptor {
+        /**
+         * Base URI / imageId of the icon.
+         */
+        public URI baseUri;
+        
+        /**
+         * Supplemental IDs from composed merged-in images or applied filters.
+         */
+        public String[] composition;
+    }
 
     public TreeItem() {
     }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistry.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistry.java
@@ -20,6 +20,8 @@ package org.netbeans.modules.java.lsp.server.explorer;
 
 import java.awt.Image;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.concurrent.CompletionStage;
 import org.openide.nodes.Node;
 
@@ -40,6 +42,19 @@ public interface TreeNodeRegistry {
     CompletionStage<TreeViewProvider> createProvider(String id);
     
     public class ImageDataOrIndex {
+        /**
+         * Base URL of the image, if it could be extracted from the Image object.
+         */
+        public URI          baseURI;
+        
+        /**
+         * URLs(?) of additional image constituents.
+         */
+        public String[]     composition;
+        
+        /**
+         * Image URI, data: protocol.
+         */
         public final URI    imageURI;
         public final int    imageIndex;
 
@@ -56,6 +71,25 @@ public interface TreeNodeRegistry {
         public ImageDataOrIndex(int imageIndex) {
             this.imageIndex = imageIndex;
             this.imageURI = null;
+        }
+        
+        public ImageDataOrIndex baseURL(URL u) {
+            try {
+                baseURI = u == null ? null : u.toURI();
+            } catch (URISyntaxException ex) {
+                throw new IllegalArgumentException(ex);
+            }
+            return this;
+        }
+        
+        public ImageDataOrIndex baseURL(URI u) {
+            this.baseURI = u;
+            return this;
+        }
+        
+        public ImageDataOrIndex composition(String[] composition) {
+            this.composition = composition;
+            return this;
         }
     }
 }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistryImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistryImpl.java
@@ -259,21 +259,18 @@ public class TreeNodeRegistryImpl implements TreeNodeRegistry {
 
 
     public static URI findImageURI(Image i) {
-        Object o = i.getProperty("url", null);
+        String s = ImageUtilities.findImageBaseId(i);
+        if (s == null) {
+            return null;
+        }
         try {
-            if (o instanceof URL) {
-                return ((URL)o).toURI();
-            }
-            o = i.getProperty("uri", null);
-            if (o instanceof URI) {
-                return (URI)o;
-            } else if (o instanceof String) {
-                return new URI(o.toString());
+            if (s.contains(":")) {
+                return new URI(s);
             } else {
-                return null;
+                return new URI("nbres:/" + s);
             }
         } catch (URISyntaxException ex) {
-            LOG.log(Level.WARNING, "Unable to interpret image URL: {0}", o);
+            LOG.log(Level.WARNING, "Unable to interpret image ID: {0}", s);
             return null;
         }
     }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistryImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistryImpl.java
@@ -30,6 +30,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
@@ -49,6 +50,7 @@ import org.openide.explorer.ExplorerUtils;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
 import org.openide.nodes.Node;
+import org.openide.util.ImageUtilities;
 import org.openide.util.Lookup;
 import org.openide.util.lookup.Lookups;
 import org.openide.util.lookup.ProxyLookup;
@@ -220,7 +222,6 @@ public class TreeNodeRegistryImpl implements TreeNodeRegistry {
                 if (h == -1 || w == -1) {
                     try {
                         observer.cdl.await();
-                        Thread.sleep(1000);
                     } catch (InterruptedException ex) {
                     }
                     synchronized (observer) {
@@ -248,10 +249,33 @@ public class TreeNodeRegistryImpl implements TreeNodeRegistry {
         }
         synchronized (this) {
             if (res == null) {
-                res = new ImageDataOrIndex(imageURI, imageCounter++);
+                res = new ImageDataOrIndex(imageURI, imageCounter++).
+                        baseURL(findImageURI(i));
             }
             images.put(i, res);
         }
         return res;
     }
+
+
+    public static URI findImageURI(Image i) {
+        Object o = i.getProperty("url", null);
+        try {
+            if (o instanceof URL) {
+                return ((URL)o).toURI();
+            }
+            o = i.getProperty("uri", null);
+            if (o instanceof URI) {
+                return (URI)o;
+            } else if (o instanceof String) {
+                return new URI(o.toString());
+            } else {
+                return null;
+            }
+        } catch (URISyntaxException ex) {
+            LOG.log(Level.WARNING, "Unable to interpret image URL: {0}", o);
+            return null;
+        }
+    }
+    
 }
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/api/TreeItemData.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/api/TreeItemData.java
@@ -83,7 +83,7 @@ public final class TreeItemData {
         this.contextValues = contextValues;
         return this;
     }
-
+    
     public String getCommand() {
         return command;
     }
--- a/java/java.lsp.server/vscode/package.json
+++ b/java/java.lsp.server/vscode/package.json
@@ -661,8 +661,79 @@
 					"group": "inline@1"
 				}
 			]
-		}
+		},
+		
+		"netbeans.iconMapping" : [
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/gradle/resources/gradle.png",
+				"codeicon": "project"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/gradle/java/resources/javaseProjectIcon.png",
+				"codeicon": "project"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/maven/resources/jaricon.png",
+				"codeicon": "project"
+			},
+
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/gradle/resources/libraries.png",
+				"codeicon": "settings-gear"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/gradle/resources/(empty.png|module-artifact.png)",
+				"codeicon": "file-zip"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/maven/DependencyIcon.png",
+				"codeicon": "file-zip"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/maven/TransitiveDependencyIcon.png",
+				"codeicon": "library"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/spi/java/project/support/ui/package(Empty)?.gif",
+				"codeicon": "file-submodule"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/java/resources/(abstract_class_file|class|enum_file).png",
+				"codeicon": "symbol-class"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/java/resources/(annotation_file|interface_file).png",
+				"codeicon": "symbol-interface"
+			},
+			{
+				"uriExpression": "uidefaults:Tree.(open|closed)Icon",
+				"valueMatch": [ " [0-9]+([A-Z][A-z]+)?SourceRoot | Dependencies[A-Z]+ | OtherRoots | projectfiles | BootCPNode " ],
+				"codeicon": "file-submodule"
+			},
+			{
+				"uriExpression": "uidefaults:Tree.(open|closed)Icon",
+				"valueMatch": [ " [0-9]+([A-z]+).(java|groovy|generated|resources) " ],
+				"codeicon": "file-submodule"
+			},
+			{
+				"uriExpression": "uidefaults:Tree.(open|closed)Icon",
+				"codeicon": "*folder"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/java/api/common/project/ui/resources/platform.gif",
+				"codeicon": "vm"
+			}
+		],
+	
+		"jsonValidation": [
+			{
+				"fileMatch": "package.json",
+				"url": "./schemas/package.schema.json"
+			}
+		]
 	},
+
+
 	"scripts": {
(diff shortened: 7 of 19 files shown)


java/java.lsp.server/vscode/package.json
```diff
@@ -661,8 +661,79 @@
...
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/maven/TransitiveDependencyIcon.png",
+				"codeicon": "library"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/spi/java/project/support/ui/package(Empty)?.gif",
+				"codeicon": "file-submodule"
+			},
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/java/resources/class.png",
```
dbalek, line 701: What about `abstract_class_file.png`, `main-class.png`, and 'enum_file.png`

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
```diff
@@ -56,6 +56,7 @@
     public static final String EXPLORER_ROOT = "Explorers"; // NOI18N
     public static final String COOKIES_EXT = "contextValues"; // NOI18N
     
+    public static final String CTXVALUE_FILE = "is:file"; // NOI18N
```
JaroslavTulach, line 59: Probably it should be documented somewhere.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,213 @@
...
+        UIDefaults defs = UIManager.getDefaults();
+        
+        for (Enumeration en = defs.keys(); en.hasMoreElements(); ) {
+            Object ok = en.nextElement();
+            Object o = defs.get(ok);
+            try {
+
+                if (o instanceof ImageIcon) {
+                    ImageIcon ii = (ImageIcon)o;
+                    Image wrapper = wrapImage(ok.toString(), ii.getImage(),
```
JaroslavTulach, line 103: OK, wrapping the images to have attributes.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,213 @@
...
+@OnStart
+public class UIDefaultsIconMetadata implements Runnable {
+    private static final Logger LOG = Logger.getLogger(UIDefaultsIconMetadata.class.getName());
+    
+    public static final String PROP_URI = "uri"; // NOI18N
+    public static final String PROP_URL = "url"; // NOI18N
+    public static final String PROP_ORIGINAL_ICON = "originalIcon"; // NOI18N
+    public static final String PROP_ORIGINAL_IMAGE = "originalImage"; // NOI18N
+    
+    private static final String URN_DEFAULTS_PREFIX = "urn:uidefaults:"; // NOI18N
```
JaroslavTulach, line 58: Is the whole introduction of `"uri"` because you want to use this `URN_DEFAULTS_PREFIX`!? I understand that when starting from scratch the ability to have real `URI` with its "magical strings" might be beneficial, but we are not starting from scratch. We have `"url"` on images - it'd be more hidden if we could reuse it and just enhance it.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,213 @@
...
+    
+    @Override
+    public void run() {
+        EventQueue.invokeLater(() -> decorateUIDefaults());
+    }
+    
+    public static void decorateUIDefaults() {
+        UIDefaults defs = UIManager.getDefaults();
+        
+        for (Enumeration en = defs.keys(); en.hasMoreElements(); ) {
```
JaroslavTulach, line 96: Replacing `UIManager.getDefaults()` and decorating the original items when one asks for them would also be an option. Faster on startup, probably. Converting only the icons that are needed.

JaroslavTulach (approved):
I'd prefer to not introduce the `"uri"` `Image` key. Should there be one, please add a utility method to extract `URI` from an `Image`.

A bit of documentation and some tests (Java as well as TypeScript) would be ideal at the end.

sdedic:
> I'd prefer to not introduce the `"uri"` `Image` key. Should there be one, please add a utility method to extract `URI` from an `Image`.

Let's get a little philosophical (there are more hints to use URL, so let's sort that in one place). The goal is to provide an ID. And there are images that one cannot get the physical bits. Most importantly the images in the `UIDefaults`, but possibly other ones as well. 
The distinction between URI and URL is that the URI is really an **identifier** ... which sometimes might be the **location** at the same time (URL). But some URIs are not (= URN). The most important functional distinction is that URL allows to locate the physical resource = access its bits. That's what `URLStreamHandler`s are for.

I don't have the bits for images from UIDefaults. They can be forged, by painting the Icon + encoding as PNG, for example - then the special URL with an associated `URLStreamHandler` instance could eventually work. But such URLs would not be usable anywhere - an attempt to pass them around will eventually lead to .toString() and back - which will fail, since the protocol would not be probably registered - and the referenced bits are ephemeral.

I can invent an additional protocol, 'uidefaults' and register a handler for it that would access the LaF (authority) icon/image (path). But still the resource's content would be computed. Allowing for URIs does not preclude that, but allows true identifiers for ephemeral objects. So IMHO there should have been URIs from the start - so that the image was not required to be a locatable, but just identifiable.

Does that explain the introduction - does it seem acceptable ? If there's strong disagreement, then my preferred best solution is that special protocol.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/DefaultDecorationsImpl.java
sdedic: I think it is, in [`arch.xml`](https://github.com/apache/netbeans/pull/3414/files#diff-3d2e53a800fa8bf1c32071f2a26d981c1d4f698a9e85504fac40efaf3b402b25R964)

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/api/TreeItemData.java
```diff
@@ -83,7 +83,7 @@ public TreeItemData setContextValues(String... contextValues) {
         this.contextValues = contextValues;
         return this;
     }
-
+
```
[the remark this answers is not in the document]
sdedic: Leftover after refactoring, thanks

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistryImpl.java
```diff
@@ -248,10 +249,33 @@ public boolean imageUpdate(Image img, int infoflags, int x, int y, int width, in
...
+                res = new ImageDataOrIndex(imageURI, imageCounter++).
+                        baseURL(findImageURI(i));
             }
             images.put(i, res);
         }
         return res;
     }
+
+
+    public static URI findImageURI(Image i) {
```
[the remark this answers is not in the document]
sdedic: OK. I had those accessors, but removed them.

But see the note about not introducing this as the official API _now_ - but definitely until the 13 release. As noted in the description, there are still some missing pieces that would be exposed in the APIs (the decorations/badge components), I'd prefer to publish it as a whole. OK ?

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -825,22 +850,62 @@ private static final ToolTipImage doMergeImages(Image image1, Image image2, int
...
+    /**
+     * Retrieves the original icon from a possible wrapper.
+     * @param image image
+     * @return the icon, if reachable.
+     */
+    private static Icon originalIcon(Image image) {
+        if (image instanceof ToolTipImage) {
+            return ((ToolTipImage) image).getDelegateIcon();
+        }
+        Object o = image.getProperty("originalIcon", null); // NOI18N
```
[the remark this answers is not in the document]
sdedic: Yes. This one I am not proud of, but I've didn't find a sensible way how to work with a delegating Image :( Will be documented with the other magic property strings.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -1069,31 +1140,55 @@ public static ToolTipImage createNew(String toolTipText, Image image, URL url) {
             if (url == null) {
                 Object value = image.getProperty("url", null);
                 url = (value instanceof URL) ? (URL) value : null;
-            }            
-            Icon icon = (image instanceof ToolTipImage)
-                    ? ((ToolTipImage) image).getDelegateIcon() : null;
+            }        
+            if (uri == null) {
+                Object value = image.getProperty("uri", null);
```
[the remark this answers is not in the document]
sdedic: See the 'philosophical' note below.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -352,7 +369,14 @@ private static ToolTipImage icon2ToolTipImage(Icon icon, URL url) {
         should really only be called on the Event Dispatch Thread. Constructing the component once
         on the EDT fixed the problem. Read-only operations from non-EDT threads shouldn't really be
         a problem; most Icon implementations won't ever access the component parameter anyway. */
-        icon.paintIcon(dummyIconComponent, g, 0, 0);
+        try {
+            icon.paintIcon(dummyIconComponent, g, 0, 0);
+        } catch (ClassCastException ex) {
+            // java.desktop/javax.swing.plaf.metal.OceanTheme$IFIcon.paintIcon assumes a different component,
+            // so let's try second most used one type, it satisfies AbstractButton, JCheckbox. Not all cases are
+            // covered, however.
```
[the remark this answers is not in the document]
sdedic: Well, more cases will be resolved by the delegating lazy UIDefaults that you've suggested.

java/java.lsp.server/vscode/schemas/package.schema.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "LSP Image Resource Mapping contributions to package.json",
+    "type": "object",
+    "properties": {
+            "contributes": {
+                    "type": "object",
+                    "properties": {
+                            "netbeans.iconMapping" : {
```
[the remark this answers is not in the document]
sdedic: What API category should that be ? Layer ;) ?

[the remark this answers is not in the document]
JaroslavTulach: package.json!?

No need to add it into `arch.xml` - this is really VSNetBeans specific. However other extension writers should know how much they can rely on this. I'd also like to know whether we want to support this `iconMapping` _forever_ or rewrite it the next version?

JaroslavTulach:
First of all I'd like to point out I have already approved your PR. The changes are sane and acceptable from some architecture perspective. My following comments are just result of viewing them from a different angle...

> Let's get a little philosophical...

Thanks for writing down your thoughts. Yes, I expected you had something like that in mind. You worked in [Systinet](https://en.wikipedia.org/wiki/Mercury_Interactive) where **URL**, **URI**, **URN** were essential. Your focus on the nuances between them is understandable.

I also know that `mailto:` isn't technically URL. But for me `java.net.URL` is just a class, I have no mental problem to encode `mailto:` into `java.net.URL` and later throw an `org.openide.text.UserTypeSomeTextException`.

I am more scared of duplicating the _existing_ API by a (95% similar) concept, just to be _pure_ and absolutely correct.

>... goal is ... an ID ... images ... (without) ... the physical bits... URI is really an **identifier** ... URL allows to locate ... that's what `URLStreamHandler`s are for...invent an additional protocol, 'uidefaults' 

From a point of Java programmer and maintainer (creator) of the previous (messy?) system, I'd call the **URI** vs. **URL** debate an unimportant difference and I'd just encode `UIDefaults` & co. images in `java.net.URL` and later throw an exception (or paint the icon into `data:`, if useful).

> the protocol would not be probably registered - and the referenced bits are ephemeral.

Yes, there are problems passing Java URLs into OS or browser. Each has its own set of handlers and the intersection is small (`http`, `file`, `ftp`?).

> So IMHO there should have been URIs from the start 

Indeed!

However there was no `java.net.URI` when NetBeans started. The [java.net package](https://www.cs.princeton.edu/courses/archive/fall97/cs461/jdkdocs/api/Package-java.net.html) contained just `java.net.URL` and other APIs like [Class.getResource](https://www.cs.princeton.edu/courses/archive/fall97/cs461/jdkdocs/api/java.lang.Class.html#getResource(java.lang.String)) used it to refer to other objects than just those on _World Wide Web_.

If we started from scratch right now, I'd support your idea to use `java.net.URI`. However we have `java.net.URL` and I'd rather fit into existing design than complicate the conceptual area (from a Java programmer perspective) just because of little (philosophical) differences between **URL** and **URI**.

> Does that explain the introduction - does it seem acceptable ? 

Yes, your position is acceptable. It just leads to more complicated system. If we pretended **URI** could be represented by `java.net.URL` (as in the early days of Java), we'd get simpler result.

sdedic:
> First of all I'd like to point out I have already approved your PR. The changes are sane and acceptable from some architecture 

I know. This discussion is about coming to an agreement, or a practical compromise. Maybe 3rd view would help.

> I am more scared of duplicating the _existing_ API by a (95% similar) concept, just to be _pure_ and absolutely correct.

Well, not exactly. When `pseudoURL.open*()` throw an exception, you can't really know (if there's a mix of real+ephemeral URLs), if something broke or you just hit the ephemeral one, for example. If ever someone prints fakeURL.toString() and will try to convert it back, an exception is likely (handler not registered). fakeURL.toURI() will work, but fakeURL.toURI().toURL() will fail. We break the java.net.UR(I|L) behaviour at runtime ... but yes, the API would be simpler.

> However there was no `java.net.URI` when NetBeans started. The [java.net package](https://www.cs.princeton.edu/courses/archive/fall97/cs461/jdkdocs/api/Package-java.net.html) contained just `java.net.URL` and other APIs like [Class.getResource](https://www.cs.princeton.edu/courses/archive/fall97/cs461/jdkdocs/api/java.lang.Class.html#getResource(java.lang.String)) used it to refer to other objects than just those on _World Wide Web_.
> 
The Locator was not required to point to WWW. It is required to _locate_ the resource so its _content_ can be _accessed_ (other ops possible). The ClassLoader methods were right to use URLs, as the purpose was to access content (to a lesser degree identify the resource, but as each well-formed URL is URI, it's satisfied by the contract).
I believe we have a different situation here.

> Yes, your position is acceptable. It just leads to more complicated system. If we pretended **URI** could be represented by `java.net.URL` (as in the early days of Java), we'd get simpler result.

Hm, let's let some other party to weigh in :) there are yet some other reviewers available. Functionally, I am mostly afraid of weird bugs from circulating URLs, that are not URLs and break the usual URL conversion/usage patterns, through the system at runtime.

eirikbakke:
I started looking through this, but I'm boarding a plane now and haven't finished reviewing yet...

Adding URI in addition to URL seems to add a lot of complexity, and potential for confusion. Perhaps just add an optional String identifier, and then let icons be looked up by either the URL (if present) or the optional identifier?

There's a number of hacks that could add complexity here... I remember dealing with icon/image wrappers when adding SVG support, and it was a source of pain.

I will look more closely through this patch again in a day or two.

dbalek (approved):

sdedic:
> Adding URI in addition to URL seems to add a lot of complexity, and potential for confusion. Perhaps just add an optional String identifier, and then let icons be looked up by either the URL (if present) or the optional identifier?
> 
This ("url" + "id") is the conceptually the same as "url" + "uri", it is that dichotomy itself that is the centre of the heated philosophical debate. 

BTW as a improvement, there could be always an ID (String, URI), possibly derived from a supplied load URL _but_ unaffected by branding / localization (hey, another benefit of the duplication !) and sometimes possibly URL. Then a client may choose to get just the property it is interested in (location, identification). The current client code in the patch queries both, which is confusing; will fix.

Except that, URI has at least some structure :) compared to String, with Strings, we would still need to invent some convention (e.g. prefixing) so that ids from different sources remain unique (i.e. I should not probably provide just "Tree.openIcon", that's likely to clash, but "something.Tree.openIcon" ... which is quite equivalent to an URL/URI scheme +  host / path).

neilcsmith-net:
> Except that, URI has at least some structure :) compared to String, with Strings, we would still need to invent some convention (e.g. prefixing) so that ids from different sources remain unique (i.e. I should not probably provide just "Tree.openIcon", that's likely to clash, but "something.Tree.openIcon" ... which is quite equivalent to an URL/URI scheme + host / path).

That's an interesting point about specificity.  What would the path levels mean? I've said before that I'd wished NetBeans had a logical icon spec, something akin to xdg icon spec https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#guidelines  Here the level of specificity has meaning, and the most specific option found will be used, by removing the lower level and looking again.  eg. Tree.openIcon and Tree.openIcon.something would make sense.

sdedic:
> That's an interesting point about specificity. What would the path levels mean? I've said before that I'd wished NetBeans had a 
I wouldn't drag this discussion into a PR, maybe mailing list would be better. For the purpose of the PR (and accompanying documentation), I'd say that a module contributed to Apache NetBeans repository should coordinate its IDs with others, and external modules must use DNSlike prefix as if URI was created. That will satisfy uniqueness property required here, additional semantics should be IMHO solved separately

sdedic:
@JaroslavTulach @eirikbakke it seems that some consensus is bulding finally. Let me recap:
- I need some 'identifier' to identify icon for replacement; that can be sometimes derived from resource URL, but sometimes not.
- Client should make one call or extraction, not a series of get + instanceof
- URI is rather confusable with URL for other people than XML veterans, so people could choose wrong thing from URL/URI alternatives (especially when most of the time URI would be URL)

So ... I am going to:
- change the type to String, document minimal rules for uniqueness
- provide a helper static method to get URL and an (String) identifier (the instanceof cascade will be here)
- propagate "id" and "url" separately, not derive one form the other in ImageUtilities impl.

All OK ?

neilcsmith-net:
@sdedic you mentioned "invent some convention". I'm just suggesting ad(o/a)pting one that already exists.  And that branding by _unique_ IDs across modules is one of the flaws in NetBeans as it is.

eirikbakke:
@sdedic Yes, that would cover my own main concerns with ImageUtilities.

eirikbakke (commented):
I submitted a few in-code comments that I started yesterday... happy to review again when PR is updated with the latest proposed changes.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -825,22 +850,62 @@ private static final ToolTipImage doMergeImages(Image image1, Image image2, int
...
+    
+    private static Image originalImage(Image image) {
+        Object o = image.getProperty("originalImage", null); // NOI18N
+        return o instanceof Image ? (Image)o : image;
+    }
+
+    /**
+     * Draws an image on graphics. The difference between a direct {@link Graphics#drawImage(java.awt.Image, int, int, java.awt.image.ImageObserver)} call
+     * and this is that if the image is wrapped (i.e. as an attempt to add some metadata to it), the original will be used if the wrapper is cooperative. The
+     * wrapper image may not be compatible with the SurfaceManager and a direct draw would result in an IllegalArgumentException.
```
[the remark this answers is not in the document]
sdedic: I wrapped an Image instance in UIDefaults (an ImageIconUIResource, "Tree.openIcon") with a delegating implementation: created an Image delegate for the original Toolkit image, adding just the id metadata, delegating other stuff to the original. Wrapped in an `ImageIcon`, put back to UIDefaults.

When such image got painted, the SurfaceManager of the delegate was still `null`, [`SurfaceManager.getManager`](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/awt/image/SurfaceManager.java#L73) attempted to cast it to a `BufferedImage`, which it is not, then threw `IllegalArgumentException`.

..... and that's it ! My bad and thanks for make me looking into it once more. If I create a BufferedImage in the first place, the instance will work without any hassle. I thought that merely delegating `Image.getGraphics` will be sufficient, but it is not.

java/java.lsp.server/vscode/schemas/package.schema.json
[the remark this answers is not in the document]
sdedic: I would mark it as a highly experimental feature at this moment.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -1178,23 +1281,27 @@ logical pixel position (7,0) would map to device pixel position (10.5,0).
 
         @Override
         public Object getProperty(String name, ImageObserver observer) {
-            if ("url".equals(name)) { // NOI18N
+            if ("url".equals(name) || "uri".equals(name)) { // NOI18N
```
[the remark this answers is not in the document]
sdedic: Done in 2b887f0

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: Resolved using @eirikbakke suggestion to use Strings. The client instanceof duality was resolved by helper functions.

platform/openide.util.ui/src/org/openide/util/FilteredIcon.java
```diff
@@ -87,4 +89,8 @@ protected Image createAndPaintImage(
         return Toolkit.getDefaultToolkit().createImage(
                 new FilteredImageSource(img.getSource(), filter));
     }
+    
+    public String getFilterType() {
+        return filterType;
```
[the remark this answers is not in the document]
sdedic: This is a leftover from an additional feature I am not able to deliver atm because of time constraints; will follow up in another PR. There should be something as 'image composition' for images merged by `ImageUtiltiies` - now after adjustments that would be represented as a String[]. That would allow clients to really distinguish bare and decorated images, should they need so (the LSP client may need, since the badges change semantic that may be represented differently by the LSP client).

Removed in 6c944b9

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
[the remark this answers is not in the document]
sdedic: Implementation changed in 2b887f0 -- access to the original is no longer needed to render.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -825,22 +850,62 @@ private static final ToolTipImage doMergeImages(Image image1, Image image2, int
...
+     * and this is that if the image is wrapped (i.e. as an attempt to add some metadata to it), the original will be used if the wrapper is cooperative. The
+     * wrapper image may not be compatible with the SurfaceManager and a direct draw would result in an IllegalArgumentException.
+     * 
+     * @param g graphics
+     * @param image image to draw
+     * @param x x coordinate
+     * @param y y coordinate
+     * @param observer the image observer callback.
+     */
+    private static void drawImage(Graphics g, Image image, int x, int y, ImageObserver observer) {
```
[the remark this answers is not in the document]
sdedic: Removed from 2b887f0

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
```diff
@@ -42,6 +42,7 @@
     // rendered less prominent. When true, it is derived from resourceUri
     public String description;
     // ?: string | Uri | {dark: string | Uri, light: string | Uri} | ThemeIcon
+    public IconDescriptor iconDescriptor;
```
JaroslavTulach, line 45: Is the comment on line 45 supposed to apply to `IconDescriptor`? It doesn't seem to match.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
```diff
@@ -56,6 +57,11 @@
     public String resourceUri;
     // ?: string | MarkdownString | undefined
     public Object tooltip;
+    
+    public static class IconDescriptor {
+        public URI baseUri;
+        public String[] composition;
```
JaroslavTulach, line 63: Please add a comment describing the meaning of `composition` array.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistry.java
```diff
@@ -40,6 +42,19 @@
     CompletionStage<TreeViewProvider> createProvider(String id);
     
     public class ImageDataOrIndex {
+        /**
+         * Base URL of the image, if it could be extracted from the Image object.
+         */
+        public URI          baseURI;
+        
+        /**
+         * URLs(?) of additional image constituents.
```
JaroslavTulach, line 51: Here is some javadoc for `composition`. Surprises me there is `IconDescriptor` and `ImageDataOrIndex`...

JaroslavTulach (approved):
If you are OK with it Sváťo, feel free to proceed. Kudos to XML veterans!

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeNodeRegistry.java
sdedic: I know - the protocol need to be cleared up, ideally during NB13 release cycle. The client should call back with some getImageData(). I do not want to mess with existing parts because of the vscode release timing, but I am planning to clean it up so the LSP server releases clean with NB13.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
sdedic: Fixed in b20f0d6

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
sdedic: Fixed in b20f0d6

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -118,17 +125,26 @@
...
+     * 
+     */
+    private static volatile Component dummyIconComponent2;
 
     static {
         /* Could have used Mutex.EVENT.writeAccess here, but it doesn't seem to be available during
         testing. */
         if (EventQueue.isDispatchThread()) {
             dummyIconComponent = new JLabel();
+            dummyIconComponent2 = new JButton();
```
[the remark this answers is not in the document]
sdedic: Fixed in 30d43a7

java/java.lsp.server/vscode/package.json
sdedic: Added in eec8fe1

sdedic:
Will squash & merge early morning (CET) tomorrow.

java/java.lsp.server/vscode/package.json
```diff
@@ -661,8 +661,79 @@
 					"group": "inline@1"
 				}
 			]
-		}
+		},
+		
+		"netbeans.iconMapping" : [
+			{
+				"uriExpression": "nbres:/org/netbeans/modules/gradle/resources/gradle.png",
+				"codeicon": "project"
```
eirikbakke, line 669: nit: Should probably be codeIcon (with capital I).

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -447,6 +478,53 @@ public static Image createDisabledImage(Image image)  {
...
+     */
+    public static URL findImageBaseURL(Image image) {
+      Object o = image.getProperty(PROPERTY_URL, null);
+      return o instanceof URL ? (URL)o : null;
+    }
+    
+    /**
+     * Attempts to find image's ID, if it is defined. The image may have an ID even
+     * though it has no URL. An image whose {@link #findImageBaseURL(java.awt.Image)} returns
+     * non-null has always an ID. Note that image IDs may be the same for bare and badged
```
eirikbakke, line 500: Thanks for this... I like it better.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -1112,6 +1201,7 @@ public ToolTipImage(Icon delegateIcon, String toolTipText, URL url, int imageTyp
             this.delegateIcon = delegateIcon;
             this.toolTipText = toolTipText;
             this.url = url;
+            this.imageId = null;
```
eirikbakke, line 1204: Do we not need to pass the imageId along here?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,357 @@
...
+import javax.swing.UIDefaults;
+import javax.swing.UIDefaults.ActiveValue;
+import javax.swing.UIDefaults.LazyValue;
+import javax.swing.UIManager;
+import org.openide.modules.OnStart;
+import org.openide.util.ImageUtilities;
+
+/**
+ * Goes through UIDefaults and replaces Icons and Images with variants that
+ * return image's URL from the 'url' property.
```
eirikbakke, line 53: How do we know that this is run after the UIDefaults customizations in o.n.swing.plaf or o.n.swing.laf.flatlaf? Does the invokeLater call really guarantee this?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,357 @@
...
+ * @author sdedic
+ */
+@OnStart
```
eirikbakke, line 56: Is this class going to be run always, in the regular NetBeans IDE configuration? Or is it just during some special uses by VSCode or such?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,357 @@
...
+        int w = Math.max(1, image.getWidth(null));
+        int h = Math.max(1, image.getHeight(null));
+        if (icon == null && image instanceof Icon) {
+            icon = (Icon)image;
+        }
+        WritableRaster raster = model.createCompatibleWritableRaster(w, h);
+        BufferedImage newImage = new MetadataImage(image, model, raster, bitmask, props);
+
+        java.awt.Graphics g = newImage.createGraphics();
+        g.drawImage(image, 0, 0, null);
```
eirikbakke, line 251: This doesn't work for HiDPI icons or MultiResolutionImage? What kind of icons is this used for?

eirikbakke (commented):
Thanks for the changes in ImageUtilities! I'm not familiar with the VSCode, LSP stuff etc., but I had a look at everything.

My biggest question is about UIDefaultsIconMetadata. Is this going to be run as part of the regular NetBeans IDE now? If so I want to take a closer look at it...

In particular, how does UIDefaultsIconMetadata handle Icon implementations other than ImageIcon, or ImageIcon instances that use a MultiResolutionImage?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: No, it should be in `java.lsp.server/nbcode`, so it should be _present_ only in a special headless distribution of Apache NB Language Server, not in the standard IDE distribution.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
sdedic: The other constructor is called from `assignToolTipToImageInternal` that reads `imageID` property from the image. That one is called from the `image2Icon`, which is used by `mergeImages`.In addition 
```
icon2Image(image2Icon(image_with_imageId))
```
ought to provide an image that still returns `imageID` property. So I think yes, TooltipImage should remember it.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: Hm. I think I can _force_ the configured LaF initialization by getting some UIManager's value.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: In _this first_ version (which affects just NBLS headless distribution) - icons are exported from Nodes. Could you please suggest how to improve this part in the future so the HiDPI works OK ?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
eirikbakke: Ah, great. I was concerned that every icon in the IDE was going to be subject to yet another wrapping layer...

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
eirikbakke: > In this first version (which affects just NBLS headless distribution)
Is a future version of UIDefaultsIconMetadata supposed to actually run as part of the NetBeans IDE? Or is it always just in this headless mode?

(In any case, I'll have a look at the HiDPI implications here...)

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+    
+    @Override
+    public void run() {
+        EventQueue.invokeLater(() -> {
+            replaceIconsAndImages(UIManager.getDefaults());
+        });
+    }
+    
+    static void replaceIconsAndImages(UIDefaults defs) {
```
eirikbakke, line 96: Could you explain why the UIDefaultsIconMetadata wrapping is necessary? As far as I know, icons that are loaded from files already have the "url" property set by ImageUtilities.

There's a lot of LAF customizations (and hacks) that interact with UIManager, and hooking up to it this way seems fragile and likely to cause problems in the future. Furthermore, the Icon wrapping logic in ImageUtilities is already hard to understand and debug, and I'd hate to wrap all the Icon instances in yet another layer...

mbien:
i am a bit late to the URI vs URL discussion but i only wanted to add that I am avoiding using URLs whenever possible. The class has several non obvious problems, e.g it does dns lookups in the background. This means that equals and hash code has to do IO which causes a chain of issues. All what is needed is someone putting URLs as keys into a map.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: At this moment, we need to **replace** certain icons bcs their style does not match the LSP client's appearance. This does not apply to NB -- the icons would be simply changed or rebranded.

The client (vscode) has some icons already that use font glyphs, rebranding is not that much possible. Extracting glyphs, converting (e.g.) to SVGs for branding could bring some licensing issues I'd like to avoid. You're right (see other question) that the implementation is not optimal - this will hopefully improve.

Re. usage in NB apps that actually display the icons - I can think of possible usage for A11Y purposes, but did not went deep that way.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: My goal was to provide "some identification" for an Image now exported to LSP, so the client can decide what to do with it. Ideally without changing **all** the places where an image is used or loaded. In addition images are loaded from UIDefaults by Swing as well (which I can not change). 

It may be the case that eventually I'll go through NB codebase and change the UIManager.getIcon() calls one by one (after I make up some alternative and for a SPI)... then this wrapper code code would eventually vanish without impact on its clients.

This place is at this moment the central and only one that knows, for a image, some ID - the hashmap key.

sdedic:
Rebased on master & consolidated fixes.

sdedic:
@mbien this is another philosophical one ;) if interested, let's take it on the mailing list.

JaroslavTulach:
> avoiding using URLs whenever possible

Great idea when you write new code or refactor internals of existing one. However here we are dealing with 20 years old already existing API. Just replacing `java.net.URL` with `java.net.URI` would cause uncountable disruptions in NetBeans own source code as well as applications using the NetBeans Platform.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
neilcsmith-net: I share Eirik's concerns, and also have concerns about changing the UIManager calls without some coordination. An API / SPI approach is certainly where I'd like to see this evolve, but that overlaps with a bunch of other conversations going on at the moment (inc. hidpi, svg, dark icons, branding). As you said before, there's a discussion to be had on dev@ about this. I still think this PR is potentially addressing the issue in the wrong place, but also that it's a case of needs must right now.

mbien:
> > avoiding using URLs whenever possible
> 
> Great idea when you write new code or refactor internals of existing one. However here we are dealing with 20 years old already existing API. Just replacing `java.net.URL` with `java.net.URI` would cause uncountable disruptions in NetBeans own source code as well as applications using the NetBeans Platform.

@JaroslavTulach I was not proposing to change any existing API. (although JDK 8 would have all tools it needs for a migration, default methods and the fact that there are easy ways to convert between URL and URI - its basically a best case scenario).

But this should be considered for new APIs and impls which use the actual classes - since its easy to overlook. I also use the word "avoiding" in the "try not to use if possible" sense, not in the "its forbidden" sense.

> @mbien this is another philosophical one ;) if interested, let's take it on the mailing list.

@sdedic the philosophical discussion probably happened when URL was implemented, now we can only try to find technical solutions to deal with the non-obvious side effects of the URL class (e.g. [inspections](https://github.com/mbien/jackpot-inspections/blob/master/CollectionPerformance.hint#L233-L264)). JEP 418 might help mitigating some side effects, but thats JDK 18 and would be only a mitigation.

graemerocher:
Is there a snapshot build of the vsix for this that can be tested?

sdedic:
The artifacts currently being voted on (https://dist.apache.org/repos/dist/dev/netbeans/netbeans-vscode-ext/12.6.301/) should contain this change.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
eirikbakke: @neilcsmith-net Oh, I always assumed that this PR was part of the effort to let icons be branded for LAF purposes. Is there any hope of satisfying both needs with one API change?

> It may be the case that eventually I'll go through NB codebase and change the UIManager.getIcon() calls one by one (after I make up some alternative and for a SPI)... then this wrapper code code would eventually vanish without impact on its clients.

Perhaps this might actually be preferable to hacking UIManager. How many of the icons would actually require this? It's only the select few that don't have an URL to a PNG or GIF for some reason, right?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,359 @@
...
+        if (o instanceof String || o instanceof URI) {
+            return image;
+        } else {
+            
+            Hashtable props = new Hashtable();
+            for (int i = 0; i < meta.length; i+= 2) {
+                props.put(meta[i], meta[i+1]);
+            }
+            props.put(ImageUtilities.PROPERTY_ID, URN_DEFAULTS_PREFIX + key);
+            return createNew(image, icon, props);
```
eirikbakke, line 83: OK, I'm actually confused about what UIDefaultsIconMetadata does. Is it trying to use the UIManager key to identify the icon?

sdedic: OK, I should change the (now after the changes) obsolete javadoc. The only purpose of that is to 'somehow add metadata' to the Icons and Images available from `UIManager.getIcon/Image()`. It does so by wrapping the originals into a delegate that provides appropriate value(s) from `Image.getProperty` - and that's all.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: @eirikbakke note that this 'replacement' solves also icons queried by Swing codebase that might flow through the system: even if Swing calls `getIcon()`, it gets a wrapper ... and if that image/icon is passed around the meta-data are present.

Other approach would be to introduce a SPI, that would supply metadata (in general) for an image/icon. ImageUtilities could ensure the SPIs would be called just once/instance (as the images should be immutable anyway). Then wrappers could just vanish from UIDefaultsIconMetadata. The drawback would be that if one instance is shared into multiple keys, the metadata (icon id) provider would just pick one of the keys as an identification - I don't know how much could that be an issue.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: I maybe do not understand the question:  The client receives an Image instance. That instance may be possibly badged from some base Image instance the client has no access to. How could the client compute that the (base) image comes from UIDefault key xxx when it does not know the base image instance, or how the base image has been obtained ?

sdedic: Then there is the Icon > Image conversion that happens sometimes in the infrastructure (i.e. FolderNode) which also needs (somehow) to carry over the ID; in this case, the converting code does not even know that someone will query an ID from the resulting image (or an image badged from that).

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
eirikbakke: > Other approach would be to introduce a SPI, that would supply metadata (in general) for an image/icon
That sounds a bit backwards. "Metadata" here just means "logical icon ID", right? And surely it's the site that actually loads the icon that should provide the logical icon ID to look up by?

By the way, do we have an actual use case for this patch yet, i.e. actual icons that will be substituted in a future patch? Which icons are those, and how many of them are there?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
eirikbakke: > The client receives an Image instance. That instance may be possibly badged from some base Image instance the client has no access to. How could the client compute that the (base) image comes from UIDefault key xxx when it does not know the base image instance, or how the base image has been obtained (maybe even not from UIDefaults) ?

I think this is backwards. By the time you have an Image or an Icon, it's already too late to try to substitute another icon. The substitution needs to happen when you _load_ the Image or the Icon. That's when the code loading the icon should be supplying an ID.

> Then there is the Icon > Image conversion that happens sometimes in the infrastructure (i.e. FolderNode) which also needs (somehow) to carry over the ID

There was a similar problem when I implemented HiDPI icon support in ImageUtilities. The solution was to make sure that ImageUtilities.icon2image and ImageUtilities/image2icon were reversible. So the metadata is carried over.

eirikbakke: OK, so as long as all of this only ever runs in the headless LSP use case, I have a lot fewer concerns. Adding the string id field to ImageUtilities also seems fine, as long as there is no extra instance wrapping in the regular IDE.

> The use case for headless LSP server is different. The icon, for example in our case, resides in a completely other binary/product with a completely different loading and painting mechanism.

So the icons to be replaced are not icons in the NetBeans codebase? Could you give an example of such an icon, perhaps with a screenshot? (A screenshot would make it a lot easier for me to understand what, exactly, the LSP server is...) What are we replacing these icons with?

> The imageID stuff supplies some ID for non-URL images, most notably those from UIDefaults (not loaded by loadImage)

Right, but UIDefaults icons can be replaced by changing the UIDefaults itself, no?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
neilcsmith-net: > Oh, I always assumed that this PR was part of the effort to let icons be branded for LAF purposes. Is there any hope of satisfying both needs with one API change?

@eirikbakke I think so, but not necessarily with this PR.  And the question of keeping the metadata in the icon I can understand given the Swing internals at play too.  API discussions are probably better on dev@

> That sounds a bit backwards. "Metadata" here just means "logical icon ID", right? And surely it's the site that actually loads the icon that should provide the logical icon ID to look up by?

I agree with that.  IMO we need an API whereby icon loading can use a logical ID and request an icon for X.  And an SPI that provides the icons for X.  In the LSP _client_ you'd implement the same. Any such API/SPI could use the existing URLs (file locations) as a backup ID to aid transition.  So both needs would benefit from migrating to a logical ID, as well as being able to query that ID, without any icon being loaded.

eirikbakke: > IMO we need an API whereby icon loading can use a logical ID and request an icon for X. And an SPI that provides the icons for X. In the LSP client you'd implement the same. Any such API/SPI could use the existing URLs (file locations) as a backup ID to aid transition. So both needs would benefit from migrating to a logical ID, as well as being able to query that ID, without any icon being loaded.

Yes, something like this would be exactly what I'd have in mind for this.

> that would require to change UIManager.getIcon|Image to "something different" that would involve that loading SPI (an implementation could do UIManager.get* and return the image/icon with 'logical ID' ) - the change would need to be done throughout the codebase, right ?

I think the existing APIs could be just fine initially. Just keep loading icons by URL; after all, there needs to be a default icon file in the NetBeans codebase for each logical icon ID in any case. But there could indeed be an SPI that ImageUtilities will check to see if a given URL should be resolved to a different Icon instance.

> Good; and what should we do with Swing calling UIManager.get() from its own code ?

For this, just modify the icon in UIDefaults, like Swing intended. We already do this e.g. from Windows8LFCustoms.

neilcsmith-net: @sdedic I somewhat suspect there will be fallbacks very much along the lines of what you do here too! I only mentioned in the first place so we sync up conversations on the various uses for IDs / don't lock its spec to something that can't meet more needs in future.

sdedic:
FYI - I removed NB13 milestone from here, bcs. of pending discussion. I created #3459 which does not introduce new things to API (I've retained the URL property, it's there and using just "url" string is not well traceable)

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
neilcsmith-net: > The use case for **headless** LSP server is different. 

My general point is that the use case for the headless LSP server is different to what we have now, but that the IDE would benefit from fixing the process in a similar way.

> For Images obtained as `ImageUtilities.loadImage` it's rather easy as there's the URL (which can be mangled by localization etc, but that can be handled).

And the URL has _been_mangled to address branding, and dark variants, and sizes, and hidpi, and svg conversion ... Some of which interfere badly with each other, and spread files (some duplicates IIRC) all over the code base.  For the purposes of loading an _icon_ at least, there are benefits to not mangling based on URL in the first place.

Chris2011:
I would prefer to add a comment why PRs or Issues where closed. If a short description is possible, that would be nice :)

PR #3417, opened 2022-01-05
Native image breakpoints work with shared libraries.

entlicher:
When submitting a breakpoint into a Java native image, we consult the list of source files from the executable to determine the correct destination file.

The breakpoint's file is resolved relative to the `sources` folder, which is expected to be the sibling of the executable file.
When the `sources` folder is not found, breakpoints are resolved to locations that do not exist in the executable.
To verify the correctness of file paths, we use a newly introduced `NIDebugger.listSources()` method.

### DIFF
--- a/java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/breakpoints/JPDABreakpointsHandler.java
+++ b/java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/breakpoints/JPDABreakpointsHandler.java
@@ -46,6 +46,8 @@ import org.openide.filesystems.URLMapper;
  */
 public class JPDABreakpointsHandler extends DebuggerManagerAdapter implements PropertyChangeListener {
 
+    private static final String SOURCES_FOLDER = "sources"; // NOI18N
+
     private final File niFileSources;
     private final NIDebugger debugger;
     private final Set<JPDABreakpoint> attachedBreakpoints = new HashSet<>();
@@ -62,7 +64,7 @@ public class JPDABreakpointsHandler extends DebuggerManagerAdapter implements Pr
     }
 
     private static File getNativeSources(File niFile) {
-        File sources = new File(niFile.getParentFile(), "sources");
+        File sources = new File(niFile.getParentFile(), SOURCES_FOLDER);
         if (sources.isDirectory()) {
             return sources;
         } else {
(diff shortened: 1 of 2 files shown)


JaroslavTulach:
I've just tried a version of the `.vsix` extension prepared by Martin and it kinda works: If my `libxyz.so` is loaded, I can attach, place a breakpoint to it and it is hit!

If I first attach, then load my dynamic library, the breakpoint doesn't work. But I can live with it.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/spi/debug/NIDebuggerProvider.java
```diff
@@ -194,4 +194,13 @@
     default Map<SourceInfo, List<Symbol>> listVariables(String name, boolean includeNondebug, int maxResults) {
         return null;
     }
+
+    /**
+     * Provide a list of source paths in the debuggee.
+     *
+     * @since 0.7
+     */
+    default List<SourceInfo> listSources(String nameRegexp) {
```
JaroslavTulach, line 203: A note in API changes would be good. Specify that the method can return `null`. State the `List` is immutable.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
```diff
@@ -257,6 +257,16 @@ public String getVersion() {
...
     }
 
+    /**
+     * Get a list of source paths in the debuggee.
+     *
+     * @param nameRegexp regular expression that filters the list of source paths
+     * @since 0.7
+     */
+    @CheckForNull
+    public List<SourceInfo> listSources(String nameRegexp) {
```
JaroslavTulach, line 267: A note in `apichanges.xml` would be good. This method shall never return `null`, rather empty list - unless you want to use `null` for some use-case. In such case mention that in Javadoc.

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/breakpoints/JPDABreakpointsHandler.java
```diff
@@ -154,6 +159,25 @@ private Breakpoint added(JPDABreakpoint b) {
         return null;
     }
 
+    private String verifyFilePath(String filePath, String relativePath) {
+        if (relativePath != null) {
+            List<SourceInfo> sources = debugger.listSources(SOURCES_FOLDER + File.separator + relativePath);
```
JaroslavTulach, line 164: `listSources` is supposed to take RegExp - what will it do on Windows? `\` is an escape character...

JaroslavTulach (approved):
The (necessary) API changes are a little price for getting the debugging working. You have my go!

java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/breakpoints/JPDABreakpointsHandler.java
entlicher: Good point.
After all I've abandoned matching file paths in the sources listing as we can not have reliable source list for lazily loaded shared libraries.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/spi/debug/NIDebuggerProvider.java
entlicher: The API changes were abandoned for now.

ide/nativeimage.api/src/org/netbeans/modules/nativeimage/api/debug/NIDebugger.java
entlicher: The API changes were abandoned for now.

entlicher:
After all I've abandoned matching file paths in the sources listing as we can not have reliable source list for lazily loaded shared libraries.
The fix reduces to breakpoints submission based on relative paths with respect to the `sources` folder and using `-f` option to create pending breakpoints when the location is unknown.

JaroslavTulach:
The newest version seems to work for my use-case.

PR #3418, opened 2022-01-06
[NETBEANS-6285] Shield against usage of JDK-11 API.

JaroslavTulach:
Since #2761 we use `--release 8` consistently to compile NetBeans code except modules that need to change `-bootclasspath`. By specifying `-bootclasspath` to `javac` one disables `--release` and uses the actual API of the underlaying JDK. As we currently compile with JDK-11 that means API of JDK-11 and as [NETBEANS-6285](https://issues.apache.org/jira/browse/NETBEANS-6285) and its many duplicates show, it can cause problems.

The best way to shield us against that is to build on JDK-11 and test (at least something) on JDK-8. This PR adds new GitHub Action to do so. It [fails](https://github.com/JaroslavTulach/netbeans/runs/4726963523) on the CI as well as  locally:

tests failing on jdk-8

Now we can fix the problem as outlined in [NETBEANS-6349](https://issues.apache.org/jira/browse/NETBEANS-6349).

### DIFF
--- a/.github/workflows/ensure-jdk8.yml
+++ b/.github/workflows/ensure-jdk8.yml
@@ -21,29 +21,14 @@ on:
   push:
   pull_request:
 
-jobs: 
-  jdk11-jdk8: 
+jobs:
+  jdk11-jdk8:
     name: "Compile with JDK-11 and test something on JDK-8"
     runs-on: ubuntu-18.04
     env:
       OPTS: -Dcluster.config=basic -Dtest-unit-sys-prop.ignore.random.failures=true
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK 8
-        uses: actions/setup-java@v2
-        with:
-            distribution: 'zulu'
-            java-version: 8
-      - name: Record JDK8
-        run: echo "$JAVA_HOME" | tee jdk8
-      - name: Set up JDK 11
-        uses: actions/setup-java@v2
-        with:
-            distribution: 'zulu'
-            java-version: 11
-
-      - name: Checkout
-        uses: actions/checkout@v2
 
       - name: Caching dependencies
         uses: actions/cache@v2
@@ -51,6 +36,22 @@ jobs:
           path: ~/.hgexternalcache
           key: ${{ runner.os }}-${{ hashFiles('**/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v2
+        with:
+            distribution: 'zulu'
+            java-version: 8
+
+      - name: Record JDK8
+        run: echo "$JAVA_HOME" | tee `pwd`/jdk8
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+            distribution: 'zulu'
+            java-version: 11
+
 
       - name: Build
         run: ant $OPTS build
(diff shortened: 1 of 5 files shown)


.github/workflows/ensure-jdk8.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/setup-java@v2
+        with:
+            distribution: 'zulu'
+            java-version: 11
+
+
+      - name: Build
+        run: ant $OPTS build
+
+      - name: Test java.source.base on JDK-8
```
JaroslavTulach, line 59: The future plan is: Once we discover another module that has this kind of compilation problem, we should add another test run here by duplicating line 59 and 60.

mbien:
I like this. But tbh I am a bit surprised that we are actually building release builds on JDK 11. Since we can't put target to 11, I am not sure what the benefit is to build on 11 (beside validating that the build works which is a good thing).

As long NB has the requirement to run on 8, we should build on 8 and test the build everywhere else (8, 11, 17, current) [1].

This PR is a good safety measure, but this should not have happened in the first place IMO.

[1] rudimentary example of gh actions i once used for debugging, builds on 8, uploads the artifact and tests everywhere else: https://github.com/mbien/netbeans/blob/d61f3b6ccf0e92e7f570fc912608a455186813c8/.github/workflows/main.yml#L37

mbien:
I vaguely remember that there are optional modules which only build on 11+ (mac related?). Is that the reason?

neilcsmith-net:
@mbien the idea is to allow more modules to adopt 11. The whole of NetBeans doesn't have a requirement to run on 8 any more, but a lot of parts do. This was discussed on dev@ as was the choice to build the 12.6 convenience binaries on JDK 11 to make sure our build processes were ready.  I'm surprised this particular bug got through user testing, which in some ways shows we're not getting anywhere near enough of that on JDK 8 to keep supporting it.

It's worth remembering that ASF only releases sources!  And the source release is still JDK 8 compatible.  Doing an extra binary zip build of 12.6 with JDK 8 is an option if anyone wants to do it.

JaroslavTulach:
The [newest run](https://github.com/apache/netbeans/runs/4727134768) still fails, but that will need an adjustment of the test, I guess:
```
     [junit] Testcase: testGetTypeDeclaration(org.netbeans.modules.java.source.parsing.ClasspathInfoTest):	FAILED
    [junit] Should be empty [sun.security.provider.Sun$1, sun.security.rsa.SunRsaSign$1]
    [junit] junit.framework.AssertionFailedError: Should be empty [sun.security.provider.Sun$1, sun.security.rsa.SunRsaSign$1]
    [junit] 	at org.netbeans.modules.java.source.parsing.ClasspathInfoTest.testGetTypeDeclaration(ClasspathInfoTest.java:134)
    [junit] 	at org.netbeans.junit.NbTestCase.access$200(NbTestCase.java:77)
    [junit] 	at org.netbeans.junit.NbTestCase$2.doSomething(NbTestCase.java:476)
    [junit] 	at org.netbeans.junit.NbTestCase$1Guard.run(NbTestCase.java:402)
    [junit] 	at java.lang.Thread.run(Thread.java:748)
```

.github/workflows/ensure-jdk8.yml
neilcsmith-net: @JaroslavTulach wouldn't a future plan of building on JDK 11, testing on JDK 8, for all JDK 8 tests be a better option?  At this point, that seems the more important combination to test.  If capacity an issue, could run a build on JDK 8 test on a regular basis, but not per PR?

sdedic (approved):

matthiasblaesing (approved):
The idea sounds sane to me and I'm fine with the change as-is. It might be worth to consider using [animal-sniffer](https://www.mojohaus.org/animal-sniffer/) for problematic modules. The JNA project uses it to ensure compatibility with Java 1.6.

mbien (approved):

PR #3419, opened 2022-01-06
[cnd] small-1.5 More cnd tests

vieiro:
This adds more passing tests to the cnd cluster (cnd.remote, cnd, cnd.toolchain)

cnd.remote depended on "cnd.model.api", which we have not integrated in Apache NetBeans (to be replaced with LSP). Tests that required validating the model 
- FullRemoteCodeModelTestCase, 
- RemoteCodeModelTestCase, 
- FullRemoteBuildTestCase, 
- RemoteBuildTestBase
are modified accordingly dropping this dependency.

After this the cnd cluster should test properly (dlight cluster requires some extra care, though).

### DIFF
--- a/cnd/cnd.remote/nbproject/project.xml
+++ b/cnd/cnd.remote/nbproject/project.xml
@@ -225,10 +225,6 @@
                         <test/>
                     </test-dependency>
                     <test-dependency>
-                        <code-name-base>org.netbeans.modules.cnd.api.model</code-name-base>
-                        <compile-dependency/>
-                    </test-dependency>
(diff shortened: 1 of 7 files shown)


matthiasblaesing (approved):
Looks ok to me, the only question I have is regarding the removed tests: Might it make sense to comment them out to restore them if a model implemention is later added? This question is without any bias, as I really no idea how the CND cluster works.

vieiro:
Who knows, maybe somebody adds a model impl later on.  Let's restore those tests, commented out, and let's move on.

PR #3420, opened 2022-01-06
LSP console IO is made more reliable, output is not lost now.

entlicher:
The console is closed only when both standard and error outputs are closed.
Also a correct stream is referred in `BridgingIOProvider`.


sdedic (approved):
Thanks !

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspInputOutputProvider.java
```diff
@@ -165,8 +166,13 @@ public void close() {
             this.in = in;
         }
 
+        boolean isClosed() {
+            return out.closed && err.closed;
+        }
+
         private final class LspWriter extends Writer {
             private final boolean stdIO;
+            volatile boolean closed;
```
JaroslavTulach, line 175: I wanted to ask: "who sets this field?" as that isn't apparent from this diff. However there already is `close()` method in this class which was setting the outer class `closed` - by moving it, it now sets this field. OK.

PR #3424, opened 2022-01-07
LSP: Fixed broken code completion for super constructor call.

dbalek:
Fixing broken code completion for super constructor call.


jtulach (approved):
A test'd make sure the behavior won't regress.

sdedic (approved):

mbien:
this doesn't seem to be LSP specific. What did it fix? Is there an example for what didn't work before and what works now?

PR #3427, opened 2022-01-08
Convert source path to URI String.

entlicher:
The Path can be treated as an URI on UNIX, but not on Windows. Paths starting with the drive letter are not valid URIs, this is why we need to convert the Path to URI and then to String to get a valid URI String representation.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbSourceProvider.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/NbSourceProvider.java
@@ -125,7 +125,7 @@ public final class NbSourceProvider {
                 for (String path : sourcePaths) {
                     Path fullpath = Paths.get(path, relativeSourcePath);
                     if (Files.isRegularFile(fullpath)) {
-                        uri = fullpath.toString();
+                        uri = fullpath.toUri().toString();
                         break;
                     }
                 }


sdedic (approved):

mbien:
`getSourceFileURI` should probably return an actual URI. This forces everything in this class to use URIs instead of Strings and this won't happen again easily. 

`fullpath.toUri().toString()` might look like an inefficiency for some which could reintroduce the bug again.

entlicher:
> `getSourceFileURI` should probably return an actual URI. This forces everything in this class to use URIs instead of Strings and this won't happen again easily.
> 
> `fullpath.toUri().toString()` might look like an inefficiency for some which could reintroduce the bug again.

Well, we implement LSP and we need a String at the end. There's no URI in org.eclipse.lsp4j.debug.Source.

mbien:
> > `getSourceFileURI` should probably return an actual URI. This forces everything in this class to use URIs instead of Strings and this won't happen again easily.
> > `fullpath.toUri().toString()` might look like an inefficiency for some which could reintroduce the bug again.
> 
> Well, we implement LSP and we need a String at the end. There's no URI in org.eclipse.lsp4j.debug.Source.

that is not the point. The utility methods could work with URIs and return a string as last step, instead of working with strings representing URIs.

entlicher:
> that is not the point. The utility methods could work with URIs and return a string as last step, instead of working with strings representing URIs.

Looking into the code, yes, it looks like an improvement.
This was a hot fix for VSCode extension, thus I had to merge soon and with just the necessary changes for stability reasons, but I can do the change to URIs as a new PR, yes.

PR #3428, opened 2022-01-08
[NETBEANS-6372] [cnd] small-2.2 cnd+cpplite integration

vieiro:
Both cpplite and cnd declare:

- An "Options/C C++" entry.
- A "Template/Projects/Native" folder.

in the "layer.xml" generated file. This generates failures when one runs "ant commit-validation", as reported in https://issues.apache.org/jira/browse/NETBEANS-6372

This PR:
- Makes cpplite and cnd publish the "Template/Projects/Native" folder with the same "position" attribute.
- Adds a friend dependency "cpplite editor" -> "cnd.editor", and "cnd.editor" is the responsible for registering the "Options/C C++" dialog. We could use a "weight" attribute to solve the situation and avoid this dependency, as stated in https://bz.apache.org/netbeans/show_bug.cgi?id=141925 but sadly the annotation that register the option pane does not support the weight attribute.
- Adds both "cnd" and "dlight" to the "release" cluster, in order to validate the "commit-validation" test in our pipelines.


vieiro:
NOTE: The "Commit-validation" checks int this PR are expected to **pass**.

vieiro:
NOTE-II: We should think if we want cpplite depend on cnd or the other way around. One cannot live without the other. Alternatively, we could also integrate cpplite in cnd, for instance.

vieiro:
Added a commit that replaces a dependency with sun.nio in dlight.libs.common replacing it with java.nio, as detected in the Travis build.

vieiro:
The way to proceed with this integration is being discussed in the dev mailing list.

PR #3435, opened 2022-01-10
Always call mx build before running the java.mx.projects tests

JaroslavTulach:
@jlahoda reported that the `ant -f java/java.mx.projects test` is failing on subsequent executions - because the `mx build` is executed on 1st run and not then. This PR changes the build script to invoke `mx build` everytime `ant test` is invoked.


sdedic (approved):
Understood: checkout just when needed, don't block test-preinit.

dbalek (approved):

PR #3438, opened 2022-01-10
LSP: Modified to fix errors currently printed to log from ServerTest.

dbalek:
Couple of modifications to fix errors currently printed to log from ServerTest.


JaroslavTulach:
Why shouldn't errors be printed to the log? How are we going to analyze them if they are swallowed?

dbalek:
> Why shouldn't errors be printed to the log? How are we going to analyze them if they are swallowed?

Sorry. Wrong wording, errors currently printed to the log are fixed.

JaroslavTulach (approved):

PR #3439, opened 2022-01-10
[jackpot] Add sourceVersion(int) to rule file utils and deprecate enum variant. (part 2)

mbien:
I forgot to add a commit to https://github.com/apache/netbeans/pull/3395

This is the actual support of int literals for rule conditions - the rule API of part 1 alone won't do anything.

sorry!


jlahoda (approved):
My mistake I didn't remember there are limitations on the conditions.

mbien:
@jlahoda thanks for review, merging

PR #3440, opened 2022-01-11
Add default hint for switch pattern matching error

mishrasandeep:


mbien:
what shall we do with this PR? Hasn't changed since it was set to draft.

PR #3441, opened 2022-01-11
Simplified cancelHandle test.

sdedic:
An attempt to fix failing `ServerTest` from lsp on Travis CI, job `Test Java modules with nb-javac on Java 8`.

I've found out that the test probably catches a wrong process: at least on local, there was also indexing and some other processes going on, so the captured cancellation token might represent a different task than the compilation one.

I've rewritten the test so it does not rely on Maven, the mocked LSP client captures just the specific command's progress. The test now verifies that a cancel instruction from the LSP client will reach the appropriate `Cancellable` instance in the LSP server  - but nothing more; the previous version could eventually verify that a build can be cancelled.

### DIFF
--- a/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
+++ b/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
@@ -36,7 +36,6 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -59,6 +58,7 @@ import javax.swing.Icon;
 import javax.swing.event.ChangeListener;
 import javax.swing.text.Document;
 import javax.swing.text.StyledDocument;
+import static junit.framework.TestCase.assertEquals;
 import static junit.framework.TestCase.assertNotNull;
 import static junit.framework.TestCase.assertNull;
 import org.eclipse.lsp4j.ApplyWorkspaceEditParams;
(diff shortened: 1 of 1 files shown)


sdedic:
Hm, fails again ... so I'll use this PR as a vehicle to run modified stuff on CI :(

sdedic:
I know, but the test has a meaning - it checks that a LSP client is able to cancel an operation that uses `Progress API` in NBLS - for example a pending build :)

mbien (approved):
looks good to me - thanks for getting this resolved. This might be one of the rare cases where a Phaser could be used - but two CountDownLatches are just as readable.

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -4846,8 +4849,85 @@ public void saveProject(Project project) throws IOException, ClassCastException
...
+
+            return CompletableFuture.<Object>supplyAsync(() -> {
+                ProgressHandle h = ProgressHandle.createHandle("Test Command", this::cancel);
+                try {
+                    h.start(100, 100);
+                    h.progress(20);
+                    // The Progress infra sometimes coalesces events, so definitely deliver begin + progress before
+                    // waking up the client controller that will issue cancel. Otherwise begin / cancel could be merged into
+                    // a no-op.
+                    Thread.sleep(1000);
```
JaroslavTulach, line 4902: `Thread.sleep` is just a request for random failures (under load, for example). Can't the test verify an event has been delivered via the `ProgressHandle` infrastructure somehow?

JaroslavTulach (approved):

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
sdedic: No :( or not in a way I am aware of: events may be coalesced right in the loop, the even queue is private, its RequestProcessor as well. And there's also transport from the server > client which is done over sockets = at least one another thread hand-off, so even I could eventually patch the `Controller` in Progress module and watch when the event queue empties, the event could be still partially somewhere between LSP Server and Client I guess.

PR #3442, opened 2022-01-11
LSP: Change method parameters should refactor access modifier.

dbalek:
Change method parameters should refactor access modifier and reflect static methods.


JaroslavTulach (approved):
It is easy to write unit tests for the `@Model` classes, btw.

sdedic (approved):

PR #3443, opened 2022-01-11
Use boolean flag to create apidoc menu

ebarboni:
attempt to make apidoc generation according to a flag instead of url.
remove apidoc prior to 8.2 link

will need to recommit on the previous release branch.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
@@ -115,8 +115,6 @@ public class ReleaseJsonProperties extends Task {
             throw new BuildException(ex);
         }
 
-        // remove empty api doc 
-        //ri.removeIf(e -> e.apidocurl.isEmpty());
         // sort all information
         Collections.sort(ri);
         // build a sorted xml
@@ -140,7 +138,8 @@ public class ReleaseJsonProperties extends Task {
         }
         List<String> updateValues = new ArrayList<>();
         for (ReleaseInfo releaseInfo : ri) {
-            if (releaseInfo.position < requiredbranchinfo.position) {
+            // take previous version of Apache NetBeans only if published, need for scan for old NetBeans version
+            if (releaseInfo.position < requiredbranchinfo.position && releaseInfo.publishapi ) {
                 updateValues.add(releaseInfo.version);
             }
         }
--- a/nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
@@ -114,9 +114,9 @@ public class ReleaseJsonProperties extends Task {
         } catch (ParseException | IOException ex) {
             throw new BuildException(ex);
         }
-        
+
         // remove empty api doc 
-        ri.removeIf(e -> e.apidocurl.isEmpty());
+        //ri.removeIf(e -> e.apidocurl.isEmpty());
         // sort all information
         Collections.sort(ri);
         // build a sorted xml
@@ -194,7 +194,7 @@ public class ReleaseJsonProperties extends Task {
             throw new BuildException("Properties File for release cannot be created");
         }
 
-        log("Writing releasinfo file " + xmlFile);
+        log("Writing release info file " + xmlFile);
 
         xmlFile.getParentFile().mkdirs();
         try (OutputStream config = new FileOutputStream(xmlFile)) {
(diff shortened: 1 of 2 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
```diff
@@ -114,9 +114,9 @@ public void execute() throws BuildException {
         } catch (ParseException | IOException ex) {
             throw new BuildException(ex);
         }
-        
+
         // remove empty api doc 
-        ri.removeIf(e -> e.apidocurl.isEmpty());
+        //ri.removeIf(e -> e.apidocurl.isEmpty());
```
neilcsmith-net, line 119: I just spent 15min looking for the build issue that @JaroslavTulach had with older VSNetBeans release to realise it's already covered by this change! :smile:

neilcsmith-net (approved):
Looks good. And allowing apidocurl, etc. to be unspecified should make it easier for VSNetBeans to use its own json file and leave out a bunch of stuff it doesn't need?!

@JaroslavTulach please can we get a `vsnetbeansrelease.json` for the upcoming VSNetBeans release?  Will simplify things with out of band releases. Should just be a case of overriding the default json file property somewhere in the build files.

JaroslavTulach:
> @JaroslavTulach please can we get a `vsnetbeansrelease.json` for the upcoming VSNetBeans release? Will simplify things with out of band releases. Should just be a case of overriding the default json file property somewhere in the build files.

This is the [current build configuration](https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/configure) for VSNetBeans:
```
ant build-source-config -Dmetabuild.branch=$METABRANCH -Dmetabuild.jsonurl=$METAURL
ant build -Ddo.build.windows.launchers=true -Dmetabuild.branch=$METABRANCH -Dmetabuild.jsonurl=$METAURL

cd java/java.lsp.server/
ant build-vscode-ext -Dvsix.version=$VERSION -D3rdparty.modules=$MODULES -Dmetabuild.jsonurl=$METAURL
```
I can certainly change `$METAURL` to point to something else, but I don't know which script you'd like me to modify!?

neilcsmith-net:
> I can certainly change $METAURL to point to something else, but I don't know which script you'd like me to modify!?

Yes, looking at that I guess changing $METAURL will have to do. I was thinking there was an Ant file inside `java/java.lsp.server` that was triggering the IDE build - my mistake.

neilcsmith-net:
@ebarboni _if_ we keep the VSNetBeans releases in the same json file, a flag to pick up what is and isn't an IDE release might be useful for this and other problems.  On the other hand, that would still involve VSNetBeans having to configure the build differently, so still probably better as a separate file IMO.

nbbuild/antsrc/org/netbeans/nbbuild/ReleaseJsonProperties.java
neilcsmith-net: @JaroslavTulach normally I'd be the first to agree with you on data duplication, but as also mentioned there, we're losing the canonical record of Apache IDE releases in that file, which is also important because it's used elsewhere. Let's keep conversation on that there.

ebarboni:
this pr is more for cleaning up apidoc once done on other release branches,

If we add some flag on the json, (non impacting the netbeans repo) we may also have for example in the release branches
https://ci-builds.apache.org/job/Netbeans/job/netbeans-TLP/job/netbeans/job/release126/14/artifact/
containing vsix files to get ready for release for normale release and with some works handling for the vsrelease1263 so we pick as we pick maven, installer, nbm during release

Will have impact on https://github.com/apache/netbeans-jenkins-lib/blob/master/vars/asfMainNetBeansBuild.groovy that is old and can be upgraded to fit

neilcsmith-net:
Yes, I understand that.  Still all easier with separate JSON files.  There's the other stuff on the VM that reads the JSON file too - I just realised I don't think you had the current `update-netbeans-javadoc.php` in the puppet PR?

neilcsmith-net:
Well, from 13 we'll have X (and X.301 potentially).  I'm still curious to see what that breaks! :smile:

That 404 took out the entire Javadoc publishing process if I remember rightly.

I don't know what the other places we use the json are, or what we might need in future - that's my concern.  And if we start filtering for this flag inside the build, then VSNetBeans will still have to special case it.

ebarboni:
version X worked for 10, just the apidoc/update catalog need the X.0

neilcsmith-net:
Thanks!  Let's see how we get on with this.

Let's take the flags discussion elsewhere.  Having thought on this, I think if we do keep in one file, we need two truth-y values for `is_ide` and `is_vsix` rather than a false-y one `vsix-only`. To fix some of the issues we're likely to need to filter the list both ways, and some are IDE only.  Or we add a `category` array with `ide` and `vsix` - future proof for the next thing someone wants in here! :smile:

PR #3444, opened 2022-01-11
Update tests to build on JDK 11 (WIP)

neilcsmith-net:
Updating tests to build using JDK 11.

Work-in-progress - GitHub Actions first, Travis to follow.

The PHP Editor test FoldingTest::testPHPTags seems flaky on my personal repo, and definitely locally - let's see what happens!

### DIFF
--- a/.github/workflows/cv.yml
+++ b/.github/workflows/cv.yml
@@ -24,7 +24,7 @@ on:
 jobs:
   commit-validation:
     name: Commit-validation (on Linux)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     env:
       OPTS: -Dcluster.config=release -Dtest-unit-sys-prop.ignore.random.failures=true
     steps:
@@ -37,6 +37,12 @@ jobs:
           path: ~/.hgexternalcache
           key: ${{ runner.os }}-${{ hashFiles('**/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+            distribution: 'zulu'
+            java-version: 11
 
       - name: Clean
         run: ant -Dcluster.config=release clean
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,7 +24,7 @@ on:
 jobs:
   linux:
     name: Check Linux
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     env:
       ANT_OPTS: -Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json
     steps:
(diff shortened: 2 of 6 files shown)


.github/workflows/cv.yml
```diff
@@ -24,7 +24,7 @@ on:
 jobs:
   commit-validation:
     name: Commit-validation (on Linux)
```
mbien, line 26: nit: maybe we can rename it to "Commit-validation (Linux, JDK 11)" (same for the other jobs)

.github/workflows/linux.yml
```diff
@@ -24,7 +24,7 @@ on:
 jobs:
   linux:
     name: Check Linux
```
mbien, line 26: this would deserve a more descriptive name too. I am not even sure what it tests. The build harness?

mbien: if it doesn't test NB itself, this job might be able run on JDK 17, since the basic cluster should build ok by now.

mbien:
> Work-in-progress - GitHub Actions first, Travis to follow.

@jlahoda made some updates and version bumps to travis in this PR: https://github.com/apache/netbeans/pull/3342 it unfortunately never made it in.

neilcsmith-net:
Thanks @mbien Minimal changes so far to see what worked and didn't - will look at some more useful names!  Will check out the mentioned PR too.

neilcsmith-net:
@junichi11 do you have any idea why the FoldingTest fails here?  It may be due to Ubuntu 20.04?  Ant version?  It fails locally for me with JDK 8 too.

Also, is there anything in PHP editor we could run daily and/or if PR touches particular paths?

junichi11:
@neilcsmith-net 

> do you have any idea why the FoldingTest fails here? It may be due to Ubuntu 20.04? Ant version? It fails locally for me with JDK 8 too.

Perhaps, it is related to the following line, I suppose...
https://github.com/apache/netbeans/blob/ae6389a98e30c459435800920a57363351dfd543/php/php.editor/src/org/netbeans/modules/php/editor/csl/FoldingScanner.java#L134

We expect that `FOLD_PHPTAG` is `true` by default.

https://github.com/apache/netbeans/blob/ae6389a98e30c459435800920a57363351dfd543/php/php.editor/src/org/netbeans/modules/php/editor/csl/FoldingScanner.java#L174-L176

I'm using Ubuntu 20.04
Product Version: Apache NetBeans IDE 12.0
Java: 1.8.0_202; Java HotSpot(TM) 64-Bit Server VM 25.202-b08
Runtime: Java(TM) SE Runtime Environment 1.8.0_202-b08
System: Linux version 5.11.0-46-generic running on amd64; UTF-8;

The test passed in my PC.

> Also, is there anything in PHP editor we could run daily and/or if PR touches particular paths?

Other than unit tests? Umm, I'm not sure...

neilcsmith-net:
Thanks @junichi11 Been a bit distracted by release work to follow up.  Will check that out.

junichi11:
@neilcsmith-net Can we add `-J-Dnb.php.editor.doNotFoldPhptag=false` as an option?

neilcsmith-net:
Closing in favour of @mbien work on #3785 - got waylaid by release work and became obvious full building on JDK 11 for tests wasn't doable for NB13.

PR #3449, opened 2022-01-12
[NETBEANS-6347] Revert "UI support for disabled actions."

neilcsmith-net:
https://issues.apache.org/jira/browse/NETBEANS-6347

This reverts commit a401cec52837866eaf67e7a60bddd71da89730a0.

This commit appears to be the root cause of NETBEANS-6347 - adding global and project custom actions is completely broken in Maven.

I did ask about this on #3016 I'm not sure what the desired effect in the UI is.  I looked into trying to keep the renderer and disabled button, but the renderer needs adapting to handle empty added actions, and the disabled button doesn't seem to be hooked up to anything?!

Think we need to revert and rethink, as it's causing major issues for users.  cc @sdedic

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.form
+++ b/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.form
@@ -32,7 +32,7 @@
     <AuxValue name="FormSettings_listenerGenerationStyle" type="java.lang.Integer" value="0"/>
     <AuxValue name="FormSettings_variablesLocal" type="java.lang.Boolean" value="false"/>
     <AuxValue name="FormSettings_variablesModifier" type="java.lang.Integer" value="2"/>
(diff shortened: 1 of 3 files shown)


sdedic:
Please hold on with this one till Friday, I have the bug on my list but needed to finish with vscode-ext first.

neilcsmith-net:
Sure! Will wait until branching on Monday latest. I did my best not to revert!  But it was getting to the stage I'd pulled out almost all the changes anyway.  Think it needs a separation of no-goals / custom actions from disabled in this particular context.

sdedic:
OK, I believe I have a quick fix that will allow to disable (standard) actions; a proper fix requires a small extension to the ActionMapping definition, as the `disabled` state is now under-represented (as an absence of any goal). So it both complicates interaction with new actions and also prevents from disable-enable when one would expect the action retains the former definition, but it does not (goals are cleared). 
Similar with Gradle. I would track that as another bug + attempt to fix for the next release (or backport if the result would not be that disruptive).

neilcsmith-net:
Closing in favour of #3454

PR #3452, opened 2022-01-13
When OOME situation is detected during indexing, make sure it is consistently handled as OOME; enable the Truffle source test, and give it consistent amount of memory.

jlahoda:


mbien (approved):
looks good to me

sdedic:
@jlahoda still relevant ?

PR #3454, opened 2022-01-13
NETBEANS-6347: Disabled user actions fixed

sdedic:
Actions without any goals are identified as disabled. That means that any unfinished, but saved or freshly created actions become disabled. This PR fixes that by assuming that all user-defined actions are enabled always. [NETBEANS-6387](https://issues.apache.org/jira/browse/NETBEANS-6387) was filed for further improvement / consistency between Maven and Gradle.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.form
+++ b/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.form
@@ -355,6 +355,9 @@
         </Property>
         <Property name="name" type="java.lang.String" value="" noResource="true"/>
       </Properties>
+      <Events>
+        <EventHandler event="actionPerformed" listener="java.awt.event.ActionListener" parameters="java.awt.event.ActionEvent" handler="btnDisableActionPerformed"/>
+      </Events>
       <Constraints>
         <Constraint layoutClass="org.netbeans.modules.form.compat2.layouts.DesignGridBagLayout" value="org.netbeans.modules.form.compat2.layouts.DesignGridBagLayout$GridBagConstraintsDescription">
           <GridBagConstraints gridX="10" gridY="6" gridWidth="1" gridHeight="1" fill="2" ipadX="0" ipadY="0" insetsTop="6" insetsLeft="6" insetsBottom="0" insetsRight="12" anchor="18" weightX="0.0" weightY="0.0"/>
--- a/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.java
+++ b/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.java
@@ -580,6 +580,11 @@ public class ActionMappings extends javax.swing.JPanel implements HelpCtx.Provid
 
         org.openide.awt.Mnemonics.setLocalizedText(btnDisable, org.openide.util.NbBundle.getMessage(ActionMappings.class, "ActionMappings.btnDisable.text")); // NOI18N
         btnDisable.setName(""); // NOI18N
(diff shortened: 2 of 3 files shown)


neilcsmith-net:
Will review ASAP.  This seems similar to something I tried, before opening the revert PR, that still seemed to have some issues.

sdedic:
Note: the commit d36d263 is unrelated, I've just noticed that my last changes to the travis script leave 'XPath is empty' messages in the log sometimes.

neilcsmith-net:
I'm -0.5 on this.  This is almost what I tried too, but I still prefer what we had before.  Although it does fix the critical issue with user actions.

Having actions disabled elsewhere in the UI when there are no goals makes sense.  Just less sure here.  I'm not sure what the purpose of the manual `Disable` button is exactly?  Certainly having to click the button labelled with `Remove` in order to re-enable an action seems counter-intuitive.  Can this not be controlled by the existence of goals for now, rather than disabling that input field entirely?

The various labels turning bold (`Execute Goals`, `Activate Profiles`, etc.) when the disabled action is selected makes everything bounce around, although I think that was a bug before with custom actions.  

Aside : I set up a Micronaut project to test this, but the dev mode profile isn't showing up in a current build vs 12.6??  Same without this PR, so not related, but was trying to see how this interacts with that.  That mvnw isn't executable also breaks the Micronaut wizard (which I should probably look at!)

neilcsmith-net:
The counterintuitive bit is not that the disable button clears the goals, but that that disables the input fields. Keep the input fields usable and I'm all for it.

sdedic:
> The counterintuitive bit is not that the disable button clears the goals, but that that disables the input fields. Keep the input fields usable and I'm all for it.

Well, OK - but if I leave the input(s) enabled, the user edits them ... then what: the action becomes enabled again; so the Disable really becomes just a shortcut to clear the goals.

neilcsmith-net:
> Well, OK - but if I leave the input(s) enabled, the user edits them ... then what: the action becomes enabled again; so the Disable really becomes just a shortcut to clear the goals.

But it already is?! Until there's a way to differentiate between disabled and _temporarily_ without goals (which needs something like NETBEANS-6387) then this doesn't work. It broke the custom actions, but it also breaks if you move from one action to another (eg. to copy something across) and the goals field is empty.

It doesn't actually look like it is possible to get the Micronaut disabled actions to work again anyway (personally I think being able to enable _something_ with the debug action isn't a bad thing, while noting why the default isn't useful).  Having a lax UI is better than having one that is over strict however.

sdedic:
OK, so to recap (bcs I'd strongly prefer to leave the disable actions support there) - if I change the appearance to allow editing, perhaps remove `Delete` bcs of its unclear / duplicate semantics - just gray out the action without goals (as it would not work anyway if executed): would that work ?

neilcsmith-net:
@sdedic sure!  Or I can?  Not trying to make work for you.  Like I said, I had almost identical code to this PR at one point yesterday.  It's just altering (well reverting) the logic of when this applies now? https://github.com/apache/netbeans/blob/d36d26302a5ef5316d1b07bac4f44f776ee27dda/java/maven/src/org/netbeans/modules/maven/customizer/ActionMappings.java#L678-L695

sdedic:
OK, more or less reverted + some minor UI improvements: slightly more space for the properties (not for the hint area); changing goals to (non)empty will redraw list item.

neilcsmith-net (approved):
Thanks! Let's get this in for 13 and on to proper action disablement in 14.

PR #3455, opened 2022-01-13
[NETBEANS-6388] remove findbugs installer module.

mbien:
- causes trouble in the source -> inspect dialog
- findbugs is EOL (spotbugs is the successor)

https://issues.apache.org/jira/browse/NETBEANS-6388


mbien:
I didn't remove the findbugs module entirely, because 1) removing things makes @JaroslavTulach angry :) and 2) someone might want to repurpose it for "spotbugs".

Also please note that this doesn't fix actual bug. Because the dialog should not require the findbugs module to be installed when a custom NB hint config (without findbugs) is selected - there is probably something wrong in that logic.

jlahoda:
I'll leave that on @jtulach, but I'd vote for removing the module from the standard configuration (nbbuild/cluster.properties) and moving it to contrib, or deleting it. It will be in the history in case of need, and given the license of SpotBugs (and FindBugs), it seems more likely support for it would be a 3rd party plugin anyway.

mbien:
roger, say no more. Its now deleted. Going to squash if everything builds. (I removed some friend dependencies but only tested the base cluster locally)

mbien:
the "compile on 11 and test on 8" job errored twice in a row. There might be something wrong with CI right now. I haven't changed anything beside squashing the commits.

JaroslavTulach:
> removing things makes @JaroslavTulach angry

I am usually against removing API or infrastructure modules. Findbugs module is purely end-user oriented and has never provided any API as far as I know.

mbien:
> I am usually against removing API or infrastructure modules. Findbugs module is purely end-user oriented and has never provided any API as far as I know.

i was just kidding

mbien:
going to merge it if nobody disagrees. The main remark was that the module should be removed instead of just disabled, which is now the case.

source -> inspect doesn't work properly without this.

neilcsmith-net:
@mbien OK, all green, no-one has disagreed yet, and I need to merge this now if it's going in before branch, so getting ahead of you ... :smile:

PR #3458, opened 2022-01-15
lambda expression completion support.

mbien:
Adds a completion item for inserting lambda expressions.
```java
Stream.of().filter();                // cursor between ()
Stream.of().filter((t) -> false);    // generated with new expression completion
Stream.of().filter((t) -> {          // old block completion item remains available
    return false;
});      
```

edit:
it works but it has a problem: user changing the template could easily break the code responsible to mark the replacement String with `${...}` to make the generated code tab-navigatable.

https://user-images.githubusercontent.com/114367/174423136-45b86da0-04ff-4c83-9f96-a21c092fcc22.mp4

### DIFF
--- a/java/java.source/src/org/netbeans/modules/java/source/package-info.java
+++ b/java/java.source/src/org/netbeans/modules/java/source/package-info.java
@@ -20,10 +20,10 @@
 @TemplateRegistrations({
     @TemplateRegistration(folder = "Classes/Code", position = 100, content = "resources/GeneratedMethodBody.template", scriptEngine = "freemarker", displayName = "#GeneratedMethodBody", category = "java-code-snippet"),
     @TemplateRegistration(folder = "Classes/Code", position = 200, content = "resources/OverriddenMethodBody.template", scriptEngine = "freemarker", displayName = "#OverriddenMethodBody", category = "java-code-snippet"),
-    @TemplateRegistration(folder = "Classes/Code", position = 300, content = "resources/LambdaBody.template", scriptEngine = "freemarker", displayName = "#LambdaBody", category = "java-code-snippet")
+    /*see layer.xml for non-public templates*/
 })
 @NbBundle.Messages({
-    "GeneratedMethodBody=Generated Method Body", "OverriddenMethodBody=Overridden Method Body", "LambdaBody=Generated Lambda Body" //NOI18N
+    "GeneratedMethodBody=Generated Method Body", "OverriddenMethodBody=Overridden Method Body" //NOI18N
 })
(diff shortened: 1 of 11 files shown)


java/java.source/src/org/netbeans/modules/java/source/package-info.java
```diff
@@ -20,10 +20,11 @@
...
     @TemplateRegistration(folder = "Classes/Code", position = 200, content = "resources/OverriddenMethodBody.template", scriptEngine = "freemarker", displayName = "#OverriddenMethodBody", category = "java-code-snippet"),
-    @TemplateRegistration(folder = "Classes/Code", position = 300, content = "resources/LambdaBody.template", scriptEngine = "freemarker", displayName = "#LambdaBody", category = "java-code-snippet")
+    @TemplateRegistration(folder = "Classes/Code", position = 300, content = "resources/LambdaBody.template", scriptEngine = "freemarker", displayName = "#LambdaBody", category = "java-code-snippet"),
+    @TemplateRegistration(folder = "Classes/Code", position = 400, content = "resources/LambdaExpression.template", scriptEngine = "freemarker", displayName = "#LambdaExpression", category = "java-code-snippet")
 })
 @NbBundle.Messages({
-    "GeneratedMethodBody=Generated Method Body", "OverriddenMethodBody=Overridden Method Body", "LambdaBody=Generated Lambda Body" //NOI18N
+    "GeneratedMethodBody=Generated Method Body", "OverriddenMethodBody=Overridden Method Body", "LambdaBody=Generated Lambda Body", "LambdaExpression=Generated Lambda Expression" //NOI18N
 })
```
mbien, line 28: @dbalek is it possible to hide the Code Snippets for lambdas somehow so that the user can't edit them via the template window?

mbien: @jtulach do you happen to know if that would be possible? To hide the snipped from the user?

matthiasblaesing: If you don't want the user to edit the template, why put in a user accessible position in the first place? can't the template be read from the classpath or defined inline?

mbien:
> Looks good to me and seems to work ok. Before you merge, please add a license header to the template files. There are already comments i place that can also hold that.

Can we use the shorter variant as described [here](https://www.apache.org/foundation/license-faq.html#Apply-My-Software). Templates are read every time the completion happens.

mbien:
ok going to add the regular license header

mbien:
done. Going to merge once everything is green again.

PR #3459, opened 2022-01-15
Support to redefine icons for LSP client

sdedic:
I reworked the stuff originally in #3414 to a state hopefully acceptable for NB13 before freeze/branching. Overview

### Openide changes
- ImageUtilities just formalize `PROPERTY_URL` in API + implementation bugfixes

### LSP / VSCode changes
- UIDefaults are patched at startup, with `ImageUtilities.loadImage`s converted to icons for keys that are actually present. Replacement table is in a resource file. Stub image is replicated at build time, and mapped to images not mapped explicitly (just listed).
- regexps in vscode client adjusted.


java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/resources/uimanager-icons.properties
```diff
@@ -0,0 +1,63 @@
...
+Table.columnSelection
+ToolBar.handleIcon
+Nb.Explorer.Folder.icon
+Nb.Explorer.Folder.openedIcon
+Tree.gtk_collapsedIcon
+Tree.gtk_expandedIcon
+Tree.closedIcon=org/netbeans/modules/nbcode/integration/resources/uidefaults/Tree.closedIcon.png
+Tree.collapsedIcon
+Tree.expandedIcon
+Tree.leafIcon==org/netbeans/modules/nbcode/integration/resources/uidefaults/Tree.leafIcon.png
```
eirikbakke, line 62: Typo with the double equals here?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
```diff
@@ -56,6 +55,27 @@
...
+    public static class IconDescriptor {
+        /**
+         * Base URI / imageId of the icon.
+         */
+        public URI baseUri;
+        
+        /**
+         * Supplemental IDs from composed merged-in images or applied filters.
+         */
+        public String[] composition;
```
eirikbakke, line 77: As discussed earlier, trying to replace icons after they have already been loaded and even merged with badges etc. seems like a big hack. But you know the requirements here better than me, and it's localized to the LSP server stuff, so it's fine with me.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/resources/uimanager-icons.properties
```diff
@@ -0,0 +1,63 @@
...
+OptionPane.warningIcon
+PropSheet.customButtonIcon
+RadioButton.icon
+Table.columnSelection
+ToolBar.handleIcon
+Nb.Explorer.Folder.icon
+Nb.Explorer.Folder.openedIcon
+Tree.gtk_collapsedIcon
+Tree.gtk_expandedIcon
+Tree.closedIcon=org/netbeans/modules/nbcode/integration/resources/uidefaults/Tree.closedIcon.png
```
eirikbakke, line 59: Will this new png file show up somewhere in the GUI? There are better open/closed folder icons available in the codebase (and even SVG versions)... but I assume this is just a placeholder default.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.swing.UIDefaults;
+import javax.swing.UIDefaults.LazyValue;
+import javax.swing.UIManager;
+import org.openide.modules.OnStart;
+import org.openide.util.ImageUtilities;
+import org.openide.util.Lookup;
+
+/**
+ * Goes through UIDefaults and replaces Icons and Images with variants that
+ * return image's URL from the 'url' property.
```
eirikbakke, line 40: Could you add a comment to this Javadoc pointing out that this will only run during the headless LSP stuff, and not during regular NetBeans IDE execution?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+            replaceIconsAndImages(UIManager.getDefaults());
+        });
+    }
+    
+    static void replaceIconsAndImages(UIDefaults defs) {
+        ClassLoader l = Lookup.getDefault().lookup(ClassLoader.class);
+        try (InputStream is = UIDefaultsIconMetadata.class.getResourceAsStream("resources/uimanager-icons.properties")) { // NOI18N
+            Properties p = new Properties();
+            p.load(is);
+            // UIDefaults.keySet() is so lazy implemented that it does not enumerate any items ... but entrySet
```
eirikbakke, line 61: Yeah, I've noticed this in the past, too. Using entrySet to get all the keys is fine.

eirikbakke (changes requested):
Thanks for splitting out this digestible PR; looks good. I read over everything, though assuming that the LSP stuff was reviewed before by someone more familiar with that part of the codebase.

See in-code comments... main changes requested was fixing a double equals in a properties file and adding a comment to UIDefaultsIconMetadata to make it clear that it doesn't run during normal IDE execution.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
sdedic: The point is that (after some further adjustments), the client could get just icon's metadata (base URL + composition info) - and then ***decide*** to either handle the picture itself - or ask for the resource's bits back to the server: in that case, the original image's bits will be served.
This may seem awkward, but I don't think a single LSP server can satisfy various client needs / UI styles

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/resources/uimanager-icons.properties
sdedic: They might show. Pls. suggest the more appropriate ones - I'll change them when I incorporate SVG support into the LSP protocol. From the (current) vscode release's (12.6.301, 13.0) perspective, these are just placeholders.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/TreeItem.java
eirikbakke: Yeah, I still don't quite understand what the LSP server _is_. A screenshot would still be appreciated!

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/resources/uimanager-icons.properties
eirikbakke: Looking at the icon resource files, I think these are the best PNG files to use; the new ones are SVG files only. You can include apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/resources/defaultFolder.svg and
apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/resources/defaultFolderOpen.svg alongside the old PNG files for when you want to support SVG support.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/resources/uimanager-icons.properties
sdedic: Fixed in 871486f

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/UIDefaultsIconMetadata.java
sdedic: BTW the **whole subtree** `nbcode` is a seprate NB platfrom/ application: there are many registrations that would be disruptive in normal IDE. I put a `README.md` in the toplevel directory as well -  6e2382f

eirikbakke: Great, thanks!

eirikbakke (approved):
Thanks for the updates!

jtulach (approved):

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -117,18 +125,27 @@
...
+     */
+    private static volatile Component dummyIconComponentButton;
 
     static {
         /* Could have used Mutex.EVENT.writeAccess here, but it doesn't seem to be available during
         testing. */
         if (EventQueue.isDispatchThread()) {
-            dummyIconComponent = new JLabel();
+            dummyIconComponentLabel = new JLabel();
+            dummyIconComponentButton = new JButton();
```
neilcsmith-net, line 142: This still looks suspicious - `dummyIconComponentButton` is a different type depending on which thread hits this first?

neilcsmith-net (changes requested):
Can you double check the static initialization logic in ImageUtilities.  Then please squash and merge when ready.  Will be branching for NB13 later today.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
sdedic: Done - fixed & review feedback squashed into the feature commits.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -117,18 +124,27 @@
...
+     */
+    private static volatile Component dummyIconComponentButton;
 
     static {
         /* Could have used Mutex.EVENT.writeAccess here, but it doesn't seem to be available during
         testing. */
         if (EventQueue.isDispatchThread()) {
-            dummyIconComponent = new JLabel();
+            dummyIconComponentLabel = new JLabel();
+            dummyIconComponentButton = new JCheckBox();
```
eirikbakke, line 141: If we're doing a JCheckBox here instead of a JButton, the variable should be called dummyIconComponentCheckBox.

But why are we doing a JCheckBox again? I know it extends JButton but unless you know of specific cases that need it, I'd prefer to just have a JButton. Otherwise the hacks start getting too specific, and we'll be afraid to change them in the future.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
```diff
@@ -447,7 +471,22 @@ public static Image createDisabledImage(Image image)  {
...
+     * the URL may be specific for a localization or branding, and may be the same for
+     * bare and badged icons.
+     * 
+     * @param image image to inspect
+     * @return image's URL or {@code null} if not defined.
+     * @since 9.24
+     */
+    public static URL findImageBaseURL(Image image) {
+      Object o = image.getProperty(PROPERTY_URL, null);
+      return o instanceof URL ? (URL)o : null;
```
eirikbakke, line 487: Space after ")", if you're doing another round of changes...

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
sdedic: During development of the UIDefaults replacing code, I used `ImageUtilities` on real icons served by UIManager (default LaF). Several of them had special implementations that downcast the Component to JCheckbox in their `paint()`. Having a JCheckbox satisfies more icons than just JButton (I tried to choose the class that covered most of the errors).

eirikbakke: OK, as long as you saw actual cases, that's fine. What do you mean by "using ImageUtilities"; which method(s) specifically? And which LAF was this? (Which OS?) Why haven't these cases come up before?

sdedic: > which method(s) specifically
`ImageUtilities.icon2Image`; that ultimately calls `paintIcon()` to get the icon rendered to a `BufferedImage`. OS Linux / KDE, JDK 8+11, default L&F. 

> Why haven't these cases come up before?
Nobody called `icon2Image` on that particular UIManager icon ?

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
neilcsmith-net: Thanks!

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
neilcsmith-net: I think the reasoning is captured fine in the comment above too.

neilcsmith-net (approved):
OK, let's get it in before branching.

neilcsmith-net:
@sdedic actually, hold on!  Travis CI VSCode test is broken.  Lots of logged `java.lang.NoClassDefFoundError: Could not initialize class org.openide.util.ImageUtilities`.

platform/openide.util.ui/src/org/openide/util/ImageUtilities.java
eirikbakke: Yeah, it's fine with me. Just wondering if these were problems caused due to aggressive manual testing only, or whether they actually occur during real use.

eirikbakke:
@neilcsmith-net ImageUtilities.image2Icon(ImageUtilities.loadImage(resImage)) is suspect in any case, since loadImage can return null and image2Icon does not accept null.

neilcsmith-net:
@sdedic I need to proceed with branching. Once available, please rebase this on delivery.

sdedic:
Got it, running tests. Will rebase.

sdedic:
Seems that when the 1st call to ImageUtilities happened in the EDT (seemed not the case when running as a headless NBLS server **or** as an IDE, but did in the test), UIDefaultsIconMetadata attempted to call back to stil-not-initialized class `ImageUtilities` that resulted in CDNFE; at least when running locally.

neilcsmith-net:
Thanks! See, I knew that would cause a bug somewhere, or that fixing it would. :laughing:

Bit of a backlog in tests with all the branching and merging today.  Please merge when it goes green.

neilcsmith-net (approved):

PR #3460, opened 2022-01-15
[cnd] 4th and 5th Oracle donations

vieiro:
Those parts of the cnd and dlight clusters to be integrated with cpplite in the future. The rest of modules of cnd/dlight have been moved to the contrib area.
Features:
- Make based C/C++ projects & templates.
- C/C++ Toolchains
- Remote project support (builds in remote boxes).
- Cleans/builds/runs projects (no debugger: to use cpplite's)
- Editor with syntax highlighting, folding, formatting (no completion: to use cpplite's).
- No semantic highlighting (to use ccplite's).


matthiasblaesing:
I don't think it is ready right now. I saw a serious problem when I tested with this: I configured a project as "Lightweigth C/C++" project, so I would expect that the LSP server is started up. This is not the case (I checked the process output). I also noticed, that the syntax highlighting was off and looked strange. I think I'm seeing the old C/C++ support kicking in and the cpplite one in the background.

My understanding was, that we wanted to use what works from cpplite and have CND only supply additional info.

vieiro:
Perfectly fine, thanks @matthiasblaesing .
We will then enter stage "cnd-small/3" in the plan https://cwiki.apache.org/confluence/display/NETBEANS/cnd+integration+proposal%3A+cnd-small and do all those tasks in the "cnd" branch.
I've updated the main tasks to do in confluence as discussed in the mailing list.

PR #3465, opened 2022-01-16
[cnd] S3.1 cpplite/cpplite.debugger to cnd/cnd.debugger

vieiro:
These commits are for step 3.1 in 
https://cwiki.apache.org/confluence/display/NETBEANS/cnd+integration+proposal%3A+cnd-small

Tested making cpplite's dataloaders temporarily acknowledge 'z' files as C files (not commited).

Captura de pantalla de 2022-01-16 11-42-21

Since we want to keep track of IP in this branch, I'd suggest _not_ to squash.

BTW, cpplite.debugger tests fail (even in master). We should handle that later on.


mbien:
awesome, that is some great progress!

vieiro:
Hi @entlicher ,

In order to continue integrating the 4th and 5th Oracle donation, we are moving the "cpplite/cpplite.debugger" into "cnd/cnd.debugger" in the "cnd" branch, as Jan recently suggested in the mailing list https://lists.apache.org/thread/zlf78ryof5d3cg4g2gh3dsx0fzrjhrgj

He suggested we talking to you. As you can see in the 3rd commit we've already applied the change in "dlight.nativeexecution", "nativeimage.api" and "java.nativeimage.debugger" (in the cnd branch, not in master). AFAIU these are the only places where "cpplite.debugger" is used.

If you have time we'd appreciate you taking a look at this PR and see if we're missing anything.

Later on, before merging into master, we'll resolve any conflicts that may happen.

Thanks,
Antonio

vieiro:
Hi again, @entlicher ,

Since there're dependencies with the debugger in  ide/dlight.nativeexecution/nbproject/project.xml, ide/nativeimage.api/nbproject/project.xml and java/java.nativeimage.debugger/src/org/netbeans/modules/java/nativeimage/debugger/actions/NIAttachCustomizer.java , we may as well move "cpplite/cpplite.debugger" to "ide/lsp.debugger" instead, at your liking.

Thanks,
Antonio

matthiasblaesing:
I want to be truthful, I only looked at it and it looks sane to me (without going to deep into the inner workings of the debugger). 

From my perspective the first two commits look like good candidates for squashing, as they both just move files around. From a testing perspective, commits, that lead to a broken state (in this case b3290b2 and beefa36) should be avoided, as they give headaches if you try to bisect. Do we really loose something if all three commits are squashed? The IP is still clear:

- the original code import was already voted on (@jlahoda took care of that)
- all following changes went through normal committer stages and thus are also save

As a final comment: The base of binary dependency `cpplite-mi` (source is here: https://github.com/matthiasblaesing/cpplite-mi) can be found in `contrib/cnd.debugger.gdb2` and `contrib/cnd.debugger.common2`. At a later stage (maybe even after integration of CND), we should drop the binary dependency.

vieiro:
Maybe we should avoid messing with cpplite (and the whole codebase) right now. Instead ov moving cpplite/cpplite.debugger onto cnd/cnd.debugger, we could just copy it and just evolve cnd independently. Once cnd is up and running with LSP servers, we can then worry about cpplite.

For instance, cnd has this "C/C++ Project with Existing Sources" that we could adapt to migrate cpplite projects. Once this is in place (and cnd works well with lsp servers) we can drop cpplite entirely.

So let's close this one, let's copy (not move) cpplite.debugger into cnd.debugger, right?

PR #3352, opened 2021-12-02
Adding support for Oracle Cloud

jhorvath:
Adding new module with support for Oracle Cloud. For now it is able to browser Tenancy and Compartments. In selected Compartment it shows Autonomous Databases and allow to create and download Database Wallet.

PR #3468, opened 2022-01-17
Added ExprorerManager for Cloud services to nbcode

jhorvath:
Added ExprorerManager for Cloud services to nbcode.  This will enable Oracle Cloud browser in explorer view.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/cloud/OCIExplorer.java
```diff
@@ -0,0 +1,55 @@
...
+    public static final String CLOUD_PATH = "Cloud"; // NOI18N
+    private static String CLOUD_NODE_NAME = "cloud";
+
+    @Override
+    public CompletionStage<ExplorerManager> createManager(String id, Lookup context) {
+        if (!ID.equals(id)) {
+            throw new IllegalArgumentException("Unsupported view"); //NOI18N
+        }
+        ExplorerManager em = new ExplorerManager();
+        Lookup.Result<Node> nodes = Lookups.forPath("UI/Runtime").lookupResult(Node.class);
```
sdedic, line 47: Maybe a runtime dependency on `org.netbeans.modules.server` could be declared ?

java/java.lsp.server/nbcode/nbproject/platform.properties
```diff
@@ -77,7 +74,6 @@ disabled.modules=\
     org.netbeans.libs.git,\
     org.netbeans.libs.graaljs,\
     org.netbeans.libs.ini4j,\
-    org.netbeans.libs.jaxb,\
```
sdedic, line 80: Please doublecheck the dependencies. I can see some of them starting recursively from `cloud.common` (but there's no impl of cloud.common SPI ?) but which of the 'feature' modules needs bcprov jaxb or restlib ? Isn't there some impl-providing module missing ?

java/java.lsp.server/vscode/src/extension.ts
```diff
@@ -827,6 +827,7 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
             // create project explorer:
             c.findTreeViewService().createView('foundProjects', 'Projects', { canSelectMany : false });
             createDatabaseView(c);
+            c.findTreeViewService().createView('cloud.resources', undefined, { canSelectMany : false });
```
sdedic, line 830: I guess we could make some `package.json` registration of our own for this :)

java/java.lsp.server/nbcode/nbproject/platform.properties
jhorvath: This is really needed: `Warning - could not install some modules:
	org.netbeans.modules.xml.jaxb.api - No module providing the capability com.sun.xml.bind could be found.
	org.netbeans.modules.websvc.restlib - The module org.netbeans.modules.xml.jaxb.api would also need to be installed.
	org.netbeans.modules.cloud.oracle - The module org.netbeans.modules.websvc.restlib would also need to be installed.`

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/cloud/OCIExplorer.java
sdedic: Who knows. The implementation depends on it, if the Node (= the module) is not there, we get an always empty Cloud view in the LSP client... I thought it is desirable, at this moment to declare the dependency, so it is also visible at the 'component' level. Maybe not...

java/java.lsp.server/nbcode/nbproject/platform.properties
sdedic: Aha ! So this fixes an earlier addition of `cloud.oracle` (since it was added / not disabled in NBLS in #3352) ... Thanks.

sdedic (approved):

dbalek (approved):

PR #3469, opened 2022-01-17
Increment spec versions for Apache NetBeans 14 development.

neilcsmith-net:
Increment all module spec versions on master for Apache NetBeans 14 now that delivery has been branched off for 13.


mbien:
@neilcsmith-net @JaroslavTulach nb-javac is incremented separately, right? I remember there being an issue where someone imported old nb-javac from the last install and could not update it since it had the same version.

neilcsmith-net:
Now it's an actual module and no longer in its own weird, fake update centre, it should be fine I think?  Seems to be in this list, incremented to 17.1?

mbien:
> Now it's an actual module and no longer in its own weird, fake update centre, it should be fine I think? Seems to be in this list, incremented to 17.1? Hopefully that's what we want?!

ah its on the list. My browser couldn't find it. Thats good - so the same issue can't happen again basically now that it is deployed by default. I only wanted to be sure :)

PR #3470, opened 2022-01-17
failsafe:integration-test from contextual menu

abourree:
When editing JUnit class of integration test, I expect "Test File", "Debug Test File", "Run Focused Test Method" or "Debug Focused Test Method" to run "integration-test.single" or "debug.integration-test.single" actions instead of "test.single" or "debug.test.single".
 
The goal of this feature is to run integration test during integration-test phase and so after pre-integration-test.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/TestChecker.java
+++ b/java/maven/src/org/netbeans/modules/maven/TestChecker.java
@@ -59,6 +59,17 @@ public class TestChecker implements PrerequisitesChecker {
                     config.setProperty("test", test + '#' + method);
                 }
         }
+        if (ActionProviderImpl.COMMAND_INTEGRATION_TEST_SINGLE.equals(action) ||
+            ActionProviderImpl.COMMAND_DEBUG_INTEGRATION_TEST_SINGLE.equals(action) ||
+            "profile-tests".equals(action)) //NOI18N - profile-tests is not really nice but well. 
+        {
+                String test = config.getProperties().get("it.test"); //NOI18N
+                String method = config.getProperties().get(DefaultReplaceTokenProvider.METHOD_NAME);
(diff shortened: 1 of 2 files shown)


java/maven/src/org/netbeans/modules/maven/TestChecker.java
```diff
@@ -59,6 +59,17 @@
                     config.setProperty("test", test + '#' + method);
                 }
         }
+        if (ActionProviderImpl.COMMAND_INTEGRATION_TEST_SINGLE.equals(action) ||
+            ActionProviderImpl.COMMAND_DEBUG_INTEGRATION_TEST_SINGLE.equals(action) ||
+            "profile-tests".equals(action)) 
+        { //NOI18N - profile-tests is not really nice but well.
```
matthiasblaesing, line 65: Please move the NOI18N onto the line where the string is located (one line up). Else it is not connected to that string. (Yes I know that there other locations in that file, that are missing NOI18N markers or have them on the next line, but ...)

matthiasblaesing (changes requested):
Thank you - looks sane to me. Please fix the minimal nitpick I found.

java/maven/src/org/netbeans/modules/maven/TestChecker.java
matthiasblaesing: @abourree I saw the updated commit - would you mind squashing the to commits into one?

abourree:
Sorry, I try to squash commits, but look not working

matthiasblaesing:
@abourree seems you managed to do it, thank you :-) - I activated the test runs and if they run clean, I'll merge. I assume you read the license statement at the top of the file and understood, that you donate your changeset to the ASF.

matthiasblaesing (approved):

PR #3472, opened 2022-01-18
Hint/Inspection panel and dialog layout fixes

mbien:
- `source -> inspect` dialog did not resize properly
- `refactor -> inspect and refactor` dialog had no separation between row two and three (very similar dialog to the one above but uses a different layout manager :) )
- layout fixes to HintPanel as used in `options -> editor -> hints` or from various configuration dialogs
  - description resizes now vertically, required updates to many customizer components so that they only take as much space as they need
  - made dialogs which use the HintPanel a little larger

### DIFF
--- a/ide/code.analysis/src/org/netbeans/modules/analysis/RunAnalysisPanel.form
+++ b/ide/code.analysis/src/org/netbeans/modules/analysis/RunAnalysisPanel.form
@@ -36,7 +36,7 @@
     <AuxValue name="FormSettings_listenerGenerationStyle" type="java.lang.Integer" value="0"/>
     <AuxValue name="FormSettings_variablesLocal" type="java.lang.Boolean" value="false"/>
     <AuxValue name="FormSettings_variablesModifier" type="java.lang.Integer" value="2"/>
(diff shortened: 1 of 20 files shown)


neilcsmith-net:
@mbien need to rebase on delivery if you want this in NB13 - picking up commits from master we don't want.

matthiasblaesing:
In this case we don't want a rebase, but a cherry-pick. The difference is, that rebase will carry all commits from the base branch, that are not in the target branch, while cherry-picking only uses the referenced commits.

mbien:
@neilcsmith-net done

@matthiasblaesing yeah that is correct. Although I did cherry pick to delivery, I still refer to it as "rebasing the two commits to delivery" myself - I probably shouldn't since it could be taken literally.

neilcsmith-net:
@mbien thanks.

@matthiasblaesing yes, rebasing was more a turn of phrase around changing the PR base branch.  Mind you, still prefer using git rebase for this task myself, but don't mind how anyone wants to clean up their PR unless they need help with it.  Now, as long as we don't get cherry-picking from upstream master I'm happy ... :smile:

PR #3475, opened 2022-01-18
LSP: java.source.save.CasualDiff: Illegal values after using CodeCompletion.

dbalek:
After using code completion for e.g. `System.out`, a syntax error due to unfinished statement is usually reported. At the same time VSCode client asks for all available CodeActions (Hints) at the current line. One of the hints offers creating a static import for `System.out` and therefore a diff computation for such change touches the `ErroneousTree` representing the unfinished statement. Change to the `CasualDiff` was necessary to correctly handle this situation.


neilcsmith-net:
Anything for consideration for NB13 needs rebasing on delivery.

JaroslavTulach (approved):
Let' do it!

PR #3479, opened 2022-01-19
Fix for testPreTrusted gradle test failing on CI

sdedic:
This PR attempts to fix gradle test suite that is randomly failing just on CI. The root cause is that obviously I've written the test wrong ;) the original idea was to test what will happen if:
- the project is trusted
- it has been compiled, closed, IDE restarted
- the project is opened again.
Since I didn't want to bring all the NbModuleSuite machinery, the test just moved the original project, created a copy == new `FileObject` and opened the project again. New FileObject means new Project.

But some Gradle caches rely on `j.io.File`, not FileObjects, so [at this point](https://github.com/apache/netbeans/blob/master/java/gradle.java/test/unit/src/org/netbeans/modules/gradle/java/classpath/ClassPathProviderImplTest.java#L248) the `prj2` was a new instance, but its `GradleFiles` key into the cache was **the same** as the `prj` key...
... so the project info was found in the cache loaded, and was returned. And the test was failing because **sometimes the timestamps of the copied files were different** from the timestamps of the in-memory loaded cache data, if CI was lucky to execute the two project loads in different seconds. Then the cache failed, and legacy project loader did not execute the script  because it should not do so implicitly on open, but only when (somehow) asked to load the real contents - that is intended and is being tested in Gradle core testsuite.

So I split the test into two cases:
1. reopening a trusted, closed project w/ caches: this should report proper classpath immediately from the cached data.
2. reopening a trusted project wil cleared caches: this should initially report some classpath, but as the project is trusted, it should execute the configuration and correct the classpath. This behaviour is desirable when Java files are opened, so that their compilation classpath (and editor underlines) are fixed if the trust allows to execute buildscript.

I need to [flush the project info cache](https://github.com/apache/netbeans/pull/3479/files#diff-fe8817b482b8cf192d5a3f06c1ed4cc2a186d7b38e0f4436a5366a8eb8fbe02bR60) from the test, and potentially [delete the cache data for project](https://github.com/apache/netbeans/pull/3479/files#diff-fe8817b482b8cf192d5a3f06c1ed4cc2a186d7b38e0f4436a5366a8eb8fbe02bR67)

During the debugging, I've noticed that access to the disk cache  [was not synchronized](https://github.com/apache/netbeans/pull/3479/files#diff-fe8817b482b8cf192d5a3f06c1ed4cc2a186d7b38e0f4436a5366a8eb8fbe02bR48) and while that did not cause the test failures, it could cause weird things in production (as the hashtable may become corrupted).

I've also encountered a **deadlock**  caused by [`toString()`](https://github.com/apache/netbeans/pull/3479/files#diff-e833b1ad0623c45c9af375ded2ba2f854c529638f4334c624969e3ea40a6e05eR477) was trying to synchronize on the project's instance. If the project object was just being formatted by a LogHandler, and some other thread was trying to use Logger in `synchronized` section using that same project instance, the two threads would block each other.

Then there was a [nasty typo](https://github.com/apache/netbeans/pull/3479/files#diff-e833b1ad0623c45c9af375ded2ba2f854c529638f4334c624969e3ea40a6e05eR315) that caused the `CompletableFuture` NOT to complete on success - and therefore Source Groups [were not refreshed[(https://github.com/apache/netbeans/blob/master/java/gradle.java/src/org/netbeans/modules/gradle/java/classpath/ClassPathProviderImpl.java#L159) after project load. This typo was not even noticed by tests ! So I added checks for `java2` directory that should appear after full load in `ClassPath.SOURCE`.

I left some logging tweaks in the test class: it enables FINER logging for the test that was randomly failing, so if this PR does not fix the problem, CI logs will hopefully contain some hints to analyse. I'll remove the setLevel later after the fix proves OK.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
@@ -30,6 +30,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.prefs.Preferences;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
@@ -44,6 +46,8 @@ import org.openide.util.NbPreferences;
  * @author lkishalmi
  */
 public class ProjectTrust {
+    private static final Logger LOG = Logger.getLogger(ProjectTrust.class.getName());
+    
     private static final String KEY_SALT     = "secret";     //NOI18N
     private static final String NODE_PROJECT = "projects"; //NOI18N
     private static final String NODE_TRUST   = "trust";    //NOI18N
@@ -82,6 +86,7 @@ public class ProjectTrust {
     public boolean isTrusted(Project project) {
         synchronized (this) {
             if (temporaryTrustedIds.contains(getPathId(project))) {
+                LOG.log(Level.FINER, "Project {0} temporarily trusted.", project);
                 return true;
             }
         }
@@ -108,7 +113,9 @@ public class ProjectTrust {
             List<String> trust = Files.readAllLines(trustFile);
             String hash = hmacSha256(fromHex(projectId));
             ret = trust.size() == 1 && trust.iterator().next().equals(hash);
+            LOG.log(Level.FINER, "Trust for project {0} is: {1}", new Object[] { project, ret });
         } catch (IOException ex) {
+            LOG.log(Level.FINER, "Could not load trust file {0} for projec {1}.", new Object[] { trustFile, project});
         }
         return ret;        
     }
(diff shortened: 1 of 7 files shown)


sdedic:
Note: this PR need not to be merged into `release` NB13 (no real functional change), but the failing test is a pain in .... anywhere.

extide/gradle/src/org/netbeans/modules/gradle/ProjectTrust.java
```diff
@@ -108,7 +113,9 @@ public boolean isTrustedPermanently(Project project) {
             List<String> trust = Files.readAllLines(trustFile);
             String hash = hmacSha256(fromHex(projectId));
             ret = trust.size() == 1 && trust.iterator().next().equals(hash);
+            LOG.log(Level.FINER, "Trust for project {0} is: {1}", new Object[] { project, ret });
         } catch (IOException ex) {
+            LOG.log(Level.FINER, "Could not load trust file {0} for projec {1}.", new Object[] { trustFile, project});
```
lkishalmi, line 118: typo: projec -> project

lkishalmi (approved):
Looks good so far.
Could you test if this one also resolves NETBEANS-6117 ?

sdedic:
The failed build on Windows is some connectivity issue, cannot be related to the changes (fails at the beginning of build). Merging.

PR #3480, opened 2022-01-19
[NETBEANS-6117] Fix Gradle project loading in template operations.

lkishalmi:
Well, in this case the isGradleProjectLoaded() and the getGradlePorject() in NbGradleProjectImpl does not really agree, if the project is only initialized with a FALLBACK grade project. That would result infinite nesting at [AbstractProjectLoader:74](https://github.com/apache/netbeans/blob/6146b70dd3d1b2a269f934c5a457b1aa876ceb88/extide/gradle/src/org/netbeans/modules/gradle/loaders/AbstractProjectLoader.java#L74)

Strangely it did not cause problems elsewhere so far. This fix may/may not be ported to NB13. @sdedic please evaluate what would happen with this change. (I keep this one on master for the time being).

I know the main reason of this change to load the info from disk cache if that would exists. A better solution would have been to get the Disk cache loader give back better than FALLBACK info if a cached data is already available on disk.

With all the options, the project load is getting quite convoluted. I'm going to give it a try and factor these out somehow. Will file another PR on that.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/loaders/AbstractProjectLoader.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/loaders/AbstractProjectLoader.java
@@ -71,7 +71,7 @@ public abstract class AbstractProjectLoader {
 
         public ReloadContext(NbGradleProjectImpl project, NbGradleProject.Quality aim, GradleCommandLine cmd, String description) {
             this.project = project;
-            this.previous = project.isGradleProjectLoaded() ? project.getGradleProject() : FallbackProjectLoader.createFallbackProject(project.getGradleFiles());
+            this.previous = project.isGradleProjectLoaded() ? project.projectWithQuality(null, FALLBACK, false, false) : FallbackProjectLoader.createFallbackProject(project.getGradleFiles());
             this.aim = aim;
             this.cmd = cmd;
             this.description = description;
--- a/extide/gradle/src/org/netbeans/modules/gradle/spi/newproject/TemplateOperation.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/spi/newproject/TemplateOperation.java
@@ -50,6 +50,8 @@ import static org.netbeans.modules.gradle.spi.newproject.Bundle.*;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectManager;
 import org.netbeans.modules.gradle.GradleProjectLoader;
@@ -65,6 +67,8 @@ import org.openide.util.NbBundle;
  * @author Laszlo Kishalmi
  */
 public final class TemplateOperation implements Runnable {
+    private static final Logger LOG = Logger.getLogger(TemplateOperation.class.getName());
+    
     public interface ProjectConfigurator {
         void configure(Project project);
     }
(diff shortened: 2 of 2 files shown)


lkishalmi:
Hmm might related with the typo in: https://github.com/apache/netbeans/pull/3479/commits/8fbb1eba35ff9c43f89c0b4a123dc07a4667cc67 ?

lkishalmi:
Well, the fixing that typo does not fix the issue. I also think that this change might cause a few test to fail.

sdedic:
> Well, in this case the isGradleProjectLoaded() and the getGradlePorject() in NbGradleProjectImpl does not really agree, if the project is only initialized with a FALLBACK grade project. That would result infinite nesting at AbstractProjectLoader:74

Is there a test or a sample code I can try ? 

> Strangely it did not cause problems elsewhere so far.

This is more or less by accident, but the behaviour is stable: `getGradleProject()` is first called from within `NbGradleProjectFactory.loadProject`, before the new project is passed to the ProjectManager. So the first request before anyone can see `NbGradleProjectImpl` (under normal circumstances) is for `EVALUATED`. 
Should that fail, e.g. the actually "loaded" project is just `FALLBACK`, further requests for `EVALUATED` are rejected (unless `force` flag is in effect), since the result of that is already known.

As a result, if the Loader sees non-null as a project (2nd+ attempts), it should not recurse. The recusion could happen, if one would force loading of a `FALLBACK` project.

It's can be better to ensure that if a `getGradleProject` is called **from within project loading process**, it gets always some stable value. IMHO making this call from within the loader can indicate a bug either: since the project is being loaded, it cannot return "the project".

>  I'm going to give it a try and factor these out somehow. Will file another PR on that.

I would say "please don't" as there are many things that can break with such a change. 

When I reworker the loading, `isGradleProjectLoaded()` users were using its `false` as an indication that none of the project information can be used. This is the case of a `dumped` project (forcibly closed): even FALLBACK project has some information (derived from directory structure) in it already.

The main reason for the rewrite was that Gradle project loading was breaking API semantics: project queries, such as ClassPathProvider or SourceForBinaryQuery must work even though the project is not UI-opened (through OpenProjects) - that's the requirement of project APIs in general. In the UI, the breakage manifests when you open a file from a closed Gradle project, for example by going to symbol declaration from some other file. In absence of correct information (i.e. compile ClassPath), all library symbols in the opened file are underlined, find usages, implementations etc do not work. This is different from J2SE project behaviour (which is the de-facto standard to follow) or Maven.

Now the loading process works in a way that ensures that if someone asks for project information, and the information is available in a serialized form, it will be loaded. In this (implicit) scenario, it is desirable not to execute Gradle for project configuration, as it may take considerable time not anticipated by the query / calling code semantics.

The  target state (now aim = EVALUATED) has to be identified differently from FULL state, so that the loading process can stop before actually executing Gradle - and the API clients can say what level of quality should be reached. I am not saying that exactly EVALUATED must be used, but it is a logical candidate: it is "a state that was once correct, but need not to be any more" - which is OK for queries that need at least some data, and can fire events when the data change.

---
To prevent the exact recursion, I can offer different techniques, depending on how the recursion happens exactly:
- call a **different** accessor in NbGradleProjectImpl, that will really return the `current` (= the previous for the loader) project
- ensure that the loading process will **always** get the same (current/previous) reference to the project from `getGradleProject`: during the load, as it is defined in the internal APIs, `getGradleProject` can't really return the 'current' (being loaded) project, as that instance is the product of the Loader and does not exist before it completes.

But as there are not IMHO any additional calls from Loaders except ReloadContext, I'd just use that alternative (direct) accessor. Need a scenario when the recursion actually happens to write a proper test.

lkishalmi:
@sdedic thanks for the detailed summary. It seems it would be enough to adjust the wizard operations and some of the loading work in the new paradigm. I still think, that some refactoring reorganization of the code needs to be done. NbGradleProjectImpl is leaking different GradleProjects on too many methods with too many parameters. What is on my mind is to encapsulate these parameters and the loading code into a separate object to make the things clearer.

@neilcsmith-net With that change I'm rebased this PR on delivery branch as this would only fix the template operations and the possible stack overflow in AbstractProjectLoader

sdedic (approved):

PR #3482, opened 2022-01-20
Enable building on JDK17 (bump source to 1.8 in web.client.rest and profiler.j2ee)

matthiasblaesing:


matthiasblaesing:
Given the minimal impact I intent to merge this on the weekend. If there is really an appserver running outthere one something older than 1.8 it has bigger problems, then performance measurements.

PR #3488, opened 2022-01-21
javafx ide.ergonomics config should not try to load the old nb-javac.

mbien:
as discovered on the dev-list, the "Java Frontend Application" wizzard is trying install/load the old nb-javac plugin.

This should fix it. (tested it once as described on the mail to verify)

@neilcsmith-net ping


neilcsmith-net:
Thanks @mbien for taking this off my hands! I'd have looked eventually - just venting my annoyance on list at how often this seems to break in releases. :laughing:

jtulach:
Great, thank you. I can confirm that with 83fe4039dc the wizard behaves sanely.

Issue #3487, opened 2022-01-21
Find in Projects - Pressing enter does not start the search anymore


JoachimRohde:
### Apache NetBeans version

Latest release candidate

### What happened

Open the "Find in Projects" dialog, type something into the "Containing Text" text field and press enter. 
I expected the same behavior as in previous versions (12.6 and before): that the search gets triggered. Instead a line break is inserted into the text field.

### How to reproduce

Open the "Find in Projects" dialog, type something into the "Containing Text" text field and press enter. 

### Did this work correctly in an earlier version?

- [X] This used to work!

### Operating System

Linux Mint 20.2 Cinnamon

### JDK

11.0.13; OpenJDK 64-Bit Server VM 11.0.13+8-Ubuntu-0ubuntu1.20.04

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow the Apache Software Foundation's [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html)

neilcsmith-net:
You can use CTRL-Enter. There were changes to support multiline text. #3393 However this is the second report of confusion already. Maybe some rethinking of UI required? Maybe should look multiline without resize? cc @mbien

emaphis:
Maybe include a note (as a label) in the dialog box for couple of releases.  Ugly but eye-catching.  Remove later.

mbien:
the reason i haven't noticed this is because I never use enter on dialogs. Half of the software out there never does what you would expect - so it became a habit to use the mouse for ok/cancel - esp on important stuff.

Anyway, i made the text areas behave like a text fields if they are still small enough. Should cover both use cases without the need to break habits. Please test it if it works ok for you.

neilcsmith-net:
@mbien some of us are used to NetBeans behaving correctly with the keyboard and try to use the mouse as little as possible!  I'm not sure your solution is correct so far, but let's take the discussion into #3489

@emaphis might be right about an ugly but eye catching label.

PR #3489, opened 2022-01-22
search/replace dialog: set shift+enter as line break and add multi-line info message.

mbien:
https://github.com/apache/netbeans/issues/3487

with this PR applied the following should be possible again:

1. press CTRL+SHIFT+H (or F)
2. type to enter search string
3. press ENTER to trigger search
(no mouse required)

which should be to my knowledge EXACTLY as it was before.

### DIFF
--- a/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
+++ b/platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
@@ -39,6 +39,7 @@ import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.swing.*;
+import javax.swing.border.Border;
 import javax.swing.border.EmptyBorder;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
@@ -216,7 +217,7 @@ final class BasicSearchForm extends JPanel implements ChangeListener,
 
         lblTextToFind = new JLabel();
         JComboBox<String> box = new JComboBox<>();
-        box.setEditor(new MultiLineComboBoxEditor());
+        box.setEditor(new MultiLineComboBoxEditor(box));
         cboxTextToFind = ComponentUtils.adjustComboForSearchPattern(box);
         lblTextToFind.setLabelFor(cboxTextToFind.getComponent());
         btnTestTextToFind = new JButton();
@@ -227,7 +228,7 @@ final class BasicSearchForm extends JPanel implements ChangeListener,
         if (searchAndReplace) {
             lblReplacement = new JLabel();
             cboxReplacement = new JComboBox<>();
-            cboxReplacement.setEditor(new MultiLineComboBoxEditor());
+            cboxReplacement.setEditor(new MultiLineComboBoxEditor(cboxReplacement));
             cboxReplacement.setEditable(true);
             cboxReplacement.setRenderer(new ShorteningCellRenderer());
             lblReplacement.setLabelFor(cboxReplacement);
@@ -1116,11 +1117,23 @@ final class BasicSearchForm extends JPanel implements ChangeListener,
 
         private final JTextArea area = new JTextArea();
 
-        public MultiLineComboBoxEditor() {
+        public MultiLineComboBoxEditor(JComboBox reference) {
             area.setWrapStyleWord(false);
             area.setLineWrap(false);
+
+            Border border = ((JComponent)reference.getEditor().getEditorComponent()).getBorder();
+            if (border == null) {
+                border = reference.getBorder();
+            }
+            area.setBorder(border);
+
+            // retain standard focus traversal behavior
             area.setFocusTraversalKeys(KeyboardFocusManager.FORWARD_TRAVERS‌​AL_KEYS, null);
             area.setFocusTraversalKeys(KeyboardFocusManager.BACKWARD_TRAVERS‌​AL_KEYS, null);
+
+            // dispatch enter to parent; set line breaks on shift+enter
+            area.getInputMap().put(KeyStroke.getKeyStroke("shift ENTER"), "insert-break");
+            area.getInputMap().put(KeyStroke.getKeyStroke("ENTER"), "text-submit");
         }
 
         @Override
--- a/platform/api.search/src/org/netbeans/modules/search/BasicSearchProvider.java
+++ b/platform/api.search/src/org/netbeans/modules/search/BasicSearchProvider.java
@@ -226,7 +226,8 @@ public class BasicSearchProvider extends SearchProvider {
                     notifySupport.setInformationMessage(UiUtils.getText(
                             "BasicSearchForm.txtInfoNoWildcards"));     //NOI18N
                 } else {
-                    notifySupport.clearMessages();
+                    notifySupport.setInformationMessage(UiUtils.getText(
+                            "BasicSearchForm.txtInfoMultiline"));     //NOI18N
                 }
             }
             return usable;
--- a/platform/api.search/src/org/netbeans/modules/search/Bundle.properties
+++ b/platform/api.search/src/org/netbeans/modules/search/Bundle.properties
@@ -248,6 +248,7 @@ BasicSearchForm.txtErrorTextPattern=Containing Text field contains invalid regul
 BasicSearchForm.txtErrorReplacePattern=Replace With field contains invalid expression
 BasicSearchForm.txtErrorFileName=File Name Patterns field contains invalid regular expression
 BasicSearchForm.txtInfoNoWildcards=File Name Pattern contains no wildcards. It may be a mistake.
+BasicSearchForm.txtInfoMultiline=Resize Window for multi-line search (SHIFT+ENTER for line breaks).
 
 LBL_OptionsPanelTitle=Options
 
--- a/platform/api.search/src/org/netbeans/modules/search/PatternSandbox.java
+++ b/platform/api.search/src/org/netbeans/modules/search/PatternSandbox.java
@@ -82,7 +82,7 @@ public abstract class PatternSandbox extends JPanel
     protected void initComponents() {
 
         cboxPattern = new JComboBox<String>();
-        cboxPattern.setEditor(new BasicSearchForm.MultiLineComboBoxEditor());
+        cboxPattern.setEditor(new BasicSearchForm.MultiLineComboBoxEditor(cboxPattern));
         cboxPattern.setEditable(true);
         cboxPattern.setRenderer(new ShorteningCellRenderer());
         lblPattern = new JLabel();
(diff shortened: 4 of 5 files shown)


platform/api.search/src/org/netbeans/modules/search/BasicSearchForm.java
```diff
@@ -1119,8 +1121,23 @@ private void limitReached() {
         public MultiLineComboBoxEditor() {
             area.setWrapStyleWord(false);
             area.setLineWrap(false);
+
+            // retain standard focus traversal behavior
             area.setFocusTraversalKeys(KeyboardFocusManager.FORWARD_TRAVERS‌​AL_KEYS, null);
             area.setFocusTraversalKeys(KeyboardFocusManager.BACKWARD_TRAVERS‌​AL_KEYS, null);
+
+            // pretend to be a single line text field if you are small enough
+            area.addKeyListener(new KeyAdapter() {
```
matthiasblaesing, line 1130: This will break accessibility. As a developer you know, that the behavior, but a user will not know and will be surprised if the field suddenly changes bahaviour. How about a toggle button, that switches between single and multiline case? Activating that could even provide a textarea, that is large enough by default to show multiple lines of text.

mbien: it is intended to be intuitive without the need to explain or configure anything. If it looks like a text field, it behaves just like it did before (users seem to like to use the Enter key as OK dialog action according to the mailing list and issue). But if you have a multi line-string in it you expect it having a working Enter key for new lines - if it would trigger the dialog action it would look like a bug too.

toggle would create more issues. since if you turn it off again the Enter key would suddenly be the dialog action and the text field would contain multi line strings.

matthiasblaesing: Still **I** would not guess, that I have to resize the dialog to do multiline searches and this has zero discoverability. The dialog is big enough, so why would I resize it?

Also think about people with bad eyesight, using assistive technologies, they will even be worse, because not only have they use their tech, the UI suddenly changes behavior, just because it was scaled.

For the toggle button. Toggling it can remove newlines or turn the search to regexp based search, that will happily serach for newlines even before this change. It would also be clearer, because a button can be expected to change behavior, resizing in my mind can't.

mbien: I agree on the discoverability aspect but this PR does not change anything there. There is already a msg label somewhere e.g stating "the text will be used literally". It could also add "resize window for multi line Strings". Adding a button which makes the window larger is unusual UI-wise, but this could be added too if there is a need (I would prefer not).

there are many little details to consider here. e.g if you select text in the editor, the dialog will already have properly sized texfields. The text fields are also used in the regex tester etc - so we would need to add those toggle buttons everywhere.

This is a very trivial issue IMO. The worst thing which can happen is that someone adds a newline in a text field instead of triggering the OK action - and only if the text field has focus. I wouldn't have expected that anyone would file an issue for that tbh.

This PR fixes it and restores original behavior for single line search/replace.

neilcsmith-net: I agree with @matthiasblaesing that this seems worse from an accessibility / discoverability point of view.  It may seem trivial, but it's been mentioned twice already, and it's annoyed me about 10 times already, and I know to press `CTRL+ENTER`.  Keyboard control of dialogs is second nature for some people!

I think either make the dialog show at least 2 lines so that multiline behaviour is expected, add some information about `CTRL-ENTER` (and whatever it is on macOS), add another way of accessing multiline, or revert and rethink after NB13.

mbien: @neilcsmith-net please give it a try first and test it. This PR should restore the original behavior. We can improve discoverability of the multi line feature later - this has low priority for me. A hidden feature is not a bug, this PR fixes a regression caused by https://github.com/apache/netbeans/pull/3393.

just to be to be sure i repeat: this PR changes nothing regarding discoverability, it restores original behavior so that if you are used to Enter as action trigger, you can keep using it.

mbien: with this PR applied the following should be possible again:
 - press CTRL+SHIFT+H
 - type to enter search string
 - press ENTER to trigger search
(no mouse required) 

 which should be to my knowledge EXACTLY as it was before.

mbien: btw CTRL+ENTER also works fine - i just tested it. It works in both cases, even if the well hidden and new multi line search is active.

@neilcsmith-net Maybe i am doing something wrong, but you said "it's annoyed me about 10 times already, and I know to press CTRL+ENTER". How can i reproduce this?

neilcsmith-net: @mbien `CTRL+ENTER` works fine. You're reading my comment the opposite of how it was intended.  For those of use who've used this for decades, switching away from just `ENTER` takes some relearning of muscle memory! :smile:

How about we do the reverse here anyway? Other UIs (not NetBeans) use `SHIFT+ENTER` or `ALT+ENTER` to type newline in these sort of cases. Let's keep the default behaviour as it was.

matthiasblaesing: I tried a less invasive approach here: https://github.com/matthiasblaesing/netbeans/commit/99c6b0b878aff54c498c1ed38c54006cd48abf4b

- The renderer for previous searches replaces new lines with a single space, so that the text is still readable, but compact
- An information text is shown in the notification area, that the search can be submitted by pression CTRL+enter (might need wording adjustment and handling the mac case)
- the input area is widend to approx. 3 lines, so that text can be realistically be added. In the future it might be optimized by being able to open a dialog with a bigger TextArea
- the border of the textarea used as editor is ajusted to match the normal combo box editor one

image

This is not polished, but IMHO improves the situation.

mbien: > @mbien `CTRL+ENTER` works fine. You're reading my comment the opposite of how it was intended. For those of use who've used this for decades, switching away from just `ENTER` takes some relearning of muscle memory! smile

ok understood, my mistake


> How about we do the reverse here anyway? Other UIs (not NetBeans) use `SHIFT+ENTER` or `ALT+ENTER` to type newline in these sort of cases. Let's keep the default behaviour as it was.

i get the feeling that everyone made the decision already before testing this fix, because... thats exactly what it does. It keeps the default behavior as it was :)

mbien: @matthiasblaesing lets not redesign the dialog in a bugfix PR during rc2. The discoverability improvement can be addressed later.

mbien: current

thats the version right now after this PR. Enter works, no change in behavior form 12.6.

matthiasblaesing: > thats the version right now after this PR. Enter works, no change in behavior form 12.6.

Until you resize the dialog, then "magic" happens. The "magic" part is what I don't like. It is not discoverable and can't be observed by accessibility tools. As a user you have to "know" that something will happen or you don't and get surprised.

mbien: > > thats the version right now after this PR. Enter works, no change in behavior form 12.6.
> 
> Until you resize the dialog, then "magic" happens. The "magic" part is what I don't like. It is not discoverable and can't be observed by accessibility tools. As a user you have to "know" that something will happen or you don't and get surprised.

I have to disagree there since that is how everyone should expect a larger text field to work - what else should happen if you edit a multi line text area and press enter?

As before mentioned CTRL+ENTER works too (it always did). So if you want I could add it as initial info msg as in your version.

To me this is all surprisingly off topic. Since this PR restores original behavior while most discussion is about how to make the (new) multiline feature more discoverable.

mbien: v2
would this help?

neilcsmith-net: @mbien this doesn't restore the old behaviour. As Matthias says it adds "magic". So, we can restore the old behaviour by reverting the UI changes for multiline, make things clearer as Matthias' change does, or change the newline key to be `SHIFT+ENTER` or `ALT+ENTER` which is similar to other software.  I'd be happy if we got any of those 3 options in to NB13, but disagree on the behaviour in this PR.

mbien: @neilcsmith-net
the screenshot above, option 1:
- says press CTRL+ENTER just like in Matthias'es version (just wanting to make sure that you didn't overlook it)
- it addresses the discoverability issue by suggesting to resize the window for multi line use
 
option 2:
 - add resize info text just like above for discoverability (but don't mention CTRL+ENTER)
 - always dispatch ENTER as dialog action
 - use SHIFT+ENTER as new line action as you suggested (and mention it in the msg)
 
I would prefer 1 since it has identical behavior to 12.6 and the hint msg clearly describes how to use the new feature. Option 2 is close second since it requires to use unintuitive SHIFT+ENTER for newline in text areas, but since its a niche feature, i would be ok with it.

I would also like to avoid making the textfields larger by default since it makes the dialog uglier for the most common usecase: single line search

neilcsmith-net: I'll leave you and @matthiasblaesing to decide over option 2 and his suggestion.  I agree with him that anything changing the behaviour based on window sizing is the wrong approach, with or without the tip.  Still somewhat prefer his suggestion.

Incidentally `SHIFT+ENTER` and `ALT+ENTER` are commons ways to handle this in chat programs and spreadsheets, so not that unintuitive, particularly as a retrofit.

matthiasblaesing: I would prefer option 2.

mbien: changed it to option 2 and took the border change from @matthiasblaesing 's commit with slight modifications, since some LAFs use the outer components border instead of the editor's border. (e.g as visible on the screenshot https://github.com/apache/netbeans/pull/3489#discussion_r790134047, there would be no space between border and text and look inconsistent)

matthiasblaesing (approved):
Looks sane to me.

neilcsmith-net (approved):
Thanks for bearing with us @mbien Just tested, looks good.

I did notice while testing that multiline strings reports a bunch of exceptions with empty files - think needs to bail out at this line when the file size is zero? https://github.com/apache/netbeans/blob/master/platform/api.search/src/org/netbeans/modules/search/matcher/MultiLineStreamMatcher.java#L84

Something to address elsewhere though - merging. Thanks!

PR #3490, opened 2022-01-22
[NETBEANS-6323] Fix Gradle popup menu Tasks.. action not working.

neilcsmith-net:
Fixing another fallout of disabled custom tasks in the IDE UI. ActionToTaskUtils::isActionEnabled always reports the `Run Gradle / Tasks..` action in project menu as disabled, so the action does nothing.  Switch to using custom prefix for name as in TaskNode actions.

I initially used a hidden action name checked in ActionToTaskUtils, but then noticed that the working version of this in the navigator just uses the custom prefix.


lkishalmi (approved):
Thank you!
It is harmless and good for NB13!

PR #3492, opened 2022-01-23
Fixes to LSP client/server related to semantic highlighting.

jlahoda:
Trying to use the semantic highlighting in https://github.com/apache/netbeans/pull/1475, I've found a few bugs in the handling of the LSP semantic highlighting (and also a bug in the LSP client code completion update to a new LSP4J version). This patch should be fixing that. If possible, it would make sense to me to put it into NB 13.


matthiasblaesing (approved):
I tested the typescript editor component with delivery and indeed found it broken. Parts worked, but completion for example failed. With this PR applied, it looks much better.

If noone does it before me or objects, I'll merge this tomorrow evening.

neilcsmith-net:
@matthiasblaesing thanks, but please leave us to handle merges to delivery. Will merge for rc2.

PR #3493, opened 2022-01-23
[NETBEANS-6058]

AaronTheBaker:
1) Fix Jira NETBEANS-6058: new maven action is disabled permanently
2) Related updates:
- Add handler for "Disable" button. Only allow custom actions to be disabled.
- Disable "Disable" button for default actions.
- When removing custom action, automatically select previous action in list.
- When disabling action, don't clear other text fields.
- When adding new action, and immediately switching configuration via comboBox, properly save/load actions for new configuration (don't remove action listener).


neilcsmith-net:
Thanks @AaronTheBaker but I believe this is already fixed in 13-rc1 and current master (which is why you're seeing merge conflicts).  Although not fixed in the same way listed.  I've closed NETBEANS-6058 - missed some of the duplicates of this.

PR #3495, opened 2022-01-24
[NETBEANS-6411] - Add javadoc for JDK 19 early access and JDK 18 GA

pepness:
Add entry to the `OFFICIAL_JAVADOC` Map for Java 19

- https://download.java.net/java/early_access/jdk19/docs/api/

Change Java 18 to GA
- https://docs.oracle.com/en/java/javase/18/docs/api/


mbien:
i think we could also switch JDK 18 docs to non-early access. Since NB 14 will be released when JDK 18 is already available.

mbien:
@pepness I just noticed you set the target to NB 13. For that you would have to base it on the delivery branch - master would be for NB 14. My comment above assumed it being for 14.

mbien (approved):
thanks, looks good. Please squash to a single commit before merge.

PR #3496, opened 2022-01-24
Change maven 384 dependencies to codehaus groupID

ebarboni:
During testing of org.netbeans.api:org-netbeans-api-maven:jar:RELEASE130 it appears that the pom is using dependencies that should come from codehaus groupId instead of sonatype.


neilcsmith-net (approved):

matthiasblaesing (approved):
Looks sane to me - thanks for catching this. There is a test tool in the build tool to check the hash, but not the artifact.

ebarboni:
@neilcsmith-net would it be possible to have that in the release130 branch for being able to regenerate all the artefacts.

neilcsmith-net:
@ebarboni planning to sync for rc2 later today.

neilcsmith-net:
OK, it's in release130 now.

PR #3499, opened 2022-01-25
Updating vsce to the 2.6.4 version.

dbalek:
Updating `vsce` to the 2.6.4 version to prevent security vulnerabilities reported during VSCode extension build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -652,7 +652,7 @@ matrix:
         - name: Build the Visual Studio Code extension for Java
           jdk: openjdk8
           before_install:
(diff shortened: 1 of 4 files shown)


sdedic (approved):

neilcsmith-net:
Only going on description, but is this one for delivery and NB13?

dbalek:
> Only going on description, but is this one for delivery and NB13?

Since the change updates the packaging tool used to build VSCode extension (not the extension itself) there is IMHO no need for backporting to NB13.

PR #3501, opened 2022-01-25
Update README with JDK 11 requirement.

neilcsmith-net:
Minimal update to README with JDK 11 requirement for build and JDK 11+ requirement for running. As this file gets included into source release, need to get it in to delivery.


mbien (approved):
L41+L42 could be refactored to one line (i am kidding its fine :) )

neilcsmith-net:
@mbien afterwards in master, hopefully. I don't think NB13 will build on JDK 17.

PR #3502, opened 2022-01-25
LSP ProjectView test + fixes

sdedic:
This PR focus on integration-testing the projects view. It checks basic scenarios: created files appear, deleted disappear and the package nodes restructure according to the occupied folders.

I moved certain registrations / layer folders from `nbcode` integration module to LSP module - these are harmless for IDE execution, but the LSP server (which can be potentially started inside NB IDE) does not function without those folders well. Also `foundProjects` view is defined in the LSP server but its critical cookie registrations were in the integration module - a different distribution.

During testing, I've encountered a NPE thrown when the client asks for details of a non-existing node: this is fixed by the last commit.


java/java.lsp.server/test/unit/data/gradle-hello/app/gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
JaroslavTulach, line 1: Ugly file, too ugly. Would it be to dare to ask you to create a Gradle project using the wizard/template in the IDE?

JaroslavTulach (approved):
It is sad that two biggest files in the change are `gradlew` wrappers! I'd prefer to generate them using "gradle init" which is available when one uses "New From Template" on the gradle project. I know you argued that JAR files should rather be in the repository and this is similar issue, so I leave the decision up to you.

java/java.lsp.server/test/unit/data/gradle-hello/app/gradlew
sdedic: ok, why not.

sdedic: project template used in 7cce3cf

dbalek (approved):
Looks fine.

PR #3504, opened 2022-01-25
Update FlatLaf from 2.0 to 2.0.1

DevCharly:
Unfortunately there is a memory leak in FlatLaf 2.0, which is fixed in 2.0.1.

The memory leak is in UI delegates for components `JPanel`, `JSeparator` and `JToolBar.Separator`.
There is always a reference from a static hash map to the **last created component** of that type.
So if e.g. a dialog (that uses `JPanel`) is closed, it can not released until a new `JPanel` is created somewhere else...

Memory leak issue: https://github.com/JFormDesigner/FlatLaf/issues/471

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-A45856E4115136977C662298BAC02F3B428082DB com.formdev:flatlaf:2.0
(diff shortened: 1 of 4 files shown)


vieiro (approved):

vieiro:
Perfect. Thanks, @DevCharly

vieiro:
Hi @neilcsmith-net , I added the "cherry-pick" tag to this one (is this correct?), for your consideration in next release candidate.

neilcsmith-net:
Thanks @DevCharly will be in rc3 now.

@vieiro no need for the label. Milestone and delivery base is enough.

PR #3506, opened 2022-01-25
Micronaut project creation fix

matthiasblaesing:
Micronaut projects are generated by the https://micronaut.io/launch/ service, which creates a ZIP file with the project contents.

The NetBeans integration unzips that file and invokes the gradle/maven build. Micronaut uses either the maven or gradle wrappers and on unixoid systems this wrapper needs to be executable. This is fixed in 9de1d4b.

In addition the user is informed, that he should be careful to trust the system used to generate the project file:

Bildschirmfoto-20220125211631-911x565


mbien (approved):
LGTM - the added warning is great too

neilcsmith-net (changes requested):
Thanks for getting ahead of me on this one.

Minor typo to fix.

The UI code is possibly more complicated than it needs to be.  Can you not use the NotificationLineSupport::setWarningMessage for this?

Still not particularly keen on download and execute as single action, but can live with it with the warning.  Pinging @vieiro for review too.

This has missed the boat for 13-rc2 anyway, so doesn't need merging until early next week.

enterprise/micronaut/src/org/netbeans/modules/micronaut/newproject/Bundle.properties
```diff
@@ -36,3 +36,4 @@ LBL_TestFramework=Test Framework:
 LBL_JUnit=JUnit
 LBL_Spock=Spock
 LBL_Kotest=Kotest
+BasePropertiesVisual.jLabel1.text=<html>Only finish this dialog if you trust the choosen service. Code from that service is downloaded and executed on this machine.
```
neilcsmith-net, line 39: Minor typo : ~~choosen~~ service ? chosen service.

neilcsmith-net (changes requested):

vieiro (approved):

vieiro:
Hi @neilcsmith-net . I can't really review this one, but a quick look reasonable.

Two questions:
- A) Are we setting the background color in that notice? It seems to me is reducing contrast. Can't tell how this looks in a dark mode.
- B) Is this expected to solve NETBEANS-6401 ? If so I'm missing setting an executable bit in "mvnw" after unzipping, right?

Thanks,
Antonio

neilcsmith-net:
Thanks @vieiro Wanted your review primarily on principle and text.

Execute bit is set at https://github.com/apache/netbeans/blob/9de1d4bafd8d602b646b4bc8595b7fb3fb3fb33a/enterprise/micronaut/src/org/netbeans/modules/micronaut/newproject/MicronautProjectWizardIterator.java#L237

Have similar concern on UI, hence wondering if NotificationLineSupport would work better and simplify the changes.

vieiro:
> Thanks @vieiro Wanted your review primarily on principle and text.
> 
> Execute bit is set at
> 
> https://github.com/apache/netbeans/blob/9de1d4bafd8d602b646b4bc8595b7fb3fb3fb33a/enterprise/micronaut/src/org/netbeans/modules/micronaut/newproject/MicronautProjectWizardIterator.java#L237
> 
> Have similar concern on UI, hence wondering if NotificationLineSupport would work better and simplify the changes.

Yep, I've seen the executable bit later. +1 to the NotificationLineSupport if possible.

matthiasblaesing:
Will have a look at NotificationLineSupport, but I had tested the display with different LAFs and I remember definitely testing DarkNimbus. My eyes were still ok afterwards, but maybe its even easier/looks better with NotificationLineSupport.

neilcsmith-net:
@matthiasblaesing actually, it might conflict with other messages being set by the wizard here - eg. the error messages in `valid(..)`.  Any immediate concern with that, then let's just fix the typo and keep as is.

matthiasblaesing:
I finally remembered, that I ran hard against `NotificationLineSupport`: It is good for very short texts and utterly fails for anyhing longer than a handful of words. This is also the case here (hint: the sentence does not end where it is cut here):

truncated_text

Here are screenshots in the various LAFs:

Metal:
base_metal

Nimbus:
base_nimbus

CDE/Motif:
base_cde_motif

GTK:
base_gtk

Flatlaf:
base_flatlaf_light

Flatflaf (Dark):
base_flatlaf_dark

Metal (Dark):
base_metal_dark

Nimbus (Dark):
base_nimbus_dark

mbien:
I think it looks really nice + this PR fixes an important issue. I suppose it could be make more DRY by updating the notification service and letting it use it like Neil suggested but I feel this would have low priority when taken in context. Updating the notification support would also require more testing since its used everywhere. Still +1 from me.

neilcsmith-net:
Thanks @matthiasblaesing and kudos for all the screenshots - above and beyond! :smile: Tested locally too.  Will merge.

> I ran hard against NotificationLineSupport: It is good for very short texts and utterly fails for anyhing longer than a handful of words.

Yes, you need to add `<br>` tags. I may look at a wrapping fix some time, but not for this!

@mbien it wasn't from a DRY perspective, at least not code anyway! Icons and colours spread across the code base tend to cause problems over time.  But this solves the immediate concerns so I'm merging.

neilcsmith-net (approved):
@matthiasblaesing do you want to squash and merge?  Or I can.  But please do to the PR and merge that so we don't lose the link.  Thanks.

mbien:
> @mbien it wasn't from a DRY perspective, at least not code anyway! Icons and colours spread across the code base tend to cause problems over time.

sure DRY is goes beyond code.

neilcsmith-net:
@mbien you're reading that the opposite way to how it was intended. Code-wise it wasn't (mostly) about DRY. Anyway, this change is good to go  - any follow up discussion can happen elsewhere if need be.

matthiasblaesing:
@neilcsmith-net squashed the commits I wanted squashed and adjusted the title of the second commit to refer to the JIRA issue. Feel free to merge when this gets green.

neilcsmith-net:
@matthiasblaesing thanks, yes, that split into two commits makes sense.  If you see it green before me, feel free to merge too.

PR #3512, opened 2022-01-26
Making build information a bit more up to date

JaroslavTulach:
The current `BUILD.md` info is a bit out of date. Rewording, updating, etc.


sdedic (approved):

entlicher (approved):

neilcsmith-net:
Wouldn't this (or variation of it) be better in delivery?  I've just pushed a few updates to the main README for similar reasons.  Source release docs should be as current as we can be.

PR #3516, opened 2022-01-26
Parse Indexables under a source root as a group while running EmbeddedIndexers

dbalek:
This is an attempt to fix the performance problem found while trying to open and index larger mx based project (GraalVM's Truffle). Because of an EmbeddingIndexer registered for the `text/x-java` mime type (`TestMethodFinderImpl`), thousands of the `javac` compiler instances are created to index the project sources (one instance per source root via `JavaCustomIndexer` and instance per source/indexable via `RepositoryUpdater.indexEmbeddings`.
This change tries to parse all sources/indexables under a source root as a group while running EmbeddedIndexers. Is saves thousands of the `javac` instances (only two instances per source root are created - one via `JavaCustomIndexer`, the second via `RepositoryUpdater.indexEmbeddings`) and approximately 20% of indexing time for the Truffle mx project.


JaroslavTulach (approved):
My understanding of the code changes is limited, but:
- reducing the number of `javac` instances by thousands is a great thing
- allowing people to use the same API without writing special case of Java support is great too

You have my emotional support and approval!

jlahoda (approved):
Seems sensible to me.

PR #3518, opened 2022-01-27
Exposing simple UI action to open next segment of a heap dump

JaroslavTulach:
I've created a simple library to generate `.hprof` files: [HeapDump](https://www.graalvm.org/tools/javadoc/org/graalvm/tools/insight/heap/HeapDump.html). It follows the [Java Profiler Heap Dump Format](http://hg.openjdk.java.net/jdk6/jdk6/jdk/raw-file/tip/src/share/demo/jvmti/hprof/manual.html) specification, however it allows more than regular JVMs do. Regular JVMs always dump the whole heap in a single round, while the [HeapDump](https://www.graalvm.org/tools/javadoc/org/graalvm/tools/insight/heap/HeapDump.html) allows one to generate _multiple segments_.

There already is support for reading these multisegment files in the Apache NetBeans Profiler library - it just isn't exposed in the heapwalker UI. This PR fixes that. The fix isn't "fancy", but it does its job. When an additional segment is detected, there is a little _link_ in the UI to open it in a new window. 

obrazek

Simple, but functional. I'd be delighted if the support for _multisegment_ `.hprof` files landed in the NetBeans 13.

### DIFF
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/CacheDirectory.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/CacheDirectory.java
@@ -28,16 +28,19 @@ import java.io.IOException;
  */
 class CacheDirectory {
     
-    private static final String DIR_EXT = ".hwcache";   // NOI18N
+    private static final String DIR_EXT = ".nbcache";   // NOI18N
     private static final String DUMP_AUX_FILE = "NBProfiler.nphd";   // NOI18N
     
     private final File.Factory io;
     private File cacheDirectory;
     
-    static CacheDirectory getHeapDumpCacheDirectory(File.Factory io, File heapDump) {
+    static CacheDirectory getHeapDumpCacheDirectory(File.Factory io, File heapDump, int segment) {
         String dumpName = heapDump.getName();
+        if (segment != 0) {
+            dumpName += "_" + segment;
+        }
         File parent = heapDump.getParentFile();
-        File dir = io.newFile(parent, dumpName+DIR_EXT);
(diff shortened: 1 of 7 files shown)


profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/CacheDirectory.java
```diff
@@ -28,16 +28,19 @@
  */
 class CacheDirectory {
     
-    private static final String DIR_EXT = ".hwcache";   // NOI18N
+    private static final String DIR_EXT = ".nbcache";   // NOI18N
```
JaroslavTulach, line 31: This change is useful to avoid clash between Apache NetBeans profiler library and VisualVM profiler library. Both of these libraries are derived from NetBeans 8.2 profiler library, but evolve independently. Sharing the same directory leads to errors like:
```
java.io.IOException: Invalid HPROF version 3 loaded 2
        at ...HprofHeap.<init>(HprofHeap.java)
        at ...HeapFactory.loadHeap(HeapFactory.java)
        at ...HeapFactory.createHeap(HeapFactory.java)
```

neilcsmith-net (approved):
Fine with me.  Feel free to merge when review complete as long as before next Tues.

matthiasblaesing (approved):
Looks ok to me.

JaroslavTulach:
The _Profiler native binary build-windows_ is repeatedly failing. Do you think it can be because of my changes? I have renamed the cached directory - that is the only (yet unlikely) thing which could cause the profiler binaries failures.

neilcsmith-net:
Looks like it might have been failing for a while.  It's only triggered rarely by path.  Just done a manual trigger on master.

PR #3522, opened 2022-01-27
build.xml and run.xml: fixed detection of 64 bit architecture for Java 9+ (on Windows)

DevCharly:
When using Java 9+ for debugging/running NetBeans from NetBeans on Windows,
then the 64-bit architecture of the JVM was not detected correctly and
`netbeans.exe` was used instead of `netbeans64.exe` to launch NetBeans.
This affects not only NetBeans core development, but also NetBeans module development.

This PR detects 32-bit JVMs and otherwise defaults to 64-bit.

See https://lists.apache.org/thread/5v5gb40tj3t4y06qtzxhjncowtm290t1

### DIFF
--- a/harness/apisupport.harness/release/run.xml
+++ b/harness/apisupport.harness/release/run.xml
@@ -54,8 +54,11 @@
         </delete>
         
         <!-- architecture of jvm on which app will run -->
-        <condition property="jvm.arch.string" value="64" else="">
-            <available file="${run.jdkhome}\jre\lib\amd64" />
+        <condition property="jvm.arch.string" value="" else="64">
+            <or>
+                <available file="${run.jdkhome}\bin\javaaccessbridge-32.dll" />      <!-- 32-bit Java 9+ -->
+                <available file="${run.jdkhome}\jre\bin\JavaAccessBridge-32.dll" />  <!-- 32-bit Java 8 -->
+            </or>
         </condition>  
         
         <!-- use nb.exe if exists (old platform), netbeans.exe otherwise -->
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@lkishalmi busy adding milestones as I'm reading :smile:  From a NetBeans 13 perspective, happy to merge, but wouldn't mind review from someone who actually uses this more (I use Windows rarely) given we're late in the day.

mbien (approved):
I like that 64bit is now the default, makes it less likely that this will break again.

The file probe however only checks windows dlls. Unfortunately I don't think this same trick will work on linux for example since it doesn't use the bridge libs. Still +1 from me since it makes the situation better.

edit: its probably not needed on linux since the launcher is a shell script and doesn't rely on the property i suppose

lkishalmi:
@neilcsmith-net Just saw this one requested against the delivery branch, so I've added you as a reviewer and set the milestone to NB13. The choice is in your hands.

Also, I'm not a Windows user as well.

DevCharly:
> The file probe however only checks windows dlls.

@mbien the Ant property `jvm.arch.string`, which this PR affects, is used only for the Windows "exe" launcher.
Also think that the old check works only on Windows because it uses backslash in the path.

Anyway, compared 32bit and 64bit **Linux** builds of Zulu and Bellsoft:

- Java 8: 32bit has directory `jre/lib/i386`, 64bit has `jre/lib/amd64` (Java 11 and 17 do not have these dirs)
- Java 11 and 17: have not found a file/directory to detect 32bit

For macOS I have not found 32bit builds.

mbien:
@DevCharly yeah I agree its not needed to cover all cases here since the property is only used on windows.

PR #3525, opened 2022-01-28
[NETBEANS-6425] Provide outline view for Groovy file in VSCode.

ppisl:
This is the outlineview implementation not only for Groovy files, but for all GSF languages that are/will be enabled in the Vs Code extension. At the moment it is for Groovy files and YAML files that have `application` or `bootstrap` prefix.

The implementation requires adding two interfaces to the LSP API. Clients can implement and register a `StructureProvider` into `MimeLookup` for a given mimetype. `StructureProvider` provides a list of top symbols to be displayed in the outline view. The symbol structure is composed of `StrctureElements`. 

StructureElements are then translated into `DocumentSymbol`s on the VSCode API side, which are displayed in the outline view and also in Go to a symbol (CTRL+SHIFT+O by default). `DocumentSymbol`  has two methods `getName()` and `getDetail()`, which both return `String` and the resulting compound text is displayed in the outline view, where the text returned by `getName()` is slightly larger than the text returned by `getDetail()`. On the other hand, only the text returned by `getName()` is displayed in **Go to a symbol**. The `DocumentSymbol` documentation says that `getName()` should return the name of the symbol and `getDetail()` should return the detail of the symbol, such as the signature of the function. 

Currently in outline view we display the signature as the symbol name and the type as the symbol detail for java files. This means that in Go to a symbol we display not only the names but the symbol, but also signatures with the types of the function parameters, which can affect the symbol search. You can see on this picture, what I mean: https://issues.apache.org/jira/secure/attachment/13039469/Screenshot%20from%202022-01-28%2011-05-32.png

On the other hand, the java extension offered by Microsoft does the same. However, in my opinion, `getName` should really return only the name of the symbols and this would make the search in Go to symbol easier. 

This PR also moves the existing functionality that provided outline view for java files outside the LSP server. The JavaStructureProvider implementation is now in the java.editor module, which already has a dependency on the LSP API.  Provider should logically belong to java.navigation module, but this module is not enabled for VsCode extension and if we enable it, it brings with it dependency on WindowsSystem.API, which is not sufficient now.

### DIFF
--- a/ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
+++ b/ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.api.lsp;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * StructureElement is a tree item that shows the structure of the source code.
+ * 
+ * @author Petr Pisl
+ * @since 1.8
+ */
+public interface StructureElement {
+   
+    /**
+     * Kind of the structure element. 
+     */
+    public static enum Kind {
+        File,
+	Module,
+	Namespace,
+	Package,
+	Class,
+	Method,
+	Property,
+	Field,
+	Constructor,
+	Enum,
+	Interface,
+	Function,
+	Variable,
+	Constant,
+	String,
+	Number,
+	Boolean,
+	Array,
+	Object,
+	Key,
+	Null,
+	EnumMember,
+	Struct,
+	Event,
+	Operator,
+	TypeParameter
+    }
+    
+    /**
+     * Tags are extra annotations that tweak the rendering of a symbol.
+     */
+    public static enum Tag {
+        Deprecated;
+    }
+    
+    /**
+     * The name is displayed in the structure tree and other ui.
+     * @return The name of this element.
+     */
+    public String getName();
+    
+    /**
+     * The start of offset range where this element should be selected and revealed 
+     * when selected.
+     * @return start offset of the selection
+     */
+    public int getSelectionStartOffset();
+    
+    /**
+     * The end of offset range where this element should be selected and revealed 
+     * when selected.
+     * @return end offset of the selection
+     */
+    public int getSelectionEndOffset();
+    
+    
+    /**
+     * The expanded range is offset range that is typically used to determine if the cursors
+     * inside the element to reveal in the element in the UI. 
+     * @return start of the enclosed range
+     */
+    public int getExpandedStartOffset();
+    
+    /**
+     * The expanded range is offset range that is typically used to determine if the cursors
+     * inside the element to reveal in the element in the UI. 
+     * @return end of the enclosed range
+     */
+    public int getExpandedEndOffset();
+    
+    /**
+     * Kind of this symbol
+     * @return Kind of this symbol
+     */
+    public Kind getKind();
+    
+    /**
+     * Tags for this element.
+     * @return list of tags
+     */
+    public Set<Tag> getTags();
+    
+   /**
+   * More detail for this symbol, e.g the signature of a function.If not provided the name is used.
+   * @return the detail text for this element
+   */
+    public String getDetail();
+    
+    /**
+     * Children of this element, e.g. method and fields of a class. 
+     * @return 
+     */
+    public List<? extends StructureElement> getChildren();
+    
+}
--- a/ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
+++ b/ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
@@ -20,111 +20,172 @@ package org.netbeans.api.lsp;
 
 import java.util.List;
 import java.util.Set;
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.modules.lsp.StructureElementAccessor;
 
 /**
  * StructureElement is a tree item that shows the structure of the source code.
- * 
+ *
  * @author Petr Pisl
  * @since 1.8
  */
-public interface StructureElement {
-   
+public final class StructureElement {
+
+    static {
+        StructureElementAccessor.setDefault(new StructureElementAccessor() {
+            @Override
+            public StructureElement createStructureElement(String name, String detail, int selectionStartOffset, int selectionEndOffset, int expandedStartOffset, int expandedEndOffset, Kind kind, Set<Tag> tags, List<StructureElement> children) {
+                return new StructureElement(name, detail, selectionStartOffset, selectionEndOffset, expandedStartOffset, expandedEndOffset, kind, tags, children);
+            }
+        });
+    }
+
+    private final String name;
+    private final String detail;
+    private final int selectionStartOffset;
+    private final int selectionEndOffset;
+    private final int expandedStartOffset;
+    private final int expandedEndOffset;
+    private final Kind kind;
+    private final Set<Tag> tags;
+    private final List<StructureElement> children;
+
     /**
-     * Kind of the structure element. 
+     * Kind of the structure element.
      */
     public static enum Kind {
         File,
-	Module,
-	Namespace,
-	Package,
-	Class,
-	Method,
-	Property,
-	Field,
-	Constructor,
-	Enum,
-	Interface,
-	Function,
-	Variable,
-	Constant,
-	String,
-	Number,
-	Boolean,
-	Array,
-	Object,
-	Key,
-	Null,
-	EnumMember,
-	Struct,
-	Event,
-	Operator,
-	TypeParameter
+        Module,
+        Namespace,
+        Package,
+        Class,
+        Method,
+        Property,
+        Field,
+        Constructor,
+        Enum,
+        Interface,
+        Function,
+        Variable,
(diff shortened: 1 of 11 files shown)


ppisl:
There are no tests yet, I need to write them.

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
```diff
@@ -0,0 +1,130 @@
...
+ * @since 1.8
+ */
+public interface StructureElement {
+   
+    /**
+     * Kind of the structure element. 
+     */
+    public static enum Kind {
+        File,
+	Module,
```
JaroslavTulach, line 37: Strange formatting. Space vs. tab? Using 8 spaces would be more consistent.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -813,97 +814,61 @@ protected void process(CompilationInfo arg0, Document arg1, SchedulerEvent arg2)
...
+        
+            String uri = params.getTextDocument().getUri();
+            FileObject file = fromURI(uri);
+            Document doc = server.getOpenedDocuments().getDocument(uri);
+            if (file == null || !(doc instanceof LineDocument)) {
                 return CompletableFuture.completedFuture(Collections.emptyList());
             }
-            List<Either<SymbolInformation, DocumentSymbol>> result = new ArrayList<>();
-            try {
-                js.runUserActionTask(cc -> {
```
JaroslavTulach, line 822: Removing this special Java code is a great step at the right direction. Thanks.

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * StructureElement is a tree item that shows the structure of the source code.
+ * 
+ * @author Petr Pisl
+ * @since 1.8
+ */
+public interface StructureElement {
```
JaroslavTulach, line 30: How do you want to evolve this type? You may need to add new getters in the future.

I't be better to follow the pattern of [Completion](https://bits.netbeans.org/dev/javadoc/org-netbeans-api-lsp/org/netbeans/api/lsp/Completion.html) in API and 
[CompletionCollector](https://bits.netbeans.org/dev/javadoc/org-netbeans-api-lsp/org/netbeans/spi/lsp/CompletionCollector.html) in SPI.

Turn `StructureElement` into `final class`.

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Resolves a structure tree of symbols at the given document.
+     *
+     * @param doc document on which to operate.
+     * @return a list of top level structure elements
+     * 
+     * @since 1.8
+     */
+    @NonNull
+    CompletableFuture<List<? extends StructureElement>> getStructure(@NonNull Document doc);
```
JaroslavTulach, line 56: You need a builder like [CompletionCollector.Builder](https://bits.netbeans.org/dev/javadoc/org-netbeans-api-lsp/org/netbeans/spi/lsp/CompletionCollector.Builder.html).

JaroslavTulach (commented):
Great refactoring, just get ready for evolution of `StructureElement`. And write at least one test!

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
ppisl: Oh, this was copy pase from other place. I will correct it.

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
ppisl: I was trying to create the sub nodes lazily. The tree doesn't appear all at once, only the top two levels. But in the end, the whole tree is sent in response, so from that perspective it doesn't matter. But for the future ....

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
ppisl: When we agree, that we don't need to create subnodes lazily, I will do it in this way.

ppisl: I was thinking about this again and I agree with @JaroslavTulach . I will change `StructureElement` to the final class and create a builder for it.

ppisl:
> There are no tests yet, I need to write them.

There is now test for the api.

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
ppisl: Done

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
```diff
@@ -0,0 +1,244 @@
...
+import javax.swing.text.Document;
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.modules.lsp.StructureElementAccessor;
+
+/**
+ * Interface for building structure of symbols at a given document.
+ * Implementations of the interface should be registered in MimeLookup.
+ * <pre>
+ *
+ *  {@code @MimeRegistration(mimeType = "text/foo", service = StructureProvider.class)
```
JaroslavTulach, line 36: Consider using `{@codesnippet}` tag.

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
```diff
@@ -0,0 +1,244 @@
...
+            return this;
+        }
+
+        /**
+         * Adds sub element for this structure element.
+         *
+         * @since 1.8
+         */
+        @NonNull
+        public Builder addChild(@NonNull StructureElement child) {
```
JaroslavTulach, line 210: Why `addChild` vs. `children`? It'd be enough to have `children(StructureElement... children)` and possibly `children(List<StructureElement> children)` that would both add few child(ren) to the list.

JaroslavTulach (approved):

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
ppisl: Corrected.

ide/api.lsp/src/org/netbeans/spi/lsp/StructureProvider.java
ppisl: I have changed this.

ide/api.lsp/src/org/netbeans/api/lsp/StructureElement.java
```diff
@@ -0,0 +1,191 @@
...
+            @Override
+            public StructureElement createStructureElement(String name, String detail, int selectionStartOffset, int selectionEndOffset, int expandedStartOffset, int expandedEndOffset, Kind kind, Set<Tag> tags, List<StructureElement> children) {
+                return new StructureElement(name, detail, selectionStartOffset, selectionEndOffset, expandedStartOffset, expandedEndOffset, kind, tags, children);
+            }
+        });
+    }
+
+    private final String name;
+    private final String detail;
+    private final int selectionStartOffset;
```
sdedic, line 45: A weird idea: if `javax.swing.text.Position` is used here, the structure could survive some edits between the time StructureElement is created and transofmed into LSP4J wire format -- and during the translation the document could be readlocked to protect the consistency. @JaroslavTulach  ?

JaroslavTulach: Do we want the _structure_ to represent read-only snapshot and be always re-created from scratch? Or do we want the _structure_ to be updated?

I believe we strive for the first approach. As such making the fields `final` more properly represents the _snapshot_ nature of the _structure_.

ide/csl.api/src/org/netbeans/modules/csl/navigation/GsfStructureProvider.java
```diff
@@ -0,0 +1,268 @@
...
+                int startOffset = (int)item.getPosition();
+                int lineStart = startOffset;
+                int lineEnd = (int)item.getEndPosition();
+                Builder builder = StructureProvider.newBuilder(item.getName(), convertKind(item.getKind()));
+                builder.selectionStartOffset(startOffset).selectionEndOffset(lineEnd);
+                createDetail(item, builder);
+                createTags(item, builder);
+
+                try {
+                    String prefix = doc.getText(lineStart, startOffset);
```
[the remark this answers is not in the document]
ppisl: This is useful for example for yaml files like:
```
martin:
  name: Martin D'vloper
|  job: Developer
  skill: Elite
```
The ui will select the `job` item, not `martin` structure (that will be selected if the job expanded offsets will be the sema lake selected offsets.

ide/csl.api/src/org/netbeans/modules/csl/navigation/GsfStructureProvider.java
```diff
@@ -0,0 +1,268 @@
...
+                    builder.expandedStartOffset(lineStart);
+                    builder.expandedEndOffset(lineEnd);
+                } else {
+                    if (lastElement.getExpandedEndOffset() < lineStart) {
+                        builder.expandedStartOffset(lineStart);
+                        builder.expandedEndOffset(lineEnd);
+                    } else if (lastElement.getExpandedStartOffset() <= lineStart && lineEnd == lastElement.getExpandedEndOffset()) {
+                        // The same line
+                        sElements.remove(lastElement);
+                        Builder leBuilder = StructureProvider.newBuilder(lastElement.getName(), lastElement.getKind());
```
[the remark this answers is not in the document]
ppisl: Will be in the next commit.

java/java.editor/src/org/netbeans/modules/editor/java/JavaStructureProvider.java
```diff
@@ -0,0 +1,314 @@
...
+        if (info.getElements().isDeprecated(el)) {
+            builder.addTag(StructureElement.Tag.Deprecated);
+        }
+        for (Element child : el.getEnclosedElements()) {
+            StructureElement jse = element2StructureElement(info, child);
+            if (jse != null) {
+                builder.children(jse);
+            }
+        }
+        if (path.getLeaf().getKind() == Tree.Kind.METHOD || path.getLeaf().getKind() == Tree.Kind.VARIABLE) {
```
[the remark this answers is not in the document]
ppisl: I have copied the original code here. I didn't changed the logic.

PR #3540, opened 2022-01-29
[jackpot] DefaultRuleUtilities::referencedIn fix for single variable matching

mbien:
`referencedIn(v, in)` did not match if `in` was a single variable.

example:
```java
for(int $i = $from; $i < $to; $i++) {
    $array[$i] = $n;
} 
:: !referencedIn($i, $n) && !matchesAny($n, "$i") /*&& more()*/ =>
java.util.Arrays.fill($array, $n);
;;
```
needed an extra `!matchesAny($n, "$i")` to make sure `$n` is not actually equal to `$i`. (i+1 would be matched by `referencedIn` but not i). After this PR `matchesAny` is no longer needed.

* added a small optimization which adds a fast path to the recursion if a result has been already determined.
* couldn't find tests for referencedIn so I added some.

btw the in-editor test integration for rule files is super cool


JaroslavTulach (approved):
I am amazed by your contributions to Jackpot, Michael. It is great to see someone else than @jlahoda taking care of the tool!

Writing a test as well as changing code is always good thing. I didn't even know we have `referencedIn` condition. Shouldn't we also improve the [documentation](https://netbeans.apache.org/jackpot/HintsFileFormat.html) to give a nice example of this (and possibly other) conditions?

mbien:
@JaroslavTulach I do what i can but a lot of this is still black magic to me :)

I am going to rebase this PR for delivery just in case it can be still integrated. This looks fairly low risk to me (I don't think the utility functions are even used in netbeans itself).

neilcsmith-net:
> This looks fairly low risk to me (I don't think the utility functions are even used in netbeans itself).

Not sure how to assess the latter point against bug criteria then, so will take your word on the first point and sneak it in rc3! :smile: Will merge tomorrow, give @jlahoda time to comment if needed.

java/java.hints.declarative/src/org/netbeans/modules/java/hints/declarative/conditionapi/Matcher.java
```diff
@@ -90,25 +91,38 @@ public Void scan(Tree tree, Void p) {
...
-        final Element e = ctx.ctx.getInfo().getTrees().getElement(ctx.getSingleVariable(variable));
+        final Trees trees = ctx.ctx.getInfo().getTrees();
+        final Element e = trees.getElement(ctx.getSingleVariable(variable));
 
         if (e == null) { //TODO: check also error
             return false;
         }
 
+        TreePath v = ctx.getSingleVariable(in);
+        if (v != null && e.equals(trees.getElement(v))) {
```
jlahoda, line 104: Good catch, but it would be better, I think, to move this test into the for loop (and operate over `tp`). We could have something like: `Arrays.asList(0, 1, b, 3)` with a pattern list `Arrays.asList($arrgs$)`.

mbien: haven't considered that. Fixed + added another test for the mentioned case.

jlahoda:
Looks great, BTW!

Issue #3526, opened 2022-01-28
Can't rename classes in newly created maven project


mbien:
### Apache NetBeans version

Latest release candidate

### What happened

see title, log:
```
SEVERE [org.openide.util.RequestProcessor]: Error in RequestProcessor org.netbeans.modules.refactoring.spi.impl.ParametersPanel$12
java.lang.IllegalArgumentException: file:/tmp/mavenproject1/src/test/java is not a valid classpath entry; it must end with a slash.
	at org.netbeans.modules.java.classpath.SimplePathResourceImplementation.verify(SimplePathResourceImplementation.java:90)
	at org.netbeans.modules.java.classpath.SimplePathResourceImplementation.verify(SimplePathResourceImplementation.java:43)
	at org.netbeans.modules.java.classpath.SimplePathResourceImplementation.<init>(SimplePathResourceImplementation.java:108)
	at org.netbeans.spi.java.classpath.support.ClassPathSupport.createResource(ClassPathSupport.java:54)
	at org.netbeans.spi.java.classpath.support.ClassPathSupport.createClassPath(ClassPathSupport.java:124)
	at org.netbeans.modules.testng.DefaultPlugin.getOppositeLocation(DefaultPlugin.java:242)
	at org.netbeans.modules.testng.DefaultPlugin.getTestLocation(DefaultPlugin.java:183)
	at org.netbeans.modules.testng.GoToOppositeAction.appliesTo(GoToOppositeAction.java:335)
	at org.netbeans.modules.refactoring.java.ui.RenamePanel$2.run(RenamePanel.java:158)
	at org.netbeans.modules.refactoring.java.ui.RenamePanel$2.run(RenamePanel.java:113)
	at org.netbeans.api.java.source.JavaSource$MultiTask.run(JavaSource.java:504)
	at org.netbeans.modules.parsing.impl.TaskProcessor.callUserTask(TaskProcessor.java:586)
	at org.netbeans.modules.parsing.api.ParserManager$UserTaskAction.run(ParserManager.java:132)
	at org.netbeans.modules.parsing.api.ParserManager$UserTaskAction.run(ParserManager.java:116)
	at org.netbeans.modules.parsing.impl.TaskProcessor$2.call(TaskProcessor.java:181)
	at org.netbeans.modules.parsing.impl.TaskProcessor$2.call(TaskProcessor.java:178)
[...]

PR #3541, opened 2022-01-30
#3526: Retain traling / for non-existing directories.

sdedic:
I found the cause for #3526  : the implementation was changed in [bd4fd0e](https://github.com/apache/netbeans/commit/bd4fd0e4f6f323c458869ca4d77f6efd8b24f012#diff-d217c6a1644bb679cd1327c9fb77e539735b405f6874f62760a7522bda248172R54) -- the delegated-to implementation behaves differently depending on if the configured directories exist.

I fixed the `getSourceRoots` impl to always return URIs that end with "/". I found out that there's an API method `FileUtilities.convertStringToUri` which behaves this way which is not sometimes (always ?) preferred. As it is just before RCs, I don't want to add another API method, so I left a note in the code instead & will keep the issue for myself to finish after NB13 release is done to avoid different content in `delivery` and `master`.

- Maybe I can just create a new PR with a proper API change for ` master`, if merging from/to `delivery` is not an issue -- @neilcsmith-net  ?
- @dbalek  is it OK for LSP operations ?

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/NbMavenProjectImpl.java
+++ b/java/maven/src/org/netbeans/modules/maven/NbMavenProjectImpl.java
@@ -26,6 +26,7 @@ import java.lang.ref.Reference;
 import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


mbien (approved):
fix looks solid. @sdedic is the hero of NB13. This would have been annoying if it would slip though.
(we can still improve it later)

dbalek (approved):
Looks fine to me.

neilcsmith-net:
Thanks @sdedic

Delivery merges to master after each RC - once we know build is OK. I'll merge this PR into delivery now. RC3 should be built Wednesday. So, merge to master should happen by end of week.  Please hold fire on further fixes until #3515 is merged or we'll get merge conflicts.

neilcsmith-net (approved):

neilcsmith-net:
@sdedic merge to master is now done!

Issue #3511, opened 2022-01-26
New class wizard superclass selection does not provide project classes as an option


jamoamo:
### Apache NetBeans version

Latest release

### What happened

When adding a new Java class to a project, one is presented with the option of browsing for a superclass for the new class to extend. In the "Find Type" dialog, when one enters the name of a class within the same project, the class is not presented as an option to select. The expectation is that classes from the same project should be options to be extended. This expected behaviour is seen when browsing for interfaces to implement.

### How to reproduce

Create a new Ant project [Java with Ant -> Java class library] (A Maven project can also reproduce the bug)
Right click source packages -> New -> Java Class
* Enter ParentClass as Class Name
* Enter a convenient package name
* Click Finish
Right click package -> New -> Java Class
* Enter ChildClass as Class Name
* Enter a convenient package name
* next to Superclass click "browse"
* In the "Find Type" dialog enter the name "ParentClass"
* The ParentClass class created earlier is not shown in the list

Attached is a screen recording showing how to reproduce
![NewClassSuperclassbug](https://user-images.githubusercontent.com/5406148/151182641-2eda116b-bcf8-421b-b18b-400f7069a1a4.gif)

### Did this work correctly in an earlier version?

- [ ] This used to work!

### Operating System

Windows 10 Home 21H2

### JDK

OpenJDK 64-bit server VM 17+35 

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

I tracked this down to org.netbeans.modules.java.project.ui.ExtensionAndImplementationVisualPanel.java and the method `isNotFinalClass(ElementHandle<TypeElement>)` which looks as follows:
```
private boolean isNotFinalClass(ElementHandle<TypeElement> typeHandle) {
        try {
            Class<?> clazz = Class.forName(typeHandle.getQualifiedName());
            return typeHandle.getKind() == ElementKind.CLASS && !Modifier.isFinal(clazz.getModifiers());
        } catch (ClassNotFoundException ex) {
        }
        return false;
    }
```
In my understanding, Class.forName will only load classes that are on the IDE's class path. Any other classes, such as those from the project under development, will result in a ClassNotFoundException which will cause the class to be ignored.

I would provide a PR for this but am not sure how best to get Modifier information about the class to determine if it is final.

### Are you willing to submit a pull request?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #3543, opened 2022-01-31
[#3511] workaround for extends/implements panel not showing any results.

mbien:
workaround for #3511 lets trade a big bug against a small bug to improve the situation :)

commit msg:
 - this is not a clean fix, it only improves the situation by making the
   wizzard actually usable
 - logic should not use Class.forName since it doesn't know anything
   about the opened projects or used JDKs
 - proper fix would either require calling ElementHandle.resolve off
   EDT and getting CompiltaionInfo from somewhere or to add modifiers
   to ElementHandle itself which would be a fairly large change


neilcsmith-net:
Seems a good workaround, at least for now. Better a UI is too lax than too strict.

I've linked the issue, although we might want a follow up to-do on this.  Please at least reference the issue in the description (title is optional) so we get linkage, or it's almost as bad as dealing with JIRA! :smile:

Will wait to merge until tomorrow to allow time for other reviewers, but would like to get in rc3 if it's going in.

neilcsmith-net (approved):

PR #3546, opened 2022-02-01
Download wallet

jhorvath:
Download DB wallet action is available in vscode. `NotifyDescriptor.InputLine` is displayed in vscode as QuickInput. `NotifyDescriptor.PasswordLine` displays password input dialog.

### DIFF
--- a/enterprise/cloud.oracle/nbproject/project.xml
+++ b/enterprise/cloud.oracle/nbproject/project.xml
@@ -117,7 +117,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>7.57</specification-version>
+                        <specification-version>7.59</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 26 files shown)


platform/openide.dialogs/src/org/openide/NotifyDescriptor.java
```diff
@@ -993,6 +994,25 @@ public Exception(Throwable detail, Object message) {
         }
     }
 
+    /**
+     * Notification providing a password input.
+     */
+    public static class PasswordLine extends InputLine {
```
JaroslavTulach, line 1000: Please make the class `final` - we don't want random subclasses. And in case you ask: The fact that `InputLine` isn't final is more a design mistake than something done on purpose.

JaroslavTulach (approved):
Please update dependency of `java.lsp.server` on `openide.dialogs` - it needs to use the most recent version otherwise there are goingto be linkage problems.

enterprise/cloud.oracle/nbproject/project.xml
```diff
@@ -188,6 +188,23 @@
                         <specification-version>1.10</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.openide.filesystems</code-name-base>
```
JaroslavTulach, line 192: The `cloud.oracle` is using API from dialogs 7.59 - does it require version 7.59?

java/java.lsp.server/nbproject/project.xml
```diff
@@ -537,7 +537,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>7.52</specification-version>
+                        <specification-version>7.59</specification-version>
```
JaroslavTulach, line 540: Good, thanks.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/UIContext.java
```diff
@@ -85,6 +86,9 @@ public static synchronized UIContext find() {
     protected abstract CompletableFuture<MessageActionItem> showMessageRequest(ShowMessageRequestParams msg);
     protected abstract void logMessage(MessageParams msg);
     protected abstract Message showStatusMessage(ShowStatusMessageParams msg);
+    protected CompletableFuture<String> showInputBox(ShowInputBoxParams params) {
```
JaroslavTulach, line 89: Actually, this is an API change in the API of `java.lsp.server` module. It would deserve module version increment, javadoc, `@since` tag. I am sorry for not documenting it in my commit that introduced this method.

dbalek (approved):

PR #3549, opened 2022-02-02
[NETBEANS-6177] Fix an issue the stop command is not sent

junichi11:
- https://issues.apache.org/jira/browse/NETBEANS-6177
- Related to NETBEANS-5080
- Use the `canceled` field instead of `detachRequest.set(true)` because
the stop command is not sent in `sendStopCommand()` if `detacheRequest` is `true`

### DIFF
--- a/php/php.dbgp/src/org/netbeans/modules/php/dbgp/DebugSession.java
+++ b/php/php.dbgp/src/org/netbeans/modules/php/dbgp/DebugSession.java
@@ -78,6 +78,7 @@ public class DebugSession extends SingleThread {
     private AtomicReference<DebuggerEngine> engine;
     private IDESessionBridge myBridge;
     private AtomicReference<String> myFileName;
+    private volatile boolean canceled;
 
     DebugSession(DebuggerOptions options, BackendLauncher backendLauncher) {
         commands = new LinkedList<>();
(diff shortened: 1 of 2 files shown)


junichi11:
@neilcsmith-net Can we merge this into the `delivery` branch? If it's too late, let's merge it in `NB14`.

php/php.dbgp/src/org/netbeans/modules/php/dbgp/DebugSession.java
```diff
@@ -124,6 +125,14 @@ public void run() {
                 } catch (Throwable e) {
                     log(e, Level.SEVERE);
                 }
+                if (canceled) {
+                    synchronized (commands) {
```
tmysik, line 129: The change looks good to me. However, the threading in this class is not documented and is not clear to me - I can see no synchronization of `commands` in `preprocess()` method, for example.

tmysik (approved):

tmysik:
@neilcsmith-net please merge it if it is OK for you.

Thank you, guys!

junichi11:

php/php.dbgp/src/org/netbeans/modules/php/dbgp/DebugSession.java
junichi11: You are right!

junichi11: Fixed only it: https://github.com/apache/netbeans/compare/79f57ddf2076828384d9b9a10ab082e44cb1c594..582c2c57d1e053833e7dafc7aedf34abc3b70187

neilcsmith-net:
OK, thanks, just caught in time for syncing for 13-rc3 so will merge.  Not sure if we'll have a 13-rc4 yet.

PR #3550, opened 2022-02-02
Improve Gradle projects problem reporting.

sdedic:
I have noticed that an exception thrown from Gradle script execution is slightly mis-presented
- a notification is fired, project is marked as having errors
- the "Project Problems" dialog does not appear (sometimes)
- "Do priming build" fixable problem appears - but can't fix anything as the script is broken
- The **real** problem does not appear in the "Project problems" dialog

The notification itself is mangled: part of the text is duplicated (see the picture). 
gradle-notification

I have tried to change that. I added some exception stack processing to ignore / reorder the information; the resulting notification is visible here:
gradle-notification2

In addition, I have change the condition for replacing loaded projects, so that 'problematic' project instance is retained - then the project infrastructure starts to work properly, displaying "project problems" dialog. In some subsequent PR, I would like to remove Gradle's special code that makes a Notification and change the core project ui to do that - for all projects.

The Project UI 'broken references' dialog was changed to detect & respect possible HTML formatting: will change `JTextPane` content type accordingly, which results in much nicer formatting. Gradle's `GradleProjectProblemProvider` uses `<br/>` but there was noone interpreting HTML on the receiving side.

Gradle's 'priming build required' Problem is now only displayed if there are no errors reported. It works well for broken buildscripts, but perhaps may need more clever test, or maybe even an API change (i.e. a non-fatal Problem that does not prevent the script from being executed may allow priming build to happen).

This PR also adds "Resolve project problems" standard Project UI action to **Gradle + Maven** projects; this action declares 'hide-if-disabled', so it will not clutter the project context menu unless there are actually some problems on the project.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/GradleProjectProblemProvider.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/GradleProjectProblemProvider.java
@@ -80,12 +80,16 @@ public class GradleProjectProblemProvider implements ProjectProblemsProvider {
     public Collection<? extends ProjectProblem> getProblems() {
         List<ProjectProblem> ret = new ArrayList<>();
         GradleProject gp = project.getLookup().lookup(NbGradleProjectImpl.class).getGradleProject();
-        if (gp.getQuality().notBetterThan(EVALUATED)) {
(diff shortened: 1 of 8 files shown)


lkishalmi (approved):
Nice work!

lkishalmi:
I'm creating a new netbeans-dev snap today and would like to see this one in there.
Thank you!

PR #3553, opened 2022-02-03
[cnd] 3.1 makeproject compile_commands.json

vieiro:
Automatically generate compile_commands.json for make based projects.

Makefile based projects generate now a compile_commands.json compilation database with the proper project configuration, as defined in https://clang.llvm.org/docs/JSONCompilationDatabase.html this is needed for both clangd anc ccls LSP servers. 

A specific MakeProjectLookupProvider adds this feature to all make projects, background tasks run on demand, after a coalescing period, and for small projects this runs in < 100ms. on my box.

### DIFF
--- a/cnd/cnd.lsp/src/org/netbeans/modules/cnd/lsp/compilationdb/ClangCDBSupport.java
+++ b/cnd/cnd.lsp/src/org/netbeans/modules/cnd/lsp/compilationdb/ClangCDBSupport.java
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.cnd.lsp.compilationdb;
+
+import java.util.List;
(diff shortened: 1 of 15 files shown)


cnd/cnd.lsp/src/org/netbeans/modules/cnd/lsp/compilationdb/ClangCDBSupport.java
```diff
@@ -0,0 +1,219 @@
...
+            ClangCDBGenerationTask task = null;
+            try {
+                task = new ClangCDBGenerationTask(this, pendingCauses);
+            } catch (IllegalArgumentException iae) {
+                LOG.log(Level.SEVERE, "Unable to create ClangCDBGenerationTask: {0}", iae.getMessage());
+                return;
+            }
+            // We fire the regeneration task after a COALESCING_DELAY_MS milliseconds, to
+            // coalesce requests (for instance, when many files are added or removed).
+            CLANG_CDB_PROCESSOR.schedule(task, COALESCING_DELAY_MS, TimeUnit.MILLISECONDS);
```
matthiasblaesing, line 207: I'm not sure this is what you want. Jan made me aware of a mode of the request processor when working on the LSPBindings.  The background tasks are created: 

https://github.com/apache/netbeans/blob/48ff3e7790554fffae05eab5621a9f1271642951/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java#L444-L457

and then scheduled to be run:

https://github.com/apache/netbeans/blob/48ff3e7790554fffae05eab5621a9f1271642951/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java#L475-L477

The important part is here:
- if `#scheduleBackgroundTask` is called while the task is already waiting, execution is again postponed
- if the task is finished, it is scheduled again
- if the task already running it is left to be running to completion, but a new run is scheduled

https://bits.netbeans.org/12.0/javadoc/org-openide-util/org/openide/util/RequestProcessor.Task.html#schedule-int-

For the LSPBindings this means, that the updates in NetBeans happen, when the situation is settled. This might be a sane approach here too.

matthiasblaesing (commented):
Did I get this right, that the Make support is already in place and one of the projects imported and you are using it to build the compilation database from this?

I'm not a native expert, but this would match my understanding of makes approach (declaring what source files are needed to build what output file and how).

Looks sane to me (only eyeballed), but I left an inline commit, you might want to consider.

vieiro:
Ok. Let's move on then.

PR #3555, opened 2022-02-03
Hint wizzard module dependency fix.

mbien:
New "Java Hint" Wizzard added test dependencies to the pom which did not exist (`org-netbeans-modules-parsing-nb` and `org-netbeans-modules-projectapi-nb`).

PR changes it to `org-netbeans-modules-parsing-api` and `org-netbeans-modules-projectapi`.


matthiasblaesing (approved):
Change looks sane to me.

PR #3558, opened 2022-02-05
maven indexing lucene renovation

mbien:
Check if the maven repo indexing module can be upgraded from apache lucene 5.x to 9.x.
 - lucene 9.x requires Java 11+, so we have to stick with 8.x
 - upgrading lucene requires upgrading `maven-indexer` to 6.10+
 - unfortunately it has to be done in one step since there don't appear to be any `maven-indexer` releases compatible with lucene 6.x or 7.x.

note:
- code cleanup has its own commit, the interesting changes are in the other commits

progress:

- [x] wait for release of maven-indexer 6.10 (https://github.com/apache/maven-indexer/pull/169) (~~using local SNAPSHOT atm~~)
- [x] `ClassDependencyIndexCreatorTest.testFindClassUsages()` fails with no results found, other tests are green.
- [x] investigate assertion error in `BooleanQuery.hashCode` / lucene -> https://issues.apache.org/jira/browse/LUCENE-10431 -> https://github.com/apache/netbeans/pull/3558#issuecomment-1086785298
- [x] investigate index out of bounds exception during local repo indexing / maven-indexer -> https://issues.apache.org/jira/browse/MINDEXER-144 + workaround found
- [x] `maven.apisupport` tests were failing since they loaded the wrong lucene jars (3.x) while using loading maven-indexer 6.1:

```
...
[26,138s][info][class,load] org.apache.lucene.store.SimpleFSDirectory source: file:/home/mbien/NetBeansProjects/netbeans/nbbuild/netbeans/ide/modules/ext/lucene-core-3.6.2.jar
[26,138s][info][class,load] org.apache.lucene.search.BooleanQuery$TooManyClauses source: file:/home/mbien/NetBeansProjects/netbeans/nbbuild/netbeans/ide/modules/ext/lucene-core-3.6.2.jar
[26,138s][info][class,load] org.apache.maven.index.ArtifactInfoFilter source: file:/home/mbien/NetBeansProjects/netbeans/nbbuild/netbeans/java/modules/ext/maven/indexer-core-6.1.1.jar
[26,138s][info][class,load] org.apache.maven.index.updater.WagonHelper$WagonFetcher source: file:/home/mbien/NetBeansProjects/netbeans/nbbuild/netbeans/java/modules/ext/maven/indexer-core-6.1.1.jar
...
```
- [x] the same `maven.apisupport` tests are now timing out since they decided to download and index the entire maven central repo

### DIFF
--- a/apisupport/maven.apisupport/nbproject/project.xml
+++ b/apisupport/maven.apisupport/nbproject/project.xml
@@ -25,6 +25,16 @@
         <data xmlns="http://www.netbeans.org/ns/nb-module-project/3">
             <code-name-base>org.netbeans.modules.maven.apisupport</code-name-base>
             <module-dependencies>
+                <!-- keep in front of any module which uses lucene, maven indexer uses it's own verison  -->
+                <dependency>
+                    <code-name-base>org.netbeans.modules.maven.indexer</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>2.11</specification-version>
+                    </run-dependency>
+                </dependency>
                 <dependency>
                     <code-name-base>org.netbeans.api.annotations.common</code-name-base>
                     <build-prerequisite/>
@@ -163,15 +173,6 @@
                     <run-dependency>
                         <release-version>2</release-version>
                         <specification-version>2.54</specification-version>
-                    </run-dependency>
-                </dependency>
-                <dependency>
-                    <code-name-base>org.netbeans.modules.maven.indexer</code-name-base>
-                    <build-prerequisite/>
-                    <compile-dependency/>
-                    <run-dependency>
-                        <release-version>2</release-version>
-                        <specification-version>2.11</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenNbModuleImpl.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenNbModuleImpl.java
@@ -78,10 +78,6 @@ import org.openide.filesystems.FileSystem;
 import org.openide.filesystems.FileUtil;
 import org.openide.filesystems.XMLFileSystem;
 import org.openide.modules.SpecificationVersion;
-import org.openide.util.RequestProcessor;
-
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  *
@@ -93,10 +89,7 @@ public class MavenNbModuleImpl implements NbModuleProvider {
     
     private final Project project;
     private final DependencyAdder dependencyAdder = new DependencyAdder();
-    private static final RequestProcessor RP = new RequestProcessor(MavenNbModuleImpl.class);
-    
-    private final RequestProcessor.Task tsk = RP.create(dependencyAdder);
-    
+
     public static final String MAVEN_CENTRAL = "central";
     public static final String APACHE_SNAPSHOT_REPO_ID = "apache.snapshots";
     // this repository is not good anymore, dev-SNAPSHOT version are buil on apache snapshot
@@ -110,11 +103,9 @@ public class MavenNbModuleImpl implements NbModuleProvider {
     public static final String GROUPID_MOJO = "org.codehaus.mojo";
     public static final String GROUPID_APACHE = "org.apache.netbeans.utilities";
     public static final String NBM_PLUGIN = "nbm-maven-plugin";
-    private static final String LATEST_NBM_PLUGIN_VERSION = "4.3";
+    static final String LATEST_NBM_PLUGIN_VERSION = "4.7";
 
     public static final String NETBEANSAPI_GROUPID = "org.netbeans.api";
-
-    private static final Logger LOG = Logger.getLogger("org.netbeans.modules.maven.apisupport.MavenNbModuleImpl");
 
     /** Creates a new instance of MavenNbModuleImpl 
      * @param project 
@@ -128,23 +119,17 @@ public class MavenNbModuleImpl implements NbModuleProvider {
                 RepositoryPreferences.getInstance().getRepositoryInfoById(MAVEN_CENTRAL));
     }
 
+    /**
+     * Returns the latest known version of the NetBeans maven plugin.
+     * This method will not wait for the index to be downloaded, it will return a default value instead.
+     */
     public static String getLatestNbmPluginVersion() {
-        try {
-            RepositoryQueries.Result<NBVersionInfo> versionsResult = RepositoryQueries.getVersionsResult(GROUPID_APACHE, NBM_PLUGIN, null);
+        RepositoryQueries.Result<NBVersionInfo> versionsResult = RepositoryQueries.getVersionsResult(GROUPID_APACHE, NBM_PLUGIN, null);
 
-            if (versionsResult.isPartial()) {
-                versionsResult.waitForSkipped();
-            }
-
-            // Versions are sorted in descending order
-            List<NBVersionInfo> results = versionsResult.getResults();
-            if (!results.isEmpty()) {
-                return results.get(0).getVersion();
-            }
-        }
-        catch (NullPointerException ex) {
-            // This exceptions occurs during unit tests so default to LATEST_NBM_PLUGIN_VERSION
-            LOG.log(Level.WARNING, "Unable to get latest nbm-maven-plugin version number");
+        // Versions are sorted in descending order
+        List<NBVersionInfo> results = versionsResult.getResults();
+        if (!results.isEmpty()) {
+            return results.get(0).getVersion();
         }
         return LATEST_NBM_PLUGIN_VERSION;
     }
--- a/java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java
+++ b/java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java
@@ -1,0 +1,371 @@
+// <editor-fold desc="Source copied form Lucene 8.11.1, see secretInit() for patch notes" defaultstate="collapsed">
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+
+import java.io.IOException;
+import java.util.Objects;
+import java.util.logging.Logger;
+
+import org.apache.lucene.index.FilteredTermsEnum; // javadocs
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.SingleTermsEnum;   // javadocs
+import org.apache.lucene.index.Term;
+import org.apache.lucene.index.TermStates;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.search.BooleanQuery.Builder;
+import org.apache.lucene.util.AttributeSource;
+import org.apache.lucene.util.Version;
+
+/**
+ * An abstract {@link Query} that matches documents
+ * containing a subset of terms provided by a {@link
+ * FilteredTermsEnum} enumeration.
+ *
+ * <p>This query cannot be used directly; you must subclass
+ * it and define {@link #getTermsEnum(Terms,AttributeSource)} to provide a {@link
+ * FilteredTermsEnum} that iterates through the terms to be
+ * matched.
+ *
+ * <p><b>NOTE</b>: if {@link #setRewriteMethod} is either
+ * {@link #CONSTANT_SCORE_BOOLEAN_REWRITE} or {@link
+ * #SCORING_BOOLEAN_REWRITE}, you may encounter a
+ * {@link BooleanQuery.TooManyClauses} exception during
+ * searching, which happens when the number of terms to be
+ * searched exceeds {@link
+ * BooleanQuery#getMaxClauseCount()}.  Setting {@link
+ * #setRewriteMethod} to {@link #CONSTANT_SCORE_REWRITE}
+ * prevents this.
+ *
+ * <p>The recommended rewrite method is {@link
+ * #CONSTANT_SCORE_REWRITE}: it doesn't spend CPU
+ * computing unhelpful scores, and is the most
+ * performant rewrite method given the query. If you
+ * need scoring (like {@link FuzzyQuery}, use
+ * {@link TopTermsScoringBooleanQueryRewrite} which uses
+ * a priority queue to only collect competitive terms
+ * and not hit this limitation.
+ *
+ * Note that org.apache.lucene.queryparser.classic.QueryParser produces
+ * MultiTermQueries using {@link #CONSTANT_SCORE_REWRITE}
+ * by default.
+ */
+public abstract class MultiTermQuery extends Query {
+  protected final String field;
+  protected RewriteMethod rewriteMethod = CONSTANT_SCORE_REWRITE;
+
+  /** Abstract class that defines how the query is rewritten. */
+  public static abstract class RewriteMethod {
+    public abstract Query rewrite(IndexReader reader, MultiTermQuery query) throws IOException;
+    /**
+     * Returns the {@link MultiTermQuery}s {@link TermsEnum}
+     * @see MultiTermQuery#getTermsEnum(Terms, AttributeSource)
+     */
+    protected TermsEnum getTermsEnum(MultiTermQuery query, Terms terms, AttributeSource atts) throws IOException {
+      return query.getTermsEnum(terms, atts); // allow RewriteMethod subclasses to pull a TermsEnum from the MTQ
+    }
+  }
+
+  /** A rewrite method that first creates a private Filter,
+   *  by visiting each term in sequence and marking all docs
+   *  for that term.  Matching documents are assigned a
+   *  constant score equal to the query's boost.
+   *
+   *  <p> This method is faster than the BooleanQuery
+   *  rewrite methods when the number of matched terms or
+   *  matched documents is non-trivial. Also, it will never
+   *  hit an errant {@link BooleanQuery.TooManyClauses}
+   *  exception.
+   *
+   *  @see #setRewriteMethod */
+  public static final RewriteMethod CONSTANT_SCORE_REWRITE = new RewriteMethod() {
+    @Override
+    public Query rewrite(IndexReader reader, MultiTermQuery query) {
+      return new MultiTermQueryConstantScoreWrapper<>(query);
+    }
+  };
+
+  /** A rewrite method that first translates each term into
+   *  {@link BooleanClause.Occur#SHOULD} clause in a
+   *  BooleanQuery, and keeps the scores as computed by the
+   *  query.  Note that typically such scores are
+   *  meaningless to the user, and require non-trivial CPU
+   *  to compute, so it's almost always better to use {@link
+   *  #CONSTANT_SCORE_REWRITE} instead.
+   *
+   *  <p><b>NOTE</b>: This rewrite method will hit {@link
+   *  BooleanQuery.TooManyClauses} if the number of terms
+   *  exceeds {@link BooleanQuery#getMaxClauseCount}.
+   *
+   *  @see #setRewriteMethod */
+  public final static RewriteMethod SCORING_BOOLEAN_REWRITE = ScoringRewrite.SCORING_BOOLEAN_REWRITE;
+
+  /** Like {@link #SCORING_BOOLEAN_REWRITE} except
+   *  scores are not computed.  Instead, each matching
+   *  document receives a constant score equal to the
+   *  query's boost.
+   *
+   *  <p><b>NOTE</b>: This rewrite method will hit {@link
+   *  BooleanQuery.TooManyClauses} if the number of terms
+   *  exceeds {@link BooleanQuery#getMaxClauseCount}.
+   *
+   *  @see #setRewriteMethod */
+  public final static RewriteMethod CONSTANT_SCORE_BOOLEAN_REWRITE = ScoringRewrite.CONSTANT_SCORE_BOOLEAN_REWRITE;
+
+  /**
+   * A rewrite method that first translates each term into
+   * {@link BooleanClause.Occur#SHOULD} clause in a BooleanQuery, and keeps the
+   * scores as computed by the query.
+   *
+   * <p>
+   * This rewrite method only uses the top scoring terms so it will not overflow
+   * the boolean max clause count.
+   *
+   * @see #setRewriteMethod
+   */
+  public static final class TopTermsScoringBooleanQueryRewrite extends TopTermsRewrite<BooleanQuery.Builder> {
+
+    /**
+     * Create a TopTermsScoringBooleanQueryRewrite for
+     * at most <code>size</code> terms.
+     * <p>
+     * NOTE: if {@link BooleanQuery#getMaxClauseCount} is smaller than
+     * <code>size</code>, then it will be used instead.
+     */
+    public TopTermsScoringBooleanQueryRewrite(int size) {
+      super(size);
+    }
+
+    @Override
+    protected int getMaxSize() {
+      return BooleanQuery.getMaxClauseCount();
+    }
+
+    @Override
+    protected BooleanQuery.Builder getTopLevelBuilder() {
+      return new BooleanQuery.Builder();
+    }
+
+    @Override
+    protected Query build(Builder builder) {
+      return builder.build();
+    }
+
+    @Override
+    protected void addClause(BooleanQuery.Builder topLevel, Term term, int docCount, float boost, TermStates states) {
+      final TermQuery tq = new TermQuery(term, states);
+      topLevel.add(new BoostQuery(tq, boost), BooleanClause.Occur.SHOULD);
+    }
+  }
+
+  /**
+   * A rewrite method that first translates each term into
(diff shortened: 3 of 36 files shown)


mbien:
This ended up being more interesting than anticipated.

It turned out that the reason for `findClassUsages()` not returning any hits was `maven-indexer`producing lower case based queries, while `ClassDependencyIndexCreator` encodes class dependencies in searchable Base64 encoded CRC32 Strings. This worked for some reason with old lucene (maybe searches were case insensitive?) but fails now after the lucene upgrade. But even if it worked before, it is likely still a bug since it could produce false positives if I see this correctly.

This [hack](https://github.com/apache/netbeans/pull/3558/commits/98d4f336875323f0ed05b00d0241189e61daf55b) simulates the old behavior and lets all tests pass locally.

potential fix: encode with Base32 instead (in lowercase) or workaround maven-indexer which uses toLowerCase [everywhere](https://github.com/apache/maven-indexer/blob/f09e411d6ced2027856fd24e0292e6ea57b9ff1b/indexer-core/src/main/java/org/apache/maven/index/DefaultQueryCreator.java) (custom query instead of using indexer.constructQuery() as seen below?)
https://github.com/apache/netbeans/blob/98d4f336875323f0ed05b00d0241189e61daf55b/java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java#L170-L175

ebarboni:
maven indexer 6.1.1 is in vote at the time writing.

mbien:
regarding the timeout of `NBMNativeMWITest`
The junit tests expected the query to fail with a NPE in the indexer code
https://github.com/apache/netbeans/blob/c084119009d2e0f736f225d706bc1827af283501/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenNbModuleImpl.java#L145-L149

post migration, the NPE does not occur. The query appears to work. However, this causes the download+indexing of the maven central repo. Which causes a timeout after 3 minutes.

So my question is: was the test originally supposed to download the index?

suppressed exception for reference (pre-migration):
```
java.lang.NullPointerException: Cannot invoke "org.apache.maven.index.Indexer.createIndexingContext(String, String, java.io.File, java.io.File, String, String, boolean, boolean, java.util.List)" because "this.indexer" is null
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.addIndexingContextForced(NexusRepositoryIndexerImpl.java:272)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.loadIndexingContext2(NexusRepositoryIndexerImpl.java:335)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.access$700(NexusRepositoryIndexerImpl.java:122)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl$6.run(NexusRepositoryIndexerImpl.java:940)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl$6.run(NexusRepositoryIndexerImpl.java:937)
	at org.netbeans.modules.openide.util.DefaultMutexImplementation.writeAccess(DefaultMutexImplementation.java:229)
	at org.openide.util.Mutex.writeAccess(Mutex.java:252)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.iterate(NexusRepositoryIndexerImpl.java:937)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.getVersions(NexusRepositoryIndexerImpl.java:1153)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.getVersions(NexusRepositoryIndexerImpl.java:1148)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.lambda$getVersionsResult$4(RepositoryQueries.java:314)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.getQueryResult(RepositoryQueries.java:217)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.getVersionsResult(RepositoryQueries.java:311)
	at org.netbeans.modules.maven.apisupport.MavenNbModuleImpl.getLatestNbmPluginVersion(MavenNbModuleImpl.java:133)
	at org.netbeans.modules.maven.apisupport.NBMNativeMWI$AdditionalOperations.performOperation(NBMNativeMWI.java:250)
	at org.netbeans.modules.maven.apisupport.NBMNativeMWI$AdditionalOperations.performOperation(NBMNativeMWI.java:141)
	at org.netbeans.modules.maven.model.Utilities.performPOMModelOperations(Utilities.java:318)
	at org.netbeans.modules.maven.spi.newproject.CreateProjectBuilder.create(CreateProjectBuilder.java:146)
	at org.netbeans.modules.maven.apisupport.NBMNativeMWI.instantiate(NBMNativeMWI.java:69)
	at org.netbeans.modules.maven.apisupport.NBMNativeMWITest.testPathNoParent(NBMNativeMWITest.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at org.netbeans.junit.NbTestCase.runBare(NbTestCase.java:495)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at org.netbeans.junit.NbTestCase.run(NbTestCase.java:291)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)

```

mbien:
everything green after brute forcing job 13 :)

still some work left

mbien:
switched to lowercase base32 encoding since I couldn't find any other way to make it work.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/CustomArtifactContextProducer.java
```diff
@@ -36,6 +37,16 @@ public CustomArtifactContextProducer(ArtifactPackagingMapper mapper) {
...
 
+    @Override
+    protected Gav getGavFromPath(IndexingContext context, String repositoryPath, String artifactPath) {
+        // filter out maven-metadata[-repo_name].xml to avoid IndexOutOfBoundsExceptions in
+        // org.apache.maven.index.artifact.M2GavCalculator.getSnapshotGav(M2GavCalculator.java:187)
+        if (artifactPath.endsWith(".xml") && artifactPath.substring(artifactPath.lastIndexOf("/")+1).startsWith("maven-metadata")) {
+            return null;
+        }
+        return super.getGavFromPath(context, repositoryPath, artifactPath);
+    }
```
mbien, line 48: added a workaround for the IOOBE in maven-indexer.

issue filed under:
https://issues.apache.org/jira/browse/MINDEXER-144

ebarboni:
Hi @mbien, I'm not sure my issue is related to your task but the Apache NetBeans indexer was not able to index Apache Archiva repository for a long time. It seems that your patch make it working again. 
Typical stacktrace.
``
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Downloaded maven index file has size 38?912 (zipped). The usable space in **mydrive** is 348?321?529?856.
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: could not (re-)index **arepoid**
java.lang.IllegalArgumentException: The legacy format is no longer supported by this version of maven-indexer.
	at org.apache.maven.index.updater.DefaultIndexUpdater.loadIndexDirectory(DefaultIndexUpdater.java:210)
	at org.apache.maven.index.updater.DefaultIndexUpdater.access$300(DefaultIndexUpdater.java:76)
	at org.apache.maven.index.updater.DefaultIndexUpdater$LuceneIndexAdaptor.setIndexFile(DefaultIndexUpdater.java:518)
	at org.apache.maven.index.updater.DefaultIndexUpdater.fetchAndUpdateIndex(DefaultIndexUpdater.java:752)
	at org.apache.maven.index.updater.DefaultIndexUpdater.fetchAndUpdateIndex(DefaultIndexUpdater.java:161)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.indexLoadedRepo(NexusRepositoryIndexerImpl.java:523)
Caused: java.io.IOException: Failed to load maven-index for: **repourl**
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.indexLoadedRepo(NexusRepositoryIndexerImpl.java:536)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.access$200(NexusRepositoryIndexerImpl.java:122)
[catch] at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl$3.run(NexusRepositoryIndexerImpl.java:654)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl$3.run(NexusRepositoryIndexerImpl.java:647)
	at org.netbeans.modules.openide.util.DefaultMutexImplementation.writeAccess(DefaultMutexImplementation.java:229)
	at org.openide.util.Mutex.writeAccess(Mutex.java:252)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.indexRepo(NexusRepositoryIndexerImpl.java:647)
	at org.netbeans.modules.maven.indexer.api.RepositoryIndexer.indexRepo(RepositoryIndexer.java:42)
	at org.netbeans.modules.maven.repository.RepositoryNode$RefreshIndexAction$1.run(RepositoryNode.java:228)
``

mbien:
@ebarboni thanks Eric for testing. That is good news. This PR was a little bit of a minefield :)

mbien:
for future reference, the assertion error in BooleanQuery.hashcode() attached below.

https://github.com/apache/netbeans/blob/04fa8fba812566a211462fc3eef73597fbf3a975/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java#L1451-L1457

can be mitigated by not changing the rewrite method to `MultiTermQuery.CONSTANT_SCORE_BOOLEAN_REWRITE` recursively via setBooleanRewrite(query).

```
java.lang.AssertionError
	at org.apache.lucene.search.BooleanQuery.hashCode(BooleanQuery.java:614)
	at java.base/java.util.Objects.hashCode(Objects.java:103)
	at java.base/java.util.HashMap$Node.hashCode(HashMap.java:298)
	at java.base/java.util.AbstractMap.hashCode(AbstractMap.java:527)
	at org.apache.lucene.search.Multiset.hashCode(Multiset.java:119)
	at java.base/java.util.EnumMap.entryHashCode(EnumMap.java:717)
	at java.base/java.util.EnumMap.hashCode(EnumMap.java:709)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4498)
	at java.base/java.util.Objects.hash(Objects.java:133)
	at org.apache.lucene.search.BooleanQuery.computeHashCode(BooleanQuery.java:597)
	at org.apache.lucene.search.BooleanQuery.hashCode(BooleanQuery.java:611)
	at java.base/java.util.HashMap.hash(HashMap.java:340)
	at java.base/java.util.HashMap.put(HashMap.java:612)
	at org.apache.lucene.search.Multiset.add(Multiset.java:82)
	at org.apache.lucene.search.BooleanQuery.<init>(BooleanQuery.java:154)
	at org.apache.lucene.search.BooleanQuery.<init>(BooleanQuery.java:42)
	at org.apache.lucene.search.BooleanQuery$Builder.build(BooleanQuery.java:133)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl$1.run(NexusRepositoryIndexerImpl.java:1457)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.lambda$iterate$11(NexusRepositoryIndexerImpl.java:890)
	at org.netbeans.modules.openide.util.DefaultMutexImplementation.writeAccess(DefaultMutexImplementation.java:229)
	at org.openide.util.Mutex.writeAccess(Mutex.java:252)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.iterate(NexusRepositoryIndexerImpl.java:872)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.find(NexusRepositoryIndexerImpl.java:1387)
	at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.find(NexusRepositoryIndexerImpl.java:1382)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.lambda$findResult$13(RepositoryQueries.java:505)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.getQueryResult(RepositoryQueries.java:217)
	at org.netbeans.modules.maven.indexer.api.RepositoryQueries.findResult(RepositoryQueries.java:502)
	at org.netbeans.modules.maven.nodes.AddDependencyPanel$QueryPanel$2.run(AddDependencyPanel.java:1144)
(...)
```

JaroslavTulach:
Maven continues to support JDK8 and even older. E.g. Maven itself must be [using compatible version of Lucene](https://github.com/apache/maven-indexer/blob/master/pom.xml#L95). I suggest NetBeans to use the [same version](https://github.com/apache/maven-indexer/blob/master/pom.xml#L95).

mbien:
> Maven continues to support JDK8 and even older. E.g. Maven itself must be [using compatible version of Lucene](https://github.com/apache/maven-indexer/blob/master/pom.xml#L95). I suggest NetBeans to use the [same version](https://github.com/apache/maven-indexer/blob/master/pom.xml#L95).

you mean maven-indexer? Thats exactly what this PR does. Latest 8.x is in use which is the same what maven-indexer 6.1x uses, Compatibility windows are described in the PR text and why this all had to be done in one step (5.x -> 8.x). The encountered problems are listed too. 

The only potential showstopper remaining (IMO) is the discovered bug in lucene's hashcode logic, although it got quickly fixed thanks to lucene devs after I filed the issue, it doesn't really help us if it won't be backported to 8.x which we are stuck to due to the JDK 8 requirement of this module.

mbien:
to meet the JDK 8 requirement of this module I backported the hashcode workaround of lucene 9.0.0 to EOL lucene 8.11.1.

This is done by eagerly loading a patched version of MultiTermQuery. Two guards were added to make sure this hack doesn't break anything and ensure that it is removed on next upgrade.

1) if the unpatched class would be already loaded, this static init would fail and we would notice
https://github.com/apache/netbeans/blob/63485d7744cf209f589ac9fc382df860a6e18ff0/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java#L126-L131

2) if lucene is upgraded without removing this hack, this assert will fail, so we can't forget about it:
https://github.com/apache/netbeans/blob/63485d7744cf209f589ac9fc382df860a6e18ff0/java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java#L332-L341

java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java
```diff
@@ -168,50 +166,49 @@ private static boolean isArchiveFile(File jar) throws IOException {
     }
 
     static void search(String className, Indexer indexer, Collection<IndexingContext> contexts, List<? super ClassUsage> results) throws IOException {
-        String searchString = crc32base64(className.replace('.', '/'));
-        Query refClassQuery = indexer.constructQuery(ClassDependencyIndexCreator.FLD_NB_DEPENDENCY_CLASS.getOntology(), new StringSearchExpression(searchString));
-        TopScoreDocCollector collector = TopScoreDocCollector.create(NexusRepositoryIndexerImpl.MAX_RESULT_COUNT, null);
+        String searchString = crc32base32(className.replace('.', '/'));
```
matthiasblaesing, line 169: If I understand this correctly the indexed values are changed and thus an old index can't be reused. Can we make sure, that the old index is considered broken/obsolete/invalid and force a reindex or am I missing something?

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
```diff
@@ -123,34 +122,30 @@
...
-        // that will assign value to the ACCESSOR field above
-        Class<?> c = RepositoryQueries.class;
-        try {
-            Class.forName(c.getName(), true, c.getClassLoader());
-        } catch (ClassNotFoundException x) {
-            throw new ExceptionInInitializerError(x);
-        }
-    }     
+        // XXX patch for LUCENE-10431
+        // should be removed as soon migraiton from EOL lucene 8 to lucene 9+ is possible
```
matthiasblaesing, line 127:
```suggestion
        // should be removed as soon migration from EOL lucene 8 to lucene 9+ is possible
```

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
```diff
@@ -958,47 +899,39 @@ private void iterate(List<RepositoryInfo> repos, final RepoAction action, final
...
+        );
         return filterGroupIds(prefix, result, repos, true);
     }
     
     private ResultImplementation<String> filterGroupIds(final String prefix, final ResultImpl<String> result, 
                                                             final List<RepositoryInfo> repos, final boolean skipUnIndexed) {
-        final Set<String> groups = new TreeSet<String>(result.getResults());
-        final List<RepositoryInfo> slowCheck = new ArrayList<RepositoryInfo>();
+        final Set<String> groups = new TreeSet<>(result.getResults());
+//        final List<RepositoryInfo> slowCheck = new ArrayList<>();
```
matthiasblaesing, line 917: Should slowCheck be removed? Comment below says, that is is used rarely, now it is commented out - maybe it should be removed?

matthiasblaesing (commented):
In general looks good. A quick local test showed, that searching for dependencies (locally by class, remote by artifact/group name). I left a few inline comments though.

apisupport/maven.apisupport/nbproject/project.xml
```diff
@@ -25,6 +25,16 @@
         <data xmlns="http://www.netbeans.org/ns/nb-module-project/3">
             <code-name-base>org.netbeans.modules.maven.apisupport</code-name-base>
             <module-dependencies>
+                <!-- keep in front of any module which uses lucene, maven indexer uses it's own verison  -->
```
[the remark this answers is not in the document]
mbien: it did solve the problem. You should be able to reproduce it by swapping it around again. It should fail rather quickly in jar version hell.

apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NBMNativeMWI.java
```diff
@@ -247,7 +247,7 @@ public void performOperation(POMModel model) {
                 //nbm-maven-plugin
                 boolean addPlugin = true;
                 String managedPVersion = null;
-                String pVersion = MavenNbModuleImpl.getLatestNbmPluginVersion();
+                String pVersion = MavenNbModuleImpl.LATEST_NBM_PLUGIN_VERSION;
```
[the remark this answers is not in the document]
mbien: it is intended, Its this commit: https://github.com/apache/netbeans/pull/3558/commits/5b9efc25e765714a31e094e05be661afc88ff4b0

explanation is here:
https://github.com/apache/netbeans/pull/3558#issuecomment-1045585199

short version:
code swallowed a NPE which basically made the tests test nothing, post upgrade the NPE did not occur and the tests tried to download the maven repo index which caused them to time out

by returning a constant (again, i basically reverted a older commit) and not ignoring the exception, everything works again.

Yes, semantics changed a little to the old behavior before 57e21b83150cc4d7e380c53c2b990c6c6e69bc52

java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java
```diff
@@ -0,0 +1,366 @@
...
+ * performant rewrite method given the query. If you
+ * need scoring (like {@link FuzzyQuery}, use
+ * {@link TopTermsScoringBooleanQueryRewrite} which uses
+ * a priority queue to only collect competitive terms
+ * and not hit this limitation.
+ *
+ * Note that org.apache.lucene.queryparser.classic.QueryParser produces
+ * MultiTermQueries using {@link #CONSTANT_SCORE_REWRITE}
+ * by default.
+ */
```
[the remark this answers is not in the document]
mbien: comment is there. Should i move it to the very top of the file? Its over the secretInit() method.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java
mbien: good question. I haven't tested that situation at all tbh since a NB upgrade does not copy cache files. Lucene 5 index would be incompatible with 8 anyway - Its likely the exiting code would reset the index on failure. 

I am going to copy the index from NB 13 to a test setup and see how it behaves.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
mbien: fixed

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
mbien: removed it.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java
mbien: i ran the experiment as described above and it reindexed everything the next time the index was used.

apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NBMNativeMWI.java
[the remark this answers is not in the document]
matthiasblaesing: @oyarzun could you please have a look at this - it is your commit, for which the behavior is changed, so this needs to be reevaluated.

java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java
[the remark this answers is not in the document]
matthiasblaesing: I saw the comment at the method, but I knew what I was looking for. Without that, having a comment at the top would be helpful.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java
matthiasblaesing: Thank you.

java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java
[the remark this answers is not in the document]
mbien: i wasn't sure where to put it to grab the developers attention, it was on top first actually, but i moved it down to the hashCode() method later where the assert is. I added code folds now with additional description and set them to collapsed by default. Let me know what you think.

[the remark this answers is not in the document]
matthiasblaesing: Nice idea - I feared that rat check would fail on this, but the ALv2-ASF license is correctly picked up. Thank you for the update.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/ClassDependencyIndexCreator.java
```diff
@@ -382,21 +382,20 @@ void read(File jar, JarClassEntryConsumer consumer) throws IOException {
...
         crc.reset();
         crc.update(s.getBytes(UTF8));
         long v = crc.getValue();
-        byte[] b64 = Base64.encodeBase64(new byte[] {(byte) (v >> 24 & 0xFF), (byte) (v >> 16 & 0xFF), (byte) (v >> 8 & 0xFF), (byte) (v & 0xFF)});
+        byte[] b64 = base32.encode(new byte[] {(byte) (v >> 24 & 0xFF), (byte) (v >> 16 & 0xFF), (byte) (v >> 8 & 0xFF), (byte) (v & 0xFF)});
         assert b64.length == 8;
-        assert b64[6] == '=';
         assert b64[7] == '=';
-        return new String(b64, 0, 6, LATIN1).replace('+', '_');
+        return new String(b64, 0, 7, LATIN1).toLowerCase();
```
mbien: forgot to rename the b64 variable to b32

mbien:
@matthiasblaesing what is your opinion on this. This is a fairly mission critical area so we really should make sure that everything works before integrating IMO... but we have only a week left till freeze - would be great to get more reviews.

matthiasblaesing:
@mbien sorry for the late reply: The outstanding question is the handling of the plugin version. Would it be possible to implement something like this:

1. Check if maven index is available
2. If maven index is available find the latest version and use that
3. else fallback to the hardcoded value

That way we have the best of both worlds.

apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/MavenNbModuleImpl.java
```diff
@@ -119,13 +119,13 @@ static List<RepositoryInfo> netbeansRepo() {
...
+    /**
+     * Returns the latest known version of the NetBeans maven plugin.
+     * This method will not wait for the index to be downloaded, it will return a default value instead.
+     */
     public static String getLatestNbmPluginVersion() {
         RepositoryQueries.Result<NBVersionInfo> versionsResult = RepositoryQueries.getVersionsResult(GROUPID_APACHE, NBM_PLUGIN, null);
 
-        if (versionsResult.isPartial()) {
-            versionsResult.waitForSkipped();
-        }
```
mbien: @matthiasblaesing lets see if this works. Its now essentially doing what you suggested - assuming all tests are green.

mbien: did some more manual testing and everything worked fine. I think we could try to get this in NB 14 (assuming travis is green again, queue is full ATM)

This PR would reduce on-disk storage size by about 800mb for a fresh index and queries appear to have slightly better performance too when tested side by side.

matthiasblaesing (approved):
I suggest to get this into master, once it settles down, the earlier, the long we have time to fix bugs ;-)

mbien:
ok awesome. Travis is working here too. I squashed a few less important commits, i would like to merge the rest without squashing so that it simplifies debugging with git bisections if problems occur.

PR #3561, opened 2022-02-06
[jackpot] more robust condition handling + code renovations

mbien:
first commit is a general language level upgrade (to 8) of the declarative hints module (diamonds, multicatch, overrides, lamdas, collections, ARM-blocks, minor refactorings, warning removal etc)

second commit makes rule conditions more robust.
 - catches and logs the exception with information helpful to see which rule/condition causes problems
 - this should make it also more robust since one broken rule won't affect processing of other rules


mbien:
rebased to resolve merge conflicts

lkishalmi (approved):
Looks good to me!

mbien:
@lkishalmi thank you for reviewing.

PR #3564, opened 2022-02-07
WIP: Download wallet

sdedic:
Testing why #3546 fails on CI


sdedic:
Note this PR is marked as do not merge, but did not contain more commits than the original one from @jhorvath -- seems that GH marked it as merged when the original commit hash was merged to master. Ignore, a new PR (this time with an unique commit on top) will be created.

PR #3568, opened 2022-02-08
Alternative ProjectProblems presenter for LSP server.

sdedic:
Project problems are presented in a rich UI dialog in NB, but that's not suitable for LSP protocol that only allows for limited dialogs or quickpicks. Instead of a rich-interaction view, this PR replaces (just for the LSP !!) the process by a series of dialogs.
- Errors are presented all at once, with some limit (~10). If there are more errors, they're displayed as soon as the user closes/confirms earlier messages.
- Fixable problems are presented one by one: each confirmation offers the possibility to fix rest of the issues without asking
- If an error occurs during a fix (the resolution is not "resolved"), the err message is printed, and the user is asked to cancel or continue with fixes
- If a new unfixable problem appears during the process, the error is displayed before the next confirmation/question

There's a timeout implemented for the user's response to questions or messages; the process will continue after the timeout. This is because the LSP client user may completely ignore (even not cancel) the messages, but I'd still want the identified errors to be displayed.

A `showAlert` API call will immediately continue the problem resolution process, as if the timeout had elapsed; for example on project reload or reopen. But ony problems not yet presented to the user are processed.

A `showCustomizer` API call will restart the whole process with all the currently known problems (incl. the already shown ones, if they still exist).

This alternative implementation is present in `java/java.lsp.server` module, but the implementation is only registered in `nbcode/integration` module which is part of `nbcode` distribution (published as Apache NetBeans Language Server extension for vscode).

### DIFF
--- a/java/java.lsp.server/nbproject/project.xml
+++ b/java/java.lsp.server/nbproject/project.xml
@@ -462,6 +462,17 @@
                         <specification-version>1.101</specification-version>
                     </run-dependency>
                 </dependency>
+
+                <dependency>
+                    <code-name-base>org.netbeans.modules.projectuiapi</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>1</release-version>
+                        <implementation-version/>
+                    </run-dependency>
+                </dependency>
+
                 <dependency>
                     <code-name-base>org.netbeans.modules.queries</code-name-base>
                     <build-prerequisite/>
@@ -738,6 +749,7 @@
                     <test-dependency>
                         <code-name-base>org.netbeans.modules.projectui</code-name-base>
                         <recursive/>
+                        <compile-dependency/>
                     </test-dependency>
                     <test-dependency>
                         <code-name-base>org.netbeans.modules.settings</code-name-base>
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
@@ -1,0 +1,318 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.netbeans.modules.java.lsp.server.project;
+
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import javax.swing.AbstractListModel;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.api.project.Project;
+import org.netbeans.api.project.ProjectManager;
+import org.netbeans.api.project.ProjectUtils;
+import org.openide.util.ChangeSupport;
+import org.netbeans.spi.project.ui.ProjectProblemsProvider;
+import org.netbeans.spi.project.ui.ProjectProblemsProvider.ProjectProblem;
+import org.openide.util.Mutex;
+import org.openide.util.NbBundle;
+import org.openide.util.WeakListeners;
+
+/**
+ * This is a copy of broken references model from Project UI API, slightly improved.
+ * 
+ */
+public final class BrokenReferencesModel extends AbstractListModel implements PropertyChangeListener, ChangeListener {
+
+    private static final Logger LOG = Logger.getLogger(BrokenReferencesModel.class.getName());
+
+    private final Context ctx;
+    private final boolean global;
+    private final Object lock = new Object();
+    //@GuardedBy("lock")
+    private final Map<ProjectProblemsProvider,PropertyChangeListener> providers;
+    //@GuardedBy("lock")
+    private final List<ProblemReference> problems;
+
+
+    BrokenReferencesModel(@NonNull final Context ctx, boolean global) {
+        assert ctx != null;
+        this.ctx = ctx;
+        this.global = global;
+        problems = new ArrayList<ProblemReference>();
+        providers = new WeakHashMap<ProjectProblemsProvider, PropertyChangeListener>();
+        refresh();
+        ctx.addChangeListener(this);
+    }
+
+    BrokenReferencesModel(@NonNull final Project project) {
+        this(new Context(), false);
+        this.ctx.offer(project);
+    }
+
+    @Override
+    public Object getElementAt(int index) {
+        return getOneReference(index);
+    }
+
+    @Override
+    public int getSize() {
+        synchronized (lock) {
+            return problems.size();
+        }
+    }
+    
+    public List<Project> projects() {
+        synchronized (this) {
+            return problems.stream().sequential().map(r -> r.project).collect(Collectors.toList());
+        }
+    }
+    
+    public List<ProblemReference> projectProblems(Project project, boolean includeSeen) {
+        synchronized (this) {
+            return problems.stream().sequential().filter(r -> r.project == project).collect(Collectors.toList());
+        }
+    }
+    
+    void refresh() {
+        refresh(true);
+    }
+    
+    
+    /**
+     * Refreshes the model + allows to suppress change events. Call refresh(false) from problem processing methods to avoid
+     * recursive invocation, as the change events are fired even though no problems are added/removed.
+     * @param fire 
+     */
+    void refresh(boolean fire) {
+        AtomicBoolean changed = new AtomicBoolean(false);
+        final int size = ProjectManager.mutex().readAccess(new Mutex.Action<Integer>() {
+            @Override
+            public Integer run() {
+                synchronized (lock) {
+                    final Map<ProjectProblemsProvider,Project> newProviders = new LinkedHashMap<ProjectProblemsProvider,Project>();
+                    for (Project bprj : ctx.getBrokenProjects()) {
+                        final ProjectProblemsProvider provider = bprj.getLookup().lookup(ProjectProblemsProvider.class);
+                        if (provider != null) {
+                            newProviders.put(provider, bprj);
+                        }
+                    }
+                    for (Iterator<Map.Entry<ProjectProblemsProvider,PropertyChangeListener>> it = providers.entrySet().iterator(); it.hasNext();) {
+                        final Map.Entry<ProjectProblemsProvider,PropertyChangeListener> e = it.next();
+                        if (!newProviders.containsKey(e.getKey())) {
+                            e.getKey().removePropertyChangeListener(e.getValue());
+                            it.remove();
+                        }
+                    }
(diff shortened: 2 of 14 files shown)


java/java.lsp.server/nbproject/project.xml
```diff
@@ -454,6 +454,17 @@
...
                     </run-dependency>
                 </dependency>
+
+                <dependency>
+                    <code-name-base>org.netbeans.modules.projectuiapi</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>1</release-version>
+                        <implementation-version/>
```
JaroslavTulach, line 464: Implementation version? It'd be better to use real API, not implementation dependencies.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
```diff
@@ -0,0 +1,318 @@
...
+import org.netbeans.spi.project.ui.ProjectProblemsProvider.ProjectProblem;
+import org.openide.util.Mutex;
+import org.openide.util.NbBundle;
+import org.openide.util.WeakListeners;
+
+/**
+ * This is a copy of broken references model from Project UI API, slightly improved.
+ * 
+ */
+public final class BrokenReferencesModel extends AbstractListModel implements PropertyChangeListener, ChangeListener {
```
JaroslavTulach, line 55: The alternative might be to use `Node` instead of `javax.swing.AbstractListModel` and expose the hierarchy from Project UI API module rather than copying the code.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
```diff
@@ -0,0 +1,622 @@
...
+            @Override
+            public boolean cancel(boolean mayInterruptIfRunning) {
+                boolean r = super.cancel(mayInterruptIfRunning);
+                if (r) {
+                    ProjectAlertPresenter.this.cancel();
+                }
+                return r;
+            }
+        };
+        model.addListDataListener(new ListDataListener() {
```
JaroslavTulach, line 211: Using Swing in headless module is probably not appropriate.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
```diff
@@ -0,0 +1,622 @@
...
+     */
+    // @GuardedBy(this)
+    boolean allProcessed = true;
+    
+    /**
+     * Wait threads that only record user's acknowledge for a fatal error. No real interaction is required, so there can be more
+     * of these messages to confirm
+     */
+    // @GuardedBy(RP.this)
+    RequestProcessor fatalMsgs = new RequestProcessor(ProjectAlertPresenter.class.getName(), MAX_PRESENTED_ERRORS);
```
JaroslavTulach, line 182: `RequestProcessor` with ~ ten parallel threads!? I would hope the `CompletableStage` approach leads towards a single threaded solution, not a UI that needs pool of many threads.

java/java.lsp.server/nbproject/project.xml
sdedic: This will be tough. Event `projectui` impl-depends on `projectuiapi` for this type. Should we make `BrokenReferencesImplementation` a regular API ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
sdedic: I need a thread pool (now) so multiple messages can be displayed in parallel + waited on the result: `DialogDisplayer.notify()` blocks, .notifyLater does not return a value. An option would be to introduce `DialogDisplayer.notifyCompletion` that returns a `CompletableFuture`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
sdedic: Well... it's weird, but the data model does not use any graphics-related stuff.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
sdedic: Do you mean to use Node / Children as a model, ProjectProblem as a service in its Lookup ?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
JaroslavTulach: Making callback on button press is the future! I have implemented that for the [HTML UI dialogs](https://bits.netbeans.org/dev/javadoc/org-netbeans-api-htmlui/org/netbeans/api/htmlui/HTMLDialog.OnSubmit.html) during NetBeans 13 development. Providing similar solution for regular dialogs is good idea.

java/java.lsp.server/nbproject/project.xml
JaroslavTulach: Making `BrokenReferencesImplementation` (a two method interface) an SPI in `org.netbeans.spi.project.ui` is certainly better than to spread implementation dependencies.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
JaroslavTulach: Btw. if you had `Node` representing the problems, you could show it in the logical project view under the project node... that might be nicer UI than a sequence of dialogs...

sdedic: Hm, as an alternative - possibly. I did not thought about chaning NB IDE UI by this PR ... and I still need to think this through - the error messages might be quite long -> display as node's description (tooltip):  does not seem as good usability pattern.
The API `ProjectProblems.alert()` is used & useful (it is called after project open).

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
sdedic: CompletableFuture support is proposed in #3613

java/java.lsp.server/nbproject/project.xml
sdedic: SPI created in 27ad3df6

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/ProjectAlertPresenter.java
sdedic: Rewritten using `CompletableFutures` in 325c279

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/BrokenReferencesModel.java
sdedic: I'll keep the data model as it is now -- it will not escape the implementation package, and although types are in Swing lib, they do not make threading manipulations.

ppisl (approved):
Currently I don't see any problem.

sdedic:
Consolidated fixes ... actually everything into a single commit.

sdedic:
I'd like to hold this PR for a moment - during testing, I've encountered [NETBEANS-5846](https://issues.apache.org/jira/browse/NETBEANS-5846) this time even in vscode; together with this PR, NETBEANS-5846 looks rather ugly (several dialogs displayed on project open). I'd prefer to fix NETBEANS-5846 first the way @lkishalmi suggested there.

sdedic:
Prerequisite PR #3668 merged, so this is ready.

sdedic:
Just random failures from ServerTest / ProjectViewTest. Merging.

ide/projectuiapi.base/src/org/netbeans/spi/project/ui/ProjectProblemsProvider.java
```diff
@@ -187,7 +187,7 @@ public static Result create(
         private final String description;
         private final ProjectProblemResolver resolver;
 
-        private ProjectProblem(
+        public ProjectProblem(
```
JaroslavTulach, line 190: The constructor is now part of the API. Needed? If so, add Javadoc, `@since` tag, etc.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/LspBrokenReferencesImpl.java
```diff
@@ -0,0 +1,161 @@
...
+ * <p/>
+ * A request to <b>display a customizer</b> will invalidate the current presenter, and display all the issues
+ * anew. This allows for an explicit CodeAction or Command that will reiterate questions the user may have closed.
+ * <p/>
+ * Note: if a ShowMessage request is delivered to vscode client and the same ShowMessage request is pending, the client 
+ * will cancel the former ShowMessage as if the user pressed ESC or dismissed the message. The new ShowMessage will be 
+ * displayed
+ * @author sdedic
+ */
+// BAD BAD BAD - DO NOT REGISTER THIS IN LSP SERVER, put it in the integration module !!!
```
JaroslavTulach, line 68: Would making this class `abstract` with `protected` constructor help to remove the `BAD BAD` message?

JaroslavTulach: But how can the integration module access this class, when it is not in public package?

JaroslavTulach (approved):

ide/projectuiapi.base/src/org/netbeans/spi/project/ui/ProjectProblemsProvider.java
sdedic: Good catch ! It's a leftover prototyping, actually the factory methods are sufficient. Will remove.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/LspBrokenReferencesImpl.java
sdedic: It would not :) it's there just for case that someone won't get the great idea to add `@ServiceProvider` since the lsp.server module is a part of NetBeans IDE distro and this override would screw the IDE up.

The integration module uses an explicit META-INF/Services registration that references this class by name. For that the default constructor has to be public. If too ugly, the alternative would be to expose a factory method for the implementation as an API, but I really do not want to expose the implementation class in the APIs.

JaroslavTulach: There is a common pattern used for this kind of registrations and there are multiple [AbstractLspXYZ](https://github.com/apache/netbeans/tree/8789e44fef1d416288f6fb62ec9a79442dee9f99/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui) classes. What is so special on this class that requires a special treatment?

Unless you have some intrinsic answer to the above question, then I **suggest** to create `AbstractLspProjectProblemsImplementation`, make it `abstract`, have `protected` constructor like other implementations do. However, it is just a **suggestion**, do as you like.

ide/projectuiapi.base/src/org/netbeans/spi/project/ui/ProjectProblemsProvider.java
sdedic: Fixed in 39596e2d6f

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/project/LspBrokenReferencesImpl.java
sdedic: Dear colleague(s). Your peristence in commenting my laziness finally worked... So this should be fixed in 39596e2d6f

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspBrokenReferences.java
```diff
@@ -0,0 +1,48 @@
...
+import org.netbeans.modules.java.lsp.server.project.BrokenReferencesImpl;
+import org.netbeans.spi.project.ui.ProjectProblemsImplementation;
+
+/**
+ * Implementation of {@link ProjectProblemsImplementation} implemented using
+ * dialogs over LSP.
+ * @since 1.19
+ * @author sdedic
+ */
+public abstract class AbstractLspBrokenReferences implements ProjectProblemsImplementation {
```
JaroslavTulach, line 32: Great!

JaroslavTulach (approved):

PR #3574, opened 2022-02-08
update github tags.

mbien:
any other tags I should add?

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -19,7 +19,10 @@ github:
   homepage: https://netbeans.apache.org/
   labels:
     - netbeans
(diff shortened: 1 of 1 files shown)


neilcsmith-net (approved):
Looks OK to me. Assume you've compared with other similar projects on GitHub? Not sure of pros / cons / value of hacktoberfest being there.

mbien:
@neilcsmith-net +1, lets remove hacktober. Its hackfebruary anyway.

mbien:
added "java" since NetBeans and the platform is written in it/runs on it (not in the sense that it is a supported language)

PR #3575, opened 2022-02-09
fixed wrong nb-javac module name so that it can be installed.

mbien:
when nb-javac is uninstalled by the user and NB is started on a JDK < 17 this dialog will show up:
install-nb-javac

However, pressing install did nothing since the module name was wrong (!). This fixes it.

PR is based on delivery branch. I think we should add this even if there won't be another rc.

how to test:
1) start NB, load java cluster
2) uninstall nb-javac
3) start again on JDK 8, dialog should show up
4) install nb-javac via dalog, restart -> everything should be working again (no dialog)

### DIFF
--- a/java/java.j2semodule/src/org/netbeans/modules/java/j2semodule/ui/customizer/Bundle.properties
+++ b/java/java.j2semodule/src/org/netbeans/modules/java/j2semodule/ui/customizer/Bundle.properties
@@ -601,7 +601,7 @@ ACSD_librariesLocation=Folder containing location of the library definition file
 ACSD_jButtonEdit=Edit classpath items
 ACSD_BuildJarAfterCompile=Build JAR after compiling option
 LBL_CustomizeRun_Enable_Quick_Run=Enable Quick Run
-CustomizerCompile.CompileOnSave=Compile on &Save (requires nb-javac plugin)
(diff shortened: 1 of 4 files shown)


JaroslavTulach (approved):

neilcsmith-net (commented):
> PR is based on delivery branch. I think we should add this even if there won't be another rc.

Except we don't add things unless there's another RC. Release vote is always on git hash of last RC.

I'll leave @geertjanw and @ebarboni to make call if they want to merge and roll another RC - I can't until next week anyway, and would like to kick off the vote then if possible.

It's a good catch, but potentially a minor annoyance as most users shouldn't ever see this?

ebarboni:
hi I can go for another RC but we can still start vote next week

mbien:
@neilcsmith-net understood. I do also agree that this is not severe enough to be the cause for another rc - its just a safety net for users which misconfigured NB.

neilcsmith-net:
@mbien we might be looking at another issue that requires an rc4 anyway, in which case this would be good to get in as well. Can we switch it back to delivery / NB13 again for now? :smile:

neilcsmith-net:
Thanks @mbien - merging!

PR #3594, opened 2022-02-12
[NETBEANS-5596] No syntax coloring with the Java Editor Kit

nbauma109:
This will enable syntax coloring in the JAVA editors of standalone swing applications such as https://github.com/GeertjanWielenga/netbeans-visual-diff-standalone


matthiasblaesing:
Please squash the two commits and adjust the author information in the commit to hold your real name and a valid email address.

java/java.editor/test/unit/src/org/netbeans/modules/editor/java/SyntaxColoringTest.java
```diff
@@ -0,0 +1,69 @@
...
+			if (tokenHierarchy != null) {
+				TokenSequence<?> tokenSequence = tokenHierarchy.tokenSequence();
+				if (tokenSequence != null) {
+					tokenCount = tokenSequence.tokenCount();
+				}
+			}
+		});
+
+		try {
+			Thread.sleep(5000);
```
mbien, line 57: lets try to implement this without `sleep`. 5s are long for a test but making it shorter will cause trouble under load.

e.g `invokeAndWait` or a semaphore would be an option.

matthiasblaesing: If the intention is, that the complete test is run on the EDT, `NbTestCase` can be told to do that:

http://bits.netbeans.org/dev/javadoc/org-netbeans-modules-nbjunit/org/netbeans/junit/NbTestCase.html#runInEQ--

return `true` from it and you are done.

matthiasblaesing:
@nbauma109 this change causes a problem in other unittests. I had a quick look at it and indeed the problem is locally reproduceable. It seems, that the unittest relied on original document returned. My gut feeling is, that the indention seems to be sane after the change. This was the change needed:

```diff
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/java/java.source.base/test/unit/src/org/netbeans/api/java/source/gen/CompilationUnitTest.java
+++ b/java/java.source.base/test/unit/src/org/netbeans/api/java/source/gen/CompilationUnitTest.java
@@ -1280,7 +1280,7 @@
             "    public I(java.util.Collection c) {\n" +
             "        super(c);\n" +
             "    }\n" +
-            "  \n" +//XXX
+            "    \n" +//XXX
             "}\n";
 
         JavaSource javaSource = JavaSource.create(cpInfo, FileUtil.toFileObject(testFile));
@@ -1309,7 +1309,7 @@
         result.commit();
         String res = TestUtilities.copyFileToString(new File(getDataDir().getAbsolutePath() + "/zoo/I.java"));
         //System.err.println(res);
-        assertEquals(res, golden1);
+        assertEquals(golden1, res);
     }
     
     String getGoldenPckg() {
```

You can run the failing unittest locally:

```bash
ant -f java/java.source.base/ test
```

The failure can be found in java/java.source.base/build/test/unit/results/TEST-org.netbeans.api.java.source.gen.CompilationUnitTest.xml (or by running the unittest in the IDE).

nbauma109:
The change seems to cause more serious side effects on other test cases than this indent. Can you also confirm that ?

jlahoda:
@matthiasblaesing , sorry for delays. I'll take a look to see if I can find out what is going on.

jlahoda:
So, I believe the reason for the failing `CompilationUnitTest.java` is because, when the Language is installed, reformatter runs when the file is created. I see the `PartialReparseTest` is failing again, and the reason is probably similar (reformatter changes the code, so that original positions are no longer reasonable and the test modifies wrong parts of the code). Any other test that is failing?

I believe NB is normally finding the Languages using an implementation of `org.netbeans.spi.lexer.LanguageProvider` (the NB implementation will look the Language in a MimeLookup, but I believe it can get the Language anywhere). I wonder if that would be a viable alternative?

nbauma109:
`LanguageManager.getInstance().findLanguage("text/x-java")` didn't find the Java language (at least in my scenario) so I added the Language.class property directly. What are the side effects aside formatting, especially in the test cases with "(ERROR)" in them ?

jlahoda:
Without working MimeLookup, filesystems, etc., the LanguageManager needs a different implementation of LanguageProvider. What I had in mind is something along the lines of (untested):
```
@ServiceProvider(service=LanguageProvider.class)
public class LanguageProviderImpl extends LanguageProvider {

    @Override
    public Language<?> findLanguage(String mimeType) {
        switch (mimeType) {
            case "text/x-java": return JavaTokenId.language();
        }
        return null;
    }

    @Override
    public LanguageEmbedding<?> findLanguageEmbedding(Token<?> token, LanguagePath languagePath, InputAttributes inputAttributes) {
        return null;
    }

}
```

In case of PartialReparseTest, it is probably "just" a matter of formatting changing the code so that existing positions (in the test) no longer work. You may be able to use Positions, to keep track of the changes.

nbauma109:
It's exactly what I'm looking for, and it's tested at my end, outside of netbeans. I think that SPI implementation belongs to the "client" code, doesn't it ? Or should it be included in netbeans as a default implementation ? Since there is no need to workaround the JavaKit like I did, I think I can also just give up the PR ? It was for use outside of netbeans for standalone swing applications with netbean modules as dependencies. Let me know your thoughts.

Issue #3593, opened 2022-02-12
java.lang.NullPointerException occurs when we open a css file.


junichi11:
### Apache NetBeans version

Latest release candidate

### What happened

java.lang.NullPointerException occurs when we open a css file.

```
java.lang.NullPointerException
	at org.netbeans.modules.csl.navigation.ElementNode.getNodeForOffset(ElementNode.java:245)
	at org.netbeans.modules.csl.navigation.ElementNode.getMimeRootNodeForOffset(ElementNode.java:230)
	at org.netbeans.modules.csl.navigation.ElementNode.getMimeRootNodeForOffset(ElementNode.java:202)
	at org.netbeans.modules.csl.navigation.ClassMemberPanelUI.doSelectNodes0(ClassMemberPanelUI.java:263)
	at org.netbeans.modules.csl.navigation.ClassMemberPanelUI.access$500(ClassMemberPanelUI.java:74)
	at org.netbeans.modules.csl.navigation.ClassMemberPanelUI$3.run(ClassMemberPanelUI.java:254)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:136)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:190)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:235)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:233)
[...]

PR #3597, opened 2022-02-12
[#3593] Fix csl elementhandle

matthiasblaesing:
The contract for StructureItem#getElementHandle is not always honored and implementors return `null`.
This causes NullPointer Exceptions in `ElementNode` which relies on it being present.

This changesets adds a guard to prevent invalid implementations to raise exceptions and fixes the CSS
code to provide the required `ElementHandle` implementations.

### DIFF
--- a/ide/csl.api/src/org/netbeans/modules/csl/navigation/ElementNode.java
+++ b/ide/csl.api/src/org/netbeans/modules/csl/navigation/ElementNode.java
@@ -29,6 +29,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import javax.swing.Action;
 import javax.swing.Icon;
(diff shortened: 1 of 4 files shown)


ide/csl.api/src/org/netbeans/modules/csl/navigation/ElementNode.java
```diff
@@ -242,7 +243,15 @@ public ElementNode getNodeForOffset(int offset) {
             Node[] children = ch.getNodes();
             for (int i = 0; i < children.length; i++) {
                 ElementNode c = (ElementNode) children[i];
-                if (this.getFileObject() != c.getDescription().getElementHandle().getFileObject()) {
+                // The promise of the API is broken at several places in the
+                // codebase and thus this needs to be guarded. The assert is in
+                // place to find the violating places.
+                assert c.getDescription().getElementHandle() != null;
```
junichi11, line 249: Nice!

junichi11 (approved):
Looks good to me. Thank you for your great work!
I've verified that NPE doesn't occur with css, less, and scss files.

neilcsmith-net (approved):
Did quick check on the reported issue. Looks great!  Thanks @matthiasblaesing Merging to delivery and will be in rc4.

PR #3598, opened 2022-02-13
Handle classfiles with too new versions.

jlahoda:
javac normally refuses to load classfiles that are too new (unlike old nb-javac). Classfile versions change fairly frequently, but javac can often load the newer classfiles anyway. So let's try to load them, as did nb-javac. (It will be a bit slower to read the too-new classfile, but I think that is mostly fine.)

Based on a report from here:
https://lists.apache.org/thread/gtv4nnhvjtzqrk8zmsojj70blw0n18yv

### DIFF
--- a/java/lib.nbjavac/src/org/netbeans/lib/nbjavac/services/NBClassReader.java
+++ b/java/lib.nbjavac/src/org/netbeans/lib/nbjavac/services/NBClassReader.java
@@ -18,23 +18,40 @@
  */
 package org.netbeans.lib.nbjavac.services;
 
+import com.sun.tools.javac.code.ClassFinder.BadClassFile;
 import java.util.Set;
 import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.jvm.ClassFile;
 import com.sun.tools.javac.jvm.ClassFile.Version;
 import com.sun.tools.javac.jvm.ClassReader;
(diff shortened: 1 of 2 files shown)


mbien:
awesome :)
nitpick: typo in commit msg and PR title. two -> too

curious, what will `org.netbeans.api.java.source.CompilationInfo#getSourceVersion()` return after this? The modified version or the version set for the project?

mbien (approved):
ran some manual tests and I could open and edit ea JDK 19 loom/panama projects while using nb-javac and running NB on JDK 17. Auto completion, rename refactoring, jackpot scripts etc all worked with no exceptions in the log.

mbien:
There will be another rc for NB 13 on Tuesday. From a certain perspective this could be seen as a bug fix. Might be worth to rebase to delivery.

jlahoda:
> There will be another rc for NB 13 on Tuesday. From a certain perspective this could be seen as a bug fix. Might be worth to rebase to delivery.

I can rebase, I think, but I'll leave it up to @neilcsmith-net is he would be interested to include this in NB 13. To me, seems it could wait for the next version, though.

jlahoda:
(Thanks for testing and spotting the typo, BTW!)

neilcsmith-net:
Well, I've no more say than anyone else what goes in NB13!  However, I do have reservations about dropping this in so late.  I'm also curious what the implications are with things like Compile on Save?  Hard fail with workaround in NB13, and aiming to look at a better experience for this in NB14 seems a good move to me.

matthiasblaesing:
This was proposed to NB13, anyone objecting merging it?

lkishalmi:
Got approved, tests seems to be happy. I'm merging this one.

PR #3599, opened 2022-02-13
Update JS support to ECMAScript 13 / 2022

matthiasblaesing:
See individual commits for details


matthiasblaesing:
The only failing test is `java.lsp.server`, which does not touch the modified area and is known to be flaky.

matthiasblaesing:
I intent to merge this by mid next week. If anyone wants to object, please do so. The changesets to the parse itself can be found here:

https://github.com/matthiasblaesing/graaljs-nb

PR #3618, opened 2022-02-16
Update Apache Felix to 7.0.3 release

sjaenick:
Minor change: Update Apache Felix from 7.0.0 to 7.0.3


mbien:
changelog: https://github.com/apache/felix-dev/blob/master/framework/doc/changelog.txt

pepness:
Change look sane to me.

matthiasblaesing (commented):
The NOTICE file needs to be updated (at least the year has changed) and the big question: Was this tested? If I remember correctly, NetBeans IDE OSGI support runs on nebinox and not felix, so running the IDE would not be a test.

sjaenick:
Where's the upstream notice file coming from? Wasn't able to find it.

Also, untested, since this is only a minor bugfix update.

matthiasblaesing:
> Where's the upstream notice file coming from? Wasn't able to find it.

For Apache Projects I would always look first in the JAR itself. You can't be required to include a NOTICE you don't have. There are two main places to look: 

- the root of the JAR
- the META-INF folder

If that fails, I would look into the repository of the project (but if a project does not follow its own rules, this might already be to much).

> Also, untested, since this is only a minor bugfix update.

I changes behavior in classloaders and using Unsafe. Two notoriously incompatible places. Changes without checks are bad. Always. At least a manual check has to be done.

mbien:
i ran a fulltext search to find out the modules with direct dependencies to this osgi wrapper module and quite a few do.
```
org.netbeans.modules.bugzilla
org.netbeans.modules.java.mx.project
org.netbeans.modules.maven.osgi
org.netbeans.core.netigso
org.netbeans.modules.netbinox
```

I don't really know how this all connects since netbinox itself wraps a very old version of osgi (3.9). I just know that the last time i tried to run tests on JDK 17, that version got in the way. So I assume that 3.9 isn't actually used at runtime (otherwise it would fail there too?). Maybe @JaroslavTulach can help out here, I am curious how this works.

JaroslavTulach:
How does it all connect?
* first of all there was [Netigso](http://wiki.apidesign.org/wiki/Netigso)
* then we realized we need [Netbinox](http://wiki.apidesign.org/wiki/Netbinox) to
* be able to run [Mylyn](http://wiki.apidesign.org/wiki/Mylyn)

NetBeans usually prefer so called [modular libraries](http://wiki.apidesign.org/wiki/Modular_library) - as such NetBeans rather include direct OSGi JARs than using their copy included in Felix or Equinox.

mbien:
@JaroslavTulach  I remember reading your wiki back when i tried to get things to work on JDK 17. How did it evolve from there? Is netigso running all nb modules? Or did it stay a optional bridge for client apps to load their osgi libs with. I don't think mylyn is used anywhere in the IDE or is it? Maybe we are updating here something which isn't even used anywhere :)

JaroslavTulach:
Both Felix and Netbinox (enhanced Equinox) are included in the platform cluster. Many applications built on the NetBeans Platform (including for example VSCode, right @sdedic?) are fine with Felix. However NetBeans IDE needs Netbinox because Eclipse originated OSGi bundles like https://github.com/apache/netbeans/tree/master/ide/o.eclipse.mylyn.commons.core & co. use `Require-Bundle` directive in manifest that can only be satisfied by including Equinox.

sjaenick:
> > Where's the upstream notice file coming from? Wasn't able to find it.
> 
> For Apache Projects I would always look first in the JAR itself. You can't be required to include a NOTICE you don't have. There are two main places to look:
> 
>     * the root of the JAR
> 
>     * the META-INF folder

Thanks for the pointer, indeed it exists within the META-INF folder of the jar; I updated the PR.

matthiasblaesing:
Thank you for the update. I looked into the code base and core.kit, core.osgi, autoupdate.services and core.netigso have direct dependencies on felix, at least for tests. So we have at least partitial coverage. I restartet the failing tests and if noone objects, I would be willing to merge.

lkishalmi:
@matthiasblaesing merge?

matthiasblaesing:
@lkishalmi thanks for the reminder ;-)

PR #3620, opened 2022-02-17
Updated Gradle CLI option support.

lkishalmi:
Updated GradleCommandLine to support options according to Gradle 7.4

Even if it not nice regarding binary compatibility, I has to arrange the enums in alphabetic order, in order not to get mixed up when working on them.

Also Gradle options which expecting parameters are shown in the code completion. Those options which expect well defined values are stored in the parameter enum. Later on I plan to expand the code completion to use them as well.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
@@ -37,6 +37,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.ArrayList;
 import java.util.EnumMap;
+import java.util.HashSet;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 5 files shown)


extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -186,26 +209,57 @@ private Property(Argument.Kind kind, String prefix, String flag) {
...
-    public enum Parameter {
+    public enum Parameter implements GradleOptionItem {
 
-        SETTINGS_FILE(UNSUPPORTED, "-c", "--settings-file"),
-        CONSOLE(UNSUPPORTED, "--console"),
+        BUILD_FILE(UNSUPPORTED, "-b", "--build-file"),
+        CONFIGURATION_CACHE_PROBLEMS(PARAM, argValues("fail", "warn"), "--configuration-cache-problems"),
+        CONSOLE(UNSUPPORTED, argValues("plain", "auto", "rich", "verbose"), "--console"),
+        DEPENDENCY_VERIFICATION(PARAM, argValues("strict", "lenient", "off"), "-F", "--dependency-verification"),
+        EXCLUDE_TASK(PARAM, "-x", "--exclude-task"),
```
sdedic, line 218: `EXCLUDE_TASK` `PROJECT_DIR` and `INIT_SCRIPT` are missing description. Maybe write a test :) that will just fetch descrs from all options.
The `unsupported` parameters are never presented as items -- is their description never read ?

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -82,42 +83,55 @@
      */
     public static final String CHECK_TASK = "check"; //NOI18N
 
+    public interface GradleOptionItem {
```
sdedic, line 86: Add `@since` + APIchange.

lkishalmi: Thanks!

sdedic: Apologies :( Also please bump spec version of the module. Not only GradleOptionItem, but more importantly the Flag and Parameter changed, so a module that uses the new symbol need to depend on new spec version to avoid potential issues if distributed through Plugin Portal.

extide/gradle/test/unit/src/org/netbeans/modules/gradle/api/execute/GradleCommandLineTest.java
```diff
@@ -304,4 +305,22 @@ public void testJVMArgs2() throws IOException {
...
+        all.remove(GradleCommandLine.Parameter.IMPORT_BUILD);
+        List<GradleCommandLine.GradleOptionItem> missing = new ArrayList<>();
+        for (GradleCommandLine.GradleOptionItem item : all) {
+            try {
+                assertNotNull(item.getDescription());
+            } catch (MissingResourceException ex){
+                missing.add(item);
+            }
+        }
+        assertTrue(missing.toString(), missing.isEmpty());
```
sdedic, line 324: Perfect ! Ensures that the desc is filled if an option / parameter is added in the future.

sdedic (approved):

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
lkishalmi: Well, the spec version increment has happened when Neil made the master ready for NB14 development. So 2.23 is pretty new, unreleased.

PR #3621, opened 2022-02-17
Fixed endless CompletionStage chaining.

sdedic:
The PR fixes `TreeNodeProvider.getNodeId()` call, which was supposed to find (or create) node LSP ID. The original code chained CompletableStages endlessly, eventually runnin OOM. The code was also not updated sufficiently as the rest of code evolved. Since `holdChildren` is now supposed to contain entries for all nodes known to the client, the fix will create these records for all nodes up to the root.


JaroslavTulach (approved):
Seeing test: assuming it fails before and succeeds now. You have my go!

PR #3624, opened 2022-02-17
Use the blessed modifier order

mklaehn:
This replaces patterns such as "final public" with "public final" to bring the modifiers in the standard order.

See: https://rules.sonarsource.com/java/RSPEC-1124

---


mbien:
@mklaehn wait please. If you want do do this for all/most modules, do multiple modules per PR. Maybe a cluster per PR if possible. Travis has already a full queue right now.

mklaehn:
@mbien I am prepared to do all modules. A PR containing all files in all modules at the same time would result in changes to 1800+ files. I'd hate to review a PR of that size.

So I am going to do PRs for each affected top level folder to keep the number of reviewable files down. cpplite and a select few only have a few affected files. Others have multiple hundred affected files.

Following your comment I'm going to try and combine a few so that the number of required PRs is reduced somewhat.

mbien:
@mklaehn split it into commits, this allows reviewing just as if its split into PRs. Keep in mind that every PR is basically 2h of testing for the CI server, twice. One PR per cluster should be doable for cleanups like this. If you would have continued with one PR per module this would be a month of CI time easily, just for this one cosmetic cleanup.

vieiro:
Quoting:
_Not following this convention has no technical impact, but will reduce the code’s readability because most developers are used to the standard order._
I apologize having to say that I don't see any enhancement in code's readability with these commits.

mklaehn:
> @mklaehn split it into commits, this allows reviewing just as if it would be split into PRs. Keep in mind that every PR is basically 2h of testing for the CI cluster, twice. NB has ~20 top level folders. Using one PR per top level folder would be too disruptive just for cosmetics IMO.

As you wish

PR #3628, opened 2022-02-17
Add Support for Gradle Configuration Cache

lkishalmi:
This one is depends on #3620 (that's the first commit in this PR) will be rebased when #3620 is mainlined.
Till then the 2nd commit allows users to use the Gradle Configuration Cache feature which is in incubation phase.

That required the ```nb-toolig.gradle``` init script modified not to read the System properties for the NetBeans tooling jar file, so that file is used as a template and NB create a one in the cache dir using the absolute location of the tooling jar.

As this feature is somewhat new, the default setting is disabled. 
image


sdedic:
The cached processed init script will not work as expected if:
- another installation of NB will use the same userdir, especially when a different NB version does so.
- NB installation is moved to another directory
- Gradle module upgrades w/ changed gradle tooling

I would suggest to still read either system property or pass the location as project property. Other scenarios seem more complicated - i.e. incorporate location + tooling .jar + init script digest hash into the cached filename.

lkishalmi:
Well, the init script is in the cache dir not in the user dir, so it should not considered to be shareable by different instances. Though changed the code to generate the script on the first call/classloader instance so it should work even with runtime module upgrades as well.

If it's possible I'd try to avoid passing extra properties and things to a build.

sdedic:
> Well, the init script is in the cache dir not in the user dir, so it should not considered to be shareable by different instances. 
Note that if you pass `--user-directory` (but without an explicit `--cachedir` to NetBeans (and I do, as I maintain several sets of configs for several projects), then the cache dir is AFAIK placed in `$userdir/var/cache`; at least on Linux.

sdedic (approved):

PR #3633, opened 2022-02-18
Added Gradle '--warning-mode all' output action

lkishalmi:
This one adds a convenient rebuild action when Gradle displays: ```You can use '--warning-mode all' to show the individual...```

the ```--warning-mode all``` will be displayed as an action, clicking on that the build would be restarted with that option.


sdedic (approved):
Neat :)

PR #3635, opened 2022-02-18
[NETBEANS-6458] A bogus Cyclic Reference issue is resolved by marking fields as volatile.

entlicher:
In the GraalVM debugging tests (`java/debugger.jpda.truffle`) there was often a Cyclic reference issue from `FolderInstance`.
It turned out, that there was not really any cyclic reference, but as the field `waitFor` was not marked as volatile, the code sometimes did not see the new value updated by the creation task.
By marking the fields as `volatile` we assure that their updated values are seen by other threads.

### DIFF
--- a/platform/openide.loaders/src/org/openide/loaders/FolderInstance.java
+++ b/platform/openide.loaders/src/org/openide/loaders/FolderInstance.java
@@ -107,28 +107,28 @@ public abstract class FolderInstance extends Task implements InstanceCookie { //
     protected DataFolder folder;
     
     /** container to work with */
-    private DataObject.Container container;
+    private final DataObject.Container container;
 
     /** map of primary file to their cookies (FileObject, HoldInstance) */
-    private HashMap<FileObject, HoldInstance> map = new HashMap<FileObject, HoldInstance> (17);
+    private final HashMap<FileObject, HoldInstance> map = new HashMap<>(17);
 
     /** Array of tasks that we have to check before we are ok. These are the tasks
      *  associated with children of the current folder.
      */
-    private Task[] waitFor;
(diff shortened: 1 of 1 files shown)


lkishalmi (approved):
Nice catch!

platform/openide.loaders/src/org/openide/loaders/FolderInstance.java
```diff
@@ -115,14 +115,14 @@
     /** Array of tasks that we have to check before we are ok. These are the tasks
      *  associated with children of the current folder.
      */
-    private Task[] waitFor;
+    private volatile Task[] waitFor;
```
JaroslavTulach, line 118: Volatile applies only to reading the array pointer, but not to accessing the elements of the array. Is that OK?

JaroslavTulach (approved):
I am usually ashamed when I look back at my old multithreaded code after all those years. This is the case as well. The code using `FolderInstance` was randomly failing for many years, should your change fix that, then you deserve tons of kudos, Martine!

Making as much fields `final` is the best option. Some of the fields however have to be mutable. There is some `synchronize(this)` guard for some and the question is whether more appropriate to guard more fields by `this` or make more fields `volatile`?

platform/openide.loaders/src/org/openide/loaders/FolderInstance.java
entlicher: It's a pity we can not make `folder` final, since it's API.

entlicher: In addition to the original change fields `container`, `map`, `listener` and `err` are now final, field `creationTask` is volatile.

PR #3636, opened 2022-02-18
JSF: Declaration of the xmlns namespace is reported as missing library

matthiasblaesing:


vieiro (approved):

vieiro:
It's surprising this was reported as a missing library.

matthiasblaesing:
I added a missing "//NOI18N" marker and will merge when Unittests come back.

The reason for the missing library report is, that all namespace mappings are expected to be JSF libraries. What was overlooked is, that there might be a reason to declare a schema location (in our case the XHTML namespace is referenced). Other might map other normal XML namespaces, but I wanted to keep this minimal and a straight forward fix.

vieiro:
@matthiasblaesing Yep. In the future I think we should have a list of most used namespaces in the web we should be skipping (say SVG). 

Note that is you skip the XML Schema namespace you may be wanting to skip XLink as well, which is a common companion of XML Schema.

PR #3638, opened 2022-02-19
[cnd] 3.2.1 Merge master into cnd

vieiro:
Let's bring master bits (that include LSP 3.16 specification support in ide/lsp.client) into cnd.


matthiasblaesing (approved):
Sounds sane to me..

PR #3642, opened 2022-02-19
Improve CSS Support

matthiasblaesing:
Introduce partitial support for CSS Selector Level 4 (:not, :where, :is)

Before:

before2

After:

after2

Improve parsing of the contents of @page declarations

Before:

before

After:

after


matthiasblaesing:
I intent to merge this next weekend. If you want to object, feel free to do so.

java/java.api.common/build.xml
```diff
@@ -21,5 +21,21 @@
 -->
 <project basedir="." default="build" name="java/java.api.common">
     <import file="../../nbbuild/templates/projectized.xml"/>
+
+    <target name="generate-antlr-parser" description="Regenerate the CSS parser from the ANTLR3 grammar">
```
vieiro, line 25: Do we want to add a dependency with the compile task like so?
```
<target name="compile" depends="generate-antlr-parser">
</target>
```

vieiro (approved):
Looks sane. Good to see we're generating files from grammars. Just a little comment: we may want to regenerate files from grammars before compiling (in case one forgets to regenerate them by hand).

java/java.api.common/build.xml
matthiasblaesing: Thank you for the review. I actually thought about this (and planned to remove the generated java files from the repository), but the generated code is not reproducible.
Antlr3 injects the generation time and paths into the files and at least from a quick look, this could not be overriden. Debian carries a patch the also changes the source, so that reproducible datastructures are used (https://sources.debian.org/patches/antlr3/3.5.2-9/05-reproducible-parsers.patch/). 
So the code will change on each build and from my perspective this does not scale. In this case IMHO it is better to keep the generated parsers and regenerate them when needed.

vieiro: I see. Good to avoid chaning the code on each build then.

ebarboni:
Hi, I have a very naive question. I only see that for css we use antlr 3.x and the current antlr is 4.x. I'm not aware how many grammar we have in codebase. But would it be an idea to migrate to last ?

matthiasblaesing:
Without a good argument I think it would be no benefit. Rewriting the grammar just poses chances to break the code.

vieiro:
@ebarboni @matthiasblaesing  AFAIK Antlr4 exposes the AST through the visitor pattern (or through events), i.e., you don't have it in memory. The migration Antlr3 to Antlr4 is not straightforward. You'll have to build the AST yourself (see, https://stackoverflow.com/questions/29971097/how-to-create-ast-with-antlr4 for instance).

PR #3644, opened 2022-02-19
LSPBindings synchronization fix

vieiro:
Adds some extra synchronized blocks around those non thread-safe maps.


matthiasblaesing (approved):
If I read this correctly, the intention is to protect `workspace2Extension2Server` and `project2MimeType2Server` from access from multiple threads. Looks sane to me.

vieiro:
Exactly. There're some "synchronized" methods in LSPBiindings, that do protect these maps from being accessed from multiple threads, but these small parts were missing.

PR #3645, opened 2022-02-19
[NETBEANS-6468] Fix Windows LAF on Windows 11 and Java 17

eirikbakke:
When running NetBeans with the Windows LAF on Windows 11 and Java 17, parts of the LAF revert to an older look. See screenshot below.

This is due to the "os.name" system property being checked for the value "Windows 10" in various places. With Java 17 and Windows 11, this property can now have the value "Windows 11". This PR goes through each occurrence of "Windows 10" and also permits "Windows 11" in these cases.

220219 Windows LAF problem on Windows 11

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/view/ui/CloseButtonTabbedPane.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/view/ui/CloseButtonTabbedPane.java
@@ -339,6 +339,11 @@ final class CloseButtonTabbedPane extends JTabbedPane implements PropertyChangeL
             || (osName.equals( "Windows NT (unknown)" ) && "10.0".equals( System.getProperty("os.version") ));
     }
 
+    private static boolean isWindows11() {
+        String osName = System.getProperty ("os.name");
+        return osName.indexOf("Windows 11") >= 0;
+    }
+
     private boolean isWindowsVistaLaF() {
         String osName = System.getProperty ("os.name");
         return osName.indexOf("Vista") >= 0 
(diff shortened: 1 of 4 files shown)


matthiasblaesing (approved):
Change looks sane to me.

eirikbakke:
Thanks! Will merge.

neilcsmith-net:
While I agree with @matthiasblaesing that the changes look sane in themselves, I do wonder whether we should be simplifying some of this?  Adding extra complexity while keeping logic and LaF support for versions of Windows that NetBeans (or platform) can't actually run on doesn't seem a good idea.

eirikbakke:
> for versions of Windows that NetBeans (or platform) can't actually run on doesn't seem a good idea

Are you proposing we should remove the old Windows XP LAF code?

neilcsmith-net:
It would certainly simplify the logic here.  AFAIK, neither Microsoft or OpenJDK 11+ supports a version of Windows lower than 8.1, and then probably not for much longer.  So I do think that dropping unused code for older versions when adding newer versions should at least be considered.

neilcsmith-net:
I'd agree until we end up with more and more convoluted `if` clauses where one side cannot be hit or properly tested anyway.

Incidentally, we're no longer using JIRA and have no need for tickets to link PRs against.

eirikbakke:
Yeah, maybe the next time this breaks we can consider a bigger LAFectomy. There's also the old separate "Windows Classic" LAF theme, which really doesn't need to be shown as an option to users in the Options panel.

(The repo README still says "Bugs should be reported to https://issues.apache.org/jira/projects/NETBEANS/issues/", so I assumed the switch wasn't official yet.)

neilcsmith-net:
> (The repo README still says "Bugs should be reported to https://issues.apache.org/jira/projects/NETBEANS/issues/", so I assumed the switch wasn't official yet.)

PR welcome on that one! :smile:  That's annoying actually as I updated the README for 13 build requirements but forgot that.  Links elsewhere (eg. website) have been updated.  Switch is done and infra@ asked to partly switch off JIRA.

mbien:
> My usual approach is to avoid touching big pieces of old code until it actually gets in the way of something new. There's always the danger of breaking something unintentionally.

In projects like netbeans (old, big code base with the intention to stick around, while not very concerned about having to backport bugfixes) I try to do both. Fix the issue, but also leave the general area of the fix a bit cleaner than it was before if there is such opportunity. Cleanups should be its own commit if possible to make reviews easier and not influence git bisects - but i think its ok to have it in the same PR, right after the fix.

PR #3646, opened 2022-02-19
Update eclipselink from 2.7.9 to 2.7.10

pepness:
Library Notes:

- This is mainly a maintenance release and contains bug fixes
- Use jakarta.persistence instead of javax.persistence (still using javax.* in packages)
- Deprecated Functionality: None

NetBeans Testing:
- Full build done
- Verify successful execution of unit tests for modules `websvc.restlib`, `j2ee.eclipselink`, `j2ee.eclipselinkmodelgen`, `j2ee.jpa.refactoring`, `j2ee.metadata.model.support`, `j2ee.persistence` and `j2ee.persistenceapi`
- Checked the files are in the Ant Library Manager
- Successfully create project as per [NetBeans Documentation](https://netbeans.apache.org/kb/docs/websvc/rest.html), with Payara and Glassfish

[EclipseLink Web Page](https://www.eclipse.org/eclipselink/)
[EclipseLink Release Notes](https://github.com/eclipse-ee4j/eclipselink/releases/tag/2.7.10)


java/j2ee.eclipselink/nbproject/project.xml
```diff
@@ -26,29 +26,126 @@
             <code-name-base>org.netbeans.modules.j2ee.eclipselink</code-name-base>
             <module-dependencies/>
             <public-packages>
-                <subpackages>org.eclipse</subpackages>
+                <subpackages>org.eclipse.persistence</subpackages>
```
matthiasblaesing, line 29: What is the reason for this change? Are there internal packages, that shall be hidden? If only a single package should be exposed, I would expect `package`, `subpackages` matches all packages where the given name is a prefix.

matthiasblaesing (approved):
In general this looks good. There is one change, that I don't understand, please see inline.

mbien:
sigtests are failing on this PR even after restart.

java/j2ee.eclipselink/nbproject/project.xml
pepness: I was under the impression that I should only include `subpackages` that have java classes inside and exclude internal packages like `org.eclipse.persistence.internal`. What would be the proper way of configure `<public-packages>` and `<subpackages>`, include all packages excluding internal packages?

matthiasblaesing: If you want to expose only the package and not the package where it is a prefix of, `package` is the right element. Your assessment is right, internal should not be exposed.

The sigtest fails from my interpretation because classes are missing:

https://app.travis-ci.com/github/apache/netbeans/jobs/560459188#L12806-L12808

This smells like these:
https://search.maven.org/artifact/jakarta.resource/jakarta.resource-api/1.7.4/jar
https://search.maven.org/artifact/jakarta.transaction/jakarta.transaction-api/1.3.3/jar

pepness: I restore the entry `<subpackages>org.eclipse</subpackages>`, deleted the added entries, regenerate sigtest and everything worked fine. I suppose that if another module need a class from this packages it would fail the build process.

mbien:
tests are green, PR is approved / no outstanding change requests as far as I see -> merging.

PR #3649, opened 2022-02-20
Cleanup deprecated constructor use

BradWalker:
Cleaning some old code to make it more current..

- removed use of old deprecated constructor use
- use autobox/unbox for encapsulation
- change chain-of-if statement to switch w/ string
- some code cleanup

---

### DIFF
--- a/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
+++ b/profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
@@ -32,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
-
 /**
  *
  * @author Tomas Hurka
@@ -59,9 +58,9 @@ class ClassDumpSegment extends TagBounds {
     final int superClassIDOffset;
     ClassDump java_lang_Class;
     boolean newSize;
-    Map<JavaClass,List<Field>> fieldsCache;
-    private List<JavaClass> classes;
-    private Map /*<Byte,JavaClass>*/ primitiveArrayMap;
+    Map<JavaClass, List<Field>> fieldsCache;
+    private List<ClassDump> classes;
+    private Map<Integer, ClassDump> primitiveArrayMap;
 
     //~ Constructors -------------------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


profiler/lib.profiler/src/org/netbeans/lib/profiler/heap/ClassDumpSegment.java
```diff
@@ -235,55 +233,118 @@ void addInstanceSize(ClassDump cls, int tag, long instanceOffset) {
...
+                    java_lang_Class = jcls;
+                    break;
+                    
+                case "java.lang.Object":
+                    // NOI18N
+                    java_lang_Object = jcls;
+                    break;
+                    
+                default:
+                    break;
```
mbien, line 347: the intention is good but i think this makes it more difficult to read.

try to format it like:
```java
               case "[Z":  typeObj = HprofHeap.BOOLEAN;  break; // NOI18N
```
Its non-standard but would be compact and easier to read in this particular case. The "// NOI18N" hint must be on the same line as the String literal. 

Formatting it like this is also really close to switch expressions of later JDKs. Updating to them would only require the "->" and removal of "break".

But most importantly: we have to be careful with null here. What if vmName is null? The switch will throw an exception, the if-else would do fine.

profiler/lib.profiler/src/org/netbeans/lib/profiler/server/ProfilerRuntimeCPU.java
```diff
@@ -338,14 +338,29 @@ static void writeServletDoMethod(ThreadInfo ti, String method, String servletPat
...
-            methodId = 1;
-        } else if ("POST".equals(method)) { // NOI18N
-            methodId = 2;
-        } else if ("PUT".equals(method)) { // NOI18N
-            methodId = 3;
-        } else if ("DELETE".equals(method)) { // NOI18N
-            methodId = 4;
+        if (null != method) switch (method) {
+            case "GET":
+                // NOI18N
```
mbien, line 343: same here: // NOI18N must be on the same line behind case "foo"

mbien:
@BradWalker  please don't merge PRs without giving others a chance to review - even if it seems trivial.

BradWalker:
@mbien , sorry about that.. I got a little bit aggressive with this.. My sincere apologies.. I'll get it reverted..

PR #3653, opened 2022-02-21
ClientCommandProvider split from CodeActionsProvider. NbLanguageClient moved to API package.

sdedic:
The PR starts to separate (so far) internal NBLS APIs from the implementation classes. Two refactorings are here:
- `CodeActionsProvider` was misused in the past to serve as a SPI for adding commands. The interface was split into two
- `NbCodeLanguageClient`, `NbCodeClientCapabilities` and `LspServerUtils` were moved to an API package.


java/java.lsp.server/src/org/netbeans/api/lsp/server/NbCodeLanguageClient.java
```diff
@@ -26,10 +26,18 @@
...
+import org.netbeans.modules.java.lsp.server.protocol.ShowStatusMessageParams;
+import org.netbeans.modules.java.lsp.server.protocol.TestProgressParams;
 
 /**
  * An extension to the standard LanguageClient that adds several messages missing
- * from the official LSP protocol.s
+ * from the official LSP protocol.
  * @author sdedic
  */
 public interface NbCodeLanguageClient extends LanguageClient {
```
JaroslavTulach, line 43: Was this class ever meant to be an API? It is an interface. How do you plan to evolve it in the future?

java/java.lsp.server/src/org/netbeans/spi/lsp/server/ClientCommandProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Processes a command. Only commands previously reported by {@link #getCommands} are to be expected, the infrastructure dispatches
+     * commands only to provider(s) that have declared support for them.
+     * 
+     * @param client the client which executed the command.
+     * @param command the command
+     * @param arguments arguments
+     * @return completion handle or {@code null} to reject the command.
+     */
+    public CompletableFuture<Object> processCommand(NbCodeLanguageClient client, String command, List<Object> arguments);
```
JaroslavTulach, line 42: Possible way to avoid moving an internal interface into API: Replace `NbCodeLanguageClient` with `Context` and define:

```java
public static final class Context {
  private final NbCodeLanguageClient client;
}
```
an inner class of this class? Add only methods that are supposed to be part of the API.

java/java.lsp.server/src/org/netbeans/spi/lsp/server/ClientCommandProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    
+    /**
+     * Returns a set of supported commands. Called early during client initialization handshake to collect all supported
+     * commands for the LSP client.
+     * @return Set of supported commands. 
+     */
+    public Set<String> getCommands();
```
JaroslavTulach, line 49: Some annotation based registration would be preferable - especially if multiple modules start to contribute these commands.

JaroslavTulach (commented):
Thanks for picking up this challenge, @sdedic ! It is desperately needed to turn the currently hard-coded commands into such `ClientCommandProvider` abstraction and de-couple them as much as possible from the actual server implementation.

java/java.lsp.server/src/org/netbeans/spi/lsp/server/ClientCommandProvider.java
sdedic: OK, should the `Context` (if used specifically for `ClientCommandProvider`) contain also the `command` and `arguments` parameters ?

java/java.lsp.server/src/org/netbeans/api/lsp/server/NbCodeLanguageClient.java
sdedic: It has to (somehow) - otherwise the code wouldn't be able to call things like `showInputBox`. Additional messages can be added as `default` methods throwing UnsupportedOpEx. There are two additional ways: `final` class with protocol messages that will delegate all methods to NbCodeLanguageClient (internal API). Or the same with the `Context` proposed below.

java/java.lsp.server/src/org/netbeans/spi/lsp/server/ClientCommandProvider.java
JaroslavTulach: That implies a life-cycle: new `Context` for each invocation of the command. But that's probably OK.

java/java.lsp.server/src/org/netbeans/api/lsp/server/NbCodeLanguageClient.java
JaroslavTulach: Why not call `DialogDisplayer.getDefault().notifyFuture(new NotifyDescriptor.InputLine(...))` instead of `showInputBox`?

sdedic: Eh ;) bad example then. Let's pick a message that is NOT bridged from regular NetBeans APIs, e.g. `disposeTextEditorDecoration` or `notifyNodeChange`.

JaroslavTulach: If you want to expose whole LSP - then providing access to `LanguageClient` may be the right thing. If you want to expose it including NetBeans extensions, possibly exposing `NbCodeLanguageClient` is the right thing - but then the protocol has to become (relatively) stable API - a review of our adhoc messages may be desirable (that's what we haven't paid much attention to yet).

java/java.lsp.server/nbproject/project.xml
```diff
@@ -757,6 +757,9 @@
             </test-dependencies>
             <public-packages>
                 <package>org.netbeans.modules.java.lsp.server.ui</package>
+                <package>org.netbeans.api.lsp.server</package>
+                <package>org.netbeans.spi.lsp.server</package>
+                <package>org.netbeans.modules.java.lsp.server.ui</package>
```
dbalek, line 762: Already marked as public ?

Issue #3559, opened 2022-02-05
ide/lsp.client race condition


vieiro:
### Apache NetBeans version

Latest release

### What happened

This applies to the ide/lsp.client in the cnd branch, possibly in master as well.

LanguageServerProvider.startServer is invoked twice even though the first invocation does initialize the server properly. 

The first time the initialization comes from [BreadcrumbsImpl:337](https://github.com/apache/netbeans/blob/8fc91fa12c4c0e04c42bef9907604aeb1229cb6f/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/BreadcrumbsImpl.java#L337) and the second one from [CustomIndexerImpl.java:68](https://github.com/apache/netbeans/blob/8fc91fa12c4c0e04c42bef9907604aeb1229cb6f/ide/lsp.client/src/org/netbeans/modules/lsp/client/bindings/CustomIndexerImpl.java#L68). 

The result is that even though the server was initialized properly, the ServerRestarter is invoked to restart the server. 

### How to reproduce

Use the "clangd" LSP server. Do not implement a ServerRestarter (or do not restart the server). The server complains with:

```
V[21:26:15.044] <<< {"id":"1","jsonrpc":"2.0","method":"initialize","params":{"capabilities":{"textDocument":{"documentSymbol":{"hierarchicalDocumentSymbolSupport":true,"symbolKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]}}},"workspace":{"symbol":{"symbolKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]}},"workspaceEdit":{"documentChanges":true,"resourceOperations":["create","delete","rename"]}}},"processId":0,"rootPath":"/home/antonio/NetBeansProjects/CppApplication_1","rootUri":"file:///home/antonio/NetBeansProjects/CppApplication_1/"}}

I[21:26:15.044] <-- initialize("1")
I[21:26:15.048] --> reply:initialize("1") 4 ms
[...]

PR #3655, opened 2022-02-21
ide/lsp.client + MultiMimeLanguageServerProvider

vieiro:
Some LSP servers can handle files with different mime-types within the same project. 

For instance, "clangd", "ccls" and "apple/source-kit" are able to handle "text/x-c", "text/x-cpp" and others. C++ projects may include simple C files within the same project. 

This adds a new SPI interface "MultiMimeLanguageServerProvider" that returns a set of mime-types handled by the LSP server. 

The commit registers/deregisters all these mime-types at once. Otherwise the same LSP server may be started (restarted?)  several times for different mime types ( [in this loop](https://github.com/apache/netbeans/blob/76be2da4de40da05cd9f6c978d1e29bc16f99c1f/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java#L270)) possibly losing some previous state or causing multiple `initialize` requests to be sent (as reported in #3559).

### DIFF
--- a/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
+++ b/ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
@@ -255,8 +255,10 @@ public class LSPBindings {
             synchronized (LSPBindings.class) {
                 ServerDescription description = project2MimeType2Server.getOrDefault(baseUri, Collections.emptyMap()).remove(mt);
                 // Remove any other mimetypes as well.
-                for(String anotherMT: description.mimeTypes) {
-                    project2MimeType2Server.get(baseUri).remove(anotherMT);
+                if (description != null) {
+                    for(String anotherMT: description.mimeTypes) {
+                        project2MimeType2Server.get(baseUri).remove(anotherMT);
+                    }
                 }
                 Reference<LSPBindings> bRef = description != null ? description.bindings : null;
                 LSPBindings b = bRef != null ? bRef.get() : null;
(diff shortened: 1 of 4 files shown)


ide/lsp.client/src/org/netbeans/modules/lsp/client/LSPBindings.java
```diff
@@ -269,6 +280,12 @@ private static LSPBindings buildBindings(ServerDescription inDescription, Projec
 
         for (LanguageServerProvider provider : MimeLookup.getLookup(mt).lookupAll(LanguageServerProvider.class)) {
             final Lookup lkp = prj != null ? Lookups.fixed(prj, mimeTypeInfo, restarter) : Lookups.fixed(mimeTypeInfo, restarter);
+            inDescription.mimeTypes = Collections.singleton(mt);
+            // If this is a MultiMimeLanguageServerProvider, then retrieve all 
+            // mime types handled by this server.
+            if (provider instanceof MultiMimeLanguageServerProvider) {
+                inDescription.mimeTypes = ((MultiMimeLanguageServerProvider)provider).getMimeTypes();
+            }
```
matthiasblaesing, line 288: I would make a defensive copy in line 283, creating a new HashSet at that point from the core mimetype and adding the other mimetypes to that. Else the provider could change the backing set after the description was used for starting, causing an imbalance (yes not really realistic, but ...).

vieiro: Thanks for the tip!

vieiro:
If nobody opposes, this will be merged by next weekend.

PR #2498, opened 2020-10-28
WIP: Removed primitive wrapper constructor calls from the platform cluster

lkishalmi:

PR #3656, opened 2022-02-22
Clean up the use of deprecated numeric constructors

BradWalker:
This change cleans up the use of old deprecated numeric constructors..

Warnings like this will no longer happen:

```
[repeat] /home/bwalker/src/netbeans/java/testng/src/org/netbeans/modules/testng/AbstractTestGenerator.java:1700: warning: [deprecation] Float(float) in Float has been deprecated
[repeat]                     defValue = maker.Literal(new Float(0.0F));
[repeat]    
```
The newer implementation favor factory methods like `valueOf`() or `parseFloat`() since they are likely to
yield significantly better space and time performance by caching frequently requested values.

Also, the compiler/VM can do a better job of accessing the actual numeric values using autobox/unbox.

Most of these changes are in the enterprise/ide/java/profiler modules.

There are also some places that were changed to use the diamond operator.

---

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/wizard/NewNbModuleWizardIterator.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/wizard/NewNbModuleWizardIterator.java
@@ -259,7 +259,7 @@ public class NewNbModuleWizardIterator implements WizardDescriptor.AsynchronousI
             if (c instanceof JComponent) { // assume Swing components
                 JComponent jc = (JComponent) c;
                 // step number
-                jc.putClientProperty(WizardDescriptor.PROP_CONTENT_SELECTED_INDEX, new Integer(i)); // NOI18N
+                jc.putClientProperty(WizardDescriptor.PROP_CONTENT_SELECTED_INDEX, i); // NOI18N
                 // names of currently used steps
                 jc.putClientProperty(WizardDescriptor.PROP_CONTENT_DATA, steps); // NOI18N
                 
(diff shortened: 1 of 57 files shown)


mbien:
hi Brad,
the current code inspection in NB for this is not optimal. There is a PR in queue which attempts to fix some of it https://github.com/apache/netbeans/pull/3240

for example
```diff
-            int timeout = new Integer(startingStateTimeout);
+            int timeout = Integer.valueOf(startingStateTimeout);
```
should be `Integer.parseInt(startingStateTimeout)`, since the target type (`int timeout`) is primitive.


@lkishalmi has also a PR here https://github.com/apache/netbeans/pull/2498 which attempts the same cleanup. I don't know the state of it, maybe he can comment.

BradWalker:
> hi Brad, the current code inspection in NB for this is not optimal.

I fixed the cases where the code should be calling `parseInt()` instead of `valueOf()`. 

> There is a PR in queue which attempts to fix some of it #3240

I looked at this. My efforts stem from trying to reduce the warning messages. This PR #3240 , while admirable, are more comprehensive than what I'm proposing.

> @lkishalmi has also a PR here #2498 which attempts the same cleanup. I don't know the state of it, maybe he can comment.

I looked at the work he did and I'm happy to take it over if that helps. But, the work that @lkishalmi  was doing was really more comprehensive than what I was trying to do.

I'm simply trying to keep my changes "small" and self contained. This way the impact is very minimal.

@mbien , how do  the changes look now?

lkishalmi:
As of mine at #2498 some tests are failing especially around the Clipboard handling (if I remember well) which I was not able to resolve. Probably doing this in smaller steps would be better. I'm going to close #2498 in favor of Brad's efforts.

mbien:
@BradWalker I took a quick look and it looks fine to me, could you mention in the commit which areas of NB the cleanup is covering? Otherwise they all will look the same in the log.

mbien:
I am sure you know, but I just wanted to add that we have to be super careful in refactorings like this, if some code does identity comparisons with ==, it can break.  Thats the reason why #2498 got stuck.
```java
        System.out.println(new Integer(5) == new Integer(5));       // false
        System.out.println(new Integer(5) == Integer.valueOf(5));   // false
        System.out.println(5 == new Integer(5));                    // true
        System.out.println(5 == Integer.valueOf(5));                // true
...
```

BradWalker:
> I am sure you know, but I just wanted to add that we have to be super careful in refactorings like this, if some code does identity comparisons with ==, it can break. Thats the reason why #2498 got stuck.

Agreed.. That's why I really try to do everything by hand. So, hopefully, I don't introduce an error as a result of the work.

PR #3660, opened 2022-02-23
about window update

mbien:
before:
About

after:
About2

 - layout tweaks
 - some code cleanup / language level upgrade

context:
inspired by a discussion on NB slack.

The two text panels are in scroll areas which makes it a bit difficult to compute their size for the window. (the point of a scroll area is to not care about the dimensions of its content).

The lower panel has predictable content, so I disabled the vertical scrollbar and set the horizontal scrollbar to show-always so that it is part of the layout computation. Height is now set to default, since it can size itself now.

The upper panel had fixed size, it still has, it just lost around 50px in height. Btw should we add a github link there?

Rest are minor cleanups in the involved areas.

### DIFF
--- a/platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.form
+++ b/platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.form
@@ -41,9 +41,9 @@
               <EmptySpace max="-2" attributes="0"/>
               <Group type="103" groupAlignment="1" attributes="0">
                   <Component id="imagePanel" alignment="0" pref="190" max="32767" attributes="0"/>
-                  <Component id="jScrollPane2" alignment="0" pref="190" max="32767" attributes="0"/>
-                  <Component id="jScrollPane3" alignment="1" pref="190" max="32767" attributes="0"/>
-                  <Component id="jPanel1" alignment="0" pref="190" max="32767" attributes="1"/>
+                  <Component id="descriptionScrollPane" alignment="0" pref="190" max="32767" attributes="0"/>
+                  <Component id="copyrightScrollPane" alignment="1" pref="190" max="32767" attributes="0"/>
+                  <Component id="buttonPanel" alignment="0" pref="190" max="32767" attributes="1"/>
               </Group>
               <EmptySpace max="-2" attributes="0"/>
           </Group>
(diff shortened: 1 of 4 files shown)


vieiro:
I can't see this red face in my "About" dialog. I do want one of these!!
imagen

platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.form
```diff
@@ -90,13 +98,15 @@
         </Component>
       </SubComponents>
     </Container>
-    <Container class="javax.swing.JScrollPane" name="jScrollPane3">
+    <Container class="javax.swing.JScrollPane" name="copyRightScrollPane">
```
neilcsmith-net, line 101: (very) minor niggle - should be `copyrightScrollPane`

neilcsmith-net (commented):
Not sure about the bottom panel changes. Stops the text wrapping?  Do we really want that or not?

> Btw should we add a github link there?

No! Well, at least no direct link anyway. All inbound links from the IDE should be via a prefix of `https://netbeans.apache.org/nb/***`. That way we can manage them all without changes in the IDE - see https://github.com/apache/netbeans-website/blob/master/netbeans.apache.org/src/content/.htaccess  Could add a relevant link there forwarding directly to GitHub if required?  Almost did this with the link for `Help / Report Issue` but decided to keep the intervening web page - at least we have the option to change at any time though.

> I can't see this red face in my "About" dialog. I do want one of these!!

Seconded! :smile:

neilcsmith-net:
> On this note, on my windows setup I still have an update label the claims I'm on updated version "NetBeans 8.2 Patch 2"

@emaphis this is because you've added the 8.2 update centre? It's taken from the info in https://updates.netbeans.org/netbeans/updates/8.2/uc/final/distribution/catalog.xml.gz The updates line in ProductInformationPanel should probably be branded out entirely, but in the IDE branding only (ie. in `nb/ide.branding`)

platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.form
mbien: oh, yeah i can fix that :)

mbien:
> Not sure about the bottom panel changes. Stops the text wrapping? Do we really want that or not?

Not 100% what you mean, you can see on the screenshot that there is a newline after "Cache directory:". So it does some wrapping. It has a horizontal scrollbar too (unchanged). The main difference is that it has no vertical scroll bar, so that it can size itself. Keep in mind the second screenshot is a dev build with much longer paths - the original version would scroll too.

The reason why i set the horizontal one to always show (FlatLAF hides it anyway if its not needed, which is nice) is because otherwise it would appear after the size has been calculated, and once it appears it will also trigger the vertical scrollbar since the viewport is now smaller - this is a common problem with layouts like this one.

neilcsmith-net:
@mbien sorry, yes, took another look. The `System:` line not wrapping confused me, but I guess that's just an artefact of the longer paths. And yes, understand the scrollbar / layout calculation issue.

All good - sorry for the noise! :+1:

eirikbakke:
Thanks for this patch! I will test this in my next NetBeans build, in a day or two.

eirikbakke (approved):
Looks good to me! Tested in a private NetBeans IDE build on Windows, and on a NetBeans Platform application build on Windows and MacOS (FlatLAF Light in both cases).

eirikbakke:
Another issue, which I guess was mentioned earlier. When there are long lines in the bottom panel, we end up having both horizontal scrollbars and text wrapping. Shouldn't it be just one or the other? In the example below, why is there no text wrapping after "User directory:"?

Wrap

mbien:
changed it so that the lower panel won't wrap text anymore. 

Looking at your screenshot again, it appears as if the window is actually slightly larger than it should be, since the image is centered, it isn't on the same line as the start of the text. Without the scaling the image defines the size of the window, and everything is aligned properly I suppose. But you should see the same effect with the old version of the window, right?

mbien:
@eirikbakke if this PR doesn't make it worse for high dpi / UI scaling modes i would like to merge this anyway. Someone else interested in this would have to debug it. I am running 1080p screens.

eirikbakke:
@mbien Fine with me.

Issue #3662, opened 2022-02-23
Gradle Compile a file in buildSrc folder fails with a NullPointer


maerevoet-wim:
### Apache NetBeans version

Apache NetBeans 12.6

### What happened

If you open a Custom Build Logic project and try to compile a file you get an NullPointerException

	at org.netbeans.modules.gradle.spi.GradleFiles.searchPropertyFiles(GradleFiles.java:108)
	at org.netbeans.modules.gradle.spi.GradleFiles.getPropertyFiles(GradleFiles.java:179)

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10

### JDK

Java 11 AdoptOpenjdk

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

neilcsmith-net:
Removing triage label - discussion can happen in #3663

mbien:
@lkishalmi is this fixed via #3795? If yes, please link and close.

lkishalmi:
Yes, shall be resolved by #3795.

PR #3663, opened 2022-02-23
Fix getFile returning null

maerevoet-wim:
Fix issue #3662 : Gradle Compile a file in buildSrc folder fails with a NullPointer

Situation:
Gradle project has a custom build logic project under it. ( Build Script - Custom Build Logic )
In Netbeans you can open the project
If you select a separate file - Right Click - Compile, you get the exception
If you try to build the project, you get the exception

	at org.netbeans.modules.gradle.spi.GradleFiles.searchPropertyFiles(GradleFiles.java:108)
	at org.netbeans.modules.gradle.spi.GradleFiles.getPropertyFiles(GradleFiles.java:179)
	at org.netbeans.modules.gradle.api.execute.GradleCommandLine.addGradleSettingJvmargs(GradleCommandLine.java:804)
	at org.netbeans.modules.gradle.api.execute.GradleCommandLine.configure(GradleCommandLine.java:838)
	at org.netbeans.modules.gradle.execute.GradleDaemonExecutor.run(GradleDaemonExecutor.java:207)
	at org.netbeans.core.execution.RunClassThread.doRun(RunClassThread.java:132)
[catch] at org.netbeans.core.execution.RunClassThread.run(RunClassThread.java:81)

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/spi/GradleFiles.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/spi/GradleFiles.java
@@ -105,7 +105,7 @@ public final class GradleFiles implements Serializable {
         List<File> ret = new ArrayList<>(3);
         for (Kind kind:Kind.PROPERTIES){
             File f = getFile(kind);
-            if (f.exists()){
+            if (f != null && f.exists()){
                 ret.add(f);
             }
         }


neilcsmith-net:
Thanks @maerevoet-wim - note you don't need to open an issue to file a pull request, and we'd rather you didn't.  But more description in the pull request of what triggers this issue would be good.  Have requested review from @lkishalmi and @sdedic. Curious if this is fixing one bug by hiding another one?

maerevoet-wim:
I have no idea if this hides another bug. 
But the code getFile(Kind) does return only for a specific set of 'Kind' a file in a buildSrc project, otherwise it returns null.

sdedic:
@maerevoet-wim could you please elaborate more on the reproduction steps ? I have checked out `gradle` project, can indeed open `build-logic` ... but there are no compilable sources in it and NetBeans Dev complain about build-logic's subprojects ('org.codehaus.groovy.runtime.metaclass.MissingPropertyExceptionNoStack: No such property: getFreeCompilerArgs for class: org.jetbrains.kotlin.gradle.dsl.KotlinJvmOptionsImpl`).

Where should I right-click to get Compile ?

lkishalmi:
This exception just happened with me yesterday. I put aside to create a unittest for this one.

maerevoet-wim:
@sdedic 
For example if you use a conventin plugin in a multi project structure.
See https://docs.gradle.org/current/userguide/declaring_dependencies_between_subprojects.html#sec:depending_on_output_of_another_project

lkishalmi:
Resolved as #3795

PR #3666, opened 2022-02-23
Export the Truffle packages explicitly, since TruffleJDKServices class was removed.

entlicher:
In Truffle debugger we call Truffle API from an unnamed module. In order to be able to call Truffle APIs we need to export the Truffle packages to our module. We used `TruffleJDKServices.exportTo()`, but that class was recently removed from GraalVM.
We suspend at `Truffle.getRuntime()` (impl in `TruffleDebugManager`) and call `Module.addExports(...)` with the packages we are using (impl in `RemoteServices`).


JaroslavTulach (approved):
Shouldn't there be a test/gate failing somewhere? Removals happen, but it would be good to get an early warning next time.

entlicher:
Gate does not fail, because it uses an old GraalVM. As soon as it uses 22.1, it would fail.
The integration test in GraalVM internal repo started to fail right after the removal of `TruffleJDKServices`.
Our tests pass in this issue after we've removed the dependency on `TruffleJDKServices`. That's the verification of this change.

PR #3668, opened 2022-02-24
Ignore dependencies for BOM-like projects.

sdedic:
Following suggestion in [NETBEANS-5846](https://issues.apache.org/jira/browse/NETBEANS-5846). The ProjectInfoBuilder in `gradle-tooling` was changed so it does not produce `unresolved_problems` if the project uses `javaPlatform` plugin w/ dependencies enabled. 

The 2nd commit is a bugfix - during testing, I used `Reload Project` action and the project did not degrade to the new (broken) state after change in buildscript + reload. My fault, the `force` parameter to loadProject was not used at all.


ppisl (approved):

lkishalmi:
Looks good, though you might also would like to bump this number: https://github.com/apache/netbeans/blob/443c86cb8233eccb28bb9edb0b46c8d5529f4e2d/extide/gradle/src/org/netbeans/modules/gradle/cache/ProjectInfoDiskCache.java#L42

That would make disk cache invalid and force NB to re-load the project info.

sdedic:
Feedback incorporated, 3 days from approval; merging.

PR #3674, opened 2022-02-25
Adding Open Service Console action

jhorvath:
Adding action for opening DB Service Console to vscode


.gitignore
```diff
@@ -46,6 +46,7 @@
 /platform/o.n.bootstrap/launcher/windows/*.dll
 /platform/o.n.bootstrap/launcher/windows/*.exe
 /platform/o.n.bootstrap/launcher/windows/*.res
+/java/java.lsp.server/vscode/*.vsix
```
JaroslavTulach, line 49: Why would be a `.vsix` file created in this directory?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/VSCodeURLDisplayer.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.net.URL;
+import org.netbeans.modules.java.lsp.server.protocol.HtmlPageParams;
+import org.openide.awt.HtmlBrowser.URLDisplayer;
+
+/**
+ *
+ * @author Jan Horvath
+ */
+public class VSCodeURLDisplayer extends URLDisplayer {
```
JaroslavTulach, line 29: How is this instance registered?

The preferred way of registering these services is to create `AbstractLspURLDisplayer` with all the implementation, but without registering it. Then Create a subclass in `nbcode/integration` module with `@ServiceProvider` registration.

Make the class [abstract with protected constructor](https://github.com/apache/netbeans/blob/36443ae276eee4a2b03fd59ce28af02b576e53fd/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspStatusDisplayer.java#L29).

JaroslavTulach (changes requested):
I am not sure how it can work? How is the `VSCodeURLDisplayer` registered, how is its instance created?

.gitignore
jhorvath: It was created by manually invoking `vsce package`, removing.

enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/OpenServiceConsoleAction.java
```diff
@@ -16,33 +16,33 @@
...
 import org.openide.awt.ActionReferences;
 import org.openide.awt.ActionRegistration;
+import org.openide.awt.HtmlBrowser.URLDisplayer;
 import org.openide.util.Exceptions;
 import org.openide.util.NbBundle.Messages;
 
 @ActionID(
         category = "Tools",
-        id = "org.netbeans.modules.cloud.oracle.OpenServiceConsoleAction"
+        id = "org.netbeans.modules.cloud.oracle.actions.OpenServiceConsoleAction"
```
JaroslavTulach, line 36: Changing ID of already published action will make all the previous usages invalid. The reason for introducing `id` attribute and not using an ID derived from name of the class what exactly this: refactoring moves shall not break other modules referencing the action.

JaroslavTulach (approved):

PR #3677, opened 2022-02-26
HuntDiff: make public final

nbauma109:
This class is useful for standalone applications that use netbeans-modules-diff (i.e. https://github.com/GeertjanWielenga/netbeans-visual-diff-standalone).
While it says 'Internal', I think it would be more useful to make it public.
I'd rather make it public than see it copy/pasted in client programs.
Since everything is private static, it makes sense to make it also final.

---

### DIFF
--- a/ide/diff/src/org/netbeans/modules/diff/builtin/provider/HuntDiff.java
+++ b/ide/diff/src/org/netbeans/modules/diff/builtin/provider/HuntDiff.java
@@ -29,12 +29,12 @@ import java.util.regex.Matcher;
 import org.netbeans.api.diff.Difference;
 
 /**
- * Internal Diff algorithm.
+ * Diff algorithm.
  * 
  * @author Maros Sandor
  * @author Martin Entlicher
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
The provider implementation can be found in ([BuildInDiffProvider.java](https://github.com/apache/netbeans/blob/master/ide/diff/src/org/netbeans/modules/diff/builtin/provider/BuiltInDiffProvider.java)):

https://github.com/apache/netbeans/blob/ee2ceb2cbdad2ef5c05691c02116f31ea345a7ab/ide/diff/src/org/netbeans/modules/diff/builtin/provider/BuiltInDiffProvider.java#L69-L75

the `BuiltInDiffProvider.java` is made available via the layer file:

https://github.com/apache/netbeans/blob/ee2ceb2cbdad2ef5c05691c02116f31ea345a7ab/ide/diff/src/org/netbeans/modules/diff/mf-layer.xml#L69-L77

PR #3679, opened 2022-02-28
Remove old & deprecated openide.util.enumerations module

BradWalker:
The Enumerations module is really old, no longer used as it was replaced by an Enum class in JDK 1.5, and is not
used anywhere in the code.

Time to clean up the code base by removing this old module. That's what this code change does.

---


BradWalker:
This work is a little bit different than the usual rust removal that I do. So I'm really interested in feedback so that I can learn!

harness/nbjunit/test/unit/src/org/netbeans/testjunit/AskForOrgOpenideUtilEnumClass.java
```diff
@@ -1,50 +0,0 @@
...
-import org.netbeans.junit.*;
-import junit.framework.TestCase;
-
-/** This class needs to be in other package than
- * org.netbeans.junit in order to not be loaded by the application
- * classloader but by the started NetBeans IDE
- * 
- * @author Jaroslav Tulach
- */
-public class AskForOrgOpenideUtilEnumClass extends TestCase {
```
JaroslavTulach, line 32: Isn't this class referenced/used by some other class in the test suite?

JaroslavTulach (approved):
I am not fan of removing API only because it is no longer used by us. [APIs are like stars](http://wiki.apidesign.org/wiki/Star) - you never know who's watching.

However the [12.6](https://repo1.maven.org/maven2/org/netbeans/api/org-openide-util-enumerations/RELEASE126/) is already broken (empty) because of building on JDK-11. The `RELEASE130` JAR is also going to be empty. It makes little sense to keep distributing an empty JAR. We need a fix.

The last useful JAR comes from [version 12.5](https://search.maven.org/artifact/org.netbeans.api/org-openide-util-enumerations/RELEASE125/jar). The most compatible solution would be to turn the `openide.util.enumerations` module into a wrapper module with `binaries-list` referencing `org.netbeans.modules:org-openide-util-enumerations:12.5` and distribute that content with every new version of NetBeans.

Whether or not that makes sense is another question.

neilcsmith-net:
> I am not fan of removing API only because it is no longer used by us.

While I mostly agree with this (I do think deprecated for removal has a place), this seems somewhat different?  We're not removing because we no longer use it, but because we have to actively put in workarounds to keep it - well, bring it back.  By that point, in the very unlikely scenario someone needs this, surely _they_ can do the wrapping themselves?

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/SuiteCustomizerLibraries.java
```diff
@@ -780,7 +780,6 @@ public ExplorerManager getExplorerManager() {
         DISABLED_PLATFORM_MODULES.add("org.netbeans.modules.favorites"); // NOI18N
         // And these are deprecated:
         DISABLED_PLATFORM_MODULES.add("org.openide.compat"); // NOI18N
-        DISABLED_PLATFORM_MODULES.add("org.openide.util.enumerations"); // NOI18N
```
JaroslavTulach, line 783: The `apisupport` can work with older versions of the platform (that contain the  module). It may be wiser to keep this line.

harness/nbjunit/test/unit/src/org/netbeans/testjunit/AskForOrgOpenideUtilEnumClass.java
BradWalker: > Isn't this class referenced/used by some other class in the test suite?

Yep.. It is now fixed..

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/SuiteCustomizerLibraries.java
BradWalker: Agreed.. It is now fixed..

BradWalker:
Hey @JaroslavTulach , can you look over this once last time.. I'm really interested in your opinion since you originally created this back in the day.. 8-) Thanks.

BradWalker:
Hey @neilcsmith-net , since you looked at this.. Can you provide some comments about this.. Thanks!

JaroslavTulach:
FYI: #3744 has run to [org.openide.util.enum issue](https://app.travis-ci.com/github/apache/netbeans/jobs/562936571) as well. Clearly building with JDK-11 and testing with JDK-8 doesn't work. I should now support integrating this PR, otherwise #3744 needs to invent some other fix.

Let me attempt to behave pragmatically: Let's integrate!

mbien:
this line needs to be removed too since travis tries to run the removed test:
https://github.com/apache/netbeans/blob/754c2cd6b3e21acec851e41f994f30aa54e18277/.travis.yml#L184

as seen:
```
Buildfile: platform/openide.util.enumerations does not exist!

Build failed

The command "hide-logs.sh ant $OPTS -f platform/openide.util.enumerations test" exited with 1.
```

mbien (approved):
@BradWalker of course :)

PR #3689, opened 2022-03-03
project wide oprimization: improved map traversal

mbien:
some project wide, map/collection related optimizations (low hanging fruits).

 Now that `referencedIn` is fixed in NB 13, we might as well use the opportunity to improve the codebase a bit with the help of jackpot :)

I picked some rules from [this](https://github.com/mbien/jackpot-inspections/) hint collection.

first commit:
If the right iterator view is used (keySet/entrySet/values), the loop body does not have to call map.get(key) on every iteration.
```java
// avoid inefficient map traversal
for ($K $k : $map.keySet()) {
    $preamble$;
    $V $v = $map.get($k);
    $body$;
} :: $map instanceof java.util.Map && !referencedIn($k, $preamble$)
=>
for ($V $v : $map.values()) {
    $preamble$;
    $body$;
}
:: !referencedIn($k, $body$)
=>
for (java.util.Map.Entry<$K, $V> entry : $map.entrySet()) {
    $preamble$;
    $K $k = entry.getKey();
    $V $v = entry.getValue();
    $body$;
}
:: otherwise
;;
```

second commit - no need to use contains() before remove()
```java
// contains + remove in succession
if ($col.contains($v)) {
    $col.remove($v);
} :: $col instanceof java.util.Collection
=>
$col.remove($v);
;;

if ($map.containsKey($k)) {
    $map.remove($k);
} :: $map instanceof java.util.Map
=>
$map.remove($k);
;;
```

third commit (not project wide, only two selected places to make review easier)
This one is risky, so I only applied it selectively.
```java
// contains + get in succession (dangerous! values can be null)
if ($map.containsKey($key)) {
    $preamble$;
    $V $v = $map.get($key);
    $body$;
} :: $map instanceof java.util.Map && !referencedIn($key, $preamble$)
=>
$V $v = $map.get($key);
if ($v != null) {
    $preamble$;
    $body$;
}
;;
```

### DIFF
--- a/enterprise/glassfish.javaee/src/org/netbeans/modules/glassfish/javaee/RunTimeDDCatalog.java
+++ b/enterprise/glassfish.javaee/src/org/netbeans/modules/glassfish/javaee/RunTimeDDCatalog.java
@@ -365,8 +365,7 @@ public class RunTimeDDCatalog extends GrammarQueryManager implements CatalogRead
     public void removeCatalogListener(CatalogListener l) {
         if (null == l)
             return;
-        if (catalogListeners.contains(l))
-            catalogListeners.remove(l);
+        catalogListeners.remove(l);
     }
     
     public  void fireCatalogListeners() {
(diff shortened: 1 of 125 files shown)


BradWalker:
Nice! And I agree, looks good to me.

BradWalker (approved):
Nice! Looks good to me..

mbien:
thanks for the quick reviews Brad & Martin. I looked through it (again) here on gh and couldn't find any issues. 

Everything is green -> lets merge it before it causes conflicts.

Issue #3697, opened 2022-03-04
NetBeans 13+ macOS installer not notarized correctly


mcdonnell-john:
### Apache NetBeans version

Apache NetBeans 13 release candidate

### What happened

Due to a bug with the Apple's Notarization service, NetBeans always came back as Approved with a single warning.  Once you fix that invalid reference, the Notarization service now correctly shows the errors.

It appears as though we now need to sign embedded native libraries within Jars: https://stackoverflow.com/a/53528020/612072

### Did this work correctly in an earlier version?

Apache NetBeans 12.6

### Operating System

macOS

### JDK

11

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

Link to Apple Developer forum, where this is is under discussion: https://developer.apple.com/forums/thread/701438

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mcdonnell-john:
[curent-notarize-log.json.txt](https://github.com/apache/netbeans/files/8189273/curent-notarize-log.json.txt)
Current log from the notarization service

PR #3699, opened 2022-03-04
[3697] Removed invalid cpp package reference and introduces support for sign…

mcdonnell-john:
…ing the .app with an Application Cert

---

### DIFF
--- a/nbbuild/installer/mac/newbuild/build.sh
+++ b/nbbuild/installer/mac/newbuild/build.sh
@@ -42,12 +42,12 @@ buildnumber=$3
 build_jdk7=0
 build_jdk8=0
 build_jdk11=0
-binaryname=$5
-installer_sign_idname=$6
-application_sign_idname=$7
-version_number=$8
-if [ -n "${9}" ] ; then
-  nb_locales=",${9}"
+binaryname=$4
+installer_sign_idname=$5
+application_sign_idname=$6
+version_number=$7
+if [ -n "${8}" ] ; then
+  nb_locales=",${8}"
 fi
 
 basename=`dirname "$0"`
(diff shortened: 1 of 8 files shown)


mcdonnell-john:
Also related: https://github.com/apache/netbeans-jenkins-lib/pull/56

neilcsmith-net:
Thanks @mcdonnell-john Looks like my concerns about notarization issues are starting to realise! :smile: 

There's no need for the GitHub issue ( #3697 ) - PR fine - linking in so we have the log.

Do these changes get notarization to pass?  I'm surprised it's not erroring with other libs in jars - eg. JNA.  Also, including x86 libs and/or lack of hardened runtime being allowed through?  This seems like it still wouldn't meet the "requirements" at https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution ??

Base branch is master.  Can you build from NetBeans 13 release using master?  Or do we need to merge to delivery and then release branch?

mcdonnell-john:
Just seeing this now...

> Thanks @mcdonnell-john Looks like my concerns about notarization issues are starting to realise! 😄
> 

Yeah :)

> There's no need for the GitHub issue ( #3697 ) - PR fine - linking in so we have the log.
> 
I'll close the issue then, wasnt sure if I needed an issue to link to a PR or not.

> Do these changes get notarization to pass? I'm surprised it's not erroring with other libs in jars - eg. JNA. Also, including x86 libs and/or ~lack of hardened runtime being allowed through~? This seems like it still wouldn't meet the "requirements" at https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution ??
> 
Yes, they are passing notarization, however its worth pointing out the notarization logs to call out a warning on other .pkg files, so I might need to raise another ticket with Apple as the notarization log doesn't provide any feedback on what those warnings are about.

I'm not sure about the entitlements, I can take a further look before doing any release(just incase)

> Base branch is master. Can you build from NetBeans 13 release using master? Or do we need to merge to delivery and then release branch?
> 
If I go with the approach in the email thread I will keep this against master as I don't think we need to rebuild 13, I will generate the new distpreparations zip myself.

> EDIT: Sorry, just noticed you do have `--options runtime`, but does it need an entitlements file if we do this?
Nothing came up when I added this in, however based on what I've learned to date that means nothing 😄

PR #3711, opened 2022-03-05
[cnd] small 3.2.2 - Merging master into cnd

vieiro:
Bringing MultiMimeLanguageServer (and the rest of goodies) from master into cnd.


matthiasblaesing (approved):
Looks like a straight merge to me. I only saw a single change for the `ignored-overlaps` and that made sense to me. I restartet the failing Travis checks, but they looked like the "normally failing" bunch.

vieiro:
@matthiasblaesing , thanks for the review. 

We may want to do another merge  later on, I think we're using ssh somewhere in the remote stuff for cnd, we may want to do some tests with the new ssh library versions.

matthiasblaesing:
@vieiro I think you mean the jsch integration found in `dlight.nativeexecution`. My first thought is, that it looks safe. The corner case (agent integration) is not used by that and I did not see subclassing of the jsch Identity class.

Issue #3710, opened 2022-03-05
Editor Windows require twice the minimum height of the content (caused by BasicScrollingTabDisplayerUI)


George-Devel:
### Apache NetBeans version

Apache NetBeans 13

### What happened

 `public Dimension getMinimumSize(JComponent c)`

in

`org.netbeans.swing.tabcontrol.plaf.BasicScrollingTabDisplayerUI`

returns the minimum height of the selected content instead of the height of tabs (buttons in the north of the internally used border layout). This basically doubles the total minimum height. The height of the content is counted twice: once as center component and once as north component.

I created a test application to demonstrate the problem. 
[MinSizeTest.zip](https://github.com/apache/netbeans/files/8190728/MinSizeTest.zip)

### How to reproduce

Open the attached test application (or any Netbeans Application and set the Minimum Size in a window.) 

[MinSizeTest.zip](https://github.com/apache/netbeans/files/8190728/MinSizeTest.zip)

The window must be displayed as editor. (Otherwise the BasicScrollingTabDisplayerUI is not used. The Look and Feel must also use this class, but this is the case for most Look and Feels)

Open the output window und check the effective minimum size of the editor window by moving the divider upwards. (The green box is the minimum size, set in the window):

<img width="335" alt="image" src="https://user-images.githubusercontent.com/101003414/156885407-24ff7c35-6c5d-4e9d-9943-9c16b97dea33.png">

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10

### JDK

11

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

George-Devel:
I fixed this. I'll try to create a PR.

PR #3712, opened 2022-03-05
Using the size of the tabs as minimum size.

George-Devel:
See Netbeans issue #3710

### DIFF
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
@@ -468,7 +468,7 @@ public abstract class BasicScrollingTabDisplayerUI extends BasicTabDisplayerUI {
     
     @Override
     public Dimension getMinimumSize(JComponent c) {
-        return new Dimension( 100,  displayer.getSize().height);
+        return new Dimension( 100,  displayer.getPreferredSize().height);
     }
     
     /**
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
@@ -468,20 +468,7 @@ public abstract class BasicScrollingTabDisplayerUI extends BasicTabDisplayerUI {
     
     @Override
     public Dimension getMinimumSize(JComponent c) {
-        int index = displayer.getSelectionModel().getSelectedIndex();
-        TabDataModel model = displayer.getModel();
-        if (model.size() == 0) {
-            return new Dimension(0, 0);
-        }
-        if( index < 0 || index >= model.size() )
-            index = 0;
-        Component comp = model.getTab(index).getComponent();
-        if( null == comp )
-            comp = displayer.getComponentConverter().getComponent( model.getTab( index ) );
-        Dimension minSize = null == comp ? new Dimension( 100, 10 ) : comp.getMinimumSize();
-        minSize.width = Math.max(minSize.width, 100);
-        minSize.height = Math.max(minSize.height, 10);
-        return minSize;
+        return new Dimension( 100,  displayer.getSize().height);
     }
     
     /**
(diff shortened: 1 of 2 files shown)


George-Devel:
Hi, my code change may look weird, but this component is responsible for drawing the tab-bar in the editor area. This is only the part of the "North" component of a border layout. The minimum size of the actual content is the "Center" component and is considered by the panel.

You can check, that this works correctly by trying the Platform Application attached to the bug.
It contains only one relevant file: an editor window, registered with annotations, with a minimum size set and an overridden paint method to show the minimum size. If you like I can this file separately.

platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
```diff
@@ -468,20 +468,7 @@ protected Rectangle getControlButtonsRectangle( Container parent ) {
...
-        if( index < 0 || index >= model.size() )
-            index = 0;
-        Component comp = model.getTab(index).getComponent();
-        if( null == comp )
-            comp = displayer.getComponentConverter().getComponent( model.getTab( index ) );
-        Dimension minSize = null == comp ? new Dimension( 100, 10 ) : comp.getMinimumSize();
-        minSize.width = Math.max(minSize.width, 100);
-        minSize.height = Math.max(minSize.height, 10);
-        return minSize;
+        return new Dimension( 100,  displayer.getSize().height);
```
DevCharly, line 471: It would be better to use `displayer.getPreferredSize().height` here.
`displayer.getSize()` returns the current size, which is initially `0,0` before first layout pass.

DevCharly (changes requested):
I agree that `BasicScrollingTabDisplayerUI.getMinimumSize()` should return only the minimum size of the tabs.
Currently it returns the the minimum size of the tabs plus the minimum size of the tab content.
`getPreferredSize()` in `BasicScrollingTabDisplayerUI` subclasses also return only preferred size of the tabs.

For some unknown reason, this was added in 2008 in commit: https://github.com/emilianbold/netbeans-releases/commit/82a48e93252068b86905bcbfac1e7e508e6aa171#diff-2f04bb43144858669541be39da8de41c29eef37380e8a2b3ce26dce479c92534
This is the related issue: https://bz.apache.org/netbeans/show_bug.cgi?id=136636

Don't understand the need for this `BasicScrollingTabDisplayerUI.getMinimumSize()` implementation. Maybe some part of the windowing system worked different in 2008 that it does today...

There is a similar `getMinimumSize()` method in `AbstractViewTabDisplayerUI`, which was added in the same commit:
https://github.com/emilianbold/netbeans-releases/commit/82a48e93252068b86905bcbfac1e7e508e6aa171#diff-c2ceef123f96cb6d9bbb42f06afa4cdc636d72746b85139890a4774e6caddac2

Maybe this also needs to be changed? But maybe only if `isUseStretchingTabs()` returns `true`.

platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/BasicScrollingTabDisplayerUI.java
George-Devel: Ok, I will change this. I'll also have look at `AbstractViewTabDisplayerUI `. I see that this is used if "Multi-row tabs" is selected in the Appearance options . It seems that `AbstractViewTabDisplayerUI.getMinimumSize` returns height of a single tab even then multiple rows are displayed.

George-Devel:
@DevCharly - I changed the line as requested. I also replaced 'getMinimumSize()' in `AbstractViewTabDisplayerUI` with the same code. I checked the 'BasicScrollingTabDisplayerUI' in the Editor area, and  `AbstractViewTabDisplayerUI`  in the output area ,including the variant with "winsys.stretching_view_tabs" set to true.

My assumption that the `AbstractViewTabDisplayerUI` is used if "Multi-row tabs" was wrong.  This another independent issue.

DevCharly (approved):

eirikbakke (approved):
Thanks for reviewing, @DevCharly! The change sound fine with me... I'll test this in my IDE and NetBeans Platform application the next time I do a build, but no need to wait for that.

lkishalmi:
Well, it is approved. Let's merge it then!
Thank you @George-Devel for your contribution!

Issue #3700, opened 2022-03-04
Outdated Jcraft JSCH Library Causes Incorrect Credentials For Remote Git Repository


Ojambo:
### Apache NetBeans version

Apache NetBeans 12.6

### What happened

Originally reported on 2021-09-27 here https://issues.apache.org/jira/browse/NETBEANS-6067

I am running Fedora 35 and NetBeans 12.6 and just upgraded to 13.  I get incorrect credentials when I attempt a git push.  It works fine on the command line so I can up with a workaround until this issue is resolved.  A quick alternative for Netbeans is to use a fork of JSch-01.55 here [https://github.com/mwiede/jsch ](https://github.com/mwiede/jsch)that is actually updated allowing modern encryption algorithms.

**Workaround**:
Perform a commit using the GUI in NetBeans then on the command line run "git push" in the folder containing your project and/or local git repository.

**Problems**:
- JSch only supports weak ssh-rsa which is disabled by default in OpenSSH 8.8
- JSch has not been updated since 2018 according to SourceForce 
- Problematic key headers due to OpenSSH 7.8 or newer having  "----BEGIN OPENSSH PRIVATE KEY---" instead of "---BEGIN RSA PRIVATE KEY----"
- Replacing the official JSch JAR file with the one by mwiede did not work for me, looks like NetBeans hard-coded something in their version of EGit.

**References**:
1. Matez Blog the future of jsch without ssh-rsa https://www.matez.de/index.php/2020/06/22/the-future-of-jsch-without-ssh-rsa
2. Bug 520927: moving away from JSch https://www.eclipse.org/lists/jgit-dev/msg03696.html
3. Gerrit 3.4.0 Disables JCraft JSch client library by default https://www.gerritcodereview.com/3.4.html
4. JSch Last SourceForge Update https://sourceforge.net/projects/jsch/files/jsch/
5. Invalid Private Key When Using JSch https://stackoverflow.com/questions/53134212/invalid-privatekey-when-using-jsch

### How to reproduce

Git push will result in incorrect credentials in Netbeans if using recent versions of OpenSSH on the remote server or stronger encryption methods.

### Did this work correctly in an earlier version?

No

### Operating System

Locally Linux Fedora 35: Remotely Debian 11, and CentOS 7.9

### JDK

OpenJDK 11.0.14.1

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

This occurs for every GIT Push.
[...]

PR #3713, opened 2022-03-05
[NETBEANS-6067] Update JSch to 0.1.72 and junixsocket to 2.4.0

matthiasblaesing:
Closes: #3700


matthiasblaesing:
@sdedic you might want to have a look at this. I remember, that you asked to be included in changes, that might touch vscode. This is the case here, as vscode had a reference to the removed ssh agent integration module.

For the IDE this should work, as all relevant dependency modules (jna, jna-platform, bouncy castle and junixsocket) are either in the IDE or the Platform cluster, it might be necessary to adjust the vscode build to include them too.

ide/libs.git/src/org/netbeans/libs/git/jgit/JGitSshSessionFactory.java
```diff
@@ -167,9 +177,25 @@ protected Session createSession (Host hc, String user, String host, int port, FS
...
+            try {
+                 agentConnector = new SSHAgentConnector();
+                 if(! agentConnector.isAvailable()) {
+                     agentConnector = null;
+                 }
+            } catch (Throwable ex) {
+                LOG.log(Level.FINE, null, ex);
+            }
+            try {
+                agentConnector = new PageantConnector();
```
vieiro, line 190: Why create a new "PageantConnector()" on line 190 if we've already created a SSHAgentConnector on line 182 above?

matthiasblaesing: There are two possible outcomes of the creation of the SSHAgentConnector:

- instantiation of `SSHAgentConnector` could fail because `junixsocket` can't be instatiated
- the class might be loadable, but there is no agent running

On window this is quite possible. In my work setup I have pagent running and an ssh-agent -> pagent bridge in cygwin, but NetBeans uses PAgent directly.

norrisjeremy:
I am one of the primary caretakers of this JSch fork and want to point out that `0.2.0` is the most recent release.

norrisjeremy:
Also, if you have any questions, please let me know!

matthiasblaesing:
> I am one of the primary caretakers of this JSch fork and want to point out that `0.2.0` is the most recent release.

I noticed that, but there already breaking changes and 0.2.0 explicitly states, that algorithms are disabled. If someone makes a case for 0.2.0 I can bump it.

ide/libs.git/src/org/netbeans/libs/git/jgit/JGitSshSessionFactory.java
vieiro: What I was meaning is that you may want an `if (agentconnector != null) {}` wrapping the second try-catch, or not?

ebresie:
Are there any existing unit tests for this?  Or maybe add a new one?

matthiasblaesing:
@ebresie wrong question. The right question is: What would you test? Jsch is basicly just the original library with OSGI headers (specific for netbeans). Same is junixsocket. The rest are integrations that used now broken code paths.

matthiasblaesing:
I pushed a different implementation a few minutes ago. Sorry for the complete replacement, but the differences would have touch many files anyway:

- @vieiro rightly pointed out, that in the case, that if Pagent is used, the SSHAgent instatiation becomes useless. The updated implementation prefers the SSHAgent. If that is present, it is directly returned. That might change the situation for windows users, when both a Pagent _and_ an SSHAgent is present. However in this case I'd expect the SSHAgent to act as a proxy to the Pagent. So in sane setups nothing should change.
- the dependencies for the OSGI module were also declared on the JSch wrapper module. It is unclear whether or not this is necessary, but it can't hurt.
- Specification versions were bumped for junixsocket, libs.git and php.project
- The removal of the libs.jsch.agentproxy module was reverted and instead the factory method for the AgentConnector was updated to the new JSch API (Release and specification version were bumped accordingly)

An updated test build can be found here: http://doppel-helix.eu/NetBeans-dev-dev-5573b9c9e8794ba43374347606509a5a4206a00a-release.zip

ebresie:
> @ebresie wrong question. The right question is: What would you test? Jsch is basicly just the original library with OSGI headers (specific for netbeans). Same is junixsocket. The rest are integrations that used now broken code paths.

Why the authentication failed given good user and password I'm not clear of but...so unit test the "now broken code path".  Test (with existing or new test) a git fetch via ssh connection type that fails to authentication (and triggers the behavior described) to reproduces the problem before the fix and works after the fix.

matthiasblaesing:
> > @ebresie wrong question. The right question is: What would you test? Jsch is basicly just the original library with OSGI headers (specific for netbeans). Same is junixsocket. The rest are integrations that used now broken code paths.
> 
> Why the authentication failed given good user and password I'm not clear of but...

The authentication failed using an SSH key, that was generated using a new enough `ssh-keygen` version. The reason is, that that the format changed and also new algorithms were introduced. Both not supported by the JSch version we bundled. JSch in its original version was not updated and a fork was created: https://github.com/mwiede/jsch, to which this PR switches.

> so unit test the "now broken code path". Test (with existing or new test) a git fetch via ssh connection type that fails to authentication (and triggers the behavior described) to reproduces the problem before the fix and works after the fix.

Thank you - I know what a test is and how it works. This integration test (it tests interdependent behaviour) would need to setup an SSH server, a git repository and the right keys and then check what you describe. But this would most probably not have caught this issue, as I would have used pre generated keys and this would continue to work. Even freshly generated keys need to have the algorithm selected, which would again yield compatible keys. What is more, you are talking about serious work here, the external SSH server must work on all architectures tests are run and needs to be reliable. We have enough such complex tests, that only cause headaches and lead to people to discarding problems reported by them.

Feel free to create a test, I won't.

matthiasblaesing:
I intent to merge this on the weekend.

matthiasblaesing:
LSP test is know to fail.

PR #3715, opened 2022-03-06
[NETBEANS-5829] Use --release instead of downgrading the source level.

jtulach:
Downgrading source level of a project when the right classes aren't on bootclasspath is a major NetBeans IDE problem. Now, when it is 99% guaranteed nb-javac is available, there is no reason for doing so. Should the bootclasspath be incorrect, let's substitute it with `ct.sym` - e.g. let's use `--release` tag.

This PR may not cover all corner cases, but it does its job. One can execute the IDE on JDK8 and use `var` or text blocks without any issues. the [NETBEANS-5829 bug](https://issues.apache.org/jira/browse/NETBEANS-5829) is fixed.

Please, @jlahoda & @dbalek, guide me in further improving this PR and making production ready! Thank you.


jlahoda:
I think the end goal is good, but I wonder if the solution is at the right level. I wonder if projects should continue to be able to precisely specify the bootclasspath and/or system module path. E.g.:
-remove the source level validation
-provide a method somewhere in Java project APIs like:
ClassPath validateBoot(ClassPath origin, String release)
which would do the source level validation and would return a sensible CP for given release. Projects would then use this method.

I am however a bit concerned with debugger, whether step into will work for projects that have fallback bootclasspath.

jtulach:
Thinking about it I have to admit I see no reason for anything more complicated than what is in this PR. Clearly the NetBeans 13 state is wrong and this PR improves it:

> projects should continue to be able to precisely specify the bootclasspath and/or system module path. 

Projects can certainly specify that. However, if they fail to properly match the source/target level with bootclasspath, then instead of

> the source level validation

...and subsequent downgrading of source/target level, the `java.source` subsystem steps in and supplies bootclasspath appropriate to the source/target level by adding `--release` flag.

> I am however a bit concerned with debugger, whether step into will work for projects that have fallback bootclasspath.

No problems with debugger will occur, as the change is going to be done only when invoking `javac`. Behavior of other systems remains unchanged.

JaroslavTulach:
The changes in this PR allow me to edit [ScriptEngineTest](https://github.com/apache/netbeans-html4j/blob/b7bbe85ec63333990ff50d0a59912ea18b562e28/boot-script/src/test/java/net/java/html/boot/script/ScriptEngineTest.java#L97) with NetBeans running on JDK-8 without any errors:

obrazek

That's certainly better experience than editing the same file with NetBeans 13:

obrazek

JaroslavTulach:
The bug [NETBEANS-5829](https://issues.apache.org/jira/browse/NETBEANS-5829) is real. This PR fixes it. I understand that @jlahoda doesn't like my fix (observation based on rumors), but I am not going to wait any longer - if anyone has a problem with this PR, speak up publicly now.

jlahoda:
I think I've publicly stated my opinion. I understand that what I am proposing is potentially much more work, but there's a chance it would work without side-effects/further bugs. I suspect this patch will introduce two bugs:
1. The editor will not report errors on usage of some APIs, even of the build will consequently fail. This will apply to HTML4j, I believe.
2. The Java index won't have information about the bootclasspath, so that e.g. adding imports may not see all the required classes.

So, while I understand this fixes one problem you are seeing, I suspect it will introduce new ones. And this patch won't help with fixing these, i think - it will need to be deleted and redone from scratch.

JaroslavTulach:
Failing test written in [ff6fb43](https://github.com/apache/netbeans/pull/3715/commits/ff6fb4314093f4bf10eedb74bb17004acd311efc). It passes since [75cc6dc](https://github.com/apache/netbeans/pull/3715/commits/75cc6dc858efe69bfdf5e6ff489e3fe467b65e9e) which merges the test commit with the fix introduced in https://github.com/apache/netbeans/pull/3715/commits/fc1e300c643ae864cabfaf50ae0233a80559bd1b - while @jlahoda may not agree with my fix, it shall now be undeniable that this PR fixes _something_ and does that with solid engineering practice. 

Will somebody find enough courage to approve or are we going to release another NetBeans/VSCode version with the [NETBEANS-5829 bug](https://issues.apache.org/jira/browse/NETBEANS-5829) present?

jlahoda:
Well, I've never said this does not fix something. I just said it also will likely break something (and I have one more below). Anyway, if all you want to fix is [NETBEANS-5829](https://issues.apache.org/jira/browse/NETBEANS-5829) and make sure the test passes, I suspect simply not doing the downgrade would do the same, with hopefully a smaller fallout. If there's some other usecase you intend to fix, you should have a test for that as well, otherwise it may regress at any time.

There may be another problem I didn't realize before - when the source level selected is newer than the highest source level supported by the given compiler, it seems it may crash with an exception, possibly rendering the IDE useless for some users (me possibly included, depending on the circumstances).

Anyway - you've asked for a review/opinion, I provided one. I can't stop you from doing this, but it is not clear to me who will be responsible for the fallout.

JaroslavTulach:
> I suspect simply not doing the downgrade would do ...

Yes, you keep talking about not performing the downgrade for years. It just needs someone to try it. Is [this what you mean](https://github.com/apache/netbeans/compare/master...JaroslavTulach:jtulach/JustDisableSourceLevelDowngrading)? If so, it doesn't work. Neither for the new `TextBlocksOnAnyJavaTest`, neither for existing tests which would have to be modified somehow.

> who will be responsible

`TextBlocksOnAnyJavaTest` deserves to be fixed.  Once it is integrated, anyone doing changes into the IDE compiler will be responsible for not breaking it.

> another problem ... may crash

If you lead me I can write a test to simulate such crash. Then it can be fixed. Without a reproducer such a _maybe_ doubt is not helping us to move forward.

JaroslavTulach (approved):

JaroslavTulach:
Commit 854b196 is going to be verified by [job 11759](https://app.travis-ci.com/github/apache/netbeans/builds/248424613).

MartinBalin (approved):

jlahoda:
> > I suspect simply not doing the downgrade would do ...
> 
> Yes, you keep talking about not performing the downgrade for years. It just needs someone to try it. Is [this what you mean](https://github.com/apache/netbeans/compare/master...JaroslavTulach:jtulach/JustDisableSourceLevelDowngrading)? If so, it doesn't work. Neither for the new `TextBlocksOnAnyJavaTest`, neither for existing tests which would have to be modified somehow.

Not quite. I meant simply running as if DISABLE_SOURCE_LEVEL_DOWNGRADE was set to true. But right, for source > 8, a modular boot path is needed, and that is not readily available from JDK 8. So you are right it is not as simple as long as we need to support 8.

> 
> > who will be responsible
> 
> `TextBlocksOnAnyJavaTest` deserves to be fixed.  Once it is integrated, anyone doing changes into the IDE compiler will be responsible for not breaking it.

I meant new bugs introduced by this patch. For example, with your patch, when I write `of(1, 2, 3,4)` into AsyncJavaScriptAction.storeLater, there will be no fix to add static import java.util.List.of. But if I add the import manually, there is no error in the editor. And build will fail. So there's apparently some bug somewhere in your code. At least no error in the editor, but an error during build. So if it is not fixed here, then it must be a bug in HTML4j and will be fixed there, right?

> 
> > another problem ... may crash
> 
> If you lead me I can write a test to simulate such crash. Then it can be fixed. Without a reproducer such a _maybe_ doubt is not helping us to move forward.

I see no issue with reviewers expressing concerns, and PR author evaluating those concerns. And yes, may turn out the concerns are not valid.

Anyway, in this case, set the source level in the test to some very high level '9999' for example. This crashes the test for me with a NPE at JavacParser:963.

java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
```diff
@@ -1078,6 +1086,23 @@ private static JavacTaskImpl createJavacTask(
         return task;
     }
 
+    private static String useRelease(final String requestedSource, com.sun.tools.javac.code.Source validatedSourceLevel) {
+        if (requestedSource == null || validatedSourceLevel == null) {
+            return null;
+        }
+        com.sun.tools.javac.code.Source sourceLevel = com.sun.tools.javac.code.Source.lookup(requestedSource);
+        if (sourceLevel == null) {
```
jtulach, line 1094: Working with OpenJDK sources out of the box is an important use-case for NetBeans. By creating the `useRelease` method and adding this `sourceLevel == null` check I am able to open `java.base` from https://github.com/openjdk/jdk/commit/c2c0cb2a4372d78658326461562363de9a1a194f without major errors.

jtulach: These are the non-fatal errors I see on console:

```
org.netbeans.modules.java.editor.base.semantic.UnusedDetector.isUnusedInPkg(UnusedDetector.java:322)
org.netbeans.modules.java.editor.base.semantic.UnusedDetector.findUnused(UnusedDetector.java:138)
org.netbeans.modules.java.editor.base.semantic.SemanticHighlighterBase.process(SemanticHighlighterBase.java:185)
```
```
org.netbeans.modules.java.source.indexing.VanillaCompileWorker$3.visitVariable(VanillaCompileWorker.java:585)
org.netbeans.modules.java.source.indexing.VanillaCompileWorker$3.visitVariable(VanillaCompileWorker.java:535)
com.sun.tools.javac.tree.JCTree$JCVariableDecl.accept(JCTree.java:1045)
com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:86)
org.netbeans.modules.java.source.indexing.VanillaCompileWorker$3.scan(VanillaCompileWorker.java:886)
```

```
WARNING [org.netbeans.modules.java.source.usages.BinaryAnalyser]: Invalid class file format: file:/home/devel/NetBeansProjects/netbeans/nbbuild/netbeans/java/modules/ext/fakeJdkClasses.zip!/java/lang/AutoCloseable.class
INFO [org.netbeans.modules.java.source.usages.BinaryAnalyser]: Class File Exception Details
java.io.EOFException
        at java.base/java.io.DataInputStream.readInt(DataInputStream.java:397)
        at org.netbeans.modules.classfile.ClassFile.loadClassHeader(ClassFile.java:183)
        at org.netbeans.modules.classfile.ClassFile.load(ClassFile.java:168)
Caused: org.netbeans.modules.classfile.InvalidClassFormatException
        at org.netbeans.modules.classfile.ClassFile.load(ClassFile.java:174)
        at org.netbeans.modules.classfile.ClassFile.<init>(ClassFile.java:126)
        at org.netbeans.modules.classfile.ClassFile.<init>(ClassFile.java:73)
        at org.netbeans.modules.java.source.usages.BinaryAnalyser.analyse(BinaryAnalyser.java:670)
```

jtulach:
> set the source level in the test to some very high level '9999' ... crashes the test ... with a NPE at JavacParser:963.

Or source level `19`, just like OpenJDK, right? Addressed in https://github.com/apache/netbeans/pull/3715/commits/fdea3e515c13e392c540ac3569c733982179a9cc - returning `null`, if the `Source` level cannot be looked up. The goal for `useRelease` method is to be conservative and only offer _use of --release_ when it has a chance to be useful - I'll do my best to reach such state - e.g. to fix [NETBEANS-5829](https://issues.apache.org/jira/browse/NETBEANS-5829), but not break other important use-cases.

JaroslavTulach:
If you find some issue with this change, please report it to me. Let's evaluate it in the context of [NETBEANS-5829](https://issues.apache.org/jira/browse/NETBEANS-5829) and address it the best I can. Thank you for your patience.

PR #3721, opened 2022-03-07
Do not search unnecessarily deep.

sdedic:
The Windows Github workflow occasionaly fails on timeout, and as @JaroslavTulach  experiments indicated, the path expression `**/external/binaries-list` is too broad and it the filesystem search seems to take ages on Windows, Bill knows why.

The search can be narrowed as we do not have `binaries-list` on 3rd and lower levels (now). I made the change to all workflows, so the computed hash allows to share the binaries.


JaroslavTulach (approved):
Let's try it. Less random failures is better.

PR #3722, opened 2022-03-07
SQL completion for Query annotations added.

dbalek:
SQL completion for Micronaut Query annotations added in addition to the existing JPQL completion.


sdedic (approved):
Looks fine.

PR #3724, opened 2022-03-07
project wide refactoring: avoid String.replaceAll() if possible.

mbien:
project wide micro optimization (low hanging fruits)

`String.replaceAll()` is expecting regexps as arguments, while `String.replace()` can be used for simple substitutions. Due to its sub-optimal name `replaceAll` is sometimes unintentionally used even without any regexps involved (as seen in this PR :)).

`String.replace()`
 - is cleaner / doesn't require additional character escaping
 - and safer / less error prone
 - is 10-20x faster /  produces less garbage
 
this might fix some bugs too, for example `localizedBundlepath.replaceAll(".properties", ...)` didn't escape the '.' which could potentially cause issues with certain filenames.

used [jackpot rules](https://github.com/mbien/jackpot-inspections/blob/0832be385a6512eceebeef7651423d71c5315b90/ProbableBugs.hint#L25-L30) for the simple cases and to find candidates, adjustments were made manually.

I looked through it at least twice by now + ran some manual tests with a dev build.

split into 6 commits to make review easier, planning to squash after approval.

### DIFF
--- a/apisupport/apisupport.project/src/org/netbeans/modules/apisupport/project/spi/BrandingModel.java
+++ b/apisupport/apisupport.project/src/org/netbeans/modules/apisupport/project/spi/BrandingModel.java
@@ -719,7 +719,7 @@ public abstract class BrandingModel {
         for (BundleKey bundleKey : internationalizedResourceBundleKeys) {
             String localizedBundlepath = bundlepath;
             if(!localizedBundlepath.endsWith("_" + this.locale.toString() + ".properties"))
-                localizedBundlepath = bundlepath.replaceAll(".properties", "_" + this.locale.toString() + ".properties");
+                localizedBundlepath = bundlepath.replace(".properties", "_" + this.locale.toString() + ".properties");
             if (key.equals(bundleKey.getKey()) &&
                 backslashesToSlashes(bundleKey.getBundleFilePath()).endsWith(localizedBundlepath) &&
                 codenamebase.equals(bundleKey.getModuleEntry().getCodeNameBase()))
@@ -736,7 +736,7 @@ public abstract class BrandingModel {
         Set<BrandingSupport.BundleKey> brandedBundleKeys = branding.getBrandedBundleKeys();
         String localizedBundlepath = key.getBundleFilePath();
         if(!localizedBundlepath.endsWith("_" + this.locale.toString() + ".properties"))
-                localizedBundlepath = localizedBundlepath.replaceAll(".properties", "_" + this.locale.toString() + ".properties");
+                localizedBundlepath = localizedBundlepath.replace(".properties", "_" + this.locale.toString() + ".properties");
         File bundleFile = new File(localizedBundlepath);
         if(!bundleFile.exists()) {
             for(BrandingSupport.BundleKey keyIter:brandedBundleKeys) {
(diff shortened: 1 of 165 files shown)


apisupport/apisupport.project/src/org/netbeans/modules/apisupport/project/spi/BrandingModel.java
```diff
@@ -719,7 +719,7 @@ private String backslashesToSlashes (String text) {
         for (BundleKey bundleKey : internationalizedResourceBundleKeys) {
             String localizedBundlepath = bundlepath;
             if(!localizedBundlepath.endsWith("_" + this.locale.toString() + ".properties"))
-                localizedBundlepath = bundlepath.replaceAll(".properties", "_" + this.locale.toString() + ".properties");
```
mbien, line 722: dot in ".properties" should have been escaped: "\.properties" (multiple occurrences in this class)

java/performance/src/org/netbeans/modules/performance/guitracker/ActionTracker.java
```diff
@@ -758,11 +758,11 @@ public void exportAsXML(Document style, PrintStream out)
     }
 
     private static String getShortenName(String name) {
-        name = name.replaceAll("javax.swing", "j");
-        name = name.replaceAll("org.netbeans.modules", "o.n.m");
-        name = name.replaceAll("org.netbeans", "o.n");
-        name = name.replaceAll("org.openide.awt", "o.o.a");
-        name = name.replaceAll("org.openide", "o.o");
```
mbien, line 765: same here. "." not escaped

ide/editor.search/src/org/netbeans/modules/editor/search/SearchComboBoxEditor.java
```diff
@@ -176,14 +176,14 @@ public static void changeToOneLineEditorPane(JEditorPane editorPane) {
                     @Override
                     public void insertString(DocumentFilter.FilterBypass fb, int offset, String string, AttributeSet attr) throws BadLocationException {
                         if (string != null) {
-                            fb.insertString(offset, string.replaceAll("\\t", "").replaceAll("\\n", ""), attr); //NOI18N
```
mbien, line 179: i don't think you have to escape `\n` or `\t` but it doesn't seem to hurt.
```java
        System.out.println("a\nb\nc".replaceAll("\\n", ""));
        System.out.println("a\nb\nc".replaceAll("\n", ""));
        System.out.println("a\nb\nc".replace("\n", ""));
```
give equal results

matthiasblaesing (approved):
Looks sane to me.

mklaehn (approved):
Looks good to me.

One thing to note: `replace(String, String)` in case of both Strings being of length one could instead use `replace(char, char)`.

mbien:
> Looks good to me.

thanks for reviewing!
> 
> One thing to note: `replace(String, String)` in case of both Strings being of length one could instead use `replace(char, char)`.

This used to be a good advice, but this method is intensified and has the same performance as the char variant on modern JVMs. The only String/char based method left which still behaves differently performance wise is `indexOf`. So its more of a matter of preference now - no longer a performance optimization.

sidenote: I actually tried to add a fast path for `indexOf` https://github.com/openjdk/jdk/pull/6509 but it turned out that under certain circumstances (String lengths) the String variant is actually faster than the char variant due to different vectorization heuristics (this shouldn't happen). This would have to be fixed first otherwise it is not possible to make assumptions of which path is actually faster.

mbien:
squashed everything to one commit. going to merge tomorrow if everything is still green.

PR #3735, opened 2022-03-09
Support for Call Hierarchy for LSP

sdedic:
This PR adds support for `textDocument/prepareCallHierarchy`, `callHierarchy/incomingCalls` and `callHierarchy/outgoingCalls` messages (see [LSP Protocol Spec 3.16](https://microsoft.github.io/language-server-protocol/specification#textDocument_prepareCallHierarchy) ).

I decided to **reuse** `StructureElement` invented as a part of Outline support: elements reported by the Call Hierarchy have exactly the same meaning, plus the `uri`, which is implicit in the Outline. I have added `file` property to `StructureElement` and relevant `StructureProvider.Builder` method.

In order to share code between Java Editor module (that implements Outline) and Java Refactoring module (which implements the Call Hierarchy, but does not export it through any API), I've introduced helper methods for LSP into `java.sourceui` module: `lsp.api` module represents an abstract UI, after all, so adapting Java-specific structures to abstract ones seemed reasonable. This API allows to move Java Element -> StructureElement conversion to `java.sourceui`, and can be called from both `java.editor` and `java.refactoring`.

The other reason to choose `java.sourceui` for introducing the API is that the `ElementOpen` implementation is here, and can be adjusted to support StructureElement creation without exposing additional APIs.

### DIFF
--- a/ide/api.lsp/src/org/netbeans/api/lsp/CallHierarchyEntry.java
+++ b/ide/api.lsp/src/org/netbeans/api/lsp/CallHierarchyEntry.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.api.lsp;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.netbeans.api.annotations.common.CheckForNull;
+import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.spi.lsp.CallHierarchyProvider;
+
+/**
+ * Represents an entry in a call hierarchy chain. The entry identifies an element in a source which is the origin or target
+ * of an invocation/call. 
+ * 
+ * @since 1.9
+ * @author sdedic
+ */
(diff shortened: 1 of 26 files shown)


java/refactoring.java/nbproject/project.xml
```diff
@@ -52,6 +52,15 @@
                         <specification-version>1.0</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.api.lsp</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>1</release-version>
+                        <specification-version>1.8</specification-version>
```
ppisl, line 61: The new version of LSP API is 1.9. Version 1.8 introduced StructureElement, which is changed in this new version 1.9.

java/java.sourceui/src/org/netbeans/modules/java/source/ui/LspElementUtils.java
```diff
@@ -0,0 +1,335 @@
...
+        StructureProvider.Builder builder = StructureProvider.newBuilder(createName(info, el), ElementHeaders.javaKind2Structure(el));
+        builder.detail(createDetail(info, el));
+        setOffsets(info, el, builder);
+        if (info.getElements().isDeprecated(el)) {
+            builder.addTag(StructureElement.Tag.Deprecated);
+        }
+
+        if (childAcceptor != null) {
+            for (Element child : el.getEnclosedElements()) {
+                TreePath p = info.getTrees().getPath(el);
```
ppisl, line 82: `el` is not changed in this `for` cycle, so why we get `TreePath` in every cycle? Probably it should be `child` the param of `getPath()`?

ppisl (approved):

ide/api.lsp/src/org/netbeans/api/lsp/CallHierarchyEntry.java
```diff
@@ -0,0 +1,138 @@
...
+        public List<CallRange> getRanges() {
+            return ranges;
+        }
+    }
+    
+    /**
+     * Represents a textual range. Range covers text from `startOffset' to `endOffset' (exclusive).
+     */
+    public static final class CallRange {
+        private final int startOffset;
```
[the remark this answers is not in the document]
sdedic: OK, will introduce a toplevel `Range` ... plus I guess I could add `getSelectionRange` and `getExtendedRange` to the `StructureElement` - right ?

dbalek (approved):

sdedic:
Sorry, errors found during testing...

Issue #3730, opened 2022-03-08
Add simple code completion for Spock test framework


ppisl:
### Description

The Spock test framework is widely used framework for creating test in Groovy

### Use case/motivation

It would be useful to offer Spock block names and Spock test method parameters in cc. 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow the Apache Software Foundation's [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html)

neilcsmith-net:
No need to open an issue if you have a PR!

PR #3736, opened 2022-03-09
[NETBEANS-3730]  Add simple code completion for Spock test framework

ppisl:
This is implemented issue #3730.

Added basic code completion for Spock test framework:
  *  Spock block names are offered inside methods if the class extends the Spock specification.
  *  Parameter names are offered if the parameters are defined in a method name that is annotated with @Unroll


mbien:
hi @ppisl, you can link to issues on the right side under "Development", this will close them automatically post integration ->

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockBlockNamesCompletion.java
```diff
@@ -0,0 +1,103 @@
...
+            return getName() + ": ";
+        }
+        
+    }
+    
+    @Override
+    public boolean complete(Map<Object, CompletionProposal> proposals, CompletionContext request, int anchor) {
+        if (request.location == CaretLocation.INSIDE_METHOD 
+                && SpockUtils.isInSpecificationClass(request)) {
+            for (SpockBlock block: SpockBlock.values()) {
```
sdedic, line 93: Is the context sufficiently narrow - will these compl items NOT be offered if  i.e.  inside an expression like`obj.wh|` ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockMethodParamCompletion.java
```diff
@@ -0,0 +1,149 @@
...
+                    methodNode = (MethodNode) next;
+                    break;
+                }    
+            }
+            String name = null;
+            if (methodNode != null) {
+                name = methodNode.getName();
+            }
+            
+            if (name != null && name.contains("#")) {   //NOI18N
```
sdedic, line 123: Q: it seems that syntax 
```
def "#name should have length #length"() {
```
is supported, but the syntax
```
 @Unroll("#name should have length #length")
 def "name length"() {
```
is not ?

I can't see a check for `@Unroll` annotation either

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockMethodParamCompletion.java
```diff
@@ -0,0 +1,149 @@
...
+                String[] parts = name.split("#");       //NOI18N
+                
+                for (int i = 1; i < parts.length; i++) {
+                    String part = parts[i].trim();
+                    if (!part.isEmpty()) {
```
sdedic, line 128: Since the `name` is already matched by a regex in split() ... wouldn't it be faster to iterate using `Pattern.compile("#\\p{Alpha}[\\p{Alnum}$_]*)[. ]?*).matcher(name)` ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockUtils.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Petr Pisl
+ */
+public class SpockUtils {
+    
+    public static boolean isInSpecificationClass(CompletionContext context) {
+        
+        Set<String> visited = new HashSet<>();
+        String name;
+        ClassNode classNode = context.declaringClass;
```
sdedic, line 41: Idea: If there was a `GroovyParserResut` available from the `CompletionContext`, one could do 
```
context.declaringClass.isDerivedFrom(
   parserResult.resolveClassName("spock.lang.Specification")
)
```

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockUtils.java
```diff
@@ -0,0 +1,73 @@
...
+            visited.add(name);
+            classNode = classNode.getSuperClass();
+        }
+        return false;
+    }
+    
+    static boolean isFirstStatement(final CompletionContext request) {
+        TokenSequence<GroovyTokenId> ts = LexUtilities.getGroovyTokenSequence(request.doc, 1);
+
+        if (ts != null) {
```
sdedic, line 55: Same method in `KeywordCompletion`, consider reusing this utility there.

sdedic (changes requested):

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockMethodParamCompletion.java
ppisl: You are right, the params in the name method is newer and in parameter of @Unroll is obsolete. But we should support for both cases. Can we treat it as a bug?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockUtils.java
ppisl: Good idea.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockMethodParamCompletion.java
ppisl: I have implemented `@Unroll` with value. So this is resolved now.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockUtils.java
ppisl: Done

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockUtils.java
ppisl: Done

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/SpockBlockNamesCompletion.java
ppisl: No, because it's called only from complete keywords context, which means that it has to be the first token after a whitespace.

sdedic (approved):

PR #3742, opened 2022-03-09
Create JavaParser dumpfile on FINE log level only.

lkishalmi:
This is a trivial PR addressing the issue on #3728. Which puts the dumpfile creation under FINE loglevel.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/JavacParser.java
@@ -800,7 +800,9 @@ public class JavacParser extends Parser {
                     invalidate(false);
                 } else {
                     parserError = currentPhase;
-                    dumpSource(currentInfo, ex);
+                    if (LOGGER.isLoggable(Level.FINE)) {
+                        dumpSource(currentInfo, ex);
+                    }
                     throw ex;
                 }
             }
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
The change in `VanillaPartialReparser` is unnessary, as the code path can only be hit if assertions are enabled (with assertions disabled (i.e. a non dev build), 332 is not hit and a stays `false`):

https://github.com/apache/netbeans/blob/22368c42d1041f7b536c1cac576f4cd010a1a672/java/java.source.base/src/org/netbeans/modules/java/source/parsing/VanillaPartialReparser.java#L331-L333

IMHO then it is correct, that the problem is reported.

I don't think these changes address the problem reported. The reported dump files are created by `VerifyPartialReparse`:

https://github.com/apache/netbeans/blob/22368c42d1041f7b536c1cac576f4cd010a1a672/java/java.source.base/src/org/netbeans/modules/java/source/parsing/VanillaPartialReparser.java#L512-L558

line 534 creates the diff between the trees. As dumping is already partially protected it might be an option to tie the activation of `VerifyPartialReparse` to assertions:

https://github.com/apache/netbeans/blob/22368c42d1041f7b536c1cac576f4cd010a1a672/java/java.source.base/src/org/netbeans/modules/java/source/parsing/VanillaPartialReparser.java#L606-L614

```java
 @MimeRegistration(service=TaskFactory.class, mimeType="text/x-java") 
 public static final class FactoryImpl extends TaskFactory { 
  
     @Override 
     public Collection<? extends SchedulerTask> create(Snapshot snapshot) { 
         boolean  enableVerifier = false; 
         assert enableVerifier = true; 
         if (enableVerifier) { 
             return Collections.singletonList(new VerifyPartialReparse()); 
         }
     } 
  
 } 
```

@jlahoda what do you think about this?

jlahoda (approved):
Ok, more or less as a stop gap measure.

PR #3747, opened 2022-03-10
[cnd] small 3.2.3 LSP server support in cnd

vieiro:
- Drops LSP server support (and options, dataobjects) in cpplite.
- Adds LSP server support to cnd (clangd).

Currently a single clangd process is serving all opened projects.

cpplite projects will be recovered later on in step 3.5 in the proposed plan https://cwiki.apache.org/confluence/display/NETBEANS/cnd+integration+proposal%3A+cnd-small 

We now have code completion and hints in cnd based projects.


vieiro:
Delayed to the weekend, if rest permits :-).

PR #3754, opened 2022-03-10
increase git clone depth so that we can restart older jobs post merge.

mbien:
travis fails to restart old jobs of already integrated commits, for example:
```
fatal: reference is not a tree: 49e167edc46e335e1587ba22749650c5617f9954

The command "git checkout -qf 49e167edc46e335e1587ba22749650c5617f9954" failed and exited with 128 during .
```
setting `--depth` to 50 should improve the situation a bit (I hope) without having to clone the entire repo.

I also added some java version info printout to ant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ cache:
 
 git:
   quiet: true
-  depth: 3
+  depth: 50
 
 matrix:
     include:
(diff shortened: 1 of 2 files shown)


matthiasblaesing (approved):
Fine with me.

PR #3764, opened 2022-03-11
fix enterprise cluster build on JDK 17 + add upstream-JDK build job.

mbien:
it turns out that the enterprise cluster was the last cluster with build problems on JDK 17. Java cluster was fixed back in #3278 and gradle #3326. 

@matthiasblaesing must have fixed the javascript cluster without telling anyone :)

everything builds locally, lets see if CI agrees.

this fixes the following build error:
```
 [repeat] Note: generating beans in org.netbeans.modules.j2ee.deployment.impl.gen.nbd
   [repeat] Note: parsing: file:///mnt/ram/netbeans/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/plugins/api/netbeans-deployment.dtd
   [repeat] /mnt/ram/netbeans/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/impl/gen/nbd/package-info.java:28: error: Failed to process
   [repeat] package org.netbeans.modules.j2ee.deployment.impl.gen.nbd;
   [repeat] ^
   [repeat] error: java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.util.HashMap$Node[] java.util.HashMap.table accessible: module java.base does not "opens java.util" to unnamed module @392c130a
   [repeat]   	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
   [repeat]   	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
   [repeat]   	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:178)
   [repeat]   	at java.base/java.lang.reflect.Field.setAccessible(Field.java:172)
   [repeat]   	at org.netbeans.modules.schema2beansdev.gen.JavaUtil.getOptimialHashMapSize(JavaUtil.java:855)
   [repeat]   	at org.netbeans.modules.schema2beansdev.gen.JavaUtil.getOptimialHashMapSize(JavaUtil.java:839)
   [repeat]   	at org.netbeans.modules.schema2beansdev.JavaBeanClass.genExtendBaseBean(JavaBeanClass.java:4421)
   [repeat]   	at org.netbeans.modules.schema2beansdev.JavaBeanClass.genAllParts(JavaBeanClass.java:119)
   [repeat]   	at org.netbeans.modules.schema2beansdev.JavaBeanClass.generate(JavaBeanClass.java:54)
   [repeat]   	at org.netbeans.modules.schema2beansdev.BeanBuilder.doGeneration(BeanBuilder.java:906)
   [repeat]   	at org.netbeans.modules.schema2beansdev.BeanBuilder.process(BeanBuilder.java:556)
   [repeat]   	at org.netbeans.modules.schema2beansdev.GenBeans.doIt(GenBeans.java:841)
   [repeat]   	at org.netbeans.modules.schema2beansdev.Schema2BeansProcessor.handle(Schema2BeansProcessor.java:163)
   [repeat]   	at org.netbeans.modules.schema2beansdev.Schema2BeansProcessor.process(Schema2BeansProcessor.java:68)
   [repeat]   	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:1023)
...
```

### DIFF
--- a/ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
+++ b/ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
@@ -835,6 +835,7 @@ public class JavaUtil {
         return cls.getName();
     }
 
+    @Deprecated
     public static int getOptimialHashMapSize(Object[] keys) {
         return getOptimialHashMapSize(keys, keys.length * 8);
     }
@@ -847,12 +848,13 @@ public class JavaUtil {
      *
      * @param maxSize the point at which to give up (the maximum size to try)
      */
+    @Deprecated
     public static int getOptimialHashMapSize(Object[] keys, int maxSize) {
         int keyLength = keys.length;
-        int defaultAnswer = keyLength + 1;
+        int defaultAnswer = keyLength * 3 / 2;
         try {
             java.lang.reflect.Field tableField = HashMap.class.getDeclaredField("table");
-            tableField.setAccessible(true);
+            tableField.setAccessible(true); // requires --add-opens for util package
             for (int tableSize = keyLength + 1; tableSize <= maxSize;
                  tableSize <<= 1) {
                 //System.out.println("tableSize="+tableSize);
@@ -874,10 +876,16 @@ public class JavaUtil {
                 }
                 return table.length;
             }
-        } catch (java.lang.IllegalAccessException e) {
+        } catch (IllegalAccessException | NoSuchFieldException e) {
             return defaultAnswer;
-        } catch (java.lang.NoSuchFieldException e) {
-            return defaultAnswer;
+        } catch (RuntimeException ex) {
+            // todo: remove this workaround post JDK 9+ migration (or entire method)
+            // modules on modern JDKs should prefere immutable collections anyway, e.g Map.of();
+            if (ex.getClass().getName().equals("java.lang.reflect.InaccessibleObjectException")) {
+                return defaultAnswer;
+            } else {
+                throw ex;
+            }
         }
         return defaultAnswer;
     }
--- a/.github/workflows/linux-upstream-jdk.yml
+++ b/.github/workflows/linux-upstream-jdk.yml
@@ -1,0 +1,56 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Apache Netbeans
+
+on:
+  push:
+  pull_request:
+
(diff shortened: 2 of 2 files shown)


ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
```diff
@@ -852,7 +852,7 @@ public static int getOptimialHashMapSize(Object[] keys, int maxSize) {
         int defaultAnswer = keyLength + 1;
         try {
             java.lang.reflect.Field tableField = HashMap.class.getDeclaredField("table");
```
JaroslavTulach, line 854: What about removing the reflection all together and just assuming that `keys.length * 3 / 2` is usually good enough?

ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
```diff
@@ -874,10 +874,16 @@ public static int getOptimialHashMapSize(Object[] keys, int maxSize) {
...
                 return table.length;
             }
-        } catch (java.lang.IllegalAccessException e) {
-            return defaultAnswer;
-        } catch (java.lang.NoSuchFieldException e) {
+        } catch (IllegalAccessException | NoSuchFieldException e) {
             return defaultAnswer;
+        } catch (RuntimeException ex) {
+            // todo: remove this workaround post JDK 9+ migration
+            // modules on modern JDKs should prefere immutable collections anyway, e.g Map.of();
```
JaroslavTulach, line 881: There is a way to use `Map.of`. Described in my [Run on JDK8, use JDK11 APIs post](http://wiki.apidesign.org/wiki/AlternativeImplementation) - if you believe this cases deserves that design pattern (I don't), feel free to do implement it.

ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
mbien: i actually wanted to deprecate the method - i forgot, Lets see if it builds, if it does I push again.

ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
mbien: please no - things are complex enough already lets try to remove workarounds instead of adding more :)

mbien:
it works, full-cluster github job is green

ide/schema2beans/src/org/netbeans/modules/schema2beansdev/gen/JavaUtil.java
mbien: deprecated it + added `keylength * 3 / 2` as default return value. Its what the JDK uses anyway if i remember correctly.

matthiasblaesing (approved):
Looks sane to me, though I would make more sense just to return `defaultAnswer` straight away. That way we would get consistent results from this and not rely on runtime behavior. I have no strong feelings either way, just a thought.

mbien:
> Looks sane to me, though I would make more sense just to return `defaultAnswer` straight away. That way we would get consistent results from this and not rely on runtime behavior. I have no strong feelings either way, just a thought.

the only usage of this method is a code generator here:
https://github.com/apache/netbeans/blob/4ae01ea70f4530443343beee3292e880a74099bd/ide/schema2beans/src/org/netbeans/modules/schema2beansdev/JavaBeanClass.java#L4421

this is called at buildtime. It might be also used at runtime somewhere - I haven't investigated further, my main intention was to make NB buildable on JDK 17 without changing the behavior too much. This can be still improved later if needed.

PR #3765, opened 2022-03-11
Adding internal command for registering Orcl DB connection

jhorvath:
Adding internal command for registering Orcl DB connection from Oralce DB Extension


matthiasblaesing (changes requested):
I think it is an error to tie an independend module to a single DB vendor. This won't scale or will we get:

- DB_ADD_CONNECTION_MYSQL
- DB_ADD_CONNECTION_POSTGRESQL
- DB_ADD_CONNECTION_H2
- DB_ADD_CONNECTION_WHATEVERDB

This should at least be contained so that it can be removed, once Oracle goes out of favor. In the DB area we have seen it and see bitrot:

The special code to handle MySQL and JavaDB is becoming useless or even a liability, because no one really cares and it rots. With these we might have a chance to remove them, as they are in their own module and can be ripped out.

jhorvath:
I have modified the command to be generic. Not Oracle specific.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -66,12 +73,67 @@
...
+    private final Gson gson = new Gson();
 
     @Override
     public CompletableFuture<Object> processCommand(NbCodeLanguageClient client, String command, List<Object> arguments) {
         if (!DB_ADD_CONNECTION.equals(command)) {
             return null;
         }
+        
+        String userId = null;
+        String dataSource = null;
```
matthiasblaesing, line 93: dataSource seems to be unused and this should be removed.

matthiasblaesing (changes requested):
I think this is the right direction. I would suggest though, that the parameters expected from the client cover all the parameters required by `DatabaseConnection#create` that way no further dialogs are needed and the implementation is more straight forward.

What is more, parameters should not be assigned DB specific meaning. You use the uppercase username variant as schema. This might be right for your usecase but in general schema and username are not connected, so should be provided independently.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: Removed

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -66,12 +70,66 @@
...
+            if (driver != null && driver.length > 0) {
+                CompletableFuture<String> usernameFuture = userId != null ? CompletableFuture.completedFuture(userId) : client.showInputBox(new ShowInputBoxParams(
+                        Bundle.MSG_EnterUsername(), userId));
+                
+                usernameFuture.thenAccept((username) -> { //NOI18N
+                    if (username == null) {
+                        return;
+                    }
+                    char[] password = m.get(PASSWORD) == null ? null 
+                            : ((List<Double>) m.get(PASSWORD)).stream().map(n -> Character.toString((char) n.byteValue())).collect(Collectors.joining()).toCharArray();
```
matthiasblaesing, line 113: This is wrong. The common reason to collect passwords into char arrays is that they can be cleared on the heap, while strings might be interned and thus retained indefinitetely. However at this point the data already went through several transformations (including a string variant) and thus the argument is moot.

This also breaks non-ascii passwords. There are multiple problematic assumptions:

- the sending side must know, that the password is expected to be a sequence of UTF-16 code units (you go through Character)
- the value seems transferred as a JSON number array and the each element is treated as a byte. Once you get out of the range of a single byte the value is truncated and thus wrong.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -66,12 +70,66 @@
...
+                    CompletableFuture<String> passwordFuture = password != null ? CompletableFuture.completedFuture(new String(password)) : client.showInputBox(new ShowInputBoxParams(
```
matthiasblaesing, line 114: Can't the sending side be expected to provide the password with the data instead of relying on additional callbacks? The same question applies to the username.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -66,12 +70,66 @@
...
     public CompletableFuture<Object> processCommand(NbCodeLanguageClient client, String command, List<Object> arguments) {
         if (!DB_ADD_CONNECTION.equals(command)) {
             return null;
         }
+        
+        String userId = null;
+        String dbUrl = null;
+        String driverClass = null;
+        
+        final Map m = gson.fromJson(gson.toJson(arguments.get(0)), Map.class);
```
matthiasblaesing, line 94: This looks fishy I read this as deserialized value -> gson serialization -> gson deserialization. Why? Without further context a java map is the natural mapping of a JS object.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -66,12 +70,66 @@
...
+        if (m != null) {
+            userId = (String) m.get(USER_ID);
+            dbUrl = (String) m.get(DB_URL);
+            driverClass = (String) m.get(DRIVER);
+            
+        }
+        if (dbUrl != null && driverClass != null) {
```
lbownik, line 101: I dould be more readable is this if block was nested in the previous one - I think.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: This callbacks are used for user input in case sending side didn't provide username and password

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
matthiasblaesing: This is still broken. You just rewrote the stream to a normal iteration. You did not fix the core problem. Again: WHY do you insist on some number array? Without any pain you force yourself to have encoding issues, without any benefits.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
matthiasblaesing: This does not answer the question why the sending side does not directly ask the user and provides everything necessary?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
matthiasblaesing: This is not resolved, as no answert is provided.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: All the user input for this command is in this particular command. Written in Java. I don't see any sense to split this command and implement part of it in typescript.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: Ah, you didn't notice that the stream was transforming via String, I thought it was part of you complain. Javascript doesn't know array of primitives as java does. It is sending array of `number` which is transformed to a `List<Double>` in the protocol. I don't see any reason to have encoding issues.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: This is resolved at line 90: `final Map m = gson.fromJson((JsonObject) arguments.get(0), Map.class);`

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
dbalek: I'm not sure that passing username and password as the command arguments over LSP is a good idea.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
dbalek: Seems to be fixed in the last revision

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
jhorvath: IMO this is better readable as it is now

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
matthiasblaesing: The still unresolved question is: What is the client sending here? How is the array of `number` generated? Storing a password in an array of `number` will need a definition how the password must be encoded, so where is the that defined? I still claim, that you destroy all characters outside the first 255 codepoints.

jhorvath: Unfortunately I cannot check the code where the array of `numbers` is created. I tried to check what happens when I set a password containing characters outside the first 255. I was not able to establish the connection from NetBeans or DB Admin console. So apparently UTF characters are badly handled even before the password gets to this command. I'm switching password to `String` and will try to handle it before the command is called.

MartinBalin (approved):
I see that all substantial requested changes were done. It should be merged.

matthiasblaesing (approved):

PR #3767, opened 2022-03-11
Fix maven daemon detection on windows platform

philippefichet:
In the "bin" folder of "mvnd-0.7.1-windows-amd64" (latest current release), the list of files is as follows:
```
├── mvnd-bash-completion.bash
├── mvnd-sync.exe
├── mvnd.cmd
├── mvnd.exe
└── mvnd.sh
```
I take the liberty of adding the suffix ".cmd" when detecting under windows only

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/execute/cmd/ShellConstructor.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/cmd/ShellConstructor.java
@@ -44,7 +44,9 @@ public class ShellConstructor implements Constructor {
     public List<String> construct() {
 
         // use mvnd if its the home of a daemon
-        String ex = Files.exists(Paths.get(mavenHome.getPath(), "bin", "mvnd")) ? "mvnd" : "mvn"; //NOI18N
+        String mavenDaemonSuffixDetection = Utilities.isWindows() ? ".exe" : "";
+        String ex = Files.exists(Paths.get(mavenHome.getPath(), "bin", "mvnd" + mavenDaemonSuffixDetection)) ? "mvnd" : "mvn"; //NOI18N
+
         List<String> command = new ArrayList<>();
 
         if (Utilities.isWindows()) {


java/maven/src/org/netbeans/modules/maven/execute/cmd/ShellConstructor.java
```diff
@@ -44,7 +44,9 @@ public ShellConstructor(@NonNull File mavenHome) {
     public List<String> construct() {
 
         // use mvnd if its the home of a daemon
-        String ex = Files.exists(Paths.get(mavenHome.getPath(), "bin", "mvnd")) ? "mvnd" : "mvn"; //NOI18N
+        String mavenDaemonSuffixDetection = Utilities.isWindows() ? ".cmd" : "";
+        String ex = Files.exists(Paths.get(mavenHome.getPath(), "bin", "mvnd" + mavenDaemonSuffixDetection)) ? "mvnd" : "mvn"; //NOI18N
```
mbien, line 48: i think we should probe here for `mvnd.exe` instead of `.cmd` because `mvnd` is a binary. It is not super important but it would make it symmetrical.

Thanks for fixing this for windows btw. Feel free to force push.

philippefichet: Change done :)

mbien (approved):
thanks for the fix. looks good.

matthiasblaesing:
Looks also sane to me. Lets get this in.

matthiasblaesing:
Ah - minimal nitpick: @philippefichet could you please squash the two commits into one?

philippefichet:
@matthiasblaesing squash is done ;) (sorry for the delay)

matthiasblaesing:
After 4 tries, travis is also happy. Lets get this in. Thank you.

PR #3503, opened 2022-01-25
Testing sigfile generation for NB13 review. Not for merging!

neilcsmith-net:
Testing sigfile generation for NetBeans 13.  Generated with JDK 8 for review of API purposes.

Generation on JDK 11 fails still with error below - needs to be fixed before release.

```
Warning: incorrect classpath parameter: /usr/lib/jvm/java-11-openjdk-amd64/lib/tools.jar (/usr/lib/jvm/java-11-openjdk-amd64/lib/tools.jar). This directory or jar file will be ignored!
Constant checking: on
Found in total: 948 classes
Fatal error: class com.sun.jdi.connect.IllegalConnectorArgumentsException not found
Fatal error: class com.sun.jdi.connect.IllegalConnectorArgumentsException not found
Fatal error: class com.sun.jdi.AbsentInformationException not found
Fatal error: class com.sun.jdi.AbsentInformationException not found
Fatal error: class com.sun.jdi.AbsentInformationException not found
Fatal error: class com.sun.jdi.AbsentInformationException not found
Fatal error: class com.sun.jdi.connect.IllegalConnectorArgumentsException not found
Fatal error: class com.sun.jdi.connect.IllegalConnectorArgumentsException not found
Warning: class com.sun.jdi.connect.ListeningConnector not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.connect.IllegalConnectorArgumentsException not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.connect.VMStartException not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.connect.Connector$Argument not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.StackFrame not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.VirtualMachine not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.Value not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.connect.AttachingConnector not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.AbsentInformationException not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.ReferenceType not found. Please, add required jar or directory to the classpath.
Warning: class com.sun.jdi.ObjectReference not found. Please, add required jar or directory to the classpath.
Selected by -Package: 83 classes
```

@JaroslavTulach any thoughts?

PR #3778, opened 2022-03-14
Explicitly expose JPDA classes and compile with --release 8 argument

JaroslavTulach:
Fixes to make `ant -q gen-sigtests-release` work on JDK-11 and address issues raised in #3503


JaroslavTulach:
Yes, this commit has been merged into master as part of #3744.

neilcsmith-net:
Ah, OK. I'll sync this to `release130`, trigger sig files, and delete `delivery` then.  Only real reason for moving this through `delivery` was that's how we get things in to both main and release branches.

PR #3781, opened 2022-03-15
exile job for unreliable tests.

mbien:
lets give this a try to make job 13 more reliable

1. identify unreliable tests
2. isolate them into their own job (easier restartable without having to run well behaving tests multiple times)
3. maybe use `travis_retry`

- - -
logs from a different PR where i had to restart job 13 four times in a row:
```
https://app.travis-ci.com/github/apache/netbeans/jobs/563140526
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest

      failed: testOnceExistedDeletedItemInfo

org.netbeans.modules.java.lsp.server.project.LspBrokenReferencesImplTest

      failed: testIgnoredErrorsTimeout

org.netbeans.modules.java.source.parsing.PartialReparseTest

      failed: testIntroduceParseError2
---
https://app.travis-ci.com/github/apache/netbeans/jobs/563138079
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest

      failed: testEmptyPackageRemains
---
https://app.travis-ci.com/github/apache/netbeans/jobs/563294404
org.netbeans.modules.java.lsp.server.project.LspBrokenReferencesImplTest

      failed: testIgnoredErrorsTimeout
---
https://app.travis-ci.com/github/apache/netbeans/jobs/563294404
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest

      failed: testProjectSourcePackages
```

PR #3785, opened 2022-03-15
github actions ci pipeline.

mbien:
simple CI pipeline
 - main job builds everything and uploads the build artifact tar (retention set to 1 day)
 - all other jobs extract the tar and run their tests
 - everything is converted into matrix jobs, see commit msg for details
 - bumped OS versions and JDK versions where possible

Job dependencies only work within a workflow, so we have to merge them all (again). The reason they were split was to be able to restart them individually on failure.

Today, Github finally released a feature which allows individual jobs to be restarted. Assuming it works as expected, we won't need separate workflows anymore.

Lets see how this works and if you can actually restart a job.

pipeline:
pipeline

regarding job restarts:
Seems to be working as promised, however: restarts are only possible after the complete workflow finished.

rerun


mbien:
windows tar.gz extractor action doesn't work anymore:
`ihiroky/extract-action@v1 is not allowed to be used in apache/netbeans. Actions in this workflow must be: within a repository owned by apache, created by GitHub, verified in the GitHub Marketplace or match the following: `

disabled win job for now

neilcsmith-net:
JFTR - various ASF projects are using git submodules if there is a need for third-party actions - eg. see https://github.com/apache/superset/pull/12709

mbien:
> JFTR - various ASF projects are using git submodules if there is a need for third-party actions - eg. see [apache/superset#12709](https://github.com/apache/superset/pull/12709)

hi @neilcsmith-net, I didn't know that windows learned what a tar is. It turns out that there is no need for that action. But good to know that there is a trick available if it would be still needed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,435 @@
...
+
+      - name: Archive Build
+        run: tar -czf /tmp/build.tar.gz --exclude ".git" .
+
+      - name: Upload Build
+        uses: actions/upload-artifact@v2
+        with:
+          name: build
+          path: /tmp/build.tar.gz
+          retention-days: 1
```
mbien, line 61: @neilcsmith-net do you think it is a problem from the apache resources perspective that a build would create a 1.5 GB upload ([accessible from bottom of this page](https://github.com/apache/netbeans/actions/runs/1989525198))? 1 day retention is already the minimal setting.

mbien:
@junichi11 @oyarzun the PHP code folding test appears to have some issues on JDK 11, is this known?

build on 11, test on 11:
https://github.com/apache/netbeans/actions/runs/2000384843
both, win and linux nodes fail

build on 8, test on 8 and 11 (only enabled for debugging purposes):
https://github.com/apache/netbeans/actions/runs/2001457064
linux/jdk11 node fails, other three nodes are green

@oyarzun I have a question, why did you change $OPTS to $env:OPTS? Is this valid syntax? (I changed it back when i merged the workflows into a matrix job since i had to pick one variant - but if that was intended i can change it back)

junichi11:
@mbien Maybe, the same as https://github.com/apache/netbeans/pull/3444 ? (See my comments)

oyarzun:
> @oyarzun I have a question, why did you change $OPTS to $env:OPTS? Is this valid syntax? (I changed it back when i merged the workflows into a matrix job since i had to pick one variant - but if that was intended i can change it back)

@mbien According to the GitHub docs Windows uses powershell and the environment variables should use $env:VARNAME.

https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables

>By default, Linux runners use the bash shell, so you must use the syntax $NAME. If the workflow specified a Windows runner, you would use the syntax for PowerShell, $env:NAME.

mbien:
> https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables
> 
> > By default, Linux runners use the bash shell, so you must use the syntax $NAME. If the workflow specified a Windows runner, you would use the syntax for PowerShell, $env:NAME.

oh boy. I feared it would be something like that. This will complicate matrix jobs. Thanks for the info.

mbien:
@junichi11 i am going to leave the PHP tests on JDK 8, the flag didn't help unfortunately. Would be good if someone else could take a look at this so that we can bump it to 11 - but no need to hurry.

@oyarzun `shell: bash` worked: we can now use the same shell on all nodes

assuming this build is green, I will do a force push with another matrix job update for the base-build and set this PR to ready for review and will maybe drop a mail to the dev list about this and the other draft PR #3781.

junichi11:
@mbien OK.

.github/workflows/ensure-jdk8.yml
```diff
@@ -1,60 +0,0 @@
...
-        uses: actions/setup-java@v2
-        with:
-            distribution: 'zulu'
-            java-version: 11
-
-
-      - name: Build
-        run: ant $OPTS build
-
-      - name: Test java.source.base on JDK-8
```
JaroslavTulach, line 59: I am not sure what replaces testing on JDK8? I am against deleting any jobs that perform testing on JDK8 unless they are replaced by appropriate alternative.

.github/workflows/main.yml
```diff
@@ -0,0 +1,342 @@
...
+        with:
+          name: build
+
+      - name: Extract
+        run: tar -xzf build.tar.gz --exclude nbbuild/build/test
+
+      - run: brew install ant
+
+      - name: Test platform/masterfs.macosx
+        run: ant -f platform/masterfs.macosx test
```
JaroslavTulach, line 200: Could we use _test distribution_ instead of the tar?

JaroslavTulach (commented):
Great to see us to finally use the power of pipelines. Thanks for the work. There used to be a [test distribution](https://lists.apache.org/thread/qdltpyjrznj5bqdrq4symcl02npg63z3) effort and according to Petr Zajac, it [still seems to work](https://lists.apache.org/thread/kwhgqfs9sv3yqs9ptt1h391hv5cy4lkc):

> It would be great to reconfigure the GitHub actions to do the build of IDE ZIP 
> and test distribution ZIP only once.
> Only then other "testing" jobs are started, download the two ZIP files 
> and use them to perform the tests only

The solution proposed here doesn't seem to use the test distribution neither test against binary ZIP file. Rather it TARs all the sources+binaries and passes them to another job. Is there any benefit of this approach over test distribution (except that it has been implemented)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,342 @@
...
+# secondary jobs
+  linux-commit-validation:
+    needs: base-build
+    name: Commit Validation on Linux/JDK ${{ matrix.java }} 
+    runs-on: ubuntu-latest
+    env:
+      ANT_OPTS: -Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json
+    strategy:
+      matrix:
+        java: [ '11' ]
```
JaroslavTulach, line 81: Commit validation shall run on JDK8 as well.

.github/workflows/main.yml
mbien: potentially. What is the ant target for creating the test distribution? Some secondary jobs probably won't work with that though. E.g the nbm one.

mbien: but let me ask first: is there a reason you prefer a test distribution over the full workspace approach? It is basically like testing a fresh clone, just from a snapshot which is already built.

My main argument against it is the artifact size. But I am not sure if that is a problem with 1 day retention size. Its likely that we will have to ask someone from apache infra anyway - even if we reduce the artifact size.

.github/workflows/ensure-jdk8.yml
matthiasblaesing: Testing on JDK 8 is covered by travis: https://github.com/apache/netbeans/blob/0665eb12aaa64cbe3b7bae657f46594a579a41cc/.travis.yml#L383

.github/workflows/main.yml
mbien: @matthiasblaesing good to know! Could we leave it here and remove the jobs on travis instead? I prefer gh actions over travis personally - not sure what others think (and it is using a pipeline which means less building which makes about 10min per job).

We might have to leave the job for 15 on travis since I don't think anyone maintains EOL JDK setups for gh actions. But if someone does - even better.

matthiasblaesing: For travis vs. github action I have no strong feelings either way, so from my perspective we should just see, that they run only once.

For 15 I would consider moving to 17 and limit the test matrix to LTS versions (8, 11, 17)  and the current early access version. I.e. from this list:

https://github.com/java-native-access/jna/pull/1428/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR16

I would drop 18.

mbien: @matthiasblaesing cv doesn't work on 15+ unfortunately (like a lot of other tests). Blocked by netbinox the last time i tried. If that is unblocked (and no other issue shows up) its likely that it would also require the `--add-opens` lists (which we [now have](https://github.com/apache/netbeans/tree/master/nbbuild/jms-config), so we made some progress in that area at least).

mbien: added (8, 11, 15) to the commit-validaiton matrix + removed the same jobs from travis.

matthiasblaesing: > @matthiasblaesing cv doesn't work on 15+ unfortunately (like a lot of other tests). Blocked by netbinox the last time i tried. If that is unblocked (and no other issue shows up) its likely that it would also require the `--add-opens` lists (which we [now have](https://github.com/apache/netbeans/tree/master/nbbuild/jms-config), so we made some progress in that area at least).

I had a look at the output and this is the offender:

```
    [junit] März 24, 2022 8:33:03 PM org.netbeans.ProxyURLStreamHandlerFactory register
    [junit] SCHWERWIEGEND: No way to find original stream handler for jar protocol
    [junit] java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.net.URLStreamHandler java.net.URL.handler accessible: module java.base does not "opens java.net" to unnamed module @15eb5ee5
    [junit]     at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
    [junit]     at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
    [junit]     at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:184)
    [junit]     at java.base/java.lang.reflect.Field.setAccessible(Field.java:178)
    [junit]     at org.netbeans.ProxyURLStreamHandlerFactory.register(ProxyURLStreamHandlerFactory.java:59)
    [junit]     at org.netbeans.JarClassLoader.<clinit>(JarClassLoader.java:117)
    [junit]     at org.netbeans.MainImpl.execute(MainImpl.java:153)
    [junit]     at org.netbeans.MainImpl.main(MainImpl.java:60)
    [junit]     at org.netbeans.Main.main(Main.java:58)
    [junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    [junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:76)
    [junit]     at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:51)
```

and indeed `commit-validation`  succeeds if you run:

```bash
export JAVA_TOOL_OPTIONS='--add-opens=java.base/java.net=ALL-UNNAMED
```

before. I tested with a jdk-18-ea version.

mbien: Thats interesting! Going to try to replace the JDK 15 job with 17 + add 18. I am wondering why i had issues last time i tried.

.github/workflows/main.yml
```diff
@@ -0,0 +1,347 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: build
+
+      - name: Extract
+        run: tar -xzf build.tar.gz --exclude nbbuild/build/test
+        
+      - name: JMS config
+        if: ${{ matrix.java != '8' }}
+        run: echo "JAVA_TOOL_OPTIONS=--add-opens=java.base/java.net=ALL-UNNAMED --add-opens=java.desktop/javax.swing=ALL-UNNAMED" >> $GITHUB_ENV
```
mbien, line 101: 19-ea fails currently. It appears JDK 19 requires now `-Djava.security.manager=allow` to be set, since the warning became now a runtime exception:
```
[junit] java.lang.UnsupportedOperationException: The Security Manager is deprecated and will be removed in a future release
    [junit] 	at java.base/java.lang.System.setSecurityManager(System.java:416)
    [junit] 	at org.netbeans.TopSecurityManager.install(TopSecurityManager.java:525)
    [junit] 	at org.netbeans.core.NbLifecycleManager.advancePolicy(NbLifecycleManager.java:69)
    [junit] 	at org.netbeans.core.GuiRunLevel.run(GuiRunLevel.java:84)
    [junit] 	at org.netbeans.core.startup.Main.start(Main.java:312)
    [junit] 	at org.netbeans.core.startup.TopThreadGroup.run(TopThreadGroup.java:98)
    [junit] 	at java.base/java.lang.Thread.run(Thread.java:828)
```
but even if the flag is set (tested it locally) it would fail later:
```
 [nb-javac] Compiling 1 source file to /home/mbien/NetBeansProjects/netbeans/java/java.hints/build/classes
 [nb-javac] /home/mbien/NetBeansProjects/netbeans/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unused.java:82: error: cannot find symbol
 [nb-javac]                 fix = JavaFixUtilities.safelyRemoveFromParent(ctx, Bundle.FIX_RemoveUsedElement(name), ud.unusedElementPath);
 [nb-javac]                                       ^
 [nb-javac]   symbol:   method safelyRemoveFromParent(HintContext,String,TreePath)
 [nb-javac]   location: class JavaFixUtilities
 [nb-javac] 1 error

```
I will swap 19-ea with 18 for now since more work is needed there.

mbien: 19-ea job is now green too

vieiro (approved):
This is a great job!
I think we should merge it and see if we have green lights in PRs from now on.

lkishalmi (approved):
Have not checked in detail. What I read seems to be sane. The direction we are going with this is good. Let's have it merged!

.github/workflows/main.yml
mbien: @matthiasblaesing @neilcsmith-net if I take a look at https://github.com/pricing

I see "Free for public repositories" under the Package storage item. Does this mean we shouldn't have to worry here?

I could try to add a cleanup job which removes the upload again. (for example a variant of: https://github.com/actions/upload-artifact/issues/5#issuecomment-876499762 but only for the current workflow)

Or should we give it a try first?

.github/workflows/main.yml
mbien: found the OpenJDK PR: https://github.com/openjdk/jdk/pull/5204

.github/workflows/main.yml
neilcsmith-net: @mbien I'd run it past infra@.  Not sure of the answer.  Pricing may not be relevant info - ASF is sponsored by GitHub (with enterprise org features IIRC) but what that means for reasonable resource usage in practice I don't know for sure.

oyarzun (commented):
@mbien can you also update the GitHub Actions badge in the readme?

```markdown
* [![Build Status](https://github.com/apache/netbeans/actions/workflows/main.yml/badge.svg)](https://github.com/apache/netbeans/actions/workflows/main.yml)
```

.github/workflows/main.yml
mbien: @neilcsmith-net excellent, meanwhile I am going to try to add a cleanup job: [9e719ee](https://github.com/apache/netbeans/pull/3785/commits/9e719ee57d8b1468b609cd264f4880901ec9e7d1)

mbien:
```
Run ./.github/actions/delete-artifact/
Successfully deleted artifact "build".
```
it works :)

mbien:
@oyarzun done

artifact retention:
I added a cleanup job which removes the artifact at the very end of the workflow run.
last run: https://github.com/apache/netbeans/actions/runs/2056554946
(notice there is no artifact linked at the bottom of the page)

Retention time doesn't matter anymore, it is now removed right away. The action is deployed via a git sub module as suggested by @neilcsmith-net https://github.com/apache/netbeans/pull/3785#issuecomment-1069037190

oyarzun (approved):

mbien:
squashed all commits.

.github/workflows/main.yml
mbien: @neilcsmith-net ok to merge or should we wait for infra reply? I am no longer concerned about the storage issue due to the cleanup job which was much easier to implement than anticipated thanks to the third party gh action (deployed via the submodule trick).

neilcsmith-net: @mbien I reckon it's reasonable as is. Let's merge and deal with any fallout later! :smile:

neilcsmith-net (approved):
Thanks @mbien - looks good!

PR #3787, opened 2022-03-16
Tame Gradle Project Loading

lkishalmi:
@sdedic this is an attempt to tame the Gradle project loading methods and calls. My motivation was, that I started to get lost in the calls when a Gradle project, loaded re-loaded.

What I'm trying here is to move the options of the different kind of re-loads into a separate class as [GradleLoadOptions](extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleLoadOptions.java) so the code can be more descriptive, and the usage of certain load flags could be more easily tracked. I could be completely wrong with this PR and my approach could be wrong as well. Your feedback is essential.

I've tried not to break things, though made a few not compatible changes. loadedProjectSerial and currentSerial in NBGradleProjectImpl were moved to a new sequence field to GradleProject, so dumpProject() cannot set the loadedProjectSerial to zero.

Tests seems to be fine, save NbGradleProjectImpl.testEventsProcessedBeforeCompletion() which hangs in infinite wait. Unfortunately I could not crack what that test really does, so I need some help there. 

There are some unused code mentioning GradleLoadContext that shall be ignored.

I have purposefully not changed the javadoc on the load methods yet, so the former parameters can be looked up while replacing the calls with GradleLoadOptions arguments.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/ActionProviderImpl.java
@@ -322,9 +322,10 @@ public class ActionProviderImpl implements ActionProvider {
                 final ActionProgress g = ActionProgress.start(context);
                 GradleLoadOptions opts = GradleLoadOptions
                         .loadForQuality(maxQualily)
+                        .force()
                         .withMessage(loadReason)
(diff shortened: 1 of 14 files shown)


sdedic:
ok to review now, or is it still being overhauled (it's marked as wip) ?

PR #3788, opened 2022-03-16
Add Apache NetBeans Snap Package as an option to GitHub Bug report

lkishalmi:
Just a small addition.


neilcsmith-net (approved):
Yes makes sense to be separate from provided installers.

PR #3793, opened 2022-03-16
Maven error/warnings refactored to use Parsing API.

sdedic:
Originally I was trying to bridge project problems to LSP. During the course I have realized that Maven's errors, underlined in the editor, are provided by `UpToDateStatusProvider` implementation - this one is in charge or reporting start/stop computations in annotation's gutter.
The Provider however monitored **on-disk file changes**, so potential errors were only reported after file save, not (as usual) when a change to the opened document has been made by the user. This approach bypasses alll the machinery of Parsing API which is more suitable for tracking document changes and/or caret/selection in the live document. In addition, this approach wouldn't work in the LSP client scenario at all.

The StatusProvider implementation was broken anyway as for example [StatusProvider](https://github.com/sdedic/incubator-netbeans/blob/master/java/maven.hints/src/org/netbeans/modules/maven/hints/pom/StatusProvider.java#L284) kept the status `UP_TO_DATE_PROCESSING` if there was a non-empty selection ... The `UpToDateStatusProvider` + hints is generally broken, see [NETBEANS-6477](https://issues.apache.org/jira/browse/NETBEANS-6477). I will not fix that in this PR.

The PR adapts Maven implementation to use Parsing API - this allows to react immediately, and since I provided a custom `ModelSource` (see `M2S` class) to maven processing, even the live editor content can be analyzed by the Maven library.

Tests for this module were broken - test data file was missing; I've created some + added necessary env setup so the test passes now.

I've also discovered that Tasklist Maven integration is broken for some time - see [NETBEANS-6476](https://issues.apache.org/jira/browse/NETBEANS-6476), will not fix in this PR either.

### DIFF
--- a/java/maven.embedder/nbproject/project.xml
+++ b/java/maven.embedder/nbproject/project.xml
@@ -198,6 +198,7 @@
                 <package>org.apache.maven.artifact.resolver.filter</package>
                 <package>org.apache.maven.artifact.resolver</package>
                 <package>org.apache.maven.artifact.versioning</package>
+                <package>org.apache.maven.building</package>
                 <package>org.apache.maven.cli</package>
                 <package>org.apache.maven.cli.configuration</package>
                 <package>org.apache.maven.embedder</package>
--- a/java/maven.hints/nbproject/project.properties
+++ b/java/maven.hints/nbproject/project.properties
@@ -21,3 +21,5 @@ javac.source=1.8
 
 test.config.stableBTD.includes=**/*Test.class
 requires.nb.javac=true
+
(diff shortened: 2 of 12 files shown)


ppisl (commented):
I have not seen an issue.

ppisl (approved):

sdedic:
Sorry :( I've accidentally committed here changes that belong to other PR; reverted back to the reviewed state.

JaroslavTulach (approved):
Nice architecture improvement, I think.

dbalek (approved):
Looks fine. Thanks.

PR #3794, opened 2022-03-16
Upgrade Commons-IO from 2.6.0 to 2.11.0

pepness:
Library Notes:
 - Require Java 8
- 63 Bug fixes
- 72 New Features
- 64 Improvements

NetBeans Notes:
- Remove duplicated jar and add dependency to `java.lsp.server` module, this is the only module that requires `org.apache.commons.io`
- Since version 2.7 commons-io change it's `Bundle-SymbolicName` in the MANIFEST.MF from `org.apache.commons.io` to `org.apache.commons.commons-io`, **this newly added hyphen breaks** the Ant target `commit-validation` with the 2 respective GitHub Tasks `Commit-validation (on macOS)` and `Commit-validation (on Linux)`. This should be fixed by adding a hyphen `-` to a regular expression in `platform/openide.modules/Dependency.java line 180` but I could not make it work. I did add a two new tests to validate codename values with a hyphen and underscore. According to the spec this is a valid name:
    - name: alphanum *headerchar
    - headerchar: alphanum | - | _

Testing:
- Full build done
- Verify successful execution of unit tests for modules `java/java.lsp.server` (the one that use this library)
- Checked the file is in `{nb_all}/platform/modules/org-apache-commons-io.jar`

[Web Page](https://commons.apache.org/proper/commons-io/)
[Release Notes](https://commons.apache.org/proper/commons-io/changes-report.html)
[JARManifest Spec](https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Name-Value_pairs_and_Sections)


pepness:
Since version 2.7 commons-io change it's Bundle-SymbolicName in the MANIFEST.MF from org.apache.commons.io to org.apache.commons.commons-io, this newly added hyphen breaks the Ant target commit-validation with the 2 respective GitHub Tasks Commit-validation (on macOS) and Commit-validation (on Linux). This should be fixed by adding a hyphen - to a regular expression in platform/openide.modules/Dependency.java line 180 but I could not make it work. I did add a two new tests to validate codename values with a hyphen and underscore. According to the spec this is a valid name. [JARManifest Spec](https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Name-Value_pairs_and_Sections)

neilcsmith-net:
@dbalek @matthiasblaesing any thoughts?

matthiasblaesing:
My gut feeling is, that it should be save to relax the naming rules for code name bases. The dash is the only character that can't be used in code name base, but the NetBeans module name. This needs further checks though.

matthiasblaesing:
@pepness sorry for the late reply. I had a look at this and this needs a documented API change, that will need a review, we should also rename the NetBeans module (yes commons-io broke compatibilty).

This is my suggestion how to fix it: https://github.com/matthiasblaesing/netbeans/tree/pr-3794 (this should be rebased onto master and most probably squashed). Please have a look.

mbien:
restarting CI with all tests enabled

mbien:
`core.startup` is failing after restart too:
```
junit.framework.AssertionFailedError
    at org.netbeans.core.startup.ConsistencyVerifierTest.testIAE(ConsistencyVerifierTest.java:114)
```

mbien:
PR has conflicts

mbien:
moving to NB20

mbien:
would be good to resolve the conflicts to get a fresh CI run in

matthiasblaesing:
We have now #6480, the approach @lahodaj took to fix the hyphen vs. underscore problems looks easier to me and does not involve changing the netbeans module system.

mbien:
closing, since the other PR is now integrated

PR #3802, opened 2022-03-17
[NETBEANS-4826] Fixing java indexing in Groovy

ppisl:
This will turn back on indexing of Groovy files in Java. This means that Groovy classes can now be offered in java source files.  The Groovy file is used to create virtual Java source code that is indexed and stored in the java index. This will not only enable indexing, but features like code completion, hyperlinks, goto navigation should also work. 

More fixes were needed to get this working again. There are also a few fixes that improve the current staff, such as:

- static fields and methods are offered as fields and methods in import statements code completion - they were previously offered as classes in code completion. 
- correcting types of constructors that are strored in Groovy Index
- JavaElementHandles now are created with the fully qualifed names, there were created only with simple names. 
- adding @Trusted annotation for ForwardingJavaFileObject implementations

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/util/CompletionContext.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/util/CompletionContext.java
@@ -298,19 +298,28 @@ public final class CompletionContext {
         boolean openBraceBeforePosition = false;
         // is there package statement?
         boolean afterPackagePosition = false;
-
+        boolean canBeImport = true;
+        
         ts.move(position);
 
         while (ts.isValid() && ts.movePrevious() && ts.offset() >= 0) {
             Token<GroovyTokenId> t = ts.token();
             if (t.id() == GroovyTokenId.LBRACE) {
                 openBraceBeforePosition = true;
+                canBeImport = false;
             } else if (t.id() == GroovyTokenId.LITERAL_class || t.id() == GroovyTokenId.LITERAL_interface || t.id() == GroovyTokenId.LITERAL_trait) {
                 classDefBeforePosition = true;
                 break;
             } else if (t.id() == GroovyTokenId.LITERAL_package) {
                 afterPackagePosition = true;
                 break;
+            } else if (canBeImport && t.id() == GroovyTokenId.LITERAL_import) {
+                return CaretLocation.INSIDE_IMPORT;
+            }
+            
+            if (canBeImport && !(t.id() == GroovyTokenId.DOT || t.id() == GroovyTokenId.IDENTIFIER
+                    || t.id() == GroovyTokenId.WHITESPACE || t.id() == GroovyTokenId.LITERAL_static)) {
+                canBeImport = false;
             }
         }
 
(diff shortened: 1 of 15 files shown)


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/util/CompletionContext.java
```diff
@@ -298,19 +298,28 @@ private CaretLocation getCaretLocationFromRequest() {
...
                 classDefBeforePosition = true;
                 break;
             } else if (t.id() == GroovyTokenId.LITERAL_package) {
                 afterPackagePosition = true;
                 break;
+            } else if (canBeImport && t.id() == GroovyTokenId.LITERAL_import) {
+                return CaretLocation.INSIDE_IMPORT;
+            }
+            
+            if (canBeImport && !(t.id() == GroovyTokenId.DOT || t.id() == GroovyTokenId.IDENTIFIER
```
sdedic, line 320: Q: how will this react in context like `Whatever.class.` ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/TypesCompletion.java
```diff
@@ -313,12 +314,24 @@ private void addToProposalUsingFilter(Set<TypeHolder> alreadyPresent, TypeHolder
...
             return;
         }
+        
+        if (type.getHandle() != null
+                &&!(type.getHandle().getKind().isClass() || type.getHandle().getKind().isInterface())
+                && request.location != CaretLocation.INSIDE_IMPORT) {
+            return;
+        }
 
+        String ownerFQN = GroovyUtils.stripClassName(fqnTypeName);
```
sdedic, line 324: Use `GroovyUtils.getPackageName()`, returns name w/o trailing dot.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/TypesCompletion.java
```diff
@@ -313,12 +314,24 @@ private void addToProposalUsingFilter(Set<TypeHolder> alreadyPresent, TypeHolder
...
+        if (!ownerFQN.isEmpty() && ownerFQN.endsWith("."))  {
+            ownerFQN = ownerFQN.substring(0, ownerFQN.length() - 1);
+        }
+        
         // We are dealing with prefix for some class type
         JavaElementHandle jh = null;
         if (type.getHandle() != null) {
-            jh = new JavaElementHandle(fqnTypeName, typeName, type.getHandle(), Collections.emptyList(), Collections.emptySet());
+            jh = new JavaElementHandle(typeName, ownerFQN, type.getHandle(), Collections.emptyList(), Collections.emptySet());
```
sdedic, line 332: Uh-oh ... this was a stupid bug; thanks.

sdedic (commented):
Please leave a comment in the TypeHolder, that it is actually being (mis)used also for static fields and methods, despite its name.

dbalek (approved):

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/completion/util/CompletionContext.java
ppisl: Can we have import statement like `import Whatever.class` ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/completion/TypesCompletion.java
ppisl: Done

sdedic (approved):

PR #3808, opened 2022-03-18
ide/lsp.client ack mime-type fonts&colors

vieiro:
- Allow users of `ide/lsp.client` to use mime-type specific fonts&color settings (defined as usual), otherwise default to `text/textmate`.
- Slight performance improvement by extracting `MimeLookup.getLookup(...).lookup(FontColorSettings.class)` once per file, not per token.


vieiro:
While we're at it, we may also want to acknowledge LSP's SemanticTokenModifiers and SemanticTokenTypes. NetBeans will fall-back to proper coloring for these.

PR #3828, opened 2022-03-21
Made GradleCommandLine aware of GradleVersion

lkishalmi:
Well, the original motivation for this is that Gradle configuration cache command line argument is only supported from Gradle 6.5 and above. That makes projects before 6.5 non-buildable if the config cache global option is set. So this change makes the GradleCommandLine aware of GradleVersion and removes options which are not yet or no longer supported.

@JaroslavTulach Is it Ok to expose a third party class GradleVersion that much throughout an API?

This is a working draft. I'll add additional, tests, logging and documentation.


extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleVersionRange.java
```diff
@@ -0,0 +1,81 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.gradle.api.execute;
+
+import org.gradle.util.GradleVersion;
+
+/**
+ *
```
sdedic, line 24: Javadoc + `@since`

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -88,58 +89,82 @@
...
+         * @return {@code true} if this option is supported by the IDE
+         */
         boolean isSupported();
+        /**
+         * Shall return {@code true} if this option is supported by the provided
+         * GradleVersion.
+         * 
+         * @param ver the GradleVersion to check.
+         * @return {@code true} if the provided Gradle version supports this option.
+         */
```
sdedic, line 104: `@since`

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -525,15 +596,22 @@ public ParametricArgument parse(String arg, Iterator<String> args) {
     }
     //</editor-fold>
 
+    final GradleVersion gradleVersion;
     final Set<Argument> arguments = new LinkedHashSet<>();
     final Set<String> tasks = new LinkedHashSet<>();
 
-    public GradleCommandLine(GradleCommandLine cmd) {
+    public GradleCommandLine(GradleVersion gradleVersion, GradleCommandLine cmd) {
```
sdedic, line 603: Javadoc + `@since`

extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
```diff
@@ -194,7 +199,15 @@ private static GradleProject loadGradleProject(ReloadContext ctx, CancellationTo
         }
         return ret;
     }
-    
+
+    private static GradleVersion compatibleGradleVersion(Project prj) {
```
sdedic, line 203: Nitpick: Does this return a gradle version **compatible** with the current project, or the **configured** one ? I mean if the project does not specify anything, the current default version is returned regardless of project's structure

extide/gradle/src/org/netbeans/modules/gradle/loaders/ModelCache.java
```diff
@@ -100,4 +109,12 @@ private void load() {
     public synchronized State getState() {
         return state;
     }
+
+    private GradleVersion compatibleGradleVersion() {
```
sdedic, line 113: Duplicate method - could it be moved to some module-private Util class ?

sdedic (commented):
The `--configuration-cache` is a non-essential option; nothing (much - except the performance) will probably happen if it would be missing from new gradles - and it is added by the IDE (not by the user). For other options, or user-provided options it would be more appropriate to fail the build, as the build will do "something different than the user intended" or "different what the IDE expects".

I would suggest to tag individual GradleOptionItem with some `boolean isEssential()` and remove only those that are not essential - otherwise throw an exception and fail the operation.

Second, since the code that does `cmd.setFlag(GradleCommandLine.Flag.CONFIGURATION_CACHE,` already knows the project's configured gradle version, could it do it just conditionally, if the flag is compatible with the gradle version of the GradleCommandLine being constructed ?

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
lkishalmi: There will be an updated commit with javadoc and tests and arch changes. Let's just check the code itself for now.

extide/gradle/src/org/netbeans/modules/gradle/loaders/ModelCache.java
lkishalmi: Could find a place for it somewhere...

extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
lkishalmi: Oh, good catch, I left the checking of the IDE Java Version out from this computation!

sdedic:
> do you see value in this PR for the future?

I personally like the version meta-data for options / parameters; especially the user action mapping (custom actions) may eventually go out of sync. Can be used to report project problems. Code completion can also offer only relevant items in action editors.

The only debatable issue is the leakage of `GradleVersion` from Gradle libs outside of Gradle Projects API to its clients -- @jtulach should have an option on that

lkishalmi:
OFFTOPIC: I wish I could have that debate with you guys next to a beer somewhere in Prague. It's been 20 almost twenty years since I last visited there.

I was thinking about use GradleDistribution instead of GradleVersion in the API, and the GradleVersionRange can be moved out of the API package to somewhere else...

lkishalmi:
Well, I modified the API to use GradleDistribution instead directly exposing GradleVersion. Added javadoc and some apichange documentation plus tests.

PR #3829, opened 2022-03-21
Avoid spurious errors when loading projects.

sdedic:
When loading more complex projects like [`micronaut-test`](https://github.com/micronaut-projects/micronaut-test), NetBeans IDE reports a number of 'unresolvable artifact' errors:
unresolvable-artifact

Indeed, even the commandline gradle produces them:
```
$ ./gradlew dependencies

[...]

testCodeCoverageReportExecutionData - Supplies JaCoCo coverage data to the testCodeCoverageReport.  External library dependencies may appear as resolution failures, but this is expected behavior.
+--- project :test-core
|    +--- io.micronaut:micronaut-inject FAILED
|    +--- io.micronaut:micronaut-runtime FAILED
|    +--- io.micronaut:micronaut-bom:3.3.4
```
... but note the comment at the top of the configuration.

We can exclude configurations from loading entirely, but it seems that the 'unresolvable' problems appear in either hidden (not visible) or otherwise poorly usable (!canBeConsumed) configurations. So I've disabled unresolvable artifact reporting in this case. 

In addition, in the project `test-kotest`, the IDE complained that a property `getFreeCompilerArgs` is not present at `JvmKotlinOptions` - that's caused by using the getter method name instead of property name in the call to `getProperty`.

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -292,7 +292,7 @@ class NbProjectInfoBuilder {
                                 try {
                                     compilerArgs = (List<String>) getProperty(compileTask, "options", "compilerArgs");
                                 } catch (Throwable ex2) {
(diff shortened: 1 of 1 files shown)


lkishalmi (approved):
Looks good! Thank you!

sdedic:
@lkishalmi Thanks ... I've just wanted to give 72h grace period for Matthias to comment, he committed into the NB Tooling code as well.

Issue #3801, opened 2022-03-17
Gradle Project opens with problem and it cannot be resolved


makiam:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

Once I opened Gradle project I see message Project Cannot be loaded...
Once I try to invoke Reolve Project Problems I see next message in dialog:
Could not open cp_init generic class cache for initialization script 'C:\Users\MaksK\AppData\Local\NetBeans\Cache\dev\gradle\nb-tooling.gradle' (C:\Users\MaksK\.gradle\caches\7.4.1\scripts\lq5c633n4lhixxj9vx3i6swd). Could not compile initialization script 'C:\Users\MaksK\AppData\Local\NetBeans\Cache\dev\gradle\nb-tooling.gradle'. startup failed: initialization script 'C:\Users\MaksK\AppData\Local\NetBeans\Cache\dev\gradle\nb-tooling.gradle': 22: Unexpected character: '\'' @ line 22, column 25. classpath files('C:\netbeans\extide\modules\gradle\netbeans-gradle-tooling.jar') ^

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Windows 10 version 10.0 running on amd64

### JDK

Java: 13.0.1; Java HotSpot(TM) 64-Bit Server VM 13.0.1+9

### Apache NetBeans packaging

Own source build

### Anything else

IDE options set to use Custom Gradle installed 7.41
Prefer to use wrapper is turned off

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

lkishalmi:
Thank you for catching this one the new code was not tested on Windows (as I'm lucky enough to have none around). I'll provide a fix today.

PR #3831, opened 2022-03-21
Escape backslash in Gradle tooling jar path on Windows

lkishalmi:
This is a fix for #3801. Moved two constants from public to private as they shall be not used elsewhere and were not part of the public API.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleDaemon.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleDaemon.java
@@ -53,8 +53,8 @@ public final class GradleDaemon {
     static final String INIT_SCRIPT_NAME = "modules/gradle/nb-tooling.gradle"; //NOI18N
     static final String TOOLING_JAR_NAME = "modules/gradle/netbeans-gradle-tooling.jar"; //NOI18N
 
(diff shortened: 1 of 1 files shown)


extide/gradle/src/org/netbeans/modules/gradle/loaders/GradleDaemon.java
```diff
@@ -53,8 +53,8 @@
     static final String INIT_SCRIPT_NAME = "modules/gradle/nb-tooling.gradle"; //NOI18N
     static final String TOOLING_JAR_NAME = "modules/gradle/netbeans-gradle-tooling.jar"; //NOI18N
 
-    public static final String PROP_TOOLING_JAR = "NETBEANS_TOOLING_JAR";
-    public static final String TOOLING_JAR = InstalledFileLocator.getDefault().locate(TOOLING_JAR_NAME, NbGradleProject.CODENAME_BASE, false).getAbsolutePath();
+    private static final String PROP_TOOLING_JAR = "NETBEANS_TOOLING_JAR";
+    private static final String TOOLING_JAR = InstalledFileLocator.getDefault().locate(TOOLING_JAR_NAME, NbGradleProject.CODENAME_BASE, false).getAbsolutePath().replace("\\", "\\\\");
```
mbien, line 57: thanks for using the right replace method :)

mbien (approved):
looks good

PR #3833, opened 2022-03-22
Enable Maven editor hints in LSP

sdedic:
An extension of #3793, this hint bridges Maven hints to LSP. The ErrorProvider runs explicitly parser task internals and report to LSP client.

Note that these changes would not be necessary if the LSP server would have been running the parser cycle ("kolecko") for document changes - but it does not -- see [NETBEANS-6477](https://issues.apache.org/jira/browse/NETBEANS-6477).

I needed also to move Maven MIME type registration to lower layer module - but the base Maven module already contained registrations based on that MIME type - but did not define the resolver for it.


JaroslavTulach (approved):

dbalek (approved):
Looks fine.

PR #3843, opened 2022-03-23
Invoke arbitrary action from vscode

jhorvath:
Adding possibility to invoke arbitrary action from vscode. Any action registered in `java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/layer.xml` can be called as a vscode command:
`vscode.commands.executeCommand('nbls:<Category>:<actionId>', <context>);`. Context will be converted to Action type defined in layer.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/NodeActionsProvider.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/NodeActionsProvider.java
@@ -38,9 +38,12 @@ import org.netbeans.modules.java.lsp.server.protocol.NbCodeLanguageClient;
 import org.netbeans.modules.parsing.api.ResultIterator;
 import org.openide.awt.Actions;
 import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
 import org.openide.nodes.Node;
 import org.openide.util.ContextAwareAction;
+import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
+import org.openide.util.lookup.Lookups;
 import org.openide.util.lookup.ProxyLookup;
 
 /**
@@ -61,7 +64,7 @@ public class NodeActionsProvider extends CodeActionsProvider {
     private final Set<String>  commands;
     private final Gson gson = new Gson();
 
-    private NodeActionsProvider(Set<String> commands) {
+    NodeActionsProvider(Set<String> commands) {
         this.commands = commands;
     }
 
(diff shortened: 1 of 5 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/NodeActionsProvider.java
```diff
@@ -161,4 +166,40 @@ public static NodeActionsProvider forFile(FileObject f) {
         return CompletableFuture.completedFuture(true);
     }
     
+    private CompletableFuture<Object> invokeAction(NbCodeLanguageClient client, String category, String aid, List<Object> arguments) {
+        String path = "Actions/" + category + "/" + aid.replace('.', '-') + ".instance"; //NOI18N
+        FileObject config = FileUtil.getConfigFile(path);
+        String contextType = (String) config.getAttribute("type"); //NOI18N
```
JaroslavTulach, line 172: Please write a test for this new API/behavior.

JaroslavTulach (approved):
Can you document (at least in the PR or in some `arch.xml`) how one invokes such action?

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/NodeActionsProvider.java
```diff
@@ -161,4 +166,40 @@ public static NodeActionsProvider forFile(FileObject f) {
         return CompletableFuture.completedFuture(true);
     }
     
+    CompletableFuture<Object> invokeAction(NbCodeLanguageClient client, String category, String aid, List<Object> arguments) {
+        String path = "Actions/" + category + "/" + aid.replace('.', '-') + ".instance"; //NOI18N
+        FileObject config = FileUtil.getConfigFile(path);
+        String contextType = (String) config.getAttribute("type"); //NOI18N
+        try {
+            Class<?> clazz = Thread.currentThread().getContextClassLoader().loadClass(contextType);
+            Object context = gson.fromJson(gson.toJson(arguments.get(0)), clazz);
```
sdedic, line 175: Q: this deserializes the 1st argument, which was deserialized at line ~97 as JSONObject, in an attempt to read inputs like `{ data : { id: <nodeId> } }`. Does the code that deserializes a JSONObject / id behave "well" when it is fed with a typed parameter (other node-based actions) ?

sdedic (approved):

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/NodeActionsProvider.java
jhorvath: In case the new method `invokeAction()` is called the first argument doesn't contain input you mention above (`data`) as it is not called from tree view. It is expected to contain fields for a action type to deserialise properly

PR #3844, opened 2022-03-24
Use constants defined in StandardCharsets instead of Charset name lookup.

mbien:
project wide refactoring using this [jackpot inspection](https://github.com/mbien/jackpot-inspections/blob/master/UseStandardCharsets.hint)

`Charset`s defined in `StandardCharsets` are always supported by the runtime, which means:
 - less handling of the checked`UnsupportedEncodingException`
 - potentially slightly better performance in some cases since no lookup is needed

static imports are only used in a handful of classes with many occurrences of the fields.

Each commit contains at least one cluster so that reviewers don't have to review everything (please mention what you reviewed).

going to squash everything before merge.

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/FileEncodingQueryImpl.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/FileEncodingQueryImpl.java
@@ -20,6 +20,7 @@
 package org.netbeans.modules.apisupport.project.queries;
 
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import org.netbeans.spi.queries.FileEncodingQueryImplementation;
 import org.openide.filesystems.FileObject;
 
@@ -37,7 +38,7 @@ public class FileEncodingQueryImpl extends FileEncodingQueryImplementation {
     public FileEncodingQueryImpl() {}
 
     public Charset getEncoding(FileObject file) {
-        return Charset.forName("UTF-8");
+        return StandardCharsets.UTF_8;
     }
 
 }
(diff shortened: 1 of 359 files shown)


matthiasblaesing:
First impression: Nice. The changes I saw looked sane to me and should not affect compatibility.

lkishalmi (approved):
Looks nice!

mbien:
> First impression: Nice. The changes I saw looked sane to me and should not affect compatibility.

right. All it does is to replace charset lookups with the equivalent instance from `StandardCharsets.*`

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/FileEncodingQueryImpl.java
```diff
@@ -37,7 +38,7 @@
     public FileEncodingQueryImpl() {}
 
     public Charset getEncoding(FileObject file) {
-        return Charset.forName("UTF-8");
+        return StandardCharsets.UTF_8;
```
lbownik, line 41: How about using static import?

mbien:
everything green + got one approval so far + I looked through everything again and found no anomalies as last check.

squashing the commits, planning to merge soon before we get conflicts.

apisupport/apisupport.ant/test/unit/src/org/netbeans/modules/apisupport/project/ui/wizard/common/CreatedModifiedFilesTest.java
```diff
@@ -222,7 +223,7 @@ public void testAddLookupRegistration() throws Exception {
         assertNotNull(registry);
         InputStream is = registry.getInputStream();
         try {
-            BufferedReader r = new BufferedReader(new InputStreamReader(is, "UTF-8"));
+            BufferedReader r = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8));
```
lbownik, line 226: how about try-with-resources here?

mbien: this PR does only one inspection. It would be too much at once otherwise and would make review very tiresome.

mbien:
all green again. merging.

PR #3850, opened 2022-03-24
[#3728] Only enable VerifyPartialReparse if assertions are enabled

matthiasblaesing:
It was observed, that NetBeans can create an excessive number of source
dump files in situations, where the partial reparse verifier bails out.

The VanillaPartialReparser already suppresses dumping of source files
in case of exceptions if assertions are not enabled. The same logic is
now applied to the activation of VerifyPartialReparse.
VerifyPartialReparse has no side effects and thus can be disabled if
dumping should be turned off.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/VanillaPartialReparser.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/VanillaPartialReparser.java
@@ -163,6 +163,7 @@ public class VanillaPartialReparser implements PartialReparser {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@lkishalmi @jlahoda this is an alternative to #3742 and should also fix #3728.

lkishalmi (approved):
Looks fine! Thank you!

matthiasblaesing:
If noone objects, I'll merge this on the weekend.

matthiasblaesing:
If noone objects, I'll merge this on the weekend.

lkishalmi:
Go for it!

PR #3851, opened 2022-03-24
add-open flag for archetype

ebarboni:
Attempt to pass the add open flags to archetype linked to use for https://github.com/apache/netbeans-mavenutils-archetype-nbm-archetype/pull/9

I should alter nbbuild/build.xml to populate branding so it can be used for replacement in the module

### DIFF
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/Bundle.properties
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/Bundle.properties
@@ -46,3 +46,5 @@ NbmWizardPanelVisual.cbAddModule.accessibledesc=Select to create a module projec
 NbmWizardPanelVisual.txtAddModule.accessiblename=module project name
 NbmWizardPanelVisual.txtAddModule.accessibledesc=Enter module project name.
 TXT_ProjectXmlFileElementValueRename=Update Element: {0} Value: {1}
+NbmWizardIterator.AddOpenFlag=@@metabuild.jms-flags@@
+NbmWizardIterator.AddOpenFFlag=@@metabuild.jms-fflags@@
--- a/nbbuild/build.xml
+++ b/nbbuild/build.xml
@@ -469,42 +469,6 @@ metabuild.hash=${metabuild.hash}</echo>
     </unzip>
   </target>
 
-  <target name="-load-module-flags">
-      <loadfile property="jms-base.flags" srcFile="jms-config/base.flags">
-          <filterchain>
-              <tokenfilter>
-                  <replaceregex pattern="--" replace=" -J--" />
-              </tokenfilter>
-              <striplinebreaks/>
-          </filterchain>
-      </loadfile>
-      <loadfile property="jms-desktop.flags" srcFile="jms-config/desktop.flags">
-          <filterchain>
-              <tokenfilter>
-                  <replaceregex pattern="--" replace=" -J--" />
-              </tokenfilter>
-              <striplinebreaks/>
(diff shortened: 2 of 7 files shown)


nbbuild/build.xml
```diff
@@ -138,6 +138,7 @@
     <property file="${nb_all}/nbbuild/build/netbeansrelease.properties"/>
     <!-- javadoc content filtering -->
     <setjavadoc branch="${metabuild.branch}"/>
+    <antcall target="-load-module-flags"/>
```
mbien, line 141: does this work? I thought antcall would run in its own context and the property would be empty after it. (I am a bit rusty with ant)

mbien (approved):
LGTM + tested it and it works great. good job.

nbbuild/build.xml
ebarboni: it works but the -load module append the file I guess that replacement reread the file

ebarboni:
it add flag like this '''-J--add-opens=java.base/java.net=ALL-UNNAMED -J--add-opens=java.base/java.lang.ref=ALL-UNNAMED''' maybe we should have the without -J version too.

ebarboni:
@mbien add the two properties for maven user will be easy to get both I guess even if user doesn't want them at the end. Easier to delete than to create.

ebarboni:
#3862 needs also the jvm version for flags

ebarboni:
wait PR to be merged and git checkout build.xml and default.xml. Idea of @matthiasblaesing was my first intention but did'nt proper understand where to put things.

apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/Bundle.properties
```diff
@@ -45,4 +45,6 @@ NbmWizardPanelVisual.cbAddModule.text=Create &Module Project:
 NbmWizardPanelVisual.cbAddModule.accessibledesc=Select to create a module project.
 NbmWizardPanelVisual.txtAddModule.accessiblename=module project name
 NbmWizardPanelVisual.txtAddModule.accessibledesc=Enter module project name.
-TXT_ProjectXmlFileElementValueRename=Update Element: {0} Value: {1}
\ No newline at end of file
+TXT_ProjectXmlFileElementValueRename=Update Element: {0} Value: {1}
+NbmWizardIterator.AddOpenFlag=@@metabuild.jms-flags@@
+NbmWizardIterator.AddOpenFFlag=@@metabuild.jms-fflags@@
```
mbien, line 50: do we really need two properties? Is there nothing which could add the `-J` for the mavenproject.conf?

I assume the updated archetype will insert the properties to surefire configuration/argline etc?

mbien: just a thought: would it be possible to make the launcher smarter so that it could use something like:
-J'-flag1 -flag2 -flag3' or -JVM_FLAGS='-flag1 -flag2 -flag3'?

ebarboni: It's a more heavy change, for windows it would be there https://github.com/apache/netbeans-native-launchers unsure on how it works on mac

neilcsmith-net:
@ebarboni status of this one?

mbien:
maybe we should just add a comment to the maven template (which links to our JMS file) and call it a day, since this ended up more difficult than anticipated :)

(i really don't like the fact that this would need two properties, once with -J and once without)

PR #3860, opened 2022-03-25
Multi release jar support.

dbalek:
Added test for multi release jar support.


JaroslavTulach (approved):

JaroslavTulach:
The test is passing. Does that mean NetBeans already supports multi release JARs?

jlahoda:
> The test is passing. Does that mean NetBeans already supports multi release JARs?

Did you try to run the test on both JDK 8 and JDK 11? Because, obviously, for a multi-release jar test, one must run on multiple JDKs.

Travis was not running(!) any(!) platform tests on anything newer that JDK 8(!!), so I've added at least `o.n.bootstrap` on JDK 11. I wonder what this says about NetBeans platform support for JDK 11, 17, ... .

JaroslavTulach:
```
Test org.netbeans.JarClassLoaderTest FAILED
Test org.netbeans.nbexec.NbExecPassesCorrectlyQuotedArgsTest FAILED
```

oyarzun:
@dbalek I took a stab at implementing muti-release support. https://github.com/oyarzun/netbeans/commit/c879b33c2e3296c33a917a75a4d89d69cfb80675

oyarzun:
@dbalek @JaroslavTulach any thoughts on getting this into NB15?

platform/o.n.bootstrap/test/unit/src/org/netbeans/JarClassLoaderTest.java
```diff
@@ -414,4 +425,65 @@ public static void initialize(String path, Semaphore sync) {
...
+        File jar = new File(getWorkDir(), "multi-release.jar");
+        try (OutputStream out = new FileOutputStream(jar)) {
+            TestFileUtils.writeZipFile(out, jarContent);
+        }
+        JarClassLoader jcl = new JarClassLoader(Arrays.asList(jar), new ProxyClassLoader[0]);
+        Class<?> api = jcl.loadClass("api.API");
+        Method run = api.getDeclaredMethod("run");
+        String output = (String) run.invoke(null);
+        String expected;
+        try {
```
lbownik, line 465: How about estracting this try-catch into a private method?

platform/o.n.bootstrap/test/unit/src/org/netbeans/JarClassLoaderTest.java
```diff
@@ -414,4 +425,65 @@ public static void initialize(String path, Semaphore sync) {
...
+                      jarContent.put(classesPath.relativize(p).toString(), TestFileUtils.readFileBin(p.toFile()));
+                  } catch (IOException ex) {
+                      throw new IllegalStateException(ex);
+                  }
+             });
+        File jar = new File(getWorkDir(), "multi-release.jar");
+        try (OutputStream out = new FileOutputStream(jar)) {
+            TestFileUtils.writeZipFile(out, jarContent);
+        }
+        JarClassLoader jcl = new JarClassLoader(Arrays.asList(jar), new ProxyClassLoader[0]);
```
lbownik, line 460: It is difficult to find where GIVEN block ends and WHEN block  begins (as in GIVEN-WEN-THEN). A blank line would be helpful here.

platform/o.n.bootstrap/test/unit/src/org/netbeans/JarClassLoaderTest.java
```diff
@@ -414,4 +425,65 @@ public static void initialize(String path, Semaphore sync) {
...
+        Method run = api.getDeclaredMethod("run");
+        String output = (String) run.invoke(null);
+        String expected;
+        try {
+            Class.forName("java.lang.Runtime$Version");
+            expected = "9";
+        } catch (ClassNotFoundException ex) {
+            expected = "base";
+        }
+        assertEquals(expected, output);
```
lbownik, line 471: a blank line before would be helpfull here

neilcsmith-net:
@dbalek when (re)targetting delivery, please add your reasoning, thanks!

dbalek:
> @dbalek when (re)targetting delivery, please add your reasoning, thanks!

OK - target milestone changed back to NB16.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -272,6 +289,12 @@ public Map<String, Attributes> getEntries() {
                     }
                 }
                 Manifest man = new DelayedManifest();
+                
+                src.setMultiRelease(man.getMainAttributes().containsKey(MULTI_RELEASE));
```
matthiasblaesing, line 293: The value of the attribute needs to be checked, that it equals case-insensitive to `true`

matthiasblaesing (commented):
Looks sane to me. I left one inline comment. The "Multi-Release" is a boolean attribute, it could be set to false, so that should be caught (yes it does not make sense to add that attribute and set it to false, but tooling could do this).

neilcsmith-net:
@dbalek that wasn't a request to move it back to NB16. I just wanted to understand the reasoning, pros and cons for getting this into NB15.  This will affect library loading I believe?  So do we know which dependencies we have are MRJ and therefore might behave differently?  Are there related issues this will fix / affect besides the obvious one in the platform?

cc @oyarzun

oyarzun:
@neilcsmith-net @dbalek Since this changes the class loader, I vote +1 for targeting NB16 to have a longer period of testing. As far as I know the only issue it fixes is allowing the platform to use multi-release jars on JDK9+. (e.g. log4j which came up on the dev mailing list.)

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
oyarzun, line 293:
```suggestion
                if (man.getMainAttributes().containsKey(MULTI_RELEASE)) {
                   String multiRelease = (String) man.getMainAttributes().get(MULTI_RELEASE);
                   src.setMultiRelease(multiRelease.equalsIgnoreCase("true"));
                }
```

neilcsmith-net:
Thanks @oyarzun - I'd missed that reference on list.  We also have MRJs in the IDE.

If my bash-fu is correct, then running this over NB15-rc1

```bash
for i in `find . -name "*.jar"`; do jar tvf $i | grep META-INF/versions/ > /dev/null ; if [ $? == 0 ]; then echo "$i"; fi; done
```

gives us the following files in the IDE that _might_ start behaving differently (well, correctly!) _in some fashion_?

```
./enterprise/modules/ext/metro/webservices-tools.jar
./enterprise/modules/ext/metro/webservices-api.jar
./enterprise/modules/ext/metro/webservices-rt.jar
./enterprise/modules/ext/jersey2/jersey-common-2.34.jar
./enterprise/modules/ext/jersey2/jakarta.xml.bind-api-2.3.3.jar
./java/modules/ext/byte-buddy-1.10.6.jar
./java/modules/ext/maven/lucene-core-8.11.1.jar
./java/modules/ext/jaxws22/jaxws-rt.jar
./java/modules/ext/jaxws22/jaxws-tools.jar
./java/modules/ext/nb-javac-jdk-18-api.jar
./platform/modules/ext/flatlaf-2.4.jar
./webcommon/modules/ext/js-20.3.0.jar
./ide/modules/com-jcraft-jsch.jar
./ide/modules/bcpg.jar
./ide/modules/bcprov.jar
./ide/modules/bcpkix.jar
./ide/modules/bcutil.jar
./ide/modules/ext/truffle-api-20.3.0.jar
./ide/modules/ext/jaxb/jaxb-impl.jar
./ide/modules/ext/jaxb/api/jaxb-api.jar
./ide/modules/ext/jaxb/jaxb-xjc.jar
./ide/modules/ext/junixsocket-common-2.4.0.jar
./ide/modules/ext/junixsocket-native-common-2.4.0.jar
```

matthiasblaesing:
I just tried to use MR support to see if this fixes #4393. JSch bundles a class that uses the UnixDomainSocket support in JDK 16+, but it requires Multi-Release support. So I thought this PR should fix the issue, turns out, that I was wrong. JSch is loaded by OSGI support (netbinox), which does not support MR jars.

neilcsmith-net:
@mbien just realised that one of the JARs listed above that has a lot of MRJ affected classes is lucene.  Anything of concern?  Possible source of any current issues?

mbien:
> @mbien just realised that one of the JARs listed above that has a lot of MRJ affected classes is lucene. Anything of concern? Possible source of any current issues?

@neilcsmith-net i looked into the jar and it has a version 9 in it. And the default is 8. Diffing with the decompiler showed that lucene is doing MRJ to get some extra APIs from `java.lang.Objects` and similar earlier. If thats all they are using it for, it wouldn't even matter what version is loaded on 11+, since they provided their own "FutureObjects" impl for 8 obviously (looks like unnecessary complexity to me tbh, but I haven't diffed everything).

They don't track the lucene 8.x branch on github anymore, we are using an EOL version and lucene 9+ moved to JDK 11+. We patched a single file
https://github.com/mbien/netbeans/blob/ec298fb6fae9185dfd2716054e124ec37647fd31/java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java in #3558, this file has only one version in lucene-core-8.11.1.jar - nothing can go wrong there either.

I assume you are asking because the "default" jar was loaded before, but after the change the "correct version" will be picked, which would be 9 in lucene's case?

I don't think this will cause problems.

(also holy cow I just edited your post instead of commenting in my own, i need a coffee haha)

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -77,6 +78,22 @@
...
+        int version;
+        try {
+                Object runtimeVersion = Runtime.class.getMethod("version").invoke(null);
+                version = (int) runtimeVersion.getClass().getMethod("major").invoke(runtimeVersion);
+        }
+        catch (Throwable ex) {
+                version = BASE_VERSION;
+        }
+        RUNTIME_VERSION = version;
+    }
```
mbien, line 95: looks good to me but fix indentation please. It would be also good to use default NB formatting for brace positioning.

The rest of the file does *not* use this formatting

platform/o.n.bootstrap/test/unit/src/org/netbeans/JarClassLoaderTest.java
mbien: have to agree here. Some empty lines which keep logical parts separate, like paragraphs in text, do a lot for readability.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -574,6 +606,17 @@ protected URL doGetResource(String name) throws IOException  {
...
+                {
+                    int ver = RUNTIME_VERSION;
+                    while (ver > BASE_VERSION) {
+                        byte[] data = archive.getData(this, "META-INF/versions/" + ver + "/" + path);
+                        if (data != null) {
+                            return data;
+                        }
+                        ver--;
+                    }
+                }
```
mbien, line 619: This might scale badly. Lets say we load lucene which has a 8 and a 9 version and run on JDK 19.

This will call `getData()` (which isn't trivial) 10 times for each class before it finds something. I would put all versions contained in `META-INF/versions/` into an array during init and go through the array backwards. This would produce hits with fewer attempts on average.

mbien (commented):
one performance concern which would require more investigation IMO and two style nitpicks.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
dbalek: Fixed.

dbalek: Done.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
dbalek: Indentation fixed.

matthiasblaesing (approved):

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
mbien: looking at the JDK impl

https://github.com/openjdk/jdk/blob/0bcf17674e1c8c07ede3aef6449e0631e526ba86/src/java.base/share/classes/java/util/jar/JarFile.java#L599-L623

JarFile lists all versions first and iterates only through them, which doesn't scale negatively with the runtime version.

dbalek: Good point. Fixed.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -574,13 +609,81 @@ protected URL doGetResource(String name) throws IOException  {
...
         }
-        
+
+        private int[] getVersions() {
+            int[] ret;
+            if (versions != null && (ret = versions.get()) != null) {
+                return ret;
+            }
+            try {
+                JarFile src = getJarFile("versions");
```
mbien, line 643: is "versions" correct here? Shouldn't this be a path within the jar?

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -574,13 +609,81 @@ protected URL doGetResource(String name) throws IOException  {
...
+                Enumeration<JarEntry> en = src.entries();
+                while (en.hasMoreElements()) {
+                    JarEntry je = en.nextElement();
+                    if (je.isDirectory()) {
+                        String itm = je.getName();
+                        if (itm.startsWith("META-INF/versions/")) {
+                            String res = itm.substring(18);
+                            int idx = res.indexOf('/');
+                            if (idx > 0 && idx == res.length() - 1) {
+                                vers.add(Integer.parseInt(res.substring(0, idx)));
```
mbien, line 654: would recommend to catch and ignore `NumberFormatException` in case something unknown shows up (defensive forward compatibility).

i converted this section to streams to check if it improves readability, I let you decide what you prefere:
```java
    int[] ret = getJarFile(path).stream()
        .filter(JarEntry::isDirectory)
        .filter((e) -> e.getName().startsWith("META-INF/versions/"))
        .mapToInt((e) -> {
            String res = e.getName().substring(18);
            int idx = res.indexOf('/');
            if (idx > 0 && idx == res.length() - 1) {
                try {
                    return Integer.parseInt(res.substring(0, idx));
                } catch (NumberFormatException ignore) {}
            }
            return -1;
        })
        .sorted().distinct().filter((v) -> v != -1).toArray();
```

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
oyarzun: Agreed looks like "versions" is just the name of the jar in the test case.

matthiasblaesing:
@dbalek could you please check whether this can be merged now? I looked at the OSGI side and I thought this was already in, we are early in the NB16 cycle, so now would be a good time.

matthiasblaesing:
There is at least one remaining problem:

> The command "hide-logs.sh ant $OPTS -f platform/core.startup test" exited with 1.

matthiasblaesing:
Changes are merged via #4589

PR #3866, opened 2022-03-25
[NETBEANS-6329] Tools > Show In Finder don't works on JDK 17

dyorgio:
Include exports to can execute macOs 'Show In Finder' action in jdk17.

### DIFF
--- a/nbbuild/jms-config/desktop.flags
+++ b/nbbuild/jms-config/desktop.flags
@@ -14,3 +14,4 @@
 --add-exports=java.desktop/sun.swing=ALL-UNNAMED
 --add-exports=java.desktop/sun.awt.im=ALL-UNNAMED
 --add-exports=java.desktop/com.sun.java.swing.plaf.motif=ALL-UNNAMED
+--add-exports=java.desktop/com.apple.eio=ALL-UNNAMED


matthiasblaesing:
Looks sane to me. The affected class is `com.apple.eio.FileManager` called from `org.netbeans.modules.applemenu.ShowInFinder`. It would be nice though to investigate if using `java.awt.Desktop` would be an option (the `browse` method in that class does the right thing (at least looks like it) on linux.

dyorgio:
Hi @matthiasblaesing, yes it does, but only if you call getParentFile:
```java
// Current Implementation
final Class<?> fmClz = Class.forName("com.apple.eio.FileManager");      //NOI18N
final Method m = fmClz.getDeclaredMethod("revealInFinder", File.class); //NOI18N
m.invoke(null, new File("/PATH/FILE.EXT"));

// Suggested Implementation
Desktop.getDesktop().browse(new File("/PATH/FILE.EXT").getParentFile().toURI());
```

But I only tested on macOS 12.3 (latest), with 1.8 and 17 java versions.

matthiasblaesing:
Unittests are green - @dyorgio I suggest to merge this as is. It is low risk and fixes a problem, so it is an improvement in anycase.

Independent of that it would be good to get rid of the add-exports if possible, so if you would follow up with the suggested implementation.

Please indicate if you agree that I merge now.

dyorgio:
Yes I also prefer the original solution.
We can guarantee that another versions of macOS will work with Desktop.browser, I suspect that it is the reason to use platform specific implementation.

PR #3868, opened 2022-03-25
SQL History loss: Reimplement atomatic saving by writing to tempfile and moving to target

matthiasblaesing:
It was observed, that sometimes the sql history was lost and it is
assumed, that writing the new history began (emptying the file as the
first step) and was not completed. By using a temporary file, the file
is either written completely or not. It is assumed, that a move
operation inside the same directory is atomic on all OSes.

### DIFF
--- a/ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
+++ b/ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
@@ -22,6 +22,9 @@ import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.xml.parsers.DocumentBuilder;
@@ -32,7 +35,6 @@ import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 import org.netbeans.modules.db.sql.execute.ui.SQLHistoryPanel;
 import org.openide.filesystems.FileObject;
-import org.openide.filesystems.FileSystem;
 import org.openide.filesystems.FileUtil;
 import org.openide.util.NbPreferences;
 import org.openide.util.RequestProcessor;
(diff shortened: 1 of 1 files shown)


mbien (approved):
LGTM

if loadHistory() fails, it would also result in an empty history and the next save would override it on disk. Might be worth to bump the log level to WARN there?

ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
```diff
@@ -205,40 +209,34 @@ void removePropertyChangeListener(PropertyChangeListener listener) {
...
+                        xsw.writeCharacters(sqe.getSql());
+                        xsw.writeEndElement();
+                        xsw.writeCharacters(CONTENT_NEWLINE);
                     }
-                });
+                    xsw.writeEndElement();
+                    xsw.flush();
+                    xsw.close();
+                } catch (IOException | XMLStreamException ex) {
+                    LOGGER.log(Level.INFO, ex.getMessage(), ex);
```
vieiro, line 236: If we catch an exception here, are we sure we want to invoke `Files.move` (below) immediately after? 

Or do we want to add a `return` statement after we log the exception so no files (with potential errors) are `Files.move`d?

ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
```diff
@@ -97,7 +101,7 @@ public int getListSize() {
         return NbPreferences.forModule(SQLHistoryPanel.class).getInt("OPT_SQL_STATEMENTS_SAVED_FOR_HISTORY", DEFAULT_SQL_STATEMENTS_SAVED_FOR_HISTORY);
     }
 
-    protected FileObject getHistoryRoot(boolean create) throws IOException {
+    private FileObject getHistoryRoot(boolean create) throws IOException {
```
vieiro, line 104: `FileUtil.createFolder` [accepts names with forward slashes](https://bits.netbeans.org/dev/javadoc/org-openide-filesystems/org/openide/filesystems/FileUtil.html#createFolder-org.openide.filesystems.FileObject-java.lang.String-) so we're good with that `"Databases/SQLHISTORY"` string.

ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
matthiasblaesing: Valid point - indeed this catch is invalid and should not be there. I pushed an update (new commit, I intent to squash before merge).

ide/db.core/src/org/netbeans/modules/db/sql/history/SQLHistoryManager.java
matthiasblaesing: If I remember correctly there is exactly one platform, that does not use "/" as separator, that is windows and I know, that windows also accepts forward slashes. I have some hope that in the future people learn from past mistakes and don't try to reinvent the wheel and either use URIs or at least use these as base.

matthiasblaesing:
Based on the comments from @mbien and @vieiro I pushed an update. Thank you both.

vieiro (approved):
+1 Looks cleaner to me now. 
Good also to clean on the catch (the try will move the temporary file, so no need to delete it.).

vieiro:
Mmm... there're some test errors you may want to check out, @matthiasblaesing . You may want to recover those `protected` methods.

PR #3869, opened 2022-03-25
Fix KeyRing issues

matthiasblaesing:


matthiasblaesing:
Both fixed issues were verified (one tested by myself, the second by the reporter), unittests are green I did a self-review.

PR #3873, opened 2022-03-26
Save cause of download error for BuildException

jgneff:
When downloading external binaries, save any `IOException` for the `cause` parameter of the `BuildException`. Otherwise, the root cause of the build failure is always `null`.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/DownloadBinaries.java
@@ -312,7 +312,8 @@ public class DownloadBinaries extends Task {
                 log("Trying: " + url, Project.MSG_VERBOSE);
                 return downloadFromServer(this, url);
             } catch (IOException ex) {
-                //Try the next URL
+                // Saves the exception and tries the next URL, if present
+                firstProblem = ex;
             }
         }
         throw new BuildException("Could not download " + cacheName + " from " + server + ": " + firstProblem, firstProblem, getLocation());


jgneff:
When the build fails to download an external binary because of an `IOException`, the root cause of the failure is always `null`. This pull request saves the exception so that it can be reported in the build error message.

### Before

Before this change, the error is reported with a `null` root cause as shown below:

```
[downloadbinaries] Could not download
EC27...6746-wordlist-en_GB-large-2017.08.24.zip to
/root/.hgexternalcache/EC27...6746-wordlist-en_GB-large-2017.08.24.zip:
/build/.../parts/netbeans/build/nbbuild/build.xml:276:
  Could not download
  EC27...6746-wordlist-en_GB-large-2017.08.24.zip from
  https://netbeans.osuosl.org/binaries/: null
```

### After

After this change, the original `IOException` is printed as the root cause:

```
[downloadbinaries] Could not download
EC27...6746-wordlist-en_GB-large-2017.08.24.zip to
/root/.hgexternalcache/EC27...6746-wordlist-en_GB-large-2017.08.24.zip:
/build/.../parts/netbeans/build/nbbuild/build.xml:276:
  Could not download
  EC27...6746-wordlist-en_GB-large-2017.08.24.zip from
  https://netbeans.osuosl.org/binaries/: java.io.IOException:
    Skipping download from
    https://netbeans.osuosl.org/binaries/EC27...6746-wordlist-en_GB-large-2017.08.24.zip
    due to response code 503
```

Now I see that the HTTPS proxy server in my build environment is returning a response status code of [503 Service Unavailable][1] for this file:

> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay.

### Tests

I ran the unit tests for the NetBeans Ant task `downloadbinaries` as follows:

```console
$ ant -Dincludes=**/DownloadBinariesTest.class localtest
  ...
localtest:
    [junit] Testsuite: org.netbeans.nbbuild.extlibs.DownloadBinariesTest
    [junit] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.162 sec
    [junit]

BUILD SUCCESSFUL
Total time: 1 second
```

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.4

vieiro (approved):
Thanks, @jgneff !

matthiasblaesing:
Thank you - looks good, unittests are green. Lets get this in.

PR #3875, opened 2022-03-26
Order Gradle sub-projects Nodes

### LINKED ISSUE #3861: Netbeans (sub-)projects tree list sub-projects in
### Description

* I recommend we list Gradle subprojects tree (or sub-sub-projects) in Alphabetical order of sub-project name
* The Gradle subprojects tree is in some random order as far as I can tell.
   * A picture is worth 1,000 words ...
     image
   * The current arrangement makes it very difficult find a specific project, especially things that belong together.
   * The order can change too between one session and the next Netbeans session
   * It like hunt and peck all over again.
* I do not have Ant or Maven projects to compare with, but I wonded if they are also in some strange order

### Use case/motivation

* Find and open sub-projects faster and more effectively
* I am looking for a specific sub-project (say TestSupport) well, it is not after "S" it might be anywhere
* With TestSupport looking for "Test..." is as fast as it might be because 
   *  3 or 4 sub-projects may be available in
   * they can be in  completely different locations on the tree
*  Want projects of similar nomencluature say TestSupport, TestingData, TestUi, etc.
    * These are named so to keep them together and then Netbeans project tree just shuffles the up
*  This implementaton just slows me / others down scanning the list

### Related issues

*  Consider issue `org.gradle.java.home` issues:
    * #3858
    * #3857
*  One of the major reasons within the GUI to need to reopen subprojects is to ensure that thissetting is correct for our project
   * I think it would be wonderfull to be able to `right-click` on a sub-project in the Gradle sub-projects tree and select from the same  'project' actions menu to change something like an attirbute/property without needing to actually re-open the project ...
      *  That is not to say that every opeion be available on closed a sub-project; some may well be greyed-out
   * After all we are talking about settings / They ought not need a project open to be altered
   * Setting changes should persist -- Not as "cache" files
*  

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow the Apache Software Foundation's [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html)

lkishalmi:
Fix #3861. Let's keep the Gradle Sub-Project nodes in order.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/nodes/SubProjectsNode.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/nodes/SubProjectsNode.java
@@ -26,8 +26,10 @@ import java.awt.Image;
 import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -50,10 +52,13 @@ import static org.netbeans.modules.gradle.nodes.Bundle.*;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.prefs.PreferenceChangeListener;
 import org.netbeans.api.project.ProjectUtils;
 import org.netbeans.modules.gradle.api.GradleBaseProject;
+import org.netbeans.modules.gradle.spi.GradleSettings;
 import org.netbeans.modules.gradle.spi.Utils;
 import org.openide.nodes.Children;
+import org.openide.util.WeakListeners;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


extide/gradle/src/org/netbeans/modules/gradle/nodes/SubProjectsNode.java
```diff
@@ -108,26 +113,39 @@ public Image getOpenedIcon(int type) {
...
             project = proj;
-            listener = (PropertyChangeEvent evt) -> {
+            propListener = (PropertyChangeEvent evt) -> {
                 if (NbGradleProject.PROP_PROJECT_INFO.equals(evt.getPropertyName())) {
                     ProjectManager.getDefault().clearNonProjectCache();
                     refresh(false);
                 }
             };
-            NbGradleProject.addPropertyChangeListener(project, listener);
+            NbGradleProject.addPropertyChangeListener(project, WeakListeners.propertyChange(propListener, proj));
```
sdedic, line 127: good catch ! Allows the nodes to GC.

sdedic (approved):

Issue #3678, opened 2022-02-28
Creating new application (Java With Maven, Java Application) has main class built ignoring templates


dgreen:
### Apache NetBeans version

Apache NetBeans 13 release candidate

### What happened

Creating a new Java with Maven, Java Application generates a Java file with a main.  The file is generated ignoring the available template for Java Main Class.

Somewhat unrelated to this issue, clicking on the nbfs: links did not cause any noticeable activity.

### How to reproduce

1. File | New Project 
2. In popup, choose Java with Maven then Java Application
3. Hit finish on next pop up
4. Generated project has a MavenProject1.java file with a main

It has 

`
/*
 * Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
 * Click nbfs://nbhost/SystemFileSystem/Templates/Project/Maven2/JavaApp/src/main/java/${packagePath}/${mainClassName}.java to edit this template
 */

package edu.uab.mavenproject1;

/**
 *
 * @author name <email>
 */
public class Mavenproject1 {

    public static void main(String[] args) {
        System.out.println("Hello World!");
    }
}
`

There is an existing template viewable using Tools | Templates , select Java, then Java Main Class, open in editor

`
/*
 * Author: ${user}
 * Assignment:  ${project.displayName} - ${course}
 *
 * Credits:  (if any for sections of code)
 */

<#if package?? && package != "">
package ${package};

</#if>
<#if superclass?? && superclass != "" && !superclass?starts_with("java.lang") && !superclass?matches(package + "\\.\\w+")>
import ${superclass};

</#if>
<#if interfaces?? && interfaces != "">
    <#list "${interfaces}"?split("\n") as interface>
        <#if interface?? && interface != "" && !interface?starts_with("java.lang") && !interface?matches(package + "\\.\\w+")>
import ${interface};
        </#if>
    </#list>

</#if>
/**
 *
 */
<#if superclass?? && superclass != "">
    <#assign extension = "${superclass}"[("${superclass}"?last_index_of(".") + 1)..]>
</#if>
<#if interfaces?? && interfaces != "">
    <#assign implementation = "">
    <#list "${interfaces}"?split("\n") as interface>
        <#assign implementation += "${interface}"[("${interface}"?last_index_of(".") + 1)..] + ", ">
    </#list>
    <#assign implementation = "${implementation}"?remove_ending(", ")>
</#if>
[...]

PR #3876, opened 2022-03-27
removed some less useful template links and other minor updates

mbien:
proposal to remove some of the in-editor links to their template files.

Some templates, specifically in the code snippet category, are unlikely to be ever edited by the user - except for the lone purpose of removing the comment with the link.

example:
```java
        List.of(1,2,3).sort((o1, o2) -> {
            return 0; // Generated from nbfs://nbhost/SystemFileSystem/Templates/Classes/Code/LambdaBody
        });
```
It is unlikely that the user will ever want to change the template for a lambda body, since there is simply not a lot to change. I think a better default would be to not have a link, the templates can be still found via Tools -> Templates.

links are removed from: GeneratedMethodBody, LambdaBody and OverriddenMethodBody

edit: made some other minor improvements, see commit msg.

I found also a bug in the URL highlighter, it adds the ')' to the url which creates an invalid link. Does anyone know where the code for that is?


matthiasblaesing (commented):
I think it is valid to remove the links here. The templates are indeed mostly place holders, with little incentive for customization. 

If the VSCODE and IDE variants of the templates are now identical I think it would be good to remove the duplication.

mbien:
> If the VSCODE and IDE variants of the templates are now identical I think it would be good to remove the duplication.

I don't want to change too much there since I don't know the requirements. I also don't understand why for example a maven project has its own new project template instead of just using the template for "new java main class" as it used to be the case if I remember correctly. (this added new bugs and confusion as seen on the mailing list or issue #3678)

this reminds me: going to try to update the txt on the other templates to mention that users have to hold down ctrl while clicking since that was also brought up on the ml.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
```diff
@@ -223,7 +222,7 @@ public void test223011b() throws Exception {
                         "        Field f = method(c1, c2);\n" +
                         "    }\n" +
                         "    private <T extends Number & CharSequence, E extends Integer> Field method(Class<E> c1, Class<T> c2) {\n" +
-                        "        throw new UnsupportedOperationException(\"Not supported yet.\"); // Generated from nbfs://nbhost/SystemFileSystem/Templates/Classes/Code/GeneratedMethodBody\n" +
```
JaroslavTulach, line 226: This is a useful link that I used as I am not satisfied with `new UnsupportedOperationException(\"Not supported yet.\")`.

ide/projectui/src/org/netbeans/modules/project/ui/resources/license-default.txt
```diff
@@ -1,4 +1,4 @@
 ${licenseFirst!""}
-${licensePrefix}Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
-${licensePrefix}Click nbfs://nbhost/SystemFileSystem/${.data_model["org.openide.filesystems.FileObject"].getPath()} to edit this template
+${licensePrefix}Ctrl/Cmd+Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
+${licensePrefix}Ctrl/Cmd+Click nbfs://nbhost/SystemFileSystem/${.data_model["org.openide.filesystems.FileObject"].getPath()} to edit this template
```
JaroslavTulach, line 3: `Ctrl/Cmd` sounds horrible. I don't have Cmd on my keyboard! Would `See nbfs:/....` work?

JaroslavTulach (commented):
Until NetBeans continues to provide template modifying functionality, it is important to promote it to the user. Removing the links goes against that effort. I'd rather not see this PR integrated.

ide/projectui/src/org/netbeans/modules/project/ui/resources/license-default.txt
mbien: the problem was that many didn't know how to click the link as seen in issues and on the mailing list. I don't have Cmd either. If I wouldn't have added it someone would have commented that they had no Ctrl on their kb.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
mbien: its also editable via Tools -> Templates -> Java/Code Snippets. It should be picked up during a NB version upgrade too as far as i know, so you have edit it only once.

ide/projectui/src/org/netbeans/modules/project/ui/resources/license-default.txt
JaroslavTulach: I have noticed the complaint on mailing list as well. Yes, we have a usability problem. 

> I don't have Cmd either. If I wouldn't have added it someone would have commented that they had no Ctrl on their kb.

Probably a sign that the user experience deserves deeper fix than `Ctrl/Cmd` prefix.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
JaroslavTulach: I have always known the template is in Tools -> Templates. But I never bothered searching for it! Having the link/action/hint next to the generated text block at the time the incorrect (as perceived by user) block is inserted is essential for people to really invoke the customization.

ide/projectui/src/org/netbeans/modules/project/ui/resources/license-default.txt
mbien: It could be a sign. But maybe its sufficient to explain how the link works at least until someone is interested to implement a deeper fix.

JaroslavTulach: I'd like to point out that users not knowing how to "click" in the editor (e.g. with Ctrl or Cmd) probably don't know how to navigate to symbol definitions either!  _"Addressing"_ just the URL problem without fixing the underlying usability issue makes little sense to me.

I am fine with some rewording - for example `Visit nbfs:...` or `Navigate to nbfs:` - and then letting the user intelligence work - but I find `Ctrl/Cmd+Click` so ugly, that I cannot adore it... even if it helped even a single user.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
JaroslavTulach: Moreover the `Tools -> Templates` UI isn't in VSCode. Clicking the URL is the only way to customize the template in VSCode.

neilcsmith-net: @JaroslavTulach what happens if the user deletes the link in the template then?!

ide/projectui/src/org/netbeans/modules/project/ui/resources/license-default.txt
neilcsmith-net: +1 to `Visit nbfs:...`, and possibly thinking about better ways of guiding the user into how to click things separately.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
mbien: > Moreover the `Tools -> Templates` UI isn't in VSCode. Clicking the URL is the only way to customize the template in VSCode.

now I understand the problem. VSCode is missing a feature.

sdedic: @mbien not entirely. The text suggesting that a generated snippet can be customized has been there for a long time. Especially novice users reminded that they can adjust the IDE to his/her own style (i.e. the UnsupportedException default body is seen as good or bad, depending on coding style).

*I opt for leaving* the live links in there. If a presentation does not seem nice (?), we can enhance Folding SPI (now allows only to provide plaintext for the placeholder, could provide i.e. JComponent or TextLayout for the entire fold contents.

jtulach (changes requested):
Looks like the overall concensus is that this needs some changes:
- Use `Visit nbfs://....` as [proposed by Neil](https://github.com/apache/netbeans/pull/3876#discussion_r843557034)
- Avoid [removing of URLs from code snippets](https://github.com/apache/netbeans/pull/3876/files#r843557034)

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/CreateMethodTest.java
mbien: would be curious what templates you guys are using for the generated methods

mbien:
going to close this since I don't think there is going to be a consensus
 - Visit/Open/etc sounds good but it doesn't really solve the original issue of communicating that the user has to ctrl-click the link, I don't want to obfuscate this further with this PR
 - I am not a fan of IDE specific tags landing in generated code, this simply does not seem to be good default behavior to me, I think its ok for file/class headers, but its too much for methods and doesn't make sense for lambdas IMO. But I underestimated how popular those links are under maintainers (i would be curious what custom templates for auto-generated methods everyone is using though)
 - I might try to remove the lambda snipped with #3458 If i revisit the draft again

PR #3881, opened 2022-03-28
Fetch richer problem info from gradle daemon

sdedic:
This PR is more a preparation for possibly better error reporting and/or assistive features. I've noticed that Gradle internally reports (sometimes !) file and location of the error. Especially unknown property or syntax errors have such annotations. Then the errors are internally yet richer than reported in the console. I.e. unresolved artifact errors track which URL(s) gradle tries to download the artifact etc.

This PR changes the protocol between Gradle Tooling Plugin and the NetBeans Gradle Projects module - exceptions are now sent including the file/line annotations and with the complete exception chain. The IDE then formats the output appropriately.

String "problems" were replaced by `GradleReport` structures, they are copied from the wire Report structures. ProjectProblemProvider was updated appropriately, as well as gradle caches (reports are saved/cached).

Then I converted `GradleReport` to editor hints. Since the Gradle does not use Parsing API to analyze the text, delivering the hints to the editor was somewhat tricky: once the first Gradle project activates, a listener is hooked to EditorRegistry and annotations are refreshed (or obtained) as soon as an editor became active. This code should not load at all if Gradle projects are not used in the IDE session. No fixes are offered for the hints - this is a big room for improvement still.

Outline of changes:
- bugfix: Gradle regsitered PlainEditorKit for its files, which is essentially a NbEditorKit; internals were ripped a long time ago (and PlainEditorKit extends NbEditorKir). By registering NbEditorKit, one can get eventually rid of dependency on plain text editor. The motivation was, in fact, that vsnetbeans does not distribute plain editor and hints didn't work without that change.
- feature: use `Report` or `GradleReport` to communicate exceptions from Gradle daemon to IDE
- enhancement: use richer problem chain to report project problems
- feature: convert `GradleReport` to editor error hints. 
- feature: convert `GradleReport` to LSP Diagnostics for LSP clients.


lkishalmi:
Really nice piece of work! Thank you!
My only question is that problems and reports sometimes used in parallel. Would it be better to use the new Report only? Or we would kick in some backward API compatibility issues?

sdedic:
> Really nice piece of work! Thank you! My only question is that problems and reports sometimes used in parallel. Would it be better to use the new Report only? Or we would kick in some backward API compatibility issues?

I didn't want to trash simple string-based "problems", as they are easy to use; but in fact, only `NbProjectInfoBuilder:352` seems to use the String-based `problem` now. 

re. API - the changed API is actually between Gradle Projects Plugin and NB Tooling Gradle Plugin ... not sure if that's API in the NetBeans compatibility sense, as these two are released always as one binary and noone except Gradle Projects module should use the NB Tooling, right ? @JaroslavTulach  ?

lkishalmi (approved):
Looked through once again. This would be fine.

About the API between the gradle-tooling and the module. I regard that as a private API, though I've made a mistake to put the NbProjectInfo into a public API package. Maybe it would be good for a later PR to move that one into an appropriate private package.

JaroslavTulach:
> that's API in the NetBeans compatibility sense

What's an API? [Everything someone else can depend on](http://wiki.apidesign.org/wiki/Determining_What_Makes_a_Good_API).

lkishalmi:
NbProjectInfo is in a public API package, but in a jar which is not exposed to other modules, so it is fine.

JaroslavTulach:
Possibly doing `new URLClassLoader("..../path/to/the/...jar").loadClass("name.of.the.Clazz")` in the `build.gradle` would be a way to _depend on presence of the class_, 

> so it is fine.

I am not saying it is not _fine_. Just that it is probably an API.

lkishalmi:
Well, that's true. IMHO, whoever were apply such a dependency, shall be rewarded by breaking that "API" in an un-announced non-compatible way.

It seems the tests are good (only the unreliable jdk8 job failed). I've restarted that. Though I think, let's have this one merged!

sdedic:
OK, so what about **documenting** these "API types" (which weren't meant to be actually exposed as API types) in NB Tooling gradle plugin as internal protocol scheduled for removal in NB 14+1 release - then refactor it to another package, so it does not clash with the real API ?

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/api/NbProjectInfo.java
```diff
@@ -41,5 +41,22 @@
      */    
     Map<String, Object> getExt();
     Set<String> getProblems();
+
+    /**
+     * @since 2.23
+     */
```
ppisl, line 47: It's API, shouldn't we add some basic JavaDoc?

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/api/NbProjectInfo.java
```diff
@@ -41,5 +41,22 @@
...
+    Set<Report> getReports();
+    
     boolean getMiscOnly();
+    
+    /**
+     * @since 2.23
+     */
+    interface Report {
+        public String getErrorClass();
+        public String getScriptLocation();
```
ppisl, line 57: It would be useful to have a JavaDoc here, without an explanation seems to be strange that a method called **getScriptLocation()** returns a String.

ppisl (approved):

lkishalmi:
Moving those two files out of the api package can be done by NB14. The sooner the better.
I can create a PR for that once this one is in.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/api/NbProjectInfo.java
sdedic: According to @lkishalmi assessment, this should not be treated as an API, but rather removed :)

PR #3897, opened 2022-03-30
Support for Micronaut configuration enhanced.

dbalek:
- Navigation to custom properties declared via @Property annotation added.

### DIFF
--- a/enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
+++ b/enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
@@ -18,13 +18,46 @@
  */
 package org.netbeans.modules.micronaut;
 
+import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.AssignmentTree;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.LiteralTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.TreePathScanner;
+import com.sun.source.util.Trees;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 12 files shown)


enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
```diff
@@ -60,6 +94,16 @@ public static boolean hasConfigMetadata(Project project) {
                     Exceptions.printStackTrace(ex);
```
lbownik, line 94: shouldn't exception's propagete instead of being swallowed?

enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
```diff
@@ -60,6 +94,16 @@ public static boolean hasConfigMetadata(Project project) {
...
                 }
             }
+            String customRepository = getCustomRepository(project);
+            if (customRepository != null) {
+                try {
+                    InputStream stream = new ByteArrayInputStream(customRepository.getBytes(StandardCharsets.UTF_8));
+                    ConfigurationMetadataRepository repository = ConfigurationMetadataRepositoryJsonBuilder.create().withJsonResource(stream).build();
+                    props.putAll(repository.getAllProperties());
+                } catch (Exception ex) {
+                    Exceptions.printStackTrace(ex);
```
lbownik, line 104: shouldn't exception's propagete instead of being swallowed?

enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
dbalek: Why? This way, the exception is printed to log, user gets immediately notified that something was wrong at the particular place, while the upper level functionality could stay (at least partially) working.

enterprise/micronaut/src/org/netbeans/modules/micronaut/MicronautConfigProperties.java
dbalek: The same as above.

dbalek: The same as above.

MartinBalin (approved):

PR #3905, opened 2022-03-31
Handle the case when the connection name is not known before wallet download

jhorvath:
In case the connection name is not known before the DB Wallet is downloaded, parse wallet's `tnsnames.ora` and take one from there.

### DIFF
--- a/enterprise/cloud.oracle/licenseinfo.xml
+++ b/enterprise/cloud.oracle/licenseinfo.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


MartinBalin (approved):

mbien:
hmm. why can't the secondary jobs find the artifact in this PR. This never happened during testing.

mbien:
@jhorvath i deleted the complete workflow run. Could you do me a favor and force push into this PR to trigger a fresh run?

jhorvath:
@mbien It is pushed already, but just 3 checks are queued.

mbien:
> It is pushed already, but just 3 checks are queued.

@jhorvath it works now like a pipeline. Two primary jobs build, rest starts later after they succeed and use the artifact of the primary build for tests.

the summary shows a nice graph https://github.com/apache/netbeans/actions/runs/2071857027

anyway. Everything is working fine now. This might just have been a weird github anomaly. At least we know how to fix it if it happens again.

dbalek (approved):

PR #3906, opened 2022-03-31
Fixing user home in project wizard and filtering outgoing messages fr…

ppisl:
This PR contains two simple fixes. 

1) The suggested folder for new java project in VSCode (new java project wizard) was taken from `user.dir` system property. This means that was something like `{user_home}/.config/Code/User/globalStorage/asf.apache-netbeans-java/userdir` . The change in `LspTemplateUI` correct this and the suggested folder should be in this case `user.home` folder. 

2) The change in `Server.java` file prevents processing outgoing messages wiht MessageConsumer that is targeted for incomming messages. As a side effects for every outgoing message was created `ProxyLookup` and `OperationContext`, which are useless for outgoing messeges.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
@@ -72,6 +72,7 @@ import org.eclipse.lsp4j.TextDocumentSyncOptions;
 import org.eclipse.lsp4j.WorkDoneProgressCancelParams;
 import org.eclipse.lsp4j.WorkDoneProgressParams;
 import org.eclipse.lsp4j.WorkspaceFolder;
+import org.eclipse.lsp4j.jsonrpc.Endpoint;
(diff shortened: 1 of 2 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/Server.java
```diff
@@ -195,6 +196,9 @@ synchronized void attachClient(NbCodeLanguageClient client) {
         public MessageConsumer attachLookup(MessageConsumer delegate) {
             // PENDING: allow for message consumer wrappers to be registered to add pre/post processing for
             // the request plus build the request's default Lookup contents.
+            if (!(delegate instanceof Endpoint)) {
```
sdedic, line 199: Thanks for catching this! I just didn't get right the idea that lsp4j sends BOTH the incoming and outgoing messages go throught the same wrapper through my head :(

sdedic (approved):

PR #3908, opened 2022-03-31
Move NbProjectInfo and ModelFetcher out of Gradle API package

lkishalmi:
Simple refactor we were talking about in #3881, that moves NbProjectInfo and ModelFetcher out of the org.netbeans.modules.gradle.api package.


neilcsmith-net:
> VSCode extension that should be built next Thursday. From that perspective, it would be much safer to merge them after next Thursday.

@ppisl aside - and freeze for NB14 is scheduled for Friday!  Will likely be after the weekend now anyway, but if master is not open for merging because of VSCode update releases, it might be time to consider a delivery branch there too?!

lkishalmi:
I have no problem to have this with NB15.

sdedic:
What about this one - can be merged (after necessary updates) ?

PR #3924, opened 2022-04-04
[3911] Filter source roots to avoids clashes between module-info files.

jlahoda:
The source path currently contains directories for all modules, and the compiler is confused and reads wrong `module-info.java`. As an attempt to fix this, lets try to filter the source path to contain only the one appropriate source root, and lets see how it works.


lkishalmi:
For me it looks good. A small enough fix to be reverted even in the NB14 cycle if anything goes south.

PR #3932, opened 2022-04-05
cancel previous CI run on repeated PR update.

mbien:
testing...
looks like it worked

I force pushed while this was running:
https://github.com/apache/netbeans/actions/runs/2097239880
which got canceled and the next run started:
https://github.com/apache/netbeans/actions/runs/2097326632

possible risks:
- the new run will remain queued while the old is being canceled, it might not be able to kill all jobs right away
- you will get a mail :)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,12 @@ on:
   push:
   pull_request:
 
+# cancel PR workflow run if PR is updated while jobs are still running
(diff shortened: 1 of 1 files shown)


vieiro (approved):
Hi @mbien , if you did the test and it worked then you should merge it, I think. We can save some trees with this! :-)

mbien:
@vieiro agreed. Lets merge this. Its small enough so that it can be easily reverted in case something happens.

PR #3940, opened 2022-04-07
[NETBEANS-5599] PHP 8.1 Support: Enumerations

junichi11:
- https://issues.apache.org/jira/browse/NETBEANS-5599
- https://wiki.php.net/rfc/enumerations

## Part 1:

- Fix the parser, the lexer, and `PHP81UnhandledError` 

nb-php81-enumerations-syntax-1

nb-php81-enumerations-syntax-2

PHP 8.1:

nb-php81-enumerations-syntax-4

PHP 8.0:

nb-php81-enumerations-syntax-php80

## Part 2:

- Fix the indexer
- Add elements for an enum and an enum case

## Part 3:

- Fix the Navigator

```php
enum UserStatus: string {

    case Pending = 'P';
    case Active = 'A';
    case Suspended = 'S';
    case CanceledByUser = 'C';

    public function label(): string {
        return match ($this) {
            static::Pending => 'Pending',
            static::Active => 'Active',
            static::Suspended => 'Suspended',
            static::CanceledByUser => 'Canceled by user',
        };
    }
}
````

nb-php81-enumerations-navigator-window

## Part 4:

- Fix the `SemanticAnalysis` and the `PHPBracesMatcher`
- Add unit tests for the SemanticAnalysis, the FoldingScanner, and the PHPBrace Matcher

## Part 5:

- Fix the parser (e.g. `Name::CASE1->method()`, `$i::CASE1->method()`, `Name:::CASE1::staicMethod()`, `Name::CASE1::CONSTANT`)

nb-php81-enumerations-syntax-3

## Part 6:

- Fix the formatter

## NOTE:

CC, GotoDeclaration, MarkOccurrences, Hints etc. are not fixed yet.

### CODE REVIEW

**Comment by junichi11** [comment]
@tmysik Could you please have a look at this when you have time? Sorry for the huge changes...
I've added unit tests as much as possible. If there is no problem, let's merge it. Thanks!

**Review by tmysik — APPROVED** [review]
Really nice (and huge, of course :) patch! Thanks for it!

**Comment by tmysik** [comment]
@junichi11 My only comment - it would be nice to have a different icon for enum, but I can imagine that there is nobody who could draw it...

**Comment by junichi11** [comment]

> My only comment - it would be nice to have a different icon for enum, but I can imagine that there is nobody who could draw it...

Yes, I think so.
I've copied the enum icon from CSL in this change, but the icon of enum cases is the same as `const`...

```php
class ExampleClass {
    const CONSTANT = 1;
}

enum ExampleEnum: string {
    case ENUM_CASE = "";
    const CONSTANT = 1;
}
```
nb-php81-enumerations-icon

tmysik:
@junichi11 Exactly...

PR #3942, opened 2022-04-08
Fix project load caching

sdedic:
The goal of this PR is to ensure that sample project (`micronaut-core` was used) is loaded through Gradle daemon just for the first time and second time it is opened is loaded solely from the cache, provided the buildscripts did not change.

In fact currently the project loads each time because of several issues - explanations are done inline in the diff. Please review carefully. Overview:
- race condition in project load. 2nd opinion of fix strategy is needed here.
- excluded more configuration, useless from the (current state of the) IDE - we have only rudimentary support for Kotlin
- artifacts files fetched recursively, similar to modules, to populate `jvm_resolved_artifacts`
- excluded abstract modules that only serve as variant selectors
- artifact cache logic fixes

I've added a 'development feature' - if the loglevel is set to FINER or above, the output from the Gradle daemon will be redirected to IDE's log. If eventually the `NbProjectInfoBuilder` prints some diagnostic messages, they will be visible in the log. If the log is set to `FINEST`, gradle daemon gets `--debug` to spam the log with gradle's own debugging.

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -32,6 +32,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.codehaus.groovy.runtime.InvokerHelper;
 import org.gradle.api.Project;
@@ -41,9 +42,11 @@ import org.gradle.api.artifacts.FileCollectionDependency;
 import org.gradle.api.artifacts.ModuleDependency;
 import org.gradle.api.artifacts.ProjectDependency;
 import org.gradle.api.artifacts.ResolveException;
+import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ModuleComponentSelector;
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.artifacts.result.ArtifactResolutionResult;
 import org.gradle.api.artifacts.result.ArtifactResult;
 import org.gradle.api.artifacts.result.ComponentArtifactsResult;
@@ -82,7 +85,12 @@ class NbProjectInfoBuilder {
         "jacocoAnt",
         "jdepend",
         "pmd",
+        ".*DependenciesMetadata"
     }));
+    
+    private static final Pattern CONFIG_EXCLUDES_PATTERN = Pattern.compile(
+        CONFIG_EXCLUDES.stream().reduce("", (s1, s2) -> s1 + "|" + s2)
+    );
 
     private static final Set<String> RECOGNISED_PLUGINS = new HashSet<>(asList(new String[]{
         "antlr",
@@ -172,7 +180,8 @@ class NbProjectInfoBuilder {
         model.getInfo().put("project_buildDir", project.getBuildDir());
         model.getInfo().put("project_projectDir", project.getProjectDir());
         model.getInfo().put("project_rootDir", project.getRootDir());
-        model.getInfo().put("gradle_user_home", project.getGradle().getGradleHomeDir());
+        model.getInfo().put("gradle_user_home", project.getGradle().getGradleUserHomeDir());
+        model.getInfo().put("gradle_home", project.getGradle().getGradleHomeDir());
 
         Set<Configuration> visibleConfigurations = configurationsToSave();
         model.getInfo().put("configurations", visibleConfigurations.stream().map(conf->conf.getName()).collect(Collectors.toCollection(HashSet::new )));
@@ -433,7 +442,12 @@ class NbProjectInfoBuilder {
                             ResolvedDependencyResult rdr = (ResolvedDependencyResult) it2;
                             if (rdr.getRequested() instanceof ModuleComponentSelector) {
                                 ids.add(rdr.getSelected().getId());
-                                componentIds.add(rdr.getSelected().getId().toString());
+                                // do not bother with components that only select a variant, which is itself a component
+                                // TODO: represent as a special component type so the IDE shows it, but the IDE knows it is an abstract
+                                // intermediate with no artifact(s).
+                                if (!rdr.getResolvedVariant().getExternalVariant().isPresent()) {
+                                    componentIds.add(rdr.getSelected().getId().toString());
+                                }
                             }
                         }
                         if (it2 instanceof UnresolvedDependencyResult) {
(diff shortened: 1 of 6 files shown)


extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -172,7 +180,8 @@ private void detectProjectMetadata(NbProjectInfoModel model) {
         model.getInfo().put("project_buildDir", project.getBuildDir());
         model.getInfo().put("project_projectDir", project.getProjectDir());
         model.getInfo().put("project_rootDir", project.getRootDir());
-        model.getInfo().put("gradle_user_home", project.getGradle().getGradleHomeDir());
+        model.getInfo().put("gradle_user_home", project.getGradle().getGradleUserHomeDir());
```
sdedic, line 183: Passing `gradleHomeDir` instead of userdir caused the `GradleModuleFileCache21` to work over **gradle distribution** directory rather than **user cache** directory, effectively disabling the accessor.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -433,7 +442,12 @@ private void detectDependencies(NbProjectInfoModel model) {
                             ResolvedDependencyResult rdr = (ResolvedDependencyResult) it2;
                             if (rdr.getRequested() instanceof ModuleComponentSelector) {
                                 ids.add(rdr.getSelected().getId());
-                                componentIds.add(rdr.getSelected().getId().toString());
+                                // do not bother with components that only select a variant, which is itself a component
+                                // TODO: represent as a special component type so the IDE shows it, but the IDE knows it is an abstract
+                                // intermediate with no artifact(s).
+                                if (!rdr.getResolvedVariant().getExternalVariant().isPresent()) {
```
sdedic, line 448: I found out that regular JAR dependencies do not have `externalVariant` populated.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -487,6 +501,15 @@ private void detectDependencies(NbProjectInfoModel model) {
 
             if (resolvable(it)) {
                 try {
+                    Set<ResolvedArtifact> arts = it.getResolvedConfiguration()
```
sdedic, line 504: The `collectArtifacts` [at line 516](https://github.com/apache/netbeans/pull/3942/files#diff-9fd2231c8ce99fda4d6d8295c69707bf67f91b6b492ee886a736f8a4a6d55a08R516) fetches modules recursively, but JARs are not. Therefore the `jvm_resolved_artifacts` map was incomplete: did not assign artifacts for all components turned into `ModuleDependency` for the `micronaut-core` project.
Adding this provides as much JAR locations as possible. Checked that does not need much additional time compared to original.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -627,7 +650,7 @@ private void collectModuleDependencies(final NbProjectInfoModel model, String co
     private Set<Configuration> configurationsToSave() {
         return project
                 .getConfigurations()
-                .matching(c -> !CONFIG_EXCLUDES.contains(c.getName()))
+                .matching(c -> !CONFIG_EXCLUDES_PATTERN.matcher(c.getName()).matches())
```
sdedic, line 653: The added regexp excludes all `*DependencyMetadata` configurations like `compileOnlyDependencyMetadata` used by Kotlin plugin. Useless for the IDE, but since the module has no JVM artifacts, it confuses the caching / validation mechanism.

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
```diff
@@ -477,9 +477,10 @@ public boolean satisifes(LoadingCF other) {
                     this.loading = null;
                 }
             }
+            f.complete(prj);
```
sdedic, line 480: This is tricky. I've encountered a situation when:
* the project was loading
* fired the [event at line 481](https://github.com/apache/netbeans/pull/3942/files#diff-e833b1ad0623c45c9af375ded2ba2f854c529638f4334c624969e3ea40a6e05eL481),
* it went from the listener through a series of other listeners (classpath changes), 
* then eventually attempted to obtain the GradleProject - in the same execution thread
* the future was NOT completed yet (as it would have been completed after the event dispatch ended)
* the loader was blocked
* all other gradle reloads were blocked, as the RP has throughput 1

An opinion on the fix is needed. This proposed version is simple: fire event only after the future completes. A code that does `NbGradleProject.toQuality(...).thenApply()` will progress with the loaded project, BUT then it gets a change event ... and may perform additional processing / updates, although the state is clean.
It could be possible to introduce a `ThreadLocal` that would return completed future just for the event firing thread; event will be delivered **before** the (public) Future actually completes. But this does not prevent a deadlock between more threads: firing thread waiting on X, while X is blocked on Future.get()

ppisl (approved):

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
ppisl: This is really tricky and i'm not sure about this. I would vote that this fix is ok for now.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
lkishalmi: Nice catch!

extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectImpl.java
lkishalmi: Well I think this patch is reasonable. There might be better options we can come up out there, but right now it seems to be good enough. Just confirm that I'm reading it right. In this patch, we simply make the completable future completed before we are firing the notification events, so the event handlers can get the project in non-transient state?

sdedic: That was the point; the drawback now is that the client may 'over-react' since it first computes its state when the future becomes fullfilled, then receives a change event (and recomputes, hopefully to the same state).

extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
```diff
@@ -425,10 +439,95 @@ private static NbProjectInfo retrieveProjectInfo(NbGradleProjectImpl projectImpl
...
+                    action.addArguments("--debug"); // NOI18N
+                }
+                PipedOutputStream pos = new ImmediatePipedOutputStream();
+                try {
+                    logStream = pos;
+                    DAEMON_LOG_RP.post(new LogDelegate(new PipedInputStream(pos)));
+                } catch (IOException ex) {
+                    throw new IllegalStateException(ex);
+                }
+                action.setStandardOutput(pos);
```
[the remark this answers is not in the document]
sdedic: Update: No, gradle output from the `LegacyProjectLoader` BuildAction goes to /dev/null unless explicitly not redirected somewhere. 
The Piped stuff was the simplest that I could think of as Gradle outputs byte stream and I need to detect newlines among the (potential) UTF8 multibyte garbage + decode Strings from the stream. It could be probably done directly in a special OutputStream implementation. Or maybe I could rely on `flush()` to be called after each line (?)

sdedic:
Apologies - I needed to rewrite the synchronization to the other proposed solution: the tests show that some code might rely on the priming build (that uses the loaded future) or the loaded future itself as trigger for further processing. At that time it was assumed that the project has already reloaded itself **including the lookup and action change**, which is done inside event handlers.

So now the previous state is maintained: the future completes only after the whole project is refreshed, but event receivers see a special completed future when they ask for it.

Sorry for the inconvenience.

sdedic:
Only unrelated J17 tests fail (groovy editor UI typinghook); merging.

PR #3948, opened 2022-04-09
readme updates (fixed links) + new log/config/cache section.

mbien:
added section to the readme which lists config/log/cache locations. Easily linkable to users which have configuration issues. Fixed some old or broken links.

preview link: https://github.com/mbien/netbeans/tree/readme-updates#log-config-and-cache-locations

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ Apache NetBeans is an open source development environment, tooling platform, and
 
 ### Build status
    * GitHub actions
-     * [![Build Status](https://github.com/apache/netbeans/actions/workflows/main.yml/badge.svg)](https://github.com/apache/netbeans/actions/workflows/main.yml)
-     * [![Profiler Lib Native Binaries](https://github.com/apache/netbeans/actions/workflows/native-binary-build-lib.profiler.yml/badge.svg)](https://github.com/apache/netbeans/actions/workflows/native-binary-build-lib.profiler.yml)
+     * [![Build Status](https://github.com/apache/netbeans/actions/workflows/main.yml/badge.svg?branch=master)](https://github.com/apache/netbeans/actions/workflows/main.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,8 +39,7 @@ Apache NetBeans is an open source development environment, tooling platform, and
 
   * Git
   * Ant 1.9.9 or above
-  * JDK 11 (to build NetBeans)
-  * JDK 11 or above (to run NetBeans)
+  * JDK 11 or above (to build and run NetBeans)
   * MinGW (optional), to build Windows Launchers
```
JaroslavTulach, line 43: This might deserve also a note about profiler binaries...

mbien: I don't know the requirements for the profiler build. I could add a bullet point if its simple enough. If it's requirements are more complex maybe it would deserve its own readme?

mbien:
@neilcsmith-net what do you think? I almost forgot about this PR.

README.md
neilcsmith-net: Yes, maybe a separate section for native binaries, and probably needs rewriting given - https://github.com/apache/netbeans-native-launchers cc/ @ebarboni

neilcsmith-net (approved):
Looks good!  There may be a few other changes around natives that would be good to make, but can follow.

Might have been good to get this in delivery as it's included in the source bundle, but I don't think we need further delays now!

EDIT:  also looking at source / binary zips we may want to update some of this info also in https://github.com/apache/netbeans/blob/master/nb/ide.branding/release-toplevel/README.html - I think that's where the binary zip readme is coming from.

README.md
neilcsmith-net: @ebarboni let's take the Swift discussion on list once NB14 out.  I've got some infra I can test on at least.  (And have another variant of the Swift launcher that works for IDE and RCP in NBPackage).  @oyarzun probably good to get in on it.

mbien: awesome, my understanding (taken from other comments and here) is that this is good to go, the profiler/installers(?) section will be added via a follow up PR.

PR #3949, opened 2022-04-09
Action to create a method breakpoint on a stackframe

jtulach:
Often, when debugging, I hit a breakpoint, but I'd like to know what is going to happen later, when the method call returns. Stepping out, is an option, but not really pleasant when there is a deep stack in between top place and the desired frame. NetBeans already supports _method breakpoint_ - just placing it is a huge pain. This PR simplifies that by adding _Add breakpoint..._ action to each `CallStackFrame`:

image

One the action is invoked a classical method breakpoint dialog is shown, but with pre-filled class and method names:

image

all one needs to do is to fill for example condition `n == 4` and press OK. When the execution continues, it stops in the desired frame.

This is particularly useful when debugging application without a source code, or written in a different language than Java.

### DIFF
--- a/java/debugger.jpda.ui/nbproject/project.properties
+++ b/java/debugger.jpda.ui/nbproject/project.properties
@@ -17,7 +17,7 @@
 
 cp.extra=${tools.jar}
 javac.compilerargs=-Xlint:unchecked
-javac.source=1.7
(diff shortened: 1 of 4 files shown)


entlicher (approved):
This is nice! Thank you!

PR #3950, opened 2022-04-09
Sync fixes: currentP can be seen as null by workers.

sdedic:
The shared variable `currentP` can be sometimes turned to `null` when a worker is already planned (but did not execute, or is executing) to the RP. 
The PR should pass the `currentP` to the worker task, and the task checks if its view of what is `currentP` is still accurate before executing.


lkishalmi (approved):
Looks good at first sight. Thank you!

PR #3951, opened 2022-04-10
Extract and display original file name from disassembled .class

jtulach:
While working with scala libraries I found it misleading that the original source is assumed to be `.java`. The source name is usually encoded in the `.class` file and this PR extracts it, when available:

image

Seeing the `IR.scala` name can help the developer locate the right file on disk.


jlahoda (approved):
Ok, I guess.

PR #3953, opened 2022-04-10
version bump: require javac 18.

mbien:
NetBeans 14 expects javac 18 which means we have to increment the version for the javac checks too.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/NoJavacHelper.java
@@ -34,21 +34,22 @@ import org.openide.modules.OnStart;
  */
 public class NoJavacHelper {
 
-    private static final boolean hasWorkingJavac;
+    public static final int REQUIRED_JAVAC_VERSION = 18; // <- TODO: increment on every release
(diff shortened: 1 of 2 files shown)


matthiasblaesing (approved):
Eyeballed this and looks sane to me.

neilcsmith-net (approved):
Looks good, at least for a short term solution. A proper API for this, and linked to nb-javac updates in some way so the increment _can't_ be missed, might be useful in the future?

mbien:
> Looks good, at least for a short term solution. A proper API for this, and linked to nb-javac updates in some way so the increment _can't_ be missed, might be useful in the future?

@neilcsmith-net I thought about this a bit. 

It is essentially a runtime dependency check. Someone gonna have to bump the number no matter where it is taken from. We can't put it into the nb-javac module since the check has to work without it.

an assert could potentially check:
`assert ((NB_VERSION / 2 + 11) == REQUIRED_JAVAC_VERSION)`
which would fail if someone forgets the bump REQUIRED_JAVAC_VERSION in NB 16. However, where do we get `NB_VERSION` from? The release would have a system property for that, but a dev version doesn't.

mbien:
integrating this one. We can further improve those version checks for the next release.

PR #3955, opened 2022-04-11
Support invocation expressions in LSP Call Hierarchy; bugfixes

sdedic:
Call hierarchy implementation for LSP throws an exception:
```
	at org.netbeans.modules.refactoring.java.callhierarchy.CallHierarchyTasks$CallersTask.runTask(CallHierarchyTasks.java:313)
[catch] at org.netbeans.modules.refactoring.java.callhierarchy.CallHierarchyTasks$CallTaskBase.run(CallHierarchyTasks.java:288)
```
on a trivial source:
```
package org.yourcompany.yourproject;

public class Locations {

    private void m1() {
        m2();
    }

    private void m2() {
        m3();
    }

    private void m3() {
        throw new UnsupportedOperationException("Not supported yet.");
    }

}
```
when the caret is positioned at stuff outside of method scope. It does not show any results on method invocations.

---
My bad, the implementation was oversimplified as the code only searched for the enclosing method. This PR extends that to method invocation and new expressions. Some other bugs were also fixed.


ppisl (approved):
Thanks for fixing this. Looks good to me.

dbalek (approved):

PR #3957, opened 2022-04-11
Replace _dark and svg icon variant in VSCode

sdedic:
The regexps were too strict: if NBLS was started in dark L&F it serves icons with `_dark` suffix (it has to be probably fixed in IconUtilities - the base URI should not contain these decorations). But it's sufficient now to just not specify the icon extension as the regexp matches all suffixes.


dbalek (approved):

sdedic:
Just unreliable tests fail; merging.

PR #3968, opened 2022-04-12
[vscode] Project View synchronizes with the active editor.

sdedic:
This PR adds a new message client > NBLS to the TreeView LSP protocol: find path from a certain root to some data => tuple of node ids. It is now only implemented for Projects view, which delegates to `PathFinder`s of the individual projects to do the searching job.

The vscode client uses this message to find a path in its Projects tree view in order to **reveal** the position of the current editor. This PR adds automatic synchronization (default: on), a setting to manage it (`netbeans.revealActiveInProjects`) true / false and an explicit command `java.select.editor.projects` to reveal the position manually - can be bound to some keyboard shortcut.

### DIFF
--- a/java/java.lsp.server/vscode/package.json
+++ b/java/java.lsp.server/vscode/package.json
@@ -119,6 +119,11 @@
 					"default": true,
 					"description": "Avoid conflicts with other Java extensions"
 				},
+				"netbeans.revealActiveInProjects": {
+					"type": "boolean",
+					"default": true,
+					"description": "Reveals active text editor in Projects view"
+				},
 				"java.test.editor.enableShortcuts": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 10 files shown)


JaroslavTulach (approved):
Works for me.

MartinBalin:
It is quite late for 13.0.301 release of VSIX and there is no time to properly test it. I propose to merge this and by default be OFF. After VSIX release, turn this ON by default.

ppisl (approved):
From code point of view it looks good.

java/java.lsp.server/vscode/package.json
```diff
@@ -121,7 +121,7 @@
 				},
 				"netbeans.revealActiveInProjects": {
 					"type": "boolean",
-					"default": true,
+					"default": false,
```
JaroslavTulach, line 124: NBLS 13.0.301 release is over. Is there a plan to change the default to `true` for NetBeans 14 release, @sdedic?

PR #3969, opened 2022-04-12
Fixing potential NPE.

sdedic:
It seems that `getResolvedVariant()` can be sometimes null in the case of resolved component - need to check for null here. Gradle sources (serializer) assume that getExternalVariant() is never null (as it is Optional).
The fix is conservative = if no resolved variant -> cannot check for externality -> pass on as a dependency.


ppisl (approved):
No problem

MartinBalin (approved):

lkishalmi (approved):
Nice!

sdedic:
Only unreliable tests fail; merging.

PR #3970, opened 2022-04-12
Jdbc url templates in the Add DB Connection command

jhorvath:
Adding simple JDBC URL templates into Add DB connection command.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
@@ -159,6 +159,9 @@ public class DBAddConnection extends CodeActionsProvider {
                         int i = ((Double) selectedItems.get(0).getUserData()).intValue();
                         JDBCDriver driver = drivers[i];
                         String urlTemplate = urlTemplates.get(driver.getClassName());
+                        if (urlTemplate == null) {
+                            urlTemplate = "";
+                        }
                         client.showInputBox(new ShowInputBoxParams(
                                 Bundle.MSG_EnterDbUrl(), urlTemplate)).thenAccept((u) -> {
                             if (u == null) {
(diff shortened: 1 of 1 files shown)


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/db/DBAddConnection.java
```diff
@@ -148,8 +158,9 @@ public CompletableFuture<Object> processCommand(NbCodeLanguageClient client, Str
                     if (!selectedItems.isEmpty()) {
                         int i = ((Double) selectedItems.get(0).getUserData()).intValue();
                         JDBCDriver driver = drivers[i];
+                        String urlTemplate = urlTemplates.get(driver.getClassName());
```
sdedic, line 161: give a safe fallback in unlikely case the driver classname will not be registered in the Map.

jhorvath: fixed in 4b935e8

matthiasblaesing (changes requested):
There is already a database URL templating system. It is located in the database explorer. There is no reason to duplicate that work. See `org.netbeans.modules.db.util.DriverListUtil` in `ide/db`.

jhorvath:
> There is already a database URL templating system. It is located in the database explorer. There is no reason to duplicate that work. See `org.netbeans.modules.db.util.DriverListUtil` in `ide/db`.

I know about `DriverListUtil`. I decided to duplicate couple of most used URLs for good reason. Templates in `DriverListUtil` are designed to support complex dialogs in DB Explorer, it covers almost all possible variants, supports all tiny details of URLs. It makes it less suitable for purpose I have in VSCode. UI of `DBAddConnection` quite different. It asks user for one field at a time. It is not able to display all field side by side as DB Explorer dialogs does. In VSCode the template would not be populated by values from the dialog. It will be displayed as is to let the user edit it. For this purpose `DriverListUtil` templates are less readable.   
I believe that in this case the duplication is better than modifying `DriverListUtil` to support different UI workflow.

matthiasblaesing (commented):
My comment stands, but I'm not a user of the lsp.server and I don't see myself becoming one. I'm surprised though, that noone cares enough to get better input into the specification. I won't stand in the way of this.

matthiasblaesing (approved):
Seems I have to "approve", even if I just want to move from -1 to 0, so be it.

PR #3973, opened 2022-04-12
Taking caret offset from SourceModificationEvent.

ppisl:
Usually the GSF Languages use `sanitizing` source code to get some reasonable output from parsers. This means that the source is parsed and if there is not usable result, then the edited code is somehow change and parsed again. When the second attempt doesn't returns anything useful, then is used different sanitization and again parsed. 

A sanitization needs to know, where the last change was made, which is usually taken from caret position. In NetBeans this position is taken form opened JTextComponent. In the VSCode case, the components are not opened and caret position is set to -1. Then the edited source is not sanitized correctly. 

This fix is trying to take the possition from SourceModificationEvent.


sdedic (approved):
We're lucky that the event comes in the headless mode.

PR #3974, opened 2022-04-12
Upgrade to Disco v2.

neilcsmith-net:
Upgrade to DiscoClient 2.0.23 to use Disco API v2. Now requires JDK 11. Removes HTTPClient library as DiscoClient uses JDK support.

Makes the client ID, default distribution and default download location brandable.

Moves to a non-hidden default download location (NetBeansJDKs). Using a hidden location makes re-registering difficult, and can hide a lot of disk usage. Logic is similar to that used in OpenProjectListSettings::getProjectsFolder - should be adjacent to NetBeansProjects by default.  Also uses tmp location in development.

NB. This PR includes some test updates related to use of `release 11` in the build.


neilcsmith-net:
OK, pushed an update to `ValidateClassFilesTest` to handle higher class file versions based on module dependencies. Commit validation now passing.  Approach OK @jtulach @mbien @matthiasblaesing ?

matthiasblaesing (approved):
Looks sane to me. I saw the bump in the major version of the spec version and though that the codename base needs to be bumped, but as the module has not API there is no reason to.

I eyeballed the changed test and to my understanding of the logic this looks good.

java/java.disco/nbproject/project.properties
```diff
@@ -17,9 +17,8 @@
...
 release.external/checker-qual-3.9.1.jar=modules/ext/checker-qual-3.9.1.jar
-release.external/discoclient-1.0.1.jar=modules/ext/discoclient-1.0.1.jar
-release.external/httpclient5-5.0.3.jar=modules/ext/httpclient5-5.0.3.jar
-release.external/httpcore5-5.0.2.jar=modules/ext/httpcore5-5.0.2.jar
-release.external/httpcore5-h2-5.0.2.jar=modules/ext/httpcore5-h2-5.0.2.jar
-javac.source=1.8
+release.external/jdktools-11.0.8.jar=modules/ext/jdktools-11.0.8.jar
+release.external/discoclient-2.0.23.jar=modules/ext/discoclient-2.0.23.jar
+javac.source=11
+javac.target=11
```
mbien, line 23: awesome

java/java.disco/src/org/netbeans/modules/java/disco/Client.java
```diff
@@ -87,20 +90,58 @@ public synchronized MajorVersion getLatestSts(boolean b) {
     public synchronized List<Pkg> getPkgs(final Distribution distribution, final VersionNumber versionNumber, final Latest latest, final OperatingSystem operatingSystem,
             final Architecture architecture, final ArchiveType archiveType, final PackageType packageType,
             final Boolean javafxBundled) {
-        return getDisco().getPkgs(distribution, versionNumber, latest, operatingSystem, LibCType.NONE, architecture, Bitness.NONE, archiveType, packageType, javafxBundled, /*directlyDownloadable*/ true, ReleaseStatus.NONE, TermOfSupport.NONE, Scope.PUBLIC);
+//        return getDisco().getPkgs(distribution, versionNumber, latest, operatingSystem, LibCType.NONE, architecture, Bitness.NONE, archiveType, packageType, javafxBundled, /*directlyDownloadable*/ true, ReleaseStatus.NONE, TermOfSupport.NONE, Scope.PUBLIC);
```
mbien, line 93: nitpick: is this line supposed to stay? the vertical version is right below.

java/java.disco/src/org/netbeans/modules/java/disco/SelectPackagePanel.java
```diff
@@ -177,6 +194,11 @@ protected void updateData(Distribution distribution, Integer featureVersion, Lat
                 .execute();
     }
 
+    @Override
+    protected void updateDistributions(List<Distribution> distros) {
+        super.updateDistributions(distros); //To change body of generated methods, choose Tools | Templates.
```
mbien, line 199: someone hasn't updated his code templates :P

platform/o.n.core/test/qa-functional/src/org/netbeans/core/validation/ValidateClassFilesTest.java
```diff
@@ -120,4 +121,24 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
     }
+
+    private Map<String, Integer> calculateFileNameToVersion() {
+        Map<String, Integer> result = new HashMap<>();
+        for (Module module : Main.getModuleSystem().getManager().getModules()) {
+            for (Dependency dep : module.getDependencies()) {
+                if (dep.getType() == Dependency.TYPE_JAVA) {
+                    String javaVersion = dep.getVersion();
+                    if (javaVersion.startsWith("1.")) {
+                        break;
```
mbien, line 132: is `break` correct here? You don't want to check the other dependencies in the list? (did you mean `continue`?)

java/java.disco/external/binaries-list
```diff
@@ -15,7 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 4FCCADF37F1C61A0895BAC3D42FE0C530438826C org.checkerframework:checker-qual:3.9.1
-C4BC772689D4BB2D4128C5E4966F765A8A76E904 org.apache.httpcomponents.client5:httpclient5:5.0.3
-EA298CD1CE1D860747149FBB5F6236300219F378 org.apache.httpcomponents.core5:httpcore5:5.0.2
-7CAEB2EEBDFB7ECA30E7F64E8138D977FC85AB80 org.apache.httpcomponents.core5:httpcore5-h2:5.0.2
```
mbien, line 20: bonus points for removing dependencies

mbien (approved):
looks good + I downloaded two JDKs and everything worked fine, good job.

Please take a look at the last inline comment, I am not sure if the loop is working as intended. Other then that, +1 from me.

as sidenote:
while testing I noticed a few issues with the advanced tab of the JDK selection window, its all reproducible with NB 13, so its not caused by this PR
 - the table does not resize properly when the window resizes
 - duplicated entries appear in the table, its likely that the duplicates are musl/alpine builds, its just that the table does not have a colum which communicates the difference to the user

java/java.disco/nbproject/project.properties
neilcsmith-net: :smile:

java/java.disco/src/org/netbeans/modules/java/disco/Client.java
neilcsmith-net: No, I'll look at tidying up.

java/java.disco/src/org/netbeans/modules/java/disco/SelectPackagePanel.java
neilcsmith-net: Oh $%$*&! I'll tidy this up too.

platform/o.n.core/test/qa-functional/src/org/netbeans/core/validation/ValidateClassFilesTest.java
neilcsmith-net: I meant `break`.  Shouldn't be more than one Java dependency in there? Should probably break later too.

mbien: roger, just wanted to be sure.

neilcsmith-net:
Thanks @mbien The UI is on my list to revise for NB15.  You're right about the reason for the duplicates.  I didn't write the original NetBeans support for Disco and it predates a bunch of features in the API.  I want to bring our support more in line with other IDE plugins / [website](https://foojay.io/download/).

PR #3983, opened 2022-04-13
Update FlatLaf from 2.0.1 to 2.1

DevCharly:
Update FlatLaf to v2.1: https://github.com/JFormDesigner/FlatLaf/releases/tag/2.1

This release significantly improves the usability of submenus. It makes it much easier to move the mouse into a submenu without the risk that the submenu is hidden.

grafik

Previously, you have to move the mouse on the **red** path, otherwise the submenu will be hidden.
Now, you can move the mouse on the **green** path, over other menu items.

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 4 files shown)


neilcsmith-net:
Merged.  Great to get this in for NB14.  Could you also have a look at #3115 in context of NB14?

PR #3987, opened 2022-04-14
Show source levels of Java platform in Ant APISupport UI.

neilcsmith-net:
The Ant APISupport UI only showed a hard-coded list of possible Java source levels for platform modules. This was a complaint raised by someone on users@. This is a quick fix to show levels up to that of the selected Java platform. (I couldn't see an API to get the supported levels directly? Would be useful).

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/CustomizerSources.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/CustomizerSources.java
@@ -22,10 +22,14 @@ package org.netbeans.modules.apisupport.project.ui.customizer;
 import java.awt.EventQueue;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.util.ArrayList;
+import java.util.List;
+import org.netbeans.api.java.platform.JavaPlatform;
 import org.netbeans.modules.apisupport.project.ui.ApisupportAntUIUtils;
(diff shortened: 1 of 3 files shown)


apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/CustomizerSources.java
```diff
@@ -204,4 +209,23 @@ private void initAccessibility() {
...
+        try {
+            String platformVersion = platform.getSpecification().getVersion().toString();
+            int maxLevel = Integer.parseInt(platformVersion.split("\\.")[0]);
+            for (int level = 9; level <= maxLevel; level++) {
+                levels.add(Integer.toString(level));
+            }
+        } catch (Exception ex) {
+            Exceptions.printStackTrace(ex);
+        }
+        return levels.toArray(String[]::new);
```
mbien, line 228: that would be Java 11 API. needs to be `new String[0]`

mbien (approved):
looks good

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/CustomizerSources.java
neilcsmith-net: Good catch!  Getting mixed up with `Stream`. Doesn't help that `-release` isn't used because `requires.nb.javac=true` customizes the boot classpath.

PR #3988, opened 2022-04-15
Added more tests to BaseUtilities, Pair and MapFormat classes in org.netbeans.util package.

lbownik:
Added more tests BaseUtilities
Added more tests to Pair class
Added more tests to MapFormat class

### DIFF
--- a/platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
+++ b/platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
@@ -21,17 +21,20 @@ package org.openide.util;
 
 import java.io.File;
 import java.net.URI;
-import java.util.Locale;
+import static java.util.Locale.US;
 import org.netbeans.junit.NbTestCase;
 import static org.openide.util.BaseUtilities.*;
 import static java.lang.Boolean.*;
+import java.text.BreakIterator;
 
 /**
  *
- * @author Jiri Rechtacek et al.
+ * @author Jiri Rechtacek et al. , Lukasz Bownik
  */
 public class BaseUtilitiesTest extends NbTestCase {
+    
     private String originalOsName;
+    private final BreakIterator breakIterator = BreakIterator.getCharacterInstance(US);
 
     public BaseUtilitiesTest(String name) {
         super(name);
@@ -59,7 +62,7 @@ public class BaseUtilitiesTest extends NbTestCase {
     public void testGetOperatingSystemFreebsd () {
         System.setProperty ("os.name", "FreeBSD");
         assertEquals ("System.getProperty (os.name) returns FreeBSD", "FreeBSD", System.getProperty ("os.name"));
-        assertEquals ("System.getProperty (os.name) returns freebsd", "freebsd", System.getProperty ("os.name").toLowerCase (Locale.US));
+        assertEquals ("System.getProperty (os.name) returns freebsd", "freebsd", System.getProperty ("os.name").toLowerCase (US));
         assertEquals ("FreeBSD recognized as OS_FREEBSD", BaseUtilities.OS_FREEBSD, BaseUtilities.getOperatingSystem ());
     }
 
@@ -100,6 +103,7 @@ public class BaseUtilitiesTest extends NbTestCase {
         assertFalse ("freebsd is not isWindows", BaseUtilities.isWindows ());
         assertTrue ("freebsd isUnix", BaseUtilities.isUnix ());
     }
+    
     //--------------------------------------------------------------------------
     public void test_IsJavaIdentifier_returnsTrue_whenGivenJavaIdentifier()
             throws Exception {
@@ -125,12 +129,7 @@ public class BaseUtilitiesTest extends NbTestCase {
     public void test_toFile_throwsNullPonter_whenArgumentIsNull()
             throws Exception {
 
-        try {
-            toFile(null);
-            fail();
-        } catch (final NullPointerException e) {
-            //good
-        }
+        expectNullPointer(() -> toFile(null));
     }
     
     //--------------------------------------------------------------------------
@@ -237,12 +236,7 @@ public class BaseUtilitiesTest extends NbTestCase {
     public void test_toURI_throwsNullPonter_whenArgumentIsNull()
             throws Exception {
 
-        try {
-            toURI(null);
-            fail();
-        } catch (final NullPointerException e) {
-            //good
-        }
+        expectNullPointer(() -> toURI(null));
     }
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


platform/openide.util/src/org/openide/util/MapFormat.java
```diff
@@ -188,9 +189,14 @@ public String processPattern(String newPattern) throws IllegalArgumentException
 
             if (ridx < 0) {
                 if (exactmatch) {
-                    throw new IllegalArgumentException(
-                        NbBundle.getBundle(MapFormat.class).getString("MSG_UnmatchedBraces") + " " + lidx
-                    );
+                    throw new IllegalArgumentException() {
+                        @Override
+                        public String getMessage() {
+                            // this is ugly- but it allows unit testing
```
vieiro, line 195: Why?

platform/openide.util/src/org/openide/util/MapFormat.java
```diff
@@ -221,7 +227,7 @@ private String formatObject(Object obj) {
         if (obj instanceof Number) {
             return NumberFormat.getInstance(locale).format(obj); // fix
         } else if (obj instanceof Date) {
-            return DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT, locale).format(obj); //fix
+            return DateFormat.getDateTimeInstance(SHORT, SHORT, locale).format(obj); //fix
```
vieiro, line 230: DateFormat.SHORT is more legible to me

platform/openide.util/src/org/openide/util/MapFormat.java
```diff
@@ -276,12 +282,16 @@ public StringBuffer format(Object pat, StringBuffer result, FieldPosition fpos)
...
-                    throw new IllegalArgumentException(
-                        MessageFormat.format(
-                            NbBundle.getBundle(MapFormat.class).getString("MSG_FMT_ObjectForKey"),
-                            new Object[] { new Integer(key) }
-                        )
-                    );
+                    throw new IllegalArgumentException() {
+                        @Override
+                        public String getMessage() {
+                            // this is ugly- but it allows unit testing
```
vieiro, line 288: Why?

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
```diff
@@ -309,66 +311,74 @@ public void test_toURI_producesURI_wchichResolvesAndWrapsProperly()
...
+        assertEquals(0, parseParameters("").length);
+        assertEquals(0, parseParameters("   \t\r\n").length);
+    }
+
+    //--------------------------------------------------------------------------
+    public void test_parseParameters_returnsProperArray_forProperInput() {
+
+        String[] args;
+
+        args = parseParameters("\"c:\\program files\\jdk\\bin\\java\" -Dmessage=\"Hello /\\\\/\\\\ there!\" -Xmx128m");
```
vieiro, line 338: Why to lines here?

platform/openide.util/test/unit/src/org/openide/util/MapFormatTest.java
```diff
@@ -18,68 +18,178 @@
...
+import org.junit.Test;
+import static org.openide.util.MapFormat.format;
 
 /**
  *
- * @author Jaroslav Tulach
+ * @author Jaroslav Tulach, Lukasz Bownik
  */
-public class MapFormatTest extends TestCase {
+public class MapFormatTest {
```
vieiro, line 36: Why remove TestCase?

platform/openide.util/test/unit/src/org/openide/util/PairTest.java
```diff
@@ -19,39 +19,84 @@
...
+        
+        assertEquals("Pair[1,1]", pair11a.toString());
+        assertEquals("Pair[1,1]", pair11b.toString());
+        assertEquals("Pair[1,2]", pair12.toString());
+        assertEquals("Pair[1,null]", pair1Null.toString());
+        assertEquals("Pair[null,1]", pairNul1.toString());
+        assertEquals("Pair[null,null]", pairNullNull.toString());
+    }
+    //-------------------------------------------------------------------------- 
+    private final Pair<Integer, Integer> pair11a = Pair.of(1, 1);
```
vieiro, line 95: If we're testing "Pair" here, we should instantiate Pairs in test methods/setup methods, not as class members. What happens if the constructors themselves fail?

vieiro (changes requested):

platform/openide.util/src/org/openide/util/MapFormat.java
lbownik: Executing 
NbBundle.getBundle(MapFormat.class).getString("MSG_UnmatchedBraces") 
throws exception saying thet no bundle has been found. 
so I detered this execution to be able to run tets

better solution is to make sure bundle exists while unit tests are being executed - I don't know how to do it yet

platform/openide.util/src/org/openide/util/MapFormat.java
lbownik: ok. I'll revert

platform/openide.util/src/org/openide/util/MapFormat.java
lbownik: Executing
NbBundle.getBundle(MapFormat.class).getString("MSG_UnmatchedBraces")
throws exception saying thet no bundle has been found.
so I detered this execution to be able to run tets

better solution is to make sure bundle exists while unit tests are being executed - I don't know how to do it yet

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
lbownik: sure. I'll fix it.

lbownik: So I don't have to declare this variable agai and again. I merger eprerate test methors into one and I'm reusing this variable as you can see below.

platform/openide.util/test/unit/src/org/openide/util/MapFormatTest.java
lbownik: Moved to JUnit4 test case.

platform/openide.util/test/unit/src/org/openide/util/PairTest.java
lbownik: good point. I'll fix it

vieiro:
Hi @lbownik ,

Travis reports failing tests here:
```
after_failure.1

0.08s$ nbbuild/travis/print-junit-report.sh

=================== JUnit Report Summary / failed tests ===================

org.openide.util.MapFormatTest

====================== JUnit failure details ===============================

Suite: org.openide.util.MapFormatTest

Stdout ----------%<----------%<-------------%<-------------%<---------------

8/6/34 12:23 PM 999.

------------- End suite org.openide.util.MapFormatTest ------------
```

Regarding the use of `NbBundle` in tests, maybe you want to take a look at `NbTestCase` . We have a tutorial here https://netbeans.apache.org/tutorials/nbm-test.html (that needs a review, by the way).

lbownik:
Travis failed tests have nothing to do with my changes :|

platform/openide.util/src/org/openide/util/MapFormat.java
```diff
@@ -152,7 +152,7 @@ protected Object processKey(String key) {
     * parser found unmatched brackets (this exception should be switched off
     * using setExactMatch(false)).
     */
-    public String processPattern(String newPattern) throws IllegalArgumentException {
+    private String processPattern(String newPattern) throws IllegalArgumentException {
```
vieiro, line 155: This is a public method on an exposed API, and sigtests warns about about an incompatibility with previous versions of the API:

```java
email: api-changes@netbeans.org
SignatureTest report
Base version: 9.23
Tested version: 9.25
Check mode: bin [throws removed]
Constant checking: on


Class org.openide.util.MapFormat
  "E1.2 - API type removed" : method public java.lang.String org.openide.util.MapFormat.processPattern(java.lang.String)

BUILD FAILED
/home/travis/build/apache/netbeans/nbbuild/build.xml:1885: The following error occurred while executing this line:
/home/travis/build/apache/netbeans/nbbuild/templates/projectized.xml:778: Signature tests return code is wrong (1), check the messages above
```

vieiro (changes requested):
There's no reason to remove a `public` method from an exposed API, sigtests faiulres. See comments in teh PR.

lbownik:
I reverted MapFormat change (from "proviate" back to "public").
Failed tests have nothing to do with my changes.

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
```diff
@@ -557,4 +567,141 @@ public void test_toPrimitiveArray_returnsArrayOfPrimitives_whenGivenArrayOfObjec
...
+    public void test_wrapStringToArray_throwsNullPointer_whenGivenNullArgument() {
+        
+        try {
+            wrapStringToArray(null, 11, BreakIterator.getLineInstance(), true);
+            fail();
+        } catch (final NullPointerException e) {
+            //good
+        }
+        try {
+            //this is slopy behavior - null pointer exception happens only
```
vieiro, line 642: Why is this "sloppy" behaviour? Is it because:
1. `BreakIterator.getLineInstance()` (which is locale dependent, by the way) is not working as you expect.
2. Or because `wrapStringToArray` is not working properly.
3. Or because `wrapStringToArray` is returning a result that is unexpected to you?

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
```diff
@@ -557,4 +567,141 @@ public void test_toPrimitiveArray_returnsArrayOfPrimitives_whenGivenArrayOfObjec
...
+        } catch (final NullPointerException e) {
+            //good
+        }
+    }
+    //--------------------------------------------------------------------------
+    public void test_wrapStringToArray_assumesWidthEqalsOne_whenGivenZeroWidth() {
+
+        String[] wrapped = wrapStringToArray("a\nb", 0, BreakIterator.getLineInstance(), true);
+        assertEquals(3, wrapped.length);
+        assertEquals("a ", wrapped[0]); // Why is there a space at the end?
```
vieiro, line 656: Quoting from [BreakIterator API](https://docs.oracle.com/javase/7/docs/api/java/text/BreakIterator.html#line): 

> The BreakIterator instances returned by the factory methods of this class are intended for use with natural languages only, not for programming language text.

Question is, are you sure `BreakIterator.getLineInstance()` (which is locale dependent, so would depend on the Locale the user is using to build NetBeans) is what you need here?

vieiro (commented):
Hi @lbownik 
We need some clarification on why we're using BreakIterator with this tests, and why you're surprised there's "a space at the end", etc. Is this because of using BreakIterator or because BaseUtilities (which has been around for several years now) is not working properly?

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
lbownik: This behavior is simply unexpected (I should have used "unexpected" instead of "sloppy").

When a method is called with a null value of a parameter which is not allowd to be null, the most common hevior is for method to throw exception before it executes it's logic. 

Meanwhile wrapStringToArray when given null proceeds with it's logic without performing "not null" check first . If the argument is not dereferenced, everything is fine. If the argument is dereferenced - it throws NullPointer. 

So the right thing would be to tighten the boahavior of this method and introduced "null check" at the beginning of 
wrapStringToArray . On the other hand it seams that this method is never used within the code base (I performed text search), so meybe it can be removed entirely.

lbownik: As far as unexpected results wrapStringToArray.
I found them strange so I'd like to someone more knowledgeable with the platform tell me, whether this is ok or have I found a bug.

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
lbownik: Thanks for spotting this - I fixed it with new commit.

vieiro (changes requested):
Hi @lbownik ,

Quoting from [Wikipedia](https://en.wikipedia.org/wiki/Unit_testing#Description)

> Unit tests are typically [automated tests](https://en.wikipedia.org/wiki/Automated_test) written and run by [software developers](https://en.wikipedia.org/wiki/Software_developer) to ensure that a section of an application (known as the "unit") meets its [design](https://en.wikipedia.org/wiki/Software_design) **and behaves as intended**

So, are the classes under test **behaving as inteded** or not? 

Reading your comments it seems you're surprised about the behaviour of the classes under tests. You say the results are `unexpected` or that `It seems wrapStringToArray is not used at all, maybe we could remove it all together`.

The question I'm asking is: why aren't you sure about the behaviour of the class under test or about the test themselves? If you need clarification or have questions about a given class or a specific behaviour then please ask in the mailing lists. Source code comments are used for clarifying how the source code works, not a proper place to ask questions or express doubts.

platform/openide.util/test/unit/src/org/openide/util/BaseUtilitiesTest.java
vieiro: Why do you find them "strange"?

lbownik:
Could somebody review this once you have a chance?

platform/openide.util/test/unit/src/org/openide/util/MapFormatTest.java
```diff
@@ -18,68 +18,257 @@
...
     }
 
-    protected void tearDown() throws Exception {
+    //--------------------------------------------------------------------------
+    @Test
+    public void format_returnEmptyString_whenGivenEmptyFormatString() {
+
+        assertEquals("", format("", this.johnDoe));
     }
-    public void testFormatIssue67238() {
```
vieiro, line 40: I can't find a test for issue 67238. Have we dropped it?

platform/openide.util/test/unit/src/org/openide/util/MapFormatTest.java
```diff
@@ -18,68 +18,257 @@
...
+
+        try {
+            mf.format("Hello $1$prefix* $1$last.");
+            fail();
+        } catch (IllegalArgumentException e) {
+            //good
+        }
     }
-    
-    public void testExectLineWithTheProblemFromFormatIssue67238 () {
```
vieiro, line 54: Same here. Can we please keep the number of the issue in the test method?

platform/openide.util/test/unit/src/org/openide/util/MapFormatTest.java
```diff
@@ -18,68 +18,257 @@
...
+    //--------------------------------------------------------------------------
+    @Test
+    public void parseObject_returnsItsArgument_whenNoTagsCanBeResolved() {
+        
+        MapFormat mf = new MapFormat(this.johnDoe);
         
+        assertEquals("Hello Adam.", mf.parseObject("Hello Adam.", new ParsePosition(0)));
     }
-    
-    public void testIssue67238() {
```
vieiro, line 70: Same here. Knowing that this test is a regression test for issue 67238 is cool. I think we should keep it.

vieiro (changes requested):
Added some comments.
It's a pity to lose refrence to this issue https://bz.apache.org/netbeans/show_bug.cgi?id=67238 
Can we keep those tests?

PR #3991, opened 2022-04-17
Add Athena/Redshift to Add Connection wizard

eirikbakke:
Add support for Amazon Athena and Amazon Redshift connections in the IDE's database connection wizard.

Also ensure that the sample JDBC URL is properly shown when the user changes the URL type from the dropdown.


matthiasblaesing (approved):
Looks sane to me

eirikbakke:
Thanks! Merging.

PR #3992, opened 2022-04-17
Improve vertical centering of FlatLAF buttons on Windows

eirikbakke:
On FlatLAF on Windows, improve vertical centering of labels in buttons. The solution was to make the buttons one pixel taller, as was done in the standard Swing Windows LAF.

See before/after screenshots below.

Centering

### DIFF
--- a/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
+++ b/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
@@ -21,14 +21,23 @@ package org.netbeans.swing.laf.flatlaf;
 
 import com.formdev.flatlaf.util.UIScale;
 import java.awt.Color;
+import java.awt.Font;
+import java.awt.Insets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.swing.BorderFactory;
 import javax.swing.InputMap;
 import javax.swing.KeyStroke;
 import javax.swing.UIDefaults;
 import javax.swing.UIDefaults.LazyValue;
 import javax.swing.UIManager;
+import javax.swing.border.CompoundBorder;
+import javax.swing.plaf.InsetsUIResource;
 import org.netbeans.swing.laf.flatlaf.ui.FlatTabControlIcon;
 import org.netbeans.swing.plaf.LFCustoms;
 import org.netbeans.swing.tabcontrol.plaf.TabControlButton;
+import org.openide.util.Utilities;
 
 /**
  * LFCustoms for FlatLaf based LAFs (light, dark, etc).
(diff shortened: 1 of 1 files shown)


eirikbakke:
I've been using this patch in my working IDE and platform app since November now; I think it should be safe to apply.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
```diff
@@ -120,6 +129,34 @@ public Object[] createApplicationSpecificKeysAndValues() {
...
+          add one pixel to the button's height to ensure that it is odd-numbered. This makes the
+          text centered with either Tahoma 11 (the default Swing Windows LAF font) or Segoe UI 12
+          (the default font on modern Windows versions, and on FlatLAF on Windows). */
+          for (String key : new String[] { "Button", "RadioButton", "ToggleButton" }) {
+            UIDefaults defaults = UIManager.getDefaults();
+            Font font = defaults.getFont(key + ".font");
+            Insets bm = defaults.getInsets(key + ".margin");
+            if (font != null && bm != null &&
+                font.getFamily().equals("Segoe UI") && font.getSize() == 12)
+            {
```
neilcsmith-net, line 146: Niggle : please have brace on line above.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
```diff
@@ -120,6 +129,34 @@ public Object[] createApplicationSpecificKeysAndValues() {
...
+              if (bm.top == bm.bottom) {
+                result.add(key + ".margin");
+                /* Create an InsetsUIResource rather than an Insets instance if necessary, as
+                FlatLAF treats them differently. Not doing this caused buttons in the main toolbar
+                to become very wide. */
+                result.add(bm instanceof InsetsUIResource
```
neilcsmith-net, line 152: My inclination is that this should only ever be InsetsUIResource.  Are there instances of plain Insets?  If so, from where, and should they be replaced?

neilcsmith-net (approved):
Makes sense, although I don't have a system I can check this on.  Made couple of minor comments.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/FlatLFCustoms.java
eirikbakke: It was just a precaution, not something I observed. But since FlatLAF treats Insets and InsetsUIResource differently, I'd imagine cases where FlatLAF customizations might pick one or the other (though this would be very weird). I figured that leaving the same kind of instance was the least likely to cause problems.

neilcsmith-net: I meant more that we should probably only replace it at all if it is already a `UIResource`.

eirikbakke: Oh, sure, that makes sense. Will change it.

eirikbakke:
Applied fixes, tested on Windows with latest master to confirm that the adjustment still is applied. Now waiting for build checks...

DevCharly (approved):
Definitely looks better 👍

Issue #3921, opened 2022-04-03
When package is renamed NB doesnt rename it in module-info opens section


PashaTurok:
### Apache NetBeans version

Apache NetBeans 13

### What happened

When java package is renamed NB doesn't rename it in module-info `opens` section. At the same time in `exports` section package is renamed. For example if you have

    exports com.company.foo;
    opens com.company.foo;

Then after renaming foo package to bar, it will become

    exports com.company.bar;
    opens com.company.foo;

### Did this work correctly in an earlier version?

No

### Operating System

Ubuntu 20.04.3 LTS, Linux version 5.11.0-38-generic running on amd64; UTF-8; en_US (nb)

### JDK

14

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

PR #3993, opened 2022-04-17
[3921] Correctly handling opens directive in package rename.

jlahoda:
When a package is renamed, reference to it in the `opens` directive is not renamed - fixed. See:
https://github.com/apache/netbeans/issues/3921


matthiasblaesing (approved):
I'm no expert in the javac area, but the change brings the handling of opens in line with the existing exports handling, which makes sense to me. It is also protected by a unittest, that matches the error description. It would be good to merge this soon.

matthiasblaesing:
Lets get this in for NB14 (which is branched tomorrow)

PR #3994, opened 2022-04-17
Fix GradleReport Notification and potential NPE on Gradle earlier, than 6.8

lkishalmi:
image

My intention was to fix this notification, by adding an appropriate toString() method to GradleReport. Though it seems the project where I've got this was generating an NPE in LegacyProjectLoader. Which root cause was that, the recently introduced method call of getExternalVariant() was added only Gradle 6.8

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -445,9 +445,14 @@ class NbProjectInfoBuilder {
                                 // do not bother with components that only select a variant, which is itself a component
                                 // TODO: represent as a special component type so the IDE shows it, but the IDE knows it is an abstract
                                 // intermediate with no artifact(s).
-                                if (rdr.getResolvedVariant() == null ||
-                                    !rdr.getResolvedVariant().getExternalVariant().isPresent()) {
+                                if (rdr.getResolvedVariant() == null) {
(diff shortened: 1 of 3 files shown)


extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -445,10 +445,14 @@ private void detectDependencies(NbProjectInfoModel model) {
                                 // do not bother with components that only select a variant, which is itself a component
                                 // TODO: represent as a special component type so the IDE shows it, but the IDE knows it is an abstract
                                 // intermediate with no artifact(s).
-                                if (rdr.getResolvedVariant() == null ||
-                                    !rdr.getResolvedVariant().getExternalVariant().isPresent()) {
+                                if (rdr.getResolvedVariant() == null) {
                                     componentIds.add(rdr.getSelected().getId().toString());
                                 }
```
sdedic, line 450: I'd put the `sinceGradle` into `else` branch, so that executes just for `resolvedVariant != null` case.

sdedic (commented):
Apologies I didn't pay attention to `@since` ... thanks for the fixes !

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
lkishalmi: True! Thanks! Fixed.

sdedic (approved):

PR #4001, opened 2022-04-18
Fix WeakListener usage in Gradle modules

lkishalmi:
As I saw some WeakListener detachment warnings in the IDE log coming from the Gradle modules, I've decided to have a look on them.

There were to types of issues:
  1. No source object were given, so the detachment of the listener would not happen. That kind of Ok.
  2. The NbGradleProjectImpl was passed as source instead of its NbGradlePorject so the removal could not happened.

The changes here should be trivial.

### DIFF
--- a/java/gradle.java/src/org/netbeans/modules/gradle/java/queries/GradleCompilerOptionsQuery.java
+++ b/java/gradle.java/src/org/netbeans/modules/gradle/java/queries/GradleCompilerOptionsQuery.java
@@ -49,9 +49,8 @@ public final class GradleCompilerOptionsQuery implements CompilerOptionsQueryImp
 
     public GradleCompilerOptionsQuery(Project project) {
         this.project = project;
-        final NbGradleProject watcher = NbGradleProject.get(project);
         listener = (evt) -> {
-            if (watcher.isUnloadable()) return;
+            if (NbGradleProject.get(project).isUnloadable()) return;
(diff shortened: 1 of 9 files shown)


java/gradle.java/src/org/netbeans/modules/gradle/java/queries/GradleCompilerOptionsQuery.java
```diff
@@ -49,9 +49,8 @@ public final class GradleCompilerOptionsQuery implements CompilerOptionsQueryImp
 
     public GradleCompilerOptionsQuery(Project project) {
         this.project = project;
-        final NbGradleProject watcher = NbGradleProject.get(project);
```
sdedic, line 52: Consider leaving the `watcher` instance and its usage here. The listener is attached to a certain NbGradleProject instance and receives events from it. With the change, it will check the **current** NbGradleProject instance which may be different from the one that actually fires the event. But maybe it's intentional ?

sdedic (approved):

java/gradle.java/src/org/netbeans/modules/gradle/java/queries/GradleCompilerOptionsQuery.java
lkishalmi: The project here is final and the NbGradleProjectImpl has a 1 to 1 assignment of it's NbGradleProject on it's constructor. So I think we are safe here.

PR #4004, opened 2022-04-19
Smart stepping implemented for Groovy. SmartSteppingFilterWrapper introduced as a step-specific filter.

entlicher:
Smart-stepping changed to have a unique instance of `SmartSteppingFilter` per the ongoing step. Javadoc is improved accordingly.
Smart stepping is implemented for Groovy - `GroovySmartStepping`.

### DIFF
--- a/groovy/groovy.support/src/org/netbeans/modules/groovy/support/debug/GroovySmartStepping.java
+++ b/groovy/groovy.support/src/org/netbeans/modules/groovy/support/debug/GroovySmartStepping.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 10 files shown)


entlicher:
This is necessary for stepping through Groovy code in debugger.

ppisl (approved):
Thanks for this patch. I have played with this and it is much better. I have tried debug this script:
`print "hello world\n"

class Man {
    int age;

    Man (int age) {
        this.age = age
    }

    int getAge() {
        return this.age;
    }
}

def sayHello() {
    print "I'm saying HELLO"
}

sayHello()

m = new Man(10);
println (m.getAge());`

and it still jump into java.lang.Integer.java, when I invoke step into on `m=new Man(10)` line. But after finishing this method it allows to debug the constructor correctly.

entlicher:
Many thanks Petr for thorough review.
We do not have a special stepping filter for boxing/unboxing methods in Java debugger, hence the user gets there during stepping from Groovy.
We have a special logic around Step Into in Java debugger, which detects the available methods from the code. This is why you do not step into boxing methods like `Integer.valueOf()` during Java debugging. You will get there if "Step Into Next Method" (Shift-F7) is used, which is used in Groovy. Groovy has no special method-selection logic implemented.

ebarboni:
rebased to delivery and green so merging

PR #4007, opened 2022-04-19
new nb module project wizard improvements.

mbien:
- use dev snapshot as fallback instead of RELEASE90
 - sort the version list properly, newest first
 - filter out non-apache releases

second commit is the usual language lvl cleanup, first commit has the fixes.

### DIFF
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NbmWizardPanelVisual.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NbmWizardPanelVisual.java
@@ -22,22 +22,17 @@ package org.netbeans.modules.maven.apisupport;
 import java.awt.EventQueue;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 import javax.swing.DefaultComboBoxModel;
 import javax.swing.SwingUtilities;
-import org.apache.maven.artifact.resolver.ArtifactNotFoundException;
-import org.apache.maven.artifact.resolver.ArtifactResolutionException;
-import org.netbeans.api.progress.aggregate.AggregateProgressFactory;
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
Requested a review from @ebarboni - need to branch, but this would be a good fix to get in before release - will need rebasing on delivery once that is set up.

ebarboni:
Hi, I wonder if we can rename the constant RELEASE90_VERSION to something like "Minimum Version" and adjust to RELEASE111 (first out of incubation)

mbien:
> Hi, I wonder if we can rename the constant RELEASE90_VERSION to something like "Minimum Version" and adjust to RELEASE111 (first out of incubation) (edited) Order is well done but we should add an ignore list (for example, RELEASE120 should not be proposed, cluster pom were invalid and make project unbuildable RELEASE120-1 should be proposed )

@ebarboni sure, i can do that. As far as i see, 90 is the last release with osgi support, this means we could basically remove the entire check box if we only list >=111 (minus ignore list).

mbien:
@ebarboni there is a `RELEASE120-1` and a `RELEASE120-u1` should I put one of them on the ignore list?

mbien:
- added filter for >=111 and put 120 on the ignore list as suggested.
- osgi checkbox is set to invisible, just in case someone wants to use that in future

mbien:
rebased onto delivery just in case it is decided to include it.

ebarboni:
LGTM merging

mbien:
@ebarboni this PR is *not* in rc2, right? I am just want to be sure that I am not crazy. I just decompiled apisupport and I saw the old file.

its also not listed here:
https://github.com/apache/netbeans/pull/4032/commits

ebarboni:
arf, I see that not on master nor on release140. I have no idea what happen. sync pr created too late ?

mbien:
I don't know. but we have to check if its only this PR or if more is missing.

mbien:
@ebarboni This PR is also not in delivery:
https://github.com/apache/netbeans/commits/delivery

this explains why its not in master. Did someone force push into delivery?

ebarboni:
check a bit but only this PR + #4004 seems the missing one.

PR #4010, opened 2022-04-20
updated nb-javac18 plugin version and description

singh-akhilesh:


neilcsmith-net:
NetBeans 14 has branched.  This might have been nice to resolve before branching - now spec version on master has incremented this will be a slight complication to merge into delivery.

mbien:
we should put this on the list of "things which need to be changed every time nb-javac is updated" #3953 would be another item

mbien (approved):
this looks good to me, travis is working now again too. I agree that we should try to get this into delivery because this has a good chance to confuse users.

I am no release manager though, so maybe rebase it for delivery but wait for someone else to merge it.

as sidenote: this is luckily just "cosmetics" because the indexer doesn't use the module version as version stamp anymore. It uses` SourceVersion.latest()` since NB 13 so things *should* keep working even with a wrong module version.

mbien:
just a thought: we might integrate javac of 18.0.1 during RC phase. Maybe we should wait for the increment and do everything at once. @jtulach

neilcsmith-net:
@mbien the spec version already differs between master and delivery due to the spec version increment at branch, so as said earlier, this will need a little care with the sync PR if merging via delivery first.   cc/ @ebarboni @MartinBalin @geertjanw

ebarboni:
@neilcsmith-net when this merged to master we need to bump the version to master too be superior and not equals ?

singh-akhilesh:
> @neilcsmith-net when this merged to master we need to bump the version to master too be superior and not equals ?

Keep the same version as defined in this PR change-set.

neilcsmith-net:
@ebarboni this went into master anyway.

@singh-akhilesh all module versions get incremented once per release, but that was all done at the time of branching. NB 14 will have wrong version at the moment though.

PR #4012, opened 2022-04-20
remove jdk18 reflection calls and updated the usages

singh-akhilesh:


neilcsmith-net:
NetBeans 14 has branched. Moved to NB15 milestone.  If really required for NB14, please rebase on delivery first and re-add the milestone.

java/java.editor/test/unit/src/org/netbeans/modules/editor/java/GoToSupportTest.java
```diff
@@ -1581,7 +1581,10 @@ private static boolean hasRecords() {
     }
 
     public void testJavadocSnippetHighlightRecord() throws Exception {
-        if (!TreeShims.isJDKVersionRelease18_Or_Above()) {
+        try {
+            SourceVersion.valueOf("RELEASE_18"); //NOI18N
+        } catch (IllegalArgumentException ex) {
+            //OK, no RELEASE_18, skip tests
             return;
         }
```
mbien, line 1589: does this mean the tests don't use nb-javac? Would it be better to enable nb-javac?

mbien (approved):
looks good to me thanks. Unfortunately there still seem to be some issues with travis. 

Maybe we should wait a bit before integration. But this is basically a cleanup, so there is no need to hurry I believe.

java/java.editor/test/unit/src/org/netbeans/modules/editor/java/GoToSupportTest.java
singh-akhilesh: If nb-javac-18 plugin installed & enabled it uses nb-javac otherwise it uses the JDK Source version. If JDK source version is below 18 then this test won't be execute.

singh-akhilesh:
> looks good to me thanks. Unfortunately there still seem to be some issues with travis.
> 
> Maybe we should wait a bit before integration. But this is basically a cleanup, so there is no need to hurry I believe.

Yes, there is no hurry. We can integrate later point of time.

java/java.editor/test/unit/src/org/netbeans/modules/editor/java/GoToSupportTest.java
mbien: yes. but tests can be configured to use nb-javac.
looking at the properties:
https://github.com/apache/netbeans/blob/master/java/java.editor/nbproject/project.properties#L113-L114

it looks like nb-javac is already enabled for the tests. It should work without this check I believe.

singh-akhilesh: Yes, then I believe we can remove other conditional checks also of this class are _SourceVersion.valueOf("RELEASE_10");_ and _SourceVersion.valueOf("RELEASE_14");_

mbien: yep, it should be possible to remove the javac version checks and the reflective code and change everything to direct calls.

Even if the user uninstalls nb-javac, NB has a runtime dependency check which will tell the user to run on a specific JDK or install nb-javac again. You can see it here: https://github.com/apache/netbeans/pull/3953/files

arvindaprameya:
Approved thank you for completing , pls resolve the conflicts

mbien:
@singh-akhilesh could you remove this commit: 18422242257a1c22937adf0fe9f22c962947a167 and rebase? Would be nice to get this PR in. The less reflection we have the better.

mbien (approved):
thanks, looks good.

singh-akhilesh:
> thanks, looks good.

Okay, will merge this.

PR #4015, opened 2022-04-21
disable failing tests until resolved.

mbien:
one fails, the other errors. Lets see if disabling the two would allow the rest to pass.

so that we can have a green delivery/master until the issue is resolved.


mbien:
@neilcsmith-net pr is based on delivery.

jlahoda:
Unfortunate, but probably needed - I'll try to look at the tests over the weekend.

neilcsmith-net (approved):
OK, let's get this in, and I'll do a quick sync to master.

@jlahoda if you do look this weekend then your call whether to merge straight to master or via delivery - depends whether re-enabling these tests for the release is desirable?

PR #4018, opened 2022-04-22
Sync #4015 from delivery to master

neilcsmith-net:
Sync #4105 to master. cc @mbien - feel free to merge.

PR #4031, opened 2022-04-25
Sync delivery to master after 14-rc2

ebarboni:


neilcsmith-net:
Should be after 14-rc2? Master should be in sync with delivery + #4015 already unless I missed something, so you shouldn't need to sync until rc2 has been announced.

ebarboni:
@neilcsmith-net you're right renaming is enough ? or should it be recreated later ?

neilcsmith-net:
@ebarboni name changes on both to rc2 are fine. :+1: Hadn't seen the sync to release140 when I commented.  Making sure you knew I'd run the extra sync in #4018 already.

PR #4037, opened 2022-04-27
[NETBEANS-5523] PHP 8.0: Treat namespaced names as single token

junichi11:
We can fix the highlighting for keyword parts of namespaced names if we add new tokens.
However, many existing features using the lexer will be broken.
So, We won't fix it at the moment.

- https://issues.apache.org/jira/browse/NETBEANS-5523
- https://wiki.php.net/rfc/namespaced_names_as_token

- Fix the parser and the lexer
- Fix/Add unit tests

nb-php80-treat-namespaced-names-as-single-token


junichi11:

php/php.editor/src/org/netbeans/modules/php/editor/parser/astnodes/NamespaceName.java
```diff
@@ -64,6 +70,24 @@ public NamespaceName(int start, int end, List segments, boolean global, boolean
         this.current = current;
     }
 
+    @NonNull
+    public static NamespaceName create(int start, int end, @NonNull String namespaceName) {
+        boolean isGlobal = namespaceName.startsWith("\\"); // NOI18N
+        boolean isCurrent = namespaceName.startsWith("namespace\\"); // NOI18N
+        String[] names = namespaceName.split("\\\\"); // NOI18N
+        int startSegment = start;
+        List<Identifier> list = new ArrayList<>();
```
tmysik, line 79: Nitpick:
```suggestion
        List<Identifier> list = new ArrayList<>(names.size());
```

junichi11: Will fix it next time :) Thanks!

junichi11:

php/php.editor/src/org/netbeans/modules/php/editor/parser/astnodes/NamespaceName.java
tmysik: Sure, it is really a nitpick :)

PR #4040, opened 2022-04-27
LSP: Html4J UI modified work over LSP.

dbalek:
An attempt to rewrite `html4j` based refactoring UI for VSCode to work over LSP.


java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Buttons.java
```diff
@@ -29,16 +29,15 @@ private Buttons() {
     }
 
     @JavaScriptBody(args = {}, body = "\n"
-        + "const vscode = acquireVsCodeApi();\n" // this method can be called only once per WebView existance
         + "window.close = function() {\n"
-        + "  vscode.postMessage({\n"
-        + "    command: 'dispose',\n"
+        + "  self.vscode.postMessage({\n"
```
jtulach, line 33: `self` isn't a defined variable.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
```diff
@@ -18,538 +18,168 @@
...
-        s.setCharacterEncoding(r, "UTF-8");
-        s.addHeader(r, "Access-Control-Allow-Origin", "*");
-        s.addHeader(r, "Access-Control-Allow-Credentials", "true");
-        s.addHeader(r, "Access-Control-Allow-Headers", "Content-Type");
-        s.addHeader(r, "Access-Control-Allow-Methods", "GET, POST, DELETE, PUT");
+    private void emitScript(StringBuilder sb, String id) throws IOException {
+        sb.append("<script id='exec' type='text/javascript'>\n"
+                + "(function () {\n"
+                + "  const self = this;\n"
+                + "  this.vscode = acquireVsCodeApi();\n"
```
jtulach, line 133: The usual recommendation is to "hide" the `vscode` API object in a local variable to avoid its misuse by untrusted code running in the browser. E.g. I'd store it only in `const vscode = ` rather than putting it into `this.vscode`.

jtulach:
With [be02cff](https://github.com/apache/netbeans/pull/4040/commits/be02cffbe1fac920a47b20a0b21c71ac1312a668) one should be able to stop the Java execution before sending a message to VSCode:

image

And just by setting `pause = true` in the evaluation window the execution will then stop in chrome dev tools in VSCode:

image

JaroslavTulach:
The _Change Refactoring UI_ seems to be working with `generic` and `ko4j` modules from https://github.com/apache/netbeans-html4j/commit/f031ed4df296ac3213dca79a32d4737c2bcbab26 - I'll prepare a release 1.8.1.

jtulach:
HTML/Java 1.8.1 is available on [Maven central](https://repo1.maven.org/maven2/org/netbeans/html/ko4j/1.8.1/). @dbalek please replace 6c9ab9a with proper external binaries. I hope this will make it into NetBeans 15.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Buttons.java
dbalek: Fixed.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
dbalek: Fixed.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
```diff
@@ -18,538 +18,167 @@
...
- *   &lt;artifactId&gt;browser&lt;/artifactId&gt;
- *   &lt;version&gt;<a target="blank" href="http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.dukescript.presenters%22%20AND%20a%3A%22browser%22">1.x</a>&lt;/version&gt;
- * &lt;/dependency&gt;
- * </pre>
- */
-public final class Browser implements Fn.Presenter, Fn.KeepAlive, Flushable,
-Executor, Closeable {
-    static final Logger LOG = Logger.getLogger(Browser.class.getName());
-    private final Map<String,Command> SESSIONS = new HashMap<>();
+public class Browser implements Closeable {
```
jtulach, line 56: I suggest to rename this class. It is no longer about "browser".

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
```diff
@@ -18,538 +18,167 @@
...
-    static <Response> void cors(HttpServer<?, Response, ?, ?> s, Response r) {
-        s.setCharacterEncoding(r, "UTF-8");
-        s.addHeader(r, "Access-Control-Allow-Origin", "*");
-        s.addHeader(r, "Access-Control-Allow-Credentials", "true");
-        s.addHeader(r, "Access-Control-Allow-Headers", "Content-Type");
-        s.addHeader(r, "Access-Control-Allow-Methods", "GET, POST, DELETE, PUT");
+    private void emitScript(StringBuilder sb, String id) throws IOException {
+        
+        sb.append("<script id='exec' type='text/javascript'>\n"
+                + "const vscode = acquireVsCodeApi();\n"
```
jtulach, line 132: This is a potential security threat. VSCode team tries hard not to allow untrusted scripting code access the API. However, right now your are only running trusted code - but avoid downloading any HTML file or JS file from network.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
```diff
@@ -18,538 +18,167 @@
...
-            );
-        }
-        sb.append(""
-            + "return txt;\n"
-            + "};\n"
-        );
-        return sb.toString();
+    private String createCallbackFn(String id) {
+        return "this.toBrwsrSrvr = function(name, a1, a2, a3, a4) {\n"
+             + "  vscode.postMessage({\n"
```
jtulach, line 169: You could use `vscode` here and then delete it. Just wrap into 
```
(function() {
  var vscode = globalThis.vscode;
  delete globalThis.vscode;
  
  globalThis.toBrwsrSrvr = ....
)();
```
after deleting `vscode` from global scope the page could even load _potentially untrusted code_.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Buttons.java
```diff
@@ -29,16 +29,15 @@ private Buttons() {
     }
 
     @JavaScriptBody(args = {}, body = "\n"
-        + "const vscode = acquireVsCodeApi();\n" // this method can be called only once per WebView existance
         + "window.close = function() {\n"
         + "  vscode.postMessage({\n"
```
jtulach, line 33: Another usage of `vscode` - one would need to capture it into a local variable and then delete it:
```
(function() {
  var vscode = globalThis.vscode;
  delete globalThis.vscode;
  
  globalThis.toBrwsrSrvr = ....
)();
```

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/refactoring/ChangeMethodParametersRefactoring.java
```diff
@@ -187,7 +187,7 @@ private static String defaultValue(String type) {
         return "null";
     }
 
-    @HTMLDialog(url = "ui/ChangeMethodParameters.html")
+    @HTMLDialog(url = "ui/ChangeMethodParameters.html", resources = {"refactoring.css"})
```
jtulach, line 190: `resources` are good API extension.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
```diff
@@ -121,11 +118,14 @@ private void load() {
                 notifyClose();
                 return;
             }
-            URL pageUrl = ctx.getPage();
-            Browser.Config c = initial.clone();
+            Browser.Config c = new Browser.Config();
```
jtulach, line 121: Is the `Config` used for anything? Can't it be deleted?

java/java.lsp.server/vscode/package.json
```diff
@@ -1001,6 +1001,7 @@
 		"apisupport": "node ./out/nbcode.js -J-Dnetbeans.close=true --modules --install '(org.netbeans.libs.xerces|org.netbeans.modules.editor.structure|org.netbeans.modules.xml|org.netbeans.modules.xml.axi|org.netbeans.modules.xml.retriever|org.netbeans.modules.xml.schema.model|org.netbeans.modules.xml.tax|org.netbeans.modules.xml.text|org.netbeans.modules.ant.browsetask|.*apisupport.*|org.netbeans.modules.debugger.jpda.ant)' && node ./out/nbcode.js -J-Dnetbeans.close=true --modules --enable .*apisupport.ant"
 	},
 	"devDependencies": {
+		"@vscode/codicons": "0.0.29",
```
jtulach, line 1004: Good. Instead of carrying our own `codicons` we can just have a dependency.

platform/api.htmlui/src/org/netbeans/spi/htmlui/HTMLViewerSpi.java
```diff
@@ -190,6 +192,15 @@ public ClassLoader getClassLoader() {
             return loader;
         }
 
+        /** List of resources available to the page.
+         *
+         * @return resources used in the page displayed by the HTML user interface
+         * @since 1.25
+         */
+        public URL[] getResources() {
```
jtulach, line 200: Are we sure we want `URL[]` and not `String[]`? Isn't the relative path to the main page important?

jtulach (approved):
Looks great and if it works for VSCode, then perfect!

In addition I'd like to see usage of `resources` in the JavaFX WebView implementation to load the `refactoring.css`. Probably the implementation shall create a temporary directory and copy the resources there to get `file` protocol URL. Right now the SPI returns `URL[]` and I am afraid it it is going to be of little use - possibly `String[]` of relative paths to main page would be better. The JavaFX WebView implementation may wait for another PR, but we don't want to create an SPI that is wrong to begin with.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/htmlui/Browser.java
dbalek: Renamed to `WebView`.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ui/AbstractLspHtmlViewer.java
dbalek: Deleted.

platform/api.htmlui/src/org/netbeans/spi/htmlui/HTMLViewerSpi.java
dbalek: Fixed.

PR #4041, opened 2022-04-27
Proxy selector fixes

sdedic:
It seems we missed some API changes in truffle and are not exporting a required API package. Because of that, when NB starts on JDK11 _and_ in an environment that provides `http://wpad/wpad.dat`, the following exception is printed at startup:
```
WARNING [org.netbeans.core.network.proxy.ProxyAutoConfig]: There was a catastrophic error with the PAC script downloaded from http://wpad/wpad.dat. Will use dummy instead. Error was : 
org.graalvm.polyglot.PolyglotException: java.lang.NoClassDefFoundError: com/oracle/truffle/api/exception/AbstractTruffleException while loading com.oracle.truffle.regex.RegexSyntaxException; see http://wiki.netbeans.org/DevFaqTroubleshootClassNotFound
        at org.netbeans.ProxyClassLoader.selfLoadClass(ProxyClassLoader.java:250)
        at org.netbeans.ProxyClassLoader.doFindClass(ProxyClassLoader.java:174)
        at org.netbeans.ProxyClassLoader.loadClass(ProxyClassLoader.java:125)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
        at com.oracle.truffle.regex.RegexEngineBuilder.execute(RegexEngineBuilder.java:117)
        at com.oracle.truffle.regex.RegexEngineBuilderGen$InteropLibraryExports$Uncached.execute(RegexEngineBuilderGen.java:197)
        at com.oracle.truffle.api.interop.InteropLibrary$Asserts.execute(InteropLibrary.java:2279)
        at com.oracle.truffle.api.interop.InteropLibraryGen$UncachedDispatch.execute(InteropLibraryGen.java:4780)
        at com.oracle.truffle.js.runtime.JSContext.createTRegexEngine(JSContext.java:998)
        at com.oracle.truffle.js.runtime.JSContext.getRegexEngine(JSContext.java:978)
        at com.oracle.truffle.js.runtime.RegexCompilerInterface.validate(RegexCompilerInterface.java:77)
...
Caused by: java.lang.NoClassDefFoundError: com/oracle/truffle/api/exception/AbstractTruffleException
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
        at org.netbeans.JarClassLoader.doLoadClass(JarClassLoader.java:287)
[...]

PR #4042, opened 2022-04-28
Strip HTML tags from project problem messages

sdedic:
The messages are reported through LSP request that does not support html or markdown, html tags should be stripped.


neilcsmith-net:
No need to add release team for review, just milestone, and suggest @ebarboni or @MartinBalin anyway (I'm not doing remaining RCs for this one).

MartinBalin (approved):

sdedic:
Perhaps my mistake; rebased on delivery, single commit remains.

ebarboni:
thanks may you do it for #4041 also

PR #4045, opened 2022-04-28
Adjust spec version to differ from Apache NetBeans 14 for java/libs.nbjavac

ebarboni:
Related to #4040

### DIFF
--- a/java/libs.nbjavacapi/manifest.mf
+++ b/java/libs.nbjavacapi/manifest.mf
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 AutoUpdate-Show-In-Client: true
 OpenIDE-Module: org.netbeans.libs.nbjavacapi
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/nbjavac/api/Bundle.properties
-OpenIDE-Module-Specification-Version: 18.0
+OpenIDE-Module-Specification-Version: 18.1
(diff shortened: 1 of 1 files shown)


singh-akhilesh (approved):
LGTM

java/libs.nbjavacapi/manifest.mf
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 AutoUpdate-Show-In-Client: true
 OpenIDE-Module: org.netbeans.libs.nbjavacapi
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/nbjavac/api/Bundle.properties
-OpenIDE-Module-Specification-Version: 18.0
+OpenIDE-Module-Specification-Version: 18.1
```
mbien, line 5: could this be 18.0.0.1? Because there might be a 18.0.1 nb-javac.

neilcsmith-net: That would be ideal, but checking 8da225d7f7cee699077fa015d5355fe2d03a449a the spec version task always increments the second number even when there are further dots, so I think keeping the major number in sync is about all we can do.  Perhaps look at `18.1.1`, `18.1.2` if we really want to have the JDK update in there too?? :shrug:

neilcsmith-net (approved):
LGTM

java/libs.nbjavacapi/manifest.mf
mbien: we have to write this somewhere down. Because this all looks like an accident in the making :)

PR #4047, opened 2022-04-28
Sampler module has a lot of obsolete dependencies

jtulach:
We are trying to use [sampler module in our application](https://github.com/enso-org/enso/pull/3389), but we realized it has a lot of unnecessary dependencies. Ideally we just use the `org-netbeans-modules-sampler.jar` on its own - should in most cases work until one calls into `InternalSampler` - we don't.

However even if one uses `InternalSampler` there are some obsoleted dependencies that make the module [too heavyweight](http://wiki.apidesign.org/wiki/HeavyWeight). This PR clears them:
- using just simpler progress API
- using `FileObject.getLookup()`
- avoiding dialogs in favor of "exception like" report

### DIFF
--- a/platform/sampler/nbproject/project.properties
+++ b/platform/sampler/nbproject/project.properties
@@ -16,6 +16,6 @@
 # under the License.
 is.autoload=true
 javac.source=1.8
-javac.compilerargs=-Xlint -Xlint:-serial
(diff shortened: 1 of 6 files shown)


ebarboni:
LGTM merging

PR #4052, opened 2022-04-29
Alternative Standalone Sampler

jtulach:
Continuous work on https://github.com/enso-org/enso/pull/3389 has revealed that [it is not possible to exclude](https://github.com/enso-org/enso/pull/3389/commits/3fd211cf318448a0bb03a67e97c4219b78e4595e) five NetBeans libraries when using sampler module standalone. To address that this PR continues in https://github.com/apache/netbeans/pull/4047:

- d66ce57938eea2037648bb7ecc4d5f121476b598 duplicates the tests to check not only the `InternalSampler`, but also existing `CLISampler`
- 9f4ec39ed761ce5950a51b370061ed0f85e2aaf2 adds yet another `StandaloneSampler` and its test

The test creates own `ClassLoader` and loads just the sampler JAR, without dependencies on other NetBeans modules (so `Lookup` isn't available) and then performs the `AbstractSamplerBase` tests. This guarantees that all three samplers behaves the same.


neilcsmith-net:
Curious why this is going into delivery and not master?

JaroslavTulach:
> Curious why this is going into delivery and not master?

I need Maven bits, so I'd like to get this into NetBeans 14.

matthiasblaesing:
> > Curious why this is going into delivery and not master?
> 
> I need Maven bits, so I'd like to get this into NetBeans 14.

I think the point that @neilcsmith-net was trying to make is, that this does not look like a PR helping to stabilize NB14.  The description of the issues shows IMHO that these are feature changes, so that another dependency gets a lower number of dependencies.

neilcsmith-net:
Yes, exactly what @matthiasblaesing said!  This doesn't seem to fit the normal criteria for merges to delivery.  Not that there's a problem with bending the rules occasionally, just like to see the reasoning (cost/benefit) in the PR for doing so.

PR #4053, opened 2022-04-29
Micronaut Data finder completion should work for record entities.

dbalek:
Micronaut Data finder completion should work in a case when the corresponding entity is Java record (not a class).

### DIFF
--- a/enterprise/micronaut/src/org/netbeans/modules/micronaut/completion/MicronautDataCompletionTask.java
+++ b/enterprise/micronaut/src/org/netbeans/modules/micronaut/completion/MicronautDataCompletionTask.java
@@ -41,6 +41,7 @@ import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.element.RecordComponentElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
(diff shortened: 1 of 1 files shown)


MartinBalin (approved):

enterprise/micronaut/src/org/netbeans/modules/micronaut/completion/MicronautDataCompletionTask.java
```diff
@@ -249,9 +250,16 @@ private <T> void resolveFinderMethods(CompilationController cc, TreePath path, S
             for (ExecutableElement method : ElementFilter.methodsIn(entity.getEnclosedElements())) {
                 String methodName = method.getSimpleName().toString();
                 if (methodName.startsWith(GET) && method.getParameters().isEmpty()) {
-                    prop2Types.put(methodName.substring(GET.length()), tu.getTypeName(method.getReturnType()).toString());
+                    methodName = methodName.substring(GET.length());
+                    methodName = methodName.substring(0, 1).toUpperCase() + methodName.substring(1);
```
sdedic, line 254: Missing case for `get` method with no suffix -> no property defined. Use `toUpperCase` with locale (ENGLISH or ROOT -- just found it exists :))

enterprise/micronaut/src/org/netbeans/modules/micronaut/completion/MicronautDataCompletionTask.java
```diff
@@ -249,9 +250,16 @@ private <T> void resolveFinderMethods(CompilationController cc, TreePath path, S
...
+                    prop2Types.put(methodName, tu.getTypeName(method.getReturnType()).toString());
```
sdedic, line 255: I'd check the `method.getReturnType()` to be valid / not erroneous, just in case - with unresolved types, generics may work / display bad

dbalek: Fixed.

PR #4069, opened 2022-05-03
new nb module project wizard improvements.

ebarboni:
#4007 second merge attempt


mbien:
may i propose an alternative PR :)
#4070 has both commits, since I try to split cleanup from bugfix if somehow possible.

I simply created another pr from the original PR branch, so it should be identical to the original version, just applied later.

ebarboni:
yes sure it was just resubmit but if you do it yourself that's no pb

PR #4072, opened 2022-05-04
Allow tests to access JPDA. Pass mx path to the testsuite.

sdedic:
Yet another fixes to fix `java.mx.project` testsuite on JDK11. Potentially other testsuites, too
- tests that access JPDA API should be now failing on JDK11, as `--limit-modules` does not allow them to see JPDA classes. This should be less controversial :) as we definitely want to access debugger support in platform from java cluster at least.
- `java.mx.project` testsuite needs `mx` binary on PATH (or know full path to it); the buildscript `git clone`s one, but it's not on path. Moreover the copy made into `build/test/unit/data` directory omits executable attribute, so the git clone sources must be used directly

Note that passing the `mx` location in system properties should also fix tests on JDK8 -- see for example [Test Java modules with nb-javac on Java 8](https://api.travis-ci.com/v3/job/568921174/log.txt):
```
    [junit] java.io.IOException: error=2, No such file or directory
    [junit] 	at java.lang.UNIXProcess.forkAndExec(Native Method)
    [junit] 	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
    [junit] 	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
    [junit] 	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
    [junit] Caused: java.io.IOException: Cannot run program "mx" (in directory "/home/travis/build/apache/netbeans/java/java.mx.project/build/test/unit/data/graal/sdk"): error=2, No such file or directory
    [junit] 	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
    [junit] 	at org.netbeans.modules.extexecution.base.ExternalProcessBuilder.call(ExternalProcessBuilder.java:272)
    [junit] 	at org.netbeans.api.extexecution.base.ProcessBuilder$LocalProcessBuilder.createProcess(ProcessBuilder.java:301)
    [junit] 	at org.netbeans.api.extexecution.base.ProcessBuilder.call(ProcessBuilder.java:252)
    [junit] 	at org.netbeans.api.extexecution.base.ProcessBuilder.call(ProcessBuilder.java:71)
```
(but the test still completes successfully; strage ... will fix in a subsequent PR)

### DIFF
--- a/java/java.mx.project/nbproject/project.properties
+++ b/java/java.mx.project/nbproject/project.properties
@@ -19,3 +19,6 @@
 javac.source=1.8
 javac.compilerargs=-Xlint -Xlint:-serial -Werror -Xlint:-processing
 requires.nb.javac=true
+
+# For testing we need path to the MX executable. The executable is checked out by the buildscript.
+test.run.args=-Dorg.netbeans.modules.java.mx.project.test.mxpath=${basedir}/test/unit/data/mx/mx
--- a/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
+++ b/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteActionProvider.java
@@ -48,6 +48,7 @@ import org.openide.windows.OutputEvent;
 import org.openide.windows.OutputListener;
 
 final class SuiteActionProvider implements ActionProvider {
+    private static final String TEST_MX_PATH = System.getProperty("org.netbeans.modules.java.mx.project.test.mxpath"); // NOI18N
     private static final RequestProcessor ASYNC = new RequestProcessor("Mx Async", 10);
     private static final List<String> SUPPORTED_ACTIONS = Arrays.asList(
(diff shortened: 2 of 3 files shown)


mbien (approved):
looks good to me, no complains :)

One question though, you mentioned the tests were failing on JDK 11 prior to this PR.
Does CI not run this scenario? This sounds like something we should add.

dbalek (approved):

sdedic:
> One question though, you mentioned the tests were failing on JDK 11 prior to this PR. Does CI not run this scenario? This sounds like something we should add.
> 

Yes; tests in https://ci-builds.apache.org/job/Netbeans/view/vscode/job/netbeans-vscode/1061/ fail because of this. The same suite runs with JDK8 (= no module system) routinely.
And yes, we should definitely run more testsuites on JDK11 :) but that should be probably a separate task/effort focused just on this. Right now I'd like to fix the vscode master builds at least.

PR #4079, opened 2022-05-06
improved tests for NBBundle class

lbownik:
refactored NBBundle with some performance improvements
no behavior changes

### DIFF
--- a/platform/openide.util/src/org/openide/util/AnnotatedResourceInputStream.java
+++ b/platform/openide.util/src/org/openide/util/AnnotatedResourceInputStream.java
@@ -1,0 +1,286 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


vieiro (changes requested):
Cannot be merged. You may want to `git fetch` from NetBeans repo (say "upstream") and `git merge upstream/master`, resolve conflicts and `git push`.

mbien (changes requested):
please don't do non trivial cleanups in the same commit as actual implementation changes. Why did you extract the inner classes? A reviewer would have to manually diff the code to figure out what happened. The code history of it is also gone since it is tracked as new files.

Also avoid rewriting tests in the same commit as implementation changes. How do we know if the old tests are green with your new implementation?

This is a lot of avoidable obfuscation and makes it risky to integrate and difficult to review. Someone might have to fix a bug in future using git bisect and land on this commit - it is not obvious what the improvement is or what was actually changed.

PR #4080, opened 2022-05-06
code and test cleaning

lbownik:
minor performance improvements
code upgrade to Java 8

### DIFF
--- a/platform/openide.util/src/org/openide/util/ChangeSupport.java
+++ b/platform/openide.util/src/org/openide/util/ChangeSupport.java
@@ -21,7 +21,7 @@ package org.openide.util;
 
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.logging.Level;
+import static java.util.logging.Level.*;
 import java.util.logging.Logger;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
@@ -38,7 +38,7 @@ public final class ChangeSupport {
     private static final Logger LOG = Logger.getLogger(ChangeSupport.class.getName());
 
     // not private because used in unit tests
-    final List<ChangeListener> listeners = new CopyOnWriteArrayList<ChangeListener>();
+    final List<ChangeListener> listeners = new CopyOnWriteArrayList<>();
     private final Object source;
 
     /**
@@ -46,7 +46,8 @@ public final class ChangeSupport {
      *
      * @param  source the instance to be given as the source for events.
      */
-    public ChangeSupport(Object source) {
+    public ChangeSupport(final Object source) {
+        
         this.source = source;
     }
 
@@ -58,14 +59,15 @@ public final class ChangeSupport {
      *
      * @param  listener the <code>ChangeListener</code> to be added.
      */
-    public void addChangeListener(ChangeListener listener) {
-        if (listener == null) {
-            return;
+    public void addChangeListener(final ChangeListener listener) {
+
+        if (listener != null) {
+            if (LOG.isLoggable(FINE) && this.listeners.contains(listener)) {
+                LOG.log(FINE, "diagnostics for #167491", 
+                        new IllegalStateException("Added " + listener + " multiply"));
+            }
+            this.listeners.add(listener);
         }
-        if (LOG.isLoggable(Level.FINE) && listeners.contains(listener)) {
-            LOG.log(Level.FINE, "diagnostics for #167491", new IllegalStateException("Added " + listener + " multiply"));
-        }
-        listeners.add(listener);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


platform/openide.util/src/org/openide/util/ChangeSupport.java
```diff
@@ -119,6 +108,7 @@ private void fireChange(ChangeEvent event) {
      *         false otherwise.
      */
     public boolean hasListeners() {
-        return !listeners.isEmpty();
+        
+        return !this.listeners.isEmpty();
```
vieiro, line 112: `this` is unnecessary here, right? Why add it then?

platform/openide.util/src/org/openide/util/NbCollections.java
```diff
@@ -34,9 +33,9 @@
 import java.util.NoSuchElementException;
 import java.util.RandomAccess;
 import java.util.Set;
-import java.util.logging.Level;
+import static java.util.logging.Level.WARNING;
```
vieiro, line 36: We use `import static` sparingly. Not really requierd this change, right?

platform/openide.util/src/org/openide/util/NbCollections.java
```diff
@@ -34,9 +33,9 @@
...
 import java.util.logging.Logger;
-import org.openide.util.Enumerations;
+import static java.util.Objects.requireNonNull;
```
vieiro, line 38: Again another cosmetic change.

vieiro (changes requested):
This PR is too long to review. As far as I can see this is just cosmetic changes to support a new JDK version, without real improvements, right? Please clarify.

platform/openide.util/src/org/openide/util/ChangeSupport.java
```diff
@@ -21,7 +21,7 @@
 
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.logging.Level;
+import static java.util.logging.Level.*;
```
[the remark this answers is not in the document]
lbownik: My reasoning is:
logger.log(FINE,..... is more compact than
logger.log(Level.FILE, .......
so it reads better (phrase "Level." is a noise here).
Since log method ony accepts Level objects as first parameter there is no doubt what "FINE" is and where it comes from (well .. absolute Java beginners might have doubts the first time ... but I don't think they contribute to NetBeans).

A counter example would be turning "Paths.get("C:\.....")" into "get(C:\...")". In this case uning static import would decrease readability.

platform/openide.util/src/org/openide/util/ChangeSupport.java
```diff
@@ -38,15 +38,16 @@ public final class ChangeSupport {
...
+    final List<ChangeListener> listeners = new CopyOnWriteArrayList<>();
     private final Object source;
 
     /**
      * Creates a new <code>ChangeSupport</code>
      *
      * @param  source the instance to be given as the source for events.
      */
-    public ChangeSupport(Object source) {
+    public ChangeSupport(final Object source) {
```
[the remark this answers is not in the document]
lbownik: ok, I'm not religious about "final" - too bad "final" is not implicit in Java.

platform/openide.util/src/org/openide/util/ChangeSupport.java
lbownik: ok

lbownik:
"As far as I can see this is just cosmetic changes to support a new JDK version, without real improvements, right?"

Yes:  The PR does not change the behavior.
No: 
1. There are small improvements in readability (I can revert finals's and this'es if you insist). 
2. There are small performance improvements:
  2.1 I removed needles "ChangeSupport.fireChange(event)" method and merged it's body with the public "fireChange().
  2.2 I relplaced iterators with lambdas in NBCollections.java (lambdas are 2x farster on my machine [Win 10 + Oracle JDK8]), which additionally shaved off 2 lines of code per change (readability).
  2.3 I replaces 2 anonymous classes  with lambdas in NBCollections.java (no need for class lodaer to load these any more).

matthiasblaesing:
I think @vieiro tried to make the point, that this PR is to much noise for to little gain. Cosmetic changes by us nothing, but take time to review and see where breaks might have been introduced. PRs should fix problems, not cosmetics.

lbownik:
> I think @vieiro tried to make the point, that this PR is to much noise for to little gain."

I do understand the workload issue - I will not propose any solutions to this as I'm new to community and I know too little.

" Cosmetic changes by us nothing"
My experience tells me otherwise - often small cosmentic changes change the way I see the problem and help me find better solutions in the end. I can make small laser-focused PR's if this helps.

", but take time to review and see where breaks might have been introduced."
That's why we have unit tests. If this helps, I can intriduce separate PRs,  one for tests cleaning and code cleaning.
I would submit code cleaning PR only after test cleaning PR gets merged. 

"PRs should fix problems, not cosmetics."
I've seen this approach in action - it leads to code rot and spaghetti (and bancrupcy in case of commercial software).
Sustainable development is all about entropy, there needs to be constant conscious effort to reduce it, otherwise it kills eventually every project. 

So would smaller PRs which tests first/code second approach work better?
I woud also skip final's and this'es.

vieiro:
> I think @vieiro tried to make the point, that this PR is to much noise for to little gain. Cosmetic changes by us nothing, but take time to review and see where breaks might have been introduced. PRs should fix problems, not cosmetics.

Exactly. Cosmetic changes add noise for little gain *and* introduce some risk in the implementation. NetBeans has >500k lines of code, and we use `import static`sparingly. Same applies for using the `this.` prefix, we use this sparingly in those other >500k files.

If you feel an urge to use `import static` for logging levels then it's better to have a single PR do this in 500k files than having 500k PRs changing each file. 

Also modifying the implementation (of classes that are very sensitive to changes, since this module is probably used by everybody) and the tests at the same time is weird/risky. It's more difficult to know if the new code coming from somebody we don't know (or, as you say, _"knows too little"_) behaves the same as before. Until you gain experience I'd prefer PRs with tests only and PRs with code changes.

PR #4081, opened 2022-05-06
improved tests for MapFormat

lbownik:
improved tests for MapFormat
refactored MapFormat.format, works 2x faster and generated about 50x less garbage
removed public method MapFormat.processPattern - it was not used and considering it's functionality must have been made public by mistake

### DIFF
--- a/platform/openide.util/src/org/openide/util/MapFormat.java
+++ b/platform/openide.util/src/org/openide/util/MapFormat.java
@@ -26,16 +26,15 @@ import java.text.NumberFormat;
 import java.text.ParsePosition;
 
 import java.util.Date;
-import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
+import static org.openide.util.NbBundle.getBundle;
 
-
-/** A text format similar to <code>MessageFormat</code>
- * but using string rather than numeric keys.
- * You might use use this formatter like this:
- * <pre>MapFormat.format("Hello {name}", map);</pre>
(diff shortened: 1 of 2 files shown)


vieiro (changes requested):
We have already explained in the mailing lists that you cannot change a public API of a module.

lbownik:
> We have already explained in the mailing lists that you cannot change a public API of a module.

my reasoning is the following (based on original implementation)
- "procesPattern" is never called by any client within code base;
- calling "processPattern" by clients is useless because:
      - it mainly prepares internal state for execution of "format", which then gets overwritten anyway, since "format" calls "preparePattern" enyway overwriting effects of previous execution, which makes calling "processPattern" by client a no-op;
      - "processPattern" returns the original pattern string with steering sequences stripped,  so giving it "Hello {name}!" will return "Hello !" - which seems useless for clients;
      
So I suspect that the propability of any client calling this method is 0. I suspect that this pethod intention was to be private but it was made public by accident (it was probably made public temporarily for testing purposes during developement and then stayed like this).

But.. If we want to keep this API structure unchanged then I could provide a stubbed implementation like

@Deprecated 
public String processPattern(String newPattern) throws IllegalArgumentException {
    return newPattern;
}

would it be acceptable?

vieiro:
Discussions about API changes are to be carried out in the mailing lists, with a broader audience, not in pull requests. Pull requests are expected to contain fixes and enhancements, not discussions.

PR #4091, opened 2022-05-09
Implementation of GoTo Symbol for Groovy

ppisl:
This PR adds functionality of GoTo Symbol for Groovy. Also there are fixes of Go to Type, now it's able to navigate to the type declaration. So far only file with the type declaration was opened and the position was set up to 0. 

The location is carrently stored in index.


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -252,10 +253,18 @@ public Set<IndexedMethod> getConstructors(final String className) {
                     }
                 }
                 int flags = 0;
-                if (parts.length > 2) {
+                if (parts.length > 3) {
```
sdedic, line 256: Please document what the `signature` syntax is; from the code it seems that e.g. `type` and `attribute` part may be optional (there are checks for that, but still better to have it documented), the `range` part is (?) mandatory.

Depending on that various checks may be insufficient; e.g. there are (?) 2 optional parts (type, attributes) separted by the same separator. So from notation `name;a` it's not obvious what 'a' is: a type or an attribute ?

But from the Indexer code, it seems that `type` is always present, so maybe the checks in index reading code can be simplified ?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndexer.java
```diff
@@ -321,6 +323,9 @@ private void indexClass(ASTClass element, IndexDocument document) {
             document.addPair(FQN_NAME, element.getFqn(), true, true);
             document.addPair(CLASS_NAME, name, true, true);
             document.addPair(CASE_INSENSITIVE_CLASS_NAME, name.toLowerCase(), true, true);
+            StringBuilder sb = new StringBuilder();
```
sdedic, line 326: Q: `addPair(CLASS_NAME,...)` is called multiple times (toplevel classes, inner classes), but [it is read as a single value here](https://github.com/apache/netbeans/pull/4091/files#diff-d59a38cf1daed90d439de7eb0be4bcea950ef015a4f2321d2cfdd9a50b90a104R607) -- will this work OK with fields of inner classes ? If not I'd suggest to file a bug for Groovy & fix it in a later PR.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/language/GroovyTypeSearcher.java
```diff
@@ -59,10 +62,38 @@ public class GroovyTypeSearcher implements IndexSearcher {
...
+            }
+            
+            Set<IndexedField> fields = index.getFields(textForQuery, null, kind);
+            for (IndexedField field : fields) {
+                result.add(new GroovyTypeDescriptor(field, helper));
+            }
+            
+            Set<IndexedMethod> methods = index.getMethods(textForQuery, null, kind);
+            for (IndexedMethod method : methods) {
+                if (method.getOffsetRange(null)  != OffsetRange.NONE) {
```
sdedic, line 90: Why is this check done for methods and not for fields + classes ? Synthetic methods ? Shouldn't we rather flag them with an attribute ?

ppisl: I have to admit that can look strange. The condition is here for my testing purposes and can be deleted. Unfortunately currently we don't store into the index whether the element is synthetic. Synthetic elements have offset [0,0] so it can be a sign how to recognize them. The question is, should we offer them in the Goto Symbol?

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndexer.java
ppisl: I'm not sure what you mean, the link shows me the whole diff.  But inner classes, even with the same name as outer class should work and work.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
ppisl: Type and range are mondatory. The attribute part is optional. I tried to keep same order as the version before.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -593,18 +630,28 @@ private IndexedMethod createMethod(String signature, IndexResult map) {
 
         // Extract attributes
         int attributeIndex = signature.indexOf(';', typeIndex + 1);
+        int offsetRangeIndex = signature.indexOf(";[", attributeIndex == -1 ? typeIndex + 1 : attributeIndex + 1);
```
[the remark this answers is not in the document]
ppisl: You are right, I'm going to change the order.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -348,6 +359,25 @@ public Set<IndexedMethod> getMethods(final String name, final String clz, QueryS
...
+        }
+        String textP = text.toLowerCase();
+        String whereP = where.toLowerCase();
+        int index = -1;
+        for (int i = 0; i < textP.length(); i++) {
+            index = whereP.indexOf(textP.charAt(i), index + 1);
+            if (index == -1) {
+                return false;
+            }
+            if (whereP.length() <= (index + 1)) {
```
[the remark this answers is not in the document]
ppisl: The condition is ok. I check whether is not whereP shorter then the searched text.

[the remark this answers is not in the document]
ppisl: The method `matchInsensitiveCamelCase` solve the case when search kind is CASE_INSENSITIVE_CAMEL_CASE.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/language/GroovyTypeSearcher.java
sdedic: I don't think we can goto to anything synthetic ;)

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -348,6 +358,25 @@ public Set<IndexedMethod> getMethods(final String name, final String clz, QueryS
         return methods;
     }
 
+    private boolean matchInsensitiveCamelCase(String text, String where) {
+        if (text.length() > where.length()) {
+            return false;
+        }
+        String textP = text.toLowerCase();
```
sdedic, line 365: Sorry :) I still believe this function is not a proper impl. At least two bugs here:
`matchInsensitiveCamelCase("RgExPars", "RegExParser")` produces `true`, but `RgEx` should not match `R`e`gEx`. The culprit is IMHO this lowercase, as we loose information on the Capital-and-lowercase exact substrings to be searched.

Try in IDE's Goto Type (type `RgExPars`)

The 2nd bug is 
```
if (whereP.length() <= (index + 1)) 
```
That causes same string match (`matchInsensitiveCamelCase("RegExParser", "RegExParser")`)  to return `false` -- the last matching character satisfies the condition.

sdedic (changes requested):

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/language/GroovyTypeSearcher.java
```diff
@@ -59,10 +62,38 @@ public class GroovyTypeSearcher implements IndexSearcher {
...
+        kind = adjustKind(kind, textForQuery);
+        
+        if (kind == QuerySupport.Kind.CASE_INSENSITIVE_PREFIX /*|| kind == QuerySupport.Kind.CASE_INSENSITIVE_REGEXP*/) {
+            textForQuery = textForQuery.toLowerCase(Locale.ENGLISH);
+        }
+        
+        Set<GroovyTypeDescriptor> result = new HashSet<GroovyTypeDescriptor>();
+        
+        
+        if (textForQuery.length() > 0) {
```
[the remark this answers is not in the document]
ppisl: In reality `textForQuery` has always a length. Method `getSymbols` is called only with some `textForQuery`. So the condition is unnecessary, but other index searchers use it as well. So in the end it's consistent with Java.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
ppisl: I have rewritten the algorithm. Added tests, so it should be ok now.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -242,7 +243,7 @@ public Set<IndexedMethod> getConstructors(final String className) {
 
             for (String constructor : constructors) {
                 String[] parts = constructor.split(";");
-                String paramList = parts.length > 1 ? parts[1] : ""; // NOI18N
+                String paramList = parts[1];
```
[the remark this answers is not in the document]
ppisl: Sanity checks added. Not only on this place, but also for fields and methods.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -649,38 +673,35 @@ private IndexedField createField(String signature, IndexResult map, boolean inhe
...
+            m.setOffsetRange(range);
+        }   
         return m;
     }
 
+    private static OffsetRange createOffsetRange(String text) {
+        OffsetRange result = null;
+        int offsetStartIndex = text.indexOf('[');
+        int commaIndex = text.indexOf(',', offsetStartIndex + 1);
+        int offsetLastIndex = text.indexOf(']', commaIndex != -1 ? commaIndex : offsetStartIndex);
```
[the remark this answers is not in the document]
ppisl: Changed :)

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/GroovyIndex.java
```diff
@@ -340,14 +332,56 @@ public Set<IndexedMethod> getMethods(final String name, final String clz, QueryS
...
         return methods;
     }
-
+    
+    // protected for test reasons.
+    protected static boolean matchCamelCase(String prefix, String where, boolean insensitive) {
+        if (where == null || where.length() == 0 || prefix == null || prefix.length() == 0) {
+            return false;
+        }
+        if (!prefix.equals(cachedPrefix) || cachedInsensitive != insensitive) {
```
sdedic, line 351: Note the `cachePrefix`, `cachedInsensitive` and `cachedCamelCasePattern` are shared between all instances of `GroovyIndex` across threads (2 threads can call `GroovyIndex.getFields()` simultaneously). Without synchronizing, Java may flush just part of the triplet on write making it temporarily inconsistent for other concurrent accesses.

sdedic (approved):

PR #4093, opened 2022-05-09
Better YAML editing with auto closing quotes and mustache

lkishalmi:
https://user-images.githubusercontent.com/1381701/167496466-a89e6eba-54e7-4edf-a686-a5eac4313376.mp4

These are just some minor keyboard tweaks what makes YAML editing more convenient.


lkishalmi:
Added some test cases. I had to fix the CslTestBase as insertChar was not working with selection after caret, and fixed a bunch of asserts there as well.

mbien (approved):
mustache - thats funny :)

lets wait with integration since we are having travis issues again, it wouldn't build anyway post integration. see INFRA-23257

PR #4097, opened 2022-05-10
Fix 'node id XXX already exists' error in vscode.

sdedic:
The bug manifests in several places. Most notably in vscode DB Explorer where a `Refresh` action causes vscode-nbls communication to fail with `Node xxx is already present`. This is caused by creation of a new node - a new identity, which is equal() to the previously created node, which is still in the map (since the client may still ask for its details).

And that node is created (since keys - underlying DB data) did not actually even change is done because of some of DB children use `ChildFactory` SPI from Nodes - this is the other fix. It turned out that `add` method of the `j.u.List` passed to `ChildFactory.createKeys()` is tweaked so it detects actual addition beyond the original content - and causes interim keys refresh, which, in turn, causes Node removal for whose keys are not (yet) present, since the `ChildFactory` did not add them.
Note that this fix is not a complete fix - if a (really) new node is inserted among existing, the nodes will be destroyed + recreated. I think it is probably impossible to fix without breaking Children and its event semantics. But the fix corrects situation when the content actually does not change or is only added to.

This second fix also fixes an issue in Gradle project view in NetBeans IDE; appears on root projects only.
- open a Gradle project, expand its `Build scrips` node. Suppose there are global `gradle.properties`, project's `gradle.properties`, `build.gradle` and `settings.gradle` (or make it so)
- open `build.gradle` in the editor
- modify the buildscript, save
- :open_mouth: the `build.gradle` node disappears !

@jhorvath please decide whether the fix should go to `master` or it is critical for DB explorer (where it mainly appears, but I've encountered it elsewhere as well) so `vscode` extension could benefit from inclusion in NB14.

adding @lkishalmi since it affects Gradle projects.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/LspTreeViewServiceImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/explorer/LspTreeViewServiceImpl.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.eclipse.lsp4j.jsonrpc.services.JsonSegment;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.LanguageClientAware;
@@ -48,6 +50,8 @@ import org.openide.util.Lookup;
  */
 @JsonSegment("nodes")
 public class LspTreeViewServiceImpl implements TreeViewService, LanguageClientAware {
+    private static final Logger LOG = Logger.getLogger(LspTreeViewServiceImpl.class.getName());
+    
     private final Lookup sessionLookup;
     /**
      * The delegate tree service.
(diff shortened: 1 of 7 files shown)


lkishalmi:
Well this one affects every project which using Async children. I do not see anything bad on the first glimpse, but I'm nowhere a node api expert. Asked @jtulach as well. The change is fine for me, though I do not feel myself qualified enough to approve it.

dbalek (approved):

jhorvath:
> 

@sdedic please include this fix also NB14 release.

jhorvath (approved):

sdedic:
Let's include RMs @neilcsmith-net  and @ebarboni to decide.

ebarboni:
hi, we have a regression with a fix under review so rc4 is expected. If you rebase on delivery it can be merged. I will also merge nb-javac 18 pr but tomorrow :D.

ebarboni:
@sdedic for ping :D

sdedic:
> @sdedic for ping :D

Sorry, had offline discussion about impact / importance / stability of this. Will rebase in ~1 hour.

ebarboni:
@sdedic a test isssue occurs on platform/openide.nodes in travis.

sdedic:
Had to fix the code, wrong analysis done w.r.t. handling of the Wait node. Now the code is simpler -- there's one extra entry, which is subtracted -- Math.max(0, x) is there just in case.

PR #4105, opened 2022-05-12
migrate groovy ci matrix from travis to gh actions.

mbien:
lets start with the great migration.

picked groovy as first target since it is already a nice matrix.

can we drop JDK 8? should we add 18? Lets see if it builds.


mbien:
```
2022-05-12T01:45:02.3774828Z     [junit] Caused by: java.awt.HeadlessException: 
2022-05-12T01:45:02.3775063Z     [junit] No X11 DISPLAY variable was set, but this program performed an operation which requires it.
```
looks like groovy will need a X11 setup. Will add that.

.travis.yml
```diff
@@ -593,60 +593,6 @@ matrix:
...
-          before_script:
-            - nbbuild/travis/ant.sh $BUILD_OPTS clean
-            - nbbuild/travis/ant.sh $BUILD_OPTS build
-            - wget https://raw.githubusercontent.com/sormuras/bach/master/install-jdk.sh
-            - export TEST_JDK=`bash install-jdk.sh --feature 17 --license GPL --emit-java-home --silent | tail -1`
-            - export OPTS="-Dmetabuild.jsonurl=https://raw.githubusercontent.com/apache/netbeans-jenkins-lib/master/meta/netbeansrelease.json $OPTS  -Dtest.nbjdk.home=$TEST_JDK -Dtest.run.args=--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming -Dtest.bootclasspath.prepend.args=-Dno.netbeans.bootclasspath.prepend.needed=true"            
-          script:
-            #- ant $TEST_OPTS -f groovy/groovy test
-            #- ant $TEST_OPTS -f groovy/groovy.java test
-            - ant $TEST_OPTS -f groovy/groovy.editor test
```
mbien, line 646: can someone explain how the groovy tests actually work?

 - what does `OPTS` do? I don't think anything is picking this env var up
 - no JMS flags are set, which means this actually can't run on JDK 17 since it would fail with encapsulation exceptions
 - JAVA_HOME is in fact correctly set but i can see in the travis log lines like `[nb-javac] Support for javac --release has been added in Java9 ignoring it which further indicates that this is running on JDK 8

svenreimers: Having been involved in creating the groovy matrix I will try to find some time to dig into this... stay tuned..

mbien: @svenreimers awesome! If its of any help, thats my rough local test setup for experiments (this doesn't actually work due to some quotation issue, but you get the idea :)):

```bash
#!/bin/bash
set -e

export JAVA_HOME=/home/mbien/dev/java/jdk-11.0.15.9.1-corretto
export PATH=$JAVA_HOME/bin:$PATH

JMS_FLAGS=$(cat nbbuild/jms-config/base.flags nbbuild/jms-config/desktop.flags | xargs)

TEST_RUN_ARGS="-Dtest.run.args=\"--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming $JMS_FLAGS\""

OPTS="-Dtest-unit-sys-prop.ignore.random.failures=true $TEST_RUN_ARGS"

echo $OPTS

ant $OPTS -f groovy/groovy.editor test
```

mbien:
closed in favor of #4431

PR #4120, opened 2022-05-15
Set --enable-preview for jtreg tests.

jlahoda:
I would like to add `--enable-preview` to the javac parameters used internally by the IDE to parse jtreg (OpenJDK) tests.


mbien (approved):
restarted some tests and everything is passing now.

I like that you extracted the while loop. Makes the method less complex.

Issue #4113, opened 2022-05-13
VSNetBeans: Latest VSCode 1.67 breaks icons in projects views 


MartinBalin:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

Installed VSNetBeans 13.0.1073 build from master and icons are broken on latest VSCode 1.67, works with 1.63.1
![Screenshot 2022-05-13 at 11 50 17](https://user-images.githubusercontent.com/26064232/168258800-8bec82dc-55e8-492f-8dc0-2e5b1e990d31.png)

### How to reproduce

Install VSIX and open any gradle or maven project.

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

macOS

### JDK

graalVM 22.1 EE JDK11

### Apache NetBeans packaging

Apache VSNetBeans for VSCode

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

sdedic:
Actually it seems that this is a bug in vscode - I traced that down to commits [vscode/#d79928](https://github.com/microsoft/vscode/commit/d79928be53a807a9bff45814b6dab65bae685879) and [vscode/#a9c3c0](https://github.com/microsoft/vscode/commit/a9c3c0e4b0384fe2b2c907b94a99fda18d3f34a6) -- `ThemeIcon` specification is completely ignored in vscode 1.67, if the `TreeItem` has a `resourceUri` defined. I am going to file a bug with an example code for vscode.

ebarboni:
is this ok now ?

sdedic:
OK from the NBLS user's point of view, I still need to file a bug for vscode to make a real fix, will reference it here

MartinBalin:
Yes, please file it with your sample VSIX.

sdedic:
@MartinBalin seems I've lost the race -- there are already at least two issues filed on this: microsoft/vscode#151087 and microsoft/vscode#150610

MartinBalin:
And it also seems fixed by https://github.com/microsoft/vscode/pull/151090 so some next iteration of vscode will have it fixed. Hopefully correctly now.

PR #4121, opened 2022-05-16
#4113: pretend that folders have no URI, so their icon can be customi…

sdedic:
vscode version 1.67 released in April breaks icon display for `TreeItems`: if an item reports `resourceUri` that corresponds to a folder, system-default folder icon (by default: none) is displayed, regardless of the `TreeItem`'s configured `iconPath`. 

This was discovered by Friday's team testing. The fix affects just **Apache NetBeans Language Server** vscode extension release, not the NetBeans release. If included in NB14, this patch should have no effect on the main release (the code is not distributed with NetBeans IDE).

### DIFF
--- a/java/java.lsp.server/vscode/CHANGELOG.md
+++ b/java/java.lsp.server/vscode/CHANGELOG.md
@@ -20,6 +20,11 @@
     under the License.
 
 -->
+## Version 14.0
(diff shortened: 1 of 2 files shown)


dbalek (approved):

MartinBalin (approved):

ebarboni:
nothing just it wasn't green

PR #4128, opened 2022-05-18
LSP: Format Document and Format Selection actions implemented.

dbalek:
LSP: `Format Document` and `Format Selection` actions implemented.
The implementation can read formatting settings exported form NetBeans.

### DIFF
--- a/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/lexer/LexUtilities.java
+++ b/groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/lexer/LexUtilities.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import javax.swing.text.BadLocationException;
 import javax.swing.text.Document;
 import org.netbeans.api.annotations.common.CheckForNull;
+import org.netbeans.api.editor.document.LineDocument;
 import org.netbeans.api.lexer.Token;
 import org.netbeans.api.lexer.TokenHierarchy;
 import org.netbeans.api.lexer.TokenId;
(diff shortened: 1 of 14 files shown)


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/lexer/LexUtilities.java
```diff
@@ -446,6 +511,15 @@ public static boolean isBeginToken(TokenId id, BaseDocument doc, int offset) {
         return END_PAIRS.contains(id);
     }
 
+    /**
+     * Return true iff the given token is a token that should be matched
```
ppisl, line 515: Possible typo here 'iff' . If it's really typo, it appears in more places.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
```diff
@@ -2243,4 +2285,220 @@ protected CallHierarchyOutgoingCall createResultItem(CallHierarchyItem item, Lis
         };
         return t.processRequest();
     }
+
```
ppisl, line 2288: This is not against this PR, but this file is really long :(

ppisl (approved):
From my point of view, OK.

groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/api/lexer/LexUtilities.java
dbalek: 'iff' usually means 'if and only if'.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/TextDocumentServiceImpl.java
dbalek: Agree :-(

PR #4116, opened 2022-05-14
JUnit Report for github CI

mbien:
attempt to add the [junit-report](https://github.com/mikepenz/action-junit-report) action to display junit results on failure.

lets see if this works

PR #4129, opened 2022-05-19
test release-drafter

ebarboni:
Hi, would like to see if it can help structuring release note.
I'm unaware if I can try it on my forks as labels belongs to apache/netbeans.
So this would be a first step

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,21 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+template: |
+  ## What’s Changed
+
+  $CHANGES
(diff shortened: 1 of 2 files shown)


mbien:
i don't know what i should review, I never used this :)

Is this a github bot which will create pull request for changelogs?

ebarboni:
Well it was more for technical I'm not sure copying the file is enough. It may help with a bit of configuration to get some info like
https://github.com/apache/archiva/releases

mbien:
ah [release-drafter](https://github.com/release-drafter/release-drafter) its a github action. But it is also available as github "app" https://github.com/apps/release-drafter, however the app is [deprecated](https://github.com/release-drafter/release-drafter/issues/335) and EOL ("The plan is to keep the GitHub App running for at least another year").

-> lets not use the app

Next we would have to figure out is if the action is part of the official github actions, if not, it will require extra steps. (git sub module deployment, see #4116 as example)

I would be also careful with github action access rights, I actually don't know how to configure this since apache projects have the rights set to minimum, see https://github.com/apache/netbeans/pull/4116#issuecomment-1126684953 (I am stuck there)

In theory, this could be done via a github token, but I don't know how this works for apache projects, I don't have access to the project settings maybe someone else has or there is a different way how to solve this.

ebarboni:
maybe I should have noticed this additionnal file 

https://github.com/apache/archiva/blob/master/.github/workflows/release-drafter.yml

mbien:
> maybe I should have noticed this additionnal file
> 
> https://github.com/apache/archiva/blob/master/.github/workflows/release-drafter.yml

this is good info, because archiva is an apache project and it uses the action directly without any tricks. This means the action must be on the white list.

ebarboni:
updated with all files. this is not testable on pr I guess

mbien:
if you push it to the master branch of your fork, you might be able to run it there, it should hopefully appear as workflow in your fork.

mbien:
i don't know. but you should be able to run it and it should output something at least :) even without labels.

ebarboni:
tested a bit, it only generate the draft that you can edit to get the markdown and use at another place. 
Will need nicer label configuration than the default but may help to get back release note on confluence

ebarboni:
image

I have limited access to the API so I guess it's acceptable but ajustement will be needed to fit our label.
This create a draf text that can be taken and stored.

mbien (approved):
sounds good!

PR #4130, opened 2022-05-19
Avoid dependency on ReferenceQueue internals

anthonyvdotbe:
What happens is:
* `ActiveQueue.Impl::removeSuper` invokes `ReferenceQueue::remove` with argument 0
* `ReferenceQueue::remove` was recently changed: before `remove()` delegated to `remove(0)`, but now the dependency is inverted and it's `remove(0)` that delegates to `remove()` instead
* `ActiveQueue.Impl` overrides `remove` and unconditionally throws `InterruptedException`
* `ActiveQueue.Daemon` [invokes](https://github.com/apache/netbeans/blame/master/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java#L115) `ActiveQueue.Impl::removeSuper` within an infinite loop, [catches](https://github.com/apache/netbeans/blame/master/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java#L134) the `InterruptedException` thrown by `ActiveQueue.Impl::remove`, and then continues the loop. This causes an infinite steam of `InterruptedExceptions` being thrown, which in turn results in constant CPU load

The fix is to invoke the correct overload of `ReferenceQueue::remove`.

Note: the PR as-is will throw an AssertionError in `isReferenceQueueV2` when running NetBeans on Java SE 9, since `Runtime.Version::feature` was only added in Java SE 10. But I assume Java SE 9 is unsupported and didn't want to complicate the code to support this edge case.

### DIFF
--- a/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java
+++ b/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java
@@ -57,17 +57,23 @@ public final class ActiveQueue {
         }
 
         @Override
-        public Reference<Object> remove(long timeout) throws IllegalArgumentException, InterruptedException {
-            throw new InterruptedException();
+        public Reference<? extends Object> remove(long timeout) throws IllegalArgumentException, InterruptedException {
+            if (timeout < 0) {
+                throw new IllegalArgumentException("Negative timeout value");
+            } else if (Thread.currentThread() != Daemon.running) {
+                throw new InterruptedException();
+            }
+
+            return super.remove(timeout);
         }
 
         @Override
-        public Reference<Object> remove() throws InterruptedException {
-            throw new InterruptedException();
-        }
-        
-        final Reference<? extends Object> removeSuper() throws InterruptedException {
-            return super.remove(0);
+        public Reference<? extends Object> remove() throws InterruptedException {
+            if (Thread.currentThread() != Daemon.running) {
+                throw new InterruptedException();
+            }
+
+            return super.remove();
         }
     }
 
@@ -112,7 +118,7 @@ public final class ActiveQueue {
                     if (impl == null) {
                         return;
                     }
-                    Reference<?> ref = impl.removeSuper();
+                    Reference<?> ref = impl.remove();
                     LOGGER.log(Level.FINE, "Got dequeued reference {0}", new Object[] { ref });
                     if (!(ref instanceof Runnable)) {
                         LOGGER.log(Level.WARNING, "A reference not implementing runnable has been added to the Utilities.activeReferenceQueue(): {0}", ref.getClass());
@@ -131,9 +137,8 @@ public final class ActiveQueue {
                         // to allow GC
                         ref = null;
                     }
-                } catch (InterruptedException ex) {
-                    // Can happen during VM shutdown, it seems. Ignore.
-                    continue;
+                } catch (InterruptedException ignored) {
+                    // Can happen during VM shutdown, it seems.
                 }
             }
         }


ebarboni:
I have low expertise on that ask for reviewer. 
Do you have an idea on the cpu load we save correcting this ?
I would add a comment in the code referencing this PR to avoid reverting this later :D

anthonyvdotbe:
> Do you have an idea on the cpu load we save correcting this ?

Basically this bug results in code as follows, so instead of having 1 logical core at 100%, it'll go back to 0%.

```java
while(true) {
    try {
        throw new InterruptedException();
    } catch (InterruptedException ignored) {}
}
```

mbien:
many tests fail on travis even after restart.

anthonyvdotbe:
@mbien AFAICS there's only [1 test](https://github.com/apache/netbeans/blob/master/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/csl/FoldingTest.java#L93) that fails (at line 3727 in the output). Where do you see the many others?
For the failed test: I don't see how the changes in this PR would affect PHP editor folding in any way, even more so as the same test passed in the Windows-based check. Would you mind to retry the checks one more time please?

mbien:
@anthonyvdotbe the PHP check is likely a false positive since it is known to fail, I am more worried about travis.

here you can see the list of jobs of this PR run, everything red has something failed. 
https://app.travis-ci.com/github/apache/netbeans/builds/250864589 

i actually just canceled the sigtests job since it ran for 1.4h which is highly unusual. its usually done in 40mins.

anthonyvdotbe:
Thanks for the guidance @mbien. The updated PR fixes the generic issue, but there's still a couple failing tests.

Is there a better way than what I've done, to check whether NetBeans is running on a JDK of at least version N? I suppose there's a utility method for this somewhere?

matthiasblaesing:
I think we should stay away from the implementation details of the JDK. It _always_ comes back to haunt us. If I read the implementation correctly, the idea is, that only the implementation inside the Lookup API should be able to `remove` from the `ReferenceQueue`. So why not change the guard method (untested!):

```diff
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java
+++ b/platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java
@@ -48,6 +48,8 @@
 
     private static final class Impl extends ReferenceQueue<Object> {
         
+        private ThreadLocal<Boolean> accessOk = new ThreadLocal<>();
+
         Impl() {
         }
 
@@ -58,18 +60,31 @@
 
         @Override
         public Reference<Object> remove(long timeout) throws IllegalArgumentException, InterruptedException {
+            if(accessOk.get() != null && accessOk.get()) {
+                return (Reference<Object>) super.remove(timeout);
+            } else {
             throw new InterruptedException();
         }
+        }
 
         @Override
         public Reference<Object> remove() throws InterruptedException {
+            if(accessOk.get() != null && accessOk.get()) {
+                return (Reference<Object>) super.remove();
+            } else {
             throw new InterruptedException();
         }
+        }
         
         final Reference<? extends Object> removeSuper() throws InterruptedException {
-            return super.remove(0);
+            accessOk.set(Boolean.TRUE);
+            try {
+                return super.remove();
+            } finally {
+                accessOk.set(Boolean.FALSE);
         }
     }
+    }
 
     private static final class Daemon extends Thread {
         private static boolean initialized;

```

The permitted callers can come though `removeSuper` and are cleared, while public callers will not get through the `accessOk`check. Another alternative would be to store the thread, that is working the reference queue and reject all calls not from that thread.

platform/openide.util.lookup/src/org/openide/util/lookup/implspi/ActiveQueue.java
```diff
@@ -67,7 +70,26 @@ public Reference<Object> remove() throws InterruptedException {
...
+                        Object version = Runtime.class.getDeclaredMethod("version").invoke(null);
+                        int feature = (int) member.getDeclaredMethod("feature").invoke(version);
+                        return feature >= 19;
+                    } catch(ReflectiveOperationException e) {
+                        throw new AssertionError(e);
+                    }
+                }
+            }
+            return false;
         }
```
mbien, line 93: can you put this into a static final field and add a short comment that the JDK impl changed in 19?

example:

```java
private static final boolean jdk19 = is19();

...
     if (jdk19) {
...
```

this would make it JDK 8 compatible:
```java
    private static boolean is19() {
        // TODO remove reflection post JDK 11 migration
        try {
            Object version = Runtime.class.getDeclaredMethod("version").invoke(null);
            int feature = (int) version.getClass().getDeclaredMethod("feature").invoke(version);
            return feature >= 19;
        } catch (NoSuchMethodException e) {
            return false; // JDK 8
        } catch (ReflectiveOperationException e) {
            throw new AssertionError(e);
        }
    }
```

regarding your question if there is an API for that, unfortunately not. Usually NB only cares about javac versions, which is easier to figure out via the `SourceVersion` class, but we can't do that here, because we patch javac to a recent version via nb-javac no matter what runtime JDK is in use, so it would always return true.

mbien (changes requested):
restarted some jobs and everything is green now, comments inline.

mbien:
> Another alternative would be to store the thread, that is working the reference queue and reject all calls not from that thread.

what would you say to `ReentrantLock`? It has a `isHeldByCurrentThread()` method.

matthiasblaesing:
I coded it with the `ReentrantLock` and looks nearly identically - I have no strong feelings either way as long as it stays an implementation detail and does not leak outside that class.

mbien:
> I coded it with the `ReentrantLock` and looks nearly identically - I have no strong feelings either way as long as it stays an implementation detail and does not leak outside that class.

@matthiasblaesing I would recommend using something which isn't TLS if possible, since dynamic TLS will more and more become an anti pattern in context of virtual threads. If the impl with `ReentrantLock` looks similar, i think we should pick that one - it might even communicate a bit better what is going on there for the reader since `isHeldByCurrentThread()` is fairly descriptive. (ironically the JDK 19 impl change of `ReferenceQueue` is also motivated by VTs)

anthonyvdotbe:
Thanks for the feedback. Since the thread is already stored in a field, I've adopted the following approach:

> Another alternative would be to store the thread, that is working the reference queue and reject all calls not from that thread.

mbien (approved):
good job

ebarboni:
Well nice finding. Just the release build is stuck, infra just started investigation. We can add this PR np for me. I will go from a RC before Voting candidate to be sure release build is ok again.

mbien:
i am going to rebase the commit on top of `delivery` and force push into this PR as discussed with @ebarboni  via mail. We can still decide to not merge but at least it would be prepared.

mbien:
ok i can't do that :( PR branch is set read only.

@anthonyvdotbe could you checkout a fresh `delivery` branch and move the commit to that branch, then force push into this PR? If you need more instructions let us know.

mbien:
@anthonyvdotbe thank you!

Issue #4102, opened 2022-05-11
groupId completion stops working


mbien:
### Apache NetBeans version

Apache NetBeans latest master

### What happened

i noticed a problem:
 - after the IDE (NB 15 dev build) is restarted `groupId` completion in maven poms doesn't give any results
 - completion in other tags, like `artifactId` or `version` still works fine

most likely caused by #4077 since reverting to the commit before that restores the functionality. Haven't investigated any further.

since this PR is fairly small it should be easy to revert in case no solution is found until NB15, this does not affect NB14. I am filing this mostly so that we don't forget.

### Did this work correctly in an earlier version?

Apache NetBeans 14

### Operating System

linux

### JDK

17.0.3

### Apache NetBeans packaging

Own source build

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
https://github.com/apache/maven-indexer/pull/201 reverting this fixes the issue. Should help to figure out what the cause is.

matthiasblaesing:
I think the issue was introduced by:

https://github.com/apache/maven-indexer/commit/c231db602b5cea9ed20eba617de5a5a846ef5faa

A costly workaround is:

```diff
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
+++ b/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
@@ -261,6 +261,7 @@
             File indexDirectory, String repositoryUrl, String indexUpdateUrl, List<? extends IndexCreator> indexers) throws IOException {
 
         IndexingContext context = indexer.createIndexingContext(id, repositoryId, repository, indexDirectory, repositoryUrl, indexUpdateUrl, true, true, indexers);
+        context.rebuildGroups();
         indexingContexts.put(context.getId(), context);
         return context;
     } 
``

matthiasblaesing:
I think we need to reintroduce the caching of the groups ourselfs. Building he indexes at runtime each time the index is opened will people even more mad than the repo scanning.

PR #4136, opened 2022-05-20
Fix maven groupId completion

matthiasblaesing:
MavenIndexer dropped the caching of the allGroups and rootGroups
properties in the index:

https://github.com/apache/maven-indexer/commit/c231db602b5cea9ed20eba617de5a5a846ef5faa

After the index is loaded there are two options:

a) rebuild the groups list from the index
b) cache the group list after indexing

NetBeans expects the index to be readily be available and thus option
a is not an option (rebuilding the group list runs over the full index
and is in the minute range). As there are only a few entrypoints that
modify the index caching the group list looks like a viable option.

Closes: #4102

### DIFF
--- a/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
+++ b/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
@@ -32,6 +32,7 @@ import java.net.URI;
 import java.nio.file.FileStore;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


mbien (approved):
thanks for fixing this, I wasn't sure at first about custom caching, I probably would have tried to add it into the index, but i believe this is the most straight forward way of doing this - I like it.

I added a commit which simplifies the code a little to use nio.Files.* utilities. Tested it and everything works.

Should we wait a bit with integration? Maybe maven-indexer releases a hotfix so that we can drop the "inversion" code.

matthiasblaesing:
I looked at your change and looks sane to me, adjusted a little and squashed. I intent to merge once this gets green. maven-indexer talks about going jdk 11+, this might or might not be a problem for NetBeans and I want to get back to being able to use master.

I was surprised by the push into this PR - I'd have expected a heads-up before (yes I'll try to remember to uncheck the "allow edits" checkbox).

mbien:
I mentioned it in the review comment, going to attach a diff next time.

PR #4142, opened 2022-05-23
ElementUtilities performance optimization.

### LINKED ISSUE #4140: java completion slow
### Apache NetBeans version

Apache NetBeans 13

### What happened

Java completion is very slow with large classes due to a poor choice of algorithm selected to filter out scope visibility.

### How to reproduce

Create a class with a moderately large number of static constants or functions (>1000), import it elsewhere using a wildcard.  Try to do any completion action that might match.  It typically takes hundreds of milliseconds to show the completion window and even longer to dismiss it when you choose something (gc?).

I hit it with a vulkan binding i'm working on which puts the 3000 odd constants in a single class.

No

### Operating System

multiple systems, slackware64-current, gentoo

### JDK

openjdk 18.0.1 from openjdk.net

### Apache NetBeans packaging

### Anything else

see: https://issues.apache.org/jira/browse/NETBEANS-6504

Sorry for now i'm not willing to fork and submit a pull request through this platform, but see the potential fixes I attached to jira if you're so inclined.

No

### Code of Conduct

Yes

mbien:
fixes #4140

`ElementUtilities.getMembers()`, `getLocalMembersAndVars()`, `getLocalVars()` and `getGlobalTypes()` performance optimization.
 - optimization was contributed by @notzed, see NETBEANS-6504 or #4140
 - only minor changes were made from the proposed patch

This significantly improves auto completion performance on types with thousands of fields. 
example: https://github.com/apache/netbeans/issues/4140#issuecomment-1134414691 completion takes several seconds to show up before patch, after patch it shows up instantly.

Since `ElementUtilites` has many usages across NetBeans, this could potentially improve performance in other places too.

First commit is the interesting one, other is cleanup. I added tests and made sure they were green before and after patch. Its already squashed in the first commit.

baseline before optimization:
completion_baseline

lets see if tests are green

https://user-images.githubusercontent.com/114367/176787320-6026f789-7721-4c6e-bb5a-1f18dc1ff9a8.mp4

### DIFF
--- a/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
+++ b/java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
@@ -3234,7 +3234,12 @@ public class ServerTest extends NbTestCase {
 
             @Override
             public CompletableFuture<List<QuickPickItem>> showQuickPick(ShowQuickPickParams params) {
-                return CompletableFuture.completedFuture(params.getItems().size() > 2 ? params.getItems().subList(0, 2) : params.getItems());
+                List<QuickPickItem> selection = params.getItems().stream()
+                        .filter((i) -> i.getLabel().equals("f1.chars(): IntStream") || i.getLabel().equals("f1.codePoints(): IntStream"))
+                        .sorted((i1, i2) -> i1.getLabel().compareTo(i2.getLabel()))
+                        .collect(Collectors.toList());
+                assertEquals(2, selection.size());
+                return CompletableFuture.completedFuture(selection);
             }
 
             @Override
--- a/java/java.source.base/src/org/netbeans/api/java/source/ElementUtilities.java
+++ b/java/java.source.base/src/org/netbeans/api/java/source/ElementUtilities.java
@@ -20,7 +20,6 @@ package org.netbeans.api.java.source;
 
 import com.sun.source.tree.CompilationUnitTree;
 import com.sun.source.tree.Scope;
-import com.sun.source.util.DocTrees;
 import com.sun.source.util.TreePath;
 import com.sun.source.util.Trees;
 import com.sun.tools.javac.api.JavacScope;
@@ -32,7 +31,6 @@ import com.sun.tools.javac.code.Source;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
-import com.sun.tools.javac.code.Symbol.PackageSymbol;
 import com.sun.tools.javac.code.Symbol.TypeSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
 import com.sun.tools.javac.code.Symtab;
@@ -217,7 +215,7 @@ public final class ElementUtilities {
      * @see Elements#getAllMembers
      */
     public Iterable<? extends Element> getMembers(TypeMirror type, ElementAcceptor acceptor) {
-        ArrayList<Element> members = new ArrayList<Element>();
+        Map<String, List<Element>> members = new HashMap<>();
         if (type != null) {
             Elements elements = JavacElements.instance(ctx);
             Types types = JavacTypes.instance(ctx);
@@ -229,18 +227,19 @@ public final class ElementUtilities {
                     if (te == null) break;
                     for (Element member : elements.getAllMembers(te)) {
                         if (acceptor == null || acceptor.accept(member, type)) {
-                            if (!isHidden(member, members, elements, types))
-                                members.add(member);
+                            addIfNotHidden(member, members, elements, types);
                         }
                     }
                     if (te.getKind().isClass() || te.getKind().isInterface() && SourceLevelUtils.allowDefaultMethods(Source.instance(ctx))) {
                         VarSymbol thisPseudoMember = new VarSymbol(Flags.FINAL | Flags.HASINIT, Names.instance(ctx)._this, (ClassType)te.asType(), (ClassSymbol)te);
-                        if (acceptor == null || acceptor.accept(thisPseudoMember, type))
-                            members.add(thisPseudoMember);
+                        if (acceptor == null || acceptor.accept(thisPseudoMember, type)) {
+                            addAlways(thisPseudoMember, members);
+                        }
                         if (te.getSuperclass().getKind() == TypeKind.DECLARED) {
                             VarSymbol superPseudoMember = new VarSymbol(Flags.FINAL | Flags.HASINIT, Names.instance(ctx)._super, (ClassType)te.getSuperclass(), (ClassSymbol)te);
-                            if (acceptor == null || acceptor.accept(superPseudoMember, type))
-                                members.add(superPseudoMember);
+                            if (acceptor == null || acceptor.accept(superPseudoMember, type)) {
+                                addAlways(superPseudoMember, members);
+                            }
                         }
                     }
                 case BOOLEAN:
@@ -256,24 +255,27 @@ public final class ElementUtilities {
                     com.sun.tools.javac.util.List<Type> typeargs = com.sun.tools.javac.util.List.of((Type)type);
                     t = new ClassType(t.getEnclosingType(), typeargs, t.tsym);
                     Element classPseudoMember = new VarSymbol(Flags.STATIC | Flags.PUBLIC | Flags.FINAL, Names.instance(ctx)._class, t, ((Type)type).tsym);
-                    if (acceptor == null || acceptor.accept(classPseudoMember, type))
-                        members.add(classPseudoMember);
+                    if (acceptor == null || acceptor.accept(classPseudoMember, type)) {
+                        addAlways(classPseudoMember, members);
+                    }
                     break;
                 case ARRAY:
                     for (Element member : elements.getAllMembers((TypeElement)((Type)type).tsym)) {
-                        if (acceptor == null || acceptor.accept(member, type))
-                            members.add(member);
+                        if (acceptor == null || acceptor.accept(member, type)) {
+                            addAlways(member, members);
+                        }
                     }
                     t = Symtab.instance(ctx).classType;
                     typeargs = com.sun.tools.javac.util.List.of((Type)type);
                     t = new ClassType(t.getEnclosingType(), typeargs, t.tsym);
                     classPseudoMember = new VarSymbol(Flags.STATIC | Flags.PUBLIC | Flags.FINAL, Names.instance(ctx)._class, t, ((Type)type).tsym);
-                    if (acceptor == null || acceptor.accept(classPseudoMember, type))
-                        members.add(classPseudoMember);
+                    if (acceptor == null || acceptor.accept(classPseudoMember, type)) {
+                        addAlways(classPseudoMember, members);
+                    }
                     break;
             }
         }
-        return members;
+        return flattenMembers(members);
     }
     
     /**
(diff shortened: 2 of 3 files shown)


mbien:
@notzed I would like to add you as author to the commit, which email address should I add?

mbien:
i can't stabilize travis, some debugging gonna be required. Unfortunately thats the CI job of unreliable tests, so we can't exclude that the performance improvement itself is causing this due to different timing.

```
-------------------
org.netbeans.modules.java.lsp.server.protocol.ServerTest
      failed: testSourceActionDelegateMethod
---
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest
      errored: testEmptyInterimPackageRemoved
org.netbeans.modules.java.lsp.server.protocol.ServerTest
      errored: testDidOpenPreservesLines
---
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest
      failed: testCreatedPeerFileAppears

org.netbeans.modules.java.lsp.server.protocol.ServerTest
      failed: testSourceActionDelegateMethod
---
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest
      errored: testEmptyInterimPackageRemoved
org.netbeans.modules.java.lsp.server.protocol.ServerTest
      failed: testSourceActionDelegateMethod
---
org.netbeans.modules.java.lsp.server.explorer.ProjectViewTest
      errored: testEmptyInterimPackageRemoved
org.netbeans.modules.java.lsp.server.protocol.ServerTest
      failed: testSourceActionDelegateMethod
---
org.netbeans.modules.java.lsp.server.protocol.ServerTest
      failed: testSourceActionDelegateMethod

Suite: org.netbeans.modules.java.lsp.server.protocol.ServerTest
      testSourceActionDelegateMethod FAILED :  expected:&lt;2&gt; but was:&lt;1&gt;
          junit.framework.AssertionFailedError: expected:&lt;2&gt; but was:&lt;1&gt;
          	at org.netbeans.modules.java.lsp.server.protocol.ServerTest.testSourceActionDelegateMethod(ServerTest.java:3257)
          	at java.lang.Thread.run(Thread.java:748)

```

notzed:
Michael Zucchi <notzed@gmail.com> would probably be best.  Sorry I haven't gotten back to you I was busy chasing up some other tech problems plus a lot of anxiety issues atm.

Thanks for getting this in, interested to see the post-patch flame graph.

java/java.source.base/test/unit/src/org/netbeans/api/java/source/ElementUtilitiesTest.java
```diff
@@ -530,4 +536,42 @@ private void handleDecl() {
...
+
+            List<String> good = Arrays.asList("getClass()",
+                    "hashCode()", "equals(java.lang.Object)", "clone()", "toString()", "notify()",
+                    "notifyAll()", "wait(long)", "wait(long,int)", "wait()", "finalize()", "A()",
+                    "FIELD1", "FIELD2", "run()", "this", "super", "class");
+            assertEquals(good.size(), list.size());
+            assertTrue(list.containsAll(good));
+
+        }, true);
+    }
```
mbien, line 574: hi @jlahoda I don't want to steal your time from doing record patterns :) But would this be a good template to test `ElementUtilities.getMembers()`? Compiler phase RESOLVED is the correct phase?

A LSP test fails after this optimization here is applied and those methods are not directly tested as far as I see. I want to reproduce it in a proper test instead of going through the high level `ServerTest.testSourceActionDelegateMethod` (i also have to rebuild everything to make the LSP code pick the changes up which is annoying/time consuming)

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
```diff
@@ -3228,7 +3228,12 @@ public CompletableFuture<ApplyWorkspaceEditResponse> applyEdit(ApplyWorkspaceEdi
 
             @Override
             public CompletableFuture<List<QuickPickItem>> showQuickPick(ShowQuickPickParams params) {
-                return CompletableFuture.completedFuture(params.getItems().size() > 2 ? params.getItems().subList(0, 2) : params.getItems());
+                List<QuickPickItem> selection = params.getItems().stream()
+                        .filter((i) -> i.getLabel().equals("f1.chars(): IntStream") || i.getLabel().equals("f1.codePoints(): IntStream"))
+                        .sorted((i1, i2) -> i1.getLabel().compareTo(i2.getLabel()))
+                        .collect(Collectors.toList());
+                assertEquals(2, selection.size());
+                return CompletableFuture.completedFuture(selection);
```
mbien, line 3236: Hi @sdedic, I have a question.
`ServerTest.testSourceActionDelegateMethod()` assumes that the items are served in a certain order. This PR here however adds an optimization which would return a random order due to the fact that `ElementUtilities.getMembers()` uses HashMaps internally.

The problem is that I don't know the requirements.
 - is it ok to change this test
 - or does getMembers() have to return a certain order
 - or does the LSP code have to sort the results somewhere, so the correct way would be to revert this test and fix it in LSP code

interestingly this was the only test which fails. NB itself does display the choices in the same sorted order as before too.

mbien:
> `Michael Zucchi <notzed@gmail.com>` would probably be best. Sorry I haven't gotten back to you I was busy chasing up some other tech problems plus a lot of anxiety issues atm.

@notzed added you as co-author. No worries there is no hurry, the next release is far away.

> Thanks for getting this in, interested to see the post-patch flame graph.

I actually tried to do this already. The problem is that your optimization is too fast :) I use async-profiler, which is a sampling profiler. To get good flame graphs I probably have to increase sample rate or trigger it via an agent instead of from CLI.

I believe this here is 99% done. Only have to clear up how/where to solve the sorting issue (if its an issue in the first place).

java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java
sdedic: @mbien AFAIK `ElementUtilities.getMembers` is typed as List (= ordered) and so it is javac's `Elements.getAllMembers` this one relies on. Javac internally uses something (ScopeImpl) that seems like LinkedHashMap so iteration happens in the insert order = lexical order.

notzed: insert order == lexical order ... doesn't make a lot of sense.  lexical order != scope order != insert order.  it's easy enough to change it to use a linked hashtable I suppose - it's what i was going to use until i saw the already existing insertifnotpresent function that used a plain hashtable. it always sorts alphabetically for presentation anyway, but the scope order for identical names has not changed.

sdedic: @notzed sorry, I don't understand - what `insertifnotpresent` ?

mbien: I don't think `List` strongly implies ordered (the non-javac `ElementUtilities` returns Iterable btw).

We have to figure out what order it is supposed to be and then write it down in the doc so that nobody has to guess again :)

Further if we decide that the order is important, we have to add a test for this. Its not good that one single high level test which is completely unrelated to the given utility methods fails, everything else is green - including a working IDE.

If we decide that the order is not important, either the test has to be fixed as proposed here, or the LSP code has to sort things somewhere to make the test work.

would be good to have input from @jlahoda for this but I completely understand if he can't since he is busy with language JEPs.

mbien: @sdedic @notzed I switched to LinkedHashMap since this might make it easier to integrate this without doing more research. Completion is still as fast as before far as I can tell.

@sdedic should we keep the change in `ServerTest` or should I revert it (I reverted it for now just to proof that everything is working like before). Assuming everything is green - would be great if you could review this PR, or maybe you know someone who would want to review this :)

would be good to merge this some time soon so that it gets some testing in before release if there are no concerns.

sdedic: Please keep the changes in the test. The importance of the order **is not mentioned** in the Javadoc anywhere and as you noted, E.U.  returns `Iterable` = no guarantees. It's good not to break the de-facto behaviour (`LinkedHashMap`), but the test should stick to the API.

mbien: @sdedic understood. Travis is green which indicates that the LinkedMaps work, going to remove the revert-commit again, so that the ServerTest changes are restored.

mbien:
> BTW the usage of `LinkedHashMap` does not entirely preserve the orignal order. Originally, elements that were not `isHidden` accumulated in a list in the order of element encounter. Now, **simpleName** order is preserved, but since Elements are assigned into individual named buckets, the relative order of (i.e.) 2nd or later references to names `foo` and `bar` may change their relative positions in the final result.

Yes, thats why I was reluctant to use a linked map right away. We would have to use two data structures, one List like in the original version, and the Map from the new version just for the visibility checks. Only the list is returned. I hope this would work without much overhead - but having redundant data structures is always ugly. I wished this wouldn't be needed and we could just return the list in any order :)

mbien:
@sdedic last commit is the dual collection proposal. The list is used to collect the results, maintaining the order as they occur, the map only for the algorithm. its a bit hacky though and someone reading the code will probably not know what is going on there.

Whether to keep the last two commits or not is a decision the compiler gurus have to make. (Performance wise its fine)

I could add comments etc to explain what is happening there for devs reading the code in future.

notzed:
Oh dear, this is disappointing if this is what ends up going in.

mbien:
> Oh dear, this is disappointing if this is what ends up going in.

@notzed tbh it turned out better than anticipated. See it this way: the original version flattened the map as last step. This version flattens the map element by element in every step before that and simply returns the list once done.

@sdedic @jlahoda would you be ok with this approach? Btw I added a clip which shows the performance difference.

mbien:
@sdedic @matthiasblaesing any reasons not to merge this? Feature freeze is in one week, we should not wait till the last minute.

java/java.source.base/src/org/netbeans/api/java/source/ElementUtilities.java
```diff
@@ -365,86 +370,97 @@ public Iterable<? extends Element> getMembers(TypeMirror type, ElementAcceptor a
...
+        List<Element> namedMembers = members.get(name);
+        if (namedMembers == null) {
+            namedMembers = new ArrayList<>();
+            members.put(name, namedMembers);
+        }
+        if (list != null) {
+            list.add(element);
         }
-    };
+        namedMembers.add(element);
```
matthiasblaesing, line 384: nitpick - lines 376-380+384 could be replaced with:

```java
members.computeIfAbsent(name, (e) -> new ArrayList<>()).add(element);
```

matthiasblaesing (approved):
I did not seen anything obviously wrong, I left an inline comment with a nitpick. I assume you tested it manually and it really improves performance, then lets get this in.

java/java.source.base/src/org/netbeans/api/java/source/ElementUtilities.java
mbien: @matthiasblaesing thanks for taking a look. It is quite difficult to get reviewers in summer :)

good idea. Will try to get this in. This would improve readability I agree.

mbien:
ready to merge (without further squashing) once everything is green again.

I squashed the LinkedHashMap commit and added some comments. If we ever decide that the return order does in fact not matter, we can simply revert dd2fd7dec9e1b924db73e145149c49b8f356ced6 since its standalone.

mbien:
all green. merging.

PR #4149, opened 2022-05-25
Initial implementation of Project Dependency API

sdedic:
We have API to get inter-**project** dependencies, but we do not have an API to get a view on dependencies incl. external resources. There's a possibility to get a ClassPath (for Java) enumerating all necessary resources as a flat list, but if one would like to get structure information ... there's probably no way.

This PR is an attempt to define some abstractions, which could eventually become part of `project.api`. They probably won't fit to good old ANT projects (but if someone finds a suitable adjustment that would make `ArtifactSpec` more Ant-like, please share your thoughts !!), but for example NetBeans modules (although ant-based) could make it.

The initial goal is to provide a structured view, a dependency tree, such as `maven dependency:tree` or `gradle dependencies`. The project may start to report its product (artifact, or artifacts ?) and later a project-type-neutral way can be offered to map a resource (e.g. jar) to an artifact known to a project. All this is TBD.

The API is initially implemented as private; `implementation dependency` is reuquired to use/implement it. This will change after the constructs are validated on more project types.


ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
```diff
@@ -0,0 +1,193 @@
...
+        return new ArtifactSpec<V>(VersionKind.REGULAR, groupId, artifactId, versionSpec, type, classifier, optional, data);
+    }
+
+    public static <V> ArtifactSpec<V> createSnapshotSpec(
+            @NonNull String groupId, @NonNull String artifactId, 
+            @NullAllowed String type, @NullAllowed String classifier, 
+            @NonNull String versionSpec, boolean optional, @NonNull V data) {
+        return new ArtifactSpec<V>(VersionKind.SNAPSHOT, groupId, artifactId, versionSpec, type, classifier, false, data);
+    }
+
```
ppisl, line 192: And VersionKind.Latest? We don't need it now?

java/java.lsp.server/vscode/package-lock.json
```diff
@@ -10,23 +10,23 @@
 			"license": "Apache 2.0",
 			"dependencies": {
 				"jsonc-parser": "3.0.0",
-				"vscode-debugadapter": "1.51.0",
-				"vscode-languageclient": "8.0.1"
+				"vscode-debugadapter": "1.42.1",
+				"vscode-languageclient": "7.0.0"
```
ppisl, line 14: Is it intended to downgrade the versions? There are downgraded versions for more modules.

ppisl (approved):
I went through this PR and it seems to me ok .

java/java.lsp.server/vscode/package-lock.json
sdedic: Unintentional - seems as leftovers from local experiments were committed, no change to vscode should be made here. Will clean up.

junichi11:
Maybe, Some files are missing the license header? (`.md`, `.xml`, `.properties`)

dbalek (approved):
The proposed API looks sane to me.

ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
sdedic: Removed in 0221749c9a

java/java.lsp.server/vscode/package-lock.json
sdedic: Changes removed (= same as `master`) in f3ee411faa

lkishalmi:
It's an interesting start. Too "mavenly" though at the moment. I'd let give it a go, though when applying to Gradle there would be some interesting questions how would could we handle includedBuild substitutes, and different outgoing variants. Also the question is there, should we handle these at all?

sdedic:
> It's an interesting start. Too "mavenly" though at the moment. I'd let give it a go, though when applying to Gradle there would be some interesting questions how would could we handle includedBuild substitutes, and different outgoing variants. Also the question is there, should we handle these at all?

I guess I may need your expertise in Gradle area here. Just OTOH (given the API), we might not even care, as the includedBuild (if I understood correctly) just replaces dependency with project dependency, but the project product artifact is still there. The dependency graph is also a snapshot of the final artifact resolution, so (IMHO) no change with included builds.

I thouhgt about adding extension services for the additional queries (e.g. one could determine build segments / included builds), which could be project-type-dependent or (if we find a suitable abstraction) project independent - that's why raw (unspecified) data is exposed from ArtifactSpec / Dependency. I thought  about adding a `Lookup.Provider` to the `DependencyResult` to offer snapshot-bound additional services; but I was not able to see the details at the time of design.

BTW - maybe the Maven Graph could be adapted to this API to see  graph dependency for Gradle too ;)

sdedic:
Just unreliable tests fail, merging.

PR #4147, opened 2022-05-24
JavaWithMaven / POM Project

vieiro:
hotfix for #3809 
- Creates a maven template with a `<packaing>pom</packaging>` pom.xml

PR #4150, opened 2022-05-25
Sync delivery to release140 for 14-rc6

ebarboni:
as rc5 was not building and as we need to wait for build system to work again


neilcsmith-net:
While we wait on infra be good to see if we can get #4147 checked and in before merging this.  I guess we could have still called it rc5, but safer this way.

neilcsmith-net (approved):

PR #4153, opened 2022-05-26
Groovy debugger in a separate module, depends on Groovy parser to retrieve names of Groovy classes.

entlicher:
Breakpoints in Groovy debugger do not currently work in inner classes. The name of the class breakpoint is submitted to is obtained from the name of the script. We need to retrieve class information from the Groovy parser.
This is why we extract `groovy.debug` from `groovy.support` to be able to depend on `groovy.editor` where the parser is.


ppisl (approved):

sdedic (approved):

neilcsmith-net:
Please take a look at #4415 and confirm whether or not that issue is caused by this change.  It's a potential blocker to NB15 release.  Thanks!

PR #4157, opened 2022-05-26
Added support for properties backed up by functional interfaces

lkishalmi:
I was working on Gradle configuration nodes to expose more information on them, while I bumped into PropertySupport and realized how old school that could be, especially if the represented value needs to be converted/formatted as well.

So I put some new age in the code. I hope you do not mind that I've removed some compile warnings alongside my additions.

Also, I'm open for feedback, method name changes etc.

### DIFF
--- a/platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
+++ b/platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
@@ -26,13 +26,17 @@ import java.beans.PropertyEditor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 import org.openide.util.Exceptions;
 import org.openide.util.NbBundle;
 
 /** Support class for <code>Node.Property</code>.
-*
-* @author Jan Jancura, Jaroslav Tulach, Ian Formanek
-*/
+ *
+ * @param <T> the type of the represented property.
+ *
+ * @author Jan Jancura, Jaroslav Tulach, Ian Formanek
+ */
 public abstract class PropertySupport<T> extends Node.Property<T> {
     /** flag whether the property is readable */
     private boolean canR;
@@ -62,6 +66,7 @@ public abstract class PropertySupport<T> extends Node.Property<T> {
     * Returns the value passed into constructor.
     * @return <CODE>true</CODE> if the read of the value is supported
     */
+    @Override
     public boolean canRead() {
         return canR;
     }
@@ -70,6 +75,7 @@ public abstract class PropertySupport<T> extends Node.Property<T> {
     * Returns the value passed into constructor.
     * @return <CODE>true</CODE> if the read of the value is supported
     */
+    @Override
     public boolean canWrite() {
         return canW;
     }
@@ -78,6 +84,7 @@ public abstract class PropertySupport<T> extends Node.Property<T> {
      * Like {@link Class#cast} but handles primitive types.
      * See JDK #6456930.
      */
+    @SuppressWarnings("unchecked")
     static <T> T cast(Class<T> c, Object o) {
         if (c.isPrimitive()) {
             // Could try to actually type-check it, but never mind.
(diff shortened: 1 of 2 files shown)


platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
```diff
@@ -87,6 +92,162 @@ static <T> T cast(Class<T> c, Object o) {
...
+     * @param valueType the type of the property
+     * @param displayName the display name of the property, can be {@code null}.
+     * @param supplier the getter functional interface, can be {@code null} for write-only properties.
+     * @param consumer the setter functional interface, can be {@code null} for read-only properties.
+     *
+     * @since 7.62
+     * @return a {@link PropertySupport} instance where getter and setter are
+     *         backed by the provided functional interfaces.
+     */
+    public static <T> PropertySupport<T> readWrite(String name, Class<T> valueType, String displayName, Supplier<T> supplier, Consumer<T> consumer) {
```
sdedic, line 128: Not a review objection, rather a thought to consider.

There are 3 overloads for each flavour (readOnly, readWrite) that serve rather as a convenience methods that do not require the caller to use `null` as argument (displayName, sdescription). Is that necessary ? I mean given that the result is `PropertySupport` = `Node.Property` = `FeatureDescriptor`, then anyone can call `setShortDescription` or `setDisplayName` afterwards (since it's in FeatureDescriptor). Or just pass `null` to the most-args factory method.
The `null`-passing technique is used to make Functional propsupport readonly or  writeonly anyway.

Another alternative, to make the construction fluent - and with an effect to other potential PropertySupport subclasses would be to add 
`public final PropertySupport withShortDescription(String s) { setShortDescription(s); return this;}` and the same for display name ... number of characters typed is +- same with the current state, but can be configured in one expression.

platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
```diff
@@ -322,14 +485,13 @@ public void setValue(T val)
...
         *    any editor.
         */
+        @Override
         public PropertyEditor getPropertyEditor() {
             if (propertyEditorClass != null) {
                 try {
-                    return propertyEditorClass.newInstance();
-                } catch (InstantiationException ex) {
+                    return propertyEditorClass.getDeclaredConstructor().newInstance();
+                } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | SecurityException | IllegalArgumentException | InvocationTargetException ex) {
```
sdedic, line 493: some of them can be reduced to a `ReflectiveOperationException` ?

platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
```diff
@@ -360,6 +522,35 @@ public ReadWrite(String name, Class<T> type, String displayName, String shortDes
...
+        private final Consumer<T> consumer;
+
+        public FunctionalProperty(String name, Class<T> type, String displayName, String shortDescription, Supplier<T> supplier, Consumer<T> consumer) {
+            super(name, type, displayName, shortDescription, supplier != null, consumer != null);
+            this.supplier = supplier;
+            this.consumer = consumer;
+        }
+
+        @Override
+        public T getValue() throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
```
sdedic, line 536: I guess there's no need to declare the exceptions, probably a leftover from a reflective op copypaste.

sdedic (commented):
+ bump version number in `manifest.mf`

platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
lkishalmi: Yes, copypaste indeed. Good catch!

platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
lkishalmi: Could be. Just accepted whatever the IDE offered to convert to Multicatch.

platform/openide.nodes/src/org/openide/nodes/PropertySupport.java
lkishalmi: Thanks for the feedback, I think I'd go with the fluent construction. That's a good idea! Thanks!

lkishalmi:
The ```manifest.mf```` was already auto bumped. I guess there would be no frequent big changes expected in this module for the NB15 cycle that would require an additoinal manual bump.

matthiasblaesing (commented):
Looks sane to me.

matthiasblaesing (approved):

PR #4167, opened 2022-05-27
Default NBLS userdir to workspace local directory. Development runs on global.

sdedic:
It seems that the user experience is better when NBLS runs for each workspace (or single opened folder) = each vscode window separately. It consumes more memory, but is less confusing to users, as e.g. projects do not mix between the windows etc.

So let's have NBLS userdir `local` by default.


MartinBalin (approved):

dbalek (approved):

sdedic:
Rebased on latest `master`

PR #4168, opened 2022-05-28
Additional methods into knownBLockingMethods to match classes of JDK-11

jtulach:
While investigating [Enso parser performance](https://www.pivotaltracker.com/n/projects/2539304/stories/182234656) I needed to filter out many idle threads. Rather than doing that manually I decided to update the `knownBLockingMethods` of NetBeans profiler.

Test with sample `.npss` file attached.


matthiasblaesing (approved):
Looks sane to me - in the test the InputStream in loadSnapshot is not closed correctly, but that should not be a problem for a test.

mbien:
even though it might not make a difference in practice, I believe it would be better to close that stream anyway, otherwise code scans will pick that up every time and it would add to the noise.

ebarboni:
Sorry being late on this. I've recently splitted the netbeans-license job into two jobs the 
netbeans-apisigcheck for sig test (that are always unstable)
netbeans-license to do basicaly the rat check  verify-libs-and-licenses. so netbeans-license job should be green if we are good.

the npss is not handled so far, we should handle that.

PR #4134, opened 2022-05-20
Removing isEventDispatchThread - it causes problems in headless env

JaroslavTulach:
A colleague of mine reported following error when using `Sampler`:
```
Caused by: java.awt.AWTError: Could not create Toolkit: sun.awt.X11.XToolkit
    at java.desktop/java.awt.Toolkit$2.run(Toolkit.java:607)
    at java.desktop/java.awt.Toolkit$2.run(Toolkit.java:583)
    at java.base/java.security.AccessController.doPrivileged(Native Method)
    at java.desktop/java.awt.Toolkit.getDefaultToolkit(Toolkit.java:582)
    at java.desktop/java.awt.Toolkit.getEventQueue(Toolkit.java:1494)
    at java.desktop/java.awt.EventQueue.isDispatchThread(EventQueue.java:1086)
    at java.desktop/javax.swing.SwingUtilities.isEventDispatchThread(SwingUtilities.java:1493)
    at org.netbeans.modules.sampler.Sampler.stopSampling(Sampler.java:232)
    at org.netbeans.modules.sampler.Sampler.stopAndWriteTo(Sampler.java:209)
    at org.netbeans.modules.sampler.Test.stopSampling(Test.java:58)
    at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    ... 1 more
```

We shouldn't be calling into Swing in a library that is supposed to be used in a headless environment. Removing the (almost useless) check.

PR #4173, opened 2022-05-30
Sync delivery to master after 14-rc6

ebarboni:


mbien (approved):
looks like everything made it this time :)

#4134 was in rc5 so it doesn't appear here in case someone is wondering

ebarboni:
merging to master as rc6 announced

Issue #4141, opened 2022-05-23
very very slow output window when running ant projects


notzed:
### Apache NetBeans version

Apache NetBeans 13

### What happened

The output window has just always been very slow in general but _fails hard_ with long lines.  It makes netbeans feel very clunky and heavy and i always blamed it on swing and 'well, java is just kinda slow' even though i know it isn't.

I traced it down to two trivial problems in `extide/o.apache.tools.ant.module`, file `org.apache.tools.ant.module.bridge.impl.ForkedJoinOverride.java`, method `Copier::run()`.

1. The `if (!STACK_TRACE.matcher(str).find())` call (ForkedJoinOverride.java:318) takes an extraordinary amount of time to execute on long lines.  This could be an openjdk issue.  Changing this to `if (!STACK_TRACE.matcher(str).matches())` speeds this up considerably but it's still terribly slow.  Note: `java.project.JavaAntLogger` uses .matches() and not .find() with the same regex.
2. Calling `in.read()` and `outputLine.write()` one byte at a time. Changing this to read into a buffer first and then searching the buffer for '\n' with some simple logic to do everything in batches and avoid the conditional substring() delivers a further greater-than 20x speed improvement.  Note that simply using a `BufferedInputStream` doesn't help much.

`LOGGER_MAX_LINE_LENGTH` default of 3000 also seems unreasonably long to avoid looking for stack traces.

### How to reproduce

run this in an ant project.
```java
    public static void main(String[] args) throws InterruptedException {
        for (int i = 0; i < 10000; i++) {
            for (int j = 0; j < 1000; j++) {
                System.out.print(j);
            }
            System.out.println();
        }
    }
```

As released netbeans will take in the order of 25 minutes(!!)  to run it to completion (ryzen 3900x, estimated, it was 16s for just 100 lines).  Just changing find() to matches() as suggested in 1. above it will complete in under 2 minutes.  Fixing the code to use a locally buffered read and iterating over the byte array it will take about 5 seconds.

### Did this work correctly in an earlier version?

No

### Operating System

slackware64-current, gentoo

### JDK

openjdk 18.0.1 from openjdk.net.  17.0.1 is no better.

### Apache NetBeans packaging

Own source build

### Anything else

I have prototype code but no suitable patch for this.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

notzed:
[...]

PR #4180, opened 2022-05-31
Improve output window performance for ant java tasks #4141.

notzed:
These two together improve a test case of 10000 lines of 2890
characters from taking an estimated 25 minutes to run to taking 5
seconds:
 Change to use matches() rather than find() for the stack checking
  regex, which runs very slowly on long lines.
 Use buffered reads rather than single byte reads.

Other changes:
 Use an ExecutorService to start and manage threads.
 Remove references to deprecated ThreadDeath exception.
 Move critical section to isolated functions.
 Split non-processing and processing stream handlers
  into separate classes.
 Fix some logic errors introduced from / and requiring ancient
  'hot fixes'.

---

### DIFF
--- a/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
+++ b/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
@@ -24,7 +24,11 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Vector;
+import java.io.UnsupportedEncodingException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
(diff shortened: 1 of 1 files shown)


ebarboni:
Hi, maybe you should check your name in your git sound like an avatar

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
```diff
@@ -162,228 +166,222 @@ public synchronized void setError(File[] error) {
...
+            }
+
+            public void start() throws IOException {
+                getProject().getBuildListeners().stream()
+                    .filter(o -> o instanceof NbBuildLogger)
+                    .map(o -> (NbBuildLogger)o)
+                    .findFirst().ifPresentOrElse(buildLogger -> {
+                        tasks.submit(new MarkupCopier(stdout, Project.MSG_INFO, false, outEncoding, buildLogger, foldingHelper));
+                        tasks.submit(new MarkupCopier(stderr, Project.MSG_WARN, true, outEncoding, buildLogger, foldingHelper));
+                    }, ()->{
```
mbien, line 201: `ifPresentOrElse` is Java 9. Some modules have to stay compatible with Java 8 still, unfortunately. (thats the reason why the build fails)

mbien:
> Hi, maybe you should check your name in your git sound like an avatar

```
git commit --amend --author="Author Name <email@address.com>" --no-edit
```
this would update the author of the last commit in the current branch. After that just force push to the remote branch.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
notzed: huh, i guess there's another netbeans bug then because the source level is set to 1.8 and it didn't pick this up.

mbien:
@notzed your github name is fine. I believe Eric meant the name on the commit which is "Not Zed". This has to be a real person I believe, with a correct mail address.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
```diff
@@ -162,219 +166,216 @@ public synchronized void setError(File[] error) {
...
                                     foldingHelper.checkFolds(str, err, session);
-                                    if (str.length() < LOGGER_MAX_LINE_LENGTH) { // not too long message, probably interesting
-                                        // skip stack traces (hyperlinks are created by JavaAntLogger), everything else write directly
-                                        if (!STACK_TRACE.matcher(str).find()) {
-                                            StandardLogger.findHyperlink(str, session, null).println(session, err);
-                                        }
-                                    } else {
-                                        // do not match long strings, directly create a trivial hyperlink
+                                    if (str.length() >= LOGGER_MAX_LINE_LENGTH || !STACK_TRACE.matcher(str).matches())
                                         StandardLogger.findHyperlink(str, session, null).println(session, err);
```
mbien, line 340: gonna have to check if matches() gives the same results as find() with the given regex.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
```diff
@@ -162,219 +166,216 @@ public synchronized void setError(File[] error) {
...
+                    .map(o -> (NbBuildLogger)o)
+                    .findFirst()
+                    .orElse(null);
+                if (buildLogger != null) {
+                        tasks.submit(new MarkupCopier(stdout, Project.MSG_INFO, false, outEncoding, buildLogger, foldingHelper));
+                        tasks.submit(new MarkupCopier(stderr, Project.MSG_WARN, true, outEncoding, buildLogger, foldingHelper));
+                } else {
+                    setCopier(stdout, getOutputStream(), delegateOutputStream, false);
+                    setCopier(stderr, getErrorStream(), delegateErrorStream, true);
+                }
```
mbien, line 206: small nitpick: indentation

mbien (commented):
all tests are green. Thats a good sign. Impl looks good at first glance, but I have to take a closer look later - great cleanup for sure.

I believe building NB via shell was known to be faster than via NB. This makes me curious how it performs post patch.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
notzed: I'm pretty certain it does because it contains .* on both ends so will always match the whole string anyway.  Also probably more to the point this is just to check if it should ignore the lines because JavaAntLogger will mark them up - and JavaAntLogger uses the same regex and uses matches().

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
mbien: you can actually configure it on a per-project basis which would override the global default. But NetBeans has ~840 sub projects so its probably not worth the trouble :)

I think the only change I made from the global default is to set tab size to 4 (I believe 8 is default). But this would only show in files which actually use tabs.

I rarely use auto formatting. If i do its usually only for few selected lines since it creates too much noise in the diff.

It was not super important anyway, I only mentioned it since I noticed it while looking through it.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
mbien: yes I think you are right. The regex already matches the full line, `find()` would be redundant and likely much slower.
I mostly tried to find an excuse to use the regex checker window #4200.

notzed: For the test i was using, find vs matches is roughly ~1500x slower and there are worse cases.  Due to the greedy operator it matches the whole string then walks all the way back to the beginning testing each position looking for the next substring match.  With matches() it then terminates but because it isn't anchored to the start of the string find() advances by one character and repeats - all the way to the other end - so it's particularly bad with long lines, and is the reason for the line length checking here and in javaantlogger. I filed a jdk performance issue against java.util.regex.Pattern but my initial assessment was wrong as to the cause, and it would be tricky to address due to the way the state is managed in the compiled expression.

perl doesn't even execute the regex because it does a pre-scan and determines the string can't possibly match due to the bits that aren't optional which are not present like '('.

Removing the .* at either end makes find() run about the same as matches().  With some simplification and no capturing groups it can run somewhat faster than calling matches() but it's no longer a significant part of the processing time and doesn't seem worth maintaining another equally unreadable expression.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
```diff
@@ -162,219 +166,216 @@ public synchronized void setError(File[] error) {
...
+
+            public void start() throws IOException {
+                NbBuildLogger buildLogger = getProject().getBuildListeners().stream()
+                    .filter(o -> o instanceof NbBuildLogger)
+                    .map(o -> (NbBuildLogger)o)
+                    .findFirst()
+                    .orElse(null);
+                if (buildLogger != null) {
+                    tasks.submit(new MarkupCopier(stdout, Project.MSG_INFO, false, outEncoding, buildLogger, foldingHelper));
+                    tasks.submit(new MarkupCopier(stderr, Project.MSG_WARN, true, outEncoding, buildLogger, foldingHelper));
```
mbien, line 202: second task should use `errEncoding`.

i think we could also store the encoding as Charset, this would remove lookups inside toString(... "encoding"). 

```diff
diff --git a/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java b/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
index b9fbad1..4f91cca 100644
--- a/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
+++ b/extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
@@ -25,6 +25,7 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -104,8 +105,8 @@
 
     private class NbRedirector extends Redirector {
 
-        private String outEncoding = System.getProperty("file.encoding"); // NOI18N
-        private String errEncoding = System.getProperty("file.encoding"); // NOI18N
+        private Charset outEncoding;
+        private Charset errEncoding;
 
         // #158492. In Ant 1.8.0 output redirection cannot be distinguished by OutputStream subclass type
         // (LogOutputStream vs OutputStreamFunneler$Funnel) as OutputStreamFunneler$Funnel is
@@ -115,6 +116,14 @@
 
         public NbRedirector(Task task) {
             super(task);
+            try {
+                Charset cs = Charset.forName(System.getProperty("file.encoding")); // NOI18N
+                outEncoding = cs;
+                errEncoding = cs;
+            } catch (IllegalArgumentException ex) {
+                outEncoding = Charset.defaultCharset();
+                errEncoding = Charset.defaultCharset();
+            }
         }
 
         public @Override ExecuteStreamHandler createHandler() throws BuildException {
@@ -123,12 +132,12 @@
         }
 
         public @Override synchronized void setOutputEncoding(String outputEncoding) {
-            outEncoding = outputEncoding;
+            outEncoding = Charset.forName(outputEncoding);
             super.setOutputEncoding(outputEncoding);
         }
 
         public @Override synchronized void setErrorEncoding(String errorEncoding) {
-            errEncoding = errorEncoding;
+            errEncoding = Charset.forName(errorEncoding);
             super.setErrorEncoding(errorEncoding);
         }
 
@@ -199,7 +208,7 @@
                     .orElse(null);
                 if (buildLogger != null) {
                     tasks.submit(new MarkupCopier(stdout, Project.MSG_INFO, false, outEncoding, buildLogger, foldingHelper));
-                    tasks.submit(new MarkupCopier(stderr, Project.MSG_WARN, true, outEncoding, buildLogger, foldingHelper));
+                    tasks.submit(new MarkupCopier(stderr, Project.MSG_WARN, true, errEncoding, buildLogger, foldingHelper));
                 } else {
                     setCopier(stdout, getOutputStream(), delegateOutputStream, false);
                     setCopier(stderr, getErrorStream(), delegateErrorStream, true);
@@ -274,7 +283,7 @@
 
         private final InputStream in;
         private final int logLevel;
-        private final String encoding;
+        private final Charset encoding;
         private final RequestProcessor.Task flusher;
         private final ByteArrayOutputStream currentLine;
         private final OutputWriter ow;
@@ -282,7 +291,7 @@
         private final AntSession session;
         private final FoldingHelper foldingHelper;
 
-        public MarkupCopier(InputStream in, int logLevel, boolean err, String encoding, NbBuildLogger buildLogger, FoldingHelper foldingHelper) {
+        public MarkupCopier(InputStream in, int logLevel, boolean err, Charset encoding, NbBuildLogger buildLogger, FoldingHelper foldingHelper) {
             this.in = in;
             this.logLevel = logLevel;
             this.err = err;

```

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
```diff
@@ -162,219 +166,216 @@ public synchronized void setError(File[] error) {
...
-                    log(str, logLevel);
+                String str = take();
+                synchronized (foldingHelper) {
+                    if (!str.isEmpty()) {
+                        ow.write(str);
+                        log(str, logLevel);
+                    }
+                    if (end && err)
+                        foldingHelper.clearHandle();
                 }
```
mbien, line 368: i like that you use `foldingHelper` as monitor instead of the task. This might fix a race condition, although interleaving `out` and `err` wouldn't result in proper folding anyway most of the time I suppose.

mbien (commented):
found a little issue, nothing major. Rest looks good to me.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
notzed: Thanks for spotting the paste-o, i agree on the charset change, i just left it as is to avoid unnecessary changes and since that form is widely used everywhere else in java code and avoids the need to handle the exception.

Do you want me to apply this diff and re-push the pull request?

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
notzed: It's a bit ugly but while i was cleaning up the final patch I noticed the comment on FoldingHelper stated it was supposed to be used as a lock for interleaving output, but then it wasn't actually used.  

With this changed it will ensure stderr/stdout has full line outputs before interleaving unless the timeout/flush happens in-between.  The output from the case below is improved although each line still interleaves due to the output rate.

    public class SlowOutput {
    static void deepStackTrace(int i) {
        if (i == 0) {
            try {
                System.out.println("/home/notzed/.emacs:5:4: yo ho");
                throw new Exception();
            } catch (Exception x) {
                x.printStackTrace(System.err);
            }
        } else {
            deepStackTrace(i - 1);
        }
    }
    public static void main(String[] args) throws InterruptedException {
        for (int i = 0; i < 10000; i++) {
            for (int j = 0; j < 1000; j++) {
                System.out.print(j);
            }
            System.out.println();
            if (i % 50 == 49)   deepStackTrace(5);
        }
    }
    }

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
mbien: feel free to apply the diff or implement it yourself, whatever you prefer. I simply refactored it to check if it would work and pasted my version.

You can add it to the commit and force push into the PR branch.

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
mbien: good stuff!

extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/ForkedJavaOverride.java
mbien: btw the way i see this: the exception would happen anyway, just later in the toString(... "encoding") method. So, converting it early to `Charset` should be fine without breaking contracts.

notzed: Well now this code has to make a decision on what to do - the only exception being handled (illegalargumentexception) happens only if the property is null, and that case would've previously thrown nullpointerexception in String:lookupCharset().  Either way having a fallback is potentially a behavioural change.

I looked at an alternative approach but then noticed ByteArrayOutputStream:toString(Charset) was introduced at Java 10, so it can't be used anyway can it?

For what it's worth I had looked into a different buffering mechanism previously that could just use String:new directly but it didn't seem worth the added code.

mbien: > ByteArrayOutputStream:toString(Charset) was introduced at Java 10, so it can't be used anyway can it?

oh wow you are right. I haven't checked for that. Lets forget about this then - my mistake. Lets only fix the `errEncoding` and we are good.

just as sidenote:
all other exceptions the method throws are subtypes of IllegalArgumentException, so it is basically catching everything, not only the null case (but making this a multi-catch would be clearer probably).

mbien (approved):

mbien:
planning to merge this soon unless someone speaks up @lkishalmi @matthiasblaesing

mbien:
oh no @notzed ;) don't merge into this branch. Now we have a 
`Author: notzed <51409789+notzed@users.noreply.github.com>` commit in here.

You think you can revert that merge and only keep the commit?
`git reset HEAD^`
should remove the last commit. And simply force push again.

In future if you want to refresh PRs you can do something like this:
`git pull --rebase --autostash origin master`
this is going to put the commit on top of latest master. (assuming origin points at the main repo, but in general its only needed to refresh PRs when they are a few months old or have conflicts)

notzed:
I must've read somewhere to keep the fork up to date so i've been doing that, i wondered why the last time did something funny - the git told me to run some commands so i did and that's what must've done it.

Err I think i fubard something else now.

mbien:
all good. now hands off :)

mbien:
@notzed congrats to your first contribution

btw you can link your commit author details to your account by adding the email address you used for the commit to your github account somewhere in the settings.

notzed:
Goodo, been a while since I contributed much to anything but i was very active in GNOME back in the day (2nd iteration of gnome-terminal, evolution).  Dunno if i'll find something else to poke around in netbeans but we'll see - i quit my job 18 months ago and can't be bothered getting another one yet so i've got a lot of time on my hands and it's getting a bit boring.

PR #4184, opened 2022-06-01
Maven / dependency query fixes.

sdedic:
During testing (and usage), it turned out that the newer `DependencyGraphBuilder` does not give out as much information as the older `DependencyTreeBuilder`. Namely duplicated artifacts are omitted from the dependency report and in case of a version clash, the clashing artifact is **removed at all** from the report.

So this PR mainly rewrites the maven implementation back to the older API so that more info can be pulled out from maven.

I've realized that the client may need to allow artifact fetching (online / offline mode) as well as order to flush caches before running the query. I changed the private API so that `DependencyQuery` object is now used as a parameter (+ added a `Builder` for it). Further extensions should be API-compatible.

Strangely enough, our infrastructure (?) sometimes fails to configure the session properly (?) or maybe Maven does not propagate the offline flag everywhere, so our `RepositoryConnector` pulls on an emergency brake and throws `AssertionError` to abort the unexpected and disallowed online operation. I've added a test method to see whether the throwable is an 'offline' exception.

I've extended `DependencyTreeFactory` API to allow multiple scopes to be specified; and also wraps maven internal errors into `MavenExecutionException` instead of just returning `null` (which can't be inspected for what happened). The older method variant remains compatible (catches, logs, returns `null`).

Note: I've **removed** a method from `DependencyTreeFactory` that has been accidentally added by previous commit (ea62da49). This version was not released yet.

### DIFF
--- a/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
+++ b/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
@@ -21,6 +21,7 @@ package org.netbeans.modules.project.dependency;
 import java.util.Objects;
 import org.netbeans.api.annotations.common.NonNull;
 import org.netbeans.api.annotations.common.NullAllowed;
+import org.openide.filesystems.FileObject;
 
 /**
  * Represents an artifact. Each artifact is identified by
@@ -62,9 +63,10 @@ public final class ArtifactSpec<T> {
     private final String versionSpec;
     private final String classifier;
     private final boolean optional;
+    private final FileObject localFile;
     final T data;
 
-    ArtifactSpec(VersionKind kind, String groupId, String artifactId, String versionSpec, String type, String classifier, boolean optional, T impl) {
+    ArtifactSpec(VersionKind kind, String groupId, String artifactId, String versionSpec, String type, String classifier, boolean optional, FileObject localFile, T impl) {
         this.kind = kind;
         this.groupId = groupId;
         this.artifactId = artifactId;
@@ -73,6 +75,15 @@ public final class ArtifactSpec<T> {
         this.optional = optional;
         this.data = impl;
         this.type = type;
+        this.localFile = localFile;
+    }
+
+    public T getData() {
+        return data;
+    }
+
+    public FileObject getLocalFile() {
+        return localFile;
     }
 
     public VersionKind getKind() {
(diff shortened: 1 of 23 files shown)


ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
```diff
@@ -62,9 +63,10 @@ public enum VersionKind {
     private final String versionSpec;
     private final String classifier;
     private final boolean optional;
+    private final FileObject localFile;
     final T data;
 
-    ArtifactSpec(VersionKind kind, String groupId, String artifactId, String versionSpec, String type, String classifier, boolean optional, T impl) {
+    ArtifactSpec(VersionKind kind, String groupId, String artifactId, String versionSpec, String type, String classifier, boolean optional, FileObject localFile, T impl) {
```
ppisl, line 69: Am I rihgt, when I expect that the localFile represnets the artefact file (jar for example)?

ide/project.dependency/src/org/netbeans/modules/project/dependency/DependencyResult.java
```diff
@@ -45,8 +46,18 @@ public interface DependencyResult extends Lookup.Provider {
...
+     * @return true, if the data is valid
+     */
     public boolean isValid();
     
+    /**
+     * Returns artifacts that may be unavailable or erroneous.
+     * @return problem artifacts
+     */
+    public Collection<ArtifactSpec> getProblemArtifacts();
+
```
ppisl, line 60: So this artefacts are not part of the dependency tree?

ppisl (approved):

ide/project.dependency/src/org/netbeans/modules/project/dependency/DependencyResult.java
sdedic: They can be ... but they are 'problematic' :) e.g. they do not have the `getFile()` valid / filled in. Or some other issue like version conflict. Maybe this could rather identify the `Dependency` structure -- which means an exact position in the tree  ?

ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
sdedic: Yes.

PR #4185, opened 2022-06-01
set release drafter workflow to manual activation.

mbien:
workflow trigger button should appear here once merged:
https://github.com/apache/netbeans/actions/workflows/release-drafter.yml


ebarboni:
thanks for looking into this. Maybe merge asap to avoid spaming drafter

mbien:
roger. merging.

PR #4187, opened 2022-06-02
LSP: Open type command added.

dbalek:
Open type command added.


graemerocher:
Can we add a keyboard shortcut? IntelliJ differentiates this with two: CMD+SHIFT+O (open any type) or CMD+O (open types only in current project)

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/QuickOpen.java
```diff
@@ -0,0 +1,135 @@
...
+                ArrayList<QuickPickItem> items = new ArrayList<>();
+                for (Project prj : prjs) {
+                    for (SourceGroup sg : ProjectUtils.getSources(prj).getSourceGroups(JavaProjectConstants.SOURCES_TYPE_JAVA)) {
+                        FileObject root = sg.getRootFolder();
+                        for (ElementHandle<TypeElement> type : ClasspathInfo.create(root).getClassIndex().getDeclaredTypes("", ClassIndex.NameKind.PREFIX, EnumSet.of(ClassIndex.SearchScope.SOURCE))) {
+                            String qualifiedName = type.getQualifiedName();
+                            int idx = qualifiedName.lastIndexOf('.');
+                            String name = idx < 0 ? qualifiedName : qualifiedName.substring(idx + 1);
+                            String pkgName = idx < 0 ? DEFAULT_PKG : qualifiedName.substring(0, idx);
+                            items.add(new QuickPickItem(name, pkgName + " : " + root.toURI().toString(), null, false, new ElementData(type)));
```
sdedic, line 89: Wow, this will be performance - marvellous for transfer & client processing in projects like Micronaut :)) but we can only use the UI that  Microsoft has allowed for us.

sdedic (approved):

MartinBalin (approved):

dbalek:
> Can we add a keyboard shortcut? IntelliJ differentiates this with two: CMD+SHIFT+O (open any type) or CMD+O (open types only in current project)
I would rather leave the action without an explicitly assigned shortcut. It is more and more difficult to find free reasonable shortcuts similar on Mac, Linux and Win (almost all "nice" shortcuts are taken). It is IMHO better to let user to assign reasonable shortcuts he/she is used to use for these actions than assign the strange shortcuts just because their key combinations are free.

PR #4194, opened 2022-06-03
Try to generate maven project skeleton from our dependencies

ebarboni:
Try to generate a maven multimodule pom from externals libraries located in nbbuild/build/mavenpoms.

remove getmavencoordinate I introduced a year ago to debug nb-repository

On this generated folder I tried
 mvn org.owasp:dependency-check-maven:7.1.0:check it could works (api rate limitation) some blocker due to too old dep.

 was thinking to push the pseudo maven project to https://github.com/apache/netbeans-temp and put a dependatbot on it.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
@@ -31,8 +31,8 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Task;
 
@@ -50,7 +50,8 @@ public class MavenSkeletonProject extends Task {
     private File mavenfolderFile;
 
     /**
-     * JUnit-format XML result file to generate, rather than halting the build.
+     * Folder to put the skeleton project in.
+     * @param report
      */
     public void setMavenFolder(File report) {
         this.mavenfolderFile = report;
@@ -66,13 +67,15 @@ public class MavenSkeletonProject extends Task {
                 modules.add("nbbuild");
             } else {
                 Path nbAllPath = nball.toPath();
-                modules = new TreeSet<>(
-                        Files.walk(nbAllPath)
-                                .filter(p -> Files.exists(p.resolve("external/binaries-list")))
-                                .map(p -> nbAllPath.relativize(p))
-                                .map(p -> p.toString())
-                                .collect(Collectors.toSet())
-                );
+                modules = new TreeSet<>();
+                try ( Stream<Path> walk = Files.walk(nbAllPath)) {
+                    modules = new TreeSet<>(
+                            walk.filter(p -> Files.exists(p.resolve("external/binaries-list")))
+                                    .map(p -> nbAllPath.relativize(p))
+                                    .map(p -> p.toString())
+                                    .collect(Collectors.toSet())
+                    );
+                }
             }
             try {
                 buildLibPomForMaven();
@@ -89,7 +92,9 @@ public class MavenSkeletonProject extends Task {
     private void buildLibPomForMaven() throws IOException {
         Path pseudoMaven = mavenfolderFile.toPath();
         if (Files.exists(pseudoMaven)) {
-            Files.walk(pseudoMaven).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+            try ( Stream<Path> walk = Files.walk(pseudoMaven)) {
+                walk.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+            }
         }
         Path pseudoMavendirectory = Files.createDirectory(pseudoMaven);
         Path parentPom = Files.createFile(pseudoMavendirectory.resolve("pom.xml"));
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
@@ -1,0 +1,208 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


mbien:
interesting. I am wondering if we could add an ant target which builds this temporary mega pom. After that we could use `versions-maven-plugin` and call `versions:display-dependency-updates` (it can be configured with a [filter](https://github.com/apache/roller/blob/2deb12060cfb4b980a2075abc5b26a8b526bc0aa/pom.xml#L104) if needed).

This could be used by devs via CLI or maybe in manually triggered workflows / jenkins etc.

I don't think dependabot will be of much use since it can't translate the upgrade back to NetBeans' own dependency management system. Even if it could, it would cause too many PRs and too much CI load to be feasible in my opinion. NB has so many dependencies (many redundant on a per-module basis) that it wouldn't surprise me to get hundreds of PRs once its turned on.

ebarboni:
I may be wrong but the default config for dependabot is for security dependencies so maybe less PR. On the nbm** repo I configured dependabot manually to get the version upgrade.  

The generated pom are one parent pom + child pom for each external with maven coordinate. But it's not always the highest version the better we may need last upgrade in major. But both plugin can be run against the poms  I guess

ebarboni:
@pepness  does a lot of version upgrade maybe good to have his opinion.

you are right @mbien the PR dependencies based on maven would be useless in our case.

ebarboni:
kind of a first approach. use mvn site site:stage on the folder.
to run successfully you need to comment on parent pom idehtmlvalidation,plafromlibsjavafx, nbbuild
it will generate site maybe resulting should be discusss on private ml

ebarboni:
about to merge doing the mvn site or mvn site:stage will be left to dev on their box. Not sure is good to have the report "public"

ebarboni:
@mbien  not sure it's maven alone maybe CI too will not change the NetBeans itself

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/VerifyLibsAndLicenses.java
```diff
@@ -76,6 +79,12 @@ public void setNball(File nball) {
     public void setReport(File report) {
         this.reportFile = report;
     }
+    
+    private File mavenfolderFile;
+    /** JUnit-format XML result file to generate, rather than halting the build. */
+    public void setMavenFolder(File report) {
```
matthiasblaesing, line 85: This looks strange. The `VerifyLibsAndLicenses` task is pretty specificy. Why is this here? Could this be extracted into a seperate task?

matthiasblaesing: I would prefer a separate task - as said generated a pom.xml has little in common with checking licenses (IMHO). They are as near as the already separate download task.

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
```diff
@@ -0,0 +1,208 @@
...
+    public @Override
+    void execute() throws BuildException {
+        try { // XXX workaround for http://issues.apache.org/bugzilla/show_bug.cgi?id=43398
+            if (getProject().getProperty("allmodules") != null) {
+                modules = new TreeSet<>(Arrays.asList(getProject().getProperty("allmodules").split("[, ]+")));
+                modules.add("nbbuild");
+            } else {
+                Path nbAllPath = nball.toPath();
+                modules = new TreeSet<>(
+                        Files.walk(nbAllPath)
```
mbien, line 70: most of the streams which are returned from `Files.*` must be closed after use. Would be good to wrap the stream in an arm-block and do the rest in the block.

I have some jackpot rules here (since its easy to forget - most streams don't have to be closed):
https://github.com/mbien/jackpot-inspections/blob/5b518358408dda4008be5b7f2602aa3304a0c8fc/ProbableBugs.hint#L32-L40

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/MavenSkeletonProject.java
```diff
@@ -66,13 +67,15 @@ void execute() throws BuildException {
...
-                modules = new TreeSet<>(
-                        Files.walk(nbAllPath)
-                                .filter(p -> Files.exists(p.resolve("external/binaries-list")))
-                                .map(p -> nbAllPath.relativize(p))
-                                .map(p -> p.toString())
-                                .collect(Collectors.toSet())
-                );
+                modules = new TreeSet<>();
+                try ( Stream<Path> walk = Files.walk(nbAllPath)) {
+                    modules = new TreeSet<>(
```
mbien, line 72: the 2x new TreeSet() is because you don't want it to be null on exception? Wouldn't it be easier to initialize the private field as empty set so that you don't have to do this here twice?

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/VerifyLibsAndLicenses.java
```diff
@@ -94,13 +95,15 @@ public void setHaltonfailure(boolean haltonfailure) {
...
-            modules = new TreeSet<>(
-                    Files.walk(nbAllPath)
-                            .filter(p -> Files.exists(p.resolve("external/binaries-list")))
-                            .map(p -> nbAllPath.relativize(p))
-                            .map(p -> p.toString())
-                            .collect(Collectors.toSet())
-            );
+            modules = new TreeSet<>();
+                try ( Stream<Path> walk = Files.walk(nbAllPath)) {
+                    modules = new TreeSet<>(
```
mbien, line 100: same here

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/VerifyLibsAndLicenses.java
```diff
@@ -426,11 +429,13 @@ private void testLicenses() throws IOException {
...
     private void testLicenseinfo() throws IOException {
         Path nballPath = nball.toPath();
-        List<File> licenseinfofiles = Files.walk(nballPath)
-                .filter(p -> p.endsWith("licenseinfo.xml"))
-                .map(p -> p.toFile())
-                .collect(Collectors.toList());
-
+        List<File> licenseinfofiles = new ArrayList<>();
+        try ( Stream<Path> walk = Files.walk(nballPath)) {
+            licenseinfofiles = walk
```
mbien, line 434: and here

mbien (approved):
thanks. looks good to me!

tested with:
```bash
$ant verify-libs-and-licenses
...
$tree nbbuild/build/mavenpoms/
nbbuild/build/mavenpoms/
├── pom.xml
├── cpplitecpplitedebugger
│   └── pom.xml
├── enterprisecloudoracle
│   └── pom.xml
├── enterprisej2eeapis
│   └── pom.xml
...
```

ebarboni:
thks merging

mbien:
you forgot to squash :(

PR #4200, opened 2022-06-05
Regex checker UI improvements

mbien:
wanted to check out the new regex checker with a fairly long regex (output window exception matcher from #4180), but noticed that the text fields had all fixed sizes despite it being a top component.

So I tried to fix this first ;)
 - everything resizes now properly (the example generator window too)
 - added vertical splitter as bonus
 - moved the error msg label a little bit around (try invalid regex)
 - noticed that the form had a unused JFrame stored, removed that (might have been the first version of the RegexExampleGeneratorPanel)

before.
regex-checker-baseline

after:
regex-checker-layout

@mishrasandeep what do you think?

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/jdk/Bundle.properties
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/jdk/Bundle.properties
@@ -111,15 +111,17 @@ DESC_AnnoProcessor_ObsoleteSupportedSource=If the project's source level is grea
     another useless warning during compilation. <p/>\
     It is recommended to return at least the current project's source level. \
     For future compatibility, consider to return <b>SourceVersion.latest()</b>; most Processors are not affected by future language changes.
-CheckRegexTopComponent.regexLabel.text=Re&gular Expression:
-CheckRegexTopComponent.exampleLabel.text=E&xample:
-CheckRegexTopComponent.strictCheckBox.text=Str&ict Check
 
 #{0}: PatternSyntaxException.getDescription()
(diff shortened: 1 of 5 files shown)


mishrasandeep:
@mbien It looks good. Can a horizontal scrolling option also be added when the example text box is small but the input is lengthy (multiple lines)?

mbien:
> It looks good. Can a horizontal scrolling option also be added when the example text box is small but the input is lengthy (multiple lines)?

@mishrasandeep good catch. I'll take a look why the scroll bar doesn't show up - it should be there agreed.

lkishalmi:
Definitely an improvement. Though to my eye the "Strict Check" checkbox is "floating" around, out of context.
Also I do not see the two buttons down there in the 3rd of June dev builds. Am I missing something?

mishrasandeep:
@mbien The highlights that we above are the substrings that are matching the regular expression. Strict check will basically check for the entire string and show a no match even if there are substrings that match.

lkishalmi (approved):
Well, it is definitely better than we are having now. If I could came up a better one than this, I should file a PR for it. Until then (if that ever would come), let's give this a go!

Thank you!

mbien:
I just force pushed to add a tooltip for the check box.

The text area has a graphical "glitch". The green check mark is overdrawn if the scroll bar is used (on linux at least), I don't see this as a show stopper for now.

There are certainly some more tweaks possible to further improve it - but this could be done in other PRs as you said. Going to merge once green again.

PR #4204, opened 2022-06-07
Issue #3886: Unused in package misbehavior - fix.

### LINKED ISSUE #3886: Unused in package misbehavior
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

Consider a simple Maven project, with two classes:
In main sources:
```
package test.mavenproject;

public class Mavenproject3 {

    static void test() {
        System.out.println("Hello World!");
    }

}
```
In test sources:
```
package test.mavenproject;

import org.junit.jupiter.api.Test;

public class Mavenproject3Test {
    
    public Mavenproject3Test() {
    }
    
    @Test
    public void testMain() {
        Mavenproject3.test();
    }
    
}
```

There are at least two issues observable with these files:
 - `test` is marked as unused, and a fix is shown to remove it - but, obviously, the (test) build fails when removed
 - when having several files opened in the editor, and switching between them, the following can happen (reason: the index query returns null when the task is cancelled):

```
        at org.netbeans.modules.java.editor.base.semantic.UnusedDetector.isUnusedInPkg(UnusedDetector.java:322)
        at org.netbeans.modules.java.editor.base.semantic.UnusedDetector.findUnused(UnusedDetector.java:138)
        at org.netbeans.modules.java.editor.base.semantic.SemanticHighlighterBase.process(SemanticHighlighterBase.java:185)
        at org.netbeans.modules.java.editor.semantic.SemanticHighlighter.process(SemanticHighlighter.java:56)
        at org.netbeans.modules.java.editor.base.semantic.SemanticHighlighterBase.run(SemanticHighlighterBase.java:112)
        at org.netbeans.modules.parsing.impl.TaskProcessor.callParserResultTask(TaskProcessor.java:561)
[catch] at org.netbeans.modules.parsing.impl.TaskProcessor$RequestPerformer.run(TaskProcessor.java:786)
        at org.openide.util.lookup.Lookups.executeWith(Lookups.java:279)
        at org.netbeans.modules.parsing.impl.TaskProcessor$RequestPerformer.execute(TaskProcessor.java:702)
        at org.netbeans.modules.parsing.impl.TaskProcessor$CompilationJob.run(TaskProcessor.java:663)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

Looking at the patch in:
https://github.com/apache/netbeans/pull/3841/files

There's a number of other issues obvious:
 - no canceling (yes, it is true there was no canceling before - that was wrong even then when the task was much faster)
 - use of `TreePathScanner`: should either be `ErrorAwareTreePathScanner` or `CancellableTreePathScanner`. Both of the scan into erroneous subtrees
 - the change seems to be severely under-tested - even if the indirect testing through the semantic highlighting was enough for the query part (which I am not sure of), there's a lot of fairly tricky code added on the removal side, and there appears to be no tests whatsoever for this removal part.
 - it would probably be better to coalescence the work in `UnusedDetector`, so that a single file is not re-parsed/re-attributed multiple times
 - nit: there's no entry in apichanges for the new method in `JavaFixUtilities`.

### How to reproduce

_No response_

Apache NetBeans 12.6

### Operating System

Linux

### JDK

JDK 8

### Apache NetBeans packaging

Own source build

### Anything else

_No response_

No

### Code of Conduct

Yes

dbalek:
Attempt to fix #3886.


java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
```diff
@@ -34,6 +34,7 @@
 import com.sun.source.util.TreePathScanner;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.nio.file.attribute.FileOwnerAttributeView;
```
jlahoda, line 37: This import is probably unused?

jlahoda (approved):
Makes sense to me. I'd be happier if there were some more tests, though.

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
dbalek: Unused imports removed. Thanks.

PR #4205, opened 2022-06-07
JDK downloader improvements

mbien:
originally I only wanted to add a column so that users can distinguish musl builds from libc builds, but I noticed a few other issues I could fix.

 - layout did not resize properly which also caused the scrollbars in the advanced tab to not work
 - there were several issues with components not being centered or not having enough space between each other
 - added the LibC column, distribution, CPU arch and EA filters
 - misc changes like giving components proper names

more details in the individual commits

looks similar, just with a bit more functionality and a working layout (try to resize the version shipping with NB 14). "Quick" tab and loading panel got some minor updates too:
advanced

quick select panel:
quck


neilcsmith-net:
Generally looks good.  Can take a better look next week (busy with installers atm) but merge before then if you want to.

I was considering looking at a bigger revamp of this UI in NB15.  Updating to Disco v2 threw up some quirks and missing info.  Possibly also worth comparing with other IDE plugins (eg. https://marketplace.eclipse.org/content/discoeclipse ) and I would say online use at https://foojay.io/download/ (except doesn't work at all for me in Firefox!)

A key question might be whether we should only support listing and downloading JDKs that can be registered in the running IDE?  Or at least shouldn't register ones that cannot run.

Maybe keep advanced as is, but we add a few options on to quick selection (eg. vendor), and possibly move to selection box over slider?  The slider ran out of space hence change of min version in NB14! :smile:

java/java.disco/src/org/netbeans/modules/java/disco/BundleTableModel.java
```diff
@@ -41,57 +69,37 @@ public BundleTableModel(final List<Pkg> bundles) {
...
-            case 0 :
-            case 1 :
-            case 2 :
-            case 3 :
-            case 4 :
-            case 5 : return columnNames[col];
-            default: throw new IllegalArgumentException("Column not found " + col);
-        }
+    @Override
+    public int getRowCount() {
```
emilianbold, line 82: Minor remark: I wish the order of methods would have been kept the same. Diff would have been easier.

java/java.disco/src/org/netbeans/modules/java/disco/ArchitectureListCellRenderer.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.java.disco;
+
+import eu.hansolo.jdktools.Architecture;
```
emilianbold, line 21: I thought this used `org.foojay`? But perhaps newer JARs switched to `eu.hansolo`...

emilianbold (approved):
Looks good to me.

I think originally there were more filters but they took too much space so we picked only a few.

The default architecture pick was an interesting topic since the Apple M1s came around that time and people on M1 actually didn't have good M1 JDKs so they had to use X64 JDKs and run them through Rosetta. That's a minor concern nowadays.

java/java.disco/src/org/netbeans/modules/java/disco/BundleTableModel.java
mbien: should I move it back?

mbien:
(neil)
> A key question might be whether we should only support listing and downloading JDKs that can be registered in the running IDE? Or at least shouldn't register ones that cannot run.

(emilian)
> The default architecture pick was an interesting topic since the Apple M1s came around that time and people on M1 actually didn't have good M1 JDKs so they had to use X64 JDKs and run them through Rosetta. That's a minor concern nowadays.

I thought about that too and came to same conclusion. Because in theory selecting cpu arch in a filter could be done by the IDE (default choice is already picked by the IDE). But i added the filter anyway for three reasons:

 - there might not always be a 1:1 mapping between JDK build and user hardware (see old 32bit builds etc)
 - its difficult to figure out what the system can actually run. For example my workstation here (64 bit linux kernel) can also run aarch64 (and more) since i have qemu configured that way. I can run aarch64 containers before deploying them to raspis or server grade arm systems. Mac + rossetta would be another example.
 - even without installing anything, the tool might be useful as simple JDK browser. ("which JDKs still run on X"). In fact, I almost wanted to add an OS combo box too (pre selected).

> 
> Maybe keep advanced as is, but we add a few options on to quick selection (eg. vendor), and possibly move to selection box over slider? The slider ran out of space hence change of min version in NB14! smile

thats a good idea. I might give that a try and add it.

thanks for the review @emilianbold  & @neilcsmith-net

mbien:
- added the distribution filter on the quick select panel
- fixed latest checkbox on adv panel (it shows now everything if unchecked, as expected)
- EA builds are now displayed if selected (adv panel)
- filter out musl on for quick select
- cleared up what the radio buttons do, I actually didn't expect them to influence the package type

added second screenshot to PR text

emilianbold (approved):

mbien:
looks like I broke the JDK 8 build.

matthiasblaesing:
> looks like I broke the JDK 8 build.

You are most probably seeing this: https://github.com/apache/netbeans/issues/4211
I noticed, that rerunning test did not change anything. But I saw this randomly fail on all java versions.

mbien:
- squashed some commits (wanted to squash more but i would have to reorder them first which would involve merge conflicts in form files which I don't want to deal with)
- removed guava dependency since it could be replaced with equivalent java 11 code
- bumped jdktools and discoclient to the latest update release
- made sure that everything still works

I promise I won't change anything anymore :)

> The slider ran out of space hence change of min version in NB14! smile

@neilcsmith-net I thought about the slider and I actually quite like it. It might not run out of space again since EOL versions should disappear automatically since it lists only maintained versions. I am not sure why JDK 16 is still there. Maybe the disco db isn't up2date, or maybe some vendor has a longer "STS" period. The "MTS" versions should disappear soon too I suppose.

If they don't disappear fast enough we could switch to LTS + "current" only some time in future.

mbien:
I swear the quick select panel didn't have 19 on it yesterday. For some reason the API returns it as maintained now :(

I am going to make a small change which will limit the quick select to LTS + current only. For special wishes the user would have to sue the advanced panel. Going to add some defensive checks so that the panel does not display anything > current - no matter what the API returns.

java/java.disco/src/org/netbeans/modules/java/disco/QuickPkgSelection.java
```diff
@@ -57,6 +59,9 @@ public QuickPkgSelection(QuickPanel.QuickSelection quick) {
                     switch (OS.getOperatingSystem()) {
                         case WINDOWS:
                             return ArchiveType.MSI == p.getArchiveType() || ArchiveType.EXE == p.getArchiveType();
+                        case LINUX:
```
neilcsmith-net, line 62: Finally getting a chance to look through these changes. This doesn't look right?  Should be returning things like DEB and RPM? However, how do we return the right one of those anyway?  I do wonder if "installers" should be left for the advanced panel and we remove that option entirely here?

java/java.disco/src/org/netbeans/modules/java/disco/Client.java
```diff
@@ -68,28 +68,31 @@ private synchronized DiscoClient getDisco() {
...
-    public synchronized MajorVersion getLatestSts(boolean b) {
-        return getDisco().getLatestSts(b);
+    public synchronized MajorVersion getLatestSts(boolean includeEA) {
+        return getDisco().getLatestSts(includeEA);
     }
 
     public synchronized List<Pkg> getPkgs(final Distribution distribution, final VersionNumber versionNumber, final Latest latest, final OperatingSystem operatingSystem,
             final Architecture architecture, final ArchiveType archiveType, final PackageType packageType,
-            final Boolean javafxBundled) {
+            final boolean ea, final boolean javafxBundled) {
```
neilcsmith-net, line 95: Disco does actually use `Boolean` with `null` for unspecified here! :grimacing:

neilcsmith-net:
> I swear the quick select panel didn't have 19 on it yesterday. For some reason the API returns it as maintained now :(

I've noticed similar in the past (and reported upstream) - likely something incorrectly parsed again.  Defensive checks sound good!

Changes look good - sorry only just getting around to properly trying out.  I guess I'm not putting in a PR to remove the slider or table now then! :smile:

mbien:
> > I swear the quick select panel didn't have 19 on it yesterday. For some reason the API returns it as maintained now :(
> 
> I've noticed similar in the past (and reported upstream) - likely something incorrectly parsed again. Defensive checks sound good!
> 
> Changes look good - sorry only just getting around to properly trying out. I guess I'm not putting in a PR to remove the slider or table now then! smile

I actually chatted with han solo himself via foojay slack and we found some bugs in the current data. For example some STS releases were reported as MTS (19 EA for example, you see it in the screenshot, han said it should be fixed now). There might be also a new method soon `disco.getAllMajorVersions` taking a distribution as input parameter - which would make it significantly more useful than now since the result currently only applies to zulu.

Thats also when i started to add defensive checks. For example to figure out what is EA, the code does not query if the package is EA, it simply checks if version > latest version. It also only uses LTS now, everything else is assumed to be STS without even asking. The advanced panel will still show the actual value, but this is not used for quick select - it only shows LTS and latest.

java/java.disco/src/org/netbeans/modules/java/disco/Client.java
neilcsmith-net: Agreed, not public so can be changed later - just mentioned because it was a direct mirror of upstream as far as I recall, and I thought it was weird too!

java/java.disco/src/org/netbeans/modules/java/disco/QuickPkgSelection.java
neilcsmith-net: Well, I had started on sketching a UI rewrite (which I'll now shelve most of) but I had planned to remove this option from the quick panel, and still think we should do so. Considering the point of this wizard is to register a platform, the download installer option doesn't actually allow for that. I think we should keep the quick panel to download and register only.

mbien: this would simplify things for sure. looking through the classes I noticed there were attempts to use the shell and an external process to extract the files. I thought maybe its ultimately planned to install things the same way.
But i am a linux user - and never install JDKs (or NetBeans) personally so i don't really care about this feature ;)

mbien:
> > it simply checks if version > latest version
> 
> We should perhaps look at latest version supported by the IDE for the quick panel? We discussed an API for this related to nb-javac warnings before. This was one place I thought it would also be useful.

Maybe just a warning with a link to the NB download page? ;)

A feature which would make me want to use the downloader would be if it could update them too. Currently I am simply using a one-liner bash script which downloads all update releases of all supported JDKs. 

Might be relatively easy to implement by storing vendor, package name and version in a properties file somewhere. But the module is going to need a public api at this point, since the platform window probably going to have to interact with it somehow.

java/java.disco/src/org/netbeans/modules/java/disco/QuickPkgSelection.java
neilcsmith-net: Back looking at this.  Yes, the shell extraction appears mainly to be about execution permissions, judging from [this comment](https://github.com/apache/netbeans/blob/master/java/java.disco/src/org/netbeans/modules/java/disco/archive/JDKCommonsUnzip.java#L32). That can't work as it misses out `jspawnhelper` at least. I'm working on porting the extraction and detection solution from NBPackage.

Also a Linux user, and don't see the utility for installing the JDK here.  OTOH, I do like my NetBeans as a DEB! :wink:

mbien: @neilcsmith-net feel free to do a second pass to correct things or write me a mail to me with a list you want changed. This PR started out as me wanting to simply add another column to a table - but ended up changing a lot of other things and re-doing most of the layouts in the form designer.

I am all for removing code btw, if the shell extraction code isn't needed anymore - lets remove it (I thought there were more plans for it).

PR #4220, opened 2022-06-11
Add support for Chrome DevTools Protocol for NodeJS Debugging

matthiasblaesing:
Highlight the hit breakpoint (see Marker in the "Breakpoints" Tab):

image

In the Global Node Configuration the debug protocol to be used can be selected, it defaults to the "Chrome DevTools protocol":

image

It would have been possible to select the protocol based on the version information from node, however the node from graalvm can't be safely and quickly detected, so the user has to choose.

In the per Project Settings the Node Debug Protocol can be overridden together with the node binary, as the two depend on each other:

image


Chris2011:
Just a question, will this also work out of the box for TypeScript?

matthiasblaesing:
> Just a question, will this also work out of the box for TypeScript?

I don't think so. The foundations could be used, but it needs to be integrated.

lkishalmi (approved):
Well, this is a pretty big one. I do not have JS knowledge to judge it. Also just skimmed through. Besides of the friend api-s there is nothing sore to my eyes.

As this one is a big one, I'd propose to merge early in the NB16 phase, so people could test it out,

ebarboni:
Can we merge this one ?

svenreimers (approved):

neilcsmith-net:
@matthiasblaesing we're getting close to freeze now, so need a call whether to merge for NB16, punt to NB17, or close.

matthiasblaesing:
Moved to 17 - I'd like to have another look at this.

matthiasblaesing:
I intent to merge this by next week, so if anyone wants to object, please do so now.

ebarboni:
I see no milestone but this PR looks ok. Could we make it for NB 22 ?

matthiasblaesing:
This was tested and is reviewed. I rebased on master, gave it another mini-spin and if tests run clean, I plan to merge tomorrow.

PR #4222, opened 2022-06-13
Improve stability of gradle build info extraction

matthiasblaesing:
While trying to work with the gradle projects from liferay-portal it
was observed, that loading failed with an exception. There are two
problems:

1. The liferay build provides incomplete implementation of JavaExec,
   which do not support setting the standard input

2. The generated UnsupportedOperation exception carries a null message,
   rendering the @NonNull annotation in GradleReport invalid and
   causing NullPointerExceptions in use sites.

Fixes:

1. Setting stdin is changed to be a best-effort operation. It is
   assumed that, if the setter fails, the access to stdin is not
   expected and thus does not need to be wired.

2. If a null message is passed into GradleReport, it is replaced with
   an empty string.

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NetBeansRunSinglePlugin.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NetBeansRunSinglePlugin.java
@@ -19,19 +19,21 @@
 
 package org.netbeans.modules.gradle.tooling;
 
-import java.util.ArrayList;
 import static java.util.Arrays.asList;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.gradle.api.logging.LogLevel;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.tasks.JavaExec;
 import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.process.CommandLineArgumentProvider;
-
 /**
  *
  * @author Laszlo Kishalmi
  */
 class NetBeansRunSinglePlugin implements Plugin<Project> {
+    private static final Logger LOG = Logging.getLogger(NetBeansRunSinglePlugin.class);
 
     private static final String RUN_SINGLE_TASK = "runSingle";
     private static final String RUN_SINGLE_MAIN = "runClassName";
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
For the liferay-portal I also did:

- Do a priming build of the liferay-portal, run `ant all`
- Copy the `gradle` folder with the `gradle-wrapper.jar` from the root folder to the `modules` folder (else the gradle wrapper was not found). After the move, the properties file is found in `modules/gradle/wrapper/gradle-wrapper.properties`, while it was in `gradle/wrapper/gradle-wrapper.properties` before. Of course the reference to local copy of the gradle distribution needs to be updated then.
- Rename the patched gradle distribution from `gradle-6.6.1.LIFERAY-PATCHED-1-bin.zip` to `gradle-6.6.1-LIFERAY-PATCHED1-bin.zip` (else the extracted gradle version string is rejected)
- I had to patch `modules/apps/portal-search-elasticsearch7/portal-search-elasticsearch7-impl/build.gradle` and replace the variable `${project.'liferay.home'}` with a fixed reference to my build bundles.

I'm not a totally happy camper (it is dog slow, blows away my memory, some test classpaths are still broken), but at least I can open the projects.

@lkishalmi @sdedic what do you think?

matthiasblaesing:
The exception I observed:

```
2022-06-13T19:56:47.896+0200 [DEBUG] [org.netbeans.modules.gradle.tooling.NetBeansRunSinglePlugin] Failed to set STDIN for Plugin: task ':apps:portal-search-elasticsearch7:portal-search-elasticsearch7-impl:autoUpdateXml'
java.lang.UnsupportedOperationException
	at com.liferay.gradle.plugins.tasks.BasePortalToolsTask.setStandardInput(BasePortalToolsTask.java:88)
	at com.liferay.gradle.plugins.tasks.DirectDeployTask_Decorated.setStandardInput(Unknown Source)
	at org.netbeans.modules.gradle.tooling.NetBeansRunSinglePlugin.lambda$apply$0(NetBeansRunSinglePlugin.java:66)
	at org.gradle.configuration.internal.DefaultUserCodeApplicationContext$CurrentApplication$1.execute(DefaultUserCodeApplicationContext.java:100)
	at org.gradle.api.internal.DefaultCollectionCallbackActionDecorator$BuildOperationEmittingAction$1.run(DefaultCollectionCallbackActionDecorator.java:95)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:395)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:387)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:157)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:242)
```

matthiasblaesing:
Thanks for the review - the liferay repository can be found here (I noticed, that that information was missing):

https://github.com/liferay/liferay-portal

PR #4230, opened 2022-06-16
Fixes for not initialized final fields added.

dbalek:
Three different fixes for initializing not initialized final fields added:
- Initialize field directly in declaration with a default value
- Initialize field in constructor(s) with a default value
- Add parameter to constructor and initialize field with its value


ppisl (approved):
I don't see problem here.

Issue #4214, opened 2022-06-09
Wrong semicolon hints for JS files


qmegas:
### Apache NetBeans version

Apache NetBeans 13

### What happened

There are 2 small ES6 paring errors. I decided to report them in a single ticket, but can split if required.

**Probem 1**
Example code:

```js
class A {
}
class B extends A {
}
```

When extending class there is incorrect warning "_Expected semicolon ; after }_" on line 4

 
**Problem 2**
Example code:

```js
const a = {
	foo: 1,
	bar: 2,
};
```
On line 3 there is incorrect warning "_Unexpected ','_" however the syntax is correct.

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10

### JDK

16.0.1

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

That issue was reported by me 2 years ago https://issues.apache.org/jira/browse/NETBEANS-4584

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
To be clear: These are not parsing errors, these are errors in the hints module. The parse happily returns the AST for these constructs, it is just the analysis, that fails to properly handle these cases.

matthiasblaesing:
Ok - here is only one bug. Problem 2 is just a misconfiguration - just deactivate the hint:

![image](https://user-images.githubusercontent.com/2179736/174154603-34016532-a123-4d7f-9627-1d08a3740384.png)

Problem 1 should be solved in the referenced PR #4232 and there is also a test build linked. Please test it.

PR #4232, opened 2022-06-16
JS: Don't generate hint for missing semicolon in generated constructor node

matthiasblaesing:
For this construct:

class A {
}
class B extends A {
}

the graaljs parser generates the default constructor:

constructor(...args) {
  super(...args);
}

This irritates the semicolon hint, as the token for the semicolon can't
be found and thus the hint is generated.

Closes: #4214


matthiasblaesing:
A test build is available here: http://doppel-helix.eu/NetBeans-dev-dev-a41856a75601a4565852a022fccf4bd237c1fea5-release.zip

matthiasblaesing:
I intent to merge this next week. If anyone wants to object, please do so now.

PR #4236, opened 2022-06-17
Avoid NPE in ant module suite's "Project Properties" dialog.

errael:
`ProjectNav > some-module-suite > properties > Libraries`
when `<module-suite-dir>/nbproject/platform.properties`
has `cluster.path=... <maven-nbm-dir>/target/nbm/clusters/extra`

This fix prevents a hung dialog (cancel terminates it).
It also makes it simpler to incorporate a maven based
nbm into an existing module suite as an external cluster.

Note that specifying a maven nbm project in cluster.path currently works for running the suite. It's just this dialog that hangs.

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ClusterUtils.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ClusterUtils.java
@@ -61,16 +61,31 @@ public final class ClusterUtils {
      * @return Path to cluster dir or <tt>null</tt> if e.g. wrong type of project.
      */
     public static File getClusterDirectory(Project prj) {
-        NbModuleProvider nbmp = prj.getLookup().lookup(NbModuleProvider.class);
-        if (nbmp != null) {
-            File jar = nbmp.getModuleJarLocation();
-            if (jar != null) {
-                return jar.getParentFile().getParentFile();
-            }
+        File clusterDir = getClusterDirectory(prj.getLookup().lookup(NbModuleProvider.class));
+        if (clusterDir != null) {
+            return clusterDir;
         }
         SuiteProvider sprv = prj.getLookup().lookup(SuiteProvider.class);
         if (sprv != null) {
             return sprv.getClusterDirectory();
+        }
+        return null;
+    }
+
+    /**
+     * Returns path to cluster dir for specified module provider.
+     * Path is returned even if it does not currently exist.
+     * @param prj module provider
+     * @return Path to cluster dir or <tt>null</tt> if e.g. wrong type of project.
+     */
+    // TODO: might be useful as public method, this usage exists elsewhere
+    private static File getClusterDirectory(NbModuleProvider nbmp) {
+        if (nbmp != null) {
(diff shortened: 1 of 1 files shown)


apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ClusterUtils.java
```diff
@@ -169,8 +169,11 @@ public static Set<ClusterInfo> evaluateClusterPath(File root, PropertyEvaluator
                     prj = _prj;
                 } else {
                     NbModuleProvider prov2 = _prj.getLookup().lookup(NbModuleProvider.class);
-                    if (prov2 != null && /* XXX is there a better way? */ cd.equals(prov2.getModuleJarLocation().getParentFile().getParentFile())) {
-                        prj = _prj;
+                    if (prov2 != null) {
+                        File fi = prov2.getModuleJarLocation();
+                        if (fi != null && /* XXX is there a better way? */ cd.equals(fi.getParentFile().getParentFile())) {
```
mbien, line 174: `getParentFile().getParentFile()`
sounds like a second potential NPE ;)

errael: Technically yes; but no, not really. The contract is that a returned File is somewhere inside a project, it's parent can't be null.

For example, elsewhere in `ClusterUtils` there's
```
            File jar = nbmp.getModuleJarLocation();
            if (jar != null) {
                return jar.getParentFile().getParentFile();
            }
```
There are other examples scattered through other files.

errael: I hadn't seen this before
```
    /**
     * Returns location of built module JAR (file need not to exist).
     *
     * Currently (6.7) used only for suite-chaining. May return <tt>null</tt>,
     * module project cannot be chained into another suite in such case.
     * @return location of built module JAR
     */
    @CheckForNull File getModuleJarLocation();
```

mbien: could be put into a utility method `getRootDir(NbModuleProvider prov)`. Would be self explanatory and could do the null check there.

errael: > could ... getRootDir(NbModuleProvider prov)

Something to consider for a rework of this ancient code, not holding my breath :-) Seems out of scope for fixing a missing null check.

mbien: if you want you can add another commit. first fixes the bug, second extracts the two occurrences into a utility method. This keeps bugfix and code improvement separate.

There is no need to wait with small cleanups like this.

errael:
Turns out `ClusterUtils` already has `getClusterDirectory(Project prj)`. The only difference is whether suite or standalone is checked first.

mbien (approved):
awesome, looks good to me.

Issue #4209, opened 2022-06-08
Incorrect variable highlighting in PHP arrow function


qmegas:
### Apache NetBeans version

Apache NetBeans 13

### What happened

When there is a ternary operator inside arrow function in PHP then variable highlighting works incorrectly.

Example code:

```php
$a = 5;
$b = fn() => $a > 5 ? $a + 1 : 10;
```

When put cursor on variable `$a` on first line, then first `$a` variable inside arrow function is also highlighted (which is correct), but second variable doesn't (which is incorrect). Same bug happens when put cursor on variables inside arrow function.

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10

### JDK

16.0.1

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

junichi11:
Reproducible. Thanks for reporting it :)

PR #4250, opened 2022-06-20
Fix variable highlighting in arrow function #4209

junichi11:
- https://github.com/apache/netbeans/issues/4209
- Fix the scope for the variable in the arrow function
- Add unit tests

#### Before

nb-gh-4209-before

#### After

nb-gh-4209-after


junichi11:
@tmysik Could you please have a look at this? If there is no problem, let's merge it :) Thanks!

junichi11:
@tmysik Thank you for your time and your review!

Issue #4246, opened 2022-06-19
Global JS functions from classpath/platform are reported as undeclared global variable


excolo-rick:
### Apache NetBeans version

Apache NetBeans 14

### What happened

Working with Promise in JS, correct syntax is highlighted as errors and/or warnings.

This occurs with `Promise.all()` and `Promise.then()`. I know it affects extensions of `Promise`, such as `fetch().then()`.

Of special note: a single instance of the syntax being erroneously highlighted will result in all other instance of `Promise` (and `fetch`) within the same file being marked as errors even though they are correct and not highlighted without the triggering error being present.

In the most basic version of the error, per first example below, every `Promise` is highlighted with the warning _Global variable "Promise" is not declared_. If the `Promise.resolve()` and `Promise.all()` are removed, the remaining `new Promise` is no longer highlighted.

In the second example, the same error is shown on the `fetch()`, but if the `.then()` statements are removed, the supposed error goes away.

### How to reproduce

Use sample `Promise.all()` code from MDN:

```
const p1 = Promise.resolve(3);
const p2 = 1337;
const p3 = new Promise((resolve, reject) => {
  setTimeout(() => {
    resolve("foo");
  }, 100);
});
Promise.all([p1, p2, p3]).then(values => {
  console.log(values); // [3, 1337, "foo"]
});
```

Use sample `fetch()` code from MDN:

```
fetch('http://example.com/movies.json')
  .then(response => response.json())
  .then(data => console.log(data));
```

### Did this work correctly in an earlier version?

No

### Operating System

Windows 11

### JDK

17.0.3.1 Windows 64 bit

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

I say this didn't work in previous versions, but I don't know whether that is the case. New machine, new install, and first time I ever make a `Promise()`.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
Please test the referenced PR - a testbuild is referenced there.

PR #4252, opened 2022-06-20
Global JS functions from classpath/platform are reported as undeclared global variable

matthiasblaesing:
The JS hint, that detects undeclared global variables only queried the JS index for properties of the "window" object. This causes the variables and functions provided by the core- and dom-stubs to be mostly ignored.

This PR drops most of the hardcoded values in favor of querying the index.

The additional commits fix and reenable JS tests for CI/CD.

Closes: #4246


matthiasblaesing:
A test build is available here: https://doppel-helix.eu/NetBeans-dev-dev-305874798a712c97027ae7418159a185159c91e0-release.zip

matthiasblaesing:
I intent to merge this some time next week. If anyone wants to object, please do so.

matthiasblaesing:
What is your problem? 

image

Hell, NetBeans does not only "know" that the objects are in the global scope, it even can do code completion on them.

qmegas:
My problem is that since NB 15 it looks like that:



Both `window` and `document` marked as not declared global variable.

neilcsmith-net:
@qmegas please open an issue rather than discussing on a merged pull request, and by the looks of it, share a minimal reproducible example so that you're discussing the same thing.

Issue #4243, opened 2022-06-18
Why is java/lang/Module repackaged in NetBeans ?


nbauma109:
### Apache NetBeans version

Apache NetBeans 14

### What happened

I have a project with org-netbeans-bootstrap-RELEASE140.jar as a dependency, which has java/lang/Module repackaged in it.
Before I get into the details of the problem, I would like to raise that it's not allowed to repackage JDK classes like that (I think on the scale of don'ts, this is much worse than using restricted intenal APIs), and as much good will I had to accomodate with that, I ran into some problems and it took me time to find that it was the issue.

I work under Eclipse and as soon I do a method search, Eclipse runs into an error. I reported the bug (https://bugs.eclipse.org/bugs/show_bug.cgi?id=578279) but as I look at the code at the line of the crash (QualifiedTypeReference.java:133), I see that the compiler is precisely trying to report that exact same problem of package conflict.

QualifiedTypeReference.java:133
```java
	    			scope.problemReporter().conflictingPackagesFromModules(splitPackage, scope.module(), this.sourceStart, (int)this.sourcePositions[typeStart-1]);
````
I confirmed that the culprit was that java/lang/Module inside the jar as when I rollbacked to the latest version that didn't have it yet (RELEASE90), I stopped getting the issue.

I'm curious, why you need to do this ?

```java
package java.lang;

/** Dummy */
public class Module {
}
````

### How to reproduce

- Create a Eclipse project with org-netbeans-bootstrap-RELEASE140.jar as a dependency (RELEASE100+)
- Create a project with a record
In my minimal example, I just have one record in src/main/java :

public record MyRecord(String s1, String s2) {
}

My build.gradle at the root my project is :

plugins {
  id 'java'
}

repositories {
  mavenCentral()
}

dependencies {
    //implementation 'org.netbeans.modules:org-netbeans-bootstrap:RELEASE90' // => OK 
    implementation 'org.netbeans.modules:org-netbeans-bootstrap:RELEASE100' // => ArrayIndexOutOfBoundsException 
}

The issue happens while building but also whenever a method search is done.

### Did this work correctly in an earlier version?

Apache NetBeans 12.3 or earlier

### Operating System

Windows

### JDK

17

### Apache NetBeans packaging

Other

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
@jtulach @dukescript ping

jtulach:
[...]

PR #4254, opened 2022-06-20
Issue #4243 - delete java/lang/Module

nbauma109:
I'd like to see where the unit tests fail and why, then this can serve to discuss of a possible strategy.


jtulach:
Doesn't even compile. The module compiles against APIs of JDK8. I'd say proper fix isn't to delete the `Module.java`, but don't package `Module.class` into the final JAR. Btw. this [trick with javaresource](https://github.com/apache/netbeans/blob/master/profiler/lib.profiler/build.xml#L45) could be a decent substitute if one compiles on JDK-11+.

matthiasblaesing:
The NetBeans unittests are expensive. Please run the applicable test locally first. There is no reason to push this onto the CI/CD infrastructure.

nbauma109:
You're right. I will run that locally.

PR #4255, opened 2022-06-21
Register @ActionsProvider.Registrations as annotation processed by DebuggerProcessor

jtulach:
Small bugfix for the debugger annotation processor. There was a bug in `DebuggerProcessor`. If one used only `@ActionsProvider.Registrations` and no other annotations, the processor wasn't triggered. Extracted from #4196 as this is a non-controversial bugfix.


lkishalmi (approved):
Looks sane, though not tested.

PR #4256, opened 2022-06-21
LSP: Do not autoimport nested classes when not necessary.

dbalek:
LSP code completion should not autoimport nested classes when not necessary.


ppisl (approved):
No complicated change

sdedic (approved):

PR #4267, opened 2022-06-22
Present project view files as tree leaves.

sdedic:
The current `foundProjects` tree view exported from LSP presents certain files as expandable, although they are not - most notably XML files. This is different from NB IDE environment - in NB IDE, the XML file is handled by `org.netbeans.modules.xml` and its XMLDataObject, which behaves fine. But in the reduced NBLS environment, XML files are handled by `openide.loaders` module and its base XMLDataObject that attempts to delegate its node to several implementations. This delegation breaks `isLeaf()` detection in `Node` interface - the loader's XML node may eventually get some children, when it rebounds to a node with children, so it is not a leaf according to NetBeans APIs.

There are other nodes, like Bundle.properties that also show subnodes in the Projects view, this time in NetBeans IDE as well. But for the LSP project navigation I believe that exposing internal structure right in the Project tree is not good.

I've added a decorator to the project view, that will force `CollapsibleState = None` for `TreeItems` that correspond to a file.


sdedic:
The sigtest failures were fixed by #4258; merging.

PR #4270, opened 2022-06-22
Recursively print type parameter's type parameters.

sdedic:
This error can be reproduced i.e. on `micronaut-core/http` project; the project mixes Groovy + Java sources. Java references Groovy classes, so our `GroovyVirtualSourceProvider` generates a java stub source during indexing, so Java indexer is happy and can resolve the class' symbols.

But it seems that the stub generator stops at 1st level of type parameters and does not print potential type parameters of the actual type parameter(s). So instead of `Map<String, Map<String, ClassNode>>` it prints just `Map<String, Map>`. Javac will then complain about usage of such (incompatible) type in the java source which expects the correct one.

The fix is simple: recursively print the complete type incl. parameters and bounds.


dbalek (approved):
Look fine to me.

ppisl (approved):

Issue #4211, opened 2022-06-08
ValidateLayerConsistencyTest is flaky


matthiasblaesing:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

The ValidateLayerConsistencyTest seems to be flaky:

https://github.com/apache/netbeans/blob/2bdbb637653cf19179963a3150add8f99d8b6977/platform/o.n.core/test/qa-functional/src/org/netbeans/core/validation/ValidateLayerConsistencyTest.java

Sometimes fails in an assertion:

```
2022-06-08T18:30:24.0613903Z     [junit] FINE [org.openide.util.lookup.implspi.ActiveQueue]: Got dequeued reference org.openide.util.lookup.AbstractLookup$ReferenceToResult@597289e4
2022-06-08T18:30:24.0614562Z     [junit] FINE [org.openide.util.lookup.implspi.ActiveQueue]: Got dequeued reference org.openide.util.lookup.AbstractLookup$ReferenceToResult@67cb9a30
2022-06-08T18:30:27.1650822Z     [junit] java.lang.AssertionError: Has to be NbRepository: org.openide.filesystems.Repository@5cf749fa Initialization stack: java.lang.Exception: initialized
2022-06-08T18:30:27.1651908Z     [junit] 	at org.openide.filesystems.Repository.<init>(Repository.java:326)
2022-06-08T18:30:27.1652502Z     [junit] 	at org.openide.filesystems.Repository$3.call(Repository.java:373)
2022-06-08T18:30:27.1652987Z     [junit] 	at org.openide.filesystems.Repository$3.call(Repository.java:368)
2022-06-08T18:30:27.1653709Z     [junit] 	at org.openide.filesystems.Repository.delayFileSystemAttachImpl(Repository.java:1018)
2022-06-08T18:30:27.1654297Z     [junit] 	at org.openide.filesystems.Repository.getDefault(Repository.java:368)
2022-06-08T18:30:27.1654794Z     [junit] 	at org.netbeans.core.startup.Main.start(Main.java:298)
2022-06-08T18:30:27.1655764Z     [junit] 	at org.netbeans.core.startup.TopThreadGroup.run(TopThreadGroup.java:98)
2022-06-08T18:30:27.1656203Z     [junit] 	at java.lang.Thread.run(Thread.java:750)
2022-06-08T18:30:27.1656523Z     [junit] 
[...]

PR #4273, opened 2022-06-22
Issue 4243: delete java/lang/Module and use dynamic proxy to implement Instrumentation across JDK versions…

nbauma109:
After JDK8, the Instrumentation interface introduced 2 methods isModifiableModule and redefineModule  having Module class in their signature. NbInstrumentation implemented the newer version of Instrumentation interface using Module, which was incompatible with JDK8. The compatibility was  ensured by an empty class Module which causes package conflict with the JDK. I propose to remove that and have a dynamic proxy that will handle all versions of Instrumentation interface thanks to an invocation handler.


nbauma109:
Just noticed I forgot to throw UnsupportedOperationException. Just pushed the change.

mbien:
can't get one of the commit validation jobs green (see #4211) and the build artifact is going to expire soon. I believe we can ignore that one.

matthiasblaesing:
Ignore it - I restarted the Travis tests, as they more or less all failed.

matthiasblaesing:
Travis is happy. I had a quick look at this and it looks sane to me.

matthiasblaesing:
The code looks sane, there are unittests, that hit the modified codepaths and still work, lets get this in.

For the record @nbauma109 I assume you read the license header and you are aware, that your contribution is under the assumption that you donate the code to the ASF. Thank you for your work.

JaroslavTulach:
Nice rewrite. Thank you.

PR #4278, opened 2022-06-23
Inline redundant variable hint added.

dbalek:
Hint for inlining redundant variables added.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
@@ -28,7 +28,6 @@ import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.ParenthesizedTree;
-import com.sun.source.tree.ReturnTree;
 import com.sun.source.tree.Scope;
 import com.sun.source.tree.StatementTree;
 import com.sun.source.tree.SwitchTree;
@@ -38,10 +37,10 @@ import com.sun.source.util.SourcePositions;
 import com.sun.source.util.TreePath;
 import org.netbeans.api.java.source.support.ErrorAwareTreePathScanner;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.lang.model.SourceVersion;
@@ -53,6 +52,7 @@ import org.netbeans.api.java.source.CompilationInfo;
 import org.netbeans.api.java.source.GeneratorUtilities;
 import org.netbeans.api.java.source.TreeMaker;
 import org.netbeans.api.java.source.WorkingCopy;
+import org.netbeans.modules.java.hints.errors.Utilities;
 import org.netbeans.modules.java.hints.suggestions.FillSwitchCustomizer.CustomizerProviderImpl;
 import org.netbeans.spi.editor.hints.ErrorDescription;
 import org.netbeans.spi.editor.hints.Fix;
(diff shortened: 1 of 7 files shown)


java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
```diff
@@ -400,4 +402,83 @@ protected void performRewrite(final TransformationContext ctx) {
 
     }
 
+    @Hint(displayName = "#DN_org.netbeans.modules.java.hints.suggestions.Tiny.inlineRedundantVar", description = "#DESC_org.netbeans.modules.java.hints.suggestions.Tiny.inlineRedundantVar", category="suggestions", hintKind=Kind.ACTION, severity=Severity.HINT)
```
mbien, line 405: nitpick: I believe it would improve readability (esp for future diffs) if this wouldn't be a single line, but this is consistent with the rest of the file so its not really a complaint - just my opinion

java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
```diff
@@ -400,4 +402,83 @@ protected void performRewrite(final TransformationContext ctx) {
...
+    @TriggerPattern(value="$type $var = $init; $statements$ return $var;")
```
mbien, line 406: the trigger seems to be a bit greedy. The hint is active even before the variable is declared.
inline

mbien (commented):
I like that there is some activity in the hints area :)

I have to wonder, given how much code is needed to implement this, if this wouldn't be better solved by using the rule language:
```java
"local can be inlined":
$V $v = $init;
$body$;
return $v;
:: !referencedIn($v, $body$) =>
$body$;
return $init;
;;
```
(this isn't greedy btw, it will only activate when the cursor is over the var declaration)

java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
dbalek: Let it consistent with the rest of the file.

dbalek:
> I have to wonder, given how much code is needed to implement this, if this wouldn't be better solved by using the rule language
I don't know how to register hints using the rule language to be activated based on cursor position either.
However, inspired by your suggestion, I have simplified the hint registration to couple lines of code only.

java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
```diff
@@ -400,4 +402,18 @@ protected void performRewrite(final TransformationContext ctx) {
...
     }
 
+    @Hint(displayName = "#DN_org.netbeans.modules.java.hints.suggestions.Tiny.inlineRedundantVar", description = "#DESC_org.netbeans.modules.java.hints.suggestions.Tiny.inlineRedundantVar", category="suggestions", severity=Severity.HINT)
+    @TriggerPattern(value="$before$; $type $var = $init; $statements$; return $var;")
+    public static ErrorDescription inlineRedundantVar(HintContext ctx) {
+        TreePath var = ctx.getVariables().get("$var");
+        Collection<? extends TreePath> stats = ctx.getMultiVariables().get("$statements$");
+        if (var != null && stats != null) {
+            Element element = ctx.getInfo().getTrees().getElement(var);
+            if (element != null && element.getKind() == ElementKind.LOCAL_VARIABLE && !Utilities.isReferencedIn(ctx.getInfo(), var, stats)) {
```
sdedic, line 412: Q: what if `$statements` have some effects on the data used in the  `$init` expression ?

mbien (approved):
looks good (light bulb has the right scope now as far as I see)

good choice of using `referencedIn` from the jackpot built in functions. It already has some test coverage since #3540, might be even a little bit faster since it has an early exit fast path.

java/java.hints/src/org/netbeans/modules/java/hints/suggestions/Tiny.java
dbalek: Good point. Hint changed to a more restrictive version - limited only to
- local variables that are immediately returned
- local variables that are immediately assigned to another variable and then not used

mbien: could this be solved by listing all local vars of $init (maybe there is a utility for that) and running `!referencedIn` on $statements$ for each of them?

sdedic: @mbien  I believe it can't be computed in general. If `$statements$` have some non-local side effect, then even a method call from `$init` may be affected through some field of another object.

There are at least two solutions:
- simple one: do not call it `redundant` variable, but simply suggest to inline the expression :)) It's the user responsibility to decide, and IDE to do the refactoring
- make the constraints tight:
-- does not reference any potentially non-final symbols nonlocal (that includes any invoked methods), or `statements` does not make any outbound method/constructor calls. We do not have machinery for nonlocal flow analysis.
-- `statements` does not *assign* to *any* of the symbols used in `init`

mbien: I like your first solution. I maintain a list of jackpot rules I used myself over time (https://github.com/mbien/jackpot-inspections) and the first FAQ entry is:

Q: Can I apply suggested inspections without code review and testing?
A: No.

I don't think all hints have to be fool proof it it is communicated to the user that they can be risky. Like switching out StringBuffer with Builder will work in 99% of the time, but fail in the one case when someone actually shared it between threads. I believe this is completely OK.

Thats why I am somewhat for reverting it to the more aggressive version which uses `!referencedIn` as best effort attempt to suggest the right thing.

PR #4284, opened 2022-06-26
improve php test performance

mbien:
lets see if this has any side effects, I didn't do a full test run yet, only checked few specific tests - CI will tell for sure.

edit:
holy bytecode batman! The **total** went down from >2h to <1h

before:
`PHP on ubuntu-20.04/JDK 8 succeeded in 2h 4m 24s `
after: 
`PHP on ubuntu-20.04/JDK 8 succeeded in 41m 53s`

GC tweaks result in small improvements too, observable when comparing travis time totals, see https://github.com/apache/netbeans/pull/4284#discussion_r907747129

commit msg:

 - set event delay to 0
 - tweak JVM settings a bit (little bit more heap, use throughput GC)

the event delay caused some tests to lose >1s per test method with CPU mostly
idling. Tests with a lot of cases were observed to complete 4-5x faster after
this tweak.

e.g: OccurrencesFinderImplTest: ~10 mins -> ~2 mins

### DIFF
--- a/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
+++ b/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
@@ -68,7 +68,9 @@ import org.openide.util.WeakListeners;
 public final class PathRegistry implements Runnable {
 
     private static final boolean FIRE_UNKNOWN_ALWAYS = false;
-    /*test*/ static final int FIRER_EVT_COLLAPSE_WINDOW = 500;
+
+    // property set/field used in tests
+    static final int FIRER_EVT_COLLAPSE_WINDOW = Integer.getInteger(PathRegistry.class.getName()+".FIRER_EVT_COLLAPSE_WINDOW", 500);
 
     private static PathRegistry instance;
     private static final RequestProcessor firer = new RequestProcessor ("Path Registry Request Processor"); //NOI18N
--- a/nbbuild/templates/projectized.xml
+++ b/nbbuild/templates/projectized.xml
@@ -290,7 +290,7 @@
     <target name="test-lib-init" depends="-init-bootclasspath-prepend,init,-build-libs.junit4">
         <path id="test.unit.lib.cp"/>
         <property name="test.jms.flags" value=""/>
-        <property name="test.run.args" value="-ea -Xmx700m ${metabuild.jms-flags.jvm} ${test.jms.flags} -XX:+IgnoreUnrecognizedVMOptions"/>
+        <property name="test.run.args" value="-ea -Xms1200m -Xmx1200m -XX:+UseParallelGC ${metabuild.jms-flags.jvm} ${test.jms.flags} -XX:+IgnoreUnrecognizedVMOptions"/>
         <property name="extra.test.libs.dir" location="${test.dist.dir}/extralibs"/>
         <macrodef name="test-init-nborg">
             <attribute name="test.type"/>
--- a/ide/csl.api/test/unit/src/org/netbeans/modules/csl/api/test/CslTestBase.java
+++ b/ide/csl.api/test/unit/src/org/netbeans/modules/csl/api/test/CslTestBase.java
@@ -189,6 +189,11 @@ import static org.openide.util.test.MockLookup.setLookup;
  */
 public abstract class CslTestBase extends NbTestCase {
 
+    static {
+        // testing performance: set scanner update delay to 0
+        System.setProperty(PathRegistry.class.getName()+".FIRER_EVT_COLLAPSE_WINDOW", "0");
+    }
+
     public CslTestBase(String testName) {
         super(testName);
     }
--- a/ide/ide.kit/nbproject/project.properties
+++ b/ide/ide.kit/nbproject/project.properties
@@ -42,7 +42,7 @@ test.config.versioning.includes=\
 test.timeout=900000
 requires.nb.javac=true
 
-test.run.args=-ea -Xmx700m -XX:+IgnoreUnrecognizedVMOptions \
+test.jms.flags=\
  --add-opens=java.base/java.net=ALL-UNNAMED \
  --add-opens=java.prefs/java.util.prefs=ALL-UNNAMED \
  --add-opens=java.desktop/javax.swing=ALL-UNNAMED \
@@ -50,4 +50,4 @@ test.run.args=-ea -Xmx700m -XX:+IgnoreUnrecognizedVMOptions \
  --add-opens=java.desktop/javax.swing.text=ALL-UNNAMED \
  --add-opens=java.desktop/sun.awt=ALL-UNNAMED \
  --add-modules=jdk.jdwp.agent,jdk.attach,jdk.jdi,jdk.jshell,java.compiler,jdk.compiler,jdk.management,jdk.unsupported,jdk.internal.le,jdk.internal.ed,jdk.internal.opt,jdk.internal.jvmstat \
- --add-exports=jdk.jdi/com.sun.jdi=ALL-UNNAMED \
+ --add-exports=jdk.jdi/com.sun.jdi=ALL-UNNAMED
--- a/java/java.hints/nbproject/project.properties
+++ b/java/java.hints/nbproject/project.properties
@@ -115,8 +115,8 @@ test.config.batch2-vanilla-javac.includes=\
 test.config.batch2-vanilla-javac.excludes=\
     ${vanilla-javac.excludes}
 
-test.run.args=-ea -Xmx700m -XX:+IgnoreUnrecognizedVMOptions \
+test.jms.flags=\
  --add-opens=java.base/java.net=ALL-UNNAMED \
  --add-opens=java.prefs/java.util.prefs=ALL-UNNAMED \
  --add-opens=java.desktop/javax.swing.text=ALL-UNNAMED \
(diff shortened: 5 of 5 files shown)


mbien:
in case anyone is curious how i found this:
1) filtered the log with a shell script to find the top 10 slowest tests, started with the slowest
2) while looking at async-profiler flamegraphs and VisualVM GC graphs, I noticed that the CPU use was unusually low (basically idle!)
3) now I measured setup and teardown time and found an unusual pattern:
```
    [junit] setup: 0.773
    [junit] teardown: 0.502
    [junit] setup: 0.828
    [junit] teardown: 0.502
    [junit] setup: 0.827
    [junit] teardown: 0.503
    [junit] setup: 0.835
    [junit] teardown: 0.503
...
```
500 ms looked very suspicious -> it had to be an artificial delay.

4) Finding it was not easy though. The integration tests are synced via keywords in loggers with the NB instance, involving multiple threads and countdown latches - so flamegraphs of locks or thread diagrams didn't help - all I saw were threads waiting for something. It did also not help that i looked in the wrong area at first (RequestProcessor itself). At some point I started to search for "500" while debugging and found this in the event system code:

https://github.com/apache/netbeans/blob/c084119009d2e0f736f225d706bc1827af283501/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java#L958-L963

which is used to debounce events during api call bursts (and everything waited for this event). I set it to 0 and:
```
    [junit] setup: 0.21
    [junit] teardown: 0.001
    [junit] setup: 0.229
    [junit] teardown: 0.001
    [junit] setup: 0.223
    [junit] teardown: 0.001
    [junit] setup: 0.229
    [junit] teardown: 0.001
...
```
Done. -1s per test method (the test has 463 methods!). This made the test run 5x faster, and apparently reduced the total from >2h to <1h since many other tests were also affected.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
```diff
@@ -68,7 +68,7 @@
 public final class PathRegistry implements Runnable {
 
     private static final boolean FIRE_UNKNOWN_ALWAYS = false;
-    /*test*/ static final int FIRER_EVT_COLLAPSE_WINDOW = 500;
+    /*test*/ static int FIRER_EVT_COLLAPSE_WINDOW = 500;
```
matthiasblaesing, line 71: Suggestion: Set this from a system property (Full classname + Fieldname?) and default to 500. That way the field can be kept static final and there is no need to twiddle with accessibility of the field.

nbbuild/templates/projectized.xml
```diff
@@ -290,7 +290,7 @@
     <target name="test-lib-init" depends="-init-bootclasspath-prepend,init,-build-libs.junit4">
         <path id="test.unit.lib.cp"/>
         <property name="test.jms.flags" value=""/>
-        <property name="test.run.args" value="-ea -Xmx700m ${metabuild.jms-flags.jvm} ${test.jms.flags} -XX:+IgnoreUnrecognizedVMOptions"/>
+        <property name="test.run.args" value="-ea -Xms1200m -Xmx1200m -XX:+UseParallelGC ${metabuild.jms-flags.jvm} ${test.jms.flags} -XX:+IgnoreUnrecognizedVMOptions"/>
```
matthiasblaesing, line 293: Everytime I see people mess with the GC I have to ask this: Is there a measurable improvement switching from G1GC to ParallelGC and does it depend on the JVM version? Pinning the heapsize in my mind is ok.

ebresie:
From previous checks on the pho test at one point there were many large file request tests which in turn were run of each of the major platforms.  I assume this was to test perform on each platform but is this a possible area to improve on?  Maybe do it once instead of three times or are the differences in platforms significant enough to keep each one?

matthiasblaesing:
@ebresie different discussion. Please lets focus on this PR. And this is about: Improvement of test performance by reducing wait time in PathRegistry and tweaking JVM parameters.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
mbien: sure. This was only rudimentary. Will change it.

nbbuild/templates/projectized.xml
mbien: Hi Matthias,

ParallelGC is a generational throughput GC. It is available on all JDKs from all hotspot vendors and designed for tasks which are willing to trade latency for throughput (batch processing or tests). The GC can do that since it is not concurrent, and has therefore no additional runtime overhead (for both CPU and heap upkeep overhead).

G1 (new default), Shenandoha and ZGC are the concurrent GCs which try to keep latency low but have more overhead. G1 is general purpose, ZGC is mostly for very large heaps, Shenandoah scales well from small to large heaps and has also lower heap size overhead compared to other concurrent GCs, but is not included in Oracle JDKs. (picking between the 3 concurrent GCs is often application specific and requires testing)

SerialGC is like parallel just single threaded, nice for small heaps, e.g embedded or single threaded micro service deployment (Shenandoah is also nice for embedded if you have multiple cores and low throughput).

Our nodes (both gh actions and travis have 2 threads minimum, and 7GB RAM (travis might have 4GB)) so we can use ParallelGC instead of SerialGC. Further, increasing heap size even allows to run some tests without triggering a GC cycle at all. By using a non-concurrent GC we also get lower overhead, even if it doesn't run. We could likely further tweak this by increasing Xmx, but the gain might be fairly small.

Did I measure that for the tests? No since we have 14h of test jobs. But since we are looking for throughput and don't care about latency the choice is fairly simple since there is only one GC available on hotspot distributions which is made for this usecase.

junichi11:
@mbien Thank you for your great work!

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
mbien: done

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
```diff
@@ -68,7 +68,7 @@
 public final class PathRegistry implements Runnable {
 
     private static final boolean FIRE_UNKNOWN_ALWAYS = false;
-    /*test*/ static final int FIRER_EVT_COLLAPSE_WINDOW = 500;
+    /*test*/ static final int FIRER_EVT_COLLAPSE_WINDOW = Integer.getInteger(PathRegistry.class.getName()+".FIRER_EVT_COLLAPSE_WINDOW", 500);
```
matthiasblaesing, line 71: Just an idea: it might be good to add a comment here, that this is a system property to help tests, so that noone ever can claim this is some supported system property we have to retain even if the unittests run fast without twiddling with ti.

matthiasblaesing (approved):
Thank you - I left to inline comments, but both are in the region "my opinion, but feel free to ignore it".

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
mbien: tbh I was wondering if reducing this value in production (e.g to 200) would have any effect. Who knows what else is waiting for the delay to complete. Maybe certain editor features would refresh faster. 500ms is fairly long in computer science.

nbbuild/templates/projectized.xml
mbien: we can measure. We can compare average (non-php) travis runs from after this PR to before if there are any changes. What I mean is that it is annoying to test it locally since we have so many hours of tests.

Conceptionally, ParallelGC is the right GC for this use case since it is the only "throughput GC" there is. Setting a GC has also the benefit of getting predictable results. Tests on 8 would use ParallelGC but anything beyond 8 would likely use G1GC. I say likely since there is also ergonomics code in the JVM, If it does not detect enough cores or memory it might pick something else - this is not specified and could change from version to version, how well the container detection works etc. So setting a GC for tests gives us the opposite of noise IMO.

Serial and Parallel GCs fill an important niche and are still getting new features. I use ParallelGC for example for the maven daemon too. It has less memory and less computational overhead compared to the concurrent GCs which is exactly what you want for worker processes.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/PathRegistry.java
mbien: Btw I wanted to make it private but it is actually used by a static inner class (and not in tests). So the /*test*/ comment was basically false advertisement until now :)

I guess I could make it private but add a non-public getter and add a proper comment.

nbbuild/templates/projectized.xml
mbien: in another project we use EpsilonGC for tests - it is the most predictable and fastest GC since it doesn't run at all, the JVM exits if it runs out of memory ;)

matthiasblaesing: Noise was not meant in measurement gitter, but unnecessary configuration. I suspect changing the GC will not impact the test runtime only minimally and thus the configuration should not be there. Anyway, I made my point, so lets stop discussing and keep as is.

PR #4286, opened 2022-06-27
Alter FlatLaf tab colouring to address visibility.

neilcsmith-net:
Experiment inspired by discussion in #3115 to see if concerns can be addressed (mostly) by colour changes rather than changing away from the existing flat design.

Improve colour contrast between tabs and background.
Move accent colour (underline) to top of tabs.
Make view tabs active / selected behaviour same as editor tabs.

flatlaf-light
flatlaf-dark


eirikbakke:
This is good! A few proposed adjustments:
1) Could you make the active editor tab the same color as the editor toolbar? (Either by adjusting the tab color or the toolbar color or both.)
2) Can we have the tab background be the same in the sidebar and the editor area? I.e. get rid of the special background color for the "active" area. We already know which tab area is active by seeing which tab is active (and it's usually always the editor area anyway).

neilcsmith-net:
Thanks for taking a look @eirikbakke I had already tried 1&3, but have revised.

1. The problem is that either the editor tabs don't match the toolbar or the view tabs don't match their contents (or we have editor tab and view tabs different colours). Have matched with toolbar and moved the colours around a bit for more contrast. Couldn't make dark tab background dark enough so made it lighter instead, which I think looks a bit better anyway?
2. Sorry, I think this is really important for showing focus, and I deliberately gave it a little more contrast.  Fed up of typing in the wrong tab! :wink:
3. I tried without entirely already and didn't like it.  Went to try again and accidentally used the background colour - think this works quite well.

flatlaf-light2
flatlaf-dark2

neilcsmith-net:
1. That presumes that all view tabs have that different background, which they don't. That's an issue to consider elsewhere IMO. Reduce the inconsistency further, not add to it!
2. I'm talking about editor and view (eg. terminal) tabs. I don't find the bar by itself (no matter how bright we make it) enough of a focus hint.  I want to keep the background colour follows focus behaviour.
3. Yes, I agree, which is why it's _effectively_ removed. Making it the background colour makes the tab slightly shallower than the container, with the bar effectively appearing on top. I don't see the second set of images makes the unfocused tab look focused in the same way the first set did.  At the same time, I find current FlatLaf very difficult to see what the active tab is in an unfocused set.

eirikbakke:
1) OK, one way or another, there will be some inconsistency here; some TopComponents will have toolbars and others will not.
2) OK, I see the utility. It just looks bad with all of these different background colors.
3) If you think the line is necessary, then consider: Should we not just have borders around the active tab, like in normal tab components? I'm not convinced that removing all borders is actually a good thing for the UI. Completely "flat" may be too minimal.

neilcsmith-net:
Moved inactive underline colour back to what it was in first image (and close to original colour). Kept tab background same as toolbar.

mbien (approved):
I am happy that the line is on top so +1 from me.

pretty interesting that this all is done just by editing properties. Like CSS in a HashMap.

Chris2011:
I like it, just to be a little croaker, this looks a bit broken. I would prefer to not see the right line due to the editor tabs also not shows it.

image

neilcsmith-net:
@Chris2011 thanks. Should be fixed, and work like before if a different selected background colour not set. Please try.

> pretty interesting that this all is done just by editing properties. Like CSS in a HashMap.

@mbien yes, there's a wealth of stuff that can be styled by properties in a CSS fashion, as well as useful style functions - see https://www.formdev.com/flatlaf/properties-files/  Next steps might be - open some of that up so users can easily configure / create themes, and possibly look at the additional properties that are available for tabs upstream that our displayers don't support.

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatViewTabDisplayerUI.java
```diff
@@ -232,7 +232,9 @@ private void paintTabBackgroundAtScale1x(Graphics2D g, int index, int width, int
             g.fillRect(0, height - contentBorderWidth, width, contentBorderWidth);
         }
 
-        if (showTabSeparators && index >= 0) {
+        if (showTabSeparators && index >= 0 &&
+                ((!isSelected(index) && !isSelected(index + 1)
+                || selectedBackground.equals(activeBackground)))) {
```
eirikbakke, line 237: Good change. Do you need to check index < getDataModel().size() before isSelected(index + 1)?

neilcsmith-net: Underlying code supports out of range values.

neilcsmith-net:
Thanks @eirikbakke Options for border and sizing would be really good.  There's also a bunch of other configuration options in upstream [FlatTabbedPaneUI](https://github.com/JFormDesigner/FlatLaf/blob/f8c769644d4594ae330d543ce7588e917d658232/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTabbedPaneUI.java#L107-L165) that might be worth considering?

eirikbakke:
@neilcsmith-net Initially I just want to make sure that all the different variations can be supported by setting keys in FlatLAF.properties, e.g. by adding the properties needed to support borders. Making it easier to override FlatLAF.properties (without having to compile anything) might be good, though.

neilcsmith-net:
OK, squashed and updated. Will leave CI to do its thing.  I'll merge in a couple of days if no-one gets there first.

mbien:
as general note:
if a PR is updated and travis didn't run yet, please cancel the old runs here:
https://app.travis-ci.com/github/apache/netbeans/pull_requests

(gh actions are configured to do this automatically)

i just did this for this PR since travis is in slow mode again and a lot was in queue.

DevCharly (commented):
I like the light variant, but I'm not sure about the dark variant...

In the light theme, the color of the tab areas (red arrows in below screenshot) is clearly different to the background colors of surrounding components (green arrows), which makes the tab areas stand out. The selected tabs (blue arrows) are easy to recognize.

image


But in the dark theme, the **inactive** tab areas (red arrows in below screenshot) use background color that is very similar to surrounding components (green arrows), which makes the tab areas go under (e.g. on left side of below screenshot). The selected tabs (blue arrows) are not that easy to recognize as in the light theme IMHO. The **active** tabs area looks good.

image

I actually liked the darker "tab areas" on the screenshot in the first post better. Why did you switch to lighter tab areas?

DevCharly:
BTW please also test the Multi-Tabs implementation (see PR #1865).
The keys in the properties files start with `nb.multitabs.`.
Multi-Tabs do not yet support `nb.multitabs.underlineAtTop=true`, but this should be easy to add to `org.netbeans.core.multitabs.impl.TabDataRenderer`

mbien:
i was using this with the dark theme a few hours, here some observations:
- the contrast of inactive tabs (text vs background) is a bit low which makes it more difficult to read compared to NB 14, intuitively this looks as if it would be a disabled component which can't be interacted with
- inactive tab mousevoer renders with the same color as the components around it which looks as if the tab is gone just the text remains, i prefer the version of NB 14 there too

neilcsmith-net:
Thanks @DevCharly @mbien Agreed!  I'll have another look now.  I swapped to lighter because I was running out of value (HSV) space to contrast, but I agree keeping it darker would be better.

Was intending to look at consistency with multitabs separately.

neilcsmith-net:
OK, tried reverting to a darker tab background on FlatLaf dark. Getting adequate contrast with Eirik's request to use `@background` for selected tab means we have to go quite dark (original image used `@componentBackground`).  Still, I think it is probably better this way around.

flatlaf-dark4

Also changed hover colours to be (or derive from) `@componentBackground` so they don't merge.  Dark laf goes lighter on hover too, which is more like the menu than the previous tab design.

Also checked with various alternative `@accentBaseColor` settings.

Chris2011:
shouldn't the sub tabs also look the same as the parent tabs? 
image

neilcsmith-net:
@Chris2011 IMO, no.  The idea is to make the primary window tabs / regions stand out - making sub tabs similar would make it more cluttered and harder to see.  However, I would like to see if we can make the tab highlight colour follow which region is focused - ie. should also be light grey here.  But we already have that problem.

Chris2011:
Yeah, tabs in tabs are also a pattern that I don't like and hard to achieve from a design perspective. Just wanted to ask, thx for the clarification :)

eirikbakke (approved):

mbien:
is it intended that the tab content has no border? Empty tabs next to each other look like they are the same component (see usages + project tab). (If you dock them next to each other its even more visible)
no_border

eirikbakke:
@mbien I agree that this is a problem. Again, @neilcsmith-net it would be solved by making the split pane handles the same background color as the tab background. That's how it's done in both the Windows and the Aqua LAFs:

backgroundsame

Neil's objection was: "So having the dividers the same colour as the toolbar and status bar makes sense unless we add yet another colour." I don't understand this objection; it would be the same color as is used in the tab area?

eirikbakke:
A better mockup summarizing my proposed improvements...



But if you prefer I can open my own PR later...

neilcsmith-net:
@eirikbakke yes, think you're right on the split pane handles. Sorry, wrong call! Aqua has different colour for view tabs background and split pane handles here though, which is what I was on about.

This would also be helped by integrating your border fix, which would allow option of bringing borders back.

Also agree on right tab separator.

Still disagree on removing unfocused line, particularly with FlatLaf update allowing inner tabs to sync on focus.

Also not sure about changing the line colour unless you're proposing changing the accent colour?

eirikbakke:
> Aqua has different colour for view tabs background and split pane handles here though, which is what I was on about.

Oh yeah, now I understand. That is indeed unnecessary!

> This would also be helped by integrating your border fix, which would allow option of bringing borders back.

OK, will experiment in a later PR after this one is merged.

> Still disagree on removing unfocused line, particularly with FlatLaf update allowing inner tabs to sync on focus.

OK, as long as it's an intentional choice.

> Also not sure about changing the line colour unless you're proposing changing the accent colour?

I guess, yes. It was just an opportunity to make the focus indication a little brighter. Not too important though.

DevCharly:
@eirikbakke not sure whether those thick borders look good. Especially the vertical ones.
Would be nicer to have thin borders/dividers.

Maybe we should consider extending `MultiSplitPane` to support thin dividers,
where only 1px is visible, but the user still has a 6px wide invisible drag area:

image

Some years ago I wrote an article on how to do this with `JSplitPane`:
https://www.formdev.com/blog/swing-tip-jsplitpane-with-zero-size-divider/

Then it could look like this (red arrows mark thin dividers):

image

neilcsmith-net:
Having now tried it, I agree with @DevCharly - the thick border appears too much.  Pending a change like the above, I've reverted the border colours back to how they originally were, and added a fix to remove the extra top border from editor tab tabs.  I've also added side borders when "underline" is at top. If border colours are set to `@background` will still show as earlier images.

Added code to remove last separator (which for editor tabs is from @eirikbakke ), and lowered the contrast of the unfocused tab underline a little, too.

Possibly final images (again) :laughing: -

flatlight6
flatdark6

platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatEditorTabCellRenderer.java
```diff
@@ -219,11 +222,17 @@ private void paintInteriorAtScale1x(Graphics2D g, Component c, int width, int he
...
                 int underlineHeight = (int) Math.round(FlatEditorTabCellRenderer.underlineHeight * scale);
-                g.setColor(ren.isActive() ? underlineColor : inactiveUnderlineColor);
-                if (underlineAtTop)
+                if (underlineAtTop) {
+                    g.setColor(contentBorderColor);
+                    int borderWidth = (int) (1 * scale);
+                    g.fillRect(0, 0, borderWidth, height);
+                    g.fillRect(width - tabSeparatorWidth - borderWidth, 0, borderWidth, height);
+                    g.setColor(ren.isActive() ? underlineColor : inactiveUnderlineColor);
                     g.fillRect(0, 0, width - tabSeparatorWidth, underlineHeight);
```
eirikbakke, line 231: What is this code doing? It seems it's trying to draw side borders if underlineAtTop is true. That seems to be a bit ad hoc, and should probably be in a separate configuration property.

Changes here are going to start conflicting with the changes in https://github.com/apache/netbeans/pull/3115 , which I'm still hoping to integrate later. Perhaps we could keep this PR limited to simple changes to FlatLAF.properties, as was originally intended? I can help make these improvements later...

neilcsmith-net: I reverted the border colour change, and it looks odd without this. The "underline" sits on the content border now either way. I think it's needed for consistency, and I'd like this to work standalone if merged. You can enhance or remove as needed then, but this is a more sensible default IMO.

eirikbakke: OK, as long as we agree that this code might be adjusted in the future.

neilcsmith-net:
As I said, split pane colour not changed but border put back based on @DevCharly comment which I agree with having tried it.

eirikbakke:
@neilcsmith-net Wait, you don't think the lower part of https://user-images.githubusercontent.com/886243/176727415-beb3faed-755d-4e33-93bc-8b36d61cfe48.png looks better?

neilcsmith-net:
@eirikbakke honestly, no. Having tried it in its full context, I don't like the bleed of the tab background colour. And with the content borders going back to how they were it's not needed to differentiate regions either.

eirikbakke:
Alright, you can leave it at whichever overall solution you find most harmonious. If we end up making thin draggable sliders later, it will require some of these things to be redesigned in any case.

neilcsmith-net:
Squashed everything again.  Ready to merge when green from my perspective.

@eirikbakke thanks.  The main reason it came up here at all was the need originally to hide the content borders to make overline and contiguous tab content work.  Given other issues that raised and lack of consensus on how to fix, reverting to existing borders and limiting this PR solely to the tab regions (inc. minimal fixes to render correctly) made more sense.  I also changed my mind working with it - things sometimes seem right in a static screenshot, but not in use.

eirikbakke:
I did a build with this patch and https://github.com/apache/netbeans/pull/4298. Before/after screenshots on my Windows 11 laptop at 150% DPI scaling:

Before Neil's PR, light
After Neil's PR, light

Before Neil's PR, dark
After Neil's PR, dark

I'm fine with merging this!

DevCharly (approved):

PR #4287, opened 2022-06-27
Allow to find precise output folder for each language.

sdedic:
In a mixed-language project, such as `micronaut-core/http`, each language produces the output typically to a separate folder, although the output is in form of .class files.

Currently Gradle project knows the set of output folders, but does not know what language produces output in which particular one, so backwards mapping in `SourceForBinaryQuery` provides **all source folders**, which is not correct.

In addition, the result of SFBQ query in Gradle/java says that sources should be always preferred. This is fine in theory, but java uses (exclusively) its cached indexed folders instead of the real output folders - so it is able to react on editing changes even before they are compiled. But other languages behave differently.

These two things in effect lead to spurious "undersolved symbol" when Java code references some Groovy or Kotlin class:
- the `build/groovy/main/classes`, or `build/kotlin/main/classes` is on the `classpath/compile` ClassPath; that's right
- since `preferSources` is true, java support attempts of SFBQ to find sources. It gets a result. 
- And to avoid **parsing**, it replaces the source folder with cache folder, under assumption that the sources are indexed into .sig files Java can process.

The last thing is not true for groovy / kotlin, at least now. This could be improved in Java support if it did not entirely replace the output folder with the source, but used the output folder as a cache for the case the SFBQ result folder and its cache does not contain the desired type.

---

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -22,12 +22,14 @@ package org.netbeans.modules.gradle.tooling;
 import groovy.lang.MissingPropertyException;
 import java.io.File;
 import java.io.Serializable;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import static java.util.Arrays.asList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -58,6 +60,7 @@ import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.AttributeContainer;
 import org.gradle.api.distribution.DistributionContainer;
 import org.gradle.api.file.ConfigurableFileCollection;
+import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.initialization.IncludedBuild;
 import org.gradle.api.plugins.JavaPlatformPlugin;
 import org.gradle.api.specs.Specs;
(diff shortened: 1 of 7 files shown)


lkishalmi (approved):
Well it took a while to read through. It looks sane. Thank you for this!
The only thing I would add is to increment the cache version number at 

https://github.com/apache/netbeans/blob/5aafb4fd173a53181aa197d1c8a156d01d71c1fa/extide/gradle/src/org/netbeans/modules/gradle/cache/ProjectInfoDiskCache.java#L48

So it would eventually force reloading projects.

java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
```diff
@@ -365,6 +366,33 @@ public boolean outputContains(File f) {
...
+     * null, if the source type has no output directories. Returns UNKNOWN, if the 
+     * output location is not known.
+     * 
+     * @param srcType language type
+     * @return location or {@code null}.
+     * @since 1.19
+     */
+    public File getOutputClassDir(SourceType srcType) {
+        File f = outputs.get(srcType);
+        if (f == null) {
```
lbownik, line 388: this if seems redundant as UNKNOWN.equals(f) evaluates false it f == null
 so maybe

return UNKNOWN.equals(f) ? UNKNOWN : f;

will do?

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -307,6 +354,43 @@ private void detectSources(NbProjectInfoModel model) {
...
+                                                if (candidate != null) {
+                                                    candidate = null;
+                                                    break;
+                                                } else {
+                                                    candidate = p;
+                                                }
+                                            }
+                                        }
+                                    }
+                                    if (candidate != null) {
```
lbownik, line 384: how about ternary operator here?

sdedic: Fixed in 01cb17c54a

java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
sdedic: Fixed in 01cb17c54a

sdedic:
> 
> So it would eventually force reloading projects.

Fixed in 01cb17c54a

dbalek (approved):
Looks fine.

PR #4293, opened 2022-06-28
Java Record - changed state_component to record_component

singh-akhilesh:


java/java.source/src/org/netbeans/modules/java/ui/Icons.java
```diff
@@ -95,8 +95,8 @@ public static ImageIcon getElementIcon( ElementKind elementKind, Collection<Modi
                     icon = ImageUtilities.loadImageIcon(ICON_BASE + "record" + PNG_EXTENSION, false);
                     break;
                 }
-                if (elementKind.name().equals("STATE_COMPONENT")) {
-                    icon = ImageUtilities.loadImageIcon(ICON_BASE + "stateComponent" + PNG_EXTENSION, false);
+                if (elementKind.name().equals("RECORD_COMPONENT")) {
+                    icon = ImageUtilities.loadImageIcon(ICON_BASE + "recordComponent" + PNG_EXTENSION, false);
                     break;
                 }
```
mbien, line 101: nitpick: I think we could use enums here too and add it to the switch, instead of the if-else String based workaround.

mbien: thanks awesome!

PR #4297, opened 2022-06-29
Cleanup references to the generic emptySet()..

BradWalker:
Cleanup references to the generic emptySet()..

   [repeat] /home/bwalker/src/netbeans/enterprise/web.jsf/src/org/netbeans/modules/web/jsf/JSFConfigUtilities.java:250: warning: [rawtypes] found raw type: Set
   [repeat]         Set result = Collections.emptySet();
   [repeat]         ^
   [repeat]   missing type arguments for generic class Set<E>
   [repeat]     E extends Object declared in interface Set

---

### DIFF
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ExamineManifest.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ExamineManifest.java
@@ -51,7 +51,7 @@ public class ExamineManifest  {
     private String locBundle;
     private boolean publicPackages;
     private boolean populateDependencies = false;
-    private List dependencyTokens = Collections.emptyList();
+    private List<String> dependencyTokens = Collections.<String>emptyList();
    
     
     public void checkFile() throws MojoExecutionException {
@@ -278,7 +278,7 @@ public class ExamineManifest  {
         return dependencyTokens;
     }
 
-    public void setDependencyTokens(List dependencyTokens) {
+    public void setDependencyTokens(List<String> dependencyTokens) {
         this.dependencyTokens = dependencyTokens;
     }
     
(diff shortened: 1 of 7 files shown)


apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ExamineManifest.java
```diff
@@ -51,7 +51,7 @@ public class ExamineManifest  {
     private String locBundle;
     private boolean publicPackages;
     private boolean populateDependencies = false;
-    private List dependencyTokens = Collections.emptyList();
+    private List<String> dependencyTokens = Collections.<String>emptyList();
```
mbien, line 54: I don't think the generic type passed to the method on the right side is needed. It would be only needed if the compiler can't infer the type in some edge cases. Try without first unless there is a warning (same for the other occurences).

BradWalker: Hey @mbien , I thought I had tried doing just that.. But, couldn't remember.. When I do this work, I try to go as fast as possible, so sometimes I do forget.

But, here is the error that I get.. So I guess it's not possible.

```
   [depend] Deleted 2 out of date files in 0 seconds
 [nb-javac] Compiling 2 source files to /home/bwalker/src/netbeans/apisupport/maven.apisupport/build/classes
   [repeat] /home/bwalker/src/netbeans/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ExamineManifest.java:54: error: illegal start of type
   [repeat]     private List<String> dependencyTokens = Collections.<>emptyList();
   [repeat]                                                          ^
   [repeat] 1 error
  [nbmerge] Failed to build target: all-maven.apisupport
```

mbien: no you don't understand. you don't have to specify the generic type at all.

`private List<String> dependencyTokens = Collections.emptyList();`

since you already merged, please clean it up in follow up PRs because specifying generic types for methods is only needed for edge cases, like in lambdas, but not for simple declarations like this.

BradWalker: Ugg! Right you are..

I'll work on it..

PR #4301, opened 2022-06-30
Implemnted --direct-disable module option.

sdedic:
The module system offers several options for module management. `--enable` option seems to work well, as it directly calls `ModuleManager.enable` and enables the requested modules (+ dependencies) on the fly. But `--disable` was implemented for some reason through delayed-disable, but **unlike** plugin UI, it will not generate the next-startup information files and does not request a restart: it seems broken. I filed [NETBEANS-6516](https://issues.apache.org/jira/browse/NETBEANS-6516) to track this,

Option that would **disable modules on the fly** is missing completely. This was removed from the UI, perhaps because **removing** code from the memory is hard: there may be half-linked dangling listeners etc. But the option processing happens during startup before the UI even appears, so it seems a lot safe.

I have implemented `--direct-disable` option that allows to disable one or more modules; the state will be persisted in the user directory config area. It will error out and abort the execution in case of any error, such as another not disabled module that depends on the one that should be disabled.


dbalek (approved):
Looks fine. Thanks.

sdedic:
Just flaky layer validation fails; merging.

PR #4302, opened 2022-06-30
Adding symbol provider to obtain symbols for GoTo Symbol in workspace

ppisl:
Adds a Go To Symbol functionality for Groovy source files in VSCode. Symbols are obtained from IndexSearcher (part of the CSL API). This implementation should only be temporary and should be replaced by a suitable API for retrieving symbols with lazy location, which may not return immediately, but only when the symbol is selected.


dbalek:
Unfortunately, the current design of `SymbolProvider` that returns a list of `StructureElements` does not fit to the lazy source file loading mechanism currently implemented in Java symbol extraction.

dbalek (changes requested):
Unfortunately, the current design of `SymbolProvider` that returns a list of `StructureElements` IMHO does not fit to the lazy source file loading mechanism currently implemented in Java symbol extraction.

ppisl:
I changed the implementation. Now the implementation does not require any API change. On the other hand, LSP Server depends on CSL API and CSL Types.

sdedic:
> I changed the implementation. Now the implementation does not require any API change. On the other hand, LSP Server depends on CSL API and CSL Types.

IMHO dependency on CSL is not harmful, as CSL provides common infra for many languages (groovy included, *wink*).

dbalek (approved):
Good as a temporary solution. Later we should create an API to unify the Java and CSL parts.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
```diff
@@ -99,6 +102,57 @@ public static SymbolKind structureElementKind2SymbolKind (StructureElement.Kind
...
+            case FILE: return SymbolKind.File;
+            case FIELD: return SymbolKind.Field;
+            case MODULE: return SymbolKind.Module;
+            case VARIABLE: return SymbolKind.Variable;
+            case GLOBAL: return SymbolKind.Module;
+            case INTERFACE: return SymbolKind.Interface;
+            case KEYWORD: return SymbolKind.Key;
+            case OTHER: return SymbolKind.Object;
+            case PACKAGE: return SymbolKind.Package;
+            case PARAMETER: return SymbolKind.TypeParameter;
```
sdedic, line 147: Maybe `SymbolKind.Variable` could be better ?

sdedic (approved):
Nice reuse of CSL, thanks.

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/Utils.java
ppisl: Yes, it can be better. I will change it.

PR #4310, opened 2022-07-01
Output from pty is fully read after debugger exit.

entlicher:
When debugging a native application, not all output is visible after the application exits when the output is produced just before the exit.

The reason was that the `pty` process was closed right after `gdb` finished.
This change assures that the application output from `pty` process is fully read and that the debugging does not finish sooner than all output is provided.
An `OutputTest` is added.


dbalek (approved):
Looks fine.

PR #4313, opened 2022-07-01
Improve performance of JS implementation of AlternativeLocationImpl

matthiasblaesing:
#getStringLocation reads the original file, but is called from the EDT,
so invocations must be kept to a minimum. What is more the method is
used by #compareTo and thus is potentially called often.

Two optimisations are applied:

- the result of #getStringLocation is cached and only calculated once
- #compareTo was rewritten to use the relative path and offset directly.
  Before relative path and the line number was used to create an
  ordering, as offset and line number are related, switching for
  ordering shoud have no outside result

### DIFF
--- a/webcommon/javascript2.editor/src/org/netbeans/modules/javascript2/editor/navigation/DeclarationFinderImpl.java
+++ b/webcommon/javascript2.editor/src/org/netbeans/modules/javascript2/editor/navigation/DeclarationFinderImpl.java
@@ -398,6 +398,7 @@ public class DeclarationFinderImpl implements DeclarationFinder {
         private final int offset;
         private final DeclarationLocation location;
         private final IndexedElement element;
+        private String stringLocation;
 
         public AlternativeLocationImpl(IndexResult iResult) {
             this.iResult = iResult;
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@Chris2011 please have a look at this. You provided a self profiling of the IDE when invoking goto declaration. You observed a pseudo hang (the GUI was blocked, but recovered after some time). Looking at the data verifies this:

image

On the EDT `GoToSupport#chooseAlternative` is called. This fetches the possible declaration locations and sorts these. In the profile it is visible, that around 80% of the time is spend inside `DeclarationFinderImpl$AlternativeLocationImpl.compareTo()`, which uses `DeclarationFinderImpl$AlternativeLocationImpl.getStringLocation()` as basis for comparison.  `getStringLocation` reads the JS file to calculate the line number from the raw offset and is not cached.

I only tested with the build-in stubs, but already saw a runtime reduction of the call from around 4s for 10 calls to 1s for 10 calls. Please test if this improves the situation for you. Bigger projects and/or slow drives can significantly increase runtime.

Chris2011:
@matthiasblaesing thx for the PR, will check it out asap. If you already have a build just lemme know, if not I will add the code to my already built instance from your source.

matthiasblaesing:
Freshly build ;-)

https://doppel-helix.eu/NetBeans-dev-dev-3d0065135c1495c61451d655c956a11e921dfde2-release.zip

Chris2011:
So for me, just a waiting of 1s somehow. So I can live with it, because it will go though a lot of stubs. So it seems to work quite well and as you said you are caching smth as introduced in your changes.

matthiasblaesing:
Thanks for the check @Chris2011!

Issue #4261, opened 2022-06-21
Netbeans Output window uses wrong encoding


duoduobingbing:
### Apache NetBeans version

Apache NetBeans 14

### What happened

* Have `-J-Dfile.encoding=utf-8` inside the `netbeans_default_options` under `{INSTALLDIR}/netbeans/etc/netbeans.conf`
* Create a new Maven Project
* Have it print something like
  ```
  System.out.println("%s, %s".formatted(System.getProperty("file.encoding"), Charset.defaultCharset()));
  System.out.println("✨✨✨");
  ```
* Add `-Dfile.encoding=utf-8` to `exec.vmArgs` inside the project's `nbactions.xml` for the *Run* action 
* Run
* Instead of the expected '✨✨✨' or Tofu Boxes it spits out:
  ```
  utf-8, UTF-8
  âœ¨âœ¨âœ¨
  ```

### How to reproduce

Probably is a Windows only problem, as the Output window seems to just use the Windows default encoding, which in my case is WIN1252 and ignore the encoding that has been set JVM-wide.

Interpreting ✨✨✨ as WIN1252 instead of UTF-8 yields the mangled output 'âœ¨âœ¨âœ¨' which I got.

Having `JAVA_TOOL_OPTIONS=-Dfile.encoding=utf-8` and `MAVEN_OPTS=-Dfile.encoding=utf-8` set as system variables under `HKCU\Environment` also does not fix the problem.

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Windows 10

### JDK

17.0.3 (Eclipse Adoptium Temurin)

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

This does work in Netbeans 13 and before, so this bug was introduced in Netbeans 14.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

duoduobingbing:
Update:

Switching to JDK18 does not fix the problem. 
Running Netbeans under JDK18 and the Project under JDK18 yields:
```
utf-8, UTF-8
???
```

Setting `native.encoding=utf-8`, `file.encoding=utf-8` to `MAVEN_OPTS`, `JAVA_TOOL_OPTIONS`, `nbactions.xml` and/or `nbactions.xml` does nothing to change this result.

Running Netbeans under JDK18 and the Project under JDK17 yields:
```
utf-8, UTF-8
âœ¨âœ¨âœ¨
```

duoduobingbing:
[...]

PR #4321, opened 2022-07-02
Make it possible again to set the CommandLine input and output encoding

duoduobingbing:
#3836 made it so that `native.encoding` is used when on JDK17+. This creates a lock-in to whatever value `native.encoding` has, as this property is, unlike `file.encoding`, not settable. E.g. something like `java -jar -Dnative.encoding=utf-8 myjar.jar` does not change the value of `native.encoding`.

The problem is that prior to Netbeans 14 it was possible to set the Netbeans input/output to use UTF-8 via `-Dfile.encoding=utf-8` (inside `netbeans.conf`) on platforms like Windows where the native encoding is a Windows codepage (e.g. `windows1252`).  This however is not possible anymore, as `file.encoding` is ignored due to #3836, when `native.encoding` is set; which it is as of JDK17.

This PR restores the possibility to change the input/output encoding to something that is not the default encoding by introducing the settable property `netbeans.cmdline.inout.forcedencoding` which will be used instead of `native.encoding` when defined.

### DIFF
--- a/ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
+++ b/ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
@@ -215,19 +215,36 @@ public final class BaseExecutionService {
                         // output from command line invocations to strings. That encoding is
                         // derived from the system file.encoding. From JDK 18 onwards its
                         // default value changed to UTF-8.
-                        // JDK 18+ exposes the native encoding as the new system property
-                        // native.encoding, prior versions don't have that property and will
+                        // JDK 17+ exposes the native encoding as the new system property
+                        // 'native.encoding', prior versions don't have that property and will
                         // report NULL for it.
-                        // The algorithm is simple: If native.encoding is set, it will be used
+                        // The algorithm is simple: If 'netbeans.cmdline.inout.forcedencoding' is set prefer it, else check whether 'native.encoding' is set, it will be used
                         // else the old default will be queried via Charset#defaultCharset.
-                        String nativeEncoding = System.getProperty("native.encoding");
-                        if (nativeEncoding != null) {
-                            try {
-                                charset = Charset.forName(nativeEncoding);
-                            } catch (Exception ex) {
-                                LOGGER.log(java.util.logging.Level.WARNING, "Failed to get charset for native.encoding value : '" + nativeEncoding + "'", ex);
-                            }
+						String preferredEncoding = null;
+						String usedEncodingProperty = null;
+						
+						final String forcedEncodingProperty = "netbeans.cmdline.inout.forcedencoding";
+						final String forcedEncoding = System.getProperty(forcedEncodingProperty);
+						if(forcedEncoding != null){
+							preferredEncoding = forcedEncoding;
+							usedEncodingProperty = forcedEncodingProperty;
+						}
+						
+						if (preferredEncoding == null){
+							final String nativeEncodingProperty = "native.encoding";
+							final String nativeEncoding = System.getProperty(nativeEncodingProperty);
+							if(nativeEncoding != null) {
+								preferredEncoding = nativeEncoding;
+								usedEncodingProperty = nativeEncodingProperty;
+							}
+						}
+						
+						try {
+                            charset = Charset.forName(preferredEncoding);
+                        } catch (Exception ex) {
+                            LOGGER.log(java.util.logging.Level.WARNING, "Failed to get charset for '" + usedEncodingProperty + "' value : '" + preferredEncoding + "'", ex);
                         }
+						
                         if (charset == null) {
                             charset = Charset.defaultCharset();
                         }
--- a/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
@@ -202,7 +202,7 @@ public class CommandLineOutputHandler extends AbstractOutputHandler {
         private boolean skipLF = false;
 
         public Output(InputStream instream) {
-            str = new BufferedReader(new InputStreamReader(instream, getNativeCharset()));
+            str = new BufferedReader(new InputStreamReader(instream, getPreferredCharset()));
         }
 
         private String readLine() throws IOException {
@@ -701,7 +701,7 @@ public class CommandLineOutputHandler extends AbstractOutputHandler {
 
         public @Override void run() {
             Reader in = inputOutput.getIn();
-            try (Writer out = new OutputStreamWriter(str, getNativeCharset())) {
+            try (Writer out = new OutputStreamWriter(str, getPreferredCharset())) {
                 while (true) {
                     int read = in.read();
                     if (read != -1) {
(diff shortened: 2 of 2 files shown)


matthiasblaesing:
Please explain why you expect a programm to output data that is not in line with the default system encoding, why does it work with windows CMD. Why don't you change the encoding for the whole system (to my knowledge Windows can now support UTF-8)?

I don't like the usage of a new property, that noone apart from you will use.

duoduobingbing:
@matthiasblaesing 
> that is not in line with the default system encoding

That is very easy to explain, If you have an application that displays output or takes input and you are using Windows it uses a Windows codepage by default. Let's say WIN1252.
If you want to take in stuff that is not on the codepage you get garbled input and output as it will be converted to WIN1252.
E.g. ✨✨✨ will become âœ¨âœ¨âœ¨. The only possible way to enter or display ✨✨✨ is to use UTF-8 (or some other Unicode Encoding).

Let's say you have an application that is executed with `-Dfile.encoding=utf-8` in Netbeans and have it output something like `System.out.println("日本語文書");`. Prior to Netbeans 14 it was possible to set the Netbeans output encoding to UTF-8 as well (via `file.encoding=utf-8` inside `netbeans.conf`) and the correct output `日本語文書` was displayed. Now the output encoding is set to `native.encoding` causing the output of garbled characters.

> why does it work with windows CMD

Please explain what you mean by that.

> Why don't you change the encoding for the whole system?

That, sadly, is not possible. You cannot simply set the system wide encoding in Windows to UTF-8.<br>
See [this](https://superuser.com/questions/239810/setting-utf8-as-default-character-encoding-in-windows-7) for some additional context.

> I don't like the usage of a new property, that noone apart from you will use.

I asked in #4261 whether it would be possible to revert #3836 or if there is any other way, so that Windows users can switch their Netbeans back to UTF-8 like they could prior to Netbeans 14. You did not reply, so I created this PR. I did not want to step on anyones toes regarding default behavior, so I just created an opt-in property, so that people who are forced to use Windows still have the option to get correct utf-8 output.

matthiasblaesing:
> > why does it work with windows CMD
> 
> Please explain what you mean by that.

CMD is the the command line on windows (open the start menu and enter cmd). It is the base command line interface that windows supports, its successor is powershell. The idea is, you open a command prompt and run a programm. That programm issues output to STDOUT or STDERR and it will be displayed. To do this, programm and shell have to agree which encoding to use. It seems mac OS and Linux have settled on UTF-8 (at least have not seen a non UTF-8 locale on linux for a long time), not so much windows. My assumption is, that every non-broken programm will try to output data to the shell in a sane manner, so it must respect the encoding of the shell.

So does your programm output correctly when run in CMD? If not, I argue, that your program is broken, if it does, somewhere my assumptions are broken. If the former case is determined, I'm strictly against changing anything in NetBeans, that is so broad as suggested here and instead fix your programm, if it is the latter, we need to investigate more.

duoduobingbing:
> CMD is the the command line on windows (open the start menu and enter cmd)

I know what the term 'cmd' means. It was just that your statement was ambiguous.

> So does your programm output correctly when run in CMD?

I use Windows Terminal as a wrapper for CMD and Powershell but lets not sweat about the details. Does it output correctly? Yes. Running with Powershell 5 (the legacy powershell, instead of 7) or CMD you have to change the codepage to 65001 (UTF-8) first, but, yes then it works. (This works via `chcp 65001`)

To answer a potential question about this: No, changing the codepage does not change the output of `native.encoding` when running Java in cmd or powershell with a changed codepage.

As I have stated: Having UTF-8 output under Windows was working prior to your [PR#3836](https://github.com/apache/netbeans/pull/3836) in Netbeans 13 and now it does not. The reason why has already been stated by me.

A lot of people have changed their Netbeans to UTF-8 via `-Dfile.encoding=utf-8`. Just do a Google search for `netbeans windows utf-8`. You will see a lot of results.

> somewhere my assumptions are broken
That is indeed the case. 

Here is the way it works on Windows: Windows defaults to a windows codepage, but that is just a default. It does not necessarily reflect the encoding of the terminal. Updating the Terminal codepage does not change the value of `native.encoding` or `sun.jnu.encoding`. This requires the Windows user to set each application to UTF-8 seperately. JDK18 goal is to fix this on the Java side, by having `file.encoding` default to `UTF-8`.

> instead fix your programm
There is nothing to fix. The program outputs UTF-8 as it should, and Windows can handle this just fine. It worked prior to your PR in Netbeans, now it does not.

matthiasblaesing:
> > So does your programm output correctly when run in CMD?
> 
> I use Windows Terminal as a wrapper for CMD and Powershell but lets not sweat about the details. Does it output correctly? Yes. Running with Powershell 5 (the legacy powershell, instead of 7) or CMD you have to change the codepage to 65001 (UTF-8) first, but, yes then it works. (This works via `chcp 65001`)

So this is a no. So from my POV your programm makes wrong assumptions and does not work out-of-the-box with the windows CMD. A global property will then break properly working programms, so is IMHO not a solution.

duoduobingbing:
Please stop arguing in bad faith and fully read what I write and wrote because I am of the impression that you do just skim over my answers without really looking at and into my arguments.

> So from my POV [...]

Please refrain from being so close minded. It violates mutual respect.

When using Windows Terminal, it does work out of the box.

> A global property will then break properly working programms

It does not. Please provide evidence for this (seemingly) baseless statement (especially as this is a opt-in property!).

This is not a *It's just you* problem. A lot of Windows users **have** to set their applications to UTF-8 and will face the same problem once Netbeans 14+ adoption will become more widespread. I presented a use-case in my first answer.

You pretend that this PR would be huge problem and no one apart from me would have this use case. This is just your assumption, not a given fact. This PR is no problem at all, as this is a simple addition that is also opt-in. It does not break anything. It just restores functionality that was already present. Second: Why should it be only me? Do you have user metrics or something to back your claim up; or is it just a feeling?

matthiasblaesing:
Ok - I'll stop looking into this - lets hope another committer will look at this and you can convince him/her.

duoduobingbing:
> Ok - I'll stop looking into this - lets hope another committer will look at this and you can convince him/her.
You did not even look into this.

Your answers just consisted of *You are wrong* and *Your problem is not a problem*.

duoduobingbing:
I am sorry that this interaction went south but using phrases like "[...] that noone apart from you will use" and "[...] instead fix your programm" were just so disrespectful and plain rude that I was really discontent with them. I also feel like my arguments were not adressed at all, so it would be very nice if there would be a maintainer looking at this and at least respond to the issues/arguments I have raised.

If the maintainer community comes to the conclusion that my problem solution is the totally wrong way to handle my presented problem, I will gladly close this PR and accept whatever solution is offered.

The contributions that @matthiasblaesing made to Netbeans are truly tremendous and very useful for the entire userbase. It is very kind hearted to devote one's precious free time to make contributions to an open source project, so I am left completely baffled by the way @matthiasblaesing handled this.

duoduobingbing:
As I am basically left in the rain; let's wrap this up.

Where would a proper and open-minded discussion lead eventually?

First of all: 
* What is the problem? 
  The Problem is that the Netbeans encoding does not reflect the console's encoding in certain cases.
* Why does the problem occur? 
  Because Netbeans uses `native.encoding` now, which does not necessarily reflect the console's encoding.
* What was my solution? 
  Let the user set the Output encoding manually via a custom property.
* Why is this not the most desirable solution? 
  Because such a custom property would have a very low discoverability for affected users.
* What did JEP400 do and why did [PR#3836](https://github.com/apache/netbeans/pull/3836) help with that? 
  JEP400 set the default encoding that is reported via `Charset.defaultEncoding()` and `file.encoding` to `UTF-8`. But this is not the only change. Previously the output encoding matched `file.encoding` but as of JDK18+ it matches `sun.stdout.encoding`. Before PR#3836 Netbeans used `Charset.defaultEncoding()` implicitly to determine the output encoding. This of course failed with JDK18 when `sun.stdout.encoding` and `file.encoding` mismatched. 
* What would be the proper checking order for the output encoding?
  Quoting from [JEP400](https://openjdk.org/jeps/400): 
  > We will update the specifications of all standard Java APIs that use the default charset to cross-reference Charset.defaultCharset(). Those APIs include the ones listed above, but not System.out and System.err, whose charset will be as specified by [Console.charset()](https://bugs.openjdk.java.net/browse/JDK-8264208).
  So `native.encoding` is in fact not the right property to check for output encoding. The order thus should have been: `Console.charset()` → `sun.stdout.encoding` → `native.encoding` → `Charset.defaultEncoding()`
* Is this a Windows only problem? Actually it is not. It occurs as soon as the platform default encoding (`native.encoding`) does not match the console's/terminal's encoding. Here is an [example](https://github.com/elastic/elasticsearch/issues/82939), albeit not a Netbeans one. I really do not get the insistence that the "default console encoding" is of any importance. To output to the console the console's *current* encoding instead of its default encoding should be used.

[This link](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579383) shows how it has been handled by the Eclipse IDE. 

But despite my findings that `Console.charset()`/`sun.stdout.encoding` should be used instead of `native.encoding` to match the behavior introduced by JEP400, I will close this PR, as I think nobody really cares that `native.encoding` only works most of the time but not for all cases as it is in fact the wrong property to infer the output encoding from.

PR #4324, opened 2022-07-02
Invoke indexer in RepositoryUpdater grouped by mimepath of file

matthiasblaesing:
While analysing this it was found, that the GsfParserFactory#createParser
returned null as the parser for snapshots. Looking deeper, showed, that
the #createParser method was invoked with snapshots of differing mimetypes.

This works, but is a violation of the contract in ParserFactory#createParser:

> It is guaranteed that all snapshots in the collection will be
> of the same mime type and it will be the mime type, which this factory
> was registered for (ie. in MimeLookup)

If the snapshots are inconsistent, parsing still works, but its performance
suffers. To fix this the call to ParserManager#parse in RepositoryUpdater
is split. The scanned sources are grouped by their mimepath and each group
is individually passed.

The observed improvements are:

Indexing Variant 1 (tour-of-heroes)

Before: ~125s
After:  ~105s

Indexing Variant 2 (tour-of-heros with Indexability Filter)

Before: ~52s
After:  ~38s

The Indexability Filter blocks scanning for all paths with paths containing
node_modules and where the indexer is in the list: js, angular, requirejs,
knockoutjs, TLIndexer, tests.

---

### DIFF
--- a/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/RepositoryUpdater.java
+++ b/ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/RepositoryUpdater.java
@@ -3250,7 +3250,23 @@ public final class RepositoryUpdater implements PathRegistryListener, PropertyCh
                             }
                         }
                     }
-                    ParserManager.parse(sources.keySet(), new T());
+
+                    // Performance of ParserManager#parse suffers if the sources
+                    // are of mixed mimetype. ParserManager will then generate
+                    // snapshots with mixed mimetypes, which violates the
+                    // ParserFactory contract and leads to slower code paths.
+                    //
+                    // To work around this the sources are passed to the
+                    // ParserManager grouped by their mimetype
+
+                    Map<String,List<Source>> sourcesByMimeType = sources
+                            .keySet()
+                            .stream()
+                            .collect(Collectors.groupingBy(s->s.getMimeType()));
+
+                    for(List<Source> l: sourcesByMimeType.values()) {
+                        ParserManager.parse(l, new T());
+                    }
                 } catch (final ParseException e) {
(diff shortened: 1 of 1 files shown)


matthiasblaesing:
@dbalek could you please have a look at this? You introduced the grouped mode, so you might be in the best position to check.

ide/parsing.indexing/src/org/netbeans/modules/parsing/impl/indexing/RepositoryUpdater.java
```diff
@@ -3250,7 +3250,23 @@ public void run(ResultIterator resultIterator) throws Exception {
...
+                    // ParserManager grouped by their mimetype
+
+                    Map<String,List<Source>> sourcesByMimeType = sources
+                            .keySet()
+                            .stream()
+                            .collect(Collectors.groupingBy(s->s.getMimeType()));
+
+                    for(List<Source> l: sourcesByMimeType.values()) {
+                        ParserManager.parse(l, new T());
+                    }
```
mbien, line 3269: I am wondering if it would make sense to sort that map in some way, e.g to try to index the primary language first to get a updated IDE state as fast as possible.

The primary language is probably the one with the largest list in that map? It might make no practical difference since the other lists might be small in comparison.

There is probably also an argument to be made for indexing the smallest list first to get incremental results fast.

matthiasblaesing: In the end parsing and indexing is expensive and we won't get rid of it. This is more a bugfix, that doubles as a performance improvement. I'll continue to stare at profiles of JS parsing, but outside this.

matthiasblaesing:
Lets get this in - if it causes trouble it is trivial to roll-back.

PR #4328, opened 2022-07-04
ci test reports via gh job summaries

mbien:
The new and more secure [job summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) feature might be our way out of #4116

- [x] [test-summary](https://github.com/test-summary/action) action likely not whitelisted and might require the submodule trick too (done, see [INFRA-23445](https://issues.apache.org/jira/browse/INFRA-23445))
- [x] will have to test how aggregation works (done, it aggregates automatically)

confirmed:
```
Error: .github#L1
test-summary/action@v1 is not allowed to be used in apache/netbeans. Actions in this workflow must be: within a repository owned by apache, created by GitHub, verified in the GitHub Marketplace, or matching the following: */*@[a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]+, AdoptOpenJDK/install-jdk@*, JamesIves/github-pages-deploy-action@5dc1d5a192aeb5ab5b7d5a77b7d36aea4a7f5c92, TobKed/label-when-approved-action@*, actions-cool/issues-helper@*, actions-rs/*, al-cheb/configure-pagefile-action@*, amannn/action-semantic-pull-request@*, apache/*, burrunan/gradle-cache-action@*, bytedeco/javacpp-presets/.github/actions/*, chromaui/action@*, codecov/codecov-action@*, conda-incubator/setup-miniconda@*, container-tools/kind-action@*, container-tools/microshift-action@*, dawidd6/action-download-artifact@*, delaguardo/setup-graalvm@*, docker://jekyll/jekyll:*, docker://pandoc/core:2.9, eps1lon/actions-label-merge-conflict@*, gaurav-nelson/github-action-markdown-link-check@*, golangc...
```
-> ~use via git submodule~ commit removed, infra added it to the list

result looks promising, example run:
https://github.com/apache/netbeans/actions/runs/2607773744
junit-report

other observations:
 - the summary doesn't list errors ("Forked Java VM exited abnormally"), only failures (filed: https://github.com/test-summary/action/issues/9)
 - commit validation tests are broken, but that is off topic, see dev list

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -111,7 +111,15 @@ jobs:
       - name: Commit Validation tests
         run: ant -Dcluster.config=release commit-validation
 
-      - name: Upload Build
+      - name: Create Test Summary
+        uses: test-summary/action@v1
+        if: failure()
+        with:
+          paths: "./nbbuild/build/test/commit-validation/results/*.xml"
+
+# the test-summary action above is currently only looking for failures and ignores errors,
+# this step can be removed as soon this is fixed upstream
+      - name: Upload Test Results
         if: failure()
         uses: actions/upload-artifact@v3
         with:
@@ -223,6 +231,12 @@ jobs:
 
       - name: Commit Validation tests
         run: ant -Dcluster.config=release commit-validation
+
+      - name: Create Test Summary
+        uses: test-summary/action@v1
+        if: failure()
+        with:
+          paths: "./*/*/build/test/*/results/*.xml"
 
   php:
     needs: base-build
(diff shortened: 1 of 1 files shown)


matthiasblaesing (commented):
You might consider retaining the upload action as we don't get info about error  if I understood your summary correctly?

neilcsmith-net (approved):

mbien:
> You might consider retaining the upload action as we don't get info about error if I understood your summary correctly?

added that again. I am only trying to find the JVM crash log somehow. Its not at the usual locations. Trying to set the path explicitly now. If its still doesn't work that might mean that the JVM doesn't generate the log some reason - but I have never seen this happening before.

neilcsmith-net:
> If its still doesn't work that might mean that the JVM doesn't generate the log some reason - but I have never seen this happening before.

I have, both never and intermittent. Quick search now suggests stack overflow in native might be one reason. Might need to consider what info you can get from OS?

mbien:
ok, setting the path didn't work either, so its likely that there is indeed no jvm log generated with that particular crash. Going to remove this again since it doesn't have anything to do with the original PR anyway and merge.

PR #4218, opened 2022-06-09
Disable flaky ValidateLayerConsistencyTest

matthiasblaesing:
See https://github.com/apache/netbeans/issues/4211 for details

PR #4333, opened 2022-07-05
Uprade to Apache Maven 3.8.6

ebarboni:
This is to upgrade embedded maven to 3.8.6 but overlay libs are incorrect


matthiasblaesing (approved):
Looks sane to me. A quick test with a freshly build NetBeans let me download dependencies, got artifact details, got sources and it build a project.

mbien:
> but overlay libs are incorrect

what does this mean?

Also I believe I linked the wrong issue. This is embedded, not the bundled maven, right?

matthiasblaesing:
We carry only one maven version, and this is the update - this is a screenshot from the build (see version output):

image

mbien (approved):
lets get this in :)

neilcsmith-net (approved):

mbien:
> Yes, `maven.embedder` is the module that bundles (or embeds) Maven. Somewhat confusing collision of terminology! smile

all good, just wanted to be sure that i linked the right issue :)

ebarboni:
unsure why commit-validation fail

mbien:
> unsure why commit-validation fail

the ValidateLayer tests are a bit broken at the moment #4211 #4218

mbien:
> so I merge :D

I don't know. Probably? Depends on what 

>but overlay libs are incorrect

means :)

ebarboni:
well in fact the overlay  was looking bad because ant clean were not removing the jars. on a fresh clone seems ok.

mbien:
mystery solved. lets merge :D

PR #4342, opened 2022-07-07
Update to codesnippetdoclet 0.82 (supports JDK-11) and using JEP-413 @snippet tag

jtulach:
There are broken links in NetBeans Javadoc as of 12.6 version (since we started to build with JDK-11). For example the [modules package page](https://bits.netbeans.org/14/javadoc/org-openide-modules/org/openide/modules/package-summary.html) references [Modules API](https://bits.netbeans.org/14/javadoc/org-openide-modules/org/openide/modules/doc-files/api.html) and that one is missing. To reproduce:
```
netbeans/platform/openide.modules$ JAVA_HOME=/jdk-8 ant clean build javadoc
```
generates the link properly. When JDK-11 is used to build the Javadoc, then the link is broken.

This PR updates to [codesnippet4javadoc 0.82](https://github.com/jtulach/codesnippet4javadoc) which is properly tested on newer JDKs (up to JDK-18) and generates all the pages properly on all JDKs.

Additionally the switch to version `0.82` allows us to use [JEP-413](https://openjdk.org/jeps/413) javadoc tags. Done for platform and ide clusters in: [669e8f2](https://github.com/apache/netbeans/pull/4342/commits/669e8f2027b97d522313343ed20aa8d17ec2923e) and in [66ace01](https://github.com/apache/netbeans/pull/4342/commits/66ace0141380a6f118b8ece5dc401a803c935081) - for the rest. That allows us to disable the original syntax and allow only [JEP-413 compliant one](https://github.com/apache/netbeans/pull/4342/commits/043b6ea5705264f30385d34f830917a33382d5b4)


mbien:
we have a CI job which builds javadoc. I am wondering if this serves any purpose since it wouldn't really detect bugs like this anyway. Should I remove it next time I look at CI jobs to conserve some resources (thats ~30mins x2 per PR)?

I bet jenkins builds javadoc too for the dev builds.

JaroslavTulach:
> we have a CI job which builds javadoc. I am wondering if this serves any purpose since it wouldn't really detect bugs like this anyway. 

Running `ant build-javadoc` generates a file `nbbuild/build/javadoc/checklinks-errors.xml` - if it contains failed JUnit test, the build could fail. There used to be a time, when there were no broken links in the Javadoc - not sure how much work would be to get there again.

singh-akhilesh (approved):
Awesome, LGTM.

extide/gradle/src/org/netbeans/modules/gradle/execute/AbstractGradleExecutor.java
```diff
@@ -35,7 +35,6 @@
 import org.openide.util.NbBundle;
 import org.openide.util.RequestProcessor;
 
-import static org.netbeans.modules.gradle.execute.Bundle.*;
```
JaroslavTulach, line 38: Javadoc on JDK-11 seems to have problems with `Bundle.*` imports - then it generates no HTML pages. I had to eliminate the import and use `Bundle.xyz` directly. Btw. the same problem appears in `maven` project, but there is too many mistakes for me to fix - could we have a transformation to avoid using `import static`, please @jlahoda, @dbalek?

lkishalmi (approved):
Nice!

extide/gradle/src/org/netbeans/modules/gradle/execute/AbstractGradleExecutor.java
mbien: jackpot rule language can't trigger on imports as far as I know. This would have to be written by hand I believe.

PR #4349, opened 2022-07-08
Improve alignment of text and "X" button in FlatLAF window system tabs

eirikbakke:
This PR makes the window system tab labels align with the icon and the "X" button, with any font and at any font size.

The font-agnostic calculation is useful because FlatLAF uses a different UI font on each OS. Accurate vertical alignment also allows more compact tab styles if desired in the future.

### Before/After screenshots (at 1x scaling, taken while hovering the mouse over the X icon in the last tab):
AlignBefore
AlignAfter

### After patch, with "--fontsize 20" set in netbeans.conf:
AlignFontSize

### After patch, at 150% HiDPI scaling:
Align150

Details:
* API change: Allow AbstractTabCellRenderer subclasses to override text position calculation.
* In FlatLAF, improve the calculation of tab labels' Y position, to avoid the need for ad hoc adjustments and to work with a variety of font families and font sizes. Ensure that text in view tabs and editor tabs still line up.
* In FlatLAF, ensure a minimum tab height for smaller fonts, based on the expected icon size, regardless of inset settings. This will allow the same inset settings to work for e.g. Tahoma 11 and Segoe UI 12 (the old and new default fonts on Windows), and other fonts which may be used by default for FlatLAF on other operating systems.
* In FlatEditorTabDisplayerUI.getPreferredSize, use the correct graphics context for font metrics calculations.
* Also avoid unnecessary rounding error in the vertical centering of tab icons and buttons, by doing "(a-b)/2" instead of "a/2-b/2". This is done for all LAFs; the difference is at most one pixel compared to the old behavior.

(This PR incorporates parts 4 and 5 from the now superseded https://github.com/apache/netbeans/pull/3115 .)

### DIFF
--- a/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
+++ b/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
@@ -71,7 +71,7 @@
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
-                        <specification-version>1.63</specification-version>
+                        <specification-version>1.74</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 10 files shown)


neilcsmith-net (approved):
Looks good!

~~From looking at your screenshots and then testing, I'm convinced the hovered close button looks closer to the label with dark vs light (with latest master), but didn't notice any reason for that. Maybe optical illusion.~~ Checked, and just difference in anti-aliasing colours.

DevCharly (approved):

eirikbakke:
Thanks for getting these FlatLAF patches merged everyone!

PR #4353, opened 2022-07-08
Improvements to JDK download and registration UI.

neilcsmith-net:
Building on @mbien changes in #4205 -

- Moved search for matching package into the quick panel itself - now we have distribution, it's better to validate here.
- Distribution and version sync when moving between quick and advanced panels (where possible).
- Made some improvements to keyboard and focus handling.
- Removed installer option from quick panel - leave that for advanced as they cannot be registered anyway (yet).
- Various minor fixes.

On top of that, there's a commit that stops value painting on JSliders in GTK, which causes spurious values to be painted above the Disco quick panel slider. There seems no way to control per-component, but we don't really have other uses of JSlider and the utility of this is questionable.

Added a file attribute to store the display name on the folder when registering a J2SE platform.  This means that the previous name is offered by default when re-registering a platform.

Commits split per module.  I've kept back some other changes mentioned in linked PR as we're close to freeze.


java/java.disco/src/org/netbeans/modules/java/disco/AdvancedPanel.java
```diff
@@ -378,12 +395,12 @@ private void eaCheckBoxActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FI
...
-                    (Architecture) architectureComboBox.getSelectedItem(),
-                    latestCheckBox.isSelected() ? Latest.OVERALL : Latest.ALL_OF_VERSION,
-                    (PackageType) packageTypeComboBox.getSelectedItem(),
-                    eaCheckBox.isSelected());
+        updateData((Distribution) distributionComboBox.getSelectedItem(),
+                (Integer) versionComboBox.getSelectedItem(),
+                (Architecture) architectureComboBox.getSelectedItem(),
+                latestCheckBox.isSelected() ? Latest.OVERALL : Latest.ALL_OF_VERSION,
+                (PackageType) packageTypeComboBox.getSelectedItem(),
+                eaCheckBox.isSelected());
```
mbien, line 403: since there are only whitespace changes here, I was wondering if you wanted to call the utility getters you made from here but you forgot?

mbien (approved):
good changes. Package label has more utility than the radio boxes on a quick selection panel.

java/java.disco/src/org/netbeans/modules/java/disco/AdvancedPanel.java
neilcsmith-net: No, just being guilty of my own pet hate. Must have been a little over zealous on what regions I selected to auto-format.

neilcsmith-net:
Thanks @mbien

PR #4355, opened 2022-07-09
Analyze Stack Trace window improvements.

mbien:
- added indentation for more familiar stack trace formatting
 - fixed hyperlink text color when selected
 - fixed TopComponent background color to be consistent with other  windows

indent_traces

The top component was a little bit too bright with dark theme (looked like the old color of the NB 14 dark theme and didn't fit to the rest)
Projects tab is: `#3C3F41`, Analyze Stack window was: `#46494B`, so i set it to `UIManager.getColor("Tree.background")` which is the same as the other top components (e.g projects tab).


eirikbakke (approved):
Thanks for this, looks good!

PR #4356, opened 2022-07-09
maven project wizard should set groupId #4343.

### LINKED ISSUE #4343: New Maven Project wizard leaves groupId empty in pom.xml when clearing package path entry (marked Optional)
### Apache NetBeans version

Apache NetBeans 14

### What happened

Creating a new Maven Java application, if one clears out the optional package path, NetBeans generates a project where the only reference to groupId is an empty entity "&lt;groupId/&gt;".
The project is not loadable or buildable, and, in fact, NetBeans sometimes attempts to identify it as a multi-module project.
If one leaves the package information "unmolested," the created project works fine.

### How to reproduce

Select New Project
Select Java with Maven / Java Application
Click Next
Fill in all the data correctly, but delete the package path in the Package field (marked optional).
Click Finish
NetBeans generates a project where the only reference to groupId is an empty entity "&lt;groupId/&gt;"

Apache NetBeans 12.3 or earlier

### Operating System

Windows 11 Pro (Build 22000.778)

### JDK

JDK 17.0.2 (Oracle) x86_64 2022-01-18

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

I was using Maven 3.6.3 and the bundled. I know this used to work correctly, but I don't recall the last version I used where it wasn't a problem. This might happen under NB 13, but I didn't use it long enough to realize it was a repeatable problem.

No

### Code of Conduct

Yes

mbien:
fixes #4343

 - the groupId text field was ignored so far
 - this fixes a bug which breaks the project if no package is provided
   (since the package is used as fallback to generate a groupId which hid the issue)

(artifactId is only set for completes sake)

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/newproject/idenative/IDENativeMavenWizardIterator.java
+++ b/java/maven/src/org/netbeans/modules/maven/newproject/idenative/IDENativeMavenWizardIterator.java
@@ -94,6 +94,8 @@ public abstract class IDENativeMavenWizardIterator implements WizardDescriptor.I
         return new FileBuilder(w.getTemplate().getPrimaryFile(), w.getTargetFolder().getPrimaryFile().getParent()).
             param(TemplateUtils.PARAM_PACKAGE, (String) wiz.getProperty("package")).
             param(TemplateUtils.PARAM_PACKAGING, (String) this.packaging).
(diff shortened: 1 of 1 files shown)


matthiasblaesing (approved):
Looks sane and tested successfully.

PR #4357, opened 2022-07-09
nb project wizard fixes

### LINKED ISSUE #4316: Old nbm-maven-plugin:4.3 is generated for new project pom
### Apache NetBeans version

Apache NetBeans 14

### What happened

image

image
old dependency from 2019

### How to reproduce

Generate new nbp project

No

### Operating System

Windows 10 21H1

### JDK

17.0.1; Java HotSpot(TM) 64-Bit Server VM 17.0.1+12-LTS-39

### Apache NetBeans packaging

### Anything else

_No response_

No

### Code of Conduct

Yes

mbien:
- don't add osgi flag if it was not checked (followup on #4007)
 - upgrade nbm-plugin version in nb app project wizard (fixes #4316)
   (nbm project did already upgrade it correctly)
 - minor cleanup

reviewer note: check ignore whitespace since there was a large section on the wrong indentation level

### DIFF
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NbmWizardIterator.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NbmWizardIterator.java
@@ -53,6 +53,8 @@ import static org.netbeans.modules.maven.apisupport.MavenNbModuleImpl.APACHE_SNA
 import org.netbeans.modules.maven.embedder.EmbedderFactory;
 import org.netbeans.modules.maven.indexer.api.NBVersionInfo;
 import org.netbeans.modules.maven.indexer.api.RepositoryQueries;
+import org.netbeans.modules.maven.model.pom.Plugin;
+import org.netbeans.modules.maven.model.pom.PluginManagement;
 import org.netbeans.spi.project.ui.support.CommonProjectActions;
 
 public class NbmWizardIterator implements WizardDescriptor.BackgroundInstantiatingIterator<WizardDescriptor> {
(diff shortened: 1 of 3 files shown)


matthiasblaesing:
According to the documentation of the nbm-maven plugin this comment is wrong:

https://github.com/apache/netbeans/blob/0b2627f571c030566472b40309c3f7ff409c8db1/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/NbmWizardPanelVisual.java#L274

Looking at the history I read it, that the option should have been disabled for versions before NetBeans 6.9

https://github.com/apache/netbeans-mavenutils-nbm-maven-plugin/blob/153cbd6dbbee835833fbffaa349054c2314e8249/nbm-maven-plugin/src/site/apt/index.apt.vm#L126-L145

and thus we should default to enable OSGI support.

This would also make more sense as the NetBeans support for OSGI improved and did not regress.

mbien:
> and thus we should default to enable OSGI support.
> 
> This would also make more sense as the NetBeans support for OSGI improved and did not regress.

have you tested if it works? If it does, sure lets turn the checkbox visible. Its just that NB hooks into a [patched](https://github.com/apache/netbeans/blob/master/platform/netbinox/external/rebuild-equinox.sh) version of equinox OSGI 3.9.1 (Kepler, 2013) which technically didn't have support for JDK 8 yet. If I remember correctly, Luna (2014, the version after Kepler) was the first which supported JDK 8 - and a lot happened since JDK 8. [Equinox version archive](https://archive.eclipse.org/equinox/).

matthiasblaesing:
> > and thus we should default to enable OSGI support.
> > This would also make more sense as the NetBeans support for OSGI improved and did not regress.
> 
> have you tested if it works?

No - but I doubt this should be the deciding factor here. OSGI dependencies are a reality we can't ignore and if people have problems with. The IDE is an example that OSGI dependencies work.

matthiasblaesing (approved):
Looks sane to me.

PR #4358, opened 2022-07-10
Enable inline hints by default

jtulach:
Whenever I see people using other IDEs, they have _inline hints_ enabled and find it quite useful. Thanks to @jlahoda work NetBeans also supports _inline hins_ and I am using them without any major problems for last few releases. However the feature is turned off by default and I am afraid many users of NetBeans IDE do not even know it exists. This PR proposes to turn the feature on by default.

image

### DIFF
--- a/ide/editor.actions/src/org/netbeans/modules/editor/actions/ShowInlineHintsAction.java
+++ b/ide/editor.actions/src/org/netbeans/modules/editor/actions/ShowInlineHintsAction.java
@@ -29,7 +29,7 @@ import org.netbeans.api.editor.EditorActionRegistration;
                           preferencesDefault=ShowInlineHintsAction.DEF_LINES)
 public class ShowInlineHintsAction extends AbstractAction {
     public static final String KEY_LINES = "enable.inline.hints";
-    public static final boolean DEF_LINES = false;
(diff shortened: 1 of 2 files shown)


mbien (commented):
I don't use it personally so that wouldn't be a +1 from me, but I don't have to agree - its just my preference (its also easy to disable, so no big deal). If i would want to know the parameter details I would either press ctrl+p, hit ctrl+space at the method name to get the doc too, or jump to the code with ctrl+click directly.

I personally don't like it when IDEs display things inline which are not in the text. This creates weird habits, e.g overuse of the var keyword since some IDEs display the type next to it anyway. And the result: the classic type decl would have used up less space than var + type. Thats essentially one step forward and two back :)

However, I do have a small list of other potential "good defaults" written down which I wanted to bring up on the dev list. to figure out what setting we could update as "better default" and what to leave to the user to adjust. Maybe we should do this with this feature too to gather more opinions.

geertjanw:
Agree that enabling this would mean people actually use it, otherwise they'll never find it to enable it, and it should just be a standard part of what the Java editor provides, just like anything else that is on by default.

sdemirhan2478:
Something happened until I found how to turn off the feature, you should have written how to turn it off.

PR #4361, opened 2022-07-11
Removed unused Smack libraries

lkishalmi:
Just some cleanup.


neilcsmith-net (approved):
Fine with me.  But not a friend-only library I believe?  Will ask @jtulach to cast his eye over.

PR #4363, opened 2022-07-11
Fix missing module name in Javadoc URLs for Show Javadoc, etc.

neilcsmith-net:
Attempt to fix errors in Javadoc links where the module name is missing from the URL.

Not sure if we need to try a reverse fallback without the module name on exception?


mbien:
code looks good. I think i found another issue while testing. The javadoc popup has this little "show in external browser" button. Well, it opens without the module name which leads to 404 with the current oracle hosted doc.

neilcsmith-net:
@mbien that should be fixed by this too - it's the one I come across more often. Seems fixed here for most projects except NetBeans sources. What's the nature of code you're testing with?

neilcsmith-net:
Thanks @mbien 

@jlahoda could you run a quick pair of eyes over this?  I'm also curious why there seems to be no enclosing module for JDK classes with NetBeans sources - is this a wrong assumption?  Thanks!

Issue #4231, opened 2022-06-16
Netbeans 14 freezes spuriously when r-clicking at Design View


rapastranac:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

When editing forms, when I right-click on the form and the menu is expanded, it suddenly freezes at any of the menu items

### How to reproduce

When I right-click on the form and the menu is expanded, it suddenly freezes at any of the menu items, but it does not happens consistently. Maybe a multi-thread issue.
 

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Windows 10

### JDK

13

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

mbien:
i believe I saw that too. It happens more often if the context menu is cascaded.

Might have potentially something to do with the FlatLAF update since it added some features specific to context menus (#3983). Have to check if there is a flag for that to toggle it off for testing purposes. (or test with a different look and feel enabled)

rapastranac:
For some reason I could not uploaded the thread dump when filing this issue.

Here it is

[threaddump-1655395630439.txt](https://github.com/apache/netbeans/files/9013393/threaddump-1655395630439.txt)

I captured it with visualvm, whose original extension is tdump. I edited the extension so I could upload it here

wumpz:
Same here since Netbeans 14 maybe two times.

eirikbakke:
Hmm, this problem would be very good to fix. Could either of you try to see if you could manage to reliably reproduce the bug? And perhaps take a screenshot of NetBeans in the frozen state?

It might be good to try another thread dump as well.

cts93:
I'm using Apache NetBeans IDE 14 with this: https://draculatheme.com/netbeans

I had the same issue happen to me once I started creating and editing JFrames. Same thing: change something upon context menu (particularly the Auto Resize option) and it just freezes up everything. You have to kill NetBeans through Task Manager and open again.

I just now changed NetBeans Look and Feel theme from FlatLaf Dark to something else, checked the option "Maximize use of native look and feel". It removed that front color theme I downloaded from Dracula so I added it again and changed Look and Feel back to FlatLaf Dark. I still didn't get any freeze ever since; It seemed to have solved the problem for me.

So what we know from this is it's clearly something to do with themes, perhaps compatibility issues with custom themes.

DevCharly:
[...]

PR #4375, opened 2022-07-14
Update FlatLaf from 2.3 to 2.4

DevCharly:
Update FlatLaf to v2.4.

Changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/2.4

This version fixes issue #4231 (Netbeans 14 freezes spuriously when r-clicking at Design View)

It brings also some usability improvements to the window title bar (on Windows 10/11):
- There are now additional areas at top of embedded menu bar to resize the window. This makes it easier to resize small frames with lot of menus. (see red rectangles in below screenshot)
- The window title now has a minimum width, which avoids that it is completely hidden. The window title is important because it is needed to move the window. (green rectangle)

image

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-D7A77B74A49FC9705199426FFA73BD67ACF06306 com.formdev:flatlaf:2.3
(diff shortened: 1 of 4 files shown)


mbien (approved):
awesome, the context menu fix is an important one - otherwise we would have to turn off the submenu tweaks to avoid freezes. Thanks for fixing this upstream on such short notice.

eirikbakke (approved):

PR #4383, opened 2022-07-15
Let iterate() initialize the indexer first.

sdedic:
The following exception appears in the log when one tries to `Attach Sources` and uses `Download`:

```
java.lang.NullPointerExceptionjava.lang.NullPointerException 
at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.constructQuery(NexusRepositoryIndexerImpl.java:1154)
at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.findBySHA1(NexusRepositoryIndexerImpl.java:1279)
at org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl.findBySHA1(NexusRepositoryIndexerImpl.java:1272)
at org.netbeans.modules.maven.indexer.api.RepositoryQueries.lambda$findBySHA1$6(RepositoryQueries.java:357)
at org.netbeans.modules.maven.indexer.api.RepositoryQueries.getQueryResult(RepositoryQueries.java:217)
at org.netbeans.modules.maven.indexer.api.RepositoryQueries.findBySHA1(RepositoryQueries.java:354)
at org.netbeans.modules.maven.indexer.api.RepositoryQueries.findBySHA1Result(RepositoryQueries.java:346)
at org.netbeans.modules.maven.queries.MavenSourceJavadocAttacher.attach(MavenSourceJavadocAttacher.java:87)
[catch] 
at org.netbeans.modules.maven.queries.MavenSourceJavadocAttacher.getSources(MavenSourceJavadocAttacher.java:204)
at org.netbeans.modules.java.j2seplatform.queries.SourceJavadocAttacherUtil$Downloader.run(SourceJavadocAttacherUtil.java:140)
```

The reason si that in 656e520  the `constructQuery` was [moved outside of the `iterate()` invocation](https://github.com/apache/netbeans/commit/656e5209812854905d7c2f164cf4f648430a2797#diff-388a78e69b7729d3486a578826691d1606ea56f099d2f1c0eb7c484deba12948R1376) which lazy-initializes the whole class (instantiates maven services).

### DIFF
--- a/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
+++ b/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
@@ -1275,10 +1275,10 @@ public class NexusRepositoryIndexerImpl implements RepositoryIndexerImplementati
     private ResultImplementation<NBVersionInfo> findBySHA1(final String sha1, final ResultImpl<NBVersionInfo> result, List<RepositoryInfo> repos, final boolean skipUnIndexed) {
         final List<NBVersionInfo> infos = new ArrayList<>(result.getResults());
         final SkippedAction skipAction = new SkippedAction(result);
-        BooleanQuery bq = new BooleanQuery.Builder()
-                .add(new BooleanClause((setBooleanRewrite(constructQuery(MAVEN.SHA1, sha1))), BooleanClause.Occur.SHOULD))
-                .build();
         iterate(repos, (RepositoryInfo repo, IndexingContext context) -> {
+            BooleanQuery bq = new BooleanQuery.Builder()
+                    .add(new BooleanClause((setBooleanRewrite(constructQuery(MAVEN.SHA1, sha1))), BooleanClause.Occur.SHOULD))
+                    .build();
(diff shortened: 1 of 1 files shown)


mbien (approved):
sorry about that. So `indexer` was still null before the lambda loop is run?

I checked if there are other places which are affected by this pattern, but i believe it is only this one occurrence. Since queries are basically immutable constants it would be cleaner to create one shared object outside the loop - I didn't consider that indexer could be still null at that point in time.

Do you think it would be better to keep it as is but just call `initIndexer()` before `iterate()`?

edit: but this would probably require a critical section

sdedic:
> sorry about that. So `indexer` was still null before the lambda loop is run?

Yes, the findBySHA were the only queries / operations that were executed during my IDE run, it seems.

> Do you think it would be better to keep it as is but just call `initIndexer()` before `iterate()`?
> 

I'd keep it initialized from within `iterate()` actually the init is done from loadIndexingContext2 called from within mutex. But iterate() skips it if no repos are configured or indexing is disabled for the repo. I am not sure if the laziness is that beneficial, but moving it back to iterate seemed as the simplest fix.

neilcsmith-net:
I assume we want this in before I branch so merging. Please follow up in delivery if there's a need.

Issue #4366, opened 2022-07-12
'call hierarchy' from base class does not work anymore


sijskes:
### Apache NetBeans version

Apache NetBeans 14

### What happened

'call hierarchy' 'search from base class' option does not function anymore, when clicking 'call hierarchy' on a method of a implementation of an interface. 

### How to reproduce

clicking 'call hierarchy' on a method of a implementation of an interface. 

### Did this work correctly in an earlier version?

Apache NetBeans 12.6 or earlier

### Operating System

5.4.0-110-generic #124-Ubuntu SMP Thu Apr 14 19:46:19 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### JDK

multiple

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

    /**
     * right-click 'call hierarchy' should show calls to method1 from base class.
     * (when this options is enabled in the 'Java Call Hierachy' window, filter icon menu.)
     *
     * works on:
     * Apache NetBeans IDE 11.0 (Build incubator-netbeans-release-404-on-20190319)
     * Java: 11.0.15; OpenJDK 64-Bit Server VM 11.0.15+10-Ubuntu-0ubuntu0.20.04.1
     *
     * works on:
     * Product Version: Apache NetBeans IDE 12.0
     * Java: 11.0.15; OpenJDK 64-Bit Server VM 11.0.15+10-Ubuntu-0ubuntu0.20.04.1
     *
     * does not work on:
     * Product Version: Apache NetBeans IDE 13
     * Java: 17.0.3; OpenJDK 64-Bit Server VM 17.0.3+7-Ubuntu-0ubuntu0.20.04.1
     *
     * does not work on:
     * Product Version: Apache NetBeans IDE 13
     * Java: 11.0.15; OpenJDK 64-Bit Server VM 11.0.15+10-Ubuntu-0ubuntu0.20.04.1
     *
     * does not work on:
     * Product Version: Apache NetBeans IDE 14
     * Java: 17.0.3; OpenJDK 64-Bit Server VM 17.0.3+7-Ubuntu-0ubuntu0.20.04.1
     */

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

sijskes:
other users report from dev list: 

I also have this problem on Mac M1 with self-built NB14.

sijskes:
[callbug.zip](https://github.com/apache/netbeans/files/9092026/callbug.zip)

[...]

PR #4384, opened 2022-07-15
#4366: Handle references to other compilation units.

sdedic:
For description, see #4366 

It **was** actually a one-liner ... The problem was that Trees.getPath() won't work outside of compilation units processed by the parser - but TreePathHandle.create/from() can process any element, so I've used that as a fallback.


mbien (approved):
very efficient bugfix :) Tested it and it works as it used to.

at some point we should add a junit test for this though (maybe taking FindUsagesTest as template, however even this test is disabled if I see this correctly since it sets up a Noop suite).

neilcsmith-net:
I assume we want this in before I branch so merging.  Please follow up in delivery if there's a need.

PR #4386, opened 2022-07-15
Improved YAML curly, bracket and quotes keystroke handling

lkishalmi:
Further improve the work done in #4093. Made it more consistent. As both '{}' and '[]' are language elements Pressing ```{``` or ```[``` would create the closing pair as well.

As quotes are created (usually) in pairs as well, pressing quote for a closing quote would step over the existing quote, instead of inserting another pair. So those who usually type two quotes fast, won't get annoyed.

### DIFF
--- a/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlKeystrokeHandler.java
+++ b/ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlKeystrokeHandler.java
@@ -53,6 +53,27 @@ public class YamlKeystrokeHandler implements KeystrokeHandler {
         BaseDocument doc = (BaseDocument) document;
 
         int dotPos = caret.getDot();
+        int length = doc.getLength();
+
+        // Primitive handling of backslash as escape character, far from accurate
+        // but would work most of the time.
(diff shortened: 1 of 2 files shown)


ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlKeystrokeHandler.java
```diff
@@ -101,16 +129,26 @@ public boolean beforeCharInserted(Document document, int caretOffset, JTextCompo
...
+            for (int i = 0; i < line.length; i++) {
+                char d = line[i];
+                if ('\\' == d) {
+                    i++;
+                    continue;
+                }
                 if (c == d) quotes++;
             }
+            for (char d : line) {
+            }
```
mbien, line 141: can be probably removed :D

mbien (approved):
code looks good. Tested it on some files and it worked too.

ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlKeystrokeHandler.java
lkishalmi: Oh, yes! Good catch Thanks!

PR #4387, opened 2022-07-16
Fix NPE when there is no build.gradle for the root project

lkishalmi:
It's getting somewhat usual to have project without root build.gradle (or build.gradle.kts) script, only a settings.gradle in the root.
That setup is completely valid, though it could cause NPE-s. This trivial patch fixes that.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/GradleReport.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/GradleReport.java
@@ -51,7 +51,7 @@ public final class GradleReport {
     }
     
     public static GradleReport simple(Path script, String message) {
-        return new GradleReport(null, script.toString(), -1, message, null);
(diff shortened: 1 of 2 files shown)


sdedic (approved):
Thanks, I will try to write code more robust.

extide/gradle/src/org/netbeans/modules/gradle/GradleReport.java
```diff
@@ -51,7 +51,7 @@ public GradleReport(String errorClass, String location, int line, String message
     }
     
     public static GradleReport simple(Path script, String message) {
-        return new GradleReport(null, script.toString(), -1, message, null);
+        return new GradleReport(null, script != null ? script.toString() : null, -1, message, null);
```
mbien, line 54: nitpick: there is a `Objects.toString(obj)` util for exactly this usecase ;)

Issue #4337, opened 2022-07-06
Support the conventional gradle/libs.versions.toml as a "Build Scripts" file for Gradle-based projects


swpalmer:
### Description

To assist with managing dependency information for libraries and build plugs, Gradle has the concept of a "Version Catalogue"
See: https://docs.gradle.org/current/userguide/platforms.html

One way to make the catalog easier to define is to use a file in the TOML format.  The standard location for this TOML file is in the "gradle" subdirectory of the project, using the default name "libs.versions.toml" 
See: https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml

The first step of this feature is simply to include this file under the project's "Build Scripts" node.

A further enhancement would be for syntax highlighting and auto-complete features to work in the catalog file. TOML syntax highlighting is an independent but related feature.

### Use case/motivation

Having this feature would alleviate the need to flip to the "Files" view to get at this configuration file for Gradle projects.

### Related issues

I didn't see a separate issue yet, but basic support for editing TOML files, with syntax highlighting, would improve the user experience greatly.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow the Apache Software Foundation's [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html)

PR #4388, opened 2022-07-16
Add minimal support for Gradle default version catalog

lkishalmi:
Implement minimal support for #4337. The libs.versions.toml file will be displayed under the build scripts node.


sdedic (approved):
+ version bump in manifest

PR #4389, opened 2022-07-16
Let ClassSource check for null elements in constructor.

mbien:
addToProjectClassPath() would fail later with a less useful stacktrace as observed in #4365.

(this doesn't really fix a bug, defensive checks make finding bugs a bit easier)

### DIFF
--- a/java/form/src/org/netbeans/modules/form/project/ClassSource.java
+++ b/java/form/src/org/netbeans/modules/form/project/ClassSource.java
@@ -26,7 +26,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import org.openide.filesystems.FileObject;
-import org.openide.filesystems.FileUtil;
 import org.openide.util.Lookup;
 
 /**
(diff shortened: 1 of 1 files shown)


neilcsmith-net (approved):
Sounds good, although wonder if it should still be a `NullPointerException`?

In terms of #4365 then a bigger thing is probably validation of `jdk.home` at https://github.com/apache/netbeans/blob/master/java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java#L56  Or just relying on `java.home`.

mbien:
> Sounds good, although wonder if it should still be a `NullPointerException`?

I think either is ok. Since the list itself is not null only the entry I opted for IAE here.

> 
> In terms of #4365 then a bigger thing is probably validation of `jdk.home` at https://github.com/apache/netbeans/blob/master/java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java#L56 Or just relying on `java.home`.

is already in progress, will be a different PR.

neilcsmith-net:
> I think either is ok. Since the list itself is not null only the entry I opted for IAE here.

OK, makes sense.

> is already in progress, will be a different PR.

I'll stop working on it and wait for yours then! :laughing:

PR #4390, opened 2022-07-16
use java.home as "default jdk" if jdk.home leads to symlinks.

### LINKED ISSUE #4365: java.lang.NullPointerException when creating new JFrame Form
### Apache NetBeans version

Apache NetBeans 14

### What happened

Create a new Java project then a new JFrame Form. Expected to see elements on palette but only Beans is shown and an exception appears in notification panel.
Same happens every time I open that JFram Form

### How to reproduce

Create a new JFrame Form

No

### Operating System

Ubuntu 22.04 LTS

### JDK

openjdk version "11.0.15" 2022-04-19 OpenJDK Runtime Environment (build 11.0.15+10-Ubuntu-0ubuntu0.22.04.1) OpenJDK 64-Bit Server VM (build 11.0.15+10-Ubuntu-0ubuntu0.22.04.1, mixed mode, sharing)

### Apache NetBeans packaging

### Anything else

```
	at org.netbeans.modules.form.project.ClassSource.addToProjectClassPath(ClassSource.java:89)
	at org.netbeans.modules.form.project.ClassPathUtils$1.run(ClassPathUtils.java:233)
[catch] at org.openide.util.RequestProcessor$Processor.run(RequestProcessor.java:2033)
```

image
image

Something similar happens when I try with Netbeans 12

No

### Code of Conduct

Yes

mbien:
- /usr shouldn't be a valid path for the default JDK
   - use java.home if jdk.home/bin/java is a symlink
   - updated some related warning messages
   - fixes #4365

NB can't actually be started on a "JRE" (whatever a JRE is these days since it isn't well defined anymore). However, the "default jdk" path could be invalid still which will cause trouble at runtime. (looking at you, installer)

e.g trying to run NB on a "JRE" fails early so we don't have to worry about this scenario (at least not for the IDE):
```
./netbeans 
Error occurred during initialization of boot layer
java.lang.module.FindException: Module jdk.jshell not found
```

### DIFF
--- a/java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
+++ b/java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
@@ -23,13 +23,13 @@ import java.io.*;
 import java.net.URL;
 import java.util.*;
 import java.net.MalformedURLException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import org.openide.util.Exceptions;
-
 import org.openide.util.NbBundle;
 import org.openide.util.Utilities;
 import org.openide.filesystems.FileUtil;
 import org.netbeans.api.java.platform.*;
-import org.netbeans.api.java.classpath.*;
 
 /**
  * Implementation of the "Default" platform. The information here is extracted
@@ -42,8 +42,6 @@ public class DefaultPlatformImpl extends J2SEPlatformImpl {
 
     public static final String DEFAULT_PLATFORM_ANT_NAME = "default_platform";           //NOI18N
 
-    private ClassPath standardLibs;
-
     @SuppressWarnings("unchecked")  //Properties cast to Map<String,String>
     static JavaPlatform create(Map<String,String> properties, List<URL> sources, List<URL> javadoc) {
         if (properties == null) {
@@ -53,9 +51,9 @@ public class DefaultPlatformImpl extends J2SEPlatformImpl {
         synchronized (p) {
             p = new HashMap<>(p);
         }
-        String  jdkHome = System.getProperty("jdk.home"); // NOI18N
+        String jdkHome = System.getProperty("jdk.home"); // NOI18N
         File javaHome;
-        if (jdkHome == null) {
+        if (jdkHome == null || Files.isSymbolicLink(Paths.get(jdkHome, "bin", "java"))) {
             if (Util.getSpecificationVersion((String) p.get("java.specification.version")).compareTo(Util.JDK9) < 0) {
                 javaHome = FileUtil.normalizeFile(new File(System.getProperty("java.home")).getParentFile()); // NOI18N
             } else {
@@ -64,6 +62,7 @@ public class DefaultPlatformImpl extends J2SEPlatformImpl {
         } else {
             javaHome = FileUtil.normalizeFile(new File(jdkHome));
         }
+
         List<URL> installFolders = new ArrayList<> ();
         try {
             installFolders.add (Utilities.toURI(javaHome).toURL());
--- a/java/api.debugger.jpda/src/org/netbeans/api/debugger/jpda/Bundle.properties
+++ b/java/api.debugger.jpda/src/org/netbeans/api/debugger/jpda/Bundle.properties
@@ -22,9 +22,7 @@ OpenIDE-Module-Short-Description=JPDA Debugger API
 OpenIDE-Module-Long-Description=JPDA Debugger API
 OpenIDE-Module-Package-Dependency-Message=\
     <b>The JDK is missing and is required to run some NetBeans modules</b><br> \
-    Please use the --jdkhome command line option to specify a JDK<br>\
-    installation or see <a href="http://wiki.netbeans.org/FaqRunningOnJre">http://wiki.netbeans.org/FaqRunningOnJre</a> for<br> \
-    more information.
+    Please use the --jdkhome command line option or set netbeans_jdkhome in your netbeans.conf to specify a JDK.
 MSG_NO_DEBUGGER=No debugger engine available.
 
 ClassBrkp_Type=Class Load/Unload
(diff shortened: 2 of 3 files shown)


java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
```diff
@@ -42,8 +48,15 @@ public class DefaultPlatformImpl extends J2SEPlatformImpl {
...
 
-    private ClassPath standardLibs;
+    @StaticResource
+    private static final String WARNING_ICON = "org/netbeans/modules/java/j2seplatform/resources/warning.png"; //NOI18N
 
+    @NbBundle.Messages({
+        "TITLE_InvalidDefaultJDK=Invalid JDK Home detected",
+        "DESC_InvalidDefaultJDK=The jdk.home property does not set a path to a valid JDK." +
+                               "Please update netbeans_jdkhome in your netbeans.conf" +
+                               "or start netbeans with --jdkhome /path/to/JDK set."
```
mbien, line 58: how comes that `NotificationDisplayer` doesn't support html or `\n` line breaks? This creates a fairly long line.

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
```diff
@@ -84,6 +102,12 @@ static JavaPlatform create(Map<String,String> properties, List<URL> sources, Lis
         return new DefaultPlatformImpl(installFolders, properties, systemProperties, sources, javadoc);
     }
 
+    private static boolean isValidJDKHome(Path javaHome) {
+        return Files.exists(javaHome.resolve(Path.of("bin", "java")))
+            && Files.exists(javaHome.resolve(Path.of("bin", "javac")));
+//            && Files.exists(javaHome.resolve(Path.of("jmods", "java.base.mod")));   // TODO: uncomment once everything migrated to 11+
```
neilcsmith-net, line 108: Did you test this? Because that's probably still returning `true`.  The IDE wouldn't launch otherwise.  The problem is that those files are symlinks.  Are there any circumstances that `java.home` will not be correct?  I would either default to that, or at least default to that if `<jdk.home>/bin/java` is a symlink.

neilcsmith-net (changes requested):

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
neilcsmith-net: I assume you tested the message starting with `<html>`?

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
mbien: I tested this with the jmods line uncommented, but then I realized that this is probably used by VSCode too.

So you are saying the `jdk.home` property has no use at this point in the lifecycle anymore, I should simply change it to `java.home`, do no validation and therefore also no notifications since it can't ever be wrong at this point?

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
mbien: it simply rendered with the html tags in plain text. but Disregard this, we probably don't need a notification.

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
neilcsmith-net: Well, I was going to put in a PR along the lines of the second option. Think that's probably safer? Using `jdk.home` was presumably added for some reason.

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
neilcsmith-net: Yes, disregard for this. Although I thought there was code in there to special case html. Will take a look sometime.

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
mbien: ok changed, this makes things easier. Although having jdk.home set to /usr seems to be a fairly specific misconfiguration scenario. I am sure there are more creative ways how to have an invalid jdk.home set? I guess we will see in future :)

mbien:
tested the following:
1) set netbeans_jdkhome to `/usr`
2) started NB and made sure platform had been changed to `/usr/lib/jvm/java-18-openjdk`

neilcsmith-net (approved):
Haven't tested yet, but looks good.

I think there are only a select few cases where the jdkhome setting will get through launching the iDE but the platform be broken. In the installer, this looks like it's probably the problem in this case. https://github.com/apache/netbeans/blob/master/nbi/engine/src/org/netbeans/installer/wizard/components/actions/SearchForJavaAction.java#L515 Mind you, I'd prefer the installer didn't try and set this at all.

java/java.j2seplatform/src/org/netbeans/modules/java/j2seplatform/platformdefinition/DefaultPlatformImpl.java
mbien: @neilcsmith-net I should have let you do the PR. It probably took you as much time answering my questions as if you would have done it right away. Its just that I had it already started together with the NPE PR and split it later.

Issue #4044, opened 2022-04-28
Library Javadoc not performed from source code context menu


ms270169:
### Apache NetBeans version

Apache NetBeans 13

### What happened

After adding a library (via menu Tools->Libraries) and configuring javadoc setting, the Javadoc is properly opened in browser when selected via Project Browser -> Libraries -> Show Javadoc.
But if Javadoc should be shown via context menu "Show Javadoc" in source code, the status message "Cannot perform Show javadoc here" is shown.

Class is proper imported, because building the project ans "Navigate -> Go to source" works as expected.

### Did this work correctly in an earlier version?

Apache NetBeans 12.3 or earlier

### Operating System

Ubuntu 22.04

### JDK

openjdk version "17.0.2" 2022-01-18

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

ms270169:
I made some further investigations with Google Gson library. Javadoc from source code works with [Gson 2.8.6](https://repo1.maven.org/maven2/com/google/code/gson/gson/2.8.6/), and does not work with [Gson 2.9.0](https://repo1.maven.org/maven2/com/google/code/gson/gson/2.9.0/).

I found out the very strange fact, that it does not matter on javadoc file (or folder), it depends on the library jar file.

There are indeed structeral differences inside the library file `gson.jar`:
* version 2.8.6 (good version): file `module-info.class` on root but no folder `versions`
* version 2.9.0 (bad version): no file `module-info.class` on root but a folder `versions/9` which contains the file `module-info.class`

ms270169:
I have now modified gson.jar by copying `version/0/module-info.class` to `/module-info.class` inside the jar file (which is a zip archive). After this modification Netbeans 13 is able to open Javadoc also from source.
At the moment I don't know who is makeng the mistake (the gson library or Netbeans 13).

matthiasblaesing:
I can reproduce this with a simple Maven Projekt: [mavenproject1.zip](https://github.com/apache/netbeans/files/8596824/mavenproject1.zip)

The result is:

![Bildschirmfoto vom 2022-04-30 13-59-34](https://user-images.githubusercontent.com/2179736/166104633-610f2cf1-e45a-4a2b-aad1-1b2807d290fa.png)

In the log it can be observed:

```
FINE [org.netbeans.api.java.source.ElementHandle]: Resolving element kind: METHOD
[...]

PR #4391, opened 2022-07-17
[NETBEANS-4044] Fixing patching of modular libraries

jlahoda:
Consider a binary library with an attached source. When opening a source file for the library, if the library is non-modular, the binary will be on the classpath, and the source will be able to refer to it. When the library is modular, javac will not look at the classpath anymore, but will try to lookup module-info and other dependencies on source and output path. So, we need to make sure the output path contains the binary library, so that the lookups work. This is what this patch is trying to achieve.

---

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
@@ -21,6 +21,7 @@ package org.netbeans.modules.java.source.parsing;
 import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
@@ -29,6 +30,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
I tested this change and it works as described in https://github.com/apache/netbeans/issues/4044#issuecomment-1179779238. However the testfailure from travis of [12666.13 Test Java modules with nb-javac on Java 8](https://app.travis-ci.com/github/apache/netbeans/jobs/576885446) is locally reproducible if you run the unittests locally with JDK-8.

matthiasblaesing:
Both github-actions and travis are happy now. I think this would be worth to be targeted for NB15. @jlahoda What do you think? As branching of has happend, this would need to be retargeted for delivery.

mbien:
@matthiasblaesing @neilcsmith-net i just added the NB15 label, maybe we can get this into RC2 or so. Unless @jlahoda doesn't want to.

neilcsmith-net:
@mbien we can get it in RC1 if we're quick about it! Going to change the base branch.  If it merges cleanly I will do.

neilcsmith-net:
Am going to merge this into delivery.  It doesn't seem to be a full fix for #4044 though.  Using @matthiasblaesing example project, docs in completion and navigate to source now work, but the reported Show Javadoc failure still applies.  This can be fixed by adding a `module-info` to the project.  I don't think that's unexpected.  I think we need a better way of inferring a module name from an `Element` than I added in #4363

neilcsmith-net:
First run of RC1 and I see a bunch of NPEs with the following stack trace -

```
SEVERE [org.openide.util.RequestProcessor]: Error in RequestProcessor org.netbeans.modules.parsing.ui.indexing.errors.ErrorAnnotator$1
java.lang.NullPointerException: Cannot invoke "java.util.Set.stream()" because "this.moduleLocations" is null
	at org.netbeans.modules.java.source.parsing.PatchModuleFileManager.fixLocation(PatchModuleFileManager.java:299)
	at org.netbeans.modules.java.source.parsing.PatchModuleFileManager.getJavaFileForInput(PatchModuleFileManager.java:259)
	at org.netbeans.modules.java.source.parsing.ProxyFileManager.getJavaFileForInput(ProxyFileManager.java:392)
	at com.sun.tools.javac.code.ModuleFinder.getModuleInfoFromLocation(ModuleFinder.java:285)
	at com.sun.tools.javac.code.ModuleFinder.findSingleModule(ModuleFinder.java:200)
```

PR #4392, opened 2022-07-17
NEON - added FlatLaf Dark specific colors to improve legibility

KacerCZ:
Changed colors of valued block, reference and number to improve legibility when using FlatLaf Dark color profile.

Before:
before

After:
after


lkishalmi (approved):
Thank you!

PR #4396, opened 2022-07-18
Change output encoding fallback behavior to account for JEP400

### LINKED ISSUE #3824: Can't print multi-byte character in the Output window with Java 18
### Apache NetBeans version

Apache NetBeans 13

### What happened

With JDK 18 as NetBeans runtime, the output window can't show multi-byte character correctly.

### How to reproduce

Running the code below.
```
public class Main {
        System.out.println("こんにちは");
    }
}
```

It shows this.
image

No

### Operating System

Windows

### JDK

JDK-18

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

Adding `-J-Dfile.encoding=COMPAT` for `netbeans_default_options` in netbeans.conf should be the solution with Java 18.
image

But it affects with Java 17 incorrectly.
image

No

### Code of Conduct

Yes

duoduobingbing:
Since JEP400/JDK18 `Console.charset()`/`stdout.encoding`/`sun.stdout.encoding` is used by default for the console's encoding. Prior to this it was `Charset.defaultEncoding()`/`file.encoding`.

This PR changes the order the encoding is inferred by to accomodate for this behavior.

This should fix #3824 as well as #4261.

As `stdout.encoding`/`sun.stdout.encoding` is settable, in the rare case that the default value should not match the desired Output encoding it can be changed via the corresponding commandline argument.

Please see [this comment](https://github.com/apache/netbeans/issues/4261#issuecomment-1185669743) by @neilcsmith-net for further details

[Requested via](https://lists.apache.org/thread/pzv6899dn4w2syo6dk400qc3n9wr6slf)

### DIFF
--- a/ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
+++ b/ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
@@ -141,6 +141,45 @@ public final class BaseExecutionService {
     }
 
     /**
+     * Infers the output encoding from the relevant system properties, if those should all be <code>null</code>
+     * then this will fallback to <code>Charset.defaultCharset()</code>
+     * 
+     * Since JDK 18 and JEP400 Console.charset() is used for the console's encoding instead of <code>Charset.defaultCharset()</code>. 
+     * Console.charset() is exposed via stdout.encoding/sun.stdout.encoding.
+     * If ran with JDK<=16 stdout.encoding and native.encoding should be null and the old default of <code>Charset.defaultCharset()</code> will be used to match pre JEP400 behavior.
+     * 
+     * The checking order for the encoding is stdout.encoding, sun.stdout.encoding, native.encoding, <code>Charset.defaultCharset()</code>
+     * 
+     * @see org.netbeans.modules.maven.execute.CommandLineOutputHandler#getPreferredCharset
+     * 
+     * @return inferred encoding as Charset
+     */
+    private static Charset getInputOutputEncoding(){
+        String[] encodingSystemProperties = new String[]{"stdout.encoding", "sun.stdout.encoding", "native.encoding"};
+
+        Charset preferredCharset = null;
+        for (String encodingProperty : encodingSystemProperties) {
+            String encodingPropertyValue = System.getProperty(encodingProperty);
+            if (encodingPropertyValue == null) {
+                continue;
+            }
+
+            try {
+                preferredCharset = Charset.forName(encodingPropertyValue);
+            } catch (IllegalArgumentException ex) {
+                LOGGER.log(java.util.logging.Level.WARNING, "Failed to get charset for '" + encodingProperty + "' value : '" + encodingPropertyValue + "'", ex);
+            }
+
+            if (preferredCharset != null) {
+                return preferredCharset;
+            }
+
+        }
+
+        return Charset.defaultCharset();
+    }
+
+    /**
      * Runs the process described by this service. The call does not block
      * and the task is represented by the returned value. Integer returned
      * as a result of the {@link Future} is exit code of the process.
@@ -209,28 +248,17 @@ public final class BaseExecutionService {
                     executor = Executors.newFixedThreadPool(in != null ? 3 : 2);
 
                     Charset charset = descriptor.getCharset();
+                    // The CommandLineOutputHandler used the default charset to convert
+                    // output from command line invocations to strings. That encoding is
+                    // derived from the system file.encoding. From JDK 18 onwards its
+                    // default value changed to UTF-8.
+                    // JDK 17+ exposes the native encoding as the new system property
+                    // native.encoding, prior versions don't have that property and will
+                    // report NULL for it.
+                    // To account for the behavior of JEP400 the following order is used to determine the encoding:
+                    // stdout.encoding, sun.stdout.encoding, native.encoding, Charset.defaultCharset()
                     if (charset == null) {
-                        // If charset is not set for the descriptor, the
-                        // BaseExecutionService used the default charset to convert
-                        // output from command line invocations to strings. That encoding is
-                        // derived from the system file.encoding. From JDK 18 onwards its
-                        // default value changed to UTF-8.
-                        // JDK 18+ exposes the native encoding as the new system property
-                        // native.encoding, prior versions don't have that property and will
-                        // report NULL for it.
-                        // The algorithm is simple: If native.encoding is set, it will be used
-                        // else the old default will be queried via Charset#defaultCharset.
-                        String nativeEncoding = System.getProperty("native.encoding");
-                        if (nativeEncoding != null) {
-                            try {
-                                charset = Charset.forName(nativeEncoding);
-                            } catch (Exception ex) {
-                                LOGGER.log(java.util.logging.Level.WARNING, "Failed to get charset for native.encoding value : '" + nativeEncoding + "'", ex);
-                            }
-                        }
-                        if (charset == null) {
-                            charset = Charset.defaultCharset();
-                        }
+                        charset = BaseExecutionService.getInputOutputEncoding();
                     }
 
                     tasks.add(InputReaderTask.newDrainingTask(
--- a/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
@@ -202,7 +202,7 @@ public class CommandLineOutputHandler extends AbstractOutputHandler {
         private boolean skipLF = false;
 
         public Output(InputStream instream) {
-            str = new BufferedReader(new InputStreamReader(instream, getNativeCharset()));
+            str = new BufferedReader(new InputStreamReader(instream, getPreferredCharset()));
         }
 
         private String readLine() throws IOException {
@@ -701,7 +701,7 @@ public class CommandLineOutputHandler extends AbstractOutputHandler {
 
         public @Override void run() {
             Reader in = inputOutput.getIn();
-            try (Writer out = new OutputStreamWriter(str, getNativeCharset())) {
+            try (Writer out = new OutputStreamWriter(str, getPreferredCharset())) {
                 while (true) {
                     int read = in.read();
                     if (read != -1) {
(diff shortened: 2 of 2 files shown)


ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,43 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
...
+
+        Charset preferredCharset = null;
+        for(String encodingProperty : encodingSystemProperties){
+            String encodingPropertyValue = System.getProperty(encodingProperty);
+            if(encodingPropertyValue == null){
+                continue;
+            }
+
+            try {
+                preferredCharset = Charset.forName(encodingPropertyValue);
```
neilcsmith-net, line 166: Isn't this missing a `break`?!

Could also chain the variant of System::getProperty with default values to get rid of the loop - pros and cons to validating each value.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,43 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
...
+            } catch (Exception ex) {
+                LOGGER.log(java.util.logging.Level.WARNING, "Failed to get charset for '" + encodingProperty + "' value : '" + encodingPropertyValue + "'", ex);
+            }
+
+            if(preferredCharset != null){
```
neilcsmith-net, line 171: Please correct the spacing.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,43 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
         return new BaseExecutionService(processCreator, descriptor);
     }
 
+    /**
+     * Infers the output encoding from the relevant system properties, if those should all be <code>null</code>
+     * then this will fallback to <code>Charset.defaultCharset()</code>
+     * 
+     * Since JDK 18 and JEP400 Console.charset() is used for the console's encoding instead of <code>Charset.defaultCharset()</code>. 
+     * Console.charset() is exposed via stdout.encoding/sun.stdout.encoding.
+     * If ran with JDK<=16 these properties should be null and the old default of <code>Charset.defaultCharset()</code> will be used to match pre JEP400 behavior.
```
neilcsmith-net, line 149: At least `sun.stdout.encoding` may not be `null` on JDK<=16. That predates JEP400.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,43 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
...
+        Charset preferredCharset = null;
+        for(String encodingProperty : encodingSystemProperties){
+            String encodingPropertyValue = System.getProperty(encodingProperty);
+            if(encodingPropertyValue == null){
+                continue;
+            }
+
+            try {
+                preferredCharset = Charset.forName(encodingPropertyValue);
+            } catch (Exception ex) {
```
mbien, line 167: nitpick: would recommend changing this to `IllegalArgumentException`, since this is the common supertype of all exceptions `forName` throws.

java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
```diff
@@ -828,29 +828,38 @@ public ExecutionEventObject.Tree getExecutionTree() {
...
+        for(String encodingProperty : encodingSystemProperties){
+            String encodingPropertyValue = System.getProperty(encodingProperty);
+            if(encodingPropertyValue == null){
+                continue;
+            }
+
             try {
-                nativeCharset = Charset.forName(nativeEncoding);
+                preferredCharset = Charset.forName(encodingPropertyValue);
             } catch (Exception ex) {
```
mbien, line 852: same here

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,43 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
...
+     * 
+     * Since JDK 18 and JEP400 Console.charset() is used for the console's encoding instead of <code>Charset.defaultCharset()</code>. 
+     * Console.charset() is exposed via stdout.encoding/sun.stdout.encoding.
+     * If ran with JDK<=16 these properties should be null and the old default of <code>Charset.defaultCharset()</code> will be used to match pre JEP400 behavior.
+     * 
+     * The checking order for the encoding is stdout.encoding, sun.stdout.encoding, native.encoding, <code>Charset.defaultCharset()</code>
+     * 
+     * @return inferred encoding as Charset
+     */
+    private static Charset getInputOutputEncoding(){
```
mbien, line 155: please mention that this method is duplicated in `CommandLineOutputHandler`, e.g via `@see`.

java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
```diff
@@ -828,29 +828,38 @@ public ExecutionEventObject.Tree getExecutionTree() {
         
     }    
 
-    private static Charset getNativeCharset() {
+    private static Charset getPreferredCharset() {
```
mbien, line 831: please mention that this method is a copy of BaseExecutionService in a comment.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
duoduobingbing: Where is the location where a `break` is missing?

If the `encodingPropertyValue` is `null` it will cycle through the next property, if the property is filled it will be returned on line 171 if set, or am I missing something?

mbien (approved):
looks good to me, comments inline.

however i am not a windows expert - I can't account for edge cases. 99% of my projects these use the safe space of UTF-8, since thats the standard on the web, servers and linux boxes in general.

but the priority of `stdout` > `native` makes somewhat sense to me.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
duoduobingbing: spacing

Is all of 167-171 spaced wrongly? Because line 168 should have the same number of spaces line 172 (16 spaces)

neilcsmith-net: eg. that line should be `if (preferredCharset != null) {` - ie. please add spaces around brackets and braces.

The easiest way is to select the sections you've changed (_and only the sections you've changed!_) and auto format it (`Alt+Shift+F`)

duoduobingbing:
@neilcsmith-net, @mbien thanks for looking into this.

I (hopefully) adressed all comments with the newest commit.

duoduobingbing:
> Thanks for the changes. Can you squash into a single commit? 

Of course.

mbien:
@duoduobingbing the commit header uses your github handle, I believe this project requires real names for contributions:

> Please make sure (eg. git log) that all commits have a valid name and email address for you in the Author field.

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
```diff
@@ -140,6 +140,45 @@ public static BaseExecutionService newService(@NonNull Callable<? extends Proces
...
+     * If ran with JDK<=16 stdout.encoding and native.encoding should be null and the old default of <code>Charset.defaultCharset()</code> will be used to match pre JEP400 behavior.
+     * 
+     * The checking order for the encoding is stdout.encoding, sun.stdout.encoding, native.encoding, <code>Charset.defaultCharset()</code>
+     * 
+     * @see org.netbeans.modules.maven.execute.CommandLineOutputHandler#getPreferredCharset
+     * 
+     * @return inferred encoding as Charset
+     */
+    private static Charset getInputOutputEncoding(){
+        String[] encodingSystemProperties = new String[]{"stdout.encoding, sun.stdout.encoding, native.encoding"};
```
neilcsmith-net, line 158: This is one String not three!

java/maven/src/org/netbeans/modules/maven/execute/CommandLineOutputHandler.java
```diff
@@ -827,30 +827,45 @@ public ExecutionEventObject.Tree getExecutionTree() {
...
         // native.encoding, prior versions don't have that property and will
         // report NULL for it.
-        // The algorithm is simple: If native.encoding is set, it will be used
-        // else the old default will be queried via Charset#defaultCharset.
-        String nativeEncoding = System.getProperty("native.encoding");
-        Charset nativeCharset = null;
-        if (nativeEncoding != null) {
+        // To account for the behavior of JEP400 the following order is used to determine the encoding:
+        // stdout.encoding, sun.stdout.encoding, native.encoding, Charset.defaultCharset()
+        String[] encodingSystemProperties = new String[]{"stdout.encoding, sun.stdout.encoding, native.encoding"};
```
neilcsmith-net, line 847: This is one String not three!

neilcsmith-net (changes requested):
Just been able to actually test this.  It doesn't work currently because you're only creating a single String value with all properties, rather than multiple Strings.

eg. 

```java
"stdout.encoding, sun.stdout.encoding, native.encoding"
```

vs

```java
"stdout.encoding", "sun.stdout.encoding", "native.encoding"
```

Please check this and @mbien comment about real names in commit info. Thanks!

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
mbien: btw I just noticed that you only fixed one occurrence, The spaces are also missing at the for loop, the other if and one of the two method declarations.

duoduobingbing:
Thanks for the feedback and for pointing out my (admittedly embarrassing) mistake of not segmenting the array again after cleanup.

`git log` should now show everything correctly

ide/extexecution.base/src/org/netbeans/api/extexecution/base/BaseExecutionService.java
duoduobingbing: Hopefully I got every spacing error now; I noticed that I had set my IDE to the wrong conventions. Thus Autoformat caused these.

duoduobingbing:
Spacing should now be correct as well

PR #536, opened 2018-05-06
Per-cluster repository layout

JaroslavTulach:
Up until now the repository structure of NetBeans source code has been flat - module next to a module. That comes with advantages, but also drawbacks. Recently two new cons appeared:

* github isn't willing to show the [whole repository](https://github.com/jtulach/netbeans-releases) - it limits the list to just a thousand files
* new modules are (hopefully) going to be granted by Oracle and not all of them are of the same importance

Luckily [NetBeans Runtime Container](http://wiki.apidesign.org/wiki/NetBeans_Platform) already supports grouping modules into bigger chunks - e.g. **clusters**. Thus it is natural that from time to time people would like to group the sources into **clusters** as well. This PR is another attempt to do so.

The `platform` modules are being moved into `platform` subdirectory. Other clusters shall follow, which would leave us with five to twenty top level directories in the repository (once more of the original NetBeans code is donated). There already is support for nesting modules into subdirs (used by [contrib](http://hg.netbeans.org/main/contrib/) repository) - e.g. this is about extending it to work with all the clusters.

The list of clusters and modules is specified in `nbbuild/cluster.properties` and `nbbuild/build.properties`. In general there seem to be two approaches:
* change these files to use full path to modules - e.g. `openide.util` would become `platform/openide.util`
* change the `nbbuild/antsrc` tasks to search for the modules in cluster subdirs, but keep their names.

I started with the latter approach, but currently the patch looks more like former one. I am not sure it is the greatest idea, as that means all references to properties like `${libs.osgi}` need to be replaced by `${platform/libs.osgi}` and it would be better to avoid it. 

In addition to this change, we also need to extend the the `apisupport` modules to recognize this new layout. Such tooling change shall be integrated soon,  as otherwise the IDE will not be able to work with own sources (btw. a drawback is that NetBeans 8.2 doesn't recognize this new layout).

Anyway this is a work in progress. I am creating this PR to get an early feedback on the overall direction. CC @jglick, @tzezula, @sdedic, @thurka, @jlahoda

PR #4397, opened 2022-07-18
attemp to fix javadoc for release100

ebarboni:
Hi,
 I just started apidoc polish :D. Will take sometime I guess. To not alter release I choose the release100 to startwith.

Attempt so far
1)I reduced a lot of links issue by changing the nbbuild/antsrc/org/netbeans/nbbuild/CheckLinks.java pattern to tackle the id parameter and h3 element. 
2)The following code on element are no more tolerated on javadoc by default onclick="target='_blank'". I removed them in arch/apichanges. 
3) Replacing all http://bits.netbeans.org/html+java by https://bits.netbeans.org/html+java/dev/ in arch/apichanges

that lead to a shorter [checklinks-errors.txt](https://github.com/apache/netbeans/files/9132010/checklinks-errors.txt)

But know I have issues on a more highlevel matter.
One type of issue is for example 
org-netbeans-modules-editor-codetemplate would need the org-openide-options apidoc but we don't generate org-openide-options.
same issue for editor-lib we only have editor-libs2.
What could be the good practice to get apidoc consitent again.
Second issue
org-openide-awt/org/openide/awt/ActionState.html because ActionState.java is in another subfolder (package is ok )
[ org.openide.awt.awt folder ](https://github.com/apache/netbeans/tree/release100/platform/openide.awt/src/org/openide/awt/awt)
Could we move the file to fix ? (we will never cut a release from the old branch I guess) 

Any tips appreciated, and will later fix for other branches (once I have the good advices)

All the best


ebarboni:
@neilcsmith-net @jtulach  with the 3 commit I'm able to down the error to 0 on my own build.
1) fix pattern for id and h3 [nbbuild/antsrc/org/netbeans/nbbuild/CheckLinks.java]
2) remove invalid target in arch.xml/apichanges.xml (onclick is forbidden)
3) move 3 files in openide awt to their correct public package
4) as we use config.javadoc.stable for release, I had to append this properties with the missing modules:
   openide.options,\
    editor,\
    editor.lib,\
    editor.indent,\
    editor.indent.support,\
    editor.bracesmatching,\
    editor.util

Step 4 is a bit ugly but on master, we should try to edit the arch file to match the real dependencies.

neilcsmith-net:
> move 3 files in openide awt to their correct public package

That one is weird. They only seem to be in the `awt` subfolder for a period of time - didn't start there, aren't there on master.

ebarboni:
it compile javac side (ant build) but javadoc tool miss them.

jtulach:
Great work. The [list of broken links](https://github.com/apache/netbeans/files/9132010/checklinks-errors.txt) isn't that huge anymore!

`openide.options` has been deprecated, I believe. `editor.lib` may not be stable yet. That is why we don't include these modules in _stable javadoc_ configuration.

Files in `org.openide.awt.awt` package, that's horrible!

jtulach (approved):

ebarboni:
In master I guess we should check some of the arch.xml but for the former I would like to be lazy a bit if you don't mind.

The deprecated, private modules are flagged by themselves so it's red or striked and will be showed to user.

neilcsmith-net:
> `openide.options` has been deprecated, I believe. `editor.lib` may not be stable yet. That is why we don't include these modules in _stable javadoc_ configuration.

I'd be in favour of publishing everything on bits.netbeans.org, stable, development and (probably) deprecated.

> Files in `org.openide.awt.awt` package, that's horrible!

Seem to have ended up there as a side effect of #536

ebarboni:
waiting a few hours and will try to merge if green on build I will propagate same workflow to others branches that build with jdk 8

ebarboni:
it's green https://ci-builds.apache.org/job/Netbeans/job/netbeans-TLP/job/netbeans/job/release100/

PR #4412, opened 2022-07-20
Sync delivery to release150 for 15-rc2

neilcsmith-net:


neilcsmith-net:
Merging early to enable some build testing cc @mbien @ebarboni 

There may be a second sync for rc2 if needed.

PR #4414, opened 2022-07-20
Fix #4411 - ensure moduleLocations is initialized.

### LINKED ISSUE #4411: NPE : because "this.moduleLocations" is null
The code introduced in #4391 to address #4044 causes a lot of error notifications / NPE reports.  Adding this to track for 15-rc2.  Seems possibly to only affect projects using Java modules.

Stack trace -

```
SEVERE [org.openide.util.RequestProcessor]: Error in RequestProcessor org.netbeans.modules.parsing.ui.indexing.errors.ErrorAnnotator$1
java.lang.NullPointerException: Cannot invoke "java.util.Set.stream()" because "this.moduleLocations" is null
	at org.netbeans.modules.java.source.parsing.PatchModuleFileManager.fixLocation(PatchModuleFileManager.java:299)
	at org.netbeans.modules.java.source.parsing.PatchModuleFileManager.getJavaFileForInput(PatchModuleFileManager.java:259)
	at org.netbeans.modules.java.source.parsing.ProxyFileManager.getJavaFileForInput(ProxyFileManager.java:392)
	at com.sun.tools.javac.code.ModuleFinder.getModuleInfoFromLocation(ModuleFinder.java:285)
	at com.sun.tools.javac.code.ModuleFinder.findSingleModule(ModuleFinder.java:200)
```

neilcsmith-net:
Fix #4411 - protect against NPE by ensuring moduleLocations is initialized in PatchModuleFileManager::fixLocation.

I think this is the right approach rather than bailing on null.  At least, every other use in that class is via the initializing method.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
@@ -296,7 +296,8 @@ final class PatchModuleFileManager implements JavaFileManager {
 
     private Location fixLocation(Location input) throws IOException {
         if (input == StandardLocation.CLASS_OUTPUT && overrideModuleName != null) {
-            return this.moduleLocations.stream().filter(pl -> overrideModuleName.equals(pl.getModuleName())).map(pl -> (Location) pl).findAny().orElse(input);
+            return moduleLocations(StandardLocation.PATCH_MODULE_PATH)
+                    .stream().filter(pl -> overrideModuleName.equals(pl.getModuleName())).map(pl -> (Location) pl).findAny().orElse(input);
         } else {
             return input;
         }


java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
```diff
@@ -296,7 +296,8 @@ public FileObject getFileForOutput(Location location, String packageName, String
 
     private Location fixLocation(Location input) throws IOException {
         if (input == StandardLocation.CLASS_OUTPUT && overrideModuleName != null) {
-            return this.moduleLocations.stream().filter(pl -> overrideModuleName.equals(pl.getModuleName())).map(pl -> (Location) pl).findAny().orElse(input);
+            return moduleLocations(StandardLocation.PATCH_MODULE_PATH)
+                    .stream().filter(pl -> overrideModuleName.equals(pl.getModuleName())).map(pl -> (Location) pl).findAny().orElse(input);
```
mbien, line 300: this looks good to me, I actually wanted to propose the same but you were faster this time :)

The confusing part while looking through the code however was that `moduleLocations(foo)` only allows foo to be the constant `StandardLocation.PATCH_MODULE_PATH`, everything else throws an exception. So it would be more intuitive to simply make that a `getModueLocations()` without parameter.

mbien (approved):

matthiasblaesing (approved):
I could not reproduce the problem, but change looks sane to me.

jlahoda (approved):
Looks good to me - thanks and sorry for trouble!

java/java.source.base/src/org/netbeans/modules/java/source/parsing/PatchModuleFileManager.java
neilcsmith-net: Got to be faster than you sometime! :smile: I kept the fix minimal but had similar thoughts wondering why that method takes the location.  Mind you `moduleLocations()` would be fine - no need for `get` on everything!

neilcsmith-net:
Thanks @matthiasblaesing @jlahoda - merging.

@matthiasblaesing only seems to happen with explicit module projects - try adding a `module-info` to the project you shared in #4044 and you should see it blow up straight away.  Also affects loading projects with a module info.

Issue #4393, opened 2022-07-17
NB-14: Cloning/fetching/pushing remote repositories with Git over SSH disregards cnd.tmpbase


zzzyxwvut:
### Apache NetBeans version

Apache NetBeans 14

### What happened

In `Apache NetBeans` 14, `Git`-related support for cloning,
fetching from, or pushing to remote repositories is broken,
as outlined below, whenever the temporary-file directory
(`/tmp`) is mounted with the `noexec` option set.

In prior releases, having `cnd.tmpbase` system property
defined, allowed to select a temporary-file directory where
necessary execution would have taken place. It may, as well,
work still, but its coverage seems to slip ([junixsocket](https://github.com/kohlschutter/junixsocket)).

### How to reproduce

[P&nbsp;R&nbsp;E&nbsp;L&nbsp;I&nbsp;M&nbsp;I&nbsp;N&nbsp;A&nbsp;R&nbsp;Y&nbsp;&nbsp;&nbsp;&nbsp;S&nbsp;T&nbsp;E&nbsp;P]

Set up the `ssh` transport protocol for a target repository
(locally generate public-private key pair, upload the public
key).
Create a new directory that can be read, written, searched
by an `Apache NetBeans` process and which can be used by it
for the management of its own temporary files.
Enable keyring-related logging in `netbeans/etc/netbeans.conf`:
`netbeans_default_options="... -J-Dorg.netbeans.modules.keyring.level=0"`
<br>

Proceed with the following steps (FAILURE and SUCCESS) in
any order.
<br><br><br>

[F&nbsp;A&nbsp;I&nbsp;L&nbsp;U&nbsp;R&nbsp;E]

Manually _forbid_ execution of binaries and other executable
files from the `/tmp` directory:
`sudo mount --options remount,noexec /tmp`
`mount | grep /tmp`

Launch `Apache NetBeans` 14 with `-J-Dcnd.tmpbase=/path/to/new/dir`.
(Substitute the actual path to the newly-created directory
for `/path/to/new/dir`.)

Try either cloning or fetching from or pushing to some `Git`
repository for which the `ssh` transport protocol was set.
<br>

When CLONING...

Go Team -> Git -> Clone

[Repository URL: git@...]

Check "Private/Public Key", type in the passphrase, and
press the "Next" button.
Expect a message written just above the "Next" button:
> Cannot connect to repository at git@... .

Press "Cancel" and exit `Apache NetBeans`.

[...]

PR #4420, opened 2022-07-21
Enable direct loading of native library for junixsocket instead of loading from temp

matthiasblaesing:
junixsocket loads by default the native library using System#loadLibrary
so the native libraries only need to be placed in the

modules/lib/\<arch>/\<os>

folders, so that they can be loaded directly with the requirement to
expand them and load them from java.io.tmpdir.

Closes: #4393

### DIFF
--- a/ide/libs.c.kohlschutter.junixsocket/nbproject/project.properties
+++ b/ide/libs.c.kohlschutter.junixsocket/nbproject/project.properties
@@ -17,4 +17,17 @@
 
 release.external/junixsocket-common-2.4.0.jar=modules/ext/junixsocket-common-2.4.0.jar
 release.external/junixsocket-native-common-2.4.0.jar=modules/ext/junixsocket-native-common-2.4.0.jar
+release.external/junixsocket-native-common-2.4.0.jar!/lib/x86_64-MacOSX-clang/jni/libjunixsocket-native-2.4.0.dylib=modules/lib/x86_64/libjunixsocket-native-2.4.0.dylib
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
Out of interest, why the additional module dependencies?

matthiasblaesing:
> Out of interest, why the additional module dependencies?

Artifacts from a first try. I tried to follow the JNA example to closely (the JNA module is using a module Installer, which junixsocket does not need). I removed the installer and the manifest changes, regrettably the changes to the dependencies were retained. I pushed an updated PR.

The corresponding binary: https://doppel-helix.eu/NetBeans-dev-dev-97f2a4617b23c157eee90abbcb6159fd9bda188f-release.zip

This does not invalidate tests done with the original change, as the depencies are not the critical path.

neilcsmith-net:
Thanks @matthiasblaesing - had a feeling the JNA ModuleInstall was the reason. :smile:

Will let the tests go green again and merge for rc2.

matthiasblaesing:
I did a smoke test on mac OS (amd64) - git integration with the ssh agent still works there also.

neilcsmith-net:
@mcdonnell-john could you possibly have a look when rc2 is ready whether this is going to cause a signing / notarization issue?  I had issues with signing the junixsocket library bundled in the jar because it's built against too low a version of the macos sdk. However, the jar passed notarization then.  It's possible the unbundled libraries might cause an issue though.  If they do, we might need to consider either a workaround or upgrade to 2.5.0.

neilcsmith-net:
@matthiasblaesing OK, verified this is now failing notarization on macOS. Opened issue #4460 Any concerns with updating to 2.5.1?

PR #4421, opened 2022-07-21
Unbalanced hint should work for fields initialized in constructor

### LINKED ISSUE #4402: "Unbalanced read/write with collections" hint fails for fields initialized in constructor
### Apache NetBeans version

Apache NetBeans 14

### What happened

I've a Maven project of type *jar*. Suppose that I also have a class with a single field of type `List`. Then, I initialize that field in the declaration itself. If I never add any element to that collection, NetBeans properly warns me that *The collection is never added to*.

However, if I initialize the field in the constructor, NetBeans seems to be unable to detect that I still have not added any element to the collection.

### How to reproduce

In this piece of code, NetBeans correctly detects that no element has been added to the `collection` field:

```java
public final class Test {

	private final List<String> collection = new ArrayList<> ();
                           // ^^^ WARNING! The collection is never added to.

	public Test () {
	}

	public boolean isEmpty () {
		return collection.isEmpty ();
	}

}
```

However, if I initialize the field in the constructor, the hint is never shown:

```java
public final class Test {

	private final List<String> collection;

	public Test () {
		collection = new ArrayList<> ();
	}

	public boolean isEmpty () {
		return collection.isEmpty ();
	}

}
```

No

### Operating System

Debian GNU/Linux 11.4 (Bullseye)

### JDK

OpenJDK Runtime Environment (build 17.0.3+7-Debian-1deb11u1)

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

_No response_

No

### Code of Conduct

Yes

mbien:
- we can make this work for final fields, non-final fields would
   be more complicated [explained here](https://github.com/apache/netbeans/issues/4402#issuecomment-1192312559)
 - added missing methods to READ_METHODS set

fixes #4402

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
@@ -103,14 +103,6 @@ public class Unbalanced {
 
     @Hint(displayName = "#DN_org.netbeans.modules.java.hints.bugs.Unbalanced.Array", description = "#DESC_org.netbeans.modules.java.hints.bugs.Unbalanced.Array", category="bugs", options=Options.QUERY, suppressWarnings="MismatchedReadAndWriteOfArray")
     public static final class Array {
-        private static final Set<Kind> ARRAY_WRITE = EnumSet.of(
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
Not tested but makes sense. Need to merge final things to prepare rc2. Guess we switch to merge then review with this one! cc @jlahoda @dbalek @matthiasblaesing @sdedic

PR #4429, opened 2022-07-22
remove usages of deprecated junit.framework.Assert.

mbien:
project wide cleanup which intends to:
 - get rid of many build log warnings
 - reduce the mixing of two junit APIs

The annoying part is that junit 4 made the decision of simply throwing `java.lang.AssertionError` in [`Assert.*`](https://junit.org/junit4/javadoc/latest/src-html/org/junit/Assert.html#line.85) (and i forgot about that), while junit 3 and 5 throw a subtype, so that it is possible to distinguish between language level `assert` and test `Assert`. Further, `NBTestCase` extends `TestCase` which extends the deprecated `Assert`.

So i did the following:
- if it doesn't extend `NBTestCase` use `org.junit.Assert`
- unless any of the tests catch `AssertionFailedError` (there are not many instances, second commit), then we use `junit.framework.TestCase.*` which is essentially what `NBTestCase` uses (but there won't be deprecation warnings), since catching `AssertionError` might break tests in non obvious ways

This should hopefully clean things up without making it harder to migrate to junit 5 in future.


ide/html.editor/test/unit/src/org/netbeans/modules/html/editor/completion/HtmlCompletionTestSupport.java
```diff
@@ -114,7 +113,7 @@ public void run() {
 
         try {
             assertCompletionItemNames(expectedItemsNames, items, type);
-        } catch (AssertionFailedError e) {
+        } catch (AssertionError e) {
```
matthiasblaesing, line 116: Wouldn't it be easier to create this output in `assertCompletionItemNames`, add a sensible info in front of it and use it as the message of the delegated `assert*` calls there?

matthiasblaesing (commented):
Looks ok to me. I left an inline comment.

ide/html.editor/test/unit/src/org/netbeans/modules/html/editor/completion/HtmlCompletionTestSupport.java
mbien: the problem is that this is a public utility method with several usages. I didn't want to rewrite too much and make this PR even larger. Simply catching `AssertionError` seemed like the easiest solution.

It would have been dangerous if the utility method would call into NB code which has java assert statements and the test would expect a junit assert failure. If NB code would fail for the wrong reason, the test would pass instead of fail. But this here should be safe.

matthiasblaesing: I won't go into discussion over this and feel free to ignore this.

BradWalker:
Looks good to me..

BradWalker (approved):
Looks good to me..

mbien:
squashing the commits and preparing for merge

PR #3727, opened 2022-03-08
Fix on PHP tests on Windows

oyarzun:
- Only run PHP workflows when there are changes in the PHP cluster.
- Move Windows PHP tests to PHP workflow
- Fix and/or disable PHP test on Windows 

---
**^Add meaningful description above**

By opening a pull request you confirm that, unless explicitly stated otherwise, the changes -

 - are all your own work, and you have the right to contribute them.
 - are contributed solely under the terms and conditions of the Apache License 2.0 (see section 5 of the license for more information).

Please make sure (eg. `git log`) that all commits have a valid name and email address for you in the Author field.

If you're a first time contributor, see the Contributing guidelines for more information.

PR #4431, opened 2022-07-23
conditional CI pipeline + migrate some jobs from travis to gh

mbien:
Now that we have the basic CI features implemented (cancellable, restartable, junit-reports and a simple pipeline), we can migrate some travis jobs to github.

I don't think we can migrate everything, since apache has a limited number of job runners which are shared between all apache projects - using them all every time someone opens a PR wouldn't be polite :) 

so far migrated:
sig test, java hints, javafx, enterprise, java modules, profiler, apisupport, extide, platform modules, webcommon, rat, lib validation
 - travis: 31 -> 14 jobs (~12.5h -> ~5h)
 - gh: 12 (fixed) ->  25 (conditional) (~[4h](https://github.com/apache/netbeans/actions/runs/2894573402/usage) -> ~[3h](https://github.com/apache/netbeans/actions/runs/2832175091/usage) - [8h](https://github.com/apache/netbeans/actions/runs/2899043480/usage))
 (counts individual matrix axis as job)

The good news is that a lot of time is saved simply by having the jobs as part of the pipeline, the initial work is now paying off, travis did build in every job.
Another way to save CI time is by triggering long running jobs/steps only with certain PR labels, see below.

other changes:
 - travis uses the `travis_retry` command on many unreliable tests. Gh doesn't have that so I had to find a workaround with the help of a little script. The summary will now also show warnings:
retry

- we don't have to hide logs anymore since there doesn't seem to be a log limit. Additionally: logs are split by step so which makes inspection easier

- some steps or jobs are now skipped based on [selected PR labels](https://github.com/apache/netbeans/labels?q=ci)
  - `ci:no-build` for disabling the build entirely
  - `ci:all-tests` for running all tests
  - `Java`, `JavaScript`, `Platform` and `PHP` unlock some long running steps/jobs in their respective category
  - `Ant`, `Gradle`, `Maven` or `MX` trigger the build tools job 
  - `JavaDoc` and `API Change` run javadoc jobs
  - all jobs/steps run if the workflow was not started by a PR sync, e.g during merges
  - the full text search allows to quickly check which labels influence CI, e.g by entering "ci java" or just "ci". 
ci-labels

- merged sig tests, rat, lib val into a single job for all paperwork. The job will also print the commit header so that reviewers can check if it looks ok
- merged extide, ant, gradle, maven and mx into a cross-cluster build-tools job
- moved some steps around, unreliable should run first for fast fails


neilcsmith-net:
> I don't think we can migrate everything since apache has a limited number of job runners which are shared between all apache projects - using them all every time someone opens a PR wouldn't be polite :)

No, but the situation with Travis isn't much different.  What about taking this as a chance to try some of the heavier tests on different triggers?  A daily basis on certain branches, or on PR merges?

mbien:
> > I don't think we can migrate everything since apache has a limited number of job runners which are shared between all apache projects - using them all every time someone opens a PR wouldn't be polite :)
> 
> No, but the situation with Travis isn't much different. What about taking this as a chance to try some of the heavier tests on different triggers? A daily basis on certain branches, or on PR merges?

sure thats all no problem in principle. The main issue is to actually split the tests up and decide what should run when. Triggering based on code paths, or even PR labels is also an option, but this has been discarded every time we had the discussion. PHP for example almost became conditionally triggered as first experiment.

But i do agree, this project runs probably ~20h of testing per sync (travis alone is 13h). Thats simply too much.

neilcsmith-net:
> sure thats all no problem in principle. The main issue is to actually split the tests up and decide what should run when. Triggering based on code paths, or even PR labels is also an option, but this has been discarded every time we had the discussion. 

Yes, I know! :exploding_head:   I actually meant that as a hint that you might want to propose a suitable split, whatever makes sense to you.  Then anyone who wants to -1 it can actually do the work to propose how to move things around to keep our CI usage at a workable (_and polite to the rest of ASF_) level. :smile:

mbien:
@neilcsmith-net looks promising so far since some jobs on travis did spend most of the time building. When run as part of the pipeline the time goes down to a few minutes.

I don't really know which tests need a display. I might turn the xserver on for all tests on linux to keep things simple.

observations so far: enterprise cluster tests don't work on 11, so i leave it on 8. Java Hints batch1 fails, don't know why yet.

.github/workflows/main.yml
```diff
@@ -127,12 +131,55 @@ jobs:
...
+          unzip -qq ../nbbuild/build/release-src*
+          ant $OPTS -quiet rat -Drat-report.haltonfailure=true
+
+      - name: Run check-sigtests-release
+        if: success() || failure()
+        run: ant $OPTS -quiet check-sigtests-release -Dfail.on.error=true
+        
+      - name: Run gen-sigtests-release
+        if: success() || failure()
+        run: ant $OPTS gen-sigtests-release
```
mbien, line 176: @neilcsmith-net  I tried to merge some of the "paperwork jobs" into one job, but RAT fails.
```
/home/runner/work/netbeans/netbeans/scratch/nbbuild/build.xml:2195: Failed Rat test(s):
extide/gradle/netbeans-gradle-tooling/.gradle/7.4/gc.properties
extide/gradle/netbeans-gradle-tooling/.gradle/buildOutputCleanup/cache.properties
extide/gradle/netbeans-gradle-tooling/.gradle/vcs-1/gc.properties
extide/gradle/netbeans-gradle-tooling/.gradle/7.4/dependencies-accessors/gc.properties
Unapproved license in 1 file(s) 
profiler/lib.profiler/release/BUILDINFO.txt
Unapproved license in 1 file(s) 
```

I don't know much about RAT. Does it have to be run on a repo which is not built?

neilcsmith-net: @mbien yes, RAT should be on clean checkout. Pros and cons to exceptions for build locations. Key thing is source zip is clean.

mbien: roger. The same is probably also true for sigtests. So I can run the entire job in parallel to the build job.

mbien: actually, sigtests require a build - but i will figure this out

neilcsmith-net: @mbien maybe run RAT as part of the build job? If there's a RAT failure argument for just treating it as a failing build and bailing on everything else.

mbien: @neilcsmith-net that would be also good i think. Right now I try it with a paperwork job which runs after build and does:
 - extract build artifact
 - print commit delta (allows us to check commit headers of new contributors)
 - check-sigtests-release
 - reset workspace with: git clean -xfdq
 - verify-libs-and-licenses
 - rat
 - gen-sigtests-release

This merges multiple jobs into one, lets see if this works

mbien: nevermind, this won't work since the build is actually not a git repository at this point anymore. So this will have to run as part of the build as you said.

.github/workflows/main.yml
```diff
@@ -335,7 +907,9 @@ jobs:
       - name: Test PHP Doctrine2
         run: ant $OPTS -f php/php.doctrine2 test
 
+      # todo: longest step -> needs 2 tiers, with tier 2 conditional
       - name: Test PHP Editor
+        if: contains(github.event.pull_request.labels.*.name, 'PHP') || github.event_name != 'pull_request'
         run: ant $OPTS -f php/php.editor test
```
mbien, line 913: @neilcsmith-net this seems to work. The step runs only if the PR has the PHP label on it. If this isn't a PR, it will run too (e.g merge to master). This alone would cut down 2x40mins per sync and the fast steps would still run. Ideally someone would split php.editor into two tiers some time in future.

Requires to label the PRs correctly. Other labels could follow if more categories with long running jobs are found.

This is somewhat similar to how the OpenJDK project works, its just more sophisticated there (a bot manages everything and sets things).

We could also add a full-test label, so that if someone isn't sure or changes something in the platform or API, it would just turn everything on.

mbien: just checked: 
labels are only picked up once when the workflow starts. Restarting jobs won't update them. A sync would trigger a new workflow with updated labels though. Correct labeling would be required *before* creating a PR to conserve resources.

I added a ci:no-build label for PRs which only update readmes etc.

neilcsmith-net: What are the pros / cons of running by label vs running by path?

mbien: path triggers are per-workflow, not per-job. The pipeline does only work within a workflow where artifacts can be shared. So we would have to implement our own condition script which checks paths somehow to conditionally run jobs.

but the main difference: devs set labels. If a reviewer wants all tests to run it could be easily set. Or if tier1 is fine only fast tests are run.

Automating this does rarely work. Calculating the paths would basically trigger everything all the time, see https://github.com/apache/netbeans/pull/3727#issuecomment-1061752817
And what it it doesn't trigger but you want it to trigger? You still need some fallback.

mbien: btw I don't think we should label trigger everything, only jobs which are long enough. The profiler job runs in <2 mins, there is not much to be gained to conditionally run this.

neilcsmith-net: OK, makes sense.  Slight concern at our proliferation of labels and making sure it's well documented so there aren't too many surprises.  Agree automating has issues, although if we want it later we might always automate adding labels! :smile:

mbien: @neilcsmith-net 197e209, 4 category labels so far, for `Java`, `JavaScript`, `Platform` and `PHP`. The label `ci:all-tests` will cause everything to run, `ci:no-build` cancels everything, e.g thats for fixing typos in readmes etc.

mbien:
the gradle test is failing, both locally and on github. I don't know why this test is working on travis. I used the same command lines and tried to replicate everything locally and couldn't find a reason. (even diffed the travis log to a local log)
```
Testcase: testCompilePreTrusted(org.netbeans.modules.gradle.java.classpath.ClassPathProviderImplTest):	FAILED
Closed project should be at least EVALUATED
junit.framework.AssertionFailedError: Closed project should be at least EVALUATED
	at org.netbeans.modules.gradle.java.classpath.ClassPathProviderImplTest.testCompilePreTrusted(ClassPathProviderImplTest.java:264)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/java.lang.Thread.run(Thread.java:829)

```

.github/workflows/main.yml
```diff
@@ -27,65 +27,1282 @@ concurrency:
...
+
+  # Platform label
+  test_platform: ${{ contains(github.event.pull_request.labels.*.name, 'Platform') || contains(github.event.pull_request.labels.*.name, 'ci:all-tests') || github.event_name != 'pull_request' }}
+
+  # Ant, Gradle, Maven and MX labels trigger build-tools job
+  # see job condition (env vars don't work for job conditions https://github.com/actions/runner/issues/1189 )
+
+  # special commands:
+  # ci:all-tests enables everything
+  # ci:no-build disables the build job (and test jobs too)
```
mbien, line 54: all labels

.github/workflows/main.yml
```diff
@@ -27,65 +27,1286 @@ concurrency:
...
```
mbien, line 57: currently used labels which influence CI

.github/workflows/main.yml
```diff
@@ -27,65 +27,1286 @@ concurrency:
...
+      - name: Checkout ${{ github.ref }} ( ${{ github.sha }} )
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          submodules: false
+          fetch-depth: 10
+
+      - name: Print last 10 Commits
+        if: github.event_name == 'pull_request'
+        run: git log --oneline -n10 --pretty=format:'%h %an [%ae] %s'
```
mbien, line 186: the new Paperwork job will print the header of the last 10 commits. This allows reviewers to quickly check if the author info is valid.

.github/workflows/main.yml
```diff
@@ -27,65 +27,1286 @@ concurrency:
...
+
+  # 'Platform' label
+  test_platform: ${{ contains(github.event.pull_request.labels.*.name, 'Platform') || contains(github.event.pull_request.labels.*.name, 'ci:all-tests') || github.event_name != 'pull_request' }}
+
+  # 'Ant', 'Gradle', 'Maven' and 'MX' labels trigger build-tools job
+  # see job condition (env vars don't work for job conditions https://github.com/actions/runner/issues/1189 )
+
+  # special commands:
+  # 'ci:all-tests' enables everything
+  # 'ci:no-build' disables the build job (and test jobs too)
```
mbien, line 57: list of PR labels which influence CI steps/jobs

mbien:
- rebased to latest master
 - made javadoc step conditional (triggered by 'JavaDoc' and 'API Change' labels)

(this build will run everything using 'ci:all-tests' enables everything , want to check out what the total is in hours)

mbien:
another sync with master to resolve conflicts. Rebasing this PR can be tricky since so much changed in the files - easy to overlook something during merge.

matthiasblaesing (commented):
One concern I have is, that the labels are not there, when a PR is created and would need a trigger. Forcing users/reviewers to do a forced push or request one sounds stupid and I remembered, that gitlab had an option to manually trigger CI/CD pipelines and it turns out github has something similar:

https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

This would allow a reviewer add labels and then trigger a run. What do you think?

mbien:
> One concern I have is, that the labels are not there, when a PR is created and would need a trigger. Forcing users/reviewers to do a forced push or request one sounds stupid and I remembered, that gitlab had an option to manually trigger CI/CD pipelines and it turns out github has something similar:
> 
> https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
> 
> This would allow a reviewer add labels and then trigger a run. What do you think?

thanks for taking a look @matthiasblaesing I can always count on you for reviews :)

I fully agree that this is currently the most annoying aspect of this change, I haven't figured out how to solve this though. Here the issues:

 - if a workflow is manually triggered via `workflow_dispatch` it can't access the pr-context since there is none, it would simply run on a branch. So it would see no labels etc, it would also not show up below the PR as checkmark i am afraid (not tested but i can't see how it would)
 - there are various [other triggers](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#available-events) available. E.g it would be possible to trigger if someone adds or removes a label on the PR. but this would be overkill. Due to the limitations of the declarative yaml file I don't know how to trigger only on a certain label (I don't think its currently possible), it would run on every add or remove event.

I haven't given up on that. Ideas welcome.

We have still options though:
- we could let unlabeled PRs run everything instead of the minimum job count, (`ci:all-tests` equivalent) would this be a better default?
- or we could check if it is even a problem in practice: how often does it happen that a new contributor submits a perfect PR which would create the awkward situation of a reviewer having to label the PR and request a sync?
- (I should probably also update the PR template text to mention that PRs should be labeled before pressing create)

neilcsmith-net:
> if a workflow is manually triggered via workflow_dispatch it can't access the pr-context since there is none,

Does re-run all jobs not work after relabelling then?

> we could let unlabeled PRs run everything instead of the minimum job count, (ci:all-tests equivalent) would this be a better default?

Maybe we should start with that as default anyway, but stated as time-limited for a while as people get used to it / feed back concerns?

Either way, make sure to do a mail on dev@ with `[NOTICE]` in the subject when this merges. Then no-one should be complaining they didn't see it! :wink:

mbien:
> > if a workflow is manually triggered via workflow_dispatch it can't access the pr-context since there is none,
> 
> Does re-run all jobs not work after relabelling then?

unfortunately not, that would have been perfect. The workflow takes a snapshot of the context at the beginning it seems. It doesn't see label updates on restarts.


> 
> > we could let unlabeled PRs run everything instead of the minimum job count, (ci:all-tests equivalent) would this be a better default?
> 
> Maybe we should start with that as default anyway, but stated as time-limited for a while as people get used to it / feed back concerns?

I had the same thought too. Let me put all-tests as default. Makes the transition easier

> 
> Either way, make sure to do a mail on dev@ with `[NOTICE]` in the subject when this merges. Then no-one should be complaining they didn't see it! wink

will do! and probably cc some oracle devs right?

There is something i want to try with workflow_dispatch in a testing repo. But as tom cruise would say: we need two miracles in succession. First the manual dispatch would have to actually appear under the PR if its even possible to run them on PR branches. Second: dispatch can use up to 10 input variables on its form, i could try to map them to labels somehow. I hope you can use them in job conditions since there are some things you can't use: e.g env vars for entirely undocumented reasons.

mbien:
update: workflow_dispatch can only run on regular branches (e.g master). You can't select PR branches in the UI. So i can skip testing the rest.

neilcsmith-net:
> > Does re-run all jobs not work after relabelling then?
> 
> unfortunately not, that would have been perfect.

OK, second thought.  Does changing the pull request event types to include `labelled` work? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

mbien:
> > > Does re-run all jobs not work after relabelling then?
> > 
> > 
> > unfortunately not, that would have been perfect.
> 
> OK, second thought. Does changing the pull request event types to include `labeled` work? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request



> > > Does re-run all jobs not work after relabelling then?
> > 
> > 
> > unfortunately not, that would have been perfect.
> 
> OK, second thought. Does changing the pull request event types to include `labeled` work? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

this would trigger a new run every time a label is added which is probably too much since we have also the do-not-merge label etc

mbien:
but what if we abuse `reopened` for this?

neilcsmith-net:
> this would trigger a new run every time a label is added which is probably too much since we have also the do-not-merge label etc

Given this enables many of the tests to eventually not run without required labels, I'm not sure that's a huge issue in practice, and it's certainly more obvious.  If the label trigger works, IMO we could also consider going back to minimal tests by default from the start.

mbien:
> > this would trigger a new run every time a label is added which is probably too much since we have also the do-not-merge label etc
> 
> Given this enables many of the tests to eventually not run without required labels, I'm not sure that's a huge issue in practice, and it's certainly more obvious. If the label trigger works, IMO we could also consider going back to minimal tests by default from the start.

when I read the doc this indeed seemed to be the obvious way how to solve this issue. But this dissipated quickly:
 - labeled events are fired for each label. So if you put 3 labels on a pr it will start 3 workflows + the open event which makes 4. We have the concurrency rule which would shut down all workflows except the newest, but this would still create a lot of noise on the /actions page
 - we use do-not-merge/stale labels for meta-info. This can become a problem when someone accidentally resets a build someone else is looking at
 - when someone would bulk label a bunch of PRs at once via the /pulls page it would simply cause too much building at once IMO

how about using the `unlocked` event as "secret" trigger for the workflow? I just tested it in a sandbox and it worked well
 - I don't think it sends mails (I am not sure, difficult to test with just one account, in fact, let me test this by locking this PR, you can tell me if you got a mail :)), close/reopen does
 - it is less invasive than 'reopened'
 - maybe one day gh will support better ways of doing this. e.g via adding a custom button to the checks or restarting workflows with a fresh context optionally

I assume that most devs will try to label PRs correctly right away or will sync PRs several times anyway during review which gives enough opportunities to update labels. For new contributors and the event of a PR which does not need any changes, reviewers could use this semi-secret trick to trigger a new build after labeling the PR.

neilcsmith-net:
No email that I can see.  All the options seem pretty ugly.  

Is the actual added label in the event?  And if so, easy to filter by prefix `ci:`?

Use `gh` cli to read the PR labels directly rather than from the event?

Use a comment trigger and filter by text (eg. `<RETEST THE DAMN THING>`)???

Yep, all ugly.  From my perspective, go with what you think makes most sense initially.  Can always be tweaked later.  Anyway, _you're_ in charge of explaining any secret triggers! :smile:

.github/workflows/main.yml
```diff
@@ -20,72 +20,1307 @@ name: NetBeans
 on:
   push:
   pull_request:
+    # unlocked event is used as super secret restart button
+    types: [opened, synchronize, unlocked]
```
mbien, line 24: "Lock Conversation" button is on the right panel on PRs. Unlock will start a fresh run without having to sync. I don't think we will need this very often, but its good to have that option.

We are using some fairly new github features, until recently you couldn't restart individual jobs at all. I am sure there will be better ways to implement this one day.

mbien:
ready to integrate from my side. Added the "unlocked" trigger and put a info line into the PR template.

I think its better to integrate it as is without squashing everything. It makes it easier to check if for example there was a mistake the migration of certain jobs without having to go through everything.

matthiasblaesing (approved):
I only eyeballed this and indeed the commits make sense (also to keep them separate). I think we need to "test" this in reality and see how it goes. I like the option to lock/unlock to trigger a build. I'll feel like a "supervillain" when I use the "super secret" restart button.

vieiro (commented):
HI all,
Sorry for a late reply.
I'm afraid I don't fully understand all this, but AFAIU this saves time and energy, to +1 from me.
Since this affects how people approve PRs, maybe we want to add some instructions elsewhere (Confluence?).

mbien:
- updated the label descriptions for labels which influence CI. Description text has been prefixed with [ci] which allows filtering. 
- Added a screenshot with an example to the PR.

full list:
https://github.com/apache/netbeans/labels?q=ci

@vieiro I think it is going to be fairly intuitive. Even if you don't label anything there will be still about 3h of CI time spent (maybe we can reduce this further in future).
This won't be a fine grained configuration, Existing PRs which update maven features are already labeled with "Maven" right now. Once this is merged, the same label will also toggle maven related tests which are aggregated in a build tools job.

mbien:
merging. Going to send the [NOTICE] mail to the dev list too as @neilcsmith-net advised.

Issue #4428, opened 2022-07-22
Include "get", "set", and "is" as element prefixes when doing a Java auto-complete search


SirIntellegence:
### Description

When a user invokes code completion for a member, which may be a "property" (which we shall call "Foo"), check to see if there is any "getFoo", "setFoo", or "isFoo" in the class and add it to the complete list if it is.
The best possibly insertion point I think would be around hither: https://github.com/apache/netbeans/blob/487686125c370140bc39cc8f1d598783f187fdf5/java/java.completion/src/org/netbeans/modules/java/completion/Utilities.java#L102

### Use case/motivation

I am actually porting quite a bit of code from C# to Java, and it would be immensely helpful if Netbeans just checked to see if there is a "Bar.getFoo()" when I invoke autocomplete on "Bar.Foo". It would speed things up quite a bit since I wouldn't have to go to the start, type in the "get", "set", or "is", and then move to the end to complete it.

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
sounds reasonable to me. Looking forward to the PR :)

SirIntellegence:
Will I need to make a test too or can I just submit the change?

mbien:
> Will I need to make a test too or can I just submit the change?

would be good to have test case covering it but the completion tests are a bit netbeans specific, so if you are struggling with it I am sure someone else could help.

I would probably put it in `org.netbeans.modules.java.completion.JavaCompletionTaskBasicTest`. You will probably also have to create a test file for it which would be in `java/java.completion/test/unit/data/org/netbeans/modules/java/completion/data`

Looking at the other test method might help with figuring out how this works.

here an example of a test which added completion for the record keyword:
https://github.com/apache/netbeans/pull/3227/files

SirIntellegence:
Working branch in case I am slow and someone else wants to finish: https://github.com/SirIntellegence/incubator-netbeans/tree/NETBEANS4428

ebarboni:
it may be more manageable to create a PR even with a draft status and do not merge label for people to help.

SirIntellegence:
What is the standard for fixing other unrelated tests in the same test suite? Apparently `System.out.` has a `charset()` which is not accounted for and is failing 3 of the tests

mbien:
there is no need to hurry, NetBeans 16 is still far away.

[...]

PR #4440, opened 2022-07-27
[NETBEANS4428] Initial work on "is/get/set prefixes" bug

SirIntellegence:
It has been implemented. Just need to work on the tests

To resolve issue #4428

### DIFF
--- a/java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
+++ b/java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
@@ -3853,21 +3853,13 @@ public final class JavaCompletionTask<T> extends BaseTask {
                             }
                         } else {
                             if (!options.contains(Options.ALL_COMPLETION) && e.getModifiers().contains(STATIC)) {
-                                if (startsWith(env, sn)
-                                        && (Utilities.isShowDeprecatedMembers() || !elements.isDeprecated(e))
-                                        && isOfKindAndType(((ExecutableType) asMemberOf(e, t, types)).getReturnType(), e, kinds, baseType, scope, trees, types)
-                                        && env.isAccessible(scope, e, t, isSuperCall)
-                                        && (!Utilities.isExcludeMethods() || !Utilities.isExcluded(eu.getElementName(e.getEnclosingElement(), true) + "." + sn))) { //NOI18N
+                                if (checkMethodAccepted(sn, e, t)) { //NOI18N
                                     hasAdditionalMembers = true;
                                 }
                                 return false;
                             }
                         }
-                        return startsWith(env, sn)
-                                && (Utilities.isShowDeprecatedMembers() || !elements.isDeprecated(e))
-                                && isOfKindAndType(((ExecutableType) asMemberOf(e, t, types)).getReturnType(), e, kinds, baseType, scope, trees, types)
-                                && env.isAccessible(scope, e, t, isSuperCall)
-                                && (!Utilities.isExcludeMethods() || !Utilities.isExcluded(eu.getElementName(e.getEnclosingElement(), true) + "." + sn)); //NOI18N
+                        return checkMethodAccepted(sn, e, t); //NOI18N
                     case CLASS:
                     case ENUM:
                     case INTERFACE:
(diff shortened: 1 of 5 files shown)


java/java.completion/test/unit/src/org/netbeans/modules/java/completion/JavaCompletionTaskBasicTest.java
```diff
@@ -1298,4 +1298,9 @@ public void testAfterTypingMethodInvocationInMethodBody() throws Exception {
     public void testAfteMethodInvocationInMethodBody() throws Exception {
         performTest("SimpleMethodBody", 941, null, "methodBodyContentAndLocalVar.pass");
     }
+    public void testPropertyMethods() throws Exception {
+        performTest("ListPropertyMethods", 1170, null, "ListPropertyBool.pass");
+        performTest("ListPropertyMethods", 1199, null, "ListPropertyIntField.pass");
+        performTest("ListPropertyMethods", 1212, null, "ListPropertyIntField.pass");
+    }
```
mbien, line 1305: it might be required to create a test method per performTest call, since there is typically setup and teardown work involved which happens around each test method.

SirIntellegence: Ah, ok

java/java.completion/src/org/netbeans/modules/java/completion/JavaCompletionTask.java
```diff
@@ -3890,6 +3882,33 @@ && isOfKindAndType(e.getEnclosingElement().asType(), e, kinds, baseType, scope,
...
+                    "is" + initialPrefix,
+                    "get" + initialPrefix,
+                    "set" + initialPrefix
+                };
+                for (String prefix : prefixes) {
+                    if (checkMethodAccepted(sn, prefix, e, t)){
+                        return true;
+                    }
+                }
+                return false;
```
mbien, line 3902: alternative which might be more readable:
```java
return checkMethodAccepted(sn, initialPrefix, e, t)
    || checkMethodAccepted(sn, "is" + initialPrefix, e, t)
    || checkMethodAccepted(sn, "get" + initialPrefix, e, t)
    || checkMethodAccepted(sn, "set" + initialPrefix, e, t)
```

I am a little worried about performance here. We recently made completion in cases with large amount of fields faster (#4142).
We will have to check that this is not affecting performance negatively.

Here a file for manual tests: https://github.com/apache/netbeans/issues/4140#issuecomment-1134414691

SirIntellegence: I see... I could possibly take a look at it later... But I don't know if I will have the time. That logic is only invoked for filtering methods though, not fields, but I see your point.

mbien:
before I forget: the commits need to have a real name and valid address in the header.

SirIntellegence:
By header, you mean the commits or the pull request? I will probably need to disable a certain Github option...

mbien:
> By header, you mean the commits or the pull request? I will probably need to disable a certain Github option...

the commits. `git log` of your branch i checked out shows your nickname and a noreply mail address. Those would be the same commits which land in master. This project requires real names and valid addresses for contributions. You can change this per commit, per repository or globally in your git settings.

`git commit --amend --author="Author Name <email@address.com>" --no-edit`
would update the author info of the last commit of the current branch.

SirIntellegence:
That should have done it? Don't know if I need to fiddle with other github settings or not

SirIntellegence:
I just noticed that time stamps on the commits were lost... I am trying to decide if they are important enough to fix...

mbien:
> I just noticed that time stamps on the commits were lost... I am trying to decide if they are important enough to fix...

time stamps going to usually change every time the commit hash changes which is every time the commit itself is modified. PRs like this are usually squashed as last step after review to a single commit before merge anyway so don't worry about those.

the author details look good to me now, thanks.

SirIntellegence:
In regards to optimization, we could us a wrapper `CharSequence` to avoid creating 3 Strings when we check a method. That would require modifications to other methods as well. Another possible solution is to filter through the members in parallel.

mbien:
> In regards to optimization, we could us a wrapper `CharSequence` to avoid creating 3 Strings when we check a method. That would require modifications to other methods as well. Another possible solution is to filter through the members in parallel.

string concatenation is rarely a problem. The JVM became very good at optimizing this. Most String ops are intensified. I wouldn't optimize anything by hand before knowing that this is actually a problem. But I would change it to the version i proposed at https://github.com/apache/netbeans/pull/4440#discussion_r931630020, since it creates the strings lazily/on-demand and is more compact/easier to read as bonus.

Its just that the non-matching case has 4 times more checks than before this PR. Would be good to test and show that this is in fact a non-issue. e.g generate a class with a few thousand methods/fields and compare before and after this PR.

Even if this turns out to be a problem there would be more paths we could take: e.g making this feature configurable/optional.

SirIntellegence:
Is the "Do not merge" tag still applicable?

mbien:
> Is the "Do not merge" tag still applicable?

I think so.
 - I would still like to check that this does not introduce a performance regression since it does 4x more checks per candidate. This might be a non-issue.
 - this would also need more reviewers/opinions to see if this is something what should be enabled by default or via settings. This also somewhat depends on the results of the first bullet point.

SirIntellegence:
So, I tested this file with and without the 3 additional checks and didn't notice any difference (yes, I added methods). My attempt to profile the methods in question didn't seem to work? They never showed up in the results. Even if I used instrumentation. So someone else will probably need to test that. I did notice that going the star import route was slow for both variations.
[FieldsOfJoy10k.java.txt](https://github.com/apache/netbeans/files/9329668/FieldsOfJoy10k.java.txt)

mbien:
> So, I tested this file with and without the 3 additional checks and didn't notice any difference (yes, I added methods). My attempt to profile the methods in question didn't seem to work? They never showed up in the results. Even if I used instrumentation. So someone else will probably need to test that. I did notice that going the star import route was slow for both variations. [FieldsOfJoy10k.java.txt](https://github.com/apache/netbeans/files/9329668/FieldsOfJoy10k.java.txt)

interesting, filed an issue for this #4500

mbien:
@neilcsmith-net @matthiasblaesing would we need a setting in the options for this or is it ok to have this always enabled?

neilcsmith-net:
I suspect this might be the wrong approach, although I haven't looked in depth as yet.  How does it interact with camelcase completion?  eg. typing `bean.sF|` for `bean.setFoo`. My initial thought would be to hook into, and extend, that process.

matthiasblaesing:
Maybe I'm missing something obvious, but from my perspective the issue description is solved by enabling "Subword completion":

image

with that I get this:

image

for this code:

```java
public class Test {
        Test t = new Test();
    }

    private int demoNumber;

    public int getDemoNumber() {
        return demoNumber;
    }

    public void setDemoNumber(int demoNumber) {
        this.demoNumber = demoNumber;
    }
}
```

I don't see reason to make CC more complex.

SirIntellegence:
Ooo. How long was that there? Thanks for letting me know about that. That covers the request. Thanks!

Issue #4415, opened 2022-07-21
Every Java breakpoint is a multi breakpoint


lkishalmi:
### Apache NetBeans version

Apache NetBeans 15 release candidate

### What happened

Just updated to 15-rc1 today and it seems If I put a breakpoint to a Java file it instantly becomes a multi-breakpoint. 
![image](https://user-images.githubusercontent.com/1381701/180146582-751b88d6-a441-44b8-b7b6-b604de62a90c.png)

### How to reproduce

Put a breakpoint in a Java class.

### Did this work correctly in an earlier version?

Apache NetBeans 14

### Operating System

Ubuntu 22.04

### JDK

JDK17

### Apache NetBeans packaging

Apache NetBeans Snap Package

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

neilcsmith-net:
I tried various variations of setting breakpoints but can't replicate this with the rc1 binary zip.  Any interesting logging?  Replicate with clean userdir and cachedir?

lkishalmi:
Tested a bit further. It does not happen with clean user-dir, just with migrated ones. Still testing.

lkishalmi:
Well, find out that it only happens when the Groovy cluster is enabled as the new Groovy Debugger support is eager to add additional Breakpoints into Java files.
Testing further if I could find a fix.

lkishalmi:
Most likely caused by #4153 as the following registration was missing earlier: https://github.com/apache/netbeans/blob/487686125c370140bc39cc8f1d598783f187fdf5/groovy/groovy.debug/src/org/netbeans/modules/groovy/debug/GroovyBreakpointAnnotationListener.java#L42

@entlicher Probaly we shall make that AnnotationListener smarter.

neilcsmith-net:
Thanks @lkishalmi - verified. Having multiple breakpoints also disables the menu for breakpoint properties.  Adding high priority - don't think we can release NB15 with this.

lkishalmi:
I'm working on the fix. It was kind of easy to avoid double registration on Java classes, though there is a bug in the debugger core that prevents breakpoint registration in Groovy files. Fixing that bug, it seems both Truffle and Groovy breakpoints are created.
Now Truffle and Groovy are clashing. I hope I can sort it out soon.

PR #4442, opened 2022-07-28
Make sure, that Java, Groovy and Truffle breakpoints are not clashing

lkishalmi:
Fix the breakpoint clashing issue between Java, Groovy and Truffle debuggers.
This is a fix for #4415.


neilcsmith-net:
Looks sensible, but haven't tested as yet.  Will await review from people who know this area of the codebase better, but will test and merge by Tuesday to hit rc3 if not.

neilcsmith-net:
cc @jtulach @entlicher - appreciate if either of you can take a look by tomorrow prior to merging for rc3 - thanks!

java/debugger.jpda.truffle/src/org/netbeans/modules/debugger/jpda/truffle/actions/ToggleBreakpointsInLanguagesActionProvider.java
```diff
@@ -47,7 +47,7 @@ public class ToggleBreakpointsInLanguagesActionProvider extends ActionsProviderS
     
     private static final Set<String> IGNORED_MIME_TYPES = new HashSet<>(
             // We have JSLineBreakpoint in JavaScript
```
lkishalmi, line 49: This is telling Truffle not to handle Groovy, otherwise both Truffle and Groovy will be registered on the toggle breakpoint action. Due to a "bug" in the Debugger API Truffle wins which then do not place a breakpoint into Groovy.

groovy/groovy.debug/src/org/netbeans/modules/groovy/debug/GroovyBreakpointAnnotationListener.java
```diff
@@ -71,16 +71,22 @@ public void propertyChange (PropertyChangeEvent e) {
     @Override
     public void breakpointAdded (Breakpoint b) {
         if (b instanceof LineBreakpoint) {
-            ((LineBreakpoint) b).addPropertyChangeListener (this);
-            annotate ((LineBreakpoint) b);
+            LineBreakpoint lb = (LineBreakpoint) b;
```
lkishalmi, line 74: Only act on breakpoint which has Groovy stratum. Java files will be handled by the Java JPDA breakpoint handler.

groovy/groovy.debug/src/org/netbeans/modules/groovy/debug/GroovyBreakpointStratifier.java
```diff
@@ -78,7 +79,7 @@ public void stratify(JPDABreakpoint breakpoint) {
                 String printText = pt.replace("{groovyName}", (groovyName != null) ? groovyName : "?"); // NOI18N
                 String groovyPath = getGroovyPath(url, fo);
```
lkishalmi, line 81: Once Truffle out of the way, this stratifier gets called and adds Groovy stratum to the breakpoint.

lkishalmi (commented):
Just added a few describing comment.

mbien (approved):
makes sense to me. This also looks fairly low risk - I don't think this could break anything (famous last words).

neilcsmith-net:
Thanks @mbien - agree with you and intend to merge for rc3, which is delayed elsewhere right now.  Given this issue was caused by #4153 I was hoping for a review from original PR author.  Or reviewers there - cc. @sdedic @ppisl

neilcsmith-net (approved):
Just checked this and fixes the issue with Java files for me.

It still seems that, with or without this, Groovy files get multiple breakpoints.  Which is a regression from NB14.  Maybe we can consider that less a priority?!

lkishalmi:
I've just rebuilt the whole PR. It seems yes Groovy breakpoints are doubled. My ActionManager fix in Debugger API may be required, but give me a few hours so I can fix this.

lkishalmi:
Well, I have the issue, though not sure how to fix.
So even with this PR, when a line breakpoint added to a Groovy file the base JPDA java support places an annotation at:
https://github.com/apache/netbeans/blob/06a31ed66a97a0be7c9c9fe2a2fc0477f7b7e8b9/java/debugger.jpda.projectsui/src/org/netbeans/modules/debugger/jpda/projectsui/BreakpointAnnotationManager.java#L63

Then the Groovy one registers another annotation. That's why the duplicates. @entlicher could tell the truth how to resolve this.

What I could/would do is removing the breakpoint registration from Groovy entirely as it seems JPDA takes care of that.

neilcsmith-net:
@lkishalmi the other option here is to revert #4153 for NB15. Those changes have not been in a release as yet, and it might be safer at this point to revert to the old behaviour and reconsider the original PR and your changes in NB16?

entlicher (approved):

entlicher:
> Well, I have the issue, though not sure how to fix. So even with this PR, when a line breakpoint added to a Groovy file the base JPDA java support places an annotation at:
> 
> https://github.com/apache/netbeans/blob/06a31ed66a97a0be7c9c9fe2a2fc0477f7b7e8b9/java/debugger.jpda.projectsui/src/org/netbeans/modules/debugger/jpda/projectsui/BreakpointAnnotationManager.java#L63
> 
> Then the Groovy one registers another annotation. That's why the duplicates. @entlicher could tell the truth how to resolve this.
> 
> What I could/would do is removing the breakpoint registration from Groovy entirely as it seems JPDA takes care of that.

It is possible that GroovyBreakpointAnnotationListener is not necessary. But this class is there for a long time already, thus I doubt this is a new issue. I've just moved the class to a different package only.

neilcsmith-net:
Thanks @entlicher but are you sure that #4153 is not causing this issue?  At least I cannot reproduce in NB14 or when applying a revert on top of delivery.

entlicher:
> `GroovyBreakpointAnnotationListener` might have been there for a long time, but was it ever used?

That's what came across my mind as well. I've modified the registrations during the refactoring. I'll test the functionality with and without this class...

lkishalmi:
Well added the removal of ```GroovyBreakpointAnnotationListener``` it was never used before due to a missing registration. In my brief tests both Java and Groovy breakpoints are unique after that.

lkishalmi:
I set up a small Gradle project with both Java and Groovy codes. Breakpoints work well in both classes with the removal of ```GroovyBreakpointAnnotationListener```. also the debugger stops on the registered breakpoint either on Java or on Groovy sources.

neilcsmith-net:
Thanks @lkishalmi LGTM  Merging.

Issue #4447, opened 2022-07-28
Implementing folding for specific code blocks in java editor


Chris2011:
### Description

Code folding is a nice thing to collapse lots of code, unfortunately code folding is not possible in Java editor for condition blocks (if, else if, else), loops (do, for, while), switch/case blocks or try/catch blocks but it should as in other languages too.

### Use case/motivation

Having the possibility to collapse/expand condition blocks, loop blocks and try/catch.

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

neilcsmith-net:
@Chris2011 there's no need to create an issue if you're already working on a pull request - it says this on the form, and just adds extra work.

Chris2011:
@neilcsmith-net  Ahh ok, thx for the hint :). Maybe I overread it.

istinnstudio:
useful feature

Chris2011:
Thx, PR is still created, need to change some things first :)

OnyxianSoul:
Any chance you could also implement folding for /* */ style comments? 
It'd be really useful for hiding verbose explanations of what part of the code does or why

Chris2011:
@OnyxianSoul So normal block comments, not documentation comments, right? If yes, I will try.

istinnstudio:
It would be great if it could recognize continuous multi "//" multi  line comments but why not be able to do it on "/**/" commented parts also.
It would be useful if there was a way to fold/hide unfold/show all of them all at once.

Chris2011:
> Any chance you could also include folding for /* */ style comments? It'd be really useful for hiding verbose explanations of what part of the code does or why

I had a look and it is already working. Go into a method, add /* Foo */ somwhere in the code and you can fold it. If you encounter a position, where this is not working, please create a new ticket with explicit steps to reproduce.

Chris2011:
Coming in NB 19

PR #4448, opened 2022-07-28
#4447 - Fix missing code folding blocks in java editor for conditions, loops and try/catch

Chris2011:
The motivation about this was to implement those missing folding blocks in java editor. This was just implemented for imports, classes, inner classes, methods but not for stuff inside methods like if/else if/else, try/catch and loops because all of them are treated like as code blocks.

Please bare with me and my knowledge of the stuff. Unfortunately I dunno why the if statement was implemented, that I removed from JavaElementFoldVisitor but this prevents it from adding folding rectangle to those blocks. This is the only line that I need to change in general but, now I changed the method code block to general code-blocks because now it is not only for methods anymore.

Before:

image

After:

image

It works for the mentioned blocks, but switch is still missing.

### DIFF
--- a/java/java.editor.lib/src/org/netbeans/editor/ext/java/JavaFoldManager.java
+++ b/java/java.editor.lib/src/org/netbeans/editor/ext/java/JavaFoldManager.java
@@ -42,7 +42,9 @@ public abstract class JavaFoldManager implements FoldManager {
     
     public static final FoldType JAVADOC_FOLD_TYPE = JavaElementFoldManager.JAVADOC_FOLD_TYPE;
 
-    public static final FoldType CODE_BLOCK_FOLD_TYPE = JavaElementFoldManager.CODE_BLOCK_FOLD_TYPE;
+    public static final FoldType CODE_BLOCK_FOLD_TYPE = JavaElementFoldManager.METHOD_BLOCK_FOLD_TYPE;
+
+    public static final FoldType METHOD_BLOCK_FOLD_TYPE = JavaElementFoldManager.METHOD_BLOCK_FOLD_TYPE;
     
     public static final FoldType INNERCLASS_TYPE = JavaElementFoldManager.INNERCLASS_TYPE;
     
@@ -68,7 +70,11 @@ public abstract class JavaFoldManager implements FoldManager {
 
     @Deprecated
     public static final FoldTemplate CODE_BLOCK_FOLD_TEMPLATE
-        = new FoldTemplate(CODE_BLOCK_FOLD_TYPE, CODE_BLOCK_FOLD_DESCRIPTION, 1, 1);
+            = new FoldTemplate(CODE_BLOCK_FOLD_TYPE, CODE_BLOCK_FOLD_DESCRIPTION, 1, 1);
+
+    @Deprecated
+    public static final FoldTemplate METHOD_BLOCK_FOLD_TEMPLATE
+            = new FoldTemplate(METHOD_BLOCK_FOLD_TYPE, CODE_BLOCK_FOLD_DESCRIPTION, 1, 1);
 
     @Deprecated
     public static final FoldTemplate INNER_CLASS_FOLD_TEMPLATE
--- a/java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
+++ b/java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
@@ -76,7 +76,11 @@ public class JavaElementFoldManager implements FoldManager {
 	     new org.netbeans.api.editor.fold.FoldTemplate(3, 2, "/**...*/")); // NOI18N
 
     @NbBundle.Messages("FoldType_Methods=Methods")
-    public static final FoldType CODE_BLOCK_FOLD_TYPE = FoldType.MEMBER.derive("method", Bundle.FoldType_Methods(), 
+    public static final FoldType METHOD_BLOCK_FOLD_TYPE = FoldType.MEMBER.derive("method", Bundle.FoldType_Methods(), 
+            new org.netbeans.api.editor.fold.FoldTemplate(1, 1, "{...}")); // NOI18N
+
+    @NbBundle.Messages("FoldType_CodeBlocks=Code Blocks")
+    public static final FoldType CODE_BLOCK_FOLD_TYPE = FoldType.CODE_BLOCK.derive("code-block", Bundle.FoldType_CodeBlocks(), 
 	     new org.netbeans.api.editor.fold.FoldTemplate(1, 1, "{...}")); // NOI18N
     
     @NbBundle.Messages("FoldType_InnerClasses=Inner Classes")
@@ -271,6 +275,11 @@ public class JavaElementFoldManager implements FoldManager {
                 @Override
                 public FoldInfo createInnerClassFold(int start, int end) {
                     return FoldInfo.range(start, end, INNERCLASS_TYPE);
+                }
+
+                @Override
+                public FoldInfo createMethodFold(int start, int end) {
+                    return FoldInfo.range(start, end, METHOD_BLOCK_FOLD_TYPE);
                 }
 
                 @Override
--- a/java/java.editor.base/src/org/netbeans/modules/java/editor/base/fold/JavaElementFoldVisitor.java
+++ b/java/java.editor.base/src/org/netbeans/modules/java/editor/base/fold/JavaElementFoldVisitor.java
@@ -32,6 +32,7 @@ import com.sun.source.tree.ImportTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
+import com.sun.source.tree.Tree.Kind;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.SourcePositions;
 import com.sun.source.util.TreePath;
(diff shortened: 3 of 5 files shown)


mbien:
wanted to suggest to add a test case for it but the code folding tests don't work anyway and probably haven't been run for a long time, so I suppose its fine ;)

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
```diff
@@ -75,8 +75,8 @@ public class JavaElementFoldManager implements FoldManager {
     public static final FoldType JAVADOC_FOLD_TYPE = FoldType.DOCUMENTATION.derive("javadoc", Bundle.FoldType_Javadoc(), 
 	     new org.netbeans.api.editor.fold.FoldTemplate(3, 2, "/**...*/")); // NOI18N
 
-    @NbBundle.Messages("FoldType_Methods=Methods")
-    public static final FoldType CODE_BLOCK_FOLD_TYPE = FoldType.MEMBER.derive("method", Bundle.FoldType_Methods(),
```
sdedic, line 79: Please do not **rename** an existing fold type -- this will break user settings. Instead, add an additional FoldType  - derive from nothing or FoldType.CODE_BLOCK.

Then adjust the code to use the new fold type for blocks, fors, ...

Such setup allows the user to control inner code folds and method-level folds separately, i.e. if one sets methods to be collapsed by default, the code blocks won't be initially collapsed when the user expands a method element.

sdedic (changes requested):

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
Chris2011: You are completly right, I thought so too. I will change this to Draft and will work on it further. Thx.

Chris2011: Here is what I did:
* Fixed the visitBlock method and add check for ignoring method because visitMethod is already available
* Change code for visitMethod similar to visitClass but just for methodfolding
* Introduce new option for code block

Chris2011:
Ok, I see the failing build is related to my code changes. Will try to fix them first.

Chris2011:
So everything is back green. @sdedic can you please have a look again :)

mbien:
lets wait for #4931 before merging this. Since this PR enables the tests of `java.editor`

Chris2011:
Sure, thx for the info.

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
```diff
@@ -76,7 +76,11 @@ public class JavaElementFoldManager implements FoldManager {
 	     new org.netbeans.api.editor.fold.FoldTemplate(3, 2, "/**...*/")); // NOI18N
 
     @NbBundle.Messages("FoldType_Methods=Methods")
-    public static final FoldType CODE_BLOCK_FOLD_TYPE = FoldType.MEMBER.derive("method", Bundle.FoldType_Methods(), 
+    public static final FoldType METHOD_BLOCK_FOLD_TYPE = FoldType.MEMBER.derive("method", Bundle.FoldType_Methods(), 
+            new org.netbeans.api.editor.fold.FoldTemplate(1, 1, "{...}")); // NOI18N
+
+    @NbBundle.Messages("FoldType_CodeBlocks=Code Blocks")
```
sdedic, line 82: Please also change the reference in `java.editor.lib` - class `org.netbeans.editor.ext.java.JavaFoldManager`. This is a deprecated support for ancient modules that reference symbolically `CODE_BLOCK_FOLD_TYPE` (uses implementation dependency on `java.editor`) - otherwise the old modules would change their folding semantics.

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
```diff
@@ -273,6 +277,11 @@ public FoldInfo createInnerClassFold(int start, int end) {
                     return FoldInfo.range(start, end, INNERCLASS_TYPE);
                 }
 
+                @Override
+                public FoldInfo createMethodsFold(int start, int end) {
```
sdedic, line 281: I know that it may be awkward - given the 'createCodeBlockFold` really does not describe 'method-level' folds well (although it has been its purpose) -- but since there are known implementation dependencies, maybe the `Creator`'s method could rremain as it is - and the **real** code blocks could get a new name ?

cc @dbalek -- what's your opinion ? Impl dependencies usually do not deserve any compatibility attention (especially if this PR refactors all known ones in the codebase), but still - this support incl. the misnamed code block (both constant and the Creator method) is used for long time by plugins.

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
Chris2011: I added it as requested. Hope this was the right and only place where I need to add it.

java/java.editor/src/org/netbeans/modules/java/editor/fold/JavaElementFoldManager.java
Chris2011: I already thought about this. I mean a method is also a code block as you said. But yeah at the end I'm fine with this and not to add for really each case a method to have also an option for if, for for, etc.

java/java.editor.lib/src/org/netbeans/editor/ext/java/JavaFoldManager.java
```diff
@@ -43,6 +43,8 @@ public abstract class JavaFoldManager implements FoldManager {
     public static final FoldType JAVADOC_FOLD_TYPE = JavaElementFoldManager.JAVADOC_FOLD_TYPE;
 
     public static final FoldType CODE_BLOCK_FOLD_TYPE = JavaElementFoldManager.CODE_BLOCK_FOLD_TYPE;
```
sdedic, line 45: Maybe I did not express the concerns well: The point here is that before the change the user of `JavaFoldManager.CODE_BLOCK_FOLD_TYPE` gets a `FoldType` instance whose `code() == "method"`. That affects e.g. what setting controls the fold type's default expansion state. The user may have `fold.method.defaultCollapsed=true` (just example name) in his settings.

Now with this PR mergd, that unchanged fold producing code that link to `JavaFoldManager.CODE_BLOCK_FOLD_TYPE` will start to receive a `FoldType` with 'code() == "code-block"` ... which means the saved settings won't have suddenly the same effects.

By changing the (potentially) shared definition of `CODE_BLOCK_FOLD_TYPE`, we're breaking settings / semantic compatibility. Yes, it's very frustrating, as a constant named CODE_ was misused for actually method level folds...

dbalek: Not sure what is better -  to preserve the backward semantic compatibility and potentially cause a confusion from non intuitive names or otherwise.

Chris2011: Ok, I understand the concerns. Do we have some numbers how many people are still using java.editor.lib? I mean isn't it correct to remove deprecated stuff after years of new releases? Since when is the java.editor.lib deprecated?

Chris2011: Is there anything what I can do more? Honestly I don't want to decide, because I don't know how much developers are still using the old java.editor.lib but I mean as I mentioned before, it is deprecated. So I don't see a problem there to change this and to make an help article or smth else to show them the alternative when they want to update. Just my 2 cents.

sdedic: @Chris2011 - even though APIs get deprecated, I don't remember a situation when an API was actually *dropped*, breaking backward compatibility - at least not in such (semantically) 'central' module as java editor. In the past some APIs were broken (see that many modules have /1 or /2 suffixes, which indicates a breaking change), but even during "headless split" (cca NetBeans 8.2) we did not actually drop APIs: see the `*.compat` modules that contain compatibility layer for old applications. Sadly the (deprecated) `java.editor.lib` is such official API with 'forever' commitment.

But I have made a mistake in my reasoning ! If I suppose that an ancient unmantained and useful module still uses the old `editor.lib` APIs, it won't ship that library. So if we leave the (deprecated, misnamed) `JavaFoldManager.CODE_BLOCK_FOLD_TYPE` and relay it to the new, proper-named (but nonpublic) JavaElementFoldManager.METHOD_BLOCK_FOLD_TYPE.  So the ancient client links to the misnamed symbol - but from a 'new version of deprecated java.editor.lib`, thus getting the fold still labeled "method". No semantic change, no settings change. And NB codebase can now use properly named `CODE_BLOCK` and `METHOD_BLOCK`.  Old code lives with past sins, new code uses proper names and even can correctly read the old settings !

Re FoldCreator: the 'createCodeBlockFold` seems to be also now called for initializers and static initializers: we should preserve the fold behaviour here IMHO - that an initializer's fold is controlled by method's settings - but then `createMethodFold` would be awkward for an initializer - wouldn't it ? BTW - isn't there a typo ? The other `FoldCreator` methods use singular nouns but `createMethodsFold` uses plural.

Chris2011: So I understand your point of backward compatibilty. For me I just want to know exactly, what I need to do to get this PR merged. As far as I understand you correct, I need to change some code to map some behavior from the generic code_block_fold_type to the more specific method_block_fold_type, didn't I do this?

mbien:
@Chris2011 would be good to rebase this PR so we get a fresh CI run on the new config.

Should we move this to NB18? what do you think?

Chris2011:
> @Chris2011 would be good to rebase this PR so we get a fresh CI run on the new config.
> 
> Should we move this to NB18? what do you think?

We can move this to NB18, but I'm still waiting for feedback what I should do now. I dunno whether I still should change smth or we need to have a discussion about this. I dunno unfortunately.

Chris2011:
Hey @sdedic for now I hope that I understood what you mean so I changed the mapping in the deprecated Java Editor Lib to the method block and hope that this is what you requested.

Chris2011:
To test the old implementation, if necessary, do anyone know an old Platform Application, using the old Java Editor Lib to test my stuff?

Chris2011 (commented):
I thought I already resolved the requested changes, but I didn't. Sry for that.

java/java.editor.lib/src/org/netbeans/editor/ext/java/JavaFoldManager.java
Chris2011: @sdedic hopefully resolved.

sdedic (approved):

mbien (approved):
code looks good to me.

unfortunately there don't seem to be any tests for code folding in general but this isn't your fault.

PR #4454, opened 2022-07-30
Tomcat module cleanup

pepness:
Cleanup:
- Remove unnecessary if/return statement
- Add missing braces
- Add missing `@Override` annotations
- Use diamond inference
- Use try-with-resources
- Use StandardCharsets instead of a String
- String concatenation in logger
- Use multi-catch and a more specific catch
- Improve Collection.toArray
- Length one String literal in String.indexOf
- Looks for string concatenation in StringBuilder
- Replace StringBuffer with StringBuilder when not necessary

Testing:
- Full build done
- Verify successful execution of Sigtests
- Verify successful execution of commit-validation
- Verify successful execution of rat-exclusions
- Verify successful execution of `enterprise.tomcat5` unit tests

After this, I plan to improve Tomcat/TomEE support for Jakarta EE 9/9.1

### DIFF
--- a/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/TomcatFactory.java
+++ b/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/TomcatFactory.java
@@ -261,25 +261,16 @@ public final class TomcatFactory implements DeploymentFactory {
             if (!bootstrapJar.exists()) {
                 return null;
             }
-            try {
-                JarFile jar = new JarFile(bootstrapJar);
-                try {
-                    Manifest manifest = jar.getManifest();
-                    String specificationVersion = null;
-                    if (manifest != null) {
-                        specificationVersion = manifest.getMainAttributes()
-                                .getValue("Specification-Version"); // NOI18N
-                    }
-                    if (specificationVersion != null) { // NOI18N
-                        specificationVersion = specificationVersion.trim();
-                        return getTomcatVersion(specificationVersion, TomcatVersion.TOMCAT_55);
-                    }
-                } finally {
-                    try {
-                        jar.close();
-                    } catch (IOException ex) {
-                        LOGGER.log(Level.FINEST, null, ex);
-                    }
+            try (JarFile jar = new JarFile(bootstrapJar)) {
+                Manifest manifest = jar.getManifest();
+                String specificationVersion = null;
+                if (manifest != null) {
+                    specificationVersion = manifest.getMainAttributes()
+                            .getValue("Specification-Version"); // NOI18N
+                }
+                if (specificationVersion != null) { // NOI18N
+                    specificationVersion = specificationVersion.trim();
+                    return getTomcatVersion(specificationVersion, TomcatVersion.TOMCAT_55);
                 }
             } catch (IOException e) {
                 LOGGER.log(Level.FINE, null, e);
--- a/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/TomcatFactory.java
+++ b/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/TomcatFactory.java
@@ -97,7 +97,7 @@ public final class TomcatFactory implements DeploymentFactory {
     private static final String DISCONNECTED_URI_80 = TOMCAT_URI_PREFIX_80 + "apache-tomcat-8.0.x";   // NOI18N
     private static final String DISCONNECTED_URI_90 = TOMCAT_URI_PREFIX_90 + "apache-tomcat-9.0.x";   // NOI18N
     
-    private static final Set<String> DISCONNECTED_URIS = new HashSet<String>();
+    private static final Set<String> DISCONNECTED_URIS = new HashSet<>();
     static {
         Collections.addAll(DISCONNECTED_URIS, DISCONNECTED_URI_50,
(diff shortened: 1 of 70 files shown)


enterprise/tomcat5/src/org/netbeans/modules/tomcat5/ui/wizard/AddInstanceIterator.java
```diff
@@ -224,14 +224,14 @@ private void checkStartupScript(File homeDir) {
                                 FileUtil.copyFile(
                                     FileUtil.toFileObject(new File(bundledHome, "bin/" + CATALINA)), // NOI18N
                                     FileUtil.toFileObject(new File(homeDir, "bin")),    // NOI18N
-                                    CATALINA.substring(0, CATALINA.indexOf("."))    // NOI18N
+                                    CATALINA.substring(0, CATALINA.indexOf('.'))    // NOI18N
```
mbien, line 227: `indexOf` is the last remaining String method which has the char variant outperforming the single-char-string variant (under some circumstances).

But this is more of a performance bug of the JDK/JVM and highly depends of the length of the string which is being checked (String variant can be sometimes faster! The correct behavior would be identical performance). 

I don't think we should do those kind of refactoring since: 
 - firstly, they should perform the same if the JVM does its job correctly, 
 - secondly, right now, String going to outperform the char variant in some situations, so its a de-optimization ;)
 - thirdly, the difference is so tiny that it would only show up in measurements when used in inner loops without any other bottle necks

We should probably add a note to the description of the code inspection that it is a bit outdated (just like when IDEs suggested the wrong version of toArray() for so long).

I tried to fix it in the JDK but ended up running out of time when I noticed that the initial assumption that char variants are faster was wrong due to another performance bug:
https://github.com/openjdk/jdk/pull/6509
https://github.com/openjdk/jdk/pull/6509#issuecomment-980681069 (string variant faster than char bug)

next winter maybe

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/j2ee/JaxWsStack.java
```diff
@@ -169,21 +169,17 @@ private String[] getDetectedMetroLibs() {
 
     
     private boolean isKeystore() {
-        if (new File(catalinaHome, KEYSTORE_LOCATION).exists()) return true;
-        else return false;
+        return new File(catalinaHome, KEYSTORE_LOCATION).exists();
```
mbien, line 172: nitpick: could be `Files.exists(Paths.get(catalinaHome, KEYSTORE_LOCATION))`

but we could do this later, going to add this pattern to the list here: https://github.com/mbien/jackpot-inspections :)

mbien (approved):
good cleanup, left some remarks for you to consider.

I was also wondering if some of the single method anonymous classes which got their missing overrides added could be directly converted to lambdas where the override wouldn't be needed anymore.

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/ui/wizard/AddInstanceIterator.java
pepness: You are right the difference should be negligible when it is patched.

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/TomcatFactory.java
```diff
@@ -227,15 +227,7 @@ public static String getTomcatVersionString(File catalinaHome) throws IllegalSta
...
-        } catch (MalformedURLException e) {
-            throw new IllegalStateException(e);
-        } catch (ClassNotFoundException e) {
-            throw new IllegalStateException(e);
-        } catch (NoSuchMethodException e) {
-            throw new IllegalStateException(e);
-        } catch (InvocationTargetException e) {
-            throw new IllegalStateException(e);
-        } catch (IllegalAccessException e) {
+        } catch (MalformedURLException | ClassNotFoundException | NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
```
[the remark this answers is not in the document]
pepness: Thank you.

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/j2ee/JaxWsStack.java
pepness: I agree, we should do a io to nio in a new PR. Will look into that.

mbien: this jackpot rule should cover this particular case:
```java
new java.io.File($parent, $child$).exists() :: $parent instanceof String
=> java.nio.file.Files.exists(java.nio.file.Path.of($parent, $child$))   :: sourceVersionGE(11)
=> java.nio.file.Files.exists(java.nio.file.Paths.get($parent, $child$)) :: otherwise
;;

new java.io.File($uri).exists() :: $uri instanceof java.net.URI
=> java.nio.file.Files.exists(java.nio.file.Path.of($uri))   :: sourceVersionGE(11)
=> java.nio.file.Files.exists(java.nio.file.Paths.get($uri)) :: otherwise
;;
```

mbien:
upstream commits show up in this PR now. This usually happens when they are pulled from a different branch without setting the --rebase flag.
Its probably better to move your commits on a fresh copy of master before integration.

PR #4456, opened 2022-08-01
Treat Gradle disk cache at least as evaluated.

lkishalmi:
This patch treats the on-dosk Gradle cache data as EVALUATED even if invalid. Usually there is a good chance that the information there is still better, than we can get from the heuristics.

This one would make ```ClassPathProviderImplTest.testCompilePreTrusted``` happy.

I hope this is not too late for NB15. Please review!


neilcsmith-net:
Not too late for rc3 if we can merge tomorrow.  cc @sdedic !

Aside from `ClassPathProviderImplTest.testCompilePreTrusted`, when is this problem likely to hit an end user?

mbien (approved):
can confirm that the test is working for me now locally.

I temporarily added it to #4431 to get a test run on github too. Build is in progress. will update this post once done:
https://github.com/apache/netbeans/actions/runs/2774333178

github job passed too:
https://github.com/apache/netbeans/runs/7610265837?check_suite_focus=true

both instances, github and local didn't work at all before this PR, no matter how often retried.

sdedic (approved):

lkishalmi:
Well, the end user may not even notice the difference that this fix provides. I'd say some refactoring things in a partially opened Gradle multi-project tree would be more robust.

sdedic:
IMHO if the user uses goto type to open a source from such 'degraded' project, library dependencies could be loaded from the cache and symbols in the source could resolve much better. The same for any other navigation that targets souces from such project.

PR #4459, opened 2022-08-02
Modernize ActionsManager in Debugger API

lkishalmi:
Well, last week I had a chance to spend a few hours debugging the ActionsManager in the debugger API.
It was not a delightful experience. The code was written in ancient Java, sometimes it was hard to find out what does it want to do. So for make the debugging work easier, I've translated some of those ancient Java to a bit modern dialect.

After all I've found a small bug at https://github.com/apache/netbeans/blob/06a31ed66a97a0be7c9c9fe2a2fc0477f7b7e8b9/ide/api.debugger/src/org/netbeans/api/debugger/ActionsManager.java#L317 where the case ```p1.equals(p2)``` was not handled well. But that it is easy to avoid to hit that codepath, so I did so in #4442, so the fix in the ActionManager was not required.

In order my translation efforts on the ActionsManager not to be vain, I've finished the translation of the whole file and put it here for review.

### DIFF
--- a/ide/api.debugger/src/org/netbeans/api/debugger/ActionsManager.java
+++ b/ide/api.debugger/src/org/netbeans/api/debugger/ActionsManager.java
@@ -22,14 +22,13 @@ package org.netbeans.api.debugger;
 import java.beans.*;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
-import java.util.Vector;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


mbien (approved):
good cleanup, couldn't find anything wrong with it - looks much better now.

just a nitpick:
I personally wouldn't use LinkedLists in new code anymore. ArrayLists are friendlier to the GC and will win most benchmarks (iteration, copy, addAll etc). This usecase here also does not remove elements while iterating or adds elements in front so AL should do fine.

lkishalmi:
@mbien Well, that's true about the ArrayList, I'd expect 1-2 elements of this returned list, that's why I choose the Linked one.

Telling the truth, I would get rid of the whole method. When it is called, it means that we have two debugger support listening on the same action causing trouble like #4442. Instead of trying to resolve that issue with the heuristic applied by the path. I would pick the first one (should be ordered by the lookup position property) and write a warning message on the console. 

That behavior could have saved me a few hours

PR #4461, opened 2022-08-02
Support for vulnerability audit in Oracle cloud

sdedic:
This PR adapts code developed by @ppisl  and makes necessary fixes to the non-public Project Dependency API. Persistence of the "audit knowledgebase" is out of scope of this PR, project-persistent settings can be added later. The implementation provides `ErrorCollector` that marks identified vulnerabilities as warnings / errors so they can be collected by a LSP client.

Diagnostic code URL (should point to an explanation or documentation) is added to LSP API.

The audit implementation watches for dependency model changes, which typically happens after file save - project reload.


java/java.lsp.server/nbcode/integration/nbproject/project.xml
```diff
@@ -76,13 +93,39 @@
...
+                    </run-dependency>
+                </dependency>
                 <dependency>
                     <code-name-base>org.netbeans.modules.java.lsp.server</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
-                        <specification-version>2.0</specification-version>
+                        <implementation-version/>
```
dbalek, line 110: Instead of introducing new implementation dependencies, better longer term solution would be to crate `CommandProvider` interface inside the `api.lsp` module and implement the `ProjectAuditCommand` directly as a part of `cloud.oracle` module

dbalek (approved):

java/java.lsp.server/nbcode/integration/nbproject/project.xml
sdedic: Understood: https://issues.apache.org/jira/browse/NETBEANS-6518. I would like to introduce some (de)serialization basics, so that all modules do not need to use Jackson or GSON for trivial data structures.

ppisl (approved):

neilcsmith-net:
Please check Travis is green before merging - at least the first fail looks legit.

jhorvath (approved):

sdedic:
Real failurex fixed, merging.

PR #4462, opened 2022-08-02
Update junixsocket 2.4.0 -> 2.5.1 (API incompatible!)

matthiasblaesing:


neilcsmith-net:
Thanks @matthiasblaesing So much for minor increments! :smile: 

On the sigtests front it's obviously failing on gen-sigtests which is going to cause problems.  We can override the target in `ide/libs.c.kohlschutter.junixsocket/build.xml` for now?

```xml
<target name="gen-sigtest">
        <!-- Temporary workaround for sigtest generation failures -->
        <antcall target="projectized.gen-sigtest">
            <param name="sigtest.gen.fail.on.error" value="false"/>
        </antcall>
</target>
```

matthiasblaesing:
@neilcsmith-net thank you for the suggestion. I integrated it and added a comment what goes wrong. The parsing failure of the sigtest is just part of the problem. There are also philosophical problems, where classes that are package private are used as return values or parameters. sigtest does not like these types. My gut feeling is, that sigtest is right, but then it is outside our direct control.

neilcsmith-net:
Thanks. Associated comment looks good. You're right, exposing package private classes in that way is problematic, but also not a lot we can do about it. How we handle sigtest and API stability with exposed third-party libs might be something to look at.

Will wait for the (unrelated) test to go green and merge.

Issue #4299, opened 2022-06-30
Commit validation tests fail on 19ea


mbien:
The tests were running in headless mode until #4233, while switching the xserver on, I noticed that the commit validation tests started to fail on 19ea (locally reproducible too) (11, 17, 18 works fine).

`org.netbeans.modules.ide.kit.VerifySimpleTemplatesTest` fails on 19ea with:
```
    [junit] An error occurred during parsing of '/home/runner/work/netbeans/netbeans/ide/ide.kit/build/test/qa-functional/work/o.n.m.i.k.V/at/Test41.java'. Please report a bug against java/source and attach dump file '/home/runner/work/netbeans/netbeans/ide/ide.kit/build/test/qa-functional/work/userdir0/var/log/Test41.dump'.
    [junit] Caused: java.lang.NullPointerException: Cannot read field "tsym" because "imp.qualid.selected.type" is null
    [junit] 	at com.sun.tools.javac.comp.Check.checkImportedPackagesObservable(Check.java:4031)
    [junit] 	at com.sun.tools.javac.comp.TypeEnter.finishImports(TypeEnter.java:229)
    [junit] 	at com.sun.tools.javac.comp.TypeEnter.complete(TypeEnter.java:214)
    [junit] 	at com.sun.tools.javac.code.Symbol.complete(Symbol.java:682)
    [junit] 	at com.sun.tools.javac.code.Symbol$ClassSymbol.complete(Symbol.java:1410)
    [junit] 	at com.sun.tools.javac.comp.Enter.complete(Enter.java:610)
    [junit] 	at com.sun.tools.javac.comp.Enter.main(Enter.java:587)
    [junit] 	at com.sun.tools.javac.main.JavaCompiler.enterTrees(JavaCompiler.java:1042)
    [junit] 	at com.sun.tools.javac.api.JavacTaskImpl.enter(JavacTaskImpl.java:345)
    [junit] 	at com.sun.tools.javac.api.JavacTaskImpl.enter(JavacTaskImpl.java:282)
    [junit] 	at org.netbeans.modules.java.source.parsing.JavacParser.moveToPhase(JavacParser.java:747)
    [junit] 	at org.netbeans.modules.java.source.parsing.JavacParser.getResult(JavacParser.java:540)
    [junit] 	at org.netbeans.modules.java.source.parsing.JavacParser.getResult(JavacParser.java:141)
[...]

PR #4467, opened 2022-08-04
Update nb-javac to 19+33.

dbalek:
Updating nb-javac to version 19+33.


mbien (changes requested):
awesome! please update:
 - version check to 19 as in #3953 line: https://github.com/apache/netbeans/pull/3953/files#diff-4acb756ec7378fd25f5bc06c331506df4d42d67d21b4356f8c90ff35dfaa1f75R37
 - the module description as in #4010
 - ignored-overlaps file as seen in #4106 https://github.com/apache/netbeans/pull/4106/files#diff-afa15c4cdb24d884c580b41d4e36504eb2576b904623433f15d385cf784be6a6 (i don't know what this file does, but it id going to have the wrong version in it)
 - bump this line to 20 to enable testing on JDK 19 with xserver enabled https://github.com/apache/netbeans/blob/6155cbf31a59545ba2f2a66521449ea2e388e713/.github/workflows/main.yml#L106 (#4299)

would be good to have everything in one PR so that we can use it as template and won't forget things anymore.

neilcsmith-net (dismissed):
Obviously the binaries need distributing via Maven and the binaries-list files updating rather than deleting before this can be merged.

Is there a reason for the PR in current state?  Be good to get the binaries sorted before using Travis and GitHub CI overhead for this.

dbalek:
> Obviously the binaries need distributing via Maven and the binaries-list files updating rather than deleting before this can be merged.
> 
> Is there a reason for the PR in current state? Be good to get the binaries sorted before using Travis and GitHub CI overhead for this.

This is just a temporary state (see do-not-merge badge) to check NB build and tests with the new nb-javac binary before pushing it to Maven. Will be changed to updating binaries-list later.

java/java.source.base/src/org/netbeans/modules/java/source/transform/ImmutableTreeTranslator.java
```diff
@@ -1429,18 +1502,43 @@ private DefaultCaseLabelTree rewriteChildren(DefaultCaseLabelTree tree) {
...
+	VariableTree newVariable = (VariableTree)translate(tree.getVariable());
+	if (newDeconstructor != tree.getDeconstructor() || newVariable != tree.getVariable() || !Objects.equals(newNestedPatterns, tree.getNestedPatterns())) {
+	    DeconstructionPatternTree n = make.DeconstructionPattern(newDeconstructor, newNestedPatterns, newVariable);
+            model.setType(n, model.getType(tree));
+	    copyCommentTo(tree,n);
+            copyPosTo(tree,n);
+	    tree = n;
+	}
+	return tree;
     }
```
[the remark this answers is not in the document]
dbalek: Fixed.

neilcsmith-net:
I assumed so - thanks for the updated description.  You can use URLs in `binaries-list` files now too, which might be better for testing.  This may or may not get properly tested now - at least discussion in #4431 might lead to this failing before it gets to any tests.

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -2933,36 +2959,24 @@ private boolean handleSwitch(Tree node, Void p) {
...
-                                break;
+                    for (Iterator<? extends CaseLabelTree> it = labels.iterator(); it.hasNext();) {
+                        CaseLabelTree label = it.next();
+                        scan(label, p);
+                        if (it.hasNext()) {
+                            spaces(0);
+                            accept(COMMA);
+                            space();
                         }
                     }
```
mbien, line 2977: since it doesn't have the `removeWhiteSpace` calls anymore, this basically means it will format switch-cases less aggressively now?

mbien (approved):
looks good to me as far as I can tell - thanks for updating all the version checks and descriptions.

I only left a comment for a small refactoring opportunity I noticed and a question.

java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
dbalek: Surely not. Redundant whitespace is removed as a part of `spaces(0)` and `space()` calls.

jlahoda (approved):
Seems reasonable.

dbalek:
@neilcsmith-net: Changes requested are IMHO done, do you want to update your review before merge ?

mbien:
i think we have to update javac again before NB 16 release, right? 19 GA is 19+36, we have 19+33 right now.

PR #4471, opened 2022-08-07
Updating macOS installer process to sign libjunixsocket-native-2.5.1.dylib

mcdonnell-john:
Adding libjunixsocket-native-2.5.1.dylib to the list of native binaries that are signed when creating the macOS installer


neilcsmith-net (approved):

neilcsmith-net:
Thanks.  Change looks good.  If we have no other need for a 15-rc4, might consider merging this to master and mentioning the required alteration in the installer vote thread?  We've done similar once before.

Longer term, might be good to wildcard this? Along lines of https://github.com/apache/netbeans-tools/blob/master/nbpackage/src/main/java/org/apache/netbeans/nbpackage/macos/MacOS.java#L31 although using `find` or similar rather than Java implementation of.

PR #4472, opened 2022-08-07
Make ANTLRv4 generated Lexer integration into NetBeans easy.

lkishalmi:
Well working on some ANTLRv4 stuff lately, for upcoming TOML support.  Noticed that I need an implementations of CharStream over LexerInput and that, ANTLR Lexer is used as just convert some token type (that is an ```int``` in ANTRL4) to a ```TokenId```

The second commit is a kind of demonstration of the usage of this new module. A candidate could be our CSS lexer, though that seems to be on ANTLRv3 at the moment. I would use this module on TOML support. (I'm working on something for Terraform (HCL) as well.)  Maybe a reduced dumb grammar for Gradle DSL instead if our Groovy...

Also it might be good to try to move the ANTLRv3 stuff to ANTLRv4...


matthiasblaesing:
In general I refrain from creating a library if there is no clear advantage. The problem with libraries and NetBeans "we never break API" attitude is, that you set thinks in stone, that you might regret. If the code is copied you pay by maybe doing work twice, but on the other hand you gain bigger freedoms. Not sure where I'd make the tradeoff, but two use sites seems not enough for the introduced dependency. 

> Also it might be good to try to move the ANTLRv3 stuff to ANTLRv4...

Any reason to do so? The dependencies of antlr are more or less trivial and it does not tie into the JVM, so I don't expect problems in the future.

lkishalmi:
Well, I may was a bit too eager with this. Let's put this rest now.

PR #4476, opened 2022-08-07
Gradle project reports unresolved configurations.

lkishalmi:
Well, this is a bit oversight with the recent changes around how we handle configurations in Gradle. We started to include non-resolvable configurations in the IDE around NetBeans 14. That would give better information available, however no matter what the Configurations node would display a warning badge about there are some configurations left to be unresolved.

It turned out that we forgot to filter the non-resolvable configurations out of the configurations when we are checking whether a configuration is resolved or not.

Would be a nice trivial fix for NB15, However I hope there would be no more RC-s there.


sdedic (approved):
Great - thanks.

neilcsmith-net:
> Would be a nice trivial fix for NB15, However I hope there would be no more RC-s there.

Please change the base to delivery and mark it with NB15, then at least it's ready to go in if we do need another RC build.

Issue #4305, opened 2022-06-30
CSS position:sticky support


excolo-rick:
### Description

The CSS position property allows for a value of `sticky`. Major browsers have supported this for up to a decade.

Netbeans' CSS validator objects to this value.

Please add support to the CSS validator.

### Use case/motivation

Using CSS in a project, and use of `position: sticky` is flagged as "unexpected charactes found," giving warnings where none is relevant.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow the Apache Software Foundation's [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html)

matthiasblaesing:
Fix was merged via #4480

PR #4480, opened 2022-08-08
CSS: Implemented position:sticky

troizet:
Hello! In this PR I tried to implement `position:sticky` from issue #4305, and also make a correct help for this property from the css3 specification.


matthiasblaesing (approved):
Change looks sane to me and manual test was successful. Once the unittests come back all green (still running at this point), I'll merge.

Given that you already contributed, you know the drill: I assume, that you read the license headers and also accept, that you donate the changes to the ASF.

PR #4485, opened 2022-08-08
Add HTML rule and hint+fix for missing alt attribute for tag img, applet and area

Chris2011:
This PR adds a new HTML Rule, Hint and HintFix for a missing "alt" attribute for following tags: img, applet, area.

I copied over some code from existing ExtractInlinedStyleRule and ExtractInlinedStyleHint and checked it, how it works. Fun fact, atm ExtractInlinedStyleHintFix is not working atm, it will throw an NPE, but this is not broken due to my changes. I will try to fix this later, in a seperate PR. Please have a look into the code. I will handle this as a draft and I will change some things like to add the rule/hint to the layers.xml instead of the HtmlHintsProvider as mentioned in the comment.

Also I changed a private method which gives you the exact offset of the tag <tag> or  problem here is that I need to add the attribute before the closing token so I added a bool to change this behaviour just for my and future needs. The rest is still working as before.

I also try to add tests for the hint, but there are atm no tests at all.

I also notied the extra space before my " alt=\"\"" when it is a closing tag. I can also try to fix this to check for token WS but it is the same behaviour as for WebStorm. It is better than nothing IMHO.

insert-alt-attribute


Chris2011:
I also thought to not hardcode the supported tags, but to use the validator rule for this, but this was to much for me. If this is wanted I can try to add this too, so this will be a bit more generic if other HTML tags are coming with the same validator rule.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttribute/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,89 @@
...
+        public String getDescription() {
+            return AddMissingAltAttributeRule.SINGLETON.getDisplayName();
+        }
+
+        @Override
+        public void implement() throws Exception {
+            EventQueue.invokeLater(new Runnable() {
+
+                @Override
+                public void run() {
```
mbien, line 64: feel free to bump the language level to 1.8 so that you can use lambdas.

Chris2011: did it

Chris2011:
Now I removed the rule and hint from the HintsProvider and add it to the layer.xml. The new rule extends now HtmlRule and runs the Hint by itself.

Chris2011:
Do I need to worry about the failing travis stuff? A quick look seems not related to my stuff anyway.

mbien:
> Do I need to worry about the failing travis stuff? A quick look seems not related to my stuff anyway.

I just restarted them. Looks good so far. You don't have a restart button?

ide/html.editor/src/org/netbeans/modules/html/editor/hints/HtmlHintsProvider.java
```diff
@@ -166,6 +166,7 @@ public void computeSuggestions(HintsManager manager, RuleContext context, List<H
         if(ExtractInlinedStyleRule.SINGLETON.appliesTo(context)) {
             suggestions.add(new ExtractInlinedStyleHint(context, new OffsetRange(context.caretOffset, context.caretOffset)));
         }
+
```
matthiasblaesing, line 169: This white-space change should not be here.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,234 @@
...
+
+        if (file == null) {
+            return false;
+        }
+
+        Project project = FileOwnerQuery.getOwner(file);
+
+        if (project == null) {
+            return false;
+        }
```
matthiasblaesing, line 72: Why is the project checked - it is not used later. This results in this:

image

The cursor is in the img-Tag (see my general comment), I get the error indicator, but I don't get the fix.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,234 @@
...
+    @Override
+    public boolean appliesTo(RuleContext context) {
+        this.context = context;
+
+        HtmlParserResult result = (HtmlParserResult) context.parserResult;
+        FileObject file = result.getSnapshot().getSource().getFileObject();
+
+        if (file == null) {
+            return false;
+        }
```
matthiasblaesing, line 66: Why is the file checked - it is not used.

matthiasblaesing (commented):
I left a few inline comments, but I made a general observation, that the integration looks a bit strange right now:

image

As you can see, I get an error indicator for the img-Tag, but I don't get the fixes. To get these I have to click inside the range of img-Tag:

image

I think this should be made consistent.

Chris2011:
> > Do I need to worry about the failing travis stuff? A quick look seems not related to my stuff anyway.
> 
> I just restarted them. Looks good so far. You don't have a restart button?

I had, but I didn't want to create some noice. But next time, I will try this first :). Thx for the info.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/HtmlHintsProvider.java
Chris2011: Just a leftover after I removed my check. Will remove it.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
Chris2011: Thx for the hint, yeah I didn't test an HTML file w/o having a project :).

Chris2011:
> I left a few inline comments, but I made a general observation, that the integration looks a bit strange right now:
> 
> image
> 
> As you can see, I get an error indicator for the img-Tag, but I don't get the fixes. To get these I have to click inside the range of img-Tag:
> 
> image
> 
> I think this should be made consistent.

You are right and I already tried a bit but the offset and searching logic for tokens (Tag open, tag open symbol, tag closing, etc.) drives me crazy. Will try to fix this.

Chris2011:
@matthiasblaesing after long thinking about this again I would disagree with the consistency. The warning/error/hint is really for the tag as you also see that the tag is highlighted when a div is around. So the hint shows it exactly for line + columnStart + columnEnd. For example if you have this example:

`<div></div>`

I really don't want to have the fix inside the div which is correct on this position. So the fix should really inside of the img tag not everywhere else.

Maybe the problem here is that the hint on the sidebar has an other expectation for the user.

Here are some other things:
* I checked the HTML validator it also selects the img tag, I mean you have all the information from that.
* I checked VS Code. They have no light bulb hardcoded on the sidebar so they just underline the img tag. The light bulb comes up when I set the cursor after `<|`.
* I checked WebStorm, it treats the missing alt attribute also as warning, but they don't show the warning as a light bulb and I also can't switch this. The light bulb also appears just after I set the cursor on `<|`.

So all in all, I think how NetBeans treats and shows warnings, is not optimal but this is not what I want to fix here. I also changed from warning to suggestion on current line, but the problem here is the highlight disappears too and the "suggestion"  is just seen in the statusbar BUT it just appears also on the img, not on the whitespace tokens before the tag_open_symbol. Hope that explains a bit more. I will just fix also a multi line img which makes sense to have the fix there too but not for the whole line. This will be a hole new topic IMHO.

matthiasblaesing:
@Chris2011 something went wrong here. There is a massive set of changes where and you have conflict with master. This needs to be redone.

Chris2011:
@matthiasblaesing thx for the info, will have a look soon.

Chris2011:
@matthiasblaesing I think I got it now :). I did the pull and rebase from within NetBeans and it had some conflicts, that I thought I fixed them but unfortunately not correct, sry for that. I reset everything to my last commit.

So now we can focus again on the initial task.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,84 @@
...
+import org.netbeans.modules.csl.api.Hint;
+import org.netbeans.modules.csl.api.HintFix;
+import org.netbeans.modules.csl.api.OffsetRange;
+import org.netbeans.modules.csl.api.RuleContext;
+import org.netbeans.modules.html.editor.utils.HtmlTagContextUtils;
+import org.openide.util.Exceptions;
+
+/**
+ *
+ * @author Chris
```
matthiasblaesing, line 33: I suggest to either replace this with something more distinctive (there are multiple instances of Chris) or remove it (history holds the full author information).

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,234 @@
...
+            TokenHierarchy th = TokenHierarchy.get(doc);
+            TokenSequence<HTMLTokenId> ts = Utils.getJoinedHtmlSequence(th, from);
+            
+            if (ts == null) {
+                //XXX - try to search backward and forward to find some html code???
+                return;
+            }
+            
+            //the joined ts is moved to the from offset and a token already selected (moveNext/Previous() == true)
+            //seek for all tag's attributes with css embedding representing an inlined style
```
matthiasblaesing, line 125: this comment looks as if it does not match.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,234 @@
...
```
matthiasblaesing, line 162: This looks to complex to find the alt Attribute.  The variables here indicate extraction of class, alt, id - is this copied from somewhere else?

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,234 @@
...
+import org.netbeans.modules.web.common.api.WebUtils;
+import org.openide.filesystems.FileObject;
+import org.openide.util.Exceptions;
+import org.openide.util.NbBundle;
+
+/**
+ *
+ * @author Chris
+ */
+@NbBundle.Messages("AddMissingAltAttributeRuleName=Insert required 'alt' attribute")
```
matthiasblaesing, line 51: This message key should be `AddMissingAltAttributeRule` - you can then drop the implementation of `getDisplayName` as the superclass already implements it:

https://github.com/apache/netbeans/blob/a8edf3bf92e38ac22a83e24ec910f29094f0fee3/ide/html.editor/src/org/netbeans/modules/html/editor/hints/HtmlRule.java#L86-L89
with
https://github.com/apache/netbeans/blob/a8edf3bf92e38ac22a83e24ec910f29094f0fee3/ide/html.editor/src/org/netbeans/modules/html/editor/hints/HtmlRule.java#L60-L63

You also need to define a message named `AddMissingAltAttributeRule_Desc` (you get an exception when you click on the hint in the option panel):

https://github.com/apache/netbeans/blob/a8edf3bf92e38ac22a83e24ec910f29094f0fee3/ide/html.editor/src/org/netbeans/modules/html/editor/hints/HtmlRule.java#L65-L68

I suggest to use : `Provide Text Alternatives` for the display name (to get it more in line with the rest of the entries) and `Find img/applet/area elements where no alt attribute is provided.` (for the description)

image

matthiasblaesing (commented):
I made a second pass over the changes. Please recheck the inline comments.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
Chris2011: Yes, also as you can see the comments, it was copied from the method findInlineStyle of class ExtractInlinedStyleRule. I started to copy stuff around, test some stuff and didn't fix code that much what was already there. Sry for that. I just wanted to get it working first. I will try to find anouther solution maybe with that what I already used for the HtmlCompletionProvider ` LexerUtils.followsToken(...`.

ide/html.editor/src/org/netbeans/modules/html/editor/utils/HtmlTagContextUtils.java
```diff
@@ -0,0 +1,80 @@
...
+import org.netbeans.api.lexer.Token;
+import org.netbeans.api.lexer.TokenSequence;
+import org.netbeans.modules.csl.api.OffsetRange;
+import org.netbeans.modules.html.editor.api.Utils;
+
+/**
+ *
+ * @author Chris
+ */
+public class HtmlTagContextUtils {
```
junichi11, line 33: `public final class`.
Please add a private constructor.

matthiasblaesing:
Removed the NB18 milestone and converted this to draft.

ide/html.editor/src/org/netbeans/modules/html/editor/utils/HtmlTagContextUtils.java
Chris2011: Done

Chris2011:
After long round, I hopefully simplyfied it enough. I had a deeper look into existing hints like for extracting inline tag styles and seeing missing id rule and missing class rule, which has a visitor for finding this. @matthiasblaesing and @junichi11 please have a look and leave comments as necessary. Thx.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        @Override
+        public void implement() throws Exception {
+            EventQueue.invokeLater(() -> {
+                try {
+                    Source source = Source.create(context.getFile());
+                    OffsetRange adjustContextRange = HtmlTagContextUtils.adjustContextRange(source.getDocument(false), range.getStart(), range.getEnd(), true);
+
+                    source.getDocument(false).insertString(adjustContextRange.getEnd(), " alt=\"\"", null);
```
junichi11, line 71: `// NOI18N`

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,90 @@
...
+                } catch (BadLocationException ex) {
+                    Exceptions.printStackTrace(ex);
```
junichi11, line 73: Instead, I would add a log.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AltAttributeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+import org.netbeans.modules.html.editor.lib.api.elements.*;
+import org.netbeans.modules.html.editor.refactoring.InlinedStyleInfo;
+import org.netbeans.modules.parsing.api.Source;
+
+/**
+ *
+ * @author Christian Lenz
+ */
+public class AltAttributeVisitor implements ElementVisitor {
+    private static final String ALT_ATTR = "alt";
```
junichi11, line 43: `// NOI18N`

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AltAttributeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            OffsetRange range;
+            String tag = null;
+            String attr = null;
+            int startTagOffset = 0;
+            int endTagOffset = 0;
+
+            do {
+                Token<HTMLTokenId> t = ts.token();
```
junichi11, line 79: `t` : Need `null` check

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AltAttributeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+                    range = new OffsetRange(startTagOffset, endTagOffset);
+                    //closing tag, produce the info
+                    if (tag != null && tag.matches(tagToFindRegEx) && attr == null) {
+                        //alt attribute found
+                        hints.add(new AddMissingAltAttributeHint(context, range));
+
+                        tag = attr = null;
+                    }
+                } else if (t.id() == HTMLTokenId.ARGUMENT) {
+                    if (t.text().toString().equals(ALT_ATTR)) {
```
junichi11, line 97: Instead, use `TokenUtilities.textEquals()`.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AltAttributeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+            TokenHierarchy th = TokenHierarchy.get(doc);
+            TokenSequence<HTMLTokenId> ts = Utils.getJoinedHtmlSequence(th, node.from());
+
+            if (ts == null) {
+                return;
+            }
+
+            OffsetRange range;
+            String tag = null;
+            String attr = null;
```
junichi11, line 74: This is not used as `String`. So, change this type to `CharSequence`.

ide/html.editor/src/org/netbeans/modules/html/editor/utils/HtmlTagContextUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import org.netbeans.api.lexer.Token;
+import org.netbeans.api.lexer.TokenSequence;
+import org.netbeans.modules.csl.api.OffsetRange;
+import org.netbeans.modules.html.editor.api.Utils;
+
+/**
+ *
+ * @author Christian Lenz
+ */
+public class HtmlTagContextUtils {
```
junichi11, line 33: Please add `final`.

ide/html.editor/src/org/netbeans/modules/html/editor/utils/HtmlTagContextUtils.java
```diff
@@ -0,0 +1,82 @@
...
+                return;
+            }
+
+            int adjustedFrom = ts.offset();
+
+            //now try to find the tag's end
+            ts.move(to);
+            int adjustedTo = -1;
+            while (ts.moveNext()) {
+                Token<HTMLTokenId> t = ts.token();
```
junichi11, line 62: Need `null` Check

junichi11:
Please write an example code to verify this in the description.

Please add unit tests. (`CslTestBase` has `checkHints()` and `applyHint()`)

Chris2011:
> Please write an example code to verify this in the description.
> 
> Please add unit tests. (`CslTestBase` has `checkHints()` and `applyHint()`)

Thx, I was searching for this. Will check it out.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,91 @@
...
+        @Override
+        public void implement() throws Exception {
+            EventQueue.invokeLater(() -> {
+                try {
+                    Source source = Source.create(context.getFile());
+                    OffsetRange adjustContextRange = HtmlTagContextUtils.adjustContextRange(source.getDocument(false), range.getStart(), range.getEnd(), true);
+
+                    source.getDocument(false).insertString(adjustContextRange.getEnd(), " alt=\"\"", null); // NOI18N
+                } catch (BadLocationException ex) {
+                    LOGGER.log(Level.SEVERE, ex.getMessage());
```
junichi11, line 74:
```suggestion
                    LOGGER.log(Level.WARNING, "Invalid offset: {0}", ex.offsetRequested());
```

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openide.filesystems.FileObject;
+import org.openide.util.Exceptions;
+
+/**
+ *
+ * @author Christian Lenz
+ */
+public class AddMissingAltAttributeRule extends HtmlRule {
+
+    public static AddMissingAltAttributeRule SINGLETON = new AddMissingAltAttributeRule();
```
junichi11, line 40:
```suggestion
    private static AddMissingAltAttributeRule INSTANCE = new AddMissingAltAttributeRule();
    
    private AddMissingAltAttributeRule() {
    }
    
    public static AddMissingAltAttributeRule getInstance() {
        return INSTANCE;
    }
```
BTW, Please note that we have to make it thread-safe if we make this class singleton.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeRule.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public HintSeverity getDefaultSeverity() {
+        return HintSeverity.CURRENT_LINE_WARNING;
+    }
+
+    @Override
+    protected void run(HtmlRuleContext context, List<Hint> result) {
+        try {
+            HtmlParserResult parserResult = context.getHtmlParserResult();
+            AltAttributeVisitor visitor = new AltAttributeVisitor(this, context, result, "img|applet|area");
```
junichi11, line 68: `// NOI18N`

ide/html.editor/src/org/netbeans/modules/html/editor/utils/HtmlTagContextUtils.java
junichi11: Sorry, I'm wrong... This case need not check it(`ts.moveNext` is `true`) but it's harmless.

matthiasblaesing (commented):
Am I missing something obvious here? The Rule/Hint is not registered anywhere and running a build under debugger and having a break point inside them, they are not hit. So I'm not sure what I should be seeing.

This:

image

Is generated in the html validator module, in the `ProblemHandler`:

image

With this change applied:

```diff
--- a/ide/html.editor/src/org/netbeans/modules/html/editor/resources/layer.xml
+++ b/ide/html.editor/src/org/netbeans/modules/html/editor/resources/layer.xml
@@ -534,6 +534,9 @@
                         <file name="org-netbeans-modules-html-editor-hints-NotValidatedContent.instance">
                             <attr name="position" intvalue="190"/>                            
                         </file>
+                        <file name="org-netbeans-modules-html-editor-hints-other-AddMissingAltAttributeRule.instance">
+                            <attr name="position" intvalue="200"/>
+                        </file>
                         <!-- the following hint 'Other' must be last in the hints list!!! -->
                         <file name="org-netbeans-modules-html-editor-hints-Other.instance">
                             <attr name="position" intvalue="99999"/>   
```

I get two matches, one from the html validator, one from this new hint (with the fix):

image

With the full parser result available the visitor can be simplified to (range now includes "/>", but that can be accounted for), at least I'm missing an argument to redo the scanning the parser already did.

```java
public class AltAttributeVisitor implements ElementVisitor {

    private static final String ALT_ATTR = "alt"; // NOI18N

    private final HtmlRuleContext context;
    private final List<Hint> hints;

    public AltAttributeVisitor(Rule rule, HtmlRuleContext context, List<Hint> hints) throws IOException {
        this.context = context;
        this.hints = hints;
    }

    @Override
    public void visit(Element node) {
        // We should only be invoked for opening tags
        if(! (node instanceof OpenTag)) {
            return;
        }
        // We are only interested in img, area, applet elements
        String lowerCaseTag = node.id().toString().toLowerCase();
        if(! (lowerCaseTag.equals("img") || lowerCaseTag.equals("area") || lowerCaseTag.equals("applet"))) {
            return;
        }

        OpenTag ot = (OpenTag) node;
        if(ot.getAttribute(ALT_ATTR) == null) {
            hints.add(new AddMissingAltAttributeHint(context, new OffsetRange(node.from(), node.to())));
        }
    }
}
```

I also notice, that the fix should check, if new whitespace needs to be inserted. If you have this case `` this fix will change it to `` (there are now two spaces)

Chris2011:
Thx for your review @matthiasblaesing. Yes I already encountered the missing xml part but this was done by a merge because I already had it. So I changed this now back to draft, because I already created the first test but it is still red due to not triggering the warning/suggestion handled as a error. After this I will change this back to normal.

Also I will have a look into the part that you wrote with the visit method change and the new added empty space. Thx.

Chris2011:
It is still in draft mode but can please someone have a look into my test impl? @junichi11 or @matthiasblaesing. For checkHints I needed to make a change in the CslTestBase but I'm not 100% sure whether I needed to do this or not. w/o my change the problem was that the hint was never triggered due to not writing the correct property for hints/suggestions (L4387: https://github.com/apache/netbeans/pull/4485/files#diff-afbe604d62e2e9bec89e4ed0e712ad047106ed77e63c2c4030e76a54d762941fR4387).

For the applyHint test, I also needed to change code within the CslTestBase in the method findApplicableFix (L4624: https://github.com/apache/netbeans/pull/4485/files#diff-afbe604d62e2e9bec89e4ed0e712ad047106ed77e63c2c4030e76a54d762941fR4624) from text == null to text != null. I'm also not 100% sure but the code makes more sense for me now. Otherwise it will always return null which breaks the code in applyHint (assertNotNull(fix))- This test is also funny, often it fails with that the expectation of the result is not the correct one but when I debug 1min or so with F8 to jump over the code, etc. it says that the test is green and I don't know what's wrong here.

I also added a check for Error and Warning which was missing, because if Current_line_warning is false it is not that the rest is always an error or an warning. computeHints maybe is missnamed here. It will just check for html version, nothing else. So all in all the code is often very frustrating to check and I needed days for debugging and yes it could be that I don't understand the code 100% that's why I'm asking for a bit of help please. Thanks all.

Chris2011:
Unfortunately I see that my changes break some PHP tests. Will dig into it again and would be happy to have someone that helps me break it down what I missed.

Chris2011:
@matthiasblaesing so another research hear:
* There are no CSS Hints/Suggestions/Fixes, etc, just the CssHintsProvider where the computeHints, computeSuggestions and computeSelectionHints method bodies are empty. I already have one CSS hint/suggestion on another branch, just to test smth but without any test.

* The best implementation that I found is the PHP stuff for suggestions, but the problem is that I get a timeout when I try to run a PHP related suggestion test like this one locally: https://github.com/apache/netbeans/blob/master/php/php.editor/test/unit/src/org/netbeans/modules/php/editor/verification/InitializeFieldsSuggestionTest.java and on my branch, on the pipeline, the tests are not executed/skipped, but maybe because of the failing latte tests.

* I also had a look into JsHintsProvider and JsHints. There is just one fix, changing the EcmaLevel but everything seems a warning. The tests will not work with suggestions. There will be a Parse exception, when I change the severity warning to a current_line_waning. Ok I really don't know whether I need to change smth else.

Fazit: It seems PHP and maybe JS are better implemented in their *HintsProvider and I need to change the whole logik of the HtmlHintsProvider but JS also lacks of tests and suggestions are also not a big player for JS (not yet). Should this be part of this ticket? I guess not.

I also think that we should make this somehow consistent for all CSL languages as good as possible. PHP uses Classes for like SuggestionRule, HintRule and ErrorRule, which just set the severity to the correct one and so on. So probably I will create another PR with the "fx" of the HtmlHintsProvider with using Tests.

I think I ran into a rabbithole and yes, it is not your business but the code is a big mess and I want to rework it and hope to not break more than before. Any other thoughts/suggestions/hints? :D

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,107 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.html.editor.hints.other;
+
+import java.awt.EventQueue;
```
matthiasblaesing, line 21: Unnecessary import

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.swing.text.BadLocationException;
+import javax.swing.text.Document;
```
matthiasblaesing, line 28: Unnecessary import

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,107 @@
...
+import org.netbeans.editor.BaseDocument;
+import org.netbeans.modules.csl.api.Hint;
+import org.netbeans.modules.csl.api.HintFix;
+import org.netbeans.modules.csl.api.OffsetRange;
+import org.netbeans.modules.html.editor.hints.HtmlRuleContext;
+import org.netbeans.modules.html.editor.utils.HtmlTagContextUtils;
+import org.netbeans.modules.parsing.api.Source;
```
matthiasblaesing, line 35: Unnecessary import

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,107 @@
...
+            Collections.<HintFix>singletonList(new AddMissingAltAttributeHintFix(context, range)),
+            10);
+    }
+
+    private static class AddMissingAltAttributeHintFix implements HintFix {
+
+        private static final Logger LOGGER = Logger.getLogger(AddMissingAltAttributeHintFix.class.getSimpleName());
+
+        HtmlRuleContext context;
+        OffsetRange range;
```
matthiasblaesing, line 57: These can be `private` and `final`

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
```diff
@@ -0,0 +1,107 @@
...
+                try {
+                    OffsetRange adjustedRange = HtmlTagContextUtils.adjustContextRange(document, range.getStart(), range.getEnd(), true);
+                    String tagContent = document.getText(adjustedRange.getStart(), adjustedRange.getLength());
+
+                    // Find last self-closing or non self-closing tag
+                    Pattern closingTagPattern = Pattern.compile("(/?>)");
+                    Matcher closingTagMatcher = closingTagPattern.matcher(tagContent);
+
+                    if (closingTagMatcher.find()) {
+                        int altInsertPosition = adjustedRange.getStart() + closingTagMatcher.start(1);
```
matthiasblaesing, line 82: This might lead to wrong results. Consider PHP with embedded HTML. The document could look like this:

```php
<img src='demo.png' alt='<?php echo htmlspecialchars('x > y');?>' />
```

I would use the HTML token stream and either use the start of the closing token or, if that does not exist, the start of the last HTML token + length.

ide/csl.api/test/unit/src/org/netbeans/modules/csl/api/test/CslTestBase.java
```diff
@@ -4384,8 +4384,11 @@ protected ComputedHints computeHints(final NbTestCase test, final Rule hint, Str
                             }
                         }
                         if (HintsSettings.getSeverity(manager, ucr) == HintSeverity.CURRENT_LINE_WARNING) {
-                            manager.setTestingRules(null, Collections.EMPTY_MAP, testHints, null);
+                            manager.setTestingRules(null, testHints, testHints, null);
                             provider.computeSuggestions(manager, context, hints, caretOffset);
+                        } else if(HintsSettings.getSeverity(manager, ucr) == HintSeverity.ERROR || HintsSettings.getSeverity(manager, ucr) == HintSeverity.WARNING) {
+                            manager.setTestingRules(null, testHints, testHints, null);
+                            provider.computeErrors(manager, context, hints, new ArrayList<Error>());
```
matthiasblaesing, line 4391: Did you test, that all current tests invoking `computeHints`, potentially transitively, still work?

Chris2011: No, not yet. I just thought that the pipes will show me the result and I just saw some failing tests of Latte. PHP where not triggert here at github inside my branch. I will just revert this change as you said and will try to implement a better logic into the HtmlHintsProvider as similiar as for PHP and JS.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
Chris2011: TBH, I didn't try PHP yet, will do it too. Thx for the hint. If I guess it correct, your example will not trigger the hint, because you already have the alt attribute. But yes It would be better to take the HTML token instead of just looking for the closing tag symbol with an regex.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
Chris2011: You mean, they are unnecessary, when I change the logik, right? Because atm I need them.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
Chris2011: Same as above

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
matthiasblaesing: Neither `Document`, nor `Source` are used. I admit, that I only noticed that because NetBeans marked these:

image

Chris2011: Yes, those three are correct marked as not used, but your both other comments have different lines. That's why I'm asking. It was about the matcher, pattern, the exception, etc. That's why I was confused about it.

ide/html.editor/src/org/netbeans/modules/html/editor/hints/other/AddMissingAltAttributeHint.java
Chris2011: I added a way to make this happen in php files too. Unfortunately I didn't add new PHP tests.

ide/csl.api/test/unit/src/org/netbeans/modules/csl/api/test/CslTestBase.java
Chris2011: I revert the CslTestBase code and added new testst and run the tests in Html Editor and everysthing is green now.

Chris2011:
Ready for review again :)

Chris2011:
Ok i see failing tests, that are working on my machine :(

matthiasblaesing:
@Chris2011 I ran tests locally and they also error out for me. I wanted to look into this, but then I noticed, that there are changes that make no sense to me when looking at the title of the PR "Add HTML rule and hint+fix for missing alt attribute for tag img, applet and area".

Why are there changes to `HtmlHintsProvider.java` and `HtmlRuleContext.java`? Given that tests break that are not tied to the new functionalilty, I would look at these first. And the first question I would raise: Are the changes needed? I think I see a different bug being fixed here (missing cancelability), but I would try to make this minimal and split out, what is not strictly needed.

Chris2011:
> My gut feeling is, that the HtmlHintsProvider and the expectations of the CslTestBase/GsfHintsManager don't match. It seems suggestions were never really supported. It might be worthy looking into JS/CSS whether there are suggestions being checked. I'm not sure, that modifying CslTestBase is really a good idea.

After you said that (in 2024 :D) I checked the PHP implementation of the HtmlHintsProvider and what I found was the missing cancel functionality. And this is what I implemented first. Maybe I have a caching issue locally but after this my test went green and I started to implement more tests and as I said all the tests are green "on my machine" but I know it works on my machine is not the best approach. I will have a look into it. Thx for eyeballing it. I thought some raise condition that came up here but dunno yet. Will check it. I also added some code for the thing that @junichi11 said about embedded php in html.

The problem what I see and saw years ago, that still PHP was the best to look into it. CssHintsProvider just handle errors and the rest is empty. Ok with my knowledge nowadays help of some buddies to help me understand the code, the JsHintsProvider can also help.

For now maybe I need to make an ant clean build and check the tests again but with not failing tests I dunno how I can verif it and I don't want to spam the PR. I'm working on multiple worktrees maybe this can also be a problem?

Also I'm working on windows with JDK 25 for build, different from the JDK and the system of the github actions.

matthiasblaesing:
> Also I'm working on windows with JDK 25 for build, different from the JDK and the system of the github actions.

My basic approach on _each_ branch switch:

1. Checkout the target branch `git checkout <TARGET>`
2. Ensure worktree is clean `git clean -f -x -d`
3. Build `ant`

With that I was able to reproduce the problem on Ubuntu with Amazon Corretto 21.0.3.

Chris2011:
Will do, thx.

PR #4495, opened 2022-08-10
Support for project output artifacts

sdedic:
This PR actually consists of a series of features. I should probably make separate requests for that, but ... that would prolong the review process, I believe.

The overall goal was to add a project query to get the project's output in form of an identifiable artifact(s) and their supposed (not built yet) or existing locations - and expose the analysis result through LSP.

Overview:
## ProjectActionContext
This one I wanted for a long time already. Many project APIs or services are likely to be affected by project configurations - and it's rather hard to invoke the services for a specific configuration, which is not the active one. By default UI works against the active configuration, which is fine, but API access should have an option to specify it.
Adding configurations to all APIs is hard, and sometimes it is even impossible when the execution goes through some uncooperative layers.; the `ProjectActionContext` is a way how to pass the information 'out of band'.

I needed to pass specifically project action to distinguish different types of build; for example Micronaut plugins (gradle, maven) support a regular build, and a native-image build, that produces different output(s).

Important note for reviewers: while the ProjectActionContext allows to specify a profile or user properties, which can be ignored by the underlying project system, and give almost complete control over project model evaluation to the caller (at least for gradle and maven), I am not 100% sure if the properties + profiles is a good idea. A conservative approach would be to insist on creating a `ProjectConfiguration` for the profile+properties setup and use that.

Reviewers, share your opinion on this; these ProjectActionContext properties could be eventually added (compatibly) later.

## ProjectArtifactsQuery
this is a project query as usual, asking for project output with GAV coordinates (if applicable). I am almost sure the `ArtifactSpec` will satisfy PHP, too ... but I still keep it private until Gradle implementation is complete.

The `Filter` specification does not cover all query cases, but works for the basic ones. When one wants "any output" and a specific output type / classifier.

## Micronaut support for native image build
Added a project action and project action mapping for Maven to execute native-image build of the micronaut app. The action becomes available (and is configurable in project customizer) when one adds Micronaut plugin to the maven project build configuration.

## Maven fixes
During implementation I've realized that the native-build action, contributed to the Maven by `MavenActionsProvider` is not visible in the customizer. I was little shocked when I realized that the customizer essentially hardcodes the list of (standard) actions and only displays user-defined ones in addition to the hardcoded list. I've corrected the behaviour of `M2Configuration` in that it lists just config-specific mappings, the others are retrieved through `ActionToGoalUtils`.
I've also realized that the declarative actions lack proper support for I18N; so I've added a convenience lookup in ` Bundle.proeprties` using the action's (profile's) id, and in additiona specifying `displayName` that starts with '#' also makes a lookup, with the possbility to identify the Bundle path as well.

### DIFF
--- a/enterprise/micronaut/nbproject/project.xml
+++ b/enterprise/micronaut/nbproject/project.xml
@@ -69,6 +69,32 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.modules.maven</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>2.154</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.project.dependency</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <implementation-version/>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.maven.embedder</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>2.72</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
                     <code-name-base>org.netbeans.api.progress</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
@@ -382,6 +408,39 @@
                     </run-dependency>
                 </dependency>
             </module-dependencies>
+            <test-dependencies>
+                <test-type>
+                    <name>unit</name>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.libs.junit4</code-name-base>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.nbjunit</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.java.source.base</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.editor.mimelookup.impl</code-name-base>
+                        <recursive/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.openide.modules</code-name-base>
+                        <compile-dependency/>
+                        <test/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.projectapi.nb</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                </test-type>
+            </test-dependencies>
             <public-packages/>
             <class-path-extension>
                 <runtime-relative-path>ext/spring-boot-configuration-metadata-2.4.4.jar</runtime-relative-path>
--- a/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
+++ b/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.modules.micronaut.maven;
+
+import java.util.Arrays;
+import javax.swing.Action;
+import org.netbeans.api.project.Project;
+import org.netbeans.spi.project.ActionProvider;
+import org.netbeans.spi.project.ui.support.ProjectActionPerformer;
+import org.netbeans.spi.project.ui.support.ProjectSensitiveActions;
+import org.openide.awt.ActionID;
+import org.openide.awt.ActionReference;
+import org.openide.awt.ActionRegistration;
+import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
+
+/**
+ *
+ * @author sdedic
+ */
+public class BuildActions {
+    
+    @ActionID(id = "org.netbeans.modules.micronaut.maven.nativeBuild", category = "Project")
+    @ActionRegistration(displayName = "#LBL_NativeBuild", lazy=false)
+    @ActionReference(position = 520, path = "Projects/org-netbeans-modules-maven/Actions")
+    @NbBundle.Messages({
+        "LBL_NativeBuild=Build Native Executable"
+    })
+    public static Action    createNativeBuild() {
+        return ProjectSensitiveActions.projectSensitiveAction(new NativeBuildPerformer(), Bundle.LBL_NativeBuild(), null);
+    }
+    
+    private static class NativeBuildPerformer implements ProjectActionPerformer {
+        @Override
+        public boolean enable(Project project) {
+            ActionProvider ap = project.getLookup().lookup(ActionProvider.class);
+            return 
+                    ap != null &&
+                    Arrays.asList(ap.getSupportedActions()).contains(MicronautMavenConstants.ACTION_NATIVE_COMPILE) &&
+                    ap.isActionEnabled(MicronautMavenConstants.ACTION_NATIVE_COMPILE, Lookup.EMPTY);
+        }
+
+        @Override
+        public void perform(Project project) {
+            ActionProvider ap = project.getLookup().lookup(ActionProvider.class);
+            if (ap == null) {
+                return;
+            }
+            if (!Arrays.asList(ap.getSupportedActions()).contains(MicronautMavenConstants.ACTION_NATIVE_COMPILE) || 
+                !ap.isActionEnabled(MicronautMavenConstants.ACTION_NATIVE_COMPILE, Lookup.EMPTY)) {
+                return;
+            }
+            ap.invokeAction(MicronautMavenConstants.ACTION_NATIVE_COMPILE, Lookup.EMPTY);
+        }
+    }
+}
--- a/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
+++ b/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
@@ -123,6 +123,7 @@ public final class ArtifactSpec<T> {
         hash = 79 * hash + Objects.hashCode(this.artifactId);
         hash = 79 * hash + Objects.hashCode(this.versionSpec);
         hash = 79 * hash + Objects.hashCode(this.classifier);
+        hash = 79 * hash + Objects.hashCode(this.location);
         return hash;
     }
 
@@ -151,6 +152,9 @@ public final class ArtifactSpec<T> {
             return false;
         }
         if (!Objects.equals(this.classifier, other.classifier)) {
+            return false;
+        }
+        if (!Objects.equals(this.location, other.location)) {
             return false;
         }
         return this.kind == other.kind;
--- a/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
+++ b/ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
@@ -18,10 +18,18 @@
  */
 package org.netbeans.modules.project.dependency;
 
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.Objects;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.netbeans.api.annotations.common.NonNull;
 import org.netbeans.api.annotations.common.NullAllowed;
 import org.openide.filesystems.FileObject;
+import org.openide.filesystems.FileUtil;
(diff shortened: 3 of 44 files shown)


sdedic:
added @lkishalmi  since I'm going to make equivalent changes for Gradle in a separate PR. Please comment on the APIs.

dbalek (approved):
Proposed API looks sane to me.

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
```diff
@@ -0,0 +1,402 @@
...
+    private Map<String, String> properties;
+    private Set<String> profiles;
+    private Lookup lookup;
+
+    /**
+     * Returns the project this context applies to. Having {@link Project} as a property allows multiple
+     * contexts, individually for each project, to be active.
+     * @return the target project.
+     */
+    public @NonNull Project getProject() {
```
lkishalmi, line 86: There are many possible null returns in this class. That should be avoided unless it brings a meaning.

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
```diff
@@ -0,0 +1,402 @@
...
+
+    /**
+     * Additional information for the underlying project implementation, similar to 
+     * context lookup in {@link org.netbeans.spi.project.ActionProvider#invokeAction}.
+     * Null if no specific info is present.
+     * 
+     * @return Lookup with additional information or {@code null}.
+     */
+    public @CheckForNull Lookup getLookup() {
+        return lookup;
```
lkishalmi, line 98: Can return Lookup.EMPTY

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
```diff
@@ -0,0 +1,402 @@
...
+    
+    /**
+     * User-customized properties that should be effective during the computation. The same
+     * customization should be made during project action execution to obtain the same results as 
+     * evaluated by the query.
+     * 
+     * @return user properties, or {@code null} if unspecified.
+     */
+    public @CheckForNull Map<String, String> getProperties() {
+        return properties == null ? null : Collections.unmodifiableMap(properties);
```
lkishalmi, line 129: Collections.emtyMap()

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    /**
+     * Profiles or some project system features/tags that should be applied for the evaluation.
+     * @return applied profiles or {@code null} if unspecified.
+     */
+    public @CheckForNull Set<String> getProfiles() {
+        return profiles == null ? null : Collections.unmodifiableSet(profiles);
```
lkishalmi, line 137: Collections.emptySet()

lkishalmi (commented):
Well, not finished with the review, just made some comments.

enterprise/micronaut/nbproject/project.xml
```diff
@@ -68,6 +68,32 @@
...
+                <dependency>
+                    <code-name-base>org.netbeans.modules.project.dependency</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <implementation-version/>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.maven.embedder</code-name-base>
```
jtulach, line 89: These modules don't provide real API. Depending on them from foreign "micronaut" module is strange. Also adding a dependency on Maven support, when "micronaut" supports both, Maven and Gradle is strange.

enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
```diff
@@ -0,0 +1,78 @@
...
+    private static class NativeBuildPerformer implements ProjectActionPerformer {
+        @Override
+        public boolean enable(Project project) {
+            NbMavenProject nbmp = project.getLookup().lookup(NbMavenProject.class);
+            if (nbmp == null) {
+                return false;
+            }
+            ActionProvider ap = project.getLookup().lookup(ActionProvider.class);
+            return 
+                    nbmp.getMavenProject().getPlugin("io.micronaut.build:micronaut-maven-plugin") != null &&
```
jtulach, line 59: I was hoping this (presence of some plugin) could be specified declaratively. Isn't there already a support for a plugin to trigger some configuration and configurations adding some actions?

ide/project.dependency/src/org/netbeans/modules/project/dependency/spi/ProjectArtifactsImplementation.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public interface ProjectArtifactsImplementation {
+    /**
+     * Finds project artifact(s) that match the query. 
+     * @param query the query to answer
+     * @param previous result of previous buildsteps
+     * @param all
+     * @return 
+     */
+    public Result findArtifacts(ProjectArtifactsQuery.Filter query);
```
jtulach, line 40: I prefer [singletonizer](https://bits.netbeans.org/14/javadoc/org-netbeans-api-java-classpath/org/netbeans/spi/java/queries/BinaryForSourceQueryImplementation2.html) SPI over `ProjectArtifactsImplementation` and `Result` inner class.

ide/project.dependency/src/org/netbeans/modules/project/dependency/spi/ProjectDependenciesImplementation.java
```diff
@@ -32,16 +29,9 @@
  * @author sdedic
  */
 public interface ProjectDependenciesImplementation {
```
jtulach, line 31: Removing methods from `spi` package? Isn't that an API change?

ide/projectapi/src/org/netbeans/api/project/ProjectUtils.java
```diff
@@ -57,6 +61,65 @@ private ProjectUtils() {}
...
+     * the method throws {@link IOException} if the setting save fails.
+     * 
+     * @param <C> configuration type
+     * @param p the project
+     * @param cfg configuration or {@code null} for default configuration.
+     * @return true, if the configuration was successfully set.
+     * @throws IOException when the selected configuration cannot be persisted.
+     * @since 1.89
+     */
+    public <C extends ProjectConfiguration> boolean setActiveConfiguration(@NonNull Project p, @NonNull C cfg) throws IOException {
```
jtulach, line 94: This would type the same without generics.

java/java.lsp.server/nbcode/integration/nbproject/project.xml
```diff
@@ -214,6 +214,14 @@
                         <specification-version>9.24</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.project.dependency</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <implementation-version/>
```
jtulach, line 222: Implementation dependency!?

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/commands/ProjectMetadataCommand.java
```diff
@@ -0,0 +1,180 @@
...
+            if (!(o instanceof JsonPrimitive)) {
+                throw new IllegalArgumentException("String or null expected as parameter #2, got " + o);
+            }
+            artifactType = ((JsonPrimitive)o).getAsString();
+        }
+        ProjectArtifactsQuery.Filter filter = ProjectArtifactsQuery.newQuery(artifactType, null, ctx);
+        CompletableFuture result = new CompletableFuture();
+        METADATA_PROCESSOR.post(() -> {
+            try {
+                ArtifactsResult arts = ProjectArtifactsQuery.findArtifacts(p, filter);
```
jtulach, line 169: All the changes _just_ to get name of artifact to VSCode TypeScript side...

java/maven.embedder/nbproject/project.xml
```diff
@@ -182,6 +182,7 @@
                 <friend>io.github.jeddict.runtime</friend>
                 <friend>io.github.jeddict.rest.generator</friend>
                 <!-- XXX <subpackages> not permitted by schema -->
+                <friend>org.netbeans.modules.micronaut</friend>
```
jtulach, line 185: I don't like this. `micronaut` module shall not have direct dependency on Maven (embedder).

java/maven/src/org/netbeans/modules/maven/api/NbMavenProject.java
```diff
@@ -286,6 +287,22 @@ public void removePropertyChangeListener(PropertyChangeListener propertyChangeLi
...
+     * Returns a project evaluated for a certain purpose. The while {@link #getMavenProject}
+     * works with the <b>active configuration</b> and does not apply any action-specific properties,
+     * this method tries to apply mappings, configurations, etc when loading the project model.
+     * <p>
+     * Note that loading an evaluated project may take significant time (comparable to loading
+     * the base project itself). The implementation might optimize if the passed context does not
+     * prescribe different profiles, properties etc than have been used for the default model.
+     * 
+     * @param context the loading context
+     * @return evaluated project
```
jtulach, line 300: Missing versioning, I assume.

jtulach (commented):
Personally, I don't find adding new implementation and friend dependencies attractive.

enterprise/micronaut/nbproject/project.xml
sdedic: re. depending from Micronaut no Gradle | Maven - agreed, a solution would be to create eager micronaut-maven + micronaut-gradle modules. 

re. no real API - hm, I need to use the maven model types, i.e. `org.apache.maven.model.PluginExecution`. These are currently exposed to the module system by `maven.emebdder`. Is there a better way how to import those types ?

enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
sdedic: Hmmm, good catch, will try to use plain `isActionEnabled`.

ide/project.dependency/src/org/netbeans/modules/project/dependency/spi/ProjectDependenciesImplementation.java
sdedic: This "API" was never published, even as a friend.

java/java.lsp.server/nbcode/integration/src/org/netbeans/modules/nbcode/integration/commands/ProjectMetadataCommand.java
sdedic: erm. not just the name, but also the location of the potential output. Yes, it's large, but NetBeans Project APIs themselves lack a description of usual project metadata, some identifier etc. The location of to-be-compiled output could be useful for packaging steps in NB, too.

java/maven.embedder/nbproject/project.xml
sdedic: would be gradle+maven micronaut bridge modules more approppriate ?

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
sdedic: Hm, with all those nulls (including this one), I intended to signal explicitly "no value". For example, if no Lookup + properties + profiles are present the implementation can take a great shortcut (= no customization to the project model or services).

lkishalmi: Well if those shortcuts are important, then probably I'd intorduce a ```"boolean isShortcutAllowed()"``` method here to do the necessary checks inside here. Of course I assume that you can do shortcuts if no lookup + properties + profiles are present.

sdedic: The little trouble is that it is the project implementation that decides which part may potentially affect the project data evaluation ... all stuff except the Lookup can be checked for 'no ops' (empty), the Lookup is ... well, peculiar. Eventually I could document that the exact value `Lookup.EMPTY` may be used for optimizations.

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
sdedic: Addressed in a1c2f1b8c6b2

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
sdedic: Addressed in a1c2f1b8c6b2

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
sdedic: Addressed in a1c2f1b8c6b2

ide/projectapi/src/org/netbeans/api/project/ProjectActionContext.java
sdedic: this particular one does not ;))  the others were addressed in a1c2f1b8c6b2

ide/project.dependency/src/org/netbeans/modules/project/dependency/spi/ProjectArtifactsImplementation.java
sdedic: addressed in 70dcb7f24620

ide/projectapi/src/org/netbeans/api/project/ProjectUtils.java
sdedic: The project API itself is using generics ... (?)

enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
sdedic: Addressed in 2917d2be433e

ide/project.dependency/src/org/netbeans/modules/project/dependency/ArtifactSpec.java
```diff
@@ -75,17 +86,45 @@ public enum VersionKind {
...
     public T getData() {
         return data;
     }
 
+    /**
+     * Returns local file which the artifact represents. For library (dependencies) artifacts,
+     * the file is the library consumed, e.g. from a local repository. For outputs, the artifact
+     * represents the build output, usually in project's build directory. Note that FileObject for 
+     * a dependency and its corresponding Project may not be the same.
+     *
```
ppisl, line 102: Probably there should be mentioned, that can return null, if the local file doesn't exist?

ppisl (approved):

sdedic:
2 approvals; going to merge soon, after rebasing to master (minor clash) + consolidating fixes (less clutter).

PR #4025, opened 2022-04-23
clean up code base to remove Windows 95 & 98 support..

BradWalker:
Windows 95 & Windows 98 support is no longer provided by the base JVM.  Matter of fact, support ended in Java 5 for these platforms.

This cleans up the code base to no longer support these obsolete platforms.

---
**^Add meaningful description above**

By opening a pull request you confirm that, unless explicitly stated otherwise, the changes -

 - are all your own work, and you have the right to contribute them.
 - are contributed solely under the terms and conditions of the Apache License 2.0 (see section 5 of the license for more information).

Please make sure (eg. `git log`) that all commits have a valid name and email address for you in the Author field.

If you're a first time contributor, see the Contributing guidelines for more information.

PR #4497, opened 2022-08-11
Fix bitshfits in the OS constants after the removal of OS_WIN95 and OS_WIN98

lkishalmi:
This PR restores the bit assignment to the OS constants before #4025

PR #4487, opened 2022-08-09
Snapshot of APIs as of NetBeans 15

neilcsmith-net:
Snapshot of APIs as of NetBeans 15.  For review (sorry, bit late again due to some issues, but we can run another RC if there's a real need to fix anything).  Not to be merged as / until release vote completes.

Constant changes introduced in #4025 in Utilities might be a problem?

Built in the usual way -

```
# checkout branch (or release tag)
git checkout release150
ant clean
ant build
ant gen-sigtests-release
git add -A
git stash
git checkout master
git checkout -b apis-nb150
git stash pop
git add -A
git commit -m "Snapshot of APIs as of NetBeans 15"
# remove getPeer() calls
find . -name "*.sig" -exec sed -i '/java.awt.peer.ComponentPeer/{N;d;}' {} \;
# check git diff
git add -A
git commit --amend
```

PR #4498, opened 2022-08-12
Revert PR4025 due to changes in compile time constants.

neilcsmith-net:
Straight revert of #4025 due to breaking changes in compile time constants.

cc @lkishalmi @mbien @BradWalker alternative option to revert rather than patch. Some of the other changes could then be looked at and reviewed again in master for NB16.


mbien (approved):
looks good to me. everything is here and it applies cleanly.

BradWalker:
Good morning folks. Hey @neilcsmith-net , @mbien can I get a clearer picture of what is broken? The bug says "breaking changes" but it's not clear to me what is broken.. Clearly the platforms are dead, so maybe I can easily fix it instead of doing a revert..

mbien:
@BradWalker values of public final primitive fields can't change. If they do it will break existing code since compilers can inline those sometimes. It would be also safer to simply keep all public fields and deprecate them. see #4497 and dev list.

neilcsmith-net:
I'm not sure it's "can inline" but rather "must inline". eg. from JLS "If a final field is initialized to a constant expression in the field declaration, ... uses of that final field are replaced at compile time with the value of the [constant expression](https://docs.oracle.com/javase/specs/jls/se14/html/jls-15.html#jls-15.29)".

@BradWalker this means that the value those fields are assigned should not change. Code (eg. third-party plugins) compiled against older versions of the IDE could have different values. The masks are also constant expressions.  It's a binary incompatible change.

The values are also in the API signature files - I saw it initially when generating those for NB15.  Check the diff in #4487  The tests should really have not passed with the changes.

I'm of the opinion that revert, fix, re-apply is sometimes a better strategy for handling breaking changes.  Obviously here the re-apply would be in master for NB16 though.

lkishalmi (approved):

platform/o.n.swing.tabcontrol/beanstubs/org/openide/util/Utilities.java
```diff
@@ -56,8 +56,12 @@ private Utilities() {}
 
     /** Operating system is Windows NT. */
     public static final int OS_WINNT = 1 << 0;
+    /** Operating system is Windows 95. */
+    public static final int OS_WIN95 = OS_WINNT << 1;
```
JaroslavTulach, line 60: Thank you for restoring the previous API.

PR #4499, opened 2022-08-12
Initial implementation of (experimental) Dependency API in Gradle

sdedic:
Dependency tree is exported from Gradle, using *still nonpublic* `project.dependency` API. I admit I need it functioning for the NBLS server, so I propose to go short-term with **implementation** dependency on `project.dependency` module. I plan to stabilize and publish the project.dependency as a regular API in NB16 timeframe with `ProjectArtifactsQuery` that exposes project output artifact(s) for further analysis or processing.  I hope I will be able to adapt Maven Graph module to serve for both Maven and Gradle connecting using the new APIs.

Part of this commit is an adjustment to the OCI enterprise feature so it is able to exploit this new Gradle ability to serve dependency tree. During testing it seems that an initial dependency graph of 100 nodes would explode to 10.000+ nodes when converted to a proper tree, so the Dependency API will receive some tweak to allow implementations to compute the nodes lazily and to indicate to the client that two nodes are truly equivalent. In the meantime, I've changed the ADM feature to send a graph instead of tree - it appears working.

Minor fix to Maven implementation.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
@@ -229,7 +229,7 @@ public final class GradleBaseProject implements Serializable, ModuleSearchSuppor
             String n = getName();
             String g = getGroup();
             String v = getVersion();
-            if (n == null || "".equals(n) || g == null || "".equals(g) || v == null || "".equals(v) || "unspecified".equals(v)) {  // NOI18N
+            if (n == null || n.isEmpty() || g == null || g.isEmpty() || v == null || v.isEmpty() || "unspecified".equals(v)) {  // NOI18N
                 return null;
             }
             return String.format("%s:%s:%s", g, n, v);
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
@@ -116,6 +116,10 @@ public final class GradleBaseProject implements Serializable, ModuleSearchSuppor
     public String getDisplayName() {
         return displayName;
     }
+    
+    public boolean isVersionSpecified() {
+        return version != null && !"".equals(version) && !"unspecified".equals(version); // NOI18N
+    }
 
     public String getVersion() {
         return version;
@@ -213,13 +217,23 @@ public final class GradleBaseProject implements Serializable, ModuleSearchSuppor
     /**
      * Finds a GAV for the given project. Returns {@code null} if the project path
      * is not known (it is not referenced anywhere by this project), or has no known
-     * GAV. The project's own GAV should be always present.
+     * GAV. The project's own GAV should be always present, if defined by the project 
+     * file(s).
      * 
      * @param projectPath Gradle project path
      * @return GAV coordinates, or {@code null}
      * @since 2.27
      */
     public String findProjectGav(@NonNull String projectPath) {
+        if ("".equals(projectPath) || getPath().equals(projectPath)) {
+            String n = getName();
+            String g = getGroup();
+            String v = getVersion();
+            if (n == null || "".equals(n) || g == null || "".equals(g) || v == null || "".equals(v) || "unspecified".equals(v)) {  // NOI18N
+                return null;
+            }
+            return String.format("%s:%s:%s", g, n, v);
+        }
         return projectIds.get(projectPath);
     }
 
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
@@ -53,7 +53,7 @@ public final class GradleBaseProject implements Serializable, ModuleSearchSuppor
     String name;
     String group = "";
     String description;
-    String version = "";
+    String version;
     String path;
     String status;
     String parentName;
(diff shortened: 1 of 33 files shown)


lkishalmi:
Well, what I am curious is what the additional memory footprint of adding the dependencies into the GradleBaseProject the 10000+ nodes sounds scary.
It would be good to know how essential this information would be. What I'm thinking about, that this info could be fetched/cached in a separate Model. Though that may require more work.

I'm fine to put the dependency info (if it is not that exploded) in the GradleBaseProject. 

So it would be good to now what would it put on the IDE on memory and probably on CPU. If it could be expensive let's move it to another model. It can be done later though, if you do not expose the dependencies through GradleBaseProject.

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
```diff
@@ -52,7 +53,7 @@ public final class GradleBaseProject implements Serializable, ModuleSearchSuppor
     String name;
     String group = "";
     String description;
-    String version;
+    String version = "";
```
lkishalmi, line 56: ```null``` version value has a meaning. Gradle says it is ```unspecified```, when it is an empty string it is an empty string.

sdedic:
@lkishalmi The structure in Configuration is just (currently) just a graph, not an unrolled tree, so in "usual" projects it is rather slim. No `ArtifactSpec` structures are created in advance and the graph contains just GAVs of the target nodes. The tree is made only when someone asks for the DependencyResult.

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
```diff
@@ -319,6 +335,9 @@ GradleConfiguration createConfiguration(String name) {
      *         {@code null} for non-Gradle projects.
      */
     public static GradleBaseProject get(Project project) {
+        if (project == null) {
```
lkishalmi, line 338: Is this needed? NbGradleProject.get() should return null on a null project.

java/gradle.java/nbproject/project.properties
```diff
@@ -24,4 +24,6 @@ javadoc.apichanges=${basedir}/apichanges.xml
 
 test-unit-sys-prop.test.netbeans.dest.dir=${netbeans.dest.dir}
 test-unit-sys-prop.java.awt.headless=true
-test.use.jdk.javac=true
\ No newline at end of file
+test.use.jdk.javac=true
+spec.version.base=1.19.0
```
lkishalmi, line 28: Why are we moving the Specification-Version out of manifest.mf and use spec.version.base?
I'm just curious.

java/gradle.java/src/org/netbeans/modules/gradle/java/queries/TextDependencyScanner.java
```diff
@@ -0,0 +1,708 @@
+/*
```
lkishalmi, line 1: OMG. Just wanted to add an ANTLR4 Lexer and Parser for simple Groovy based Gradle projects. Anyway, I still have to do that, so this is welcome!

sdedic: Hm, I am working on output artifacts support and for that I'd need some other scanner, at least for simple name=value assignments to various task properties. I plan to supply the list of _possible_ property names (and structure members) from introspecting the tasks / extensions in the project, so the scanner could better ignore 'the unknown' stuff without having to be a full parser.
Stay tuned, it's still in very rough state

java/gradle.java/nbproject/project.properties
sdedic: https://netbeans.apache.org/wiki/DevFaqImplementationDependency.asciidoc -- it's a way how to autoincrease spec version when the implementation dependency changes. I will move it back when I finalize the `project.dependency` module (based on feedback / findings when implementing Gradle, I'd allow some laziness, still not unsure how).

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
```diff
@@ -206,6 +213,29 @@ public String getLicense() {
...
+     * @param projectPath Gradle project path
+     * @return GAV coordinates, or {@code null}
+     * @since 2.27
+     */
+    public String findProjectGav(@NonNull String projectPath) {
+        if ("".equals(projectPath) || getPath().equals(projectPath)) {
+            String n = getName();
+            String g = getGroup();
+            String v = getVersion();
+            if (n == null || "".equals(n) || g == null || "".equals(g) || v == null || "".equals(v) || "unspecified".equals(v)) {  // NOI18N
```
ppisl, line 232: Just minor. Using n.isEmpty(), q.isEmtpy() and v.isEmpty would be better here than the equals approach. The null check is done in the condition anyway.

ppisl (approved):

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
sdedic: addressed in 2db82c7

extide/gradle/src/org/netbeans/modules/gradle/api/GradleBaseProject.java
sdedic: addressed in 2db82c7

entlicher (approved):

sdedic:
2 approvals, only `DebugRubyTest` and `DebugBaseTest` tests are failing which are unrelated to this PR. Going to merge in a few hours.

lkishalmi (approved):

Issue #4500, opened 2022-08-12
static imported methods slow down auto completion


mbien:
NB 15

This is somewhat related to #4140 (fixed via #4142 / NB 15) but the codepath is different (see flamegraph).

when taking the FieldsOfJoy test as baseline and adding 10k methods to the fields, completion performance suffers significantly (~10s). If the methods are not there, completion finishes instantly thanks to #4142.

However: this is specific to static imports, it can't be reproduced with normal usage

```java
//import static test.fields.FieldsOfJoy10k.*;   // <- this is always fast with NB 15
import static test.fields.FieldsOfJoy10k_many_methods.*;

public class Fields {
    public static void main(String[] args) {
        System.out.println( );  // use completion here, this is slow!
        System.out.println(test.fields.FieldsOfJoy10k_many_methods.);  // this is fast!
    }
}
```
thanks to @SirIntellegence for discovering this (https://github.com/apache/netbeans/pull/4440#issuecomment-1213492886).

async profiler flamegraph shows that a significant time is spent in `JavaCompletionTask.conflictsWithLocal` (this isn't the full graph, see attached file below)
![many_methods](https://user-images.githubusercontent.com/114367/184451744-31568e87-1df1-47fd-be11-2772fec5fea6.png)

full graph:
[many_methods.html.txt](https://github.com/apache/netbeans/files/9330116/many_methods.html.txt)

mbien:
Hi @neilcsmith-net I explain it here so that I don't add more noise to the PR discussion ;)

I did three things at once, i reviewed 4440 where the contributor noticed this issue, when I noticed an even worse issue which I decided to fix right away via 4501. So I profiled both at once, fixed one thing and wrote the other down to this issue here which I didn't plan to fix. But the next day when I quckly looked at the underlying code for the first time, I saw that this is actually trivial to fix - so I did that too. I usually wouldn't file issues just so that i can fix them right away via a PR, I think writing this issue took more time than the fix :)

PR #4501, opened 2022-08-13
org.netbeans.modules.java.hints.bugs.Unused doesn't cancel properly

mbien:
based on delivery for the slim chance that this gets a review

 - streams are not very good at canceling
 - by using a plain loop we can simply break out of it
 - further: `findUnused` is actually very fast, `convertUnused` is the slow part and wasn't cancelable before, the original code might have worked if `findUnused` returned a stream instead of a list and everything would be one pipeline (this would make the fix more complicated)

issue:

 - a synthetic file with 10k fields pins a thread for about 20s
   at 100%
 - most editor features don't work during that time, e.g completion
 - hint is enabled by default

discussion:

noticed this while benchmarking other issues (#4500)

#3886 mentioned the poor performance of `org.netbeans.modules.java.hints.bugs.Unused` and also that it isn't cancelable.
#4204 attempted to fix some of it, however I am not so sure that this worked.

test:
[FieldsOfJoy10k_Unused_hint.java.txt](https://github.com/apache/netbeans/files/9330547/FieldsOfJoy10k_Unused_hint.java.txt)

scroll down to:
```java
    // hints -> probable bugs -> unused element
    // org.netbeans.modules.java.hints.bugs.Unused#unused
    private void something() {
        this.     // code change triggers hint and pins cpu to 100%,
                  // this blocks completion etc. does cancel work?
    }
```

`Unused#convertUnused `/ `JavaFixUtilities#safelyRemoveFromParent` use most of the time and they were not cancelable.

unused_element_hint_baseline

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unused.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unused.java
@@ -19,6 +19,7 @@
 package org.netbeans.modules.java.hints.bugs;
 
 import com.sun.source.tree.Tree.Kind;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import javax.lang.model.element.ElementKind;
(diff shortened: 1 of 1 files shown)


jlahoda (approved):
Looks reasonable to me.

mbien:
@jlahoda thanks for review on such short notice!

PR #4503, opened 2022-08-14
Some support for TOML files

lkishalmi:
This support provides the following:
* lexer based syntax coloring
* useful keybindings
* some code completion based on the keys in the current document. 

image


ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLexer.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public TomlLexer(LexerRestartInfo<TomlTokenId> info) {
+        this.tokenFactory = info.tokenFactory();
+        try {
+            this.lexer = new org.tomlj.internal.TomlLexer(new LexerInputCharStream(info.input()));
+            if (info.state() != null) {
+                ((LexerState) info.state()).restore(lexer);
+            }
+        } catch (Throwable ex) {
+            ex.printStackTrace();
```
mbien, line 47: catching Exception would be better, since Throwable includes Error too which is better not to be caught.

handle with `Exceptions.printStackTrace`?

ide/languages.toml/nbproject/project.properties
```diff
@@ -0,0 +1,19 @@
+is.eager=true
+# Licensed to the Apache Software Foundation (ASF) under one
```
mbien, line 2: value above license

mbien (commented):
awesome! We should use this PR as an up2date example of writing a language support module. 

I quickly looked through it as first pass (no real review) and added two minor comments inline. Good job!

I think it would be good to have at least a basic "black box" test for this which parses a file and checks that nothing explodes.

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLexer.java
lkishalmi: Well, I probably should return an ERROR token, than maybe make sure than an EOF is returned. 
The TOML lexre was able to produce an ArrayIndexOutOfBounds exception and IAE on wrong input, though it might be because I was wrong about the state saving. Anyway. Thanks for spotting this one. Need to check what shall I really do.

ide/languages.toml/nbproject/project.properties
lkishalmi: Yeah, that's bad. I need to patch the UI action that put it there one day.

lkishalmi:
As of tests, I still need to study the Yaml language support how to write them. There could be some edge cases in the keyboard handling and the code completion.

Chris2011:
Great work, will help me also to extend other languages and add more :)

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlCompletionProvider.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private class TomlCompletionQuery extends AsyncCompletionQuery {
+
+        @Override
+        protected void query(CompletionResultSet resultSet, Document doc, int caretOffset) {
+            Set<String> candidates = new HashSet<>();
+            try {
+                int prefixOfs = keyPrefixOffset(doc, caretOffset);
+                String prefix = doc.getText(prefixOfs, caretOffset - prefixOfs);
```
sdedic, line 64: Q: is `doc` read-locked at this time ?

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlCompletionProvider.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final Set<TomlTokenId> DOT_OR_KEY = EnumSet.of(TomlTokenId.DOT, TomlTokenId.KEY);
+
+    static int keyPrefixOffset(Document doc, int offset) throws BadLocationException {
+        AbstractDocument d = (AbstractDocument) doc;
+        try {
+            d.readLock();
+            TokenHierarchy th = TokenHierarchy.get(doc);
+            TokenSequence<TomlTokenId> ts = th.tokenSequence();
+            ts.move(offset);
+            ts.movePrevious();
```
sdedic, line 117: Maybe check the result of `movePrevious`; indicates end(begin)-of-token stream

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlTypedTextInterceptor.java
```diff
@@ -0,0 +1,177 @@
...
+        ts.move(context.getOffset());
+        ts.movePrevious();
+        while ((ts.token() != null) && STRING_OR_WS.contains(ts.token().id())) {
+            ts.movePrevious();
+        }
+        return (ts.token() != null) && TomlTokenId.ML_STRING_START == ts.token().id();
+    }
+
+    private static int quotesInLine(Context context, char quote) throws BadLocationException {
+        BaseDocument doc = (BaseDocument) context.getDocument();
```
sdedic, line 139: Use `LineDocument`, no dependency on editor.lib needed.

ide/languages.toml/src/org/netbeans/modules/languages/toml/layer.xml
```diff
@@ -0,0 +1,168 @@
...
+                            <attr name="instanceCreate" methodvalue="org.netbeans.modules.options.editor.spi.CustomizerFactories.createDefaultTabsAndIndentsCustomizerFactory"/>
+                            <attr name="previewTextFile" stringvalue="org/netbeans/modules/languages/toml/TomlExample.toml"/>
+                            <attr name="position" intvalue="100"/>
+                        </file>
+                    </folder>
+                </folder>
+            </folder>
+        </folder>
+    </folder>
+    <!--folder name="Templates">
```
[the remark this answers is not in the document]
lkishalmi: Resolved with the recent changes.

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlTypedTextInterceptor.java
lkishalmi: Good tip thanks!

ide/languages.toml/nbproject/project.properties
```diff
@@ -0,0 +1,19 @@
+is.eager=true
```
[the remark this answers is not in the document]
lkishalmi: Actually without that one the commit validation fails, even locally. It seems to be a valid issue (something with ergonomics, if I remember well). Checked the Yaml, yes that module is Regular, so I removed the eager, and try to find another way to fix the commit validation tests. 

Any hints?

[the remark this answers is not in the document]
mbien: is the ergonomics module still loaded during IDE use? I vaguely remember that it had something to do with usage statistics which have been disabled.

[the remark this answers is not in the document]
lkishalmi: Yes it does. If you install from zip or Snap packages, the IDE starts with minimal functionality, then the rest is either switched on by ergonomics or directly form the plugins dialog.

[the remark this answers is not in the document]
mbien: oh right! I confused that with something else - my mistake.

PR #4509, opened 2022-08-16
Refactor Grade Editor related functionality into a gradle.editor module

lkishalmi:
Dear all,

This is a part of an effort trying to split up the ```gradle``` module into smaller parts with less dependencies. So this would start a series.

For me this is the second take of the work, as at first I started with moving the editor related parts to a new module. It was just too late when I realized that the ```GradleReport``` class never reached the ```api``` package (though we have an entry on that in the ```apichanges.xml```)

I broke up the work for meaningful units in the commits, so it maybe easier to review this change commit by commit.

I don't necessary like the exploiting to the ```NbGradleProjectImpl.ACCESSOR```, though that was handy to hide the constructors from the API users.

I'm not fun of the names ```GradleReport``` maybe Ok, but ```GradleProblemReport``` is an ehh.

So suggestions, and feedback are required! Till it's a draft, the Javadoc stuff are not ready.


lkishalmi:
Closed in favor of: #4518

PR #4512, opened 2022-08-17
Alternative Refactor for gradle.editor

lkishalmi:
This one is an alternative to #4509 

Here the ```GradleReport```s are attached to the existing ```GradleBaseProject```. 

Probably a bit more streamlined than #4509, I mean less weird stuff.


lkishalmi:
Closed in favor of: #4518

PR #4517, opened 2022-08-19
[Preview] Gradle project introspection

sdedic:
This is a WIP **and will not be merged** - I'll create a new PR for the actual merge. Ignore the number of commits as this work is based on another not-yet-merged PR (worked in parallel on both). This PR is meant as a heads-up or preview.

The original goal was to 'find out what file(s) / artifact(s) a project produces". I started with JAR initially ... and then followed up by support for micronaut / native-image. And then things got very complicated. With all the Gradle convetions, defaulting from task to extension properties ...

Initially I wanted to extract metadata out of the gradle build: If the IDE knows what tasks are there, what are properties of those tasks and - if the value is structured - what are the nested properties etc, the IDE might be able to quick-parse the buildscript, ignore closures and find out property values. The same for extensions.
Then I've realized that I could eventually extract the computed values, too - it is possible to ask provider NOT to run tasks, but return values as they are computed at the start of the build. Not perfect, but sufficient.

So in this PR you can find & comment the code that sucks information out of the gradle build. With this support, finding out the directory and filename of a JAR produced is a simple matter of:

```
            BuildPropertiesSupport props = BuildPropertiesSupport.get(project);
            Property p = props.findTaskProperty("jar", "archiveFileName");
            if (p != null && p.getStringValue() != null) {
                filename = p.getStringValue();
            } else {
                // Some guess based on project name and version
            }
            p = props.findTaskProperty("jar", "destinationDirectory");
            if (p != null && p.getStringValue() != null) {
                dir = p.getStringValue();
            }
```

... and it WORKS !

In addition, I was able to enumerate **all the participating Plugins**, so we can get rid of the hardcoded list of plugins to check.

### DIFF
--- a/enterprise/micronaut/nbproject/project.xml
+++ b/enterprise/micronaut/nbproject/project.xml
@@ -69,6 +69,32 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
+                    <code-name-base>org.netbeans.modules.maven</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>2.154</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.project.dependency</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <implementation-version/>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.maven.embedder</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <release-version>2</release-version>
+                        <specification-version>2.72</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
                     <code-name-base>org.netbeans.api.progress</code-name-base>
                     <build-prerequisite/>
                     <compile-dependency/>
@@ -382,6 +408,39 @@
                     </run-dependency>
                 </dependency>
             </module-dependencies>
+            <test-dependencies>
+                <test-type>
+                    <name>unit</name>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.libs.junit4</code-name-base>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.nbjunit</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.java.source.base</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.editor.mimelookup.impl</code-name-base>
+                        <recursive/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.openide.modules</code-name-base>
+                        <compile-dependency/>
+                        <test/>
+                    </test-dependency>
+                    <test-dependency>
+                        <code-name-base>org.netbeans.modules.projectapi.nb</code-name-base>
+                        <recursive/>
+                        <compile-dependency/>
+                    </test-dependency>
+                </test-type>
+            </test-dependencies>
             <public-packages/>
             <class-path-extension>
                 <runtime-relative-path>ext/spring-boot-configuration-metadata-2.4.4.jar</runtime-relative-path>
--- a/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
+++ b/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/BuildActions.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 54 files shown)


sdedic:
@lkishalmi please look at the thing from the conceptual standpoint; it seems that it is possible to deliver quite a lot of pre-computed values "as they are seen by gradle" to nb IDE for processing.

To see captured data logged, use ` -J-Dorg.netbeans.modules.gradle.loaders.LegacyProjectLoader.level=400 ` when running NetBeans.

The attachment is a capture from Micronaut sample project reload:
[gradle.log](https://github.com/apache/netbeans/files/9385089/gradle.log)

lkishalmi:
Ugh.
So at first the size of PR scared me, then I realized that this contains #4495.

It is Ok to create a PR in your personal repo and ask for review there. I guess I need to review  WIP and WIP 2 commits.?

sdedic:
> It is Ok to create a PR in your personal repo and ask for review there. I guess I need to review WIP and WIP 2 commits.?

Yes, just those two. The most interesting part is the `NbProjectInfoBuilder` in NB gradle tooling.

lkishalmi:
Well, well...
I see some of good improvements on the code, though some of these changes are a bit disturbing.
And that is the auto-discovery of the task properties.

Gradle evolved quite a lot during the last years. One of the direction they went is doing the lazy initialization of the tasks, by using task registration instead or actual instance creation. Though when you request a task, it will be instantiated. People are very creative to create really poor quality Gradle build scripts, so task instantiation can be really expensive.

If you would like to experiment with this, then probably shall create a feature branch. Create a separate TaskInfoModel add it's own cache, and probably some loading infrastructure. I would not make the NbProjectInfo more bloated.

To find out the directory and filename of the main jar is really easy right now (not fancy and shiny, but works in most cases):

```java
GradleJavaProject gjp = GradlJavaProject.get(project);
File jar = gjp.getMainJar();
```

sdedic:
> 
> ```java
> GradleJavaProject gjp = GradlJavaProject.get(project);
> File jar = gjp.getMainJar();
> ```

Well, the  main issue is not to find exactly the jar; but 
- what the `build` actually does. Different plugins glue different tasks to `build` lifecycle one. For this, I've analyzed the inter-Task dependencies. It can be still approximated from just known Plugins though. 
- what the `native-image` output is - in the case native buld. Not just the jar. 

The idea to separate the loading process seems  good, as it duplicates the infrastructure.  I would initially (since I need some results soon) drop the task _properties_ introspection (assuming separate APIs for loading the data) - as actually the task proeprties default to (combination of) proiject extension values, it's a good approximation and a good tradeoff for me.

Are you sure we're not instantiating Tasks even with the current implementation ;) ?

sdedic:
Update: I made a quick check - put a few breakpoints into e.g. `Jar` task ... the fact is, that the tasks **are instantiated already**, because of [detection of task properties](https://github.com/apache/netbeans/blob/master/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java#L223) which is present in the `master`. 
Even some more weird things. like `computeManifest` execute, thanks to 
```
jar.manifest(manifest -> {
```
https://github.com/micronaut-projects/micronaut-gradle-plugin/blame/f3687ae6b2d8d830b7742788577b80351fa88ffc/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java#L157

... the original impl (2 years ago) used `project.afterEvaluation` - is that one of the typical things the users do mistakes with early evaluation ?

lkishalmi:
Well, I'm not sure that the instantiation does not happen with the current implementation. I'm sure it happens for the jar and war task if they are exists as well as the compile tasks.

sdedic:
> Well, I'm not sure that the instantiation does not happen with the current implementation. I'm sure it happens for the jar and war task if they are exists as well as the compile tasks.

OK, I'll try with more examples, but the task creation stacktrace I saw originated from enumeration in `detectPlugins` -- I suspect that all tasks were created, not just jar/war/ear.

lkishalmi:
Back to Our original issue. Aren't you looking for the outgoing variants by
chance? Do a gradlew oV in your project and see the output.

sdedic:
> Back to Our original issue. Aren't you looking for the outgoing variants by chance? Do a gradlew oV in your project and see the output.
> […](#)

The PR targets two things. First, the ability to find out product(s) of the project. Outgoing variants could be useful, but on the sample project (micronaut project, that includes native-image build plugin), the `gradlew oV` output does not contain an information that such thing is produced:
```
sdedic@sdedic-nb4:oci/microdemo5$ gradle oV  | grep build/
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on -Dswing.aatext=true
    - build/libs/microdemo-0.1.jar (artifactType = jar)
    - build/libs/microdemo-0.1.jar (artifactType = jar)
          - build/classes/java/main (artifactType = java-classes-directory)
          - build/resources/main (artifactType = java-resources-directory)
    - build/libs/microdemo-0.1-all.jar (artifactType = jar)
    - build/libs/microdemo-0.1-all.jar (artifactType = jar)
    - build/test-results/test/binary (artifactType = directory)
```
There's no `build/native/nativeCompile/microdemo` in the output, althought the plugin is configured. That's natural, since it shows variant products for individual consumable configurations -- the `native-image` plugin does not create any specific configuration (maybe a deficiency ? especially in the .dll/.so scenario it could be).
But as I see the output, it seems that the implementation of the project artifact query could be greatly improved following the OutgoingVariants report implementation - a theme for another PR!

sdedic:
Let's focus back on the topic of gathering information. To build `native-image` support, I'd need some information from the extension / task that is configured for the project. I agree that instantiating tasks (= user code) could be dangerous. 

It seems that instantiation happens every time, just because we're **enumerating tasks** from the buildscript, so we can (for example) present them, for example, in the `Navigator` window for a gradle project.
Not only Jar task, which will be eventually executed as the part of buildscript, but all of them - e.g. `OutgoingVariantsReportTask` (`./gradlew oV`) is instantiated.

So, if the concern that task creation is harmful is valid and important, we should remove the task listing code and corresponding feature(s) in the IDE (or implement them using another technique). Given that this feature is in the Gradle Projects module for some years already, with no significant bug reports against ... I would not go that way: it might be bad (this is why we have 'Do you trust the build script' question, among other reasons)  ... but seems acceptable.

If there Tasks already instantiate ... why don't exploit the data then, so the IDE can evaluate using the most accurate values ?

lkishalmi:
I'm finishing my ANTLR works. After that I'm going to give this another try.

sdedic:
Obsoleted by #4729

PR #4518, opened 2022-08-20
Refactor Grade Editor Related Functionality into a gradle.editor Module

lkishalmi:
The finalization of the refactor from #4512


mbien (approved):
looks good as far as I can tell.

PR #4523, opened 2022-08-21
Improvements to incremental compilation of NetBeans

jtulach:
After few months I tried following on my repository
```
$ git pull 
$ ant build
```
and run into bunch of compilation errors. These are the changes I had to make to let the incremental build pass.

### DIFF
--- a/ide/c.jcraft.jsch/nbproject/project.xml
+++ b/ide/c.jcraft.jsch/nbproject/project.xml
@@ -27,15 +27,18 @@
             <module-dependencies>
                 <dependency>
                     <code-name-base>bcprov</code-name-base>
+                    <build-prerequisite/>
                     <run-dependency/>
                 </dependency>
                 <dependency>
                     <code-name-base>com.jcraft.jzlib</code-name-base>
+                    <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency/>
                 </dependency>
                 <dependency>
                     <code-name-base>libs.c.kohlschutter.junixsocket</code-name-base>
+                    <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
@@ -44,6 +47,7 @@
                 </dependency>
                 <dependency>
                     <code-name-base>org.netbeans.libs.jna</code-name-base>
+                    <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
@@ -52,6 +56,7 @@
                 </dependency>
                 <dependency>
                     <code-name-base>org.netbeans.libs.jna.platform</code-name-base>
+                    <build-prerequisite/>
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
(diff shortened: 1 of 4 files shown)


matthiasblaesing (commented):
I saw the problems regarding the annotation processors also in the past, but as I commonly build after running a clean (`git clean -f -x -d`) I put it of to investigate. Thank you.

Can you describe the requirement for adding the `build-prerequisite` entries in the `c.jcraft.jsch` module? The module itself only repackages `jsch` as an OSGI module. According to the documentation `compile-dependency` implied `build-prerequisite`:

https://github.com/apache/netbeans/blob/a0f621d00b979dd1f577ea4dfd7c62660a760707/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/resources/nb-module-project2.xsd#L57

jtulach:
> Can you describe the requirement for adding the `build-prerequisite` entries in the `c.jcraft.jsch` module? 
> The module itself only repackages `jsch` as an OSGI module. According to the documentation `compile-dependency` implied 

I got a failure that `c.jcraft.jsch` module cannot be compiled as it requires some other module in a version `x.y`, but only older `a.b` version was found. Requiring the build order fixes that problem. Maybe removing `<compile-dependency/>` would fix the problem as well, if there is no code to compile.

matthiasblaesing:
I tried to see the problem myself and indeed it is reproducible with:

```
git clean -f -x -d
git checkout 5573b9c9e8794ba43374347606509a5a4206a00a^
ant
git checkout 5573b9c9e8794ba43374347606509a5a4206a00a
ant
```

From my perspective this is not an issue in the module, but in the build-system. A module declares a versioned dependency and the build system (according to this test) the module to be present for this (although only the minimal runtime version is declared) and the build system does not ensure the module is build - this is broken.

Anyway. For the concrete case the minimal fix is adding `<build-prerequisite />`  is enough.

I'm not sure, though that it is a good idea to do this, as next time the problem is a different module.

sdedic (approved):

sdedic:
> I'm not sure, though that it is a good idea to do this, as next time the problem is a different module.

Let's do the minimal fix now, and talk/fix buildsystem as the next step (another PR)

mbien:
conflicts would need to be resolved here to get this into NB17

sdedic:
@jtulach - still relevant ?

jtulach:
> I tried to see the problem myself and indeed it is reproducible with:

+1

> Anyway. For the concrete case the minimal fix is adding `<build-prerequisite />` is enough.

+1

> I'm not sure, though that it is a good idea to do this, 

I see adding `<build-prerequisite>` as a proper fix. It says (in my opinion) _"make sure that module is **properly** compiled"_. Which it is not, if it is outdated.

> as next time the problem is a different module.

Probably. This kind of incremental issues is hard to spot and get completely right.

neilcsmith-net:
@matthiasblaesing @mbien can we just merge #5617 and remove the option from the UI!

matthiasblaesing:
#5617 is already added my +1 on that PR, so sure ;-)

mbien:
> @matthiasblaesing @mbien can we just merge https://github.com/apache/netbeans/pull/5617 and remove the option from the UI!

fine I merge it :)

I just had the worry that this might encourage merges without *any* kind of squashing. Lets see if it improves the situation

PR #4525, opened 2022-08-22
Gradle hide empty generated folders

lkishalmi:
When added the support of Generated source directories in Gradle project, the project view become a bit crowded, as the generated folders are always generated by Gradle.

Here there are two empty folder one for java and one for groovy.
image

Same project with the new code.
image

While implementing the change, I discovered a bug which prevented the freshly opened project to listen on resource changes. That one is the fist commit.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/NbGradleProject.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/NbGradleProject.java
@@ -30,6 +30,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.logging.Logger;
 import java.util.prefs.Preferences;
 import javax.swing.Icon;
 import javax.swing.ImageIcon;
@@ -56,6 +57,8 @@ import org.openide.util.Utilities;
  * @author Laszlo Kishalmi
  */
 public final class NbGradleProject {
+
+    static final Logger LOG = Logger.getLogger(NbGradleProject.class.getName());
 
     /**
      * As loading a Gradle project information into the memory could be a time
@@ -275,6 +278,10 @@ public final class NbGradleProject {
         //Never listen on resource changes when only FALLBACK quality is needed
         if ((project.getAimedQuality() == Quality.FALLBACK) && !elevateQuality) return;
         synchronized (resources) {
+            if (!resources.isEmpty()) {
+                LOG.warning("Gradle ResourceWatcher Leak: " + resources); //NOI18N
+                resources.clear();
+            }
             Collection<? extends WatchedResourceProvider> all
                     = project.getLookup().lookupAll(WatchedResourceProvider.class);
             for (WatchedResourceProvider pvd : all) {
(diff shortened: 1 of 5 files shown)


extide/gradle/src/org/netbeans/modules/gradle/api/NbGradleProject.java
```diff
@@ -255,34 +255,37 @@ private void fireProjectReload() {
...
-            resources.addAll(pvd.getWatchedResources());
-        }
-        for (File resource : resources) {
-            try {
-                FileUtil.addFileChangeListener(FCHSL, resource);
-            } catch (IllegalArgumentException ex) {
-                assert false : "Something is wrong with the resource handling";
+        if ((project.getAimedQuality() == Quality.FALLBACK) && !elevateQuality) return;
+        synchronized (resources) {
+            Collection<? extends WatchedResourceProvider> all
```
sdedic, line 278: I'd clear `resources` at the start.

java/gradle.java/src/org/netbeans/modules/gradle/java/classpath/GradleSourcesImpl.java
```diff
@@ -159,7 +160,7 @@ public synchronized SourceGroup[] getSourceGroups(String type) {
         for (SourceType st : stype) {
             Set<File> processed = new HashSet<>();
             for (String group : gradleSources.keySet()) {
-                Set<File> dirs = gradleSources.get(group).getSourceDirs(st);
+                Set<File> dirs = filterEmptyGeneratedDirs(gradleSources.get(group).getSourceDirs(st));
```
sdedic, line 163: Is it OK to filter the empty dirs from the data layer ? I'd rather leave them discoverable (as they are valid source roots), but filter them from the UI only.

lkishalmi: Good insight! Thanks! Moving it to the UI!

java/gradle.java/src/org/netbeans/modules/gradle/java/nodes/SourcesNodeFactory.java
```diff
@@ -79,7 +97,8 @@ public List<SourceGroup> keys() {
             ret.addAll(Arrays.asList(srcs.getSourceGroups(JavaProjectConstants.SOURCES_TYPE_JAVA)));
             ret.addAll(Arrays.asList(srcs.getSourceGroups(GradleSourcesImpl.SOURCE_TYPE_KOTLIN)));
             ret.addAll(Arrays.asList(srcs.getSourceGroups(JavaProjectConstants.SOURCES_TYPE_RESOURCES)));
-            ret.addAll(Arrays.asList(srcs.getSourceGroups(GradleSourcesImpl.SOURCE_TYPE_GENERATED)));
+            generatedGroups = Arrays.asList(srcs.getSourceGroups(GradleSourcesImpl.SOURCE_TYPE_GENERATED));
```
sdedic, line 100: nitpick: `generatedGroups` access is unsynchronized; here, the value should be assigned first to local, then to generatedGroups + added to ret, to be consistent (within this thread). `generatedGroups` could be possibly `volatile`

sdedic (approved):

PR #4533, opened 2022-08-25
Implementation of auto completion, formatting adjustment and providing hints for JEP 405 record patterns

MegJayan:
JEP 405 Record Patterns : Implement auto completion, adjust formatting and provide hints.


java/java.source.base/src/org/netbeans/modules/java/source/TreeShims.java
```diff
@@ -60,6 +63,8 @@ public class TreeShims {
 //    public static final String NULL_LITERAL = "NULL_LITERAL"; //NOI18N
 //    public static final String PARENTHESIZED_PATTERN = "PARENTHESIZED_PATTERN"; //NOI18N
 //    public static final String GUARDED_PATTERN = "GUARDED_PATTERN"; //NOI18N
+    public static final String DECONSTRUCTION_PATTERN = "DECONSTRUCTION_PATTERN";
```
lahodaj, line 66: Please note that after https://github.com/apache/netbeans/pull/4467, it should be possible to refer to the deconstruction/record patterns directly, without the need for reflection and `TreeShims`.

MegJayan: > Please note that after #4467, it should be possible to refer to the deconstruction/record patterns directly, without the need for reflection and `TreeShims`.

Hi, yes. Have noticed that. Will commit the latest code in some time

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+        Set<String> recordPatternVarSet = new HashSet<>();
+        Map<PatternTree, List<PatternTree>> recordComponentMap = new LinkedHashMap<>();
+        DeconstructionPatternTree recordPattern = (DeconstructionPatternTree) t.getLeaf();
+        recordComponentMap = findNested(recordPattern, recordComponentMap);
+
+        for (PatternTree p : recordComponentMap.keySet()) {
+            BindingPatternTree bTree = (BindingPatternTree) p;
+            recordPatternVarSet.add(bTree.getVariable().getName().toString());
+        }
+        while (t != null && t.getLeaf().getKind() != Tree.Kind.IF) {
```
lahodaj, line 97: Note this will find an `if` arbitrarily far from the original pattern. Even in the current code, it might be something like:
```
if (true) {
    boolean b = o instanceof R1(R2 r2) && r2.v() != null;
}
```

As I read the code, the above would probably work (more or less by accident), but would not without the enclosing `if`. Also, the scope of a variable may extend beyond the current statement, like in:
```
if (!(o instanceof R1(R2 r2))) { return ; }
System.err.println(r2.v()); //OK to use r2 here
}
```

Maybe we want to find the innermost enclosing block (or innermost enclosing lambda or case, as lambdas and cases not necessarily have a block - OTOH, I suspect searching bigger chunk of the code shouldn't have functional effect, so it might be OK to simply scan the innermost block for simplicity)?

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+            t = t.getParentPath();
+        }
+        Set<TreePath> convertPath = new HashSet<>();
+        List<String> localVarList = new ArrayList<>();
+        Map<String, List<UserVariables>> userVars = new HashMap<>();
+        new TreePathScanner<Void, Void>() {
```
lahodaj, line 103: Maybe try to use `CancelableTreePathScanner` and try to cancel the process is the hint is canceled (esp. if we are going to search bigger parts of the source code).

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+        }.scan(t, null);
+        if (!convertPath.isEmpty()) {
+            Fix fix = new FixImpl(ctx.getInfo(), ctx.getPath(), convertPath, localVarList, userVars).toEditorFix();
+            return ErrorDescriptionFactory.forName(ctx, ctx.getPath(), Bundle.ERR_ConvertToNestedRecordPattern(), fix);
+        }
+        return null;
+    }
+
+    private static Map<PatternTree, List<PatternTree>> findNested(PatternTree pTree, Map<PatternTree, List<PatternTree>> recordComponentMap) {
+        if (pTree instanceof BindingPatternTree) {
```
lahodaj, line 134: Better use `pTree.getKind()`, and switch over that kind, having a `default` with an exception. `instanceof` is not necessarily a safe idiom when working with `com.sun.source.tree`.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+            recordComponentMap.put(pTree, new ArrayList<>());
+            return recordComponentMap;
+        } else {
+            DeconstructionPatternTree bTree = (DeconstructionPatternTree) pTree;
```
lahodaj, line 138: Please note we also have parenthesized patterns.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+                VariableTree v = bTree.getVariable();
+                type = (TypeElement) wc.getTrees().getElement(TreePath.getPath(t, v.getType()));
+                if (type == null || type.getRecordComponents().size() == 0) {
+                    continue;
+                }
+                outer:
+                for (RecordComponentElement recordComponent : type.getRecordComponents()) {
+                    String name = recordComponent.getSimpleName().toString();
+                    String returnType = recordComponent.getAccessor().getReturnType().toString();
+                    returnType = returnType.substring(returnType.lastIndexOf(".") + 1);
```
lahodaj, line 205: There shouldn't be a need to do String manipulation over the (String representation of the) type. Simply keep the `TypeMirror`, and then use `TreeMaker.Type(TypeMirror)`. That method should take care about printing the correct type, with correct imports, etc.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+                            Modifiers(EnumSet.noneOf(Modifier.class)), name, wc.getTreeMaker().Identifier(returnType), null)));
+                }
+                recordComponentMap.put(p, bindTree);
+            }
+
+            DeconstructionPatternTree d = (DeconstructionPatternTree) createNestedPattern((PatternTree) t.getLeaf(), wc, recordComponentMap);
+            while (t != null && t.getLeaf().getKind() != Tree.Kind.IF) {
+                t = t.getParentPath();
+            }
+            IfTree it = (IfTree) t.getLeaf();
```
lahodaj, line 231: Please see the comment about searching for the `if` above. If we would search the enclosing block, this may need to be changed to look for immediately enclosing block for the statements we need to remove.

As a note `WorkingCopy.resolveRewriteTarget` might be useful here - even if a tree is re-written several times, this method will return the last rewrite target.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
```diff
@@ -0,0 +1,261 @@
...
+            List<Tree> removeList = replaceOccurrences.stream().map(tph -> tph.resolve(wc).getLeaf()).collect(Collectors.toList());
+            for(Tree tree : removeList) {
+                bt = wc.getTreeMaker().removeBlockStatement((BlockTree) bt, (StatementTree) tree);
+            }
+            wc.rewrite(it, wc.getTreeMaker().If(wc.getTreeMaker().Parenthesized(cond), bt, it.getElseStatement()));
+        }
+    }
+
+    private static PatternTree createNestedPattern(PatternTree pTree, WorkingCopy wc, Map<PatternTree, List<PatternTree>> map) {
+        if (pTree instanceof BindingPatternTree) {
```
lahodaj, line 244: Please see the comment about using `instanceof` on the trees.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPatternTest.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        HintTest.create()
+                .input("package test;\n"
+                        + "record Rect(ColoredPoint upperLeft,ColoredPoint lr) {}\n"
+                        + "record ColoredPoint(Point p, Color c) {}\n"
+                        + "record Point(int x, int y){}\n"
+                        + "enum Color {RED,GREEN,BLUE}\n"
+                        + "public class Test {\n"
+                        + "    private void test(Object o) {\n"
+                        + "        if (o instanceof Rect(ColoredPoint ul, ColoredPoint lr)) {\n"
```
lahodaj, line 47: One thing to note here is that for JDK 20, the named record patterns will likely be dropped, and only "unnamed" will remain. Might be reasonable to reflect that here even now.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToRecordPattern.java
```diff
@@ -0,0 +1,190 @@
...
+        if (Utilities.isJDKVersionLower(RECORD_PATTERN_PREVIEW_JDK_VERSION) && !CompilerOptionsQuery.getOptions(ctx.getInfo().getFileObject()).getArguments().contains("--enable-preview")) {
+            return null;
+        }
+        ElementKind kind = ctx.getInfo().getTrees().getElement(ctx.getVariables().get("$typeI0")).getKind();
+        if (kind.equals(ElementKind.RECORD)) {
+            Set<TreePath> convertPath = new HashSet<>();
+            Set<String> localVarList = new HashSet<>();
+            localVarList.add(ctx.getInfo().getTrees().getElement(ctx.getVariables().get("$expr")).getSimpleName().toString());
+            Map<String, String> varNames = new HashMap<>();
+            new TreePathScanner<Void, Void>() {
```
lahodaj, line 94: Nit: possibly `CancelableTreePathScanner` (and use `HintContext.isCanceled()` to cancel).

java/java.source.base/src/org/netbeans/modules/java/source/pretty/VeryPretty.java
```diff
@@ -2091,8 +2092,30 @@ public void visitErroneous(JCErroneous tree) {
...
+    }
+    
+    @Override
+    public void visitRecordPattern(JCRecordPattern tree) {
+        print((JCExpression) tree.deconstructor);
+        print("(");
+        Iterator<JCPattern> it = tree.nested.iterator();
+        while (it.hasNext()) {
+            JCPattern pattern = it.next();
+            if (pattern instanceof JCBindingPattern) {
```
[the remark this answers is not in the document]
MegJayan: @lahodaj : doAccept(pattern, true) takes a pattern of type DocTree right? Is it proper to use this here as we are not dealing with patterns inside documentation syntax tree?

java/java.source.base/src/org/netbeans/modules/java/source/pretty/VeryPretty.java
```diff
@@ -2091,8 +2092,30 @@ public void visitErroneous(JCErroneous tree) {
...
     public void visitTree(JCTree tree) {
-	print("(UNKNOWN: " + tree + ")");
-	newline();
+        print("(UNKNOWN: " + tree + ")");
+        newline();
+    }
+    
+    @Override
+    public void visitRecordPattern(JCRecordPattern tree) {
+        print((JCExpression) tree.deconstructor);
```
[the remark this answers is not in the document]
MegJayan: addressed

java/java.source.base/src/org/netbeans/modules/java/source/pretty/VeryPretty.java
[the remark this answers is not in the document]
MegJayan: addressed

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
MegJayan: addressed, modified to use switch case

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
MegJayan: handled

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
MegJayan: added case for parenthesized and default label

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPattern.java
MegJayan: addressed

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/jdk/ConvertToNestedRecordPatternTest.java
MegJayan: this will be handled as part of JDK 20 changes

lahodaj (approved):
The hint and related parts seem sensible to me, but it might be better if @dbalek took a look at the completion stuff.

dbalek (approved):
Looks fine.

PR #4537, opened 2022-08-26
Cleanup clone usage

BradWalker:
Remove usage of clone() when a copy constructor will work.
    
This reduces the use of reflection in the VM. This has the net effect of making
the application more efficient.

---

### DIFF
--- a/contrib/groovy.grails/test/unit/src/org/netbeans/modules/groovy/grails/api/GrailsRuntimeTest.java
+++ b/contrib/groovy.grails/test/unit/src/org/netbeans/modules/groovy/grails/api/GrailsRuntimeTest.java
@@ -106,7 +106,7 @@ public class GrailsRuntimeTest extends NbTestCase {
         desc.getProps().setProperty("wrong", "wrong");
         assertEquals(props, desc.getProps());
 
-        String[] correctArgs = args.clone();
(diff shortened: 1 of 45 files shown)


contrib/groovy.grails/test/unit/src/org/netbeans/modules/groovy/grails/api/GrailsRuntimeTest.java
```diff
@@ -106,7 +106,7 @@ public void testCommandDescriptor() throws IOException {
         desc.getProps().setProperty("wrong", "wrong");
         assertEquals(props, desc.getProps());
 
-        String[] correctArgs = args.clone();
+        String[] correctArgs = new ArrayList<String>.toArray(args);
```
mbien, line 109: this doesn't compile since the `()` is missing. But what worse is that this is wrong. It would create an empty AL and copy `null` into the first element of `args` to mark the end of the (empty) array.

please use `Arrays.copyOf(args, args.length)` for example. Same applies to the other occurrences.

mbien (changes requested):

BradWalker:
This is not well thought about it.. It's problematic on several fronts.

Therefore, abandon it and come back later.

PR #4549, opened 2022-08-27
ANTLR v4 and v3 Grammar Editing Support

lkishalmi:
This is my initial work on ANTLR v4 grammar support. I do not plan something big. Maybe a structure scanner, simple code completion on Tokens and Rules. Just out of curiosity I might try to get the Java parser working with the embedded code...

@matthiasblaesing I need your wisdom on the external handling. I took advantage of the recent implementation of downloading single files with http, as I do not know anyone packaging ANTLR v4 grammar.


ide/libs.antlr4.runtime/external/binaries-list
```diff
@@ -15,3 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 E27D8AB4F984F9D186F54DA984A6AB1CCCAC755E org.antlr:antlr4-runtime:4.7.2
+93058C41A6CBB7F5CA63EDAC837F42C002DBC556 org.antlr:antlr4:4.7.2
+CD9CD41361C155F3AF0F653009DCECB08D8B4AFD org.antlr:antlr-runtime:3.5.2
```
matthiasblaesing, line 19: Why? This is about antlr4 - so why to 3 runtime?

.gitignore
```diff
@@ -108,3 +108,9 @@ derby.log
 /extide/gradle/netbeans-gradle-tooling/gradle/wrapper/gradle-wrapper.jar
 /extide/gradle/release/modules/gradle/daemon-loader/.gradle/
 /nbbuild/misc/prepare-bundles/target/
+
+# ANTLR v4 Grammar
+##################################
+/java/languages.antlr/external/*.g4
+/java/languages.antlr/external/LexerAdaptor.java
```
matthiasblaesing, line 115: Why are these files downloaded at runtime? If I'm not mistaken (3-Clause BSD), we can hold them in the repository and thus don't need a build-time download (external repositories will go away)

java/java.kit/nbproject/project.xml
```diff
@@ -237,6 +237,12 @@
                         <specification-version>2.23</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.netbeans.modules.languages.antlr</code-name-base>
+                    <run-dependency>
+                        <specification-version>1.0</specification-version>
+                    </run-dependency>
+                </dependency>
```
matthiasblaesing, line 245: Why is language support for antlr4 in the java cluster? antlr grammars can be used to generate lexers/parsers for various languages, java only one of them.

java/languages.antlr/external/binaries-list
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+59D7B72E12E4348CBB355929696B3531FA2E54EF https://raw.githubusercontent.com/antlr/grammars-v4/293c9b2a3fef07c1b549a9ac42adfd98cf9bda4a/antlr/antlr4/ANTLRv4Lexer.g4 ANTLRv4Lexer.g4
+B64F65F798215059626B0C5551576AFB2AB250C5 https://raw.githubusercontent.com/antlr/grammars-v4/293c9b2a3fef07c1b549a9ac42adfd98cf9bda4a/antlr/antlr4/ANTLRv4Parser.g4 ANTLRv4Parser.g4
+02F57464E7D6AEE602FB9F1C07A2B51231D0F462 https://raw.githubusercontent.com/antlr/grammars-v4/293c9b2a3fef07c1b549a9ac42adfd98cf9bda4a/antlr/antlr4/LexBasic.g4 LexBasic.g4
+C69EDA58DC59224F15898669B6AE84058B2D5928 https://raw.githubusercontent.com/antlr/grammars-v4/293c9b2a3fef07c1b549a9ac42adfd98cf9bda4a/antlr/antlr4/Java/LexerAdaptor.java LexerAdaptor.java
```
matthiasblaesing, line 20: As written above I don't think we should download these files.

java/languages.antlr/manifest.mf
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+OpenIDE-Module: org.netbeans.modules.languages.antlr
```
matthiasblaesing, line 2: antlr is its fourth iteration - so unless this is really support for all antlr version, I suggest to name this `org.netbeans.modules.languages.antlr4`

java/languages.antlr/build.xml
```diff
@@ -0,0 +1,64 @@
...
+        >
+            <classpath refid="cp"/>
+            <compilerarg line="${javac.compilerargs}"/>
+            <processorpath refid="processor.cp"/>
+        </nb-javac>
+        <mkdir dir="${build.dir}/parser"/>
+        <jar jarfile="${build.dir}/parser/antlr4-parser.jar">
+            <fileset dir="${build.dir}/parser-classes"/>
+        </jar>
+    </target>
```
matthiasblaesing, line 63: I don't understand why the parser is not build as part of the general build of the module. Creating a secondary artifact only makes it harder to reason over it.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrDataObject.java
```diff
@@ -0,0 +1,139 @@
...
+            position = 1500
+    )
+})
+
+@DataObject.Registration(
+        mimeType = AntlrTokenId.MIME_TYPE,
+        displayName = "#ANTLRResolver",
+        position = 304
+)
+public class AntlrDataObject extends MultiDataObject {
```
matthiasblaesing, line 110: I think the data object is unnecessary (see discussion around Yaml/Docker support)

ide/libs.antlr4.runtime/external/binaries-list
lkishalmi: It's strange but the ANTLR4 Tool needs both ANTLR3 and ANTLR4 runtimes.

.gitignore
lkishalmi: Thanks! That's the wisdom I came to seek!

java/languages.antlr/build.xml
lkishalmi: Well, that was the "easiest" way (I knew) to integrate with externals.

java/languages.antlr/manifest.mf
lkishalmi: Ok.

java/java.kit/nbproject/project.xml
lkishalmi: Well I planned to support Java actions (inline source elements) only. Right now the LanguageEmbedding is only depending on Java Language Mime-Type, so this module can be put to ide or extide clusters. Though if I'd add Java parsing I'd need more from the Java modules. That's the reason putting it there.

If we would like to be more generic, then this module could again go to ide or extide, and provide some spi-s for the supported languages to implement in their language dependent module. Right now I would be just happy if I can get the Java parsing working.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrDataObject.java
lkishalmi: Yes. TOML needs that update as well. I came across the multiview flag two days ago while I was checking what the LanguageRegistration does. Developing this module.

java/languages.antlr/build.xml
matthiasblaesing: For CSS the code generated by antlr3 is contained in the module itself. The corresponding module is ` ide/css.lib`. The grammar can be found here: `ide/css.lib/src/org/netbeans/modules/css/lib/Css3.g`. It needs to be regenerated by hand running the `generate-antlr-parser`  task.

lkishalmi:
@matthiasblaesing Please check the code generation and inclusion.

matthiasblaesing:
> @matthiasblaesing Please check the code generation and inclusion.

Looks sane to me. For the generated file I thought about the right license. I know that the implementation here aligns with the way the CSS module is build, but maybe using the BSD-3-clause license would be better for the generated java files. They are directly derived from the grammar, so it seems that should match.

If I see it correctly, the grammar is used mostly verbatim (only adding a license header), this is good and makes it easier updateable. If there would be more changes, we'd need to consider, whether or not we want to be able to update from upstream or really create a fork. But this is a discussion independent of this implementation.

matthiasblaesing:
@lkishalmi I added some ideas here:

https://github.com/matthiasblaesing/netbeans/tree/antlr4-support

Not really ground-braking, but could help you.

lkishalmi:
> @lkishalmi I added some ideas here:
> 
> https://github.com/matthiasblaesing/netbeans/tree/antlr4-support
> 
> Not really ground-braking, but could help you.

There are a lot of good stuff there! Thank you! I think I'm going to reconstruct the code here and there as two new PR-s. One for the library, and one for the support.

Till then, I'm going to work on this branch. I would like to add code completion, an antlr icon and would like to check how hard would it be to add antlr3 support.

lkishalmi:
Well this one became bigger than I'd thought. Could be improved later in many directions.
I did this to have a help to develop Terraform support.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrCompletionProvider.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public int getAutoQueryTypes(JTextComponent component, String typedText) {
+        return 0;
+    }
+
+    private static boolean isCaseSensitive() {
+        Preferences prefs = MimeLookup.getLookup(MimePath.EMPTY).lookup(Preferences.class);
+        return prefs.getBoolean(SimpleValueNames.COMPLETION_CASE_SENSITIVE, false);
```
matthiasblaesing, line 67: This might be a problem. I tested in the `Css3.g` file. That has a rule for `HASH`. I entered `ha` and got completion to `haSH` (this seems to be in line for case-insensitve completion), but then antlr complains:

```
generate-antlr-parser:
     [java] ANTLR Parser Generator  Version 3.5.2/home/matthias/src/netbeans/ide/css.lib/src/org/netbeans/modules/css/lib/Css3.g
     [java] 
     [java] error(106): /home/matthias/src/netbeans/ide/css.lib/src/org/netbeans/modules/css/lib/Css3.g:814:7: reference to undefined rule: haSH
```

So the completion created a bogus result.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrLanguage.java
```diff
@@ -0,0 +1,174 @@
...
+ * @author lkishalmi
+ */
+@NbBundle.Messages(
+        "ANTLRResolver=ANTLR Grammar"
+)
+@MIMEResolver.ExtensionRegistration(displayName = "#ANTLRResolver",
+    extension = {"g4", "g"},
+    mimeType = AntlrTokenId.MIME_TYPE,
+    position = 285
+)
```
matthiasblaesing, line 49: Please split the language definitions. Mapping antlr3 and antlr4 to the same mimetype, will make it impossible to hook on only one type of grammar. Given, that there was a reason, that antlr4 was created (I assume it is more expressive than antlr3), I can imagine such requirements. Setting the unificiation into stone will come to haunt us.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
```diff
@@ -0,0 +1,153 @@
...
+
+    public AntlrParserResult(Snapshot snapshot) {
+        super(snapshot);
+        addParseTask(snapshot, false);
+    }
+    
+
+
+    public AntlrParserResult get() {
+        while (!parsingQueue.isEmpty()) {
```
matthiasblaesing, line 68: This looked hacky to me. If I understand this correctly, this is done to evaluate imported grammar parts. I only glanced at the antlr4 documentation, but it reads to me, that the files should be parsed independently. If I remember correctly the indexer infrastructure should offer the tools to lookup data from other parsed files, i.e. when code completion is invoked in a file, that imports another, completion needs to lookup the completion from the local file and merge that with data from the index.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrTokenId.java
```diff
@@ -0,0 +1,124 @@
...
+                @Override
+                protected Collection<AntlrTokenId> createTokenIds() {
+                    return EnumSet.allOf(AntlrTokenId.class);
+                }
+
+                @Override
+                protected Lexer<AntlrTokenId> createLexer(LexerRestartInfo<AntlrTokenId> info) {
+                    FileObject fo =(FileObject) info.getAttributeValue(FileObject.class);
+                    if (fo != null) {
+                        return "g4".equals(fo.getExt()) ? new Antlr4Lexer(info) : new Antlr3Lexer(info);
```
matthiasblaesing, line 87: I saw this at least twice and I think it is a hack. It will fail hard, if the user declares a new file mapping (for example `.antlr3` => `text/x-antlr3`). In my mind this would be much cleaner, if there would be two mime-types (`text/x-antlr3` and `text/x-antlr4`).

matthiasblaesing (commented):
Thank you. It was much nicer to look at the Css3 grammer with the syntax highlighting and structure navigation. The usecases I had for using antlrworks (the "antlr3 IDE") were covered ;-).

However I left a few inline comments, where I had doubts.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrCompletionProvider.java
lkishalmi: Thanks! Right now the completion just really do completion (it only inserts and does not replace text). I'm going to check what can I do here.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrTokenId.java
lkishalmi: Yes, this is the hack behind the one language two grammar problem.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrLanguage.java
lkishalmi: Well, I've spent a lot of time on that during the weekend. I did not want to duplicate the editor preferences, so I've tried with ```text/x-antlr3+x-antlr``` and ```text/x-antlr4+x-antlr``` mime types but the CSL does not support complex mimetypes.

Will try to separate the two entirely.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
lkishalmi: Yes, that's true, and then someone has to implement the indexing support. Maybe in another iteration. I could add a TODO here.

matthiasblaesing: I had another thought and it turned out, that I was right: There is the potential that the parser integration recurses indefinitely, if two grammars refer to themselves in a circular fashion.

matthiasblaesing: I'm looking into indexing implementation.

lkishalmi: Wow! Would be great. FYI. Just committed the Language split.

lkishalmi: Added a patch for the endless loop and a TODO comment.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrCompletionProvider.java
lkishalmi: Fixed.

matthiasblaesing:
Could you please have a look here: https://github.com/matthiasblaesing/netbeans/commits/antlr4-support again. There are two commits on top of your last pushed. The first stabilizes the folding for me, which gave me many error messages in the log. The second implements "goto declaration" and completion via indexer.

I plan to add "Find Usages" via refactoring api, but I'm not sure whether I can get that done shortly, so I cut this of first.

lkishalmi:
Wow. Looks really nice! Going to integrate it later today (PDT). I still need to look at the indexing implementation, as I want to learn that. Thank you!

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrIndexer.java
```diff
@@ -0,0 +1,238 @@
...
+        seen.add(sourceFile.getName());
+        while(! toScan.isEmpty()) {
+            FileObject target = toScan.poll();
+            try {
+                qs.getQueryFactory()
+                        .file(target)
+                        .execute(AntlrIndexer.FIELD_IMPORT)
+                        .forEach(c -> {
+                            for(String value: c.getValues(AntlrIndexer.FIELD_IMPORT)) {
+                                if(seen.add(value)) {
```
lkishalmi, line 192: Here I've fixed the "recursion" detection. The seen set was not really in use for that.

lkishalmi:
Right now the indexer used only when the grammar file is part of a project and the project might need to be open (or at least open once). Otherwise the index search produce an IllegalArgumentException due to missing source roots. I try to register something in my next commit.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrIndexer.java
```diff
@@ -0,0 +1,238 @@
...
+        LinkedList<FileObject> toScan = new LinkedList<>();
+        Set<String> seen = new HashSet<>();
+        result.add(sourceFile);
+        toScan.add(sourceFile);
+        seen.add(sourceFile.getName());
+        while(! toScan.isEmpty()) {
+            FileObject target = toScan.poll();
+            try {
+                qs.getQueryFactory()
+                        .file(target)
```
lkishalmi, line 188: IllegalArgumentException can be thrown if the sourceFile does not belong to a project.

lkishalmi:
No next commit. Go to declaration works only if the project is indexed. If the grammar file is not in a project (FileOwnQuery returns null), then there won't be source roots detected, and the query would throw an IllegalArgumentException.

matthiasblaesing:
Your observation is valid, sorry that I did not test that case. I have to rethink this.

lkishalmi:
It is possible that, we just not do the index search if the file is not owned by a project, just fall back to the parsing result, or just do nothing. It would be not the common thing to have these files without project. I just happened to test NetBeans that way, opening grammars from the Favorites.

lkishalmi:
Took a chance with `FileOwnerQuery`. That's no good. Ergonomics can return a `FeatureProjectFactory$FeatureNonProject` for projects that could be supported by enabling a feature. So yes, we need something smart here.

matthiasblaesing:
I pushed an update into my branch (https://github.com/matthiasblaesing/netbeans/commits/antlr4-support).

The idea is this: The NB indexer is to limited and to big to fix the issue for antlr files. We are lucky, in that they only allow import from the local directory. With this limit in mind and considering that we have a limited number of grammar files in each directory, we should be able to get away with a local indexer. This is implemented as `AntlrLocalIndex`, which caches the parser results for the local antlr files, which should make them readily available.

Based on that modification I implemented "Where used" support.

Finally I found a problem, where I could not use the navigation entries in the structure display to jump inside the grammar files. This was fixed by the last commit.

lkishalmi:
Just looked on the code, haven't merged it. In the CACHE you put `WeakReference` into a `WeakHashMap`. Isn't it too weak for a cache? The cache entry is not referenced in any way could be gone as soon as the code execution leaves the block. Could be better to use `SoftReference` instead, so it would be kept for a while.

matthiasblaesing:
> In the CACHE you put `WeakReference` into a `WeakHashMap`. Isn't it too weak for a cache? The cache entry is not referenced in any way could be gone as soon as the code execution leaves the block. Could be better to use `SoftReference` instead, so it would be kept for a while.

The reference will be cleared once GC is invoked. I'm seeing cache hits, so no - the reference is not immediately cleared.

lkishalmi:
Well, `WeakReference` hide a caching issue (well not hide entirely, just made it difficult to tackle), where the cache entry validity has been evaluated on the `FileObject` modification date. That made go to source behave funny if there are two or more dependent grammars live in the editor modified but not saved.

In that case when we had a cache hit, the data was not invalidates as it was only available in the editor not on disk, so we got a wrong result back. Actually the cache should be updated whenever the source in the editor would change. The parsing API would call the parser in that case so we would have the result. So I made the cache and it's update to the responsibility of `AntlrParser`, so whenever a parsing is requester, it would update the cache.

lkishalmi:
It is nice to see the Find Usages refactoring action. Unfortunately I could not get that working. For some reasons I get an NPE when the `AntlrParser.getParserResult(cf)` for some yet unknown reason the `UserTask` returns with a `JavacParserResult` so no `AntlrParserResult` could be returned.

lkishalmi:
Well it would be good to shoot for merging this one. I would do that if the CI is green for this one.
@matthiasblaesing Is there anything you would add?

matthiasblaesing (commented):
Last little nitpick.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
```diff
@@ -66,6 +66,11 @@ public AntlrParserResult get() {
             parser.addParseListener(createStructureListener(fo));
             parser.addParseListener(createOccurancesListener());
             evaluateParser(parser);
+
+            // Start a second parsing phase for checking;
+            parser = createParser(getSnapshot());
+            parser.addParseListener(createCheckReferences(fo));
```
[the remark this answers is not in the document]
lkishalmi: That's true. That crossed my mind yesterday as well, but was concentrating elsewhere. Good point. Going to change that.

matthiasblaesing (approved):
Looks sane to me. I'm pretty sure, that there are issue lurking somewhere, but then which software is perfect?

PR #4550, opened 2022-08-27
Added multiview [Source|History] for Dockerfile-s

lkishalmi:
image

I've been missing the History view on Dockerfile-s for a long time. This one adds that.
Also moved the action registration out of ```layer.xml``` which would remove the always disabled ```Add...``` menu item from the popup.

### DIFF
--- a/ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileResolver.java
+++ b/ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileResolver.java
@@ -20,6 +20,9 @@ package org.netbeans.modules.docker.editor;
 
 import org.netbeans.api.annotations.common.CheckForNull;
 import org.netbeans.api.annotations.common.NonNull;
+import org.openide.awt.ActionID;
+import org.openide.awt.ActionReference;
+import org.openide.awt.ActionReferences;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.MIMEResolver;
 import org.openide.util.lookup.ServiceProvider;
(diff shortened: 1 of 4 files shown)


ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileResolver.java
```diff
@@ -20,8 +20,12 @@
 
 import org.netbeans.api.annotations.common.CheckForNull;
 import org.netbeans.api.annotations.common.NonNull;
+import org.netbeans.core.spi.multiview.MultiViewElement;
+import org.netbeans.core.spi.multiview.text.MultiViewEditorElement;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.MIMEResolver;
+import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
```
matthiasblaesing, line 28: The changed imports look unnecessary.

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
```diff
@@ -0,0 +1,134 @@
...
+        return 1;
+    }
+
+    @NbBundle.Messages("Source=&Source")
+    @MultiViewElement.Registration(
+            displayName = "#Source",
+            iconBase = "org/netbeans/modules/docker/editor/resources/docker_file.png",
+            persistenceType = 1,
+            mimeType = MIME_TYPE,
+            preferredID = "neon.source",
```
matthiasblaesing, line 127: neon? I would use "docker.source".

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    protected int associateLookup() {
+        return 1;
+    }
+
+    @NbBundle.Messages("Source=&Source")
+    @MultiViewElement.Registration(
+            displayName = "#Source",
+            iconBase = "org/netbeans/modules/docker/editor/resources/docker_file.png",
+            persistenceType = 1,
```
matthiasblaesing, line 125: I would use the named constant ` TopComponent.PERSISTENCE_ONLY_OPENED`

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
```diff
@@ -0,0 +1,134 @@
...
+            mimeType = MIME_TYPE,
+            preferredID = "neon.source",
+            position = 1
```
matthiasblaesing, line 128: I would also use 100 increments - here, as with the ActionReferences.

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
```diff
@@ -0,0 +1,134 @@
...
+        position = 303
+)
+public final class DockerfileDataObject extends MultiDataObject {
+
+    public DockerfileDataObject(FileObject pf, MultiFileLoader loader) throws DataObjectExistsException, IOException {
+        super(pf, loader);
+        registerEditor(MIME_TYPE, true);
+    }
+    @Override
+    protected Node createNodeDelegate() {
```
matthiasblaesing, line 112: This is not necessary. ` associateLookup`  is overridden to return `1`. The superclass implementation thus matches this code.

lkishalmi: Removed the whole DataObject. It seems that's not needed.

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
lkishalmi: Introduced a new constant here, otherwise I should have introduced a dependency on the windowing API just for that.

ide/docker.editor/src/org/netbeans/modules/docker/editor/DockerfileDataObject.java
lkishalmi: This whole section is a copy+paste from Yaml support. I think I have to revisit those as well.

matthiasblaesing (commented):
I did not know about `useMultiview` in `LanguageRegistration`  nice.

I would move he message definition and action references to `DockerfileLanguage.java`. The former is only used there (and was already declared there) and for the latter I would look for the languange implemenations not in the resolver, so `DockerfileLanguage.java` seems more logical to me.

For the dependency on the windowing API for the named constant: The dependency is already there, it is just a transitive dependency. If you don't want it, ok, but then I would not define a new constant, but add an inline comment with the FQN of the defining class.

matthiasblaesing (approved):
Looks sane to me.

PR #4551, opened 2022-08-27
Change YAML indent to 2 spaces by default

lkishalmi:
Well two space indented YAML is kind of trendy now in the world of everything is yaml.


sdedic (approved):

mbien (approved):
2 spaces for yamls seems to be a popular default indeed since they can grow fairly large unfortunately.

PR #4559, opened 2022-08-29
YAML Support Layer Code Cleanup

lkishalmi:
Well based on the knowledge gathered during Dockerfile multiview support, I've cleaned up the Yaml support code. Removing the need of a specific DataObject. Moved the action and mime registration into annotations.


ide/languages.yaml/src/org/netbeans/modules/languages/yaml/layer.xml
```diff
@@ -100,83 +87,6 @@
...
-                        <attr name="position" intvalue="700"/>
-                    </file>
-                    <file name="OpenAction.shadow">
-                        <attr name="originalFile" stringvalue="Actions/System/org-openide-actions-OpenAction.instance"/>
-                        <attr name="position" intvalue="100"/>
-                    </file>
-                    <file name="PasteAction.shadow">
-                        <attr name="originalFile" stringvalue="Actions/Edit/org-openide-actions-PasteAction.instance"/>
-                        <attr name="position" intvalue="500"/>
-                    </file>
```
matthiasblaesing, line 131: This is missing after this change - other file types in NetBeans offer paste - the copied file will be placed in the same folder as the target file. For consistency reasons I would keep it that way.

lkishalmi: Fixed. Thank you!

matthiasblaesing (approved):

Issue #4556, opened 2022-08-27
NetBeans should not auto-insert \n\ in Groovy triple quoted strings


blakemcbride:
### Apache NetBeans version

Apache NetBeans 15 release candidate

### What happened

In a Groovy file, when creating a triple quote string, NetBeans auto-adds \n\ at the end of each line.  This is wrong.  Groovy already does this automatically.  See the attached image.

Additionally, when the user hits the enter key, auto-indent should happen.

![pic1](https://user-images.githubusercontent.com/6467584/187050618-a573a9a7-c8b2-488c-8983-952a13cc2c97.png)
.

### How to reproduce

Create a triple-quote string in a Groovy source file.  Within the string, hit the enter key.  NetBeans auto-inserts \n\

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Fedora Linux

### JDK

17

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

lkishalmi:
Yeah. Right now what we can do is remove the ```\n``` insertion in Groovy. Probably that would be the best.

Right now the Groovy Lexer we are using could not make a difference between a simple string literal and a multi-line string literal.

ppisl:
I can take over.

ppisl:
Merged in master.

PR #4562, opened 2022-08-29
#4556 - NetBeans should not auto-insert \n\ in Groovy triple quoted strings

ppisl:
The fix is simple. Just check the start of STRING_LITERAL and decide, whether inser new line or not.


groovy/groovy.editor/src/org/netbeans/modules/groovy/editor/typinghooks/GroovyTypedBreakInterceptor.java
```diff
@@ -163,6 +163,9 @@ public void insert(MutableContext context) throws BadLocationException {
         }
 
         if (id == GroovyTokenId.STRING_LITERAL || id == GroovyTokenId.STRING_CH && offset < ts.offset()+ts.token().length()) {
+            if (id == GroovyTokenId.STRING_LITERAL && ts.offsetToken().text().toString().startsWith("\"\"\"")) {
```
lkishalmi, line 166: Good! Could you add ```"'''"``` check as well?

ppisl: Thank you for noticing. I completely forgot that Groovy has strings prefixed with '

PR #4565, opened 2022-08-30
TOML Cleanup and Improvements

lkishalmi:
Well, this one got a bit fatter I've initially planned.

It has the same cleanup applied for Yaml and Dockerfile support.

Additionally I've found the root cause of the IndexOutOfBoundException as the Lexer state was not fully saved and restored. Had to do an ugly reflection hack for that, but seems to work.

As a second commit I added the parser Syntax Error output to the UI.


ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLanguage.java
```diff
@@ -19,14 +19,90 @@
...
+            position = 300
+    ),
+    @ActionReference(
+            path = "Loaders/text/x-toml/Actions",
+            id = @ActionID(category = "Edit", id = "org.openide.actions.CopyAction"),
+            position = 400
+    ),
+    @ActionReference(
+            path = "Loaders/text/x-toml/Actions",
+            id = @ActionID(category = "Edit", id = "org.openide.actions.PushAction"),
```
matthiasblaesing, line 67: Typo?

```suggestion
            id = @ActionID(category = "Edit", id = "org.openide.actions.PasteAction"),
```

ide/languages.toml/src/org/netbeans/modules/languages/toml/LexerInputCharStream.java
```diff
@@ -55,51 +56,62 @@ public int LA(int count) {
...
         } else {
-            input.backup(count);
-            c = input.read();
+            backup(count);
+            c = read();
         }
         return c;
     }
 
+    //Marks are for buffering in ANTLR4, we do not really need them
```
matthiasblaesing, line 69: If I read the documentation correctly, we actually need buffering/seekability of the consumed stream (the documentation of CharStream declares this as required for lookahead), but the LexerInput already handles that - isn't it?

Reading further you actually buffer the whole file in memory, which kind of defeats the purpose of the `CharStream`  implementation. `getText` can only look back to places, that are protected by a mark, so the buffer is limited (assuming limited lookahead and marking).

ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLexer.java
```diff
@@ -137,21 +141,45 @@ private Token<TomlTokenId> token(TomlTokenId id) {
...
+                ARRAY_DEPTH = org.tomlj.internal.TomlLexer.class.getDeclaredField("arrayDepth");
+                ARRAY_DEPTH.setAccessible(true);
+                ARRAY_DEPTH_STACK = org.tomlj.internal.TomlLexer.class.getDeclaredField("arrayDepthStack");
+                ARRAY_DEPTH_STACK.setAccessible(true);
+            } catch (ReflectiveOperationException ex) {
+            }
+        }
 
-        LexerState(org.tomlj.internal.TomlLexer lexer) {
+        LexerState(org.tomlj.internal.TomlLexer lexer) throws IllegalArgumentException, IllegalAccessException {
```
matthiasblaesing, line 166: Using reflection is an implementation detail here - I would catch IllegalAccessException and wrap it into a RuntimeException. In the future you hopefully have access to the lexer state and thus the exceptions will not be thrown.

ide/languages.toml/src/org/netbeans/modules/languages/toml/LexerInputCharStream.java
lkishalmi: LexerInput ```readText```, ```readLength```  is scoped for the actual token being processed. CharStream ```getIndex``` and ```getText``` are work on stream level. Also ```getText``` is kind of an optional method. Fortunately it used in all Lexers. Just discovered to have a problem with the previos implementation on the Antlr lexer.

I'm tempted to look around the Lexing API and probably add a more ANTLR friendly interface. Will, see. I do not think that this would be the final implementation. It is kind of good enough for now.

lkishalmi: Well, throwing out unmarked sections of StringBuilder could be implemented one day with mark supported.

Looked around the Lexing API yesterday. Accessing the underlying buffers is not as easy as I've thought.

Issue #4545, opened 2022-08-27
Incorrect code analysis


blakemcbride:
### Apache NetBeans version

Apache NetBeans 15 release candidate

### What happened

Java variable is being used but NB says it isn't.  See two images
![pic1](https://user-images.githubusercontent.com/6467584/187008790-6575256f-c60b-4ac6-a55e-8be4a6608607.png)
![pic2](https://user-images.githubusercontent.com/6467584/187008791-8649967a-c8ee-482e-8326-fe8d33ffa50a.png)
.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Fedora Linux

### JDK

8

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

neilcsmith-net:
Confirmed. This appears to be a regression caused by #4421 as far as I can tell.  cc @mbien can you take a look?

mbien:
seems to be specific to final array fields with write ops.
```java
    private final static char[] bar1 = {'y', 'n'};  // marked unused
    private final static char[] bar2 = new char[2]; // ok

    public static void main(String[] args) {
        
        final char[] foo1 = {'y', 'n'};             // marked unused
        final char[] foo3 = new char[] {'y', 'n'};  // marked unused
        
              char[] foo2 = {'y', 'n'};             // ok
        final char[] foo4 = new char[2];            // ok
        
        char[] foo5;                                // ok
        foo5 = new char[] {'Y', 'n'};
        
        System.out.println(foo1);
        System.out.println(foo2);
        System.out.println(foo3);
        System.out.println(foo4);
        System.out.println(foo5);
        System.out.println(bar1);
        System.out.println(bar2);
        
    }
```

reverting the array handler of #4421 fixes this issue but it would break #4402 again for arrays. Can't come up with a quick way of fixing this atm. Shame we haven't noticed this earlier so we could have simply reverted the PR.

```diff
diff --git a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
index 51829b5..3c3d3ae 100644
--- a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
@@ -152,7 +152,7 @@
[...]

PR #4580, opened 2022-09-02
revert rushed unbalanced-hint fix (#4421).

mbien:
- fix for #4402 caused #4545
 - partial revert of #4421, tests are commented out, update to READ_METHODS Set is kept

@neilcsmith-net would you prefer the full PR reverted or just the actual "fix" and leave the rest as in this draft here.

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/Unbalanced.java
@@ -152,7 +152,7 @@ public class Unbalanced {
                 if (secondAccess != null) {
                     record(ctx.getInfo(), var, secondAccess);
                 }
-            } else if (!var.getModifiers().contains(Modifier.FINAL)) {
+            } else {
                 record(ctx.getInfo(), var, State.WRITE, State.READ);
             }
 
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@mbien I've forgotten where we're at with this one?  I just noticed the incorrect hint again in NB15 while working on update check.  Still need fixing for NB16?

mbien:
> @mbien I've forgotten where we're at with this one? I just noticed the incorrect hint again in NB15 while working on update check. Still need fixing for NB16?

@neilcsmith-net yes this should be reverted since my "[hotfix](https://github.com/apache/netbeans/pull/4421/files)" made the situation worse unfortunately. Since the patch also added tests and a little bit of an update to the ARRAY_WRITE, READ_METHODS and WRITE_METHODS, I was wondering if we can do a partial revert like proposed here.

Completely ok with a full revert too of course - I was only asking for feedback.

neilcsmith-net:
Your call, but probably keep the other aspects.  Slightly hard to see what's kept - with a more complex PR could have two commits - full revert than partial re-apply to make it more obvious - but probably not worth the effort here.

Sorry for delayed follow up - I think my current memory capacity is barely 2 days!

neilcsmith-net (approved):

PR #4586, opened 2022-09-06
Fixing error recovery for (current) record patterns.

jlahoda:
`tree.type` may be `null`, and `error2Object` will return `null` for `null` input, and the current code will reset the `sym.type` to `null` as well, which is obviously problematic/wrong.

---

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorker.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorker.java
@@ -582,7 +582,8 @@ final class VanillaCompileWorker extends CompileWorker {
                 } else {
                     scan(node.getInitializer(), null);
                 }
-                decl.sym.type = decl.type = error2Object(decl.type);
+                decl.type = error2Object(decl.type);
+                decl.sym.type = error2Object(decl.sym.type);
                 clearAnnotations(decl.sym.getMetadata());
                 return null;
             }
--- a/java/java.source.base/test/unit/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorkerTest.java
+++ b/java/java.source.base/test/unit/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorkerTest.java
@@ -36,6 +36,7 @@ import java.util.logging.LogRecord;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


ebarboni:
looks sane + testing will merge soon if nobody complains

mbien:
this test is currently failing on master with: 

```
junit.framework.AssertionFailedError: 
expected:
<[cache/s1/java/15/classes/test/Rect.sig, cache/s1/java/15/classes/test/ColoredPoint.sig, cache/s1/java/15/classes/test/Point.sig, cache/s1/java/15/classes/test/Test.sig]> 
but was:
<[cache/s1/java/15/classes/test/Rect.sig, cache/s1/java/15/classes/test/ColoredPoint.sig, cache/s1/java/15/classes/test/Point.sig]>
at org.netbeans.modules.java.source.indexing.VanillaCompileWorkerTest.testRecordPatterns(VanillaCompileWorkerTest.java:1951)
at org.netbeans.modules.java.source.indexing.VanillaCompileWorkerTest.runTest(VanillaCompileWorkerTest.java:1980)
```
I thought at first it was my fault with the CI pipeline migration (some flag wrong or similar), however it failed on travis before but nobody noticed, now its failing on gh actions.

ebarboni:
well I see it green, I merged it. Could it be possible to have some auto labelling that detect outdated PR ?

jlahoda:
I'll take a look in the evening. Sorry for trouble.

neilcsmith-net:
> honestly, this could have already happened with only a few hours in between the PRs, Its just bad luck tbh

I've noticed people in other ASF projects discussing use of merge queues recently. That might be something to consider here too in all the CI changes?

mbien:
another thing worth mentioning. the reason why we didn't notice this right away is because the test failed on travis, and nobody trusts travis to be reliable, so when its red on master not many bother to look since red travis is the default state - green is the exception. 

I only noticed it since we happen to move the job from travis to gh actions and it failed right away even showing the exception trace in the summary which made me happy that this worked :)

mbien:
I tested with nb-javac based on jdk 19+36 (local build) and the same test is also failing there.
Build 36 is the one which got promoted to JDK 19 GA.

@neilcsmith-net correct me if I am wrong but this is one of the releases when we have the option to upgrade nb-javac to a JDK `.1` update release during RC, right? I believe 19.0.1 should be out later this month and that would be also the timeframe when NB is starting its RC cycle.

I was wondering if we should update from 19+33 to 19+36 or wait for 19.0.1.

PR #4589, opened 2022-09-06
Multi-release jar support

matthiasblaesing:
This is an enhanced version of #3860, which implements support for multi-release jars for the JarClassLoader and the Equinox based OSGI integration.

Changes compared to the original PR:

- prevent unnecessary reads from original JAR (JAR content is cached and should be used at startup)
- use descending order for versions in `getVersions`, so that implementations for newer VMs are preferred and move runtime version limitation there
- add support for multi-release jars to NetBinox classloader by using the idea from `JarFile`, where the versioning aspect of class loading is handled below the classloader level and for NetBinox is implemented in the `JarBundleFile`
- mark the JSch module as multi-release. The JSch implementation for communication with the ssh agent can then use the JDK implementation of unix domain socket support and does not rely on junixsocket anymore.

### DIFF
--- a/platform/netbinox/src/org/netbeans/modules/netbinox/JarBundleFile.java
+++ b/platform/netbinox/src/org/netbeans/modules/netbinox/JarBundleFile.java
@@ -30,17 +30,21 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.jar.Attributes.Name;
+import java.util.jar.Manifest;
 import java.util.logging.Level;
 import org.eclipse.osgi.baseadaptor.BaseData;
 import org.eclipse.osgi.baseadaptor.bundlefile.BundleEntry;
 import org.eclipse.osgi.baseadaptor.bundlefile.BundleFile;
 import org.eclipse.osgi.baseadaptor.bundlefile.DirBundleFile;
-import org.eclipse.osgi.baseadaptor.bundlefile.DirZipBundleEntry;
 import org.eclipse.osgi.baseadaptor.bundlefile.MRUBundleFileList;
 import org.eclipse.osgi.baseadaptor.bundlefile.ZipBundleFile;
 import org.netbeans.core.netigso.spi.BundleContent;
 import org.netbeans.core.netigso.spi.NetigsoArchive;
 import org.openide.modules.ModuleInfo;
+import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
 
 /** This is fake bundle. It is created by the Netbinox infrastructure to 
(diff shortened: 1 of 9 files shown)


matthiasblaesing:
@dbalek @oyarzun @jlahoda @jtulach  could you please have a look at this? This takes #3860 (support for multi-release jars), fixes at least one unittest and extends the work to NetBinox.

I tested this manually with JSch, which uses a multi-release jar to use java native UnixDomainSockets on JDK 16+.

oyarzun (approved):
@matthiasblaesing I've also successfully tested it building a RCP app that uses a log4j multi-release jar. 

Like you said in https://github.com/apache/netbeans/pull/3860#issuecomment-1237423153 it is early in the NB16 release cycle, so we should probably merge this in sooner rather than later.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -574,13 +612,57 @@ protected URL doGetResource(String name) throws IOException  {
...
+                    if (data != null && data.length == 0) {
+                        vers.add(i);
+                    }
+                }
+                ret = new int[vers.size()];
+                int i = 0;
+                for (Integer ver : vers) {
+                    ret[i++] = ver;
+                }
+                versions = new SoftReference<>(ret);
```
mbien, line 655: management overheat of a SoftReference is probably greater than simply storing the small int array, which is either null (probably in most cases), or has 1-2 elements.

platform/netbinox/src/org/netbeans/modules/netbinox/JarBundleFile.java
```diff
@@ -49,14 +54,31 @@
  * @author Jaroslav Tulach <jtulach@netbeans.org>
  */
 final class JarBundleFile extends BundleFile implements BundleContent {
```
mbien, line 56: would be good to mention in a comment that this class shares code with `JarClassLoader` and vice versa, so that future readers know that they have to mirror changes.

platform/o.n.bootstrap/src/org/netbeans/JarClassLoader.java
```diff
@@ -77,6 +79,22 @@
...
+    private static final Name MULTI_RELEASE = new Name("Multi-Release");
+    private static final int BASE_VERSION = 8;
+    private static final int RUNTIME_VERSION;
+
+    static {
+        int version;
+        try {
+            Object runtimeVersion = Runtime.class.getMethod("version").invoke(null);
+            version = (int) runtimeVersion.getClass().getMethod("major").invoke(runtimeVersion);
+        } catch (Throwable ex) {
```
mbien, line 92: consider Exception or ReflectiveOperationException please. Catching Throwable without rethrow or handling it can cause all kinds of issues.

platform/netbinox/src/org/netbeans/modules/netbinox/JarBundleFile.java
```diff
@@ -49,14 +54,31 @@
...
+    private static final int BASE_VERSION = 8;
+    private static final int RUNTIME_VERSION;
     private static Map<Long,File> usedIds;
 
+    static {
+        int version;
+        try {
+            Object runtimeVersion = Runtime.class.getMethod("version").invoke(null);
+            version = (int) runtimeVersion.getClass().getMethod("major").invoke(runtimeVersion);
+        } catch (Throwable ex) {
```
mbien, line 68: same here

matthiasblaesing:
@mbien thank you for the check - you made valid points and I pushed an update with two new commits (to be squashed before merge)
@oyarzun thank you for the review.

mbien (approved):
good job

matthiasblaesing:
Rebased, travis and GHA are both happy, so lets get this in.

PR #4592, opened 2022-09-07
Provide hints for conversion to Virtual thread executor when thread pools are used

MegJayan:
Provide hints for converting thread pools(newFixedThreadPool, newCachedThreadPool) to virtual thread pool executor.
When ThreadFactory is passed as argument to , then only a hint suggestion for converting the thread pool to newThreadPerTaskExecutor is provided. Code is not directly rewritten in this case.


java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+    "ERR_ConvertToThreadPerTask=Convert to newThreadPerTask executor"
+})
+@Hint(displayName = "#DN_ConvertToVT", description = "#DESC_ConvertToVT", category = "rules15",
+        minSourceVersion = "19")
+public class ConvertToVT {
+
+    private static final int VT_PREVIEW_JDK_VERSION = 19;
+
+    @TriggerPatterns({
+        @TriggerPattern(value = "$var = $expr2.$meth1($var1$);"),
```
lahodaj, line 64: I'm afraid the hint is written in a way that is not ideal. Using `$expr2.$meth1`, and then checking inside the hint whether or not is error prone and relatively slow.

Doing something like:
```
@TriggerPattern(value="java.util.concurrent.Executors.newFixedThreadPool($size)" /*maybe a Constraint requiring that $size is an `int`*/)
@TriggerPattern(value="java.util.concurrent.Executors.newFixedThreadPool($size, $factory)" /*maybe a Constraint requiring that $size is an `int` and $factory is a ThreadFactory*/)
```

would avoid a lot of corner case states in the hint. (Patterns are meant to be cheap, so even some duplication should not be a problem, but if it is important to verify that the result is assigned into a variable, checking the enclosing AST node  in the hint might be more viable.) In this case, saying `java.util.concurrent.Executors.newFixedThreadPool` will tell the framework to not call the hint at all unless the `newFixedThreadPool` method on `java.util.concurrent.Executors` is called, which should be much faster (the framework should be able to eliminate non-matching patterns fairly quickly).

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+        @TriggerPattern(value = "$type $var = $expr2.$meth1($var1$);")
+    })
+    public static ErrorDescription compute(HintContext ctx) {
+        if (Utilities.isJDKVersionLower(VT_PREVIEW_JDK_VERSION) && !CompilerOptionsQuery.getOptions(ctx.getInfo().getFileObject()).getArguments().contains("--enable-preview")) {
+            return null;
+        }
+        List<String> methList = Arrays.asList("newFixedThreadPool", "newCachedThreadPool");
+        String expr = ctx.getInfo().getTrees().getElement(ctx.getVariables().get("$expr2")).getSimpleName().toString();
```
lahodaj, line 72: Please note that `Trees.getElement` may, in general, return `null`.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+        String method = ctx.getVariableNames().get("$meth1");
+        Collection<? extends TreePath> treePaths = ctx.getMultiVariables().get("$var1$");
+        boolean factory = false;
+        for (TreePath treePath : treePaths) {
+            if (treePath.getLeaf() instanceof IdentifierTree) {
```
lahodaj, line 77: Not sure why `IdentifierTree` is important here - I can imagine cases like:
`Executors.newFixedThreadPool(size)`, which will contain and identifier, and cases like:
`Executors.newFixedThreadPool(1, new ThreadFactory() {...})`, which will not.

I think that matching based on the specific method overload would be better.

Also, checking `Tree`s, `Element`s or `TypeMirror`s using `instanceof` is not completely reliable, I'm afraid.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+                factory = true;
+            }
+        }
+        if (expr.equals("Executors") && methList.contains(method)) {
```
lahodaj, line 81: Note that matching simple names is usually not very reliable. The user may have `Executors` class in their own package, and we probably should not be rewriting those.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+            Fix fix = new FixImpl(ctx.getInfo(), ctx.getPath(), factory).toEditorFix();
+            if (factory) {
+                return ErrorDescriptionFactory.forName(ctx, ctx.getPath(), Bundle.ERR_ConvertToThreadPerTask(), fix);
```
lahodaj, line 84: As far as I can tell, the fix is not doing anything for `factory==true`, so maybe it should not be used at all? It is OK to have a warning if we don't/can't have a fix, but showing a fix to the user that then doing nothing feels confusing.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/ConvertToVT.java
```diff
@@ -0,0 +1,132 @@
...
+                return Bundle.FIX_ConvertToVT();
+            } else {
+                return Bundle.ERR_ConvertToThreadPerTask();
+            }
+        }
+
+        @Override
+        protected void performRewrite(TransformationContext ctx) throws Exception {
+            if (!factory) {
+                WorkingCopy wc = ctx.getWorkingCopy();
```
lahodaj, line 113: FWIW, there is:
`org.netbeans.spi.java.hints.JavaFixUtilities.rewriteFix` which allows to specify the re-write in a textual form. E.g.:
`$var = $expr2.newVirtualThreadPerTaskExecutor($var1$);` (or, better: `java.util.concurrent.Executors.newVirtualThreadPerTaskExecutor()`, if the first suggestion is adopted.)

lahodaj (changes requested):
Added a few comments on where I think the hint could be improved. But, I wonder if "blindly" suggesting a conversion to `newVirtualThreadPerTaskExecutor` is desirable - doesn't `newVirtualThreadPerTaskExecutor` work best for tasks that wait a lot e.g. on I/O?

Also, there are probably patterns that should not be re-written, like `Executors.newFixedThreadPool(1)` - that is likely to mean run on background, but single-threaded. Changing this to multiple (virtual or non-virtual) threads may lead to problems.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/CanUseVT.java
```diff
@@ -0,0 +1,67 @@
...
+                = @ConstraintVariableType(variable = "$factory", type = "java.util.concurrent.ThreadFactory"))})
+    public static ErrorDescription compute(HintContext ctx) {
+        if (Utilities.isJDKVersionLower(VT_PREVIEW_JDK_VERSION) && !CompilerOptionsQuery.getOptions(ctx.getInfo().getFileObject()).getArguments().contains("--enable-preview")) {
+            return null;
+        }
+        if (ctx.getVariables().get("$factory") != null) {
+            return ErrorDescriptionFactory.forName(ctx, ctx.getPath(), Bundle.ERR_CanUseThreadPerTask(), new Fix[0]);
+        } else {
+            return ErrorDescriptionFactory.forName(ctx, ctx.getPath(), Bundle.ERR_CanUseVT(), new Fix[0]);
+        }
```
mbien, line 65: this could return a rewrite fix, otherwise it will just mark the line and do nothing.
something like:

```java
Fix fix = JavaFixUtilities.rewriteFix(ctx, "converts to VT msg", ctx.getPath(), "java.util.concurrent.Executors.newThreadPerTaskExecutor($factory)");
```
and...
```java
Fix fix = JavaFixUtilities.rewriteFix(ctx, "converts to VT msg", ctx.getPath(), "java.util.concurrent.Executors.newVirtualThreadPerTaskExecutor()");
```

mbien (commented):
we should discuss if this is something the IDE should recommend by default. Because switching to VTs is not something which should be done blindly.

Even though the API is the same, the concepts are not. E.g there usually is no reason to migrate purely CPU bound tasks (as correctly mentioned in the hint description).

But more importantly: fixed thread pools are not the same as a `newVirtualThreadPerTaskExecutor`, you going to need a semaphore to achieve similar behavior otherwise you just added a DDOS vuln. There are also other things to pay attention to (in both your code and in libraries), e.g usage of `ThreadLocals` which used to be an optimization, but are now an anti-pattern which doesn't scale anymore. Or synchronized sections which can block or native stack frames which could lock carrier threads etc.

VTs are great and I am all for using new features and having code transformations which help with migration but maybe not enabled by default :)

options to consider: disable the hint, or make it current-line only

@neilcsmith-net @matthiasblaesing whats your opinions on this?

neilcsmith-net:
> VTs are great and I am all for using new features and having code transformations which help with migration but maybe _not enabled by default_

Emphasis mine, but yes, agree, IMO should not be enabled by default, for now at least.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/CanUseVT.java
MegJayan: I understand your point. One of the reasons why we decided to give the hint was so that user is notified of  the support of virtual threads and he/she can take a call on whether to modify the current code to use virtual threads depending on the use case. This is why we are providing just the suggestion without rewriting the code.

In the initial commit there was rewrite logic as well, but after the first round of discussion, it was decided just to go with hints and not rewrite anything as there could be multiple corner cases like Executors.newFixedThreadPool(1) [as mentioned by @lahodaj  [here](https://github.com/apache/netbeans/pull/4592#pullrequestreview-1112236691)] which probably should not be rewritten.
This is the background. Please let me know your thoughts. If you think its best not to provide hints, I can update accordingly.

arvindaprameya:
Based on the comments we have now changed to give hints for the new feature of VT thats available. 
Instead of converting the code. So no more automatic conversion.Hope this clarifies.

java/java.hints/src/org/netbeans/modules/java/hints/jdk/CanUseVT.java
mbien: sorry for the delay.

I believe this could be definitively useful as refactor-and-transform inspection. A dev can inspect a project and check if there are pools which can be moved to VTs on a case by case basis. If the code rule can convert the easy cases to the VT executor - why not (the difficult cases would be only marked without a transformation).

I just don't think it should be enabled by default, because light bulbs can be perceived as something you should fix. But if a dev enables the hint manually, the situation is different since the choice has been already made.

Those are just my thoughts, I don't have a strong opinion on it.

MegJayan: @mbien : I have updated the code after discussion with @lahodaj. Modified the category to "suggestions" and  hintKind to ACTION in the latest code so that light bulbs are not shown by default whenever the thread pools are used. Havent implemented the conversion for easy cases as we decided against it due to the corner case [there are probably patterns that should not be re-written, like Executors.newFixedThreadPool(1)] Please check and let me know if this is ok.

mbien: what do you think about `hintKind = Hint.Kind.INSPECTION, severity = Severity.HINT`, would this be also ok?

advantage:
 - it can be used as inspection too
 - it remains current-line-only just like with `hintKind = ACTION`

actions won't be listed in the source->inspect and refactor->inspect and transform dialogs and the user can't change the setting from current-line to warning since the combo box is disabled.

MegJayan: @mbien : I think it should be ok to modify the hintKind to INSPECTION. It will be listed in source-> inspect. But as per the previous conclusion we dont provide any fix for conversion to virtual threads. So refactor-> inspect and transform will not work for transformation.

mbien: @MegJayan right, there would be no transformation offered if its used via inspect and transform. But it would be current-line-only just like an action (but also configurable by the user) with the added benefit of it being usable as an inspection too (which I imagine could be super useful in some situations).

I don't know how to hide hints from the transformation dialog if they don't offer fixes. Maybe a third kind `{ACTION, INSPECTION, TRANSFORMATION}` would be required for that?

MegJayan: I have updated the PR changing hintKind to INSPECTION. Since transformation is not available, it shows the below msg on clicking Inspect and Transform:
No fix for: Can Use Virtual Thread Executor at JavaApplication2.java:21
I think this conveys to the user that transformation isnt available and this should be ok.

mbien: thank you @MegJayan for your patience. I am looking forward to use the new inspection/hint to find candidates for VTs.

mbien (approved):
great addition - helps to find VT candidates

neilcsmith-net:
@singh-akhilesh please make sure there's a milestone set when merging, thanks!

PR #4594, opened 2022-09-07
Support projects without build.gradle

sdedic:
A multiproject build looks, [according to the documentation](https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:creating_multi_project_builds) like
```
├── app
│   ...
│   └── build.gradle
└── settings.gradle
```
there's no `build.gradle` in the root. While `GradleFiles.isSubProject` already works OK with build-less projects, the `isRootProject` does not. This PR fixes that. Simple test added.

During other work, I've realized that the `ProjectFactory2` implementation for both Gradle and Maven do not return project type -- although they could. The display name is not that easy to evaluate, but both project types use a fixed ID, so it's simple to return.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectFactory.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/NbGradleProjectFactory.java
@@ -43,7 +43,12 @@ public final class NbGradleProjectFactory implements ProjectFactory2 {
 
     @Override
     public ProjectManager.Result isProject2(FileObject dir) {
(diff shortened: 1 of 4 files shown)


lkishalmi (approved):
Sounds fine!

sdedic:
Yes; but for certain purposes it may be important to differentiate the build system, i.e. for some abstractions that map to different actions/goals/properties/whatever in each build system. It was already in the APIs, so I've just filled the gap.
It's a pity that `displayName` remains unfilled - the `isProject2` could be a good query for e.g. file selector.

PR #4595, opened 2022-09-07
Project info exported through LSP

sdedic:
LSP client may need some project structure information. This PR exposes basic info on a project on the LSP protocol.
```
nbls.project.info(locations : String | String[], options? : { projectStructure? : boolean; actions? : boolean; recursive? : boolean }) : LspProjectInfo[]
```
The command accepts one or more locations as URIs (vscode.Uri.toString() for example); it will return, for each input parameter, a structure that describes the project and its direct children (if requested). If `recursive` is true, additional entries will be provided for each identified child project, recursively.


sdedic:
Note the `LspProjectInfo.projectType` will not be filled until #4594 is merged. 

For Gradle projects, Micronaut GCN wizard generates multi-project that is not recognized by current Gradle module - #4594 should fix that as well. In the meantime, use empty `build.gradle` to fool the project factory.

dbalek (approved):
Thanks.

PR #4597, opened 2022-09-07
Set "--add-modules ALL-MODULE-PATH" when compiling MX project.

dbalek:
"--add-modules ALL-MODULE-PATH" set when compiling MX projects.


lkishalmi (approved):
Looks good to me.

jlahoda (approved):
Looks good to me.

Issue #4437, opened 2022-07-25
Editor with Keyboard focus is not active TopComponent (undo/redo don't work while editing)


errael:
### Apache NetBeans version

Apache NetBeans 14

### What happened

I encountered this problem with the output window detached from the `main NetBeans Window` and **focus-follows-mouse** is set in the linux the window manager. While editing, I moved the mouse from editor to detached window, when the detached window gets focus the _undo/redo buttons go inactive_, move the mouse back to the editor and the _buttons remain inactive_ although I can edit/save as usual.

### How to reproduce

As described in **What happened**

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Linux harmony 5.18.10-76051810-generic #202207071639~1657252310~22.04~7d5e891 SMP PREEMPT_DYNAMIC Fri J x86_64 x86_64 x86_64 GNU/Linux

### JDK

8,11,17

### Apache NetBeans packaging

Own source build

### Anything else

As a workaround, in a `JEditorPane`'s `focusListener.focusGained`, I've essentially got:
```
    TopComponent tc = NbEditorUtilities.getOuterTopComponent(getEditor());
    if(tc != TopComponent.getRegistry().getActivated())
        tc.requestActive();
```
 where getEditor() is like `FocusEvent.getComponent()`.

For NetBeans, perhaps something like this
```
        Mode m = WindowManager.getDefault().findMode(tc);
        if(m != null)
            central.setActiveMode(m);
```
in MainWindow's windowListener (but without proper exerience in core.windows, I'd be hesitant to make changes in this area) or maybe a focus listener with something like the workaround I'm using above in an editor's `TopComponent`.

Relevant stack when moving mouse to detached window
```
	at org.netbeans.core.windows.WindowManagerImpl.notifyRegistryTopComponentActivated(WindowManagerImpl.java:1199)
	at org.netbeans.core.windows.Central.setActiveMode(Central.java:233)
	at org.netbeans.core.windows.Central.userActivatedModeWindow(Central.java:1883)
	at org.netbeans.core.windows.view.DefaultView.userActivatedModeWindow(DefaultView.java:654)
	at org.netbeans.core.windows.view.ui.DefaultSeparateContainer$SharedModeUIBaseImpl$1.windowActivated(DefaultSeparateContainer.java:455)
	at java.awt.AWTEventMulticaster.windowActivated(AWTEventMulticaster.java:398)
```

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

[...]

PR #4603, opened 2022-09-09
Keyboard focus can activate top component

errael:
Fix for #4437. Discussion in that issue is useful in understanding this patch.

`platform/core.windows` is  a mystery to me; this patch is almost identical to getting a mouse click, except it is triggered by keyboard focus change to handle other situations where the MainWindow is activated. There may be better solutions, but I believe this is safe.

After getting a focus event, the activate code is only invoked when it is strictly needed. This is an optimization, invoking the activation code multiple times shouldn't matter; it would be like clicking on an already active window.

@jtulach, any idea for a good candidate to review this? This area was written by _Peter Zavadsky_.

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/view/ui/TabbedHandler.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/view/ui/TabbedHandler.java
@@ -36,13 +36,17 @@ import org.openide.util.Utilities;
 import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
+
 import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.MouseEvent;
 import java.awt.event.AWTEventListener;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import org.netbeans.core.windows.ModeImpl;
 import org.netbeans.core.windows.Switches;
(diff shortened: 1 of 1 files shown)


eirikbakke:
For others passing through, here is @neilcsmith-net's reproducibility comment from the other thread:

> Try detaching a source editor tab into a separate window and Alt-TABing between. The cursor moves from one editor to the other, but the window system active tab doesn't change - undo/redo is wrong, navigator is wrong, and wrong tab is coloured as focused. Typing goes in the right place, though.

I can confirm this on my own Windows 11 machine; this is obviously incorrect behavior. If the patch fixes this problem, then great; thank you!

Looking at the patch, it does carefully apply the modified behavior only in some specific cases, which should make it lower-risk. Looks good to me, but maybe good to have one more reviewer look at it.

The next time I make a private NetBeans build I'll include this patch and test it...

errael:
> The next time I make a private NetBeans build I'll include this patch and test it...

That would be great, I moved to linux last year, not tested on windows.

BTW, note that mouse click processing happens before keyboard focus change processing so this doesn't kick in when focus change due to mouse click.

eirikbakke:
I just tested this in a custom NetBeans build; the fix seems to work! Tested for correct active tab indication and Undo/Redo effect on Windows 11 with focus changed with Alt+Tab, Ctrl+Tab, clicking in Windows taskbar, clicking on tab within NetBeans, and clicking on window title bar.

I'll now leave this patch in my working IDE for a while to see that there are no problems.

Thanks for this!

eirikbakke (approved):
I've been running my IDE with this patch active for the past few days with no problems, and it fixes the problem. I also reviewed the code previously.

I think this PR should be safe to merge.

Issue #3732, opened 2022-03-08
Netbeans freezes while editing a file and resizing its components 


wumpz:
### Apache NetBeans version

Apache NetBeans 13

### What happened

Complete freeze, Memory got to max (I think). This exception was multiple times at the end of my log:

```
INFO [org.netbeans.ui.indexing]: Indexing started, time from last indexing 84.816 ms.
INFO [org.netbeans.ui.indexing]: Indexing finished, indexing took 15 ms.
WARNING [org.netbeans.TopSecurityManager]: use of system property netbeans.user has been obsoleted in favor of InstalledFileLocator/Places at org.netbeans.modules.websvc.saas.model.SaasServicesModel.<clinit>(SaasServicesModel.java:46)
INFO [org.netbeans.modules.gradle.loaders.GradleProjectLoaderImpl]: Load aiming EVALUATED for Unloaded Gradle Project: GradleFiles[projectDir=D:\work\NetBeansProjects\gradleproject1, rootDir=D:\work\NetBeansProjects\gradleproject1]
INFO [org.netbeans.core.netigso.Netigso]: bundle org.eclipse.osgi@3.9.1.v20140110-1610 256
INFO [org.netbeans.core.netigso.Netigso]: bundle org.eclipse.osgi@3.9.1.v20140110-1610 stopped
INFO [null]: Last record repeated again.
SEVERE [global]
java.lang.IllegalArgumentException: Zero length string passed to TextLayout constructor.
	at java.desktop/java.awt.font.TextLayout.<init>(TextLayout.java:382)
	at org.netbeans.modules.editor.lib2.view.DocumentViewOp.createTextLayout(DocumentViewOp.java:1397)
	at org.netbeans.modules.editor.lib2.view.HighlightsView.createPartTextLayout(HighlightsView.java:173)
	at org.netbeans.modules.editor.lib2.view.HighlightsViewPart.getTextLayout(HighlightsViewPart.java:135)
	at org.netbeans.modules.editor.lib2.view.HighlightsViewPart.getPreferredSpan(HighlightsViewPart.java:79)
	at org.netbeans.modules.editor.lib2.view.WrapInfoUpdater.createFragment(WrapInfoUpdater.java:548)
	at org.netbeans.modules.editor.lib2.view.WrapInfoUpdater.initWrapInfo(WrapInfoUpdater.java:152)
[...]

PR #4623, opened 2022-09-11
FIX-3732: Prevent possible IAE in DocumentViewOp.

lkishalmi:
Kind of trivial fix for a rare ```IllegalArgumentException```. There is a small chance that it would result an ```NPE``` later in the code, though as far as I've checked the call stack in #3732 the return value of ```createTextLayout(String text, Font font)``` stored, and when later retrieved it is checked for null.


mbien (approved):
awesome. good fix. 

I just learned that you have to remove the triage label from the issue, otherwise you can't link it to PRs :) (already linked it)

PR #4625, opened 2022-09-12
Minor cleanup in YAML Language Support

lkishalmi:
There are two things to see here:
1. Move the language definition from ```YamlTokenId``` to ```YamlLanguage``` along with the ```MIME_TYPE```
2. Removed the unnecessary mime-registration of the language as the CSL support takes care of that.

This is just applying the knowledge gathered during ANTLR Support implementation.


ide/languages.yaml/src/org/netbeans/modules/languages/yaml/YamlLanguage.java
```diff
@@ -155,12 +167,72 @@ public InstantRenamer getInstantRenamer() {
         return null;
     }
 
+    public static final Language<YamlTokenId> language =
```
matthiasblaesing, line 170: I saw this now multiple time - I think it is an error to expose members of a class to the outside without a safety net. I acknowledge, that ` records` dropped the `get` prefix (still think this is an error), but they still use functions. Using a function allows you to modify the getter to log usages or provide an adapter to a new implementation, that might not be available when the class is created or is costly to create. At the same time hotspot should (I did not measure) be good enough in inlining getters. For an API this would be a no-go for me, for internal module use it still feels wrong to me.

lkishalmi: Well, despite it is written all lowercase, the `language` is a constant. I see nothing wrong using public class constants without a wrapping static method. (I'm with the remove the bloat folks) That static method was required for the `MimeRegistration`, which after all is not required. 

Ideally it could be a private final field, though it is being used one time in the Parser. and about 10 times in the tests.

matthiasblaesing: As said - for an API using a constant removes the possibility to create Adapter on demand as initialization will happen at class load time.

As said from my POV this is a code smell as it exposes internals while gaining little.

For the current change it is ok, as it is hidden in the module and thus free to change at any time, but should this ever become API this would have to go.

So ignore this now, I made my point.

matthiasblaesing (approved):
Thanks ;-)

PR #4630, opened 2022-09-13
Allow to disable warning for specific lifecycle participants.

sdedic:
During implementation I have found that our Maven support warns if certain plugins are included in the project - since their code should be called after project load, or maven session open, that can have unexpected effects.

This PR adds a way how one can disable such a project warning - and fixes Micronaut support to disable the warning for harmless extensions (that do nothing important in their `afterProjectsRead` etc handlers).


dbalek (approved):
Looks reasonable.

PR #4631, opened 2022-09-13
Artifact tags, support for shaded output and source attachments.

sdedic:
This PR extends the **still private API** for project artifacts with 'tags', which are implementation- or technology-dependent labels. Currently tags are used to mark artifact that are 'shaded' (with dependencies included) or 'base' so the client can select processed or original artifacts.

Maven, for example, can produce multiple shaded artifacts with different classifiers or can **replace** the main project's artifact with the shaded one. Now the client is able to find or at least enumerate the originals, the shaded artifact and since they are tagged, the client can pick the files it needs.

Several abstract classifiers are defined (sources, test-sources, tests), they are mapped to build system's own classifiers by the implementation.


dbalek (approved):
Looks reasonable.

Issue #4646, opened 2022-09-17
JavadocImports throws exception due to incorrect string trim


naotsugu:
### Apache NetBeans version

Apache NetBeans 15

### What happened

An exception is raised if an attribute is specified in the JavaDoc to be trimmed.

```
java.lang.StringIndexOutOfBoundsException: begin 0, end 7, length 6
at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4604)
at java.base/java.lang.String.substring(String.java:2707)
at org.netbeans.modules.java.editor.base.javadoc.JavadocImports.processDocTreeNode(JavadocImports.java:824)
at org.netbeans.modules.java.editor.base.javadoc.JavadocImports.access$100(JavadocImports.java:81)
at org.netbeans.modules.java.editor.base.javadoc.JavadocImports$1.scan(JavadocImports.java:135)
at org.netbeans.modules.java.editor.base.javadoc.JavadocImports$1.scan(JavadocImports.java:117)
at com.sun.source.util.DocTreeScanner.scanAndReduce(DocTreeScanner.java:88)
at com.sun.source.util.DocTreeScanner.scan(DocTreeScanner.java:103)
at com.sun.source.util.DocTreeScanner.visitReturn(DocTreeScanner.java:419)
```

### How to reproduce

An exception will be thrown if a character containing spaces such as `{@ ` is present in the JavaDoc.

```
/**
 * {@ xxx}
 */
```
Note that this character contains spaces.

The following incorrect trim process is the cause.

org.netbeans.modules.java.editor.base.javadoc.JavadocImports.java

```java
int errorBodyLength = body.trim().substring(0, body.length()).trim().length();
```

I do not understand the intent of the above, but I speculate that the following is appropriate.

```java
int errorBodyLength = body.trim().length();
```

Added in this pull. [https://github.com/apache/netbeans/pull/2954](https://github.com/apache/netbeans/pull/2954)

### Did this work correctly in an earlier version?

Apache NetBeans 12.4

### Operating System

MacOS 12.5.1

### JDK

OpenJDK 64-Bit Server VM Corretto-18.0.0.37.1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
thanks for looking into this. Just as note: no need to create an issue if you are working on a PR anyway. PRs are sufficient :)

PR #4647, opened 2022-09-17
JavadocImports fixed for illegal string trim.

naotsugu:
JavadocImports fixed for illegal string trim raises a StringIndexOutOfBoundsException.(#4646)


matthiasblaesing (approved):
Fix looks good to me. The code you removed looks like an initial implementation that was not cleared up after experimenting. I could reproduce the problem and verified, that the fix fixes it.

I checked the commit, the author information looks good.

@naotsugu I assume you read the license header and understand, that you donate the change to the ASF. If you don't agree, please indicate so, the changeset will be backed out then.

Once the final test job comes back green I'll merge.

Issue #3720, opened 2022-03-07
try with resources formatting adding extra space


dstutz:
### Apache NetBeans version

Apache NetBeans 13

### What happened

try with resources formatting adds an extra space after the opening parens.
```
 try ( InputStream is = new FileInputStream("foo")) {
      ^   
```

Unfortunately, we use checkstyle and this causes a rule violation *and* we use format on save.

### How to reproduce

Type the following and format
`try (InputStream is = new FileInputStream("foo")) { }`

### Did this work correctly in an earlier version?

Apache NetBeans 12.6

### Operating System

Windows 10

### JDK

Eclipse temurin jdk-17.0.0.35-hotspot

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

This has been plaguing people on and off for several old versions.  It used to be related to the nb-javac plugin but I have tried with/without/activated/deactivated and it doesn't seem to make a difference, it keeps adding the extra space.

Some previous issues:
https://issues.apache.org/jira/browse/NETBEANS-236
https://issues.apache.org/jira/browse/NETBEANS-4641
https://issues.apache.org/jira/browse/NETBEANS-4676
https://issues.apache.org/jira/browse/NETBEANS-6129
https://issues.apache.org/jira/browse/NETBEANS-6400

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

PavelCibulka:
Same here on Netbeans 13.

It was symptom of much bigger problems related to nb-javac in previous versions. I usually fixed it by installing nb-javac plugin for second time (it uninstalled somehow).

I hope that extra space is only problem.

ilgrosso:
Same here with NB 14, Zulu 17 on Ubuntu 20.04.
Hope it gets fixed soon.

AndrewG10i:
Issue still present on NB 15.

adamoutler:
This issue can be clearly seen in the Project Properties->Formatting->Category: Alignment. 
`try ( Resource r1 = null;`

The space between `(` and `Resource` cannot be removed, but oddly if one visits the Project Properties->Formatting->Category: Spaces, then selects "try", the space doubles so the result now contains two spaces between `(` and `Resource`.
`try (  Resource r1 = null;`

I'm not sure how to fix this problem, but it certainly is annoying when running tools such as CheckStyle which declare errors in your code style unless you disable the rule and/or turn off all auto-formatting.  

Has anyone found a workaround?
[...]

PR #4648, opened 2022-09-17
Fix extra space added in try with resources

neilcsmith-net:
Fix for #3720 

The space seems to occur due to the start and end positions reported for the implicit hidden final modifier, which seem to be the same as the type start and end. The issue can't be seen if `final` is added explicitly.  

This adds an extra check for whether the modifiers and type start at the same position.  I haven't checked if the reported positions of `mods` have changed from earlier versions. The might explain why the issue started appearing, and it's possible the whole enclosing `if` wasn't hit in that case. I've kept localized for now.

The issue with new lines being removed may also be related to this, but haven't managed to track down exactly what's happening - it's possibly related to the token locations. Leaving that for another fix.

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
@@ -1151,7 +1151,7 @@ public class Reformatter implements ReformatTask {
                                 newline();
                             else
                                 space();
-                        } else {
+                        } else if (sp.getStartPosition(root, mods) != sp.getStartPosition(root, node.getType())) {
                             space();
                         }
                     } else if (afterAnnotation) {
--- a/java/java.source.base/test/unit/src/org/netbeans/modules/java/source/save/FormatingTest.java
+++ b/java/java.source.base/test/unit/src/org/netbeans/modules/java/source/save/FormatingTest.java
@@ -6099,6 +6099,61 @@ public class FormatingTest extends NbTestCase {
         reformat(doc, content, golden);
     }
 
+    public void testTryWithResources() throws Exception {
+        testFile = new File(getWorkDir(), "Test.java");
+        TestUtilities.copyStringToFile(testFile, "");
+        FileObject testSourceFO = FileUtil.toFileObject(testFile);
+        DataObject testSourceDO = DataObject.find(testSourceFO);
+        EditorCookie ec = (EditorCookie) testSourceDO.getCookie(EditorCookie.class);
+        final Document doc = ec.openDocument();
+        doc.putProperty(Language.class, JavaTokenId.language());
+
+        String content
+                = "package hierbas.del.litoral;\n\n"
+                + "public class Test {\n\n"
+                + "    public static void main(String[] args) {\n"
+                + "        try (PrintStream out = System.out) {\n"
+                + "            System.out.println(\"TEST\");\n"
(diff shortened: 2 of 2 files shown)


java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java
```diff
@@ -1151,7 +1151,7 @@ public Boolean visitVariable(VariableTree node, Void p) {
                                 newline();
                             else
                                 space();
-                        } else {
+                        } else if (sp.getStartPosition(root, mods) != sp.getStartPosition(root, node.getType())) {
```
mbien, line 1154: the implicit `final` would have the same start position as the type (which is the next item in the tree I guess) because its hidden?

I am wondering if this is just an impl detail or actually stable behavior, because there would be also the NOPOS constant
https://github.com/openjdk/jdk/blob/05d38604a2c620dcaf8682f02dae2fddab8e0c0b/src/jdk.compiler/share/classes/com/sun/source/util/SourcePositions.java#L41-L43
which could be potentially returned in future?

if implicit items have the same start and end pos we could extract this into:
```java
private static boolean isHidden(CompilationUnitTree cut, Tree tree) {
    long start = sp.getStartPosition(cut, tree);
    long end = sp.getEndPosition(cut, tree);
    return start == end && start != javax.tools.Diagnostic.NOPOS;
}
```
Maybe that would be less likely to break on a javac update (assuming it works)?

neilcsmith-net: > if implicit items have the same start and end pos

Thanks @mbien but this hits on the crux of the issue. The implicit `final` does not have the same start and end position (zero length), or report `NOPOS`. It reports the same start _and end_ positions as the following type element.  It has the same length and overlaps.

Were the element reported empty or at no position, it shouldn't get past the enclosing check at https://github.com/apache/netbeans/blob/master/java/java.source.base/src/org/netbeans/modules/java/source/save/Reformatter.java#L1146  I assume that the reported end position changed which started causing this issue, but haven't checked old versions to be sure.

neilcsmith-net: Bizarrely I also think that #4667 might be caused by this same issue.

neilcsmith-net: Ironically @jlahoda might have fixed this today? https://bugs.openjdk.org/browse/JDK-8293897

The hotfix shouldn't break if this is fixed. And it's covered by tests now I think. Could add a comment to revert when upstream fixed?

mbien: >The hotfix shouldn't break if this is fixed. And it's covered by tests now I think. Could add a comment to revert when upstream fixed?

Yeah I think so - lets get this in. The tests are indeed important so that we notice regressions quickly. Good that you found that bug report, it clears things up.

If the upstream fix makes it into javac 19.0.1 too, we could adjust/partially revert it later while upgrading nb-javac.

mbien (approved):
good hotfix, this will make many happy - can be potentially removed again once fixed upstream in the JDK as discussed in the review comments.

neilcsmith-net:
Thanks @mbien - merging.

PR #4649, opened 2022-09-17
Fix for not calling auto completion after end tag was already added

Chris2011:
right after opening tag, while copy and paste or closing tag token was added.

The check was not proper enough for calling auto completion after end tag. It just checked the HtmlToken.TAG_CLOSE_SYMBOL which is also the one at the end of an end tag </div">".
So each time I copy and paste "\<div>\</div>" it calles the auto completion which seemed wrong to me. See my screencapture:

end-tag-code-completion

I changed that behaviour and now it always comes up at the end of an open tag. Not on an end tag or on a self closing tag like an image, which was already working before: 

no-end-tag-code-completion

I will test this a couple of days to check whether there are other misbehavior or not and if there are no complaints I will merge this later by myself.

### DIFF
--- a/ide/html.editor/src/org/netbeans/modules/html/editor/completion/HtmlCompletionProvider.java
+++ b/ide/html.editor/src/org/netbeans/modules/html/editor/completion/HtmlCompletionProvider.java
@@ -386,16 +386,10 @@ public class HtmlCompletionProvider implements CompletionProvider {
                             } else {
                                 ts.move(dotPos - 1);
                                 if (ts.moveNext() || ts.movePrevious()) {
-                                    if (!CharSequenceUtilities.equals("/>", ts.token().text()) && null != LexerUtils.followsToken(ts, HTMLTokenId.TAG_OPEN, true, false,
-                                            HTMLTokenId.ARGUMENT,
-                                            HTMLTokenId.VALUE,
-                                            HTMLTokenId.VALUE_CSS,
-                                            HTMLTokenId.VALUE_JAVASCRIPT,
-                                            HTMLTokenId.OPERATOR,
+                                    if (!CharSequenceUtilities.equals("/>", ts.token().text()) &&
+                                        null == LexerUtils.followsToken(ts, HTMLTokenId.TAG_CLOSE, true, false,
                                             HTMLTokenId.WS,
-                                            HTMLTokenId.EL_CLOSE_DELIMITER,
-                                            HTMLTokenId.EL_CONTENT,
-                                            HTMLTokenId.EL_OPEN_DELIMITER)) {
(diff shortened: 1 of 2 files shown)


ide/html.editor/src/org/netbeans/modules/html/editor/completion/HtmlCompletionProvider.java
```diff
@@ -386,7 +386,16 @@ public void run() {
...
+                                    if (!CharSequenceUtilities.equals("/>", ts.token().text()) && null != LexerUtils.followsToken(ts, HTMLTokenId.TAG_OPEN, true, false,
+                                            HTMLTokenId.ARGUMENT,
+                                            HTMLTokenId.VALUE,
+                                            HTMLTokenId.VALUE_CSS,
+                                            HTMLTokenId.VALUE_JAVASCRIPT,
+                                            HTMLTokenId.OPERATOR,
+                                            HTMLTokenId.WS,
+                                            HTMLTokenId.EL_CLOSE_DELIMITER,
+                                            HTMLTokenId.EL_CONTENT,
+                                            HTMLTokenId.EL_OPEN_DELIMITER)) {
```
matthiasblaesing, line 398: In general I like this - the current behavior annoyed the hell out of me. The second check looks to complex - If I understand correctly, you want to check if you are behind an open tag. I would invert this and check if it is not a closing tag. 

```suggestion
                                    if (!CharSequenceUtilities.equals("/>", ts.token().text()) && null == LexerUtils.followsToken(ts, HTMLTokenId.TAG_CLOSE, true, false,
                                            HTMLTokenId.WS,
                                            HTMLTokenId.TAG_CLOSE_SYMBOL)) {
```

The idea here is, that the definition for a closing tag is pretty simple: https://www.w3.org/html/wg/spec/syntax.html#end-tags

Chris2011: Shouldn't it be null !=... I thought it gives you the closing tag back, if the method finds it? But in general you are right, less checks etc. seems more performant and cleaner.

matthiasblaesing: The suggested check returns true if we are _not_ directly behind a closing tag. `null !=` would yield true after a closing tag, which is the opposite of what we want.

Chris2011: True, got it. Thx for the hint. I fixed and tested it. Worked as before.

matthiasblaesing (approved):
Looks sane to me. Thank you.

PR #4654, opened 2022-09-19
Made FileUtil to copy POSIX permissions

lkishalmi:
Well as a DevOps engineer, I copy scripts and other executable inside the IDE. They lose the executable flag. That bothers me for years. Here is how I try to fix that.

Also added `FileObject` `Path` conversion methods, to make the life easier.

### DIFF
--- a/platform/openide.filesystems/apichanges.xml
+++ b/platform/openide.filesystems/apichanges.xml
@@ -25,6 +25,30 @@
         <apidef name="filesystems">Filesystems API</apidef>
     </apidefs>
     <changes>
+        <change id="fileutil.copyposixperms">
+            <api name="filesystems"/>
+            <summary>FileObject copy preserves source posix permissions.</summary>
+            <version major="9" minor="32"/>
+            <date day="12" month="1" year="2023"/>
+            <author login="lkishalmi"/>
+            <compatibility addition="yes" semantic="compatible"/>
+            <description>
+                <a href="@TOP@/org/openide/filesystems/FileUtil.html#copyFile-org.openide.filesystems.FileObject-org.openide.filesystems.FileObject-java.lang.String-java.lang.String-">FileUtil.copyFile(...)</a> now preserve ATTRIBUTES and POSIX permissions.
+            </description>
+            <class name="FileUtil" package="org.openide.filesystems"/>
+        </change>
+        <change id="fileutil.niofilepath">
+            <api name="filesystems"/>
+            <summary>FileUtil can convert FileObject to/from java.nio.file.Path.</summary>
+            <version major="9" minor="32"/>
+            <date day="12" month="1" year="2023"/>
+            <author login="lkishalmi"/>
+            <compatibility addition="yes" semantic="compatible"/>
+            <description>
+                <a href="@TOP@/org/openide/filesystems/FileUtil.html#">FileUtil</a> has a <code>toPath(FileObject fo)</code> and <code>toFileObject(Path path)</code> utility methods.
+            </description>
+            <class name="FileUtil" package="org.openide.filesystems"/>
+        </change>
         <change id="fileutil.copyattr.trasnform">
             <api name="filesystems"/>
             <summary>Allow to filter or transform attribute values during copying.</summary>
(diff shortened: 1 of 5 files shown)


sdedic:
Two questions:
- given that FileUtil.copy did not copy the attributes or is not specified as 'copied by stream content' ... isn't that an incompatible API change ?
- what about the `AclFileAttributesView` - this is useful on Windows. Except for file owner, ACLs may be (usually) set by a regular user.

lkishalmi:
Well that API is ancient, from those times, when the Executable permissions could not be set by Java (pre-Java 6) without some magic.

As a DevOps Engineer, I'm working with a lot of scripts recently, so I thought I fix my issue when copy them in the IDE. That's all. I do not know how important Windows ACL for other people, well if that's really missed, contributions are welcome!

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -567,6 +569,7 @@ static FileObject copyFileImpl(FileObject source, FileObject destFolder, String
             }
 
             copy(bufIn, bufOut);
+            copyPosixPerms(source, dest);
             copyAttributes(source, dest);
```
mbien, line 573: just a thought:
Files.copy(...) would copy posix permissions too (there is also a COPY_ATTRIBUTES option). Alternative impl would check if its an actual file first, if it is, it would use Files.copy() with paths, otherwise use streams.

Maybe its worth thinking about refreshing this code if it isn't too much work? Otherwise we end up reimplementing everything what Files.copy already does.

mbien (approved):
looks good to me - left a comment but feel free to ignore it.

platform/openide.filesystems/apichanges.xml
```diff
@@ -25,6 +25,30 @@
         <apidef name="filesystems">Filesystems API</apidef>
     </apidefs>
     <changes>
+        <change id="fileutil.copyposixperms">
+            <api name="filesystems"/>
+            <summary>FileObject copy preserves source posix permissions.</summary>
+            <version major="9" minor="30"/>
```
sdedic, line 31: Bump version in `manifest.mf` too

platform/openide.filesystems/apichanges.xml
```diff
@@ -25,6 +25,30 @@
...
+            <date day="18" month="9" year="2022"/>
+            <author login="lkishalmi"/>
+            <compatibility addition="yes" semantic="compatible"/>
+            <description>
+                <a href="@TOP@/org/openide/filesystems/FileUtil.html#">FileUtil.copy</a> now preserves POSIX permissions if possible during file copy.
```
sdedic, line 36: missing anchor to the FileUtil.copy chapter ?

jtulach (commented):
I always dreamed about 1:1 mapping between Path and FileObject. They could delegate to each other. Always. E.g. the new methods shall never return null.

neilcsmith-net:
@lkishalmi still a few things to address here.  I also agree with @mbien and @jtulach about 1:1 mapping behaviour with Path and Files::copy COPY_ATTRIBUTES.  Not sure whether at least minimal things can be addressed and merged for NB16 or we should push all changes to NB17?

lkishalmi:
I hope I can make those minimal changes needed. The day is just started for me. If I can't make that happen, you are right, I'll move this to NB17.

neilcsmith-net:
@lkishalmi any thoughts on progressing this one?  It's getting close to missing the NB17 cut too.

lkishalmi:
Unfortunately `Files.copy` made other tests fail in masterfs and filesystems module. I'll revert to the previous implementation.

lkishalmi:
BTW. This PR marks all checks passed as it was missing the `Platform` label.

platform/openide.filesystems/src/org/openide/filesystems/FileUtil.java
```diff
@@ -832,10 +846,22 @@ public static File toFile(FileObject fo) {
         assert assertNormalized(retVal, BaseUtilities.isMac()); // #240180
         return retVal;
     }
+    
+    /** Finds appropriate java.nio.file.Path to FileObject if possible.
+     * If not possible then null is returned.
+     * This is the inverse operation of {@link #toFileObject}.
+     * @param fo FileObject whose corresponding Path will be looked for
+     * @return java.nio.file.Path or null if no corresponding File exists.
+     * @since 9.30
```
sdedic, line 855: 9.31 if apichanges is corrected.

PR #4662, opened 2022-09-20
Fix possible NPE while editing ANTLR Grammar

lkishalmi:
Yes, it seems some things are still lurking. Let's start to patch them.

### DIFF
--- a/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
+++ b/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
@@ -211,7 +211,7 @@ public final class Antlr4ParserResult extends AntlrParserResult<ANTLRv4Parser> {
 
             @Override
             public void exitLexerRuleSpec(ANTLRv4Parser.LexerRuleSpecContext ctx) {
-                if (ctx.FRAGMENT() == null) {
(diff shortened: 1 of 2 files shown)


java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
```diff
@@ -211,7 +211,7 @@ protected ParseTreeListener createStructureListener() {
 
             @Override
             public void exitLexerRuleSpec(ANTLRv4Parser.LexerRuleSpecContext ctx) {
-                if (ctx.FRAGMENT() == null) {
+                if ((ctx.FRAGMENT() == null) && (ctx.TOKEN_REF() != null)) {
```
matthiasblaesing, line 214: This reminds me of a question I had: I understand, that fragments don't end up in the parse tree, but from a user perspective, I'm very interested in them (they are after all primitives I can use to structure my grammar), so it would be good to show fragments in the structure tree.

matthiasblaesing (approved):
The guards look save. I left an inline question, but that is independent of this fix.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
lkishalmi: True.

Issue #4629, opened 2022-09-13
Access/visibility declaration lost when changing with auto complete


alex-code:
### Apache NetBeans version

Apache NetBeans 15 release candidate

### What happened

I'm on NB15 using it for PHP dev.

When changing an existing visibility declaration on a class property or method pressing enter or clicking it from the auto complete list will remove it.

### How to reproduce

Create a new method in class.
```php
private function test() {}
```
Highlight `private` and start to type `pu` to get the auto complete list.
![image](https://user-images.githubusercontent.com/1797626/189904691-743a66ca-d5f0-441f-b0cb-2f7b338863e3.png)
Hit enter or double click `public` from the list.

The visibility declaration is cleared.

### Did this work correctly in an earlier version?

Apache NetBeans 14

### Operating System

Windows 10 Pro

### JDK

13.0.2+8

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

troizet:
The error is reproduced on NetBeans 15 on 
```
Debian GNU/Linux 10 (buster)
OpenJDK 64-Bit Server VM 11.0.16+8-post-Debian-1deb10u1
```
and on
```
Ubuntu 20.04.4 LTS
OpenJDK 64-Bit Server VM 11.0.16+8-post-Ubuntu-0ubuntu120.04
```

<details>
  <summary>Stacktrace error:</summary>

```
java.lang.NullPointerException
	at org.netbeans.lib.editor.codetemplates.CodeTemplateInsertHandler.shouldOpenCompletionAfter(CodeTemplateInsertHandler.java:518)
	at org.netbeans.lib.editor.codetemplates.CodeTemplateInsertHandler.stateChanged(CodeTemplateInsertHandler.java:469)
	at org.netbeans.lib.editor.codetemplates.textsync.TextRegionManager.fireEvent(TextRegionManager.java:215)
	at org.netbeans.lib.editor.codetemplates.textsync.TextRegionManager.activateTextSync(TextRegionManager.java:415)
	at org.netbeans.lib.editor.codetemplates.textsync.TextRegionManager.releaseLastGroups(TextRegionManager.java:425)
	at org.netbeans.lib.editor.codetemplates.textsync.TextRegionManager.removeUpdateUpdate(TextRegionManager.java:751)
[...]

PR #4670, opened 2022-09-21
Fixed a NPE from issue #4629

troizet:
Hello! This PR fixes the NPE error that occurs when you change the existing visibility declaration of a class property or method when you press Enter or click it in the autocomplete list.

### DIFF
--- a/ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
+++ b/ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
@@ -515,7 +515,7 @@ public final class CodeTemplateInsertHandler implements TextRegionManagerListene
             return false;
         }
         TextSync last = removed.get(removed.size()-1).activeTextSync();
-        return last.isCaretMarker() && last.isCompletionInvoke();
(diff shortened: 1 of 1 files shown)


ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
```diff
@@ -515,7 +515,7 @@ private static boolean shouldOpenCompletionAfter(TextRegionManagerEvent evt) {
             return false;
         }
         TextSync last = removed.get(removed.size()-1).activeTextSync();
-        return last.isCaretMarker() && last.isCompletionInvoke();
+        return (last != null && last.isCaretMarker() && last.isCompletionInvoke());
```
mbien, line 518: nitpick: would you mind removing the outer braces (....)? I don't think it is helping with readability here ;)

mbien (approved):
thanks for fixing this

ide/editor.codetemplates/src/org/netbeans/lib/editor/codetemplates/CodeTemplateInsertHandler.java
mbien: thanks

mbien (approved):
good fix. thanks!

PR #4671, opened 2022-09-21
add semantic highlight for when keyword

MegJayan:
Add semantic highlight for when keyword introduced as part of jep427


java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/SemanticHighlighterBase.java
```diff
@@ -657,6 +658,18 @@ public Void visitRequires(RequiresTree tree, Void p) {
             return super.visitRequires(tree, p);
         }
 
+        @Override
+        public Void visitBindingPattern(BindingPatternTree node, Void p) {
```
lahodaj, line 662: I believe `when` is part of the `case` structure, not the binding pattern structure. Shouldn't the detection of the contextual keyword be in `visitCase`?

MegJayan: @lahodaj : addressed

lahodaj: Is there a need to keep the `when` handling here in `visitBindingPattern`? Also, it might make sense to also have a test for the deconstruction pattern with a guard.

Thanks!

MegJayan: ok sure, will handle

lahodaj (approved):
Looks good to me.

PR #4672, opened 2022-09-21
Avoid using an invalid JavaPlatform

errael:
This fixes #4651 in my case.

PR #4678, opened 2022-09-23
Improve java platform selection logic for hints.

mbien:
followup to #4672. 

 - old logic didn't check installed platforms if the spec version
   of the default platform was null
 - cap at feature version of nb-javac (or javac, if uninstalled)
 - default platform is the fallback
 - converted logic to streams

This code runs when code inspections are manually run (e.g. refactor -> inspect and transform). The editor is probably just using the default platform?

I am actually not sure why this selection logic is needed at all - but this should make it less buggy at least.

### DIFF
--- a/java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
+++ b/java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
@@ -109,6 +109,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Stream;
+import javax.lang.model.SourceVersion;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.AnnotationValueVisitor;
@@ -160,6 +162,7 @@ import org.netbeans.spi.editor.hints.Severity;
 import org.netbeans.spi.java.classpath.support.ClassPathSupport;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.FileUtil;
+import org.openide.modules.SpecificationVersion;
 import org.openide.util.Lookup;
 import org.openide.util.NbCollections;
 import org.openide.util.WeakListeners;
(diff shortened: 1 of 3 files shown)


BradWalker (approved):

BradWalker (approved):
Looks good..

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/Utilities.java
```diff
@@ -1090,7 +1091,7 @@ public synchronized ClasspathInfo createUniversalCPInfo() {
                     .filter((p) -> "j2se".equals(p.getSpecification().getName()))
                     .filter((p) -> p.getSpecification().getVersion() != null)
                     .filter((p) -> p.getSpecification().getVersion().compareTo(maxVersion) < 0)
-                    .max((p1, p2) -> p1.getSpecification().getVersion().compareTo(p2.getSpecification().getVersion()))
+                    .max(Comparator.comparing((p) -> p.getSpecification().getVersion()))
```
mbien, line 1094: to do this I had add generics to `SpecificationVersion`'s `Comparable` interface.

i believe this should be safe since the class is final and the `compareTo` method would have thrown a `ClassCastException` anyway when anything other than SpecificationVersion would be passed as parameter.

This requires a sig file change. 

@neilcsmith-net could you take a look at the second commit if that is ok? The second commit is entirely optional, no big deal to drop it. I didn't bump any versions.

neilcsmith-net: @mbien I _think_ that's fine.  And leads to nicer code here.  Still binary compatible because of bridge method not picked up by sigtest?  @jtulach the best person to check with on sigtests and compatible API updates.

neilcsmith-net:
Looks OK at a glance.  I'm curious whether a platform that passes `isValid` should ever return a `null` spec version?

mbien:
> Looks OK at a glance. I'm curious whether a platform that passes `isValid` should ever return a `null` spec version?

@neilcsmith-net `isValid` is very basic unfortunately:
```
    public boolean isValid() {
        return !getInstallFolders().isEmpty();
    }
```
Should we move the speck version check into that method?
I am a bit afraid to move more checks there because who knows what else is affected by that. E.g what if NB thinks the default platform isn't valid on some exotic setups.

neilcsmith-net:
> isValid is very basic unfortunately:

Yes, I saw.

> Should we move the spec version check into that method?

Probably not at this point?  But could in mind better validation of platforms for NB17?

neilcsmith-net (approved):

PR #4681, opened 2022-09-25
Limit bootstrap libraries to java.se

jtulach:
When running the IDE on GraalVM-java11-21.3 and working with `mx` projects I am getting reports that `org.graalvm.polyglot` package is defined in other module. Let's eliminate that problem by `--limit-modules java.se`


jlahoda (approved):
Great, if it works for the MX projects!

PR #4694, opened 2022-09-27
Fixed JBoss EAP regex patterns to match two digit minor/patch versions

akronenw:
This PR fixes a flaw in the Wildfly server log line parsing regexes.
In the old version, JBoss EAP 6 (and 7) servers with two digit patch version (e.g. 6.4.23) are not detected, same for minor version.
In EAP 7 it is possible that there is an additional whitespace between the time it took to start the server and the time unit (e.g. 4700 ms instead of 4700ms).
That issue is also fixed.

### DIFF
--- a/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/ide/WildflyOutputSupport.java
+++ b/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/ide/WildflyOutputSupport.java
@@ -34,15 +34,14 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Pattern;
 import org.netbeans.api.extexecution.ExecutionDescriptor;
 import org.netbeans.api.extexecution.ExecutionService;
 import org.netbeans.api.extexecution.base.input.InputProcessor;
 import org.netbeans.api.extexecution.base.input.InputProcessors;
+import org.netbeans.api.extexecution.base.input.InputReader;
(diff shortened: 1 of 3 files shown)


mbien (approved):
thanks for the fix and the tests.

I diffed the patterns and changes make sense.

ebarboni:
Hi, as you are first time contributor you should take a look at https://github.com/apache/netbeans/blob/master/.github/PULL_REQUEST_TEMPLATE.md and make it explicit in the comment I guess

matthiasblaesing:
@akronenw the question @ebarboni raises is: Did you read the license statement at the head of the files you modified/created and are you aware, that you donate the change to the ASF and thus allow the ASF to relicense them if necessary at a later time.

akronenw:
Hello @ebarboni, I didn't get your point at first.
Thanks @matthiasblaesing for clarifying.

I did read the linked PR template and the guide to contributing code before creating the PR. As I don't understand it that way, was I supposed to state in the PR comment that I read those guides?

mbien:
@akronenw The text can be easily overlooked, we simply checked if you have read it since you are a first time contributor.
```
By opening a pull request you confirm that, unless explicitly stated otherwise, the changes

    are all your own work, and you have the right to contribute them.
    are contributed solely under the terms and conditions of the Apache License 2.0 (see section 5 of the license for more information).
```
do you agree with this? yes/no?

akronenw:
Yes, I agree.


Michael Bien ***@***.***> schrieb am So., 2. Okt. 2022, 07:55:

matthiasblaesing:
This is reviewed, github actions and travis are both happy, author declared understanding, that this is a donation to the ASF. Lets get this in.

@akronenw thank you for the fix.

PR #4695, opened 2022-09-27
Unable to register Payara 6.x in Apache NetBeans 15

jGauravGupta:
This PR adds Payara 6 configuration xml to enable registration of Payara Server 6.x in the Apache NetBeans IDE.

### DIFF
--- a/enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/server/config/PayaraV6.xml
+++ b/enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/server/config/PayaraV6.xml
@@ -33,6 +33,8 @@ under the License.
         <platform version="17"/>
     </java>
     <javaee version="9.1.0">
+        <profile version="9.0.0" type="web"/>
+        <profile version="9.0.0" type="full" check="full"/>
         <profile version="9.1.0" type="web"/>
         <profile version="9.1.0" type="full" check="full"/>
         <module type="war"/>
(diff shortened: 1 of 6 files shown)


enterprise/payara.tooling/src/org/netbeans/modules/payara/tooling/server/config/PayaraV6.xml
```diff
@@ -0,0 +1,103 @@
...
+        <platform version="12"/>
+        <platform version="13"/>
+        <platform version="14"/>
+        <platform version="15"/>
+        <platform version="16"/>
+        <platform version="17"/>
+    </java>
+    <javaee version="10.0.0">
+        <profile version="10.0.0" type="web"/>
+        <profile version="10.0.0" type="full" check="full"/>
```
matthiasblaesing, line 37: This looks strange - previous version of payara declare support for multiple profiles - was that support never there or was it removed? (I compared with PayaraV5.xml).

I doubt this can be parsed, as the constants (version 10.0.0) are not defined in JavaEEProfile.java and it looks as if it should be. I assume this ties back into general EE support - is the support there yet?

jGauravGupta: Thanks @matthiasblaesing for reviewing the PR, I have set the 9.0.0 and 9.1.0 to Payara 6 profile and add the 10.0.0 version to Jakarta EE 10 support PR (WIP).

matthiasblaesing:
The test failures are expected (fix is currently being worked on), so if this was tested and works, I suggest that you squash these and get them in.

Issue #4441, opened 2022-07-27
UnsatisfiedLinkError with detached window


errael:
### Apache NetBeans version

Apache NetBeans 14

### What happened

With a detached NetBeans `Output Window`, I frequently see the following exception.

```
INFO [org.netbeans.core.nativeaccess.NativeWindowSystemImpl]
java.lang.UnsatisfiedLinkError: libjawt.so: cannot open shared object file: No such file or directory
	at com.sun.jna.Native.getWindowHandle0(Native Method)
	at com.sun.jna.Native$AWT.getComponentID(Native.java:2383)
	at com.sun.jna.Native.getComponentID(Native.java:326)
	at com.sun.jna.platform.WindowUtils$X11WindowUtils.getDrawable(WindowUtils.java:1747)
	at com.sun.jna.platform.WindowUtils$X11WindowUtils.access$900(WindowUtils.java:1564)
	at com.sun.jna.platform.WindowUtils$X11WindowUtils$2.run(WindowUtils.java:1769)
	at com.sun.jna.platform.WindowUtils$NativeWindowUtils.whenDisplayable(WindowUtils.java:398)
	at com.sun.jna.platform.WindowUtils$X11WindowUtils.setWindowAlpha(WindowUtils.java:1791)
	at com.sun.jna.platform.WindowUtils.setWindowAlpha(WindowUtils.java:1992)
[catch] at org.netbeans.core.nativeaccess.NativeWindowSystemImpl.setWindowAlpha(NativeWindowSystemImpl.java:91)
	at org.netbeans.core.windows.FloatingWindowTransparencyManager.turnTransparencyOff(FloatingWindowTransparencyManager.java:155)
	at org.netbeans.core.windows.FloatingWindowTransparencyManager.toggleFloatingWindowTransparency(FloatingWindowTransparencyManager.java:141)
	at org.netbeans.core.windows.FloatingWindowTransparencyManager$1$1$1.run(FloatingWindowTransparencyManager.java:74)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:771)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:722)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:716)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
[...]

PR #4696, opened 2022-09-28
Update FlatLaf from 2.4 to 2.5

DevCharly:
Update FlatLaf to v2.5.

Changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/2.5

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-691E1F12C78F42E0DFD9FC37BC599042A1EBDD0D com.formdev:flatlaf:2.4
+53B5104BDF92CBC7A650E075ED7296DD0AB77CF3 com.formdev:flatlaf:2.5
--- a/platform/libs.flatlaf/external/flatlaf-2.5-license.txt
+++ b/platform/libs.flatlaf/external/flatlaf-2.5-license.txt
@@ -1,7 +1,7 @@
 Name: FlatLaf Look and Feel
 Description: FlatLaf Look and Feel
-Version: 2.4
-Files: flatlaf-2.4.jar
+Version: 2.5
+Files: flatlaf-2.5.jar
 License: Apache-2.0
 Origin: FormDev Software GmbH.
 URL: https://www.formdev.com/flatlaf/
--- a/platform/libs.flatlaf/manifest.mf
+++ b/platform/libs.flatlaf/manifest.mf
@@ -1,6 +1,7 @@
 Manifest-Version: 1.0
 OpenIDE-Module-Localizing-Bundle: org/netbeans/libs/flatlaf/Bundle.properties
 OpenIDE-Module: org.netbeans.libs.flatlaf/1
+OpenIDE-Module-Install: org/netbeans/libs/flatlaf/Installer.class
 OpenIDE-Module-Specification-Version: 1.12
 AutoUpdate-Show-In-Client: false
 OpenIDE-Module-Implementation-Version: 2.5
--- a/platform/libs.flatlaf/manifest.mf
+++ b/platform/libs.flatlaf/manifest.mf
@@ -3,4 +3,4 @@ OpenIDE-Module-Localizing-Bundle: org/netbeans/libs/flatlaf/Bundle.properties
 OpenIDE-Module: org.netbeans.libs.flatlaf/1
 OpenIDE-Module-Specification-Version: 1.12
 AutoUpdate-Show-In-Client: false
-
+OpenIDE-Module-Implementation-Version: 2.5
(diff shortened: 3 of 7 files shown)


matthiasblaesing:
Change looks sane to me, however I suggest to include this in this change set (if in doubt, feel free to squash it into your commit):

https://github.com/apache/netbeans/commit/66716b5506d91257f8ea136ffb11a669bac53f3a

The intention is to fix the issue described in #4619. @neilcsmith-net it would be nice if you could have a look.

DevCharly:
@matthiasblaesing Done. Thanks.

DevCharly:
@mbastian could you try whether the added implementation version works with your FlatLaf-SwingX module?

neilcsmith-net:
@matthiasblaesing that change looks a good idea.

I wonder if we should be pre-extracting the native libs similarly to [how JNA module does it](https://github.com/apache/netbeans/blob/master/platform/libs.jna/nbproject/project.properties) here?

Aside @DevCharly - while looking at FlatLaf code for native loading noticed the comments on libjawt. Similar to #4441 and https://github.com/java-native-access/jna/pull/1422 Probably good to always preload.

matthiasblaesing:
> @matthiasblaesing that change looks a good idea.

Thanks for having a look.
 
> I wonder if we should be pre-extracting the native libs similarly to [how JNA module does it](https://github.com/apache/netbeans/blob/master/platform/libs.jna/nbproject/project.properties) here?

In theory a good idea, in practice extracting libraries needs a way to force loading through `System#loadLibrary` with the basename only. In FlatLaf the loading always goes through extraction and loading target with full path. We'd need to modify this.

neilcsmith-net:
> In FlatLaf the loading always goes through extraction and loading target with full path.

Yes, I saw.  We'd need to set [this](https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatSystemProperties.java#L164) in the module installer.

Let's merge as is, but keep this in mind if we get any linking error reports.

matthiasblaesing:
@neilcsmith-net thank you for the pointer - I looked one level to deep. What do you think about this:

https://github.com/apache/netbeans/commit/6781772b899fdfeaf855ed6b20d071b8bcf5fe60

@DevCharly the referenced commit should change the library loading, so that the library is loaded from the pre-extracted version in the modules directory. The issue with self extracting libraries is two fold: The first problem is cleanup - windows only lets you remove libraries after the users are closed, the second problem is on linux the temp directory could be mounted `noexec`, which also prevends library loading. As NetBeans needs to be executed, it can be expected, that the modules are located in an executable location.

neilcsmith-net:
Thanks @matthiasblaesing  Possibly should go directly in `modules/lib` rather than `modules/lib/flatlaf` though?  The lib folder has a defined structure.

matthiasblaesing:
@neilcsmith-net valid point updated to:

https://github.com/apache/netbeans/commit/be4dba61b9f6aa74089798095663fa3b917923d8

@DevCharly if you consider this, please use this commit, not the previous one, I replaced it based on Neils comments.

DevCharly:
@matthiasblaesing have added your commit (removed an unused import)

> @DevCharly an option in FlatLaf to prefer trying to load via System::loadLibrary without full path would be great in future!

I wonder where Java/NetBeans searches for libraries if System::loadLibrary is used?

neilcsmith-net:
> I wonder where Java/NetBeans searches for libraries if System::loadLibrary is used?

In NetBeans, from `modules/lib` where the commit puts it. The classloader is used for locating native libraries, and the NetBeans classloaders look there.

DevCharly:
> I wonder if we should be pre-extracting the native libs similarly to [how JNA module does it](https://github.com/apache/netbeans/blob/master/platform/libs.jna/nbproject/project.properties) here?

Is it possible that this is broken?
When I start NB 15 (on Windows), JNA creates `jna14122183419442360685.dll` in `C:\Users\abc\AppData\Local\Temp\jna--1361632413`.

Looking at the JNA code, it seems that JNA uses `System.loadLibrary()` only if system property `jna.nosys` is set to `false`, which seems not the case.
https://github.com/java-native-access/jna/blob/bf60e51eace6dffa18548019e2ba398ff84904ef/src/com/sun/jna/Native.java#L1003

DevCharly:
The implementation in above line (in previous post) seems not fit to the Javadoc:
https://github.com/java-native-access/jna/blob/bf60e51eace6dffa18548019e2ba398ff84904ef/src/com/sun/jna/Native.java#L83

Also hint at following line is IMHO not useful because `jna.nosys=true` is the default:
https://github.com/java-native-access/jna/blob/bf60e51eace6dffa18548019e2ba398ff84904ef/src/com/sun/jna/Native.java#L234

neilcsmith-net:
Well, `jna.nosys` used to default to false.  Changed in JNA 5.x I think (@matthiasblaesing ?).  Possible this should have been updated in the installer here - https://github.com/apache/netbeans/blob/master/platform/libs.jna/src/org/netbeans/libs/jna/Installer.java#L29  Take to another issue?

matthiasblaesing:
The jna issue should be fixed by #4736

neilcsmith-net:
@DevCharly sorry, looks like our PRs conflicted in changes to `Installer`.  Are you able to update and resolve?

neilcsmith-net:
Opened #4803 with conflicts resolved instead.

Issue #4702, opened 2022-09-29
Gradle build actions NPE and visually inaccurate state


errael:
### Apache NetBeans version

Apache NetBeans 15

### What happened

NPE and inconsistent visual state

### How to reproduce

gradle projects are created with NewProject > JavaWithGradle > JavaClassLibrary

Get to this dialog with `GradleProject > properties > Build > BuildActions`

First note that although nothing is selected in the `ConfiguredActions` list, both the Disable and Reset buttons are enabled. Note that when this first comes up is the only time the configuration details shows up without a ConfiguredActions selection. If you select an action, then deselect, CTRL click, you will see `No Selected Action`. I like seeing the details area when it first comes up to get an idea of what you can do with an action. But if only came up when something was selected, then there would be no issue with the Reset button.

- When it first comes up, the Reset button is enabled. Click it and there is an NPE in the NB log.

The rest of this is about stuff that seems/looks wrong, but I'm not familiar enough with this dialog to be certain. For example, what does boldface mean in `Configured Actions` list? Should some actions start in a disabled state or is there a reason that Restore doesn't restore?

- Boldface in `Configured Actions` list inconsistent
    1. Open `Build Actions` dialog
    2. In `Override in Configuration` combo select `Continuous Mode`
    3. In `Configure Action` combo select `build`
    4. In `Configure Action` combo select `debug`
    5. Observe, `debug` is in boldface
    6. click OK to close dialog
    7. Re-open `Build Actions` dialog
    8. In `Override in Configuration` combo select `Continuous Mode`
    9. Observe that `debug` is not boldface

Starting with fresh project
- Disable button and related (first 4 steps same as previous)
    1. Open `Build Actions` dialog
    2. In `Override in Configuration` combo select `Continuous Mode`
    3. In `Configure Action` combo select `build`
    4. In `Configure Action` combo select `debug`
    5. Select `build`, `Disable`
    6. Select `debug`, `Disable`
    7. Select `build`, `Restore`
    8. Select `debug`, `Restore`
    9. Observe, `debug` is not Restored
    10. With `debug` still selected, click `Reset`
    11. Again, in `Configure Action` combo select `debug`
    12. Observe that debug is enabled.
    13. Again select `debug`, Disable
    14. OK/close Dialog
    15. Re-open dialog
    16. In `Override in Configuration` combo select `Continuous Mode`
    17. Select `debug`, click `Reset`
    18. Again, in `Configure Action` combo select `debug`
[...]

PR #4703, opened 2022-09-29
Fix npe from gradle build actions dialog

errael:
This fixes the NPE from #4702


lkishalmi (approved):
Looks sane.

mbien:
want to squash the merge commit before integration or should we leave it as is?

errael:
> squash the merge commit

Oops, not sure how that got there, I think github contributed that when I wasn't looking (or maybe wasn't paying attention). Look OK now.

I use hg-git, sometime things don't mesh quite right.

mbien:
all green && approved -> merging

@errael thanks for fixing this

PR #4711, opened 2022-10-01
Override gradle's compiler version from compiler args.

### LINKED ISSUE #4704: Gradle setting wrong NetBeans jdk version
### Apache NetBeans version

Apache NetBeans 15

### What happened

This issue probably originates from some gradle behavior. Gradle 6.8.x limits sourceCompatibility and targetCompatibility to 18; using compilerArgs bigger numbers can be passed to the compiler.

Looking at the attached png, I'm guessing that NetBeans is told that the project is at 18. Shouldn't it allow the compilerArgs override the 18 as far as NetBeans is concerned?

I'm openning this in the hopes that NB-16 will use the correct jdk version.

### How to reproduce

Can provide project if that's needed.

No / Don't know

### Operating System

Linux harmony 5.19.0-76051900-generic #202207312230166078056622.04~9d60db1

### JDK

openjdk version "11.0.2" 2019-01-15

### Apache NetBeans packaging

Own source build

### Anything else

_No response_

No

### Code of Conduct

Yes

errael:
Fix #4704. This applies to both `-source` and `-target`. This information is used by editor/hints.

With this patch and the following build.gradle excerpt
```
tasks.withType(JavaCompile).each {
    it.options.compilerArgs << '--source' << '19'
    it.options.compilerArgs << '--target' << '11'
    it.options.compilerArgs << '-Xlint:deprecation' << '--enable-preview'
}

sourceCompatibility = 9
targetCompatibility = 9
```
There's

and the following has no errors in the editor.
```
        Object o2 = 3;
        return switch (o2) {
          case Integer i when i >= 0 -> 3; // positive integers
          case Integer i -> 4;             // negative integers
          case default -> 2;
        };
```

### DIFF
--- a/java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
+++ b/java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
@@ -20,6 +20,7 @@
 package org.netbeans.modules.gradle.java.api;
 
 import org.netbeans.modules.gradle.spi.Utils;
+
 import java.io.File;
 import java.io.Serializable;
 import java.nio.file.Path;
@@ -33,7 +34,9 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
+
 import static org.openide.util.NbBundle.Messages;
 
 public final class GradleJavaSourceSet implements Serializable {
@@ -134,7 +137,7 @@ public final class GradleJavaSourceSet implements Serializable {
      * @return the defined source compatibility or "1.5"
      */
     public String getSourcesCompatibility(SourceType type) {
-        return sourcesCompatibility.getOrDefault(type, DEFAULT_SOURCE_COMPATIBILITY);
+        return fixJavaCompatibility(type, "-source", sourcesCompatibility).orElse(DEFAULT_SOURCE_COMPATIBILITY);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mbien:
please label PRs before creating them. This will be important in future #4431, otherwise this PR would run no gradle jobs for example.

errael:
> please label PRs before creating them

I knew about the labeling and CI. I'll try to keep that in mind. I didn't think I was able to change/add labels.

mbien:
> > please label PRs before creating them
> 
> I knew about the labeling and CI. I'll try to keep that in mind. I didn't think I was able to change/add labels.

oh my mistake - disregard what i said. I was sure you were already committer since you are so active (which is great!). I think you have to have at least committer role to be able to label things.

lkishalmi (changes requested):
Well, the implementation is Ok. The place is not so much. This is a kind of de-serialization code. So I would rather fix this at the source: https://github.com/apache/netbeans/blob/d27cbbce190c4aec0f61cda390876ea8cedd98e5/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java#L357

or right before the usage:

https://github.com/apache/netbeans/blob/d27cbbce190c4aec0f61cda390876ea8cedd98e5/java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java#L137

errael:
I thought a builder would be good; immutable description. Looking at your two proposed locations, couldn't get breakpoints to work in NbProjectInfoBuilder, so that made up my mind.

errael:
> [errael](https://github.com/errael) requested review from [lkishalmi](https://github.com/lkishalmi) and removed request for [sdedic](https://github.com/sdedic) 7 minutes ago

Not sure how I removed the request for @sdedic, not my intention. github bug?

java/gradle.java/src/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.java
```diff
@@ -162,7 +165,36 @@ public String getTargetCompatibility() {
...
+     * For example for source, flag is "-source", and args is
+     * "... --source 13 ..." then return "13". If not in args
+     * then don't change the compatibility, return the current value.
+     */
+    private Optional<String> fixCompatibility(SourceType sourceType, Map<SourceType, String> compatibilityMap) {
+        String compatibility = compatibilityMap.get(sourceType);
+        if(sourceType == SourceType.JAVA) { // only fixup for java
+            List<String> args = getCompilerArgs(sourceType);
+            if(args != null && !args.isEmpty()) {
+                String flag = compatibilityMap == sourcesCompatibility ? "-source" : "-target";
```
sdedic, line 185: Nitpick: maybe pass the flag source/target explicitly rather than comparing Map instances ?

errael: The first implementation did that (the one that was in the wrong place). I'll make that change after any other suggested changes have time to roll in. Thanks for the comment.

lkishalmi: Well, I'd reduce the number of ifs: getCompilerArgs() never returns with `null`, so line 184 could go. The double if in line 188 and 190 could be combined...

errael: I remember now why I compared to map instance; because -source/-target are/might-be specific to java. I'll change the method name to `fixJavaCompatibility`. I'd like to take out the souceType argument, but then the check for JAVA would, would have to be in the caller.

errael: K

mbien:
i am no heavy gradle user. But there is also the `--release` flag (JEP 247) which is in many ways better than using source+target since it enforces bytecode *and* API compatibility and avoids a bunch of nasty bugs. Not sure if this is relevant here, e.g what if the build sets release not source - is this handled too?

errael:
Yikes. Thanks for something real/detectable.

What to do if `--release` and one of the others? In any event, a compile will fail.

Arbitrarily use the value with `--release`. Unless want to consider error dialogs or somesuch, but then why stop there...

mbien:
> What to do if `--release` and one of the others? In any event, a compile will fail.

it replaces both, if you mix, javac should complain and fail as far as I know. It is similar to setting both source and target to the same value, but more powerful since javac will use ct.sym files which are basically method/class signature diffs and check that your code could actually run on the target JDK APIs.

The maven template uses the source/target tags too still, if we end up adding some kind of JDK or language level selector to the new maven project wizard we could let it use that too if java 9+ is selected.

errael:
Any further suggestions, comments, questions?

lkishalmi (approved):
Thank you @errael !

mbien (approved):
looks good, thanks for making it aware of the `--release` flag too

PR #4714, opened 2022-10-01
Check for EnvVar/SysProp for preferred screen

errael:
I've been running with this patch for over 6 months (since I started using multiple screens). It allows the user to specify a "preferred screen", this is the screen that NetBeans starts up in. First looks for an environment variable `NETBEANS_PREFERRED_SCREEN`, if found use the numeric value as index into ScreenDevices; if not found then try for a SystemProperty by that name.

Originally I was considering a PR with this and other fixes for dialogs that open up on the _wrong_ screen, typically that would be screen 0 when NetBeans' main window is on screen 1. I'm opening this as a draft PR, though it could be merged as is, or maybe extended with related changes. It provides some context/reference for stuff I'm looking at. Looking for some opinions/comments about this whole issue.

I've experienced a situation related to what's described in `[platform] DialogDescriptor/NotifyDescriptor and multiple screens`, last message April 22, 2022. But I've never seen anything reproducible. I'm beginning to think that `KeyboardFocusManager.getActiveWindow()` return null and NetBeans decides to display dialog on default screen. But I can't get it to reproduce.

There also dialogs scattered throughout NetBeans that don't specify a parent, they consistently end up on the wrong screen. They could easily be changed to parent off of the NB main window, or with more effort parent from an appropriate screen.

### DIFF
--- a/platform/openide.util.ui/src/org/openide/util/Utilities.java
+++ b/platform/openide.util.ui/src/org/openide/util/Utilities.java
@@ -28,6 +28,7 @@ import java.awt.Dimension;
 import java.awt.Event;
 import java.awt.Frame;
 import java.awt.GraphicsConfiguration;
+import java.awt.GraphicsDevice;
 import java.awt.GraphicsEnvironment;
 import java.awt.Image;
 import java.awt.Insets;
@@ -72,6 +73,7 @@ import java.util.Vector;
 import java.util.WeakHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.swing.Action;
 import javax.swing.Icon;
 import javax.swing.JMenuItem;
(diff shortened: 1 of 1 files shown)


errael:
Updated to ready for review. Though it's a simple patch, the idea of giving the user a way to specify which screen to start up in is new (AFAIK).

neilcsmith-net:
My inclination would be to try and use the display that NetBeans was on when last used rather than adding additional configuration options if at all possible.

errael:
> My inclination would be to try and use the display that NetBeans was on when last used rather than adding additional configuration options if at all possible.

I don't disagree, especially considering how to document/expose it so it's actually useful.

errael:
Forget the Env var. Focus on mis-use of null parent for dialog placement through NB APIs in  a future PR.

PR #4719, opened 2022-10-02
Add ability to load custom FlatLaf properties from user config.

neilcsmith-net:
Adds ability to load custom FlatLaf property files from userdir. Checks for a `LookAndFeel` folder in config, and registers with FlatLaf. Also registers the folder and files for options export. Give user ability to override anything using FlatLaf property support documented at https://www.formdev.com/flatlaf/properties-files/

eg. file in `<userdir>/config/LookAndFeel/FlatLaf.properties` with 

```
@accentColor = #d22
@background = #ddd
```

Screenshot from 2022-10-02 17-31-18

Added button to open file in editor inside FlatLaf options panel, along with basic default content / instructions.

Screenshot from 2022-10-05 12-23-41

### DIFF
--- a/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
+++ b/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
@@ -99,6 +99,14 @@
                         <specification-version>8.46</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.openide.filesystems</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <specification-version>9.30</specification-version>
+                    </run-dependency>
+                </dependency>
             </module-dependencies>
             <public-packages/>
         </data>
--- a/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
+++ b/platform/o.n.swing.laf.flatlaf/nbproject/project.xml
@@ -107,6 +107,22 @@
                         <specification-version>9.30</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
+                    <code-name-base>org.openide.loaders</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <specification-version>7.88</specification-version>
+                    </run-dependency>
+                </dependency>
+                <dependency>
+                    <code-name-base>org.openide.nodes</code-name-base>
+                    <build-prerequisite/>
+                    <compile-dependency/>
+                    <run-dependency>
+                        <specification-version>7.63</specification-version>
+                    </run-dependency>
+                </dependency>
             </module-dependencies>
             <public-packages/>
         </data>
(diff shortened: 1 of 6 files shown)


lkishalmi (commented):
Looks good to me, however this one could be considered as an API, so it shall be documented.

neilcsmith-net:
Thanks @lkishalmi Slightly sceptical on that point, but if so, where?  `arch.xml`?  We're missing a bunch of those around preferences and import/export.  Theoretically I guess the existing preferences here should be documented.  Got a pointer to similar situation to this one?

lkishalmi:
Yes. Theoretically it should be in the `arch.xml` and `apichanges.xml`. Though as `public-packages` is empty, I do not think that the javadoc would be included for the API.  Anyway, I can't think it would do any wrong to put the info there.

About the missing documentation of the existing preferences. Probably it would be useful to start to collect those to a single place then compile into our documentation.

neilcsmith-net:
Thanks @lkishalmi I've decided not to add in this PR (unless you consider a blocker?).  An `arch.xml` might be a good idea to cover FlatLaf preferences, etc. but that's outside of this change.

I was specifically curious whether there was an `arch.xml` that specified options export.  eg. Templates seems to treat this as private implementation detail as far as I can see?

I'm working on a basic UI to open the relevant property file, but would like to keep the storage as a private implementation detail for now.  If this works, I was already thinking of moving the `LookAndFeel` config folder registration and provide access API in `o.n.swing.plaf` later (NB17+).  That would then be documented, and is also why I named the folder that way.

neilcsmith-net:
Force pushed the first commit to use `URL` rather than `File` (in case the folder isn't a file).

Added a second commit with button to open the custom properties file on the FlatLaf tab in options. The file is created if it does not exist with some basic info on usage. Not sure about the additional dependencies - although only other way would be to consider splitting off FlatLaf options into a separate module in future.

lkishalmi (approved):

DevCharly (commented):
Great idea 👍 
Works as expected 😄 

However have some concerns regarding the extended options panel.
It is fine in the NetBeans IDE, but what about applications that are build on NetBeans platform?
They probably do not have editors.
Can we hide the "Advanced" panel if editors are not available?

neilcsmith-net:
Thanks @DevCharly 

On the editor dependency, and somewhat contrary to what I said earlier, this change doesn't bring in any additional dependencies. It just makes explicit some that are already brought in by the Options API.  So, some basic editor support will be there.

It is worth considering whether to move the options panel entirely to a separate module from the LAF customs in future for platform applications.  Platform apps may choose to hide this options panel too (I do already), or provide an alternative one.

DevCharly (approved):

eirikbakke:
Looks good! Just to confirm; if the properties file does not mention any specific properties, everything behaves in the same way as if the file didn't exist at all, right?

neilcsmith-net:
Thanks @eirikbakke - yes, nothing changes, except for loading the empty property file.

Thanks for reviews - merging.

PR #4726, opened 2022-10-03
Various fixes and improvements [1/3]

sdedic:
Note: this PR is the first from series of 3. I've split the development into 3 PRs since there are different areas that seem to be better reviewed separately. Other PRs in the series depend on this one.

Unrelated fixes:
- Maven fix- - change event is fired only after result refresh from reloaded project completes.
- `ArtfiactSpec` (still private !) API fixes; some project products do not have group at all, some have no version; the API should allow that.
- core module screws up `jdk.home` property when run without a launcher (that sets `jdk.home` beforehand) - visible mainly in tests.

Gradle fixes:
- during `toQuality` that should reload a project if `force` flag is true, the flag was sometimes ignored

Improvement:
- FINER and FINEST loglevel for `LegacyProjectLoader` will print out obtained info and increase verbosity of the gradle plugin (prints into the Output window of the debugging NB)

API:
- Previously `RunConfig` APIs did not add `ActionMapping`'s arguments that specify the actual grade tasks to be run. The new API allows to obtain RunConfig as if the action was about to be run.

### DIFF
--- a/extide/gradle/apichanges.xml
+++ b/extide/gradle/apichanges.xml
@@ -83,6 +83,27 @@ is the proper place.
     <!-- ACTUAL CHANGES BEGIN HERE: -->
 
     <changes>
+        <change id="regular-project-Lookup">
+            <api name="general"/>
+            <summary>Expose regular Lookup in NbGradleProject.</summary>
+            <version major="2" minor="28"/>
+            <date day="4" month="10" year="2022"/>
+            <author login="sdedic"/>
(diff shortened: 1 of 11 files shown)


ppisl (approved):
I don't see any problem

sdedic:
Unrelated failures: `com.sun.tools.javac.code.Symbol$CompletionFailure: class file for java.lang.MatchException not found` -- merging.

PR #4728, opened 2022-10-03
LSP Client: Use full complete item to complete, not just the missing suffix

matthiasblaesing:
This is the test case (typescript):

```typescript
export class test {
    demo() {
        this.runstranget|
    }

    runStrangeThings(): void {
    }
}
```

The codecompletion at the | mark will offer "runStrangeThings", but completion will yield "this.runstrangethings". The reason is, that the CC is invoced case-insensitive, so the original string needs to be replaced with the full result.


matthiasblaesing:
Hearing no objects, having run with this the last few weeks, travis and github action being happy. I'll merge this.

PR #4730, opened 2022-10-03
Support for jar & native image artifact for gradle projects [3/3]

sdedic:
Note: this is the last "feature" PR - it depends on #4729. This PR will be rebased on whatever version #4729 evolves into.

The PR builds on top of `BuildPropertiesSupport` API in Gradle core module and allows:
- support for `jar` task's artifacts: the project reports the produced jar
- support for shadowed jars, or "uberjars": the project will report the shadow jar. It is also possible to enumerate both jars (original shadowed) marked appropriately. The API is consistent for maven + gradle projects.
- support for native-image artifacts build by Micronaut infrastructure.


sdedic:
Rebased on latest master.

sdedic:
There's some issue with the tests: they fail on CI, but not locally; stay tuned.

sdedic:
Adding @lkishalmi  - I've identified a bug in recently introduced support for task dependencies in Gradle.  All gradle changes are in commit 7226affdb.

@ppisl, @dbalek  pls review the micronaut and/or maven implementation.

sdedic:
I think an explanation would be useful for why there were so many commits and attempts to fix a failing build. The reason is simple: I failed to create a proper test environment for the project infrastructure to work properly. The tests worked OK on a local machine all the time.
1. the tests run in a pristine environment, with empty Maven repository. Project info load uses offline mode, it did not download Micronaut build plugin that knows `native-image` packaging. As a result, the project was unloadable / broken and defaulted to `pom` packaging, which broke service registrations.
2. I attempted to fix that by 'priming' the project - which downloads the dependencies to the local repository. In theory, that was fine - but the test fixture lacked `ActiveDocumentProvider`. This caused the priming build to fail during startup, but the fail was silent - exception was swallowed. Again, this was locally visible only when one cleared the `~/.m2/repository` area.
3. And finally, the JVMDI classes were required for one of the Checkers plugged into NB maven infrastructure. Not having them on the classpath (as with Zulu JDK8) caused again priming build to silently fail.

As a result, I've added a bunch of logging, which I think should remain in the code -- it will help in when one will debug project metadata loading issues in Maven.

lkishalmi (approved):
Just skimmed through.
Looks good to me.

dbalek (approved):
Quick scan of the sources. Looks good.

ppisl (approved):

neilcsmith-net:
@sdedic please make sure milestone is set when merging, thanks!

PR #4736, opened 2022-10-05
jna.nosys is required so that JNA tries to load native library via System#loadLibrary

matthiasblaesing:
Without it the embedded native library is always used.


mbien:
test failure in the Java Modules job is currently caused by #4586 which made it into master

only mentioning it in case it causes confusion, its not your PR.

Issue #4635, opened 2022-09-14
PHP Formatting: No space between 'AND !$var'.


NotherGithubAccount:
### Apache NetBeans version

Apache NetBeans 15 release candidate

### What happened

`if ($a AND !$b)` formats to `if ($a AND!$b)`. It works correctly with &&, the behavior is the same with OR and ||.

When checking the "Unary Operator" box in Tools -> Options -> Editor -> Formatting for PHP and spaces, the result is
`($a AND ! $b)`
While that works as advertised, imho there should always be a space after AND/OR.
 

### How to reproduce

Create corresponding test code and select source->format.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Debian 11 5.18.0-0.deb11.4-amd64

### JDK

OpenJDK Runtime Environment (build 11.0.16+8-post-Debian-1deb11u1)

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

Afaik, always.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

junichi11:
Reproducible. Thanks!

NotherGithubAccount:
Thanks for the quick fix!

junichi11:
You are welcome :)

PR #4741, opened 2022-10-06
Fix the formatting for the textual operators(`AND`, `OR`, `XOR`i) #4635

junichi11:
#4635 

- `WHITESPACE_BEFORE_BINARY_OP` and `WHITESPACE_AFTER_BINARY_OP` are added to `&&` and `||`
- There is the problem if we add them to `AND`, `OR`, and `XOR`
- See: https://bz.apache.org/netbeans/show_bug.cgi?id=240274
- e.g. if they are false, `copy($old,$new) or die("error");` -> `copy($old,$new) ordie("error");`
- Instead, add `WHITESPACE_AROUND_TEXTUAL_OP` as a new token


junichi11:

tmysik (approved):

junichi11:

tmysik:
@junichi11 Sure, whenever :)

PR #4742, opened 2022-10-06
Add update centre notifications to Check for Updates dialog

neilcsmith-net:
Quick fix for a long standing annoyance.  We've had a number of users report confusion that using `Help / Check for Updates` only shows plugin updates (which are rare, especially from us) and not new releases of the IDE. In particular, that the text reads "Your IDE is up to date!" when it's not the latest release.

This change adds any update centre notification text to the dialog if no plugin updates are available.  If you want to test, change the NetBeans Distribution update link from https://netbeans.apache.org/nb/updates/dev/updates.xml.gz to https://netbeans.apache.org/nb/updates/14/updates.xml.gz and use Check for Updates.

Image without notification (default as now).

nb-noupdates

Image with update notification from NB14 catalog.

nb-hasupdate

I've left the header text as is, although I think it's somewhat confusing too for updates.

### DIFF
--- a/platform/autoupdate.ui/src/org/netbeans/modules/autoupdate/ui/wizards/LazyOperationDescriptionStep.java
+++ b/platform/autoupdate.ui/src/org/netbeans/modules/autoupdate/ui/wizards/LazyOperationDescriptionStep.java
@@ -26,6 +26,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
 import javax.swing.event.ChangeEvent;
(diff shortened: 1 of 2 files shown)


platform/autoupdate.ui/src/org/netbeans/modules/autoupdate/ui/wizards/LazyOperationDescriptionStep.java
```diff
@@ -186,25 +198,31 @@ public void taskFinished (Task task) {
...
-                }
-                else
-                {
-                    body = new OperationDescriptionPanel (
-                        getBundle ("LazyOperationDescriptionStep_NoUpdatesWithProblems_Title"), // NOI18N
-                        getBundle ("LazyOperationDescriptionStep_NoUpdatesWithProblems"), // NOI18N
-                        "", "",
-                        false);
+                if (problems == null || problems.isEmpty()) {
+                    if (notifications == null || notifications.isEmpty()) {
```
mbien, line 202: `problems` and `notifications` can never be null, right? Both are final.

mbien (approved):
great update. Little nitpick inline.

ebarboni (approved):

ebarboni:
this is more friendly.

I think we should have a script to update the notification in the old file instead of doing that manually each release

platform/autoupdate.ui/src/org/netbeans/modules/autoupdate/ui/wizards/LazyOperationDescriptionStep.java
neilcsmith-net: True. I just kept the `notifications` logic the same as the `problems` one for consistency.  I'm going to force push an update anyway - too many exclamation marks for my liking - so I'll change these too.

neilcsmith-net:
Thanks @ebarboni 

> I think we should have a script to update the notification in the old file instead of doing that manually each release

Possibly, although I've got that process down well enough.  A script that could modify that and also help with changing modules for updates would be even better! :smile:

PR #4748, opened 2022-10-07
Add debugger listening connector to LSP.

entlicher:
Add a listening connector to the VSCode extension.
It acts as an attaching connector, but the attach is reversed. VSCode opens a listening socket and the debuggee attaches to it. `"listen":"true"` property flips the attach direction.

The debuggee needs to be started after the listening debugger session is started and with `server=n`, like:
`-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=localhost:8000`


JaroslavTulach:
Very nice and desperately needed (by me and `sbt`).

jtulach (approved):
Straightforward improvement. Thank you.

PR #4749, opened 2022-10-07
use java.net.URLEncoder/Decoder instead of buggy custom impl.

### LINKED ISSUE #4732: Task categories at bugrtacking dashboard with cyrillic symbols does not load properly after restart
### Apache NetBeans version

Apache NetBeans 15

### What happened

Task categories at bugrtacking dashboard with cyrillic symbols does not load properly after IDE restart. 
Before restart IDE:
image
After restart:
image

### How to reproduce

Just create task category with name "тест" and restart IDE. It will lost it after restart. 

No / Don't know

### Operating System

windows 10

### JDK

Oracle jdk 14, x64

### Apache NetBeans packaging

### Anything else

I was try to look at sources and I think issue at step when category name converted to name store file in ide/bugtracking/src/org/netbeans/modules/bugtracking/tasks/cache/DashboardStorage.java.
On read file step executed decodeURL func https://github.com/apache/netbeans/blob/694f319b068d33bb812274d1232fc79e264ca5ff/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/TextUtils.java#L213 which work correctly only with two-digit character codes:
https://github.com/apache/netbeans/blob/694f319b068d33bb812274d1232fc79e264ca5ff/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/TextUtils.java#L222
but cyrillic symbols after encodeUrl func contains three digits. for example 'testй' -> 'test%439.c'

No

### Code of Conduct

Yes

mbien:
fixes #4732

This isn't optimal since the methods are used to encode file names (255 char limit on windows), but since this is public API its still worth fixing since it didn't support non latin chars properly as seen in the issue.

### DIFF
--- a/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/TextUtils.java
+++ b/ide/team.commons/src/org/netbeans/modules/bugtracking/commons/TextUtils.java
@@ -19,6 +19,10 @@
 
 package org.netbeans.modules.bugtracking.commons;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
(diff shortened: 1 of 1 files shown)


matthiasblaesing (approved):
Looks sane to me. You might want to add a final `.replace('+', '%20')` as space is not percent encoded, but mapped to `+`.

6PATyCb:
@mbien Please help me with right way to add my test [#4751 ] after your current PR.

lkishalmi (approved):
Looks good to me.

PR #4751, opened 2022-10-07
added test for TextUtils.encodeURL

6PATyCb:
Added test for [#4732](https://github.com/apache/netbeans/issues/4732)


matthiasblaesing:
There is no need to propose a separate PR for this please coordinate with @mbien in #4749 whether this is the right change and if so, that the two commits are merged together (not necessarily squashed).

mbien:
@6PATyCb since we simply delegate to URLEncoder now, I don't think we need a test for it. Since this would basically test JDK classes with it. (there is no custom impl to test anymore)

also this PR has a lot of whitespace changes, we would have to clean it up first.

6PATyCb:
Accepted. 
So, it was nice experience for me.

PR #4756, opened 2022-10-08
Fixing error recovery when j.l.MatchException is missing, and patterns are present.

jlahoda:
When https://github.com/apache/netbeans/pull/4586 was integrated, the `VanillaCompileWorkerTest` started to fail, as error recovery no longer work properly for pattern matching and JDK 19 javac, when running on JDK <19, as `j.l.MatchException` is missing. The `MatchException` is needed by the generated code, and hence javac cannot generate the code if it is missing. This patch proposes to improve the error recovery for patterns when the class is missing - the code using the patterns would be removed, as would be done if the code would be referring to `MatchException` explicitly.

---

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorker.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/indexing/VanillaCompileWorker.java
@@ -19,9 +19,11 @@
 
 package org.netbeans.modules.java.source.indexing;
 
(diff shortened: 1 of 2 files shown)


mbien:
thanks for fixing this @jlahoda. All java tests are green, lets merge this soon so that master is stable again.

mbien (approved):
did some manual testing and everything worked -> merging

Issue #4755, opened 2022-10-08
Unable to invoke popup menu for source classes in Gradle project


makiam:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

Once I select multiple java files under project node I can invoke popup menu over it (Ex, to open multipe files or call Diff tool). This works now for Ant-based and/or for Maven projects but not work with Gradle project

### How to reproduce

Under Gradle projject select multiple java classes in project and try to invoke popum menu over it

### Did this work correctly in an earlier version?

Apache NetBeans 14

### Operating System

Windows 10

### JDK

JDK 18.0.2

### Apache NetBeans packaging

Own source build

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

makiam:
Update description

PR #4759, opened 2022-10-09
Fix for Gradle project sources popup menu invocation error #4755

makiam:
Check that EditorCookie document may be null and avoid to pass null object to NbEditorUtilities.getFileObject


lkishalmi (approved):
Seems reasonable.

I've checked the commit it has a full name on it, however I could not validate the email address, as it shows up nowhere in google.

@matthiasblaesing the code is trivial, shall we accept this one or just apply the patch in our code?

matthiasblaesing:
The email address is consistent with other repositories, so should be good. The change looks also sane. Lets get this in.

PR #4770, opened 2022-10-11
bump versions of setup-java, setup-php and cache actions.

mbien:
github showed several warnings: "Node.js 12 actions are deprecated..." and recommended to bump versions.


mbien:
forgot a `v`. But waiting with the push until higher priority PRs are done with the build.

neilcsmith-net (approved):

mbien:
another set of warnings appeared. This time from the cache action.
https://github.com/actions/cache/issues/953

but this might fix itself once the cache action releases a minor point update, so we probably won't have to bump its version.

PR #4773, opened 2022-10-11
Improve support for ANTLRv4 Grammars

lkishalmi:
This one adds support for occurrence finder for modes and channels.

The code completion has been separated for ANTLVv3 and ANTLRv4. I've added some heuristics to make it better for v4, based on the real ANTLRv4 Lexer. Though there is still room for improvement.

I'd like to finish mu indentation support before NB16, but I thought this is complete enough for review.

### DIFF
--- a/java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
+++ b/java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
@@ -19,7 +19,6 @@
 package org.netbeans.modules.languages.antlr;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -121,16 +120,11 @@ public abstract class AntlrParserResult<T extends Parser> extends ParserResult {
         if (occurrences.containsKey(refName)) {
             ret.addAll(occurrences.get(refName));
         }
-        return !ret.isEmpty() ? ret : Collections.emptyList();
+        return ret;
     } 
     
     protected final void markOccurrence(String refName, OffsetRange or) {
-        List<OffsetRange> ol = occurrences.get(refName);
-        if (ol == null) {
-            ol = new ArrayList<>();
-            occurrences.put(refName, ol);
-        }
-        ol.add(or);
+        occurrences.computeIfAbsent(refName, s -> new ArrayList<>()).add(or);
     }
 
     protected abstract T createParser(Snapshot snapshot);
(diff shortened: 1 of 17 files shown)


java/languages.antlr/src/org/netbeans/modules/languages/antlr/refactoring/Refactoring.java
```diff
@@ -148,7 +148,7 @@ public Problem prepare(RefactoringElementsBag refactoringElements) {
                         if(ref.defOffset != null) {
                             ranges.add(ref.defOffset);
                         }
-                        ranges.addAll(ref.occurances);
+                        ranges.addAll(result.getOccurrences(name));
```
matthiasblaesing, line 151: This needs to be pulled out of the if. I tested this:

- I opened `LexBasic.g4` from the `antlr4` package
- I ran "Find Usage" in `Ws`

I only got the results from `LexBasic.g4` and was missing the occurrences from `ANTLR4Lexer.g4`. `ref` is `NULL` for the latter, but there are still occurrences. Pulling out this call fixes it.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrStructureScanner.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.netbeans.modules.languages.antlr;
 
+import java.util.ArrayList;
```
matthiasblaesing, line 21: Looks like an unnecessary change.

matthiasblaesing (commented):
I left a few inline comments, but I also found this:

image

My understanding is, that the "modes" group rules in antlr4. This seems to work for the "Id" fragment, but not for the also defined lexer rules. I assume, that the contents of the two marked areas should match.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/AntlrParserResult.java
```diff
@@ -93,20 +101,37 @@ protected boolean processingFinished() {
...
+        return !ret.isEmpty() ? ret : Collections.emptyList();
+    } 
+    
+    protected final void markOccurrence(String refName, OffsetRange or) {
+        List<OffsetRange> ol = occurrences.get(refName);
+        if (ol == null) {
+            ol = new ArrayList<>();
+            occurrences.put(refName, ol);
+        }
+        ol.add(or);
```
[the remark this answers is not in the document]
lkishalmi: Thanks! I've learned something new today!

lkishalmi:
Also fixed a potential NPE on refactoring. It'd happen when the find usages would be repeated from the UI button.

matthiasblaesing (approved):
Looks sane to me.

PR #4775, opened 2022-10-11
Limit ValidateClassFilesTest to scan only JARs referenced from modules

matthiasblaesing:
With the addition of jakarta ee 10 API jars it was found, that there is an implicit assumption, that all JARs not referenced by modules have to be JDK 8 compatible. In the case of jakarta ee 10 this is an invalid assumption as it explicitly does not support JDK 8. The module that introduces the binaries correctly declares a JDK 11 dependeny, but at runtime the module <-> binary association is lost.

To prevent false positives, the test must be limited to JARs, that are directly referenced from modules.

### DIFF
--- a/ide/extexecution.process.jdk9/manifest.mf
+++ b/ide/extexecution.process.jdk9/manifest.mf
@@ -3,5 +3,5 @@ AutoUpdate-Show-In-Client: false
 OpenIDE-Module: org.netbeans.modules.extexecution.process.jdk9
 OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/extexecution/process/jdk9/resources/Bundle.properties
 OpenIDE-Module-Specification-Version: 1.18
-OpenIDE-Module-Java-Dependencies: Java > 1.9.0
+OpenIDE-Module-Java-Dependencies: Java > 9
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
@neilcsmith-net  @jtulach you two worked previously on this test. This change is triggered by the commit validation errors reported in #4692 (Commit [da98cbd](https://github.com/apache/netbeans/pull/4692/commits/da98cbdaedfc0a80730e0d11a370dc2c11a8cc9e)). From my observation the problem is, that the test assumes, that all JARs, that may need a java version higher than 8 are referenced by modules (either as module itself or as dependency). This is not true for bundled jar files. In the referenced PR the bundled library is Jakarta EE 10, which has a minimum baseline of JDK 11.

neilcsmith-net (approved):
I think this is OK as a quick fix.  I acknowledge that my addition was also a quick fix!  Ideally we should probably add bundled JARs along with the result of `module.getAllJars()` but I wasn't sure of the best way to access this information in the test (and didn't need it to get it to pass then!)

mbien:
restarting tests since master is green again.

jGauravGupta (approved):

Issue #4628, opened 2022-09-12
Module-info.java already exists.


JavacoderBrian:
### Apache NetBeans version

Apache NetBeans 15

### What happened

In the process of creating a module-info.java in an existing project, the wizard stops at the end saying, The file module-info.java already exists.
![image](https://user-images.githubusercontent.com/2698503/189648619-a1122cbd-1c55-4c68-a1b2-e944301cf9d2.png)

### Did this work correctly in an earlier version?

Apache NetBeans 12.6

### Operating System

Windows 11 Pro Version 10.0.22000 Build 22000

### JDK

OpenJDK 18.0.1

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

Everytime

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

neilcsmith-net:
Well, that's annoying!  Can replicate.  The file does at least seem to be created correctly, even if you have to cancel the wizard to get to it.

neilcsmith-net:
Stack trace of the exception thrown -

```
WARNING [org.openide.WizardDescriptor]
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at com.sun.tools.javac.util.List.get(List.java:490)
	at org.netbeans.modules.java.project.ui.NewJavaFileWizardIterator$AddRequiresDirective.run(NewJavaFileWizardIterator.java:265)
	at org.netbeans.modules.java.project.ui.NewJavaFileWizardIterator$AddRequiresDirective.run(NewJavaFileWizardIterator.java:254)
	at org.netbeans.api.java.source.JavaSource$1.run(JavaSource.java:675)
	at org.netbeans.api.java.source.JavaSource$1.run(JavaSource.java:665)
	at org.netbeans.api.java.source.JavaSource$MultiTask.run(JavaSource.java:504)
	at org.netbeans.modules.parsing.impl.TaskProcessor.callUserTask(TaskProcessor.java:586)
	at org.netbeans.modules.parsing.api.ParserManager$UserTaskAction.run(ParserManager.java:132)
	at org.netbeans.modules.parsing.api.ParserManager$UserTaskAction.run(ParserManager.java:116)
	at org.netbeans.modules.parsing.impl.TaskProcessor$2.call(TaskProcessor.java:181)
	at org.netbeans.modules.parsing.impl.TaskProcessor$2.call(TaskProcessor.java:178)
[...]

PR #4780, opened 2022-10-12
Fix error in wizard when creating module-info file (GH4628)

neilcsmith-net:
Fixes a spurious error message that the file already exists shown when creating a module-info file, as reported in #4628 This is actually an `IndexOutOfBoundsException` caused by [javac changes](https://github.com/openjdk/jdk/commit/d6fb9d7256de94c3cc407b78dde3088aaf649db7) (`ModuleTree` no longer returned from `getTypeDecls()` but available from added `getModule()`).

The second commit filters out the created module name from the list of added requires. This is an older issue that became visible again that shows a cyclic dependency error.


mbien (approved):
makes sense. Nice find with the impl change in the JDK.

sdedic (approved):

PR #4788, opened 2022-10-14
[NETBEANS-4763] Handling switch expressions, case rules and exhaustive switches more correctly.

jlahoda:
---

### DIFF
--- a/java/java.hints/src/org/netbeans/modules/java/hints/bugs/InfiniteRecursion.java
+++ b/java/java.hints/src/org/netbeans/modules/java/hints/bugs/InfiniteRecursion.java
@@ -21,12 +21,14 @@ package org.netbeans.modules.java.hints.bugs;
 import com.sun.source.tree.BinaryTree;
 import com.sun.source.tree.BreakTree;
 import com.sun.source.tree.CaseTree;
+import com.sun.source.tree.CaseTree.CaseKind;
 import com.sun.source.tree.ClassTree;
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
@sdedic @jtulach would be good if either/both of you could take a look, confirm the fix and merge before freeze.

sdedic (approved):

mbien:
this looks ready to merge to me, tests are green too. @jlahoda @matthiasblaesing should we get this in?

matthiasblaesing:
@mbien I saw this yesterday and I agree. I ran test tonight and locally I saw failures, but these were timeouts and language depedneded problems. The corresponding tests were ran on the github actions infrastructure and there they worked. So yes this should go in.

mbien:
going to risk it and merge. `java.hints` batch1 and batch2 ran on CI and this particular test does also run locally for me (started with IDE).

jlahoda:
Thanks for merging - I don't see a reason why not!

mbien:
> Thanks for merging - I don't see a reason why not!

We are always a bit careful with reviewed PRs which didn't have their tests run for a long time. It happened before that it broke master which isn't a huge problem since it can be fixed again but it still is annoying so we try to avoid it if possible.

The problem here is that travis is now gone but was still there when this PR ran, so to run the tests again someone would have to push into this PR and rebase. Since this is a small PR and is easy to check I think its fine to simply merge it without the extra steps (master is green so everything is looking good so far :)).

Issue #4760, opened 2022-10-09
Close whole documents list at once


mbien:
_Originally posted by @6PATyCb in https://github.com/apache/netbeans/issues/4710#issuecomment-1272557189_

When too much files is opened I like to use `documents list` format for manage it:
![image](https://user-images.githubusercontent.com/33724393/194762879-ea2250c6-222e-429d-8f59-c1b70390c588.png)

but in that format would be good to have close button to group of documents:
![image](https://user-images.githubusercontent.com/33724393/194763097-3fa38dee-932c-4e11-a956-de12cda764b7.png)

Richard7111:
Hello, can you assign me this issue and help me navigate a little in the project?

Richard7111:
Hello @mbien, can you assign me this issue and help me navigate a little in the project?

mbien:
hi @Richard7111

sure. Take a look at `platform/core.multitabs`.

you could set a breakpoint here:
https://github.com/apache/netbeans/blob/1a9b29d627a0f343837dc053b56e8de32d6ff689/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java#L445-L456

Don't try to close editor tabs with unsaved changes since this can cause issues.

how to build NetBeans and work with modules was recently answered here.
https://github.com/apache/netbeans/discussions/4664#discussioncomment-3756079

Richard7111:
Hello, I am not able to find how to create documents list where should I look for this feature. Google was no help. Could you please help me?

Richard7111:
Hello @6PATyCb, I am not able to find how to create documents list, where should I look for this feature. Google was no help. Could you please help me?

troizet:
Choose menu Tools -> Options -> Appearance -> check "Sort opened documents list by project".
Then open several projects.

6PATyCb:
> Hello @6PATyCb, I am not able to find how to create documents list in the IDE, where should I look for this feature. Google was no help. Could you please help me?
> 
> EDIT: Managed to find it.

Hi! Take a look at @troizet comment

Issue #4710, opened 2022-10-01
Add a "Close Right" in menu


fredericch:
### Description

Add a "Close Right" function on the menu along "Close", "Close All" and "Close Other".

When at least 2 files have been opened, we could close all windows on the right of the current active one.

![image](https://user-images.githubusercontent.com/11049456/193430514-3cc97024-2c50-4000-a2ee-6c99fad7b78b.png)

It would be the same implementation as in Chrome or FF.

### Use case/motivation

When doing a search, I often open "new" windows which are actually not currently usefull and it clutters the tabs. So I spend a lot of time closing them individually. It would save a lot of time to be able to close them all at once.

I cannot use "Close Other" or "Close All", because the tabs on the left are usually the ones I am working on.

When a lot of windows have been opened if the search was too generic, I use this work around:
- Close All
- Open all modified files from the Git menu

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

mbien:
makes sense. Added [good first issue](https://github.com/apache/netbeans/labels/good%20first%20issue) label in case someone wants to implement it.

I just checked how FF solved this and it has `Close Tabs to Left`, `.. Right` and `Close Other Tabs` in a sub menu.

NetBeans however also supports multi row tabs which would make this a bit more difficult to implement.

6PATyCb:
When too much files is opened I like to use `documents list` format for manage it:
![image](https://user-images.githubusercontent.com/33724393/194762879-ea2250c6-222e-429d-8f59-c1b70390c588.png)

but in that format would be good to have close button to group of documents:
![image](https://user-images.githubusercontent.com/33724393/194763097-3fa38dee-932c-4e11-a956-de12cda764b7.png)

mbien:
@6PATyCb thats a good idea but next time, please open a new issue. I moved it to 4760

6PATyCb:
> @6PATyCb thats a good idea but next time, please open a new issue. I moved it to #4760

I thought this was just part of the discussion about usability of documents tabs, but since it looks good as a separate feature, why not :)

tc430gi:
I would like to implement it for you like my good first issue, need it for school credit at University.

tc430gi:
@mbien can you please?

6PATyCb:
> Hello can @fredericch, @6PATyCb, or anyone assign me this issue please?

I don't have a rights

mbien:
[...]

PR #4792, opened 2022-10-15
Close whole documents list at once

Richard7111:
Implemented closing of a whole document list at once, based on issue [#4760](https://github.com/apache/netbeans/issues/4760)
image

### DIFF
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
@@ -205,8 +205,7 @@ final class ButtonPopupSwitcher implements MouseInputListener, AWTEventListener,
     
     @Override
     public void mousePressed(MouseEvent e) {
-        int tabCount = controller.getTabModel().size();
-        if( pTable.onMouseEvent(e) && tabCount == 1 ) {
+        if( pTable.onMouseEvent(e) ) {
             hideCurrentPopup();
         }
         e.consume();
@@ -230,8 +229,8 @@ final class ButtonPopupSwitcher implements MouseInputListener, AWTEventListener,
         p = SwingUtilities.convertPoint((Component) e.getSource(), p, pTable);
         if (pTable.contains(p)) {
             if( !pTable.onMouseEvent(e) ) {
-                final SwitcherTableItem item = pTable.getSelectedItem();
-                if (item != null) {
+                final DocumentSwitcherTable.Item item = pTable.getSelectedItem();
+                if (item != null && !item.isSeparator()) {
                     hideCurrentPopup();
                     item.activate();
                 }
@@ -343,7 +342,7 @@ final class ButtonPopupSwitcher implements MouseInputListener, AWTEventListener,
                 break;
             case KeyEvent.VK_DELETE: {
                 final Item item = ( Item ) pTable.getSelectedItem();
-                if (item != null && TabDataRenderer.isClosable( item.getTabData() )) {
+                if (item != null) {
                     TabData tab = item.getTabData();
                     int tabIndex = controller.getTabModel().indexOf( tab );
                     if( tabIndex >= 0 ) {
@@ -355,6 +354,10 @@ final class ButtonPopupSwitcher implements MouseInputListener, AWTEventListener,
                         selRow = Math.min( pTable.getModel().getRowCount()-1, selRow );
                         selCol = Math.min( pTable.getModel().getColumnCount()-1, selCol );
                         switched = true;
+                    } else if ( item.isSeparator() ){
+                        if( pTable.closeSelectedDocumentList() ) {
+                            SwingUtilities.invokeLater(this::hideCurrentPopup);
+                        }
                     }
                 }
                 break;
--- a/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
+++ b/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
@@ -27,6 +27,8 @@ import java.awt.Insets;
 import java.awt.Point;
 import java.awt.Rectangle;
 import java.awt.event.MouseEvent;
+import java.util.ArrayList;
+import java.util.List;
 import javax.swing.BorderFactory;
 import javax.swing.DefaultButtonModel;
 import javax.swing.Icon;
@@ -77,10 +79,8 @@ class DocumentSwitcherTable extends SwitcherTable {
     }
 
     @Override
-    public SwitcherTableItem getSelectedItem() {
+    public Item getSelectedItem() {
         Item res = ( Item ) getValueAt(getSelectedRow(), getSelectedColumn());
-        if( null != res && res.isSeparator )
-            return null; //don't hide the popup when a project separator is selected
         return res;
     }
 
@@ -90,7 +90,7 @@ class DocumentSwitcherTable extends SwitcherTable {
 
         boolean selected = row == getSelectedRow() &&
                 column == getSelectedColumn() && item != null;
-        boolean separator = null != item && item.isSeparator;
+        boolean separator = null != item && item.isSeparator();
 
         ITEM_BORDER.color = null;
         Component renComponent = super.prepareRenderer( renderer, row, column );
@@ -105,10 +105,10 @@ class DocumentSwitcherTable extends SwitcherTable {
                 lbl.setBorder( ITEM_BORDER );
             }
         }
-        if( selected && !separator ) {
+        if( selected ) {
             JPanel res = new JPanel( new BorderLayout(5, 0) );
             res.add( renComponent, BorderLayout.CENTER );
-            if( TabDataRenderer.isClosable( item.getTabData() ) ) {
+            if( separator || TabDataRenderer.isClosable( item.getTabData() )) {
                 res.add( btnClose, BorderLayout.EAST );
             }
             res.setBackground( renComponent.getBackground() );
@@ -127,15 +127,20 @@ class DocumentSwitcherTable extends SwitcherTable {
     private int lastCol = -1;
     private boolean inCloseButtonRect = false;
 
+    /**
+     * 
+     * @return True if popup window should be closed. (e.g. when no more 
+     * tabs are left opened)
+     */
     boolean onMouseEvent( MouseEvent e ) {
         Point p = e.getPoint();
         p = SwingUtilities.convertPoint((Component) e.getSource(), p, this);
         int selRow = getSelectedRow();
         int selCol = getSelectedColumn();
-        if( selRow < 0 || selCol < 0 )
+        if ( selRow < 0 || selCol < 0 )
             return false;
         Rectangle rect = getCellRect( selRow, selCol, false );
-        if( rect.contains( p ) ) {
+        if ( rect.contains( p ) ) {
             Dimension size = btnClose.getPreferredSize();
             int x = rect.x+rect.width-size.width;
             int y = rect.y + (rect.height-size.height)/2;
@@ -143,15 +148,18 @@ class DocumentSwitcherTable extends SwitcherTable {
             boolean inButton = btnRect.contains( p );
             boolean mustRepaint = inCloseButtonRect != inButton;
             inCloseButtonRect = inButton;
-            if( inButton ) {
-                if( e.getID() == MouseEvent.MOUSE_PRESSED ) {
+            if ( inButton ) {
+                if ( e.getID() == MouseEvent.MOUSE_PRESSED ) {
                     Item item = ( Item ) getModel().getValueAt( selRow, selCol );
                     TabData tab = item.getTabData();
+                    int tabSize = controller.getTabModel().size();
                     int tabIndex = controller.getTabModel().indexOf( tab );
                     if( tabIndex >= 0 ) {
                         TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
                         controller.postActionEvent( tae );
-                        return true;
+                        return tabSize == 1;
+                    } else if ( item.isSeparator() ) {
+                        return closeSelectedDocumentList();
                     }
                 }
             }
@@ -174,7 +182,30 @@ class DocumentSwitcherTable extends SwitcherTable {
         }
         return null;
     }
-
+    /**
+     * Method closes tabs in selected document list
+     * 
+     * @return True if there are no more tabs left opened in other projects
+     */
+    boolean closeSelectedDocumentList() {
+        List<TabData> tabs = new ArrayList<>(controller.getTabModel().getTabs());
+        Item item = ( Item ) getModel().getValueAt( getSelectedRow(), getSelectedColumn());
+        ProjectProxy project = item.getProject();
+        ProjectSupport projectSupport = ProjectSupport.getDefault();
+        int numOfOtherTabs = 0;
+        for ( TabData tab : tabs ) {  
+            ProjectProxy projectForTab = projectSupport.getProjectForTab( tab );
+            if (( project == null && projectForTab == null ) || ( projectForTab != null && projectForTab.equals( project ))) {
+                int tabIndex = controller.getTabModel().indexOf( tab );
+                TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
+                controller.postActionEvent( tae );
+            } else {
+                numOfOtherTabs++;
+            }
+        }
+        return numOfOtherTabs == 0;
+    }
+    
     private JButton createCloseButton() {
         JButton res = CloseButtonFactory.createBigCloseButton();
         res.setModel( new DefaultButtonModel() {
@@ -222,6 +253,10 @@ class DocumentSwitcherTable extends SwitcherTable {
 
         public TabData getTabData() {
             return tabData;
+        }
+        
+        public boolean isSeparator() {
+            return isSeparator;
         }
 
         @Override


platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -152,7 +152,21 @@ boolean onMouseEvent( MouseEvent e ) {
                         TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
                         controller.postActionEvent( tae );
                         return true;
+                    }else if(tabIndex == -1){
+                        List<TabData> tabs = controller.getTabModel().getTabs();
+                        ProjectProxy project = item.project;
+                        ProjectSupport projectSupport = ProjectSupport.getDefault();
+                        for(int i = tabs.size(); i-- > 0;){
+                            TabData projectTab = tabs.get(i);
+                            if(projectSupport.getProjectForTab(projectTab) == (project) || projectSupport.getProjectForTab(projectTab).equals(project)){
```
mbien, line 161: the equals check alone would be sufficient. but to be sure, lets add a null check too. We can also use the opportunity to store the result in a variable so that we call the method only once (it could be expensive if the result is not cached).
```java
  ProjectProxy projectForTab = projectSupport.getProjectForTab(projectTab);
  if (projectForTab != null && projectForTab.equals(project)) {
```

mbien (commented):
hi @Richard7111, thanks for implementing this. Ran a quick test and it worked great!

It would be good if new code would use the default NetBeans formatting, so that we don't mix so many styles in one file (no need to format the rest, only for new code). Example:
```java
    // space before braces
    private void method() {
      if (this.foo != null) {
          this.foo.method();
      }
    }
```

Please try to extract the new code which closes the tabs into a package private utility method and put it into `DocumentSwitcherTable` . Since `ButtonPupupSwitcher` is in the same package as `DocumentSwitcherTable` and can simply call its method without code duplication.

Some more comments are inline.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -152,7 +152,21 @@ boolean onMouseEvent( MouseEvent e ) {
                         TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
                         controller.postActionEvent( tae );
                         return true;
+                    }else if(tabIndex == -1){
+                        List<TabData> tabs = controller.getTabModel().getTabs();
+                        ProjectProxy project = item.project;
+                        ProjectSupport projectSupport = ProjectSupport.getDefault();
+                        for(int i = tabs.size(); i-- > 0;){
+                            TabData projectTab = tabs.get(i);
```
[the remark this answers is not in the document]
Richard7111: Yeah, when closing tab one by one, the tabIndex changes which resulted in closing only the first tab under document list. Managed to solve that by closing tab from the end

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
Richard7111: The reason I am not checking for null, is because there is a case, when no project is associated when file and variables projectForTab and project are null. 

Should I add another `if`  for that case?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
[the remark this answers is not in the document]
mbien: Makes sense. Would be good to leave a comment there so that nobody changes it by accident.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
mbien: good thinking. How about:
```java
  ProjectProxy projectForTab = projectSupport.getProjectForTab(projectTab);
  if ((project == null && projectForTab == null) || (projectForTab != null && projectForTab.equals(project))) {
```

Richard7111: That should work, gonna try it. Is it okay to have such a long condition?

mbien: sure. Its not a problem as long its somewhat readable. Your version was fine too, but without knowing about the special case where both are null, it almost looks like a bug. Making it explicit answers possible questions right away ;)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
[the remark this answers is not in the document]
Richard7111: > Please try to extract the new code which closes the tabs into a package private utility method and put it into DocumentSwitcherTable . Since ButtonPupupSwitcher is in the same package as DocumentSwitcherTable and can simply call its method without code duplication.

Somehow I pushed older version of the code, difference should be visible in the next commit. When closing tabs using DELETE key, I have to call SwingUtilities if all the tabs are closed. Otherwise the popup stays open. That is why I am keeping it outside of shared method.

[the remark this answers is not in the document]
mbien: if you would move this after the loop, it should still work and we could extract the whole section into a method in `DocumentSwitcherTable`? Duplicated code is really something we should try to avoid.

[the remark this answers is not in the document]
Richard7111: Oh yeah, you are right.

Not sure if I have done it correctly, is this the usage in ButtonPopupSwitcher?
```
DocumentSwitcherTable dst = new DocumentSwitcherTable( controller, createSwitcherItems(controller), y );
dst.closeDocumentList( item, controller );
if( controller.getTabModel().size() == 1 ) {
    SwingUtilities.invokeLater(this::hideCurrentPopup);
}
```

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -132,27 +132,42 @@ boolean onMouseEvent( MouseEvent e ) {
...
+                        ProjectSupport projectSupport = ProjectSupport.getDefault();
+                        for ( int i = tabs.size(); i-- > 0; ) {  //reverse iteration to close tabs from the end so tabIndex does not change
+                            TabData projectTab = tabs.get( i );
+                            ProjectProxy projectForTab = projectSupport.getProjectForTab( projectTab );
+                            if (( project == null && projectForTab == null ) || ( projectForTab != null && projectForTab.equals( project ))) {
+                                tabIndex = controller.getTabModel().indexOf( projectTab );
+                                TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
+                                controller.postActionEvent( tae );
+                            }
+                        }
```
mbien, line 167: step 1, extract this into a package-private method in the same class here.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
```diff
@@ -355,6 +354,22 @@ private boolean changeSelection( KeyEvent event ) {
...
+                                SwingUtilities.invokeLater(this::hideCurrentPopup);
+                            }
+                            TabData projectTab = tabs.get( i );
+                            ProjectProxy projectForTab = projectSupport.getProjectForTab( projectTab );
+                            if (( project == null && projectForTab == null ) || ( projectForTab != null && projectForTab.equals( project ))) {
+                                tabIndex = controller.getTabModel().indexOf( projectTab );
+                                TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
+                                controller.postActionEvent( tae );
+                            }
+                        }
```
mbien, line 372: step 2, remove this. and call the method:
```
pTable.closeTabsOfProject(project);
```
I think project might be the only param you need.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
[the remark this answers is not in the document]
mbien: i commented inline below. closing this thread.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
Richard7111: Ohh, I over looked the pTable variable. thank you. Will commit asap.

mbien: awesome

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -174,7 +178,23 @@ public String getToolTipText( MouseEvent event ) {
         }
         return null;
     }
-
+    
+    void closeDocumentList( Controller controller ) {
```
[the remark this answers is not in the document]
Richard7111: You are right. Removed parameter, but forgot to rename method. Will commit asap.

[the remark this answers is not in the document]
mbien: great. Can you also as last step squash all of your commits into one and force push into this PR? Let me know if you need instructions.

Richard7111:
Hey @mbien, I would like to write tests for my changes as a part of school assignment. However I have only used Mockito before for mocks, are you using something else for mocks or am I missing something?

mbien:
@Richard7111 Tools like Mockito didn't exist when most of NetBeans was written, this lead to custom tooling around tests. For example most tests extend NbTestCase which can do a lot of NB specific things.

You could try to ask on the dev mailing list and see if someone has a good idea how to test things like this. However, UI tests have the tendency to break with time (sometimes randomly) and take long setup time to execute. I can't promise that we would add tests like that since we are reaching some limits. Mocking UI code is also not trivial and often not very useful.

As you can see on a full test run
https://github.com/apache/netbeans/actions/runs/3280529976/usage
 (scroll down for the total time spent), thats a lot of time spent testing.
 
Btw since we are talking about testing. I think I found a bug, I am not quite sure how to reproduce it reliably. The popup closes sometimes if a list is closed. It often happens if the first list is closed and possibly if the last list has only one item in it - had no time to investigate further.
Another thing: I believe it would be better if the popup wouldn't close if a project-name item is clicked without clicking the x just like it was before. There is no action associated with it, so I believe its better if it does nothing on click.

But there is no need to hurry to fix this since there is quite some time till the NB17 release.

Richard7111:
@mbien hey, I just tried to replicate the bug with closing popup. It happens only when there are 2 document lists, you close first and the second one contains only one item. 

Tried to fix it, but can't seem to figure out the correct condition. Will work on it a little more.

Richard7111:
@mbien managed to fix the bugs, feel free to check it out. Hope everything is fixed now. 😄

mbien (approved):
excellent. The most recent update looks good to me too.

Did some manual testing with and without project list sorting and everything worked as expected.

going to ping some more reviewers if they are interested to take a look before merging.

mbien:
@Richard7111 I just saw that your email on the commit is now `@users.noreply.github.com`. Please fix this first to a valid address.

eirikbakke:
Thanks for this patch!

Another thing to test: How does this work when there are tabs with unsaved changes? What if there are multiple tabs with unsaved changes?

Will take a look at the code in a bit...

mbien:
> Another thing to test: How does this work when there are tabs with unsaved changes? What if there are multiple tabs with unsaved changes?

basically it uses the same code to close tabs as it is using now. A popup will ask the user to: discard, save or cancel

eirikbakke:
There are some other existing bugs in ButtonPopupSwitcher to be aware of, too:
https://issues.apache.org/jira/browse/NETBEANS-5855
https://issues.apache.org/jira/browse/NETBEANS-5853

Will review...

eirikbakke:
If multiple tabs have unsaved changes, and the user presses Cancel in the dialog box that prompts to save, what happens in this case? Probably the loop should stop in this case.

In fact, this should probably be done in the Close Other and Close All actions in the tab right-click menu as well. (CloseAllButThisAction and CloseAllDocumentsAction, respectively) But that's a separate bug.

mbien:
> If multiple tabs have unsaved changes, and the user presses Cancel in the dialog box that prompts to save, what happens in this case? Probably the loop should stop in this case.

it just going to ask you the same question for each unedited document. The dialog has the name of the document in it.

eirikbakke:
@mbien Right, but if the user presses Cancel (rather than No), the loop should stop.

Also, do the dialog boxes come one after another, or all on top of each other? They should probably come one at a time.

mbien:
> @mbien Right, but if the user presses Cancel (rather than No), the loop should stop.

Not sure about this since the question is only about the concrete file. The action could simply continue with the rest.

> Also, do the dialog boxes come one after another, or all on top of each other? They should probably come one at a time.

please test it. Its a UI change which is easier to discuss if everyone played around with it. But yes they don't open all at once.

eirikbakke:
> Not sure about this since the question is only about the concrete file. The action could simply continue with the rest.

Well, that's the difference between pressing "Cancel" and "No"; the "No" button refers to the file in question, while Cancel refers to the overall action.

I'll need to do a custom build to test this... will do but it may take a couple of days.

mbien:
> Well, that's the difference between pressing "Cancel" and "No"; the "No" button refers to the file in question, while Cancel refers to the overall action.

It would be less intuitive to me if a file save dialog which asks me whether or not to save a single file would cancel more than what was asked in the question if cancel is pressed (it would need a "cancel all" button or something similar). The scope of the dialog seems to be clear to me: it is a save operation of a single file. The document group action is already running at this point. Some tabs might have been already closed.

To do this differently, one would have to scan the files first, ask the question with the multi-file-save dialog, then scan again and close the group if everything was saved and user didn't press cancel.

eirikbakke:
@mbien That's standard desktop app behavior for "Cancel" in a multi-document-interface environment:  "Cancel" cancels the whole operation, "No" (actually "Discard" in NetBeans) closes the current file without saving, while continuing the loop. That's why there's a Cancel button that's separate from Discard.

> To do this differently, one would have to scan the files first, ask the question with the multi-file-save dialog, then scan again and close the group if everything was saved and user didn't press cancel.

No, that's not necessary; you just cancel when the user presses Cancel. If they already chose to save or discard some changes in other tabs, that's fine.

I just tested with VSCode and Photoshop, for instance; it works that way there. (Not applicable in IntelliJ, since it autosaves and never prompts to save the file.)

eirikbakke:
But we can leave the Discard/Cancel distinction out of this PR... I can do a separate PR later where I also propose to adjust the similar behavior for "Close All" and "Close Other".

Richard7111:
> @Richard7111 I just saw that your email on the commit is now `@users.noreply.github.com`. Please fix this first to a valid address.

@mbien sorry, I am not sure where you see that. I see correct mail and user name when I check `git log`. Would you mind helping me?
image

neilcsmith-net:
IMO it would be good to align the dialog with the single one that shows when you try to close the IDE with modified files.

mbien:
> No, that's not necessary; you just cancel when the user presses Cancel. If they already chose to save or discard some changes in other tabs, that's fine.

@eirikbakke IMO: if the situation allows it to run an atomic operation, the software should do that. Run checks first, ask if there is something to resolve, then do the operation. Don't start the op and keep asking questions while its running.

> But we can leave the Discard/Cancel distinction out of this PR... I can do a separate PR later where I also propose to adjust the similar behavior for "Close All" and "Close Other".

awesome. Would be good to have consistent behavior in comparable actions.

eirikbakke:
@neilcsmith-net I'll look into that in a future PR.

mbien:
> @mbien sorry, I am not sure where you see that. I see correct mail and user name when I check `git log`. Would you mind helping me a bit more?

@Richard7111 you are right! everything is fine. I looked at the wrong line:
https://github.com/apache/netbeans/actions/runs/3392742442/jobs/5645247607#step:4:16
It was the automatic merge commit, not the actual commit. All good, my mistake.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -27,6 +27,8 @@
 import java.awt.Point;
 import java.awt.Rectangle;
 import java.awt.event.MouseEvent;
+import java.util.Collections;
```
eirikbakke, line 30: Unused import.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -283,4 +310,4 @@ public boolean isBorderOpaque() {
             return false;
         }
     }
-}
+}
```
eirikbakke, line 313: Unnecessary whitespace change.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
```diff
@@ -374,7 +377,7 @@ private boolean changeSelection( KeyEvent event ) {
         }
         return switched;
     }
-
+
```
eirikbakke, line 380: Unnecessary whitespace change.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
```diff
@@ -508,7 +511,7 @@ private Item[] createSwitcherItems(final Controller controller) {
         }
         return items.toArray( new Item[0] );
     }
-
+
```
eirikbakke, line 514: Unnecessary whitespace change.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
```diff
@@ -549,4 +552,4 @@ private void selectTab(TabData tab) {
             }
         }
     }
-}
+}
```
eirikbakke, line 555: Unnecessary whitespace change.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
```diff
@@ -355,6 +354,10 @@ private boolean changeSelection( KeyEvent event ) {
                         selRow = Math.min( pTable.getModel().getRowCount()-1, selRow );
                         selCol = Math.min( pTable.getModel().getColumnCount()-1, selCol );
                         switched = true;
+                    }else if( tabIndex == -1 ){
```
eirikbakke, line 357: Should have space before "else" and after "if". Some similar formatting errors elsewhere. (Don't worry about fixing formatting issues in existing code, though.)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -132,26 +132,29 @@ boolean onMouseEvent( MouseEvent e ) {
...
+                if ( e.getID() == MouseEvent.MOUSE_PRESSED ) {
                     Item item = ( Item ) getModel().getValueAt( selRow, selCol );
                     TabData tab = item.getTabData();
+                    int tabSize = controller.getTabModel().size();
                     int tabIndex = controller.getTabModel().indexOf( tab );
                     if( tabIndex >= 0 ) {
                         TabActionEvent tae = new TabActionEvent( this, TabbedContainer.COMMAND_CLOSE, tabIndex );
                         controller.postActionEvent( tae );
-                        return true;
+                        return tabSize == 1;
```
eirikbakke, line 155: What is the purpose of this condition? Under what conditions are onMouseEvent supposed to return true vs. false? (It would be good to add a Javadoc comment explaining the latter.)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -174,7 +177,27 @@ public String getToolTipText( MouseEvent event ) {
         }
         return null;
     }
-
+    
+    boolean closeSelectedDocumentList() {
```
eirikbakke, line 181: Would be good to add Javadoc explaining what the return value is supposed to mean here.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -174,7 +177,27 @@ public String getToolTipText( MouseEvent event ) {
...
+        List<TabData> tabs = controller.getTabModel().getTabs();
+        Item item = ( Item ) getModel().getValueAt( getSelectedRow(), getSelectedColumn());
+        ProjectProxy project = item.getProject();
+        ProjectSupport projectSupport = ProjectSupport.getDefault();
+        int numOfOtherTabs = 0;
+        for ( int i = tabs.size(); i-- > 0; ) {  //reverse iteration to close tabs from the end so tabIndex does not change
```
eirikbakke, line 187: Probably better to assemble tabs to be closed in a list first, then iterate over them normally. There may be "Save Changes?" dialogs popping up as a result of trying to close the tabs, and they normally come left-to-right.

eirikbakke: Also, there's an off-by-one error here. Reverse for loops over zero-based indices should go "for (int i = N - 1; i >= 0; i--)". (But better iterate left-to-right over a copy of the relevant tabs as suggested above.)

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
```diff
@@ -174,7 +177,27 @@ public String getToolTipText( MouseEvent event ) {
...
+            TabData projectTab = tabs.get( i );
+            ProjectProxy projectForTab = projectSupport.getProjectForTab( projectTab );
+            if (( project == null && projectForTab == null ) || ( projectForTab != null && projectForTab.equals( project ))) {
```
eirikbakke, line 190: Why would the tab be closed in the project == null && projectForTab == null case?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
eirikbakke: Why does tabIndex == -1 indicate that closeSelectedDocumentList should be invoked? Would probably be cleaner to check if item.isSeparator(). Nice that you're handling this keypress, though!

eirikbakke (changes requested):
Thanks again for this PR! Added some comments and questions.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
Richard7111: There is a case, when no project is associated when file and variables projectForTab and project are null. Such as here for example: 
image

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
Richard7111: I am sorry, but I dont think I understand the first suggestion.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
Richard7111: You are right, thought I would keep the conditions similiar, but reads much better when it checks if item.isSeparator().

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
Richard7111: Not sure what white space, it is the end of file, I have not changed anything there.  Can you give me a hint please?

mbien: looks like you removed the newline symbol after '}'. Its common to have one at the end. That is what the stop sign means on github.

You can fix this by putting the cursor after '}' and press return.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
mbien: a defensive copy would look like:
`List<TabData> tabs = new ArrayList<>(controller.getTabModel().getTabs());`
instead of
`List<TabData> tabs = controller.getTabModel().getTabs();`

now you could simply iterate through `tabs` with a normal for loop since it is a copy of the original and won't change while you iterate.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/ButtonPopupSwitcher.java
Richard7111: Oh okay, I thougth it would be that, but looked into master and did not see newline at the end. Will add it there, thanks.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
Richard7111: I have added Javadoc explaining return values, hope it will be clear enough.

eirikbakke: Great, thanks!

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/DocumentSwitcherTable.java
eirikbakke: Makes sense. Thanks!

eirikbakke (approved):
Thanks for the changes! Tested on my own NetBeans build and confirmed to work.

eirikbakke:
(We can probably merge this one if no one objects within a couple of days.)

mbien:
@Richard7111 congrats to your first contribution. Thanks for your patience :)

eirikbakke:
I added the follow-up PR relating to Close All/Close Other type actions now ( https://github.com/apache/netbeans/pull/4977 ). Also relevant since someone wanted to implement a "Close Right" action ( https://github.com/apache/netbeans/issues/4710 ).

PR #4781, opened 2022-10-12
Cleanup warnings related to empty statement

BradWalker:
There are a few warnings related to an empty statement.

   [repeat] /home/bwalker/src/netbeans/ide/spi.debugger.ui/src/org/netbeans/modules/debugger/ui/actions/ConnectorPanel.java:204: warning: [empty] empty statement after if
   [repeat]         if (e.getActionCommand ().equals ("SwitchMe!")); // NOI18N
   [repeat]                                                        ^

This change cleans up the code to make the warnings go away.

---
**^Add meaningful description above**

By opening a pull request you confirm that, unless explicitly stated otherwise, the changes -

 - are all your own work, and you have the right to contribute them.
 - are contributed solely under the terms and conditions of the Apache License 2.0 (see section 5 of the license for more information).

Please make sure (eg. `git log`) that all commits have a valid name and email address for you in the Author field.

If you're a first time contributor, see the Contributing guidelines for more information.

If you're a committer, please label the PR before pressing "Create pull request" so that the right test jobs can run.

PR #4793, opened 2022-10-15
NBABasedBulkSearch::encode doesn't return on cancel

mbien:
@BradWalker found in #4781 a minor bug where `cancel.get()` was checked but ignored.

The last check is redundant since it is the end of the method.

This might have been a copy and paste bug, since other methods have checks like `if (cancel.get) return null;`. I suppose it was meant to only remove the `null` but leave the `return`


sdedic (approved):
I'd say the fix is OK; not sure if there are not other places where it is desirable to check for cancel flag.

mbien:
@sdedic thanks for taking a look. I also thought that this was fairly low risk - most likely a copy&paste bug.

PR #4795, opened 2022-10-16
Use frgaal compiler to compile NetBeans

jtulach:
**Warning**: Since 098e684 the purpose of this PR is "slightly" shifted.

This PR is a practical follow up of a [discussion recently held](https://lists.apache.org/thread/z0fx3b2vrjx2rqlf07mfr493tqnmc51b) on NetBeans public forums. The discussion was based on my [Admire Kotlin, Adhere to Java](https://www.youtube.com/watch?v=ua-8ySwFgqg) talk that highlights some of the benefits of Kotlin and describes that one doesn't have to use Kotlin to get the same benefits. In particular these two benefits are subject of this pull request:

> - language specification independent of the JDK
> - compiler version specified as part of project build script

This pull request achieves both goals by using [javac based frgaal compiler](http://frgaal.org) to compile NetBeans code base. Everything worked for [nb-javac](https://cwiki.apache.org/confluence/display/NETBEANS/Overview%3A+nb-javac) got integrated in #5609. However it turns the same infrastructure is enough for [frgaal compiler](http://frgaal.org) as well.

With this change, it doesn't matter what JDK one uses to build NetBeans. Even if there is some future JDK that drops `-release 8` support in `javac`, it doesn't matter, as the compiler for language version 20.0.0 is _part of the build script_. Not only the _fixture_ of compiler is beneficial for reproducing the past builds in distant future, it is also  beneficial for accessing the future: `ant -f java/java.disco/ clean build` now works on any JDK (not just JDK11+ as is the current  `master` state). This may not seem like a huge benefit, but think about modules using JDK17 or even JDK20 APIs (read [run on JDK8, use JDK11 APIs!](http://wiki.apidesign.org/wiki/AlternativeImplementation) for details) which conditionally unlocks access to all modern JDK features! 

This PR is a work in progress and serves primarily for demonstration purposes.


jtulach:
Let the discussion happen at [Shall NetBeans use fixed version of  javac compiler?](https://lists.apache.org/thread/3897rgnyc5crn82n9tp6vjo54kdo6qbo) thread.

jtulach:
Now when #5609 is integrated, there is no point in keeping this PR open. The NetBeans project doesn't want to use nb-javac as a compiler by default right now.

jtulach:
> ... there is no point in keeping this PR open...

Or: we can use the PR to check if NetBeans sources can also be built by [frgaal compiler](http://frgaal.org) and explore the possibilities it opens.

platform/api.search/src/org/netbeans/modules/search/matcher/BufferedCharSequence.java
```diff
@@ -170,12 +167,9 @@ private BufferedCharSequence(BufferedCharSequence bufferedCharSequence) {
     @Override
     @SuppressWarnings("FinalizeDeclaration")
     protected void finalize() throws Throwable {
-        try {
```
jtulach, line 173: One of the features of [frgaal](http://frgaal.org) is that it can report _"future deprecations"_ like [here](https://github.com/apache/netbeans/actions/runs/4657374401/jobs/8241911775?pr=4795#step:6:8204) - the compiler contains list of all deprecated methods up to JDK 20 and can emit warnings for each of them. When the module is compiled with `-Werror` one has to fix such _"future warnings"_ or `@SuppressWarnings` them.

platform/sampler/src/org/netbeans/modules/sampler/Sampler.java
```diff
@@ -187,6 +187,11 @@ public void run() {
         }, SAMPLER_RATE, SAMPLER_RATE);
     }
 
+    @SuppressWarnings("deprecation")
+    private static long findThreadId() {
```
JaroslavTulach, line 191: Another [future deprecation](https://github.com/apache/netbeans/pull/4795/files#r1161680325) - this time suppressed.

jtulach:
All the "tricks" here can be achieved by simple `nb-javac`. Except the 4feb401.The 4feb401 commit shows that one can use the latest Java language features without compromising support for other JDKs.

matthiasblaesing:
So that I don't forget it at a later time: I veto this change on the grounds, that it makes the project impossible to compile with javac (at time of writing there is usage of `if($var instanceof $type $newlyBoundVariable){}`, added ).
My understanding of the project is to write an IDE/Platform based on the Java language, with its reference implementation OpenJDK. I refuse to accept, that we rely on javac fork, that could vanish anytime, instead on the reference compiler as distributed by the JDK.

jtulach:
> refuse to accept, ...  javac fork, that could vanish anytime

All the necessary code is deployed on Maven central. Whatever once compile will compile forever. That's absolutely stable situation - way more stable than with JDK that keep dropping old `javac.target` with every release.

jtulach:
> My understanding of the project is to write an IDE/Platform based on the Java language, with its reference implementation OpenJDK

My intention, when I founded the project, was to make Java better. By writing a good IDE and also by showing how Java could look like.

It is sad to see that current maintainers of NetBeans gave up on making Java better and resigned on any attempt to inovate.

neilcsmith-net:
> It is sad to see that current maintainers of NetBeans gave up on making Java better and resigned on any attempt to inovate.

Well, that seems a distinctly unfair characterization of the points made.  Frgaal is a solution looking for a problem IMO.  It doesn't address any actual issues we face. Any use of a non-standard compiler here (frgaal or nb-javac) is going to run into a massive brick wall here, for a whole load of issues people have already brought up.

Quite frankly, that's far from the best or most fruitful way this project could still continue to innovate.  You're currently working on what I consider (and have used as an example in talks for years) a potentially very interesting project in the tooling space.  Lessons from that would be far more interesting to learn IMO.  I'd love to see us be more innovative and clear about what differentiates NetBeans from other options into the future.  If all that amounts to is frgaal, that really would be a sad state of affairs!

ebarboni:
Well late to the comment but this is not clear.
Is this the dependency on a possible vanishing fork the point ? 
Because we may have a "mojohaus"-like organization, with our own fork of nb-javac, frgaal we could put on mavencentral with a groupId we own (i.e. blabla.org) . This could solve the vanish as we could have several individual able to release. Karma beeing granted by  existing members with release rules that can be decided. 
That grants us, the ability to release anytime upstream lib, not only when individual wants. Risk of vanishing could be 0.
Have to make clear that vanishing is not for version 20.0.0 because it's sealed in maven central. Vanishing is what will happens tomorrow if NetBeans needs constructs of JDK 40 and frgaal people give up releasing, we will need to make a move to openjdk 40 with all the debt from JDK8 to JDK40. 
Pessimistic at the time od JDK40 they will be a CVE in 20.0.0 frgaal that will prevent his usage.
 
In a world we have dependencies at our hands, would it be still an issue ?

matthiasblaesing:
I'll assume good faith on behalf of the JDK developers, I also assume, that know what they are doing. It is a conscious decision on behalf of OpenJDK not to go down the "backporting" compiler route and also to use JDK-1 features in the javac. That can be questioned and criticized and I'd like to see a different decision, but it is not the case.

The onset of "frgaal" now is, that a fork of javac can accomplish what "plain" javac failed to do: A backporting javac, that can run on old versions of java. This is all noble, but how high is the bus-factor of this? For "plain" javac I don't see it going anytime soon. Even if Oracle would drop OpenJDK today, there are multiple companies backing the development, which already demonstrated, that they take over where Oracle drops the ball (Oracle tried to monetize their OpenJDK build and utterly failed as multiple vendors stood up to provide builds).

What is more, I'm pissed how this was brought in. The base PR, that brought the possibility for this, was advertised as an option to test nbjavac (turns out, that might have not been the whole truth). In that PR I pointed out, that there are problems, that make this not suitable for inclusion as a general solution (one such example is violating the contract for properties). I also pointed out, that the JDK has a mechanism to patch its javac with some handwaving, that was rejected.

From my POV the build system of NetBeans is complex enough without having to consider a wild mix of java compilers.

> In a world we have dependencies at our hands, would it be still an issue ?

How many people here are qualified to work on the java compiler and how many of this (I suspect) small group are willing to do it. And for how long?

jtulach:
> I'm pissed how this was brought in

That's not necessary. When working on #5609 I wasn't aware the `CustomJavac` task would work for Frgaal out of the box. I was expecting some modifications to be needed (for example the [entry point is supposed to be different](https://github.com/eppleton/frgaal#usage-on-command-line)). The fact that frgaal can work without any modification in `CustomJavac` task is still surprising to me.

PR #4800, opened 2022-10-16
ANTLRv4 indent and code snippet support.

lkishalmi:
This is my last round of features for ANTLR support, mainly for v4 grammars. Add indent support in rule definitions, and a few code shortcuts. 

Beside of that, prevented a few of NPE-s that could happen on parsing an incomplete grammar. Also removed the ConsoleErrorListener.

### DIFF
--- a/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4CompletionProvider.java
+++ b/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4CompletionProvider.java
@@ -33,8 +33,6 @@ import org.antlr.parser.antlr4.ANTLRv4Lexer;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.Token;
 import org.netbeans.api.editor.document.EditorDocumentUtils;
-import org.netbeans.api.editor.document.LineDocument;
-import org.netbeans.api.editor.document.LineDocumentUtils;
 import org.netbeans.api.editor.mimelookup.MimeLookup;
 import org.netbeans.api.editor.mimelookup.MimePath;
 import org.netbeans.api.editor.mimelookup.MimeRegistration;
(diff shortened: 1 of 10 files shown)


java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4CompletionProvider.java
```diff
@@ -143,12 +129,12 @@ protected void query(CompletionResultSet resultSet, Document doc, int caretOffse
...
                     } else {
-                        //Empty grammar so far offer lexer and grammar
-                        addTokens("", caretOffset, resultSet, "lexer", "grammar");
+                        //Empty grammar so far offer lexer, parser and grammar
+                        addTokens("", caretOffset, resultSet, "lexer", "parser", "grammar");
                     }
                 }
             } catch (Throwable th) {
-                System.out.println(th);
+                //System.out.println(th);
```
mbien, line 137: was this intended?  Would be better if it wouldn't be `Throwable`, It is always a bit dangerous to catch that without handling the edge cases.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4Formatter.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public boolean needsParserResult() {
+        return false;
+    }
+
+    @Override
+    public int indentSize() {
+        return 3;
+    }
```
mbien, line 138: 3 space indent is evil since it looks almost right :)

mbien (approved):
looks good as far as I can tell. But I didn't "think through" the formatting code. Its been a while since I wrote an Antlr grammar.

This is also fairly low risk for inclusion. +1 for merging this before feature freeze.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4Formatter.java
lkishalmi: Just checked a few grammars today randomly. Most of them were using 4 spaces, but a few of them were using 3. It seems our ANTLR v3 and v4 grammars using 3 spaces are kind of a minority.

java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4CompletionProvider.java
lkishalmi: Well, removed that one.

lkishalmi:
Well, a formatter code is never nice, as far as I checked. The basic idea here is using a heuristic based on the Lexer to determine whether we are in a rule or not. While in rule, the whitespace is examined for each line break, and indent the line according to the inrule state.

PR #4803, opened 2022-10-17
Update FlatLaf from 2.4 to 2.5

neilcsmith-net:
Replacement for #4696 fixing the merge conflict (I didn't want to force push directly to the PR).


mbien (approved):
PR looks good + it works on my system which is even better.

What are the native libs for again? Will this break NB if someone starts it on ARM or are those only for optional flatlaf features?

DevCharly (approved):

PR #4808, opened 2022-10-17
More precise code-completion for ANTLRv4 Grammars

lkishalmi:
Well, just a quick one. Lexer grammars, can only reference fragments inside rules, pure parser grammars cannot reference fragments. Add some heuristic to support lexercommands as well. + icons

### DIFF
--- a/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
+++ b/java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
@@ -53,7 +53,7 @@ public final class Antlr4ParserResult extends AntlrParserResult<ANTLRv4Parser> {
     private final List<String> imports = new ArrayList<>();
 
     private static final Logger LOG = Logger.getLogger(Antlr4ParserResult.class.getName());
-    public static final Reference HIDDEN = new Reference("HIDDEN", OffsetRange.NONE);
+    public static final Reference HIDDEN = new Reference(ReferenceType.CHANNEL, "HIDDEN", OffsetRange.NONE);
     
     public Antlr4ParserResult(Snapshot snapshot) {
         super(snapshot);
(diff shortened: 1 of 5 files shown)


java/languages.antlr/src/org/netbeans/modules/languages/antlr/v4/Antlr4ParserResult.java
```diff
@@ -113,38 +123,52 @@ public void exitTokensSpec(ANTLRv4Parser.TokensSpecContext ctx) {
...
 
 
+    private void collectReferences(FileObject fo, Map<String, Reference> refs, Set<String> visited) {
+        if (!visited.contains(fo.getName())) {
+            refs.putAll(references);
+            visited.add(fo.getName());
+            for (String im : imports) {
+                FileObject ifo = getFileObject().getParent().getFileObject(im + ".g4");
+                if (ifo != null) {
+                    Antlr4ParserResult pr = (Antlr4ParserResult) AntlrParser.getParserResult(ifo);
```
matthiasblaesing, line 155: Won't this blow when grammar files form a cycle? Assuming we have a direct cycle A -> B, B -> A, then:

- user ask for the parser result for file A
- the parse process will hit this section and ask for the parser result for file B
- the parser result is not found in the cache, so the parsing for B is invoked
- parsing for B will hit this section and ask for the parser result of file A

The cycle is closed. I just noticed, that we access other files from the parser. I think this is an error. This kind of error checking needs to be done outside the parser, then you can access the parser results sequentially and build whatever structure you need.

matthiasblaesing: This will also create stale errors - if you add a definition to import, there is no reason to invoke the Parser for the including file.

lkishalmi: You're right. I hope the new implementation is looking better.

matthiasblaesing (approved):
If I understand this correctly this works safely because a hard reference to the `Antlr4ParserResult`s is kept and thus we will not recurse indefinitely. Lets get this in.

lkishalmi:
Yes for the time the imports are processed we keep a hard reference on the results.

PR #4811, opened 2022-10-18
Update FlatLaf from 2.5 to 2.6

DevCharly:
Update FlatLaf to v2.6.

This version now uses `System.loadLibrary()` to load native libs (as suggested [here](https://github.com/apache/netbeans/pull/4696#issuecomment-1264613945)),
always loads `libjawt` (as suggested [here](https://github.com/apache/netbeans/pull/4696#issuecomment-1264336570))
and fixes some minor issues.

Changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/2.6

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-53B5104BDF92CBC7A650E075ED7296DD0AB77CF3 com.formdev:flatlaf:2.5
+4DD2BA228E3C57EB3D80E3927B5A6A33265EB69B com.formdev:flatlaf:2.6
--- a/platform/libs.flatlaf/external/flatlaf-2.6-license.txt
+++ b/platform/libs.flatlaf/external/flatlaf-2.6-license.txt
@@ -1,7 +1,7 @@
 Name: FlatLaf Look and Feel
 Description: FlatLaf Look and Feel
-Version: 2.5
-Files: flatlaf-2.5.jar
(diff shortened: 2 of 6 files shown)


neilcsmith-net:
Thanks @DevCharly Only minor niggle would be that this now should allow the native libraries to go in arch/OS specific paths like the others.  @matthiasblaesing ??

mbien:
restarted CI with all tests active

mbien:
#4803 broke `platform/core.startup` tests (restarted twice):
https://github.com/apache/netbeans/actions/runs/3272323634
exit code 1. So probably JVM crash - haven't investigated further.

This is fixing it. We should get this in ;)

mbien (approved):

matthiasblaesing (approved):
Looks sane to me. Verified on OpenJDK 17 on amd64 that the native library is loaded via the NetBeans mechanism (single stepped through `org.netbeans.StandardModule.OneModuleClassLoader.findLibrary(String)`)

neilcsmith-net:
@mbien wasn't a JVM crash.  The `CachingPreventsFileTouchesTest` was failing, either due to use of `InstalledFileLocator` in `validate()` or my suggestion to use the parent directory directly as the argument.

Possibly a sign we should be running platform tests more often?

mbien:
@neilcsmith-net agreed, I had the same thought. I could make `UI` trigger platform tests since it doesn't do anything right now. Or we should just use `Platform` more often?

but in general: if we upgrade a library which is used everywhere, we should add `ci:all-tests`. I missed that one too, I thought commit validation would catch a problem like this one.

PR #4826, opened 2022-10-20
use default github release drafter and update categories.

mbien:
the `release-drafter` action can't generate a "new contributors" category (https://github.com/release-drafter/release-drafter/issues/1001). Since first time contributors have to go through extra steps and show patience, we should try to at least give them credit.

This PR removes the `release-drafter` workflow (manual dispatch) and switches back to the [default github drafting mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes). It is less flexible but has hopefully what we need.

Added a lot more categories. (anything missing? ignore list sufficient?)

What I don't know is if items show up multiple times in multiple categories - I really hope not, otherwise this is going to be "interesting". Docs seem to leave the important parts out. This would have to be merged before testing.

It is also possible to edit old releases:
drafter

I don't know how to automate this. So we might have to do those few clicks manually on each release.


neilcsmith-net:
Will take a proper look ASAP.  Couple of quick thoughts.

Glad you picked up that we decided last week to add the `Release process` label for this.  An earlier addition was `Highlight`.  It's not been used much yet.  Idea was to tag PRs that bring in key new features to highlight in notes, and give @geertjanw something to write about! :smile:

We do seem to have a new contributors section already - may be missing what you mean?!

ebarboni:
for me is ok to merge (maybe we could direct commit to adjust the template later)

release drafter ok as it is not used or in create incomplete (no contributor)
 
the "current release note" are done manually from tag to tag using github.

technical ? if we rework a release note from tag 13-14 will it be taken or should it be on the release 140 branch ?

mbien:
> We do seem to have a new contributors section already - may be missing what you mean?!

@neilcsmith-net https://github.com/apache/netbeans/releases we used to have one when the defaults were used, until everything was re-generated with release-drafter. If you see one and I am just blind, please tell me :)

mbien:
oops I overlooked something:
when I look at this I don't see the new-contributors category:
https://github.com/apache/netbeans/releases

but if I click on the release there is one
https://github.com/apache/netbeans/releases/tag/15


But it would be still good to have categories right?

mbien:
The release which was created yesterday does have new contributors - so everything is ok.
https://github.com/apache/netbeans/releases/tag/16

but sure lets merge this and adjust based on demand.

ebarboni:
lets try

ebarboni:
looks like it's not retroactive :/

mbien:
well this is disappointing. I suppose it doesn't see the release.yaml yet if you edit an old release?

I suppose we could clone the repo. force push the yaml to NB 10 or so. Then generate releases there and edit/paste it into the official repo?

ebarboni:
also possible to push to releaseXXX, create a draft (that will no create a tag) copy the markdown to the xxxtag :D.  
may contains cleanup PR, will not work for patch also. Catch22

mbien:
i got another idea. what if we generate release text for:
- HEAD to 16
- HEAD to 15
- ...

and paste it into files. Then we write a program which subtracts them from each other and we have the right text for each release ;)

mbien:
I wrote a little java program for this purpose. Going to update the labels on a bunch of PRs and open another PR which adds some more categories. After that we can update the old release notes.

PR #4835, opened 2022-10-21
Maven UseReleaseOptionHint fixes

mbien:
- NPE fix when build tag is empty
 - don't show hint if there is no compiler plugin in the pom
 - simplified code, added more tests

targets delivery

### DIFF
--- a/java/maven.hints/src/org/netbeans/modules/maven/hints/pom/UseReleaseOptionHint.java
+++ b/java/maven.hints/src/org/netbeans/modules/maven/hints/pom/UseReleaseOptionHint.java
@@ -21,6 +21,7 @@ package org.netbeans.modules.maven.hints.pom;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.prefs.Preferences;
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
Problem with 2 (no compiler plugin) is it now doesn't show in the default created POM.  That might be something to address in master though?  In fact, probably the whole default POM needs some consideration.

mbien:
The hint can't show with the default POM because the default compiler plugin in use without setting it, would be < 3.6. So it would complain about not knowing what the release flag is during build. I overlooked this detail at first too.

(for the same reason we can't update the pom template unless we also add the compiler plugin by default)

neilcsmith-net:
Yes, I understand that.  Would need to link to code like that which detects `module-info` possibly?  But I think we probably do need a better starter POM template.

Anyway, I think this is good for NB16. I did do a quick review over before merging!

mbien:
> But I think we probably do need a better starter POM template.

I thought about a JDK selector in the new-project-wizard. If a modern JDK is selected it would use a template which has a compiler plugin declaration and the right release option set for the selected JDK. (project would have that JDK also pre selected)

PR #4836, opened 2022-10-21
Update JDK 19 Javadoc link from EA to GA.

neilcsmith-net:
Update JDK 19 Javadoc link from early access to the release location.

Quick fix in delivery as I noticed while testing 16-rc1 on JDK 19 that Javadoc linking is broken.  However, might be good to consider a better way of handling this in future.


mbien:
oh nice, I wanted to update that but totally forgot. Was thinking to generate those links somehow. Take the date, do some math and calculate what the URL must be. JDK release schedule is super stable.

mbien (approved):

neilcsmith-net:
@mbien yes, that's similar to what I was thinking - mostly why I pinged you for review. :smile:

neilcsmith-net:
I'd rather keep this for NB16, and target better changes for NB17.  I'm also wondering whether the date approach is the right one, rather than prepopulating all the ones up to the supported source level in the IDE, and possibly pinging the docs site if a higher JDK is requested.

PR #4838, opened 2022-10-21
put nb-javac modules on the plugin import block list.

mbien:
Users which import nb-javac plugins from old NB versions complain about not being able to start NB anymore (most recent: #4799). This puts all nbjavac modules which were in NB 12.1 on the block list and will hopefully prevent this in future.

targets delivery


matthiasblaesing (approved):
Looks sane to me.

neilcsmith-net (approved):
Good call, thanks!  Should have been done when we first pre-bundled again.

lkishalmi:
Cool, I did not know we have such a feature!

Issue #4686, opened 2022-09-25
[PHP] Inheriting docs is broken


terax6669:
### Apache NetBeans version

Apache NetBeans 15

### What happened

NetBeans doesn't inherit docs for class properties, see example below.

### How to reproduce

```php
<?php

/**
 * This works fine
 */
class Foo
{
	/**
	 * This thing is broken
	 */
	public $bar;

	/**
	 * This also works fine
	 */
	public function foobar() {}
}

class Bar extends Foo
{
	public $bar = 'bar';

	public function foobar() {}
}

/**
 * {@inheritDoc}
 */
class FooBar extends Foo
{
	/**
	 * {@inheritDoc}
	 */
	public $bar = 'bar';

	/**
	 * {@inheritDoc}
	 */
	public function foobar() {}
}

/**
 * {@inheritDoc} trying to extend docs
 * won't work!
 */
class FooBarFoo extends Foo
{
	/**
	 * {@inheritDoc} this is just ridiculous
	 */
	public $bar = 'bar';

	/**
	 * {@inheritDoc} sadly this is also broken
	 */
	public function foobar() {}
}

$foo = new Foo;            // IDE hint: This works fine
$foo->bar;                 // IDE hint: This thing is broken
$foo->foobar();            // IDE hint: This is ridiculous

$bar = new Bar;            // IDE hint: This works fine
$bar->bar;                 // IDE hint:
$bar->foobar();            // IDE hint: This also works fine

$foobar = new FooBar;      // IDE hint: This works fine
$foobar->bar;              // IDE hint: {@inheritDoc}
$foobar->foobar;           // IDE hint: This also works fine

$foobarfoo = new FooBarFoo; // IDE hint: {@inheritDoc} trying to extend docs won't work!
$foobarfoo->bar;            // IDE hint: {@inheritDoc} this is just ridiculous
$foobarfoo->foobar;         // IDE hint: {@inheritDoc} sadly this is also broken
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Windows 11 Pro 22000.978

### JDK

OpenJDK 19.0.0 (from Chocolatey)

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

https://pear.php.net/package/PhpDocumentor/docs/latest/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

junichi11:
[...]

PR #4846, opened 2022-10-23
Improve `{@inheritdoc}` behavior #4686

junichi11:
#4686

- Support `{@inheritdoc}` for fields and constants
- Replace `{@inheritdoc}` with whole parent sentences

#### Example
```php
<?php

/**
 * Parent class.
 * 
 */
class ParentClass
{
	/**
	 * Parernt field.
	 */
	public $field;

        /**
         * Parent const.
         */
        public const CONSTANT = "CONSTANT";

	/**
	 * Parent method.
	 */
	public function method() {}
}

class Child extends ParentClass
{
	public $field = 'field';

        public const CONSTANT = "CONSTANT";

        public function method() {}
}

/**
 * {@inheritDoc}
 */
class ChildWithTag extends ParentClass
{
	/**
	 * {@inheritDoc}
	 */
	public $field = 'field';

        /**
         * {@inheritDoc}
         */
        public const CONSTANT = "CONSTANT";

	/**
	 * {@inheritDoc}
	 */
	public function method() {}
}

/**
 * Header.
 * {@inheritDoc} Child class.
 * 
 */
class ChildWithTagAndAdditionalComment extends ParentClass
{
	/**
	 * {@inheritDoc} child field.
	 */
	public $field = 'field';

        /**
         * {@inheritDoc} child const.
         */
        public const CONSTANT = "CONSTANT";

	/**
	 * Header.
         * {@inheritDoc} child method.
         * test.
	 */
	public function method() {}
}

$parent = new ParentClass();
$parent->field;
$parent::CONSTANT;
$parent->method();

$child = new Child();
$child->field;
$child::CONSTANT;
$child->method();

$childWithTag = new ChildWithTag();
$childWithTag->field;
$childWithTag::CONSTANT;
$childWithTag->method();

$childWithTag2 = new ChildWithTagAndAdditionalComment();
$childWithTag2->field;
$childWithTag2::CONSTANT;
$childWithTag2->method();
```

nb-php-gh-4686

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
@@ -811,6 +811,12 @@ final class DocRenderer {
             } else if (indexedElement instanceof InterfaceElement) {
                 InterfaceElement interfaceElement = (InterfaceElement) indexedElement;
                 inheritedElements.addAll(getAllInheritedInterfaces(interfaceElement));
+            } else if (indexedElement instanceof FieldElement) {
+                FieldElement fieldElement = (FieldElement) indexedElement;
+                inheritedElements.addAll(getAllOverriddenFields(fieldElement));
+            } else if (indexedElement instanceof TypeConstantElement) {
+                TypeConstantElement constElement = (TypeConstantElement) indexedElement;
+                inheritedElements.addAll(getAllOverriddenConstants(constElement));
             }
             return inheritedElements;
         }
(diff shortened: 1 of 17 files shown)


junichi11:
@tmysik Could you please have a look at this when you have time? If there is no problem, let's merge it. Thanks.

I've implemented this based on PhpDocumentor in the past. If there is exact documentation for `{@inheritdoc}`, we should follow it. However, I can't find the exact specification about `{@inheritDoc}` now...
Currently, when PHPDoc has only a single sentence(e.g. Get something.), inline `{@inheritdoc}` is not replaced by anything (e.g. `/** {@inheritdoc} something...*/`). If only `{@inheritdoc}` is written, `{@inheritdoc}` is replaced with it(Get something). So, I've changed it to the same behavior.

tmysik (approved):
Looks good to me. Thanks for all the tests.

junichi11:

tmysik:
@junichi11 Nothing to thank for, really. You did the job, as always.

Issue #4847, opened 2022-10-23
Gradle won't load project in NetBeans-16rc1


errael:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

Attached zip file has trivial projects that demo the problem

Running `NB-master, Oct 19 just after freeze` under NB-15 debugger

Open lib - resolve problem

Observe:       
```
    Cannot load: /.../lib
    Could not run build action using
        /home/err/.gradle/wrapper/dists/gradle-6.8.3-bin
            /7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3
```
On the System, there is:

```
        /home/err/.gradle/wrapper/dists/gradle-6.8.3-bin
            /7ykxq50lst7lb7wx1nijpicxn/gradle-6.8.3/bin/gradle
```

Right click lib - select build, `BUILD SUCCESSFUL`

Still has PrimingBuildRequired and Resolve still fails; Cannot load

**NOTE: There are two smallish bugs in NB-15, mentioned in the "How to reproduce" section; since can't open on 16...**

### How to reproduce

Steps are shown above, to reiterate

1. jar -xf _attached-zip-file_, 3 projects: lib,app,compose
[gradle-projects-compose.zip](https://github.com/apache/netbeans/files/9847401/gradle-projects-compose.zip)
2. open lib, resolve doesn't work, see above

### This is what happens on NB-15 (with two small bugs)

**=== problem 1 (minor problem)**

0. tar -x projects
1. clean userdir/cachedir
2. Open app,lib - trust yes, resolve problem
3. Observe lib resolves, not app.
4. Attempt to resolve app, Does not resolve app, no error indication or output.
5. Do app > build, see expected error
       Execution failed for task ':compileJava'.
       > Could not resolve all files for configuration ':compileClasspath'.
          > Could not find play:lib:1.0.0.
   Resolve should show this rather than failing silently
6. Execute task lib publishToMavenLocal
7. Now app resolves as expected.

Only problem is that step 4 does not show an error (which can be seen by manually running build as in step 5)

**=== problem 2**

With all problems resolved,
right click on app, "SetMainProject" does not appear in context menu.

### Did this work correctly in an earlier version?

Apache NetBeans 15

### Operating System

Linux harmony 5.19.16-76051916-generic #202210150742~1666053244~22.04~cf07008 SMP PREEMPT_DYNAMIC Tue O x86_64 x86_64 x86_64 GNU/Linux

### JDK

[...]

PR #4850, opened 2022-10-24
#4847: do not use newer classes in older distributions.

sdedic:
I've added version guards to the newer introspection entry points. Also added tests that ensure a basic project will load in `6.8.3`, `7.0` and `7.1` gradles.


lkishalmi (approved):
Looks good to me.

Issue #4276, opened 2022-06-23
All implementations of abstract methods are considered unused


terifan:
### Apache NetBeans version

Apache NetBeans 14

### What happened

Obvious bug that make code harder to read. Was working in previous version.

### How to reproduce
```java
package test;

public class Test
{
	public static void main(String[] args)
	{
		Abstr impl = new Impl();
		impl.get();
	}
}

abstract class Abstr
{
	abstract void get();
}

class Impl extends Abstr
{
	@Override
	void get() // <---- considered unused
	{
		System.out.println("hello world");
	}
}
```

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Windows 8.1

### JDK

17.0.2

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

mbien:
note: even though this looks like a regression, it isn't since NB didn't have the hint before NB 14 (#3841).

neilcsmith-net:
As mentioned in #4783 this affects non-abstract package-private methods too.  I think we should hide or remove the hint if not fixable by NB16 release @dbalek

neilcsmith-net:
Have noticed that even with the hint disabled, methods are still underlined with an unused message in the tooltip.

neilcsmith-net:
Opened a potential fix for this - cc @mbien @dbalek

neilcsmith-net:
Should be fixed in 16-rc2, available from #4815 Please verify /cc @dbalek

PR #4857, opened 2022-10-25
Never mark overriding methods as unused, fixes GH4276.

neilcsmith-net:
Potential fix for #4276 that never marks an overriding method as unused.  The superclass definition will still be marked unused if not called.  IMO the method can be considered "used" by the superclass, and hints should be over-cautious in preference to over-eager anyway.

### DIFF
--- a/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
+++ b/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
@@ -141,7 +141,8 @@ public class UnusedDetector {
                     }
                 }
             } else if ((el.getKind() == ElementKind.CONSTRUCTOR || el.getKind() == ElementKind.METHOD) && (isPrivate || isPkgPrivate)) {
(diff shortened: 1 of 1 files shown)


mbien (approved):
fix makes sense to me. The `overridesMethod` call is probably also at the right position from the performance perspective since it likely is the second most expensive call in the `&&` chain (has to traverse the full hierarchy).

I also agree that default hints should be more on the cautious side. Inspections on the other hand which are disabled by default could allow a bit more wiggle room if the description communicates that transformations should not be applied blindly. (#4592 had a similar discussion)

neilcsmith-net:
Thanks @mbien merging.  Strictly speaking this doesn't just affect hints either - unused elements are also highlighted even if the hint is switched off.

mbien:
hmm. why does gh not let me link the two issues to this PR.

neilcsmith-net:
@mbien it would appear they have to be open.

mbien:
i see. which is weird because I did seem some suggestions of closed issues in the list - so I thought it is probably something else.

PR #4864, opened 2022-10-26
fixes an ambiguous method reference in SearchTask.

mbien:
The compiler can't decide if this is an `ExceptionAction` or a `Runnable` and marks the line as an error in-editor. Interestingly, the module builds fine.

We can help the compiler by casting to Runnable or by using a lambda expression. Cast is more compact.

ambiguous

### DIFF
--- a/platform/api.search/nbproject/project.properties
+++ b/platform/api.search/nbproject/project.properties
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 is.autoload=true
-javac.source=1.8
-javac.compilerargs=-Xlint:all -Xlint:-serial -Xlint:-processing -Xlint:-options
+javac.release=11
+javac.compilerargs=-Xlint:all -Xlint:-serial
 javadoc.arch=${basedir}/arch.xml
 javadoc.apichanges=${basedir}/apichanges.xml
--- a/platform/api.search/src/org/netbeans/modules/search/SearchTask.java
+++ b/platform/api.search/src/org/netbeans/modules/search/SearchTask.java
@@ -83,7 +83,7 @@ final class SearchTask implements Runnable, Cancellable {
         try {
             makeResultViewBusy(true);
             searchListener.searchStarted();
-            Mutex.EVENT.writeAccess(resultViewPanel::requestFocusInWindow);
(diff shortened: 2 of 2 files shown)


matthiasblaesing:
The change looks ok, but it raises the question whether the javac or NetBeans java editing is wrong.

mbien:
> The change looks ok, but it raises the question whether the javac or NetBeans java editing is wrong.

yeah exactly. The module uses a different nb-javac version than the one in the editor. The one in the editor appears to be more strict.

matthiasblaesing:
@jlahoda @singh-akhilesh could you please have a look at this? Either there is a bug in the OpenJDK javac or in the NetBeans Java Editor integration of javac. Only one of these can be right. At least I hope, that the JLS defines which method resolution is the right one.

mbien:
I still see the same in-editor error in NB 18 using vanilla javac of JDK 20.0.2 and in NB 19 which currently uses javac based on JDK 21+26. The module builds fine on all tested JDKs without warnings in that class.

@lahodaj @jlahoda @dbalek any idea what could cause this?

mbien:
removing milestone

mbien:
revisited this old PR now that we have enough infrastructure to be able to use `javac.release=11`.. and as soon the release flag is set it does also fail in the build:
```
/home/mbien/NetBeansProjects/netbeans/platform/api.search/src/org/netbeans/modules/search/SearchTask.java:86: error: reference to writeAccess is ambiguous
            Mutex.EVENT.writeAccess(resultViewPanel::requestFocusInWindow);
  both method <T>writeAccess(ExceptionAction<T>) in Mutex and method writeAccess(Runnable) in Mutex match
  where T is a type-variable:
    T extends Object declared in method <T>writeAccess(ExceptionAction<T>)
```
this would make the behavior at least consistent between embedded javac use and CLI javac use - which was my main worry tbh. (cc @lahodaj )

Will refresh this PR and bump the module to release 11 which would make this PR essentially a part of https://github.com/apache/netbeans/issues/8813

mbien:
planning to merge this soon since: 1) trivial and 2) mystery of the editor inconsistency is resolved

edit: at least three of the module dependencies use javac.release=11 or 17 already, so the bump should have no side effects here

Issue #4758, opened 2022-10-09
[jackpot] Declarative rules don't match lambdas


mbien:
### Body

migrated issue from [NETBEANS-6025](https://issues.apache.org/jira/browse/NETBEANS-6025). Still reproducible in NB 15.

The first rule works. The second rule never captures anything, the third causes a class cast exception.

```java
$stream.mapToObj($lambda) :: $stream instanceof java.util.stream.IntStream =>
$stream.boxed()
;;

$stream.mapToObj($x -> $x) :: $stream instanceof java.util.stream.IntStream =>
$stream.boxed()
;;

$stream.mapToObj(($x) -> $x) :: $stream instanceof java.util.stream.IntStream =>
$stream.boxed()
;;
```

```java
// test snippet:
IntStream.of(1,2,3).mapToObj(x -> x);
```
<details>
<summary>exception of third rule:</summary>

```
java.lang.ClassCastException: class org.netbeans.modules.java.hints.spiimpl.JackpotTrees$VariableDeclWildcard cannot be cast to class com.sun.tools.javac.tree.JCTree$JCIdent (org.netbeans.modules.java.hints.spiimpl.JackpotTrees$VariableDeclWildcard is in unnamed module of loader org.netbeans.StandardModule$OneModuleClassLoader @24b6d0a8; com.sun.tools.javac.tree.JCTree$JCIdent is in unnamed module of loader org.netbeans.StandardModule$OneModuleClassLoader @68b9c85e)
	at com.sun.tools.javac.tree.TreeCopier.visitIdentifier(TreeCopier.java:244)
	at com.sun.tools.javac.tree.TreeCopier.visitIdentifier(TreeCopier.java:44)
	at org.netbeans.modules.java.hints.spiimpl.JackpotTrees$VariableDeclWildcard.accept(JackpotTrees.java:80)
	at com.sun.tools.javac.tree.TreeCopier.copy(TreeCopier.java:60)
	at com.sun.tools.javac.tree.TreeCopier.copy(TreeCopier.java:72)
	at com.sun.tools.javac.tree.TreeCopier.visitBlock(TreeCopier.java:135)
	at com.sun.tools.javac.tree.TreeCopier.visitBlock(TreeCopier.java:44)
	at com.sun.tools.javac.tree.JCTree$JCBlock.accept(JCTree.java:1094)
	at com.sun.tools.javac.tree.TreeCopier.copy(TreeCopier.java:60)
[...]

PR #4868, opened 2022-10-27
Fix maven classpath listening

timboudreau:
I noticed, in a project that generates a lot of classes using annotation processors, that frequently the IDE would not reparse after code generation, and so error markers on code that references generated classes and similar would not be updated.

I tracked it down to this:

The previous code calls `WeakListeners.propertyChange()` *with a lambda*. The lambda will be garbage collected instantaneously and never fire changes, which was exactly the behavior.

On JDK-8, lambdas and method references would hang around a while before being garbage collected - so this code might have looked like it worked originally.  On newer JDKs, the lambda will be garbage collected almost as soon as it is allocated, so notifications will never propagate.

I've been running this patch for a month or so without problems.

One slightly questionable thing:  There was commented out code to also listen on `proj.getProjectWatcher()`.  It seemed harmless, now that the class is implementing `PropertyChangeListener` directly, and the code wanted the same listener implementation, to uncomment it, so I did.

I can definitely verify that it works, because I had to fix in a deadlock in the icon-badging support in my Antlr plugins that could only be triggered by the events that were never fired :-)

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
+++ b/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
@@ -712,7 +712,8 @@ public final class ClassPathProviderImpl implements ClassPathProvider, ActiveJ2S
      * This selector chooses the annotation classpath, if it is not empty (has items, or is broken), or the regular
      * compile classpath if annotation path is empty. The selector reacts 
      */
-    private static class AnnotationPathSelector extends ClassPathSelector {
+    private static class AnnotationPathSelector extends ClassPathSelector
+            implements PropertyChangeListener {
         private final ClassPath annotationCP;
         private final Supplier<ClassPath> compileClassPath;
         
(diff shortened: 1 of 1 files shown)


neilcsmith-net:
Makes sense.  @timboudreau @sdedic please make a call if this should be targeted at NB16 and delivery, and if so, update as necessary - thanks!

mbien:
interesting discovery. The doc of `Cleaner` has also an extra note on lambdas:
https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/ref/Cleaner.html

would have been good to inspect the codebase with a jackpot script to catch other occurrences but jackpot can't match lambdas yet unfortunately (#4758).

how comes that @timboudreau doesn't has commit rights yet? He wrote half of the code ;)

neilcsmith-net:
I keep forgetting @timboudreau isn't on the PMC!  Still need a call on whether to target this to NB16 though.

The Cleaner documentation is almost the reverse problem.

lkishalmi (approved):
Looks sane.

mbien:
@timboudreau would you be interested in becoming a NetBeans committer? This would unlock more features on github for this repo (being able to restart CI jobs is my favorite :)). We very much think that you should be a committer at the very least given how much code originates from you. If yes we would send an email with more details.

timboudreau:
> would you be interested in becoming a NetBeans committer?

I spent nearly 20 years of my life being **Mr. NetBeans**, so I pretty intentionally did *not* sign up to be one when things moved to Apache.

But yeah, I supposed I could be talked into it.

Re the patch, it would sure be nice for it to be integrated into the next release.  It makes a difference in the IDE actually picking up changes from depended-on projects (the telltale that it works is that, with the patch installed, when you edit a parent POM file and change, say, the version of a library, you'll see the *Background scanning of projects* progress bar pop up briefly.

It wouldn't be too hard to detect use of a lambda with WeakListeners (or Lookup.Result, which would also have this problem) from inside Javac.  I don't know if we currently have an easy way to write hints that use javac's api directly rather than via Jackpot, but the test is simply "is this a lambda" == "is this one of [list of things that add to a weak set]".  Solving it for specific, common cases is probably more important than solving the general case of "added to anything that might get wrapped in a weak reference".

mbien:
@timboudreau I changed this PR to target the delivery branch so that it could potentially get into NB16.

However, you would have to cherry pick this commit on top of the current delivery branch and then force push into this PR again to make this work. Otherwise we have a messed up history.

timboudreau:
Well, shoot.  Really, there is no rebase option from Github's merge UI?  It seems a pity to have it held hostage to the original author rebasing it - github emails go to an inbox I don't check all that often.

PR #4872, opened 2022-10-27
wrap unreliable micronaut tests in retry script for now

mbien:
as discussed on the dev list -> lets enable the retry script for the micronaut tests for now

targets delivery


sdedic (approved):

neilcsmith-net (approved):
Thank you! :smile:

PR #4873, opened 2022-10-27
Revert SpecificationVersion generics update

mbien:
followup on #4678 reverts the changes discussed here: https://github.com/apache/netbeans/pull/4678#pullrequestreview-1127588052

targets delivery


matthiasblaesing (approved):
Looks sane to me. Thank you for the quick revert.

matthiasblaesing:
Please wait before merging. I was wrong with my reasoning.

mbien:
tbh I don't think this actually needs to be reverted because the only safe way to stay backwards compatible in java is to build against the minimum version of the library you target. 

And since I can't come up with a scenario where code which was written against NB 15 wouldn't run on this, I still think this is safe.
(but i am also OK with a revert)

neilcsmith-net:
I'm inclined to keep the generics update as is too.

mbien:
closing for now - feel free to reopen if there are any new insights

PR #4877, opened 2022-10-28
about panel needs a vertical scrollbar if the async plugin update notification appears

### LINKED ISSUE #4876: "Help > About" doesn't show cache directory
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

CacheDirNotVisible

Problem occurs when there are updates available. Enabling the vertical scroll bar might fix it.

### How to reproduce

In plugins need to have an update center with available updates enabled.

No / Don't know

### Operating System

Linux harmony 5.19.16-76051916-generic #202210150742~1666053244~22.04~cf07008 SMP PREEMPT_DYNAMIC Tue O x86_64 x86_64 x86_64 GNU/Linux

### JDK

openjdk version "11.0.2" 2019-01-15 OpenJDK Runtime Environment 18.9 (build 11.0.2+9) OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

### Apache NetBeans packaging

Own source build

### Anything else

_No response_

No

### Code of Conduct

Yes

mbien:
fixes #4876

we can't add it right away, otherwise the window can't layout itself without hardcoding the size of the scrollpane. Changing window size once its visible would look weird.

targets delivery

### DIFF
--- a/platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.java
+++ b/platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.java
@@ -36,6 +36,7 @@ import java.util.stream.Collectors;
 import javax.swing.Icon;
 import javax.swing.JEditorPane;
 import javax.swing.JPanel;
+import javax.swing.ScrollPaneConstants;
 import javax.swing.SwingUtilities;
 import javax.swing.UIManager;
 import javax.swing.event.HyperlinkEvent;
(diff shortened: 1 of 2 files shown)


mbien:
@neilcsmith-net this is a minor low risk bug fix for delivery.

sb

enables the scrollbar if the "Updates:" text appears.

neilcsmith-net:
@mbien already seen and on the radar. Looks good, makes sense. Thanks!

platform/o.n.core/src/org/netbeans/core/ui/ProductInformationPanel.java
```diff
@@ -102,6 +103,7 @@ public ProductInformationPanel() {
             final String updates = getUpdates();
             SwingUtilities.invokeLater(() -> {
                 description.setText(LBL_description(getProductVersionValue(), getJavaValue(), getVMValue(), getOperatingSystemValue(), getEncodingValue(), getSystemLocaleValue(), getUserDirValue(), Places.getCacheDirectory().getAbsolutePath(), updates, FONT_SIZE, getJavaRuntime()));
+                descriptionScrollPane.setVerticalScrollBarPolicy(ScrollPaneConstants.VERTICAL_SCROLLBAR_ALWAYS); // will need a scrollbar now
```
errael, line 106: It's not clear to me what `now` is referring to in the comment.

mbien: the moment the description text is updated. Which is what the lambda does. So at the point this lambda runs.

errael: Thanks, I was confused by the title ".... scrollbar if the async plugin update..." but it's a little more like
"... scrollbar in case the async plugin update...".

I had thought it might mean now that there are updates available.
So the scrollbar is there, even when there are no updates available.

PR #4878, opened 2022-10-29
Fixed end of file indenting and code completion for ANTLR v4

lkishalmi:
Well, if we would have a 16-rc3 this would be a nice to have.

This one makes the formatting work if the editing happens at the end of the file.

The code completion has a fix and a feature creep. At the end of the file due to the bug in `AntlrTokenSequence` it was offering (`lexer`, `parser`, `grammar`).

The feature creep is a trivial one. We already have the unknown references collected during parsing, so why not offer to define those while not in a rule definition.

### DIFF
--- a/java/languages.antlr/src/org/netbeans/modules/languages/antlr/AbstractAntlrLexer.java
+++ b/java/languages.antlr/src/org/netbeans/modules/languages/antlr/AbstractAntlrLexer.java
@@ -29,13 +29,13 @@ import org.netbeans.spi.lexer.TokenFactory;
  *
  * @author lkishalmi
  */
-public abstract class AbstractAntlrLexer implements Lexer<AntlrTokenId> {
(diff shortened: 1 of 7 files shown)


neilcsmith-net:
@lkishalmi we aim to have at least 3 release candidates, so 16-rc3 is happening.  Have retriggered the failing tests (failure in binary downloads so not related to changes). I'll wait for @matthiasblaesing review.

matthiasblaesing (approved):
Looks ok to me. I noticed another bug in the formatting though (not introduced here).

1. Open the `ANTLRv4Lexer.g4` file
2. Format it
3. You get this: 
image


Line 49 and 53 should not be indented in my mind.

lkishalmi:
@neilcsmith-net Thanks for the merge!

@matthiasblaesing I've tried to solve this indenting issue, but that got too convoluted, and caused other issues. I fear I have to rewrite the formatting code for that. So far what that's just heuristics.

PR #4879, opened 2022-10-29
Improving the Java AST debugging view.

jlahoda:
The AST debug view for Java (`java/java.debug`) has some trouble:
 - it has hardcoded all AST node kinds, so if a new AST node kind is added, it needs an update
 - the highlight showing the current AST node's span is not cleared when the view is unselected
 - the highlight color is really ugly

This patch attempts to improve these aspects.

---


jlahoda:
In Navigator for Java source files, there should be a few debugging panels, like `Trees (NetBeans Development)`, which this patch is updating. This is part of the standard build, but is only enabled when assertions are enabled, or `-J-Dorg.netbeans.modules.java.debug.enable=true` command line option is used.

mbien (approved):
code makes sense to me as far as I can tell. It works also as advertised.

learned something new, this certainly is a useful tool to debug some editor features.

lets get this in - should be low risk anyway since this is not enabled by default..

PR #4885, opened 2022-10-29
A prototype of code that allows code embedded using TextBlocks; also adds support for Java as an embedded language.

jlahoda:
---


mbien:
awesome feature. I remember one of my first thoughts when hearing that java was getting multi line strings was that NetBeans is pretty good with embedding other languages - this would make a great match.

ide/parsing.api/src/org/netbeans/modules/parsing/impl/TaskProcessor.java
```diff
@@ -566,6 +566,25 @@ public boolean apply(Request request) {
             return cancelTime == 0 ? 0 : now - cancelTime;
     }
 
+    /*test*/ static <T extends Parser.Result> List<Embedding> callParserBaseEmbeddingProvider (
+            final @NonNull ParserBasedEmbeddingProvider<T> task,
+            final @NullAllowed T result,
+            final @NullAllowed SchedulerEvent event) {
```
lbownik, line 572: put blank line after brace cause not obvious where method body begins

java/java.editor.base/src/org/netbeans/modules/java/editor/base/embedding/EmbeddingProviderImpl.java
```diff
@@ -0,0 +1,189 @@
...
+            return Collections.emptyList();
+        }
+        List<Embedding> embeddings = new ArrayList<>();
+        TokenSequence<?> ts = TokenHierarchy.get(info.getSnapshot().getSource().getDocument(true)).tokenSequence(); //XXX: locking!
+        ts.moveStart();
+        while (ts.moveNext()) {
+            if (ts.token().id() == JavaTokenId.MULTILINE_STRING_LITERAL && ts.token().partType() == PartType.COMPLETE) {
+                String language = findLanguageFromTarget(info, ts.offset() + 1);
+                String mimeType = language != null ? mapLanguageToMimeType(language) : null;
+                if (mimeType == null) {
```
lbownik, line 79: would inverting the comparison and removing continue improve readability

java/java.editor.base/src/org/netbeans/modules/java/editor/base/embedding/EmbeddingProviderImpl.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+            case VARIABLE: {
+                target = (VariableElement) info.getTrees().getElement(tp.getParentPath());
+                break;
+            }
+        }
+        if (target == null) {
+            return null;
+        }
+        for (AnnotationMirror am : target.getAnnotationMirrors()) {
```
lbownik, line 124: lines 124 to 133 look loke a candidate for a private static method

java/java.lexer/src/org/netbeans/lib/java/lexer/JavaMultiLineStringLexer.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        int len = 0;
+        while (len == 0)
+            len = changePoints[changePointsPointer++];
+        for (int i = 0; i < len; i++) {
+            if (input.read() == LexerInput.EOF && len == 1) {
+                return null;
+            }
+        }
+        JavaMultiLineStringTokenId id;
```
lbownik, line 74: lines 74 to 80 look like a candidate for private  method

private JavaMultiLineStringTokenId idFromPointer() {
 
        switch (changePointsPointer % 3) {
            case 0: id = return JavaMultiLineStringTokenId.NEWLINE;
            case 1: id = return JavaMultiLineStringTokenId.INDENT; 
            case 2: id = return JavaMultiLineStringTokenId.TEXT; 
            default: throw new IllegalStateException();
        }
}

java/java.lexer/src/org/netbeans/lib/java/lexer/JavaMultiLineStringLexer.java
```diff
@@ -0,0 +1,108 @@
...
+        return tokenFactory.createToken(id);
+    }
+
+    private int leadingIndent(String line) {
+        int indent = 0;
+
+        for (int i = 0; i < line.length(); i++) { //TODO: code points
+            if (Character.isWhitespace(line.charAt(i)))
+                indent++;
+            else
```
lbownik, line 90: this else block is a dead code

neilcsmith-net:
This is still marked as a draft, so removing the milestone entirely rather than bumping forward to NB19.  Please add a milestone back when you feel it's ready for merging.

matthiasblaesing:
One problem solved here also affects JShell Integration. Using `var` in JShell causes Exceptions and fails to work properly. The hack from 2f4dd82 together with a few followups fixes makes it work again.

I think the core problem is, that `var` is not a keyword. It is only valid when used in (local) variable declarations, so from my understanding `var` on the lexer level can only be an identifier and only the parser level can decide whether we see the identifier `var` or a local variable declaration. At least that is my take away from skimming through the JLS.

Chris2011:
Is this only a Java in NetBeans feature? I mean we can also have SQL in strings in JS, PHP and also HTML in JS and in PHP etc. This would be very nice to handle multi embedding in strings in different languages. Had just a quick look and it seemed more or less just for Java atm? IntelliJ handles this with a comment for permanent injection or with a hint for strings with temporarely injection: https://www.jetbrains.com/help/idea/using-language-injections.html.

Is it really common to make annotations per IDE? The Matisse GUI editor is using smth like this for make the code readonly `// <editor-fold defaultstate="collapsed" desc="Generated Code">//GEN-BEGIN:initComponents`. Don't get me wrong, I really like the feature but when I see annotation I think of "@language" and this is not what I want to see in my code. A comment is more clear IMHO. Like highlighting // TODO:

mbien:
annotations are the right way to add metadata to code. +1 from my side for using them for the intended purpose.

Chris2011:
This needs to be a language independent feature otherwise you can add this into the JDK as a Java specific feature. Some languages doesn't have annotations afaik. A comment doesn't need to be explained with special syntax. IMHO.

Do we already have NetBeans specific annotations? Maybe I got this topic wrong.

jlahoda:
Well, embedding as such is not a Java-specific feature, and already works. If you have HTML like:
```
<!DOCTYPE html>
<html>
    <head>
        <script>
            console.log("Hello, World!");
        </script>
    </head>
    <body>
    </body>
</html>
```

what you get inside the `script` tag is embedded JavaScript. There are quite a few other embedding providers, many of them in the HTML/JavaScript area.

What cannot be done at this point is for anything to be embedded inside Java. And having Java embedded inside a bigger file is also extremely limited. (And having Java embedded inside Java is a bit of a minefield.)

There are many tricky parts here, but possibly the biggest one as follows. Consider this method:
https://github.com/apache/netbeans/blob/62a3770454446bcb421043f31f89f011e87ca1af/java/java.hints.test/src/org/netbeans/modules/java/hints/test/api/HintTest.java#L309
the content of the `code` parameter here is 100% Java. Ideally, I would prefer if, on use-sites, I didn't have to say anything to get the proper Java embedding. But to get that, the embedding can only be done after the file is parsed AND attributed, so that for a method invocation we know which method is being invoked. This is possibly the most tricky thing here.

A small part of this is that, after the code is parsed and attributed, and we need to somehow detect the language for the embedding. So we look at the annotations on the parameter. This is where the annotations are used, and they are very difficult to replace here - the source code may not be available at all, and even if it is, it would be very slow and hard to look into it. Annotations are immediately ready. Note the embedding provider is ignoring the package of the annotation, and only uses its simple name, so if needed, "anyone" can create their own `Language` annotation. There's also a positive synergy with the external annotations.

The newly-introduced Language annotation is only meant for code inside NetBeans - like from the abovementioned method. And we already have quite a few more:
https://github.com/apache/netbeans/tree/62a3770454446bcb421043f31f89f011e87ca1af/platform/api.annotations.common/src/org/netbeans/api/annotations/common

There are other heuristics included - like the variable name, and tracking how the value flows. A comment may be another heuristics (and, admittedly, the commend would not require parse&attribute). But, unless we decide that we want to abandon the embedding based on attribution altogether, the comments are not really  something urgent - there are bigger fish to fry.

(Overall, I would summarize it as: I would prefer if I wasn't forced to add tags to the source code or manually specify embeddings at use sites. It would be much better if could "tag" the declarations, and the rest would "simply" work.)

platform/api.annotations.common/src/org/netbeans/api/annotations/common/Language.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Allows to mark a specific variable so that the is expected to contain source
+ * code in a given language.
+ *
+ * @since 1.57
+ */
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE})
+@Retention(RetentionPolicy.CLASS) //TODO: can be SOURCE after PR 4882
+@Documented
+public @interface Language {
```
mbien, line 37: would this allow as future enhancement to annotate on the class level (for example) and all blocks inside the element scope would inherit the annotation? That way tests would only have to set a single annotation on top.

mbien:
```java
@Language("java")
public class Foo {

    public void fooTest() {
        test(
          """
          // this is recognized as java
          """
        );
    }
}
```

mbien: or does this work the other way around: if the `String` parameter of the `test()` method is annotated, NB will know that the text block of the example above is a java snippet?

Chris2011:
@jlahoda I know that we already have embeddings in NetBeans, but this is based on language standard featuures. The IDE knows script or style and knows this is JS or CSS and so on. But this is because the parser knows those tags are not artificial.

This example here in JS will break and will not run, if we will have this as an annotation:

```JavaScript
@Language("sql")
const sqlString = "SELECT * FROM users WHERE age > 30 AND city = 'New York'";

function foo(@Language("sql") sql) {
    return null
}
```

This is why i mean this should be a language independent feature if we want to that this should be handled by the IDE. For java there is maybe not a problem of custom annotations, but I also need to ask my self, when I use VS Code or any other IDE, what this annotation is for.

This will unfortunately not work for other languages than Java.

Chris2011:
At the end, I just wanted to say, that I like the feature but I prefer a more intuitive way of using that what we already have like comments across languages. I'm also not blocking this feature if an annotation is the preferred way,

PR #4891, opened 2022-10-30
hide template link as last resort if it appears to be broken.

mbien:
based on the discussion in #3678 we decided to try to hide the template link if it can't be resolved for now.

adding `all:tests` since this could potentially cause some String comparing tests to fail. Ran some locally but everything was green so far.

targets delivery


neilcsmith-net:
Looks good.  Nice way to do it.  Just waiting on irrelevant, retriggered Travis test and will merge.

PR #4899, opened 2022-11-01
Fix maven classpath listening

neilcsmith-net:
Rebase of @timboudreau #4868 onto delivery.  Would like to get into rc3 so needs to merge soon.  Didn't want to force push to Tim's branch without OK.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
+++ b/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
@@ -712,7 +712,8 @@ public final class ClassPathProviderImpl implements ClassPathProvider, ActiveJ2S
      * This selector chooses the annotation classpath, if it is not empty (has items, or is broken), or the regular
      * compile classpath if annotation path is empty. The selector reacts 
      */
-    private static class AnnotationPathSelector extends ClassPathSelector {
+    private static class AnnotationPathSelector extends ClassPathSelector
+            implements PropertyChangeListener {
         private final ClassPath annotationCP;
         private final Supplier<ClassPath> compileClassPath;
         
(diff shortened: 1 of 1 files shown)


java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
```diff
@@ -721,18 +722,14 @@ public AnnotationPathSelector(NbMavenProjectImpl proj, ClassPath anno, Supplier<
...
-                    }, ClassPath.PROP_ROOTS, anno
-            ));
-//            proj.getProjectWatcher().addPropertyChangeListener((e) -> {
-//                if (NbMavenProject.PROP_PROJECT.equals(e.getPropertyName())) {
-//                    active = null;
-//                    support.firePropertyChange(PROP_ACTIVE_CLASS_PATH, null, null);
-//                }
-//            });
+            anno.addPropertyChangeListener(WeakListeners.propertyChange(this, anno));
+            proj.getProjectWatcher().addPropertyChangeListener(WeakListeners.propertyChange(this, anno));
```
sdedic, line 726: Should be the 2nd parameter the event source, `proj.getProjectWatcher()` in this case ?

mbien: it was 'anno' before the change. Since this is a last minute fix, it's probably safer to not change this too, right?

mbien: let me check this in the ide in case i didn't count the brackets properly

neilcsmith-net: @mbien from the docs, the property name should be passed to both `WeakListeners::propertyChange` and `addPropertyChangeListener`. As not used for the latter, I assume this might have been firing for all properties but only trying to remove a listener on `ClassPath.PROP_ROOTS`?!

@sdedic is right, that second weak listener should be passed `proj::getProjectWatcher`, or `null`.  I think the underlying code can handle events from alternative sources though, so shouldn't cause a problem?

I (or one of you) can add a second commit here to tidy this up, but need a call ASAP if we're going to get in rc3.

cc @timboudreau

neilcsmith-net: @sdedic added a second commit that fixes the source passed to WeakListeners.  Other than that are you happy with @timboudreau changes going in to 16-rc3?

sdedic: Is there a reason to react on all propchanges on project, not only PROP_PROJECT (= reload of project file) ?

neilcsmith-net: @sdedic no idea. Justification for that is covered by Tim in description for #4868

mbien (approved):
still looks good to me :)

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
neilcsmith-net: @sdedic have merged to proceed with rc3.  If you have concerns about this reacting to too many changes, we can review before release vote.

PR #4900, opened 2022-11-02
Remove Abandoned Cache Dirs Automatically [Janitor]

lkishalmi:
This one adds an option to the Janitor to remove abandoned cache directories automatically.

Well, I've set the default true, as even this is a delete operation. it is just cache.  Usually it bothers me, as a user of `netbeans-dev` Snap package, which is updated weekly and leaves a cache directory behind each week.


matthiasblaesing (approved):
Looks sane to me and removing cache dirs should always be harmless.

mbien (approved):
clean code and clean caches!

I also think that its fine to auto remove caches which are older than 30 days. This hopefully shouldn't impact users which use multiple IDE versions at the same time. Maven index is getting big, good that NB has a janitor.

PR #4901, opened 2022-11-02
Recognize Scala class constructor

jtulach:
I am developing mixed [Java+Scala application](https://github.com/enso-org/enso/tree/330612119aba1cbcb310333ca5f06274a67d935b/engine/runtime/src/main) with the NetBeans IDE. I got debugging working, but I have problems to place a breakpoint on constructor of some classes. This is the (disassembled) source I see in the editor:
image
I can place a method breakpoint on it (as the triangle in the left editor gutter indicates). However the breakpoint is then later reported as misplaced.

I tracked this down to problems in recognizing constructor of such strangely named Scala classes like: `org.enso.compiler.core.IR$Name$MethodReference`. This PR extracts the _"constructor recognizing logic"_  into its own method and adds few tests to verify current behavior. Then it adds a test and a fix to make the code work for my usage case 3205d10d00ab9c1123b39e407dcaee7bb13c92cc.

### DIFF
--- a/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/breakpoints/MethodBreakpointImpl.java
+++ b/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/breakpoints/MethodBreakpointImpl.java
@@ -291,7 +291,33 @@ public class MethodBreakpointImpl extends ClassBasedBreakpoint {
         }
         return super.exec (event);
     }
-    
+
+    static String checkForConstructor(String typeName, String methodName) {
+        String constructorName = typeName;
+        int index = Math.max(constructorName.lastIndexOf('.'),
+                constructorName.lastIndexOf('$'));
+        if (index > 0) {
+            constructorName = constructorName.substring(index + 1);
+            if (typeName.charAt(index) == '$') {
+                // test for: ...$<digits only>$<name>
+                int i = index - 1;
+                while (i > 0 && Character.isDigit(typeName.charAt(i))) {
+                    i--;
+                }
+                if (typeName.charAt(i) == '$') {
+                    if (constructorName.equals(typeName.substring(i + 1, index) + methodName)) {
+                        methodName = constructorName; // Constructor
+                    }
+                }
+            }
+        }
+        if (methodName.equals(constructorName)) {
+            return "<init>"; // Constructor
+        } else {
+            return methodName;
+        }
+    }
+
     @Override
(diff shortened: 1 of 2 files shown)


java/debugger.jpda/nbproject/project.properties
```diff
@@ -23,7 +23,7 @@ javac.source=1.8
 javadoc.arch=${basedir}/arch.xml
 jpda.classes.dir=${build.dir}/jpda/classes/
 requires.nb.javac=true
-spec.version.base=1.127.0
+spec.version.base=1.128.0
```
neilcsmith-net, line 26: Please remove the commit incrementing the version if targeting delivery - this will conflict with changes in master.

neilcsmith-net (changes requested):
Version increment commit will need removing if intended for delivery.

This probably won't be reviewed in time for rc3, so will need a call whether this change needs to trigger an rc4, otherwise it won't make it in unless something else does.

@mbien is a label required to trigger that test?

mbien:
@neilcsmith-net  `java/debugger.jpda.<foo>` are in the "Java Modules" job which is always on since its a short job
https://github.com/apache/netbeans/actions/runs/3374691868/jobs/5607816502

however` java/debugger.jpda` tests are commented out. When I migrated I left everything commented out which was also commented out on travis.

mbien:
@neilcsmith-net we will see in about 30mins if the tests still work. I enabled them on the other PR. Will update this comment with results.

java/debugger.jpda/nbproject/project.properties
JaroslavTulach: Done. Good I have made that version increment as the last commit. Now it is gone.

jtulach: Done. Good I have made that version increment as the last commit. Now it is gone.

jtulach:
> This probably won't be reviewed in time for rc3, so will need a call whether this change needs to trigger an rc4

This is a low priority bugfix PR. No need for special RC. Integrate, if you have a chance. If not let's merge to `master`.

neilcsmith-net (commented):
Thanks @jtulach Will be kept in mind if we have an rc4.

mbien:
many tests fail, so we can't turn `java/debugger.jpda` on right now:
https://github.com/apache/netbeans/actions/runs/3379257947/jobs/5611704286

entlicher (approved):
Looks good.

PR #4906, opened 2022-11-02
Maven HintsPanel UI layout fixes and minor cleanup.

mbien:
- analog to #3472, this fixes UI component resize behavior of the Maven HintsPanel
 - description panel is now filling the space vertically like in the java version of this panel. Text is now wrapping properly.
 - the option panels had to be adjusted to prefer their minimum default size
 - minor language cleanup: diamonds, lambdas

targets delivery. But this is fairly low priority. Noticed those UI issues while writing the maven release version hint.

hints-before

hints-after

### DIFF
--- a/java/maven.hints/src/org/netbeans/modules/maven/hints/pom/HintsPanel.form
+++ b/java/maven.hints/src/org/netbeans/modules/maven/hints/pom/HintsPanel.form
@@ -54,9 +54,6 @@
   <SubComponents>
     <Container class="javax.swing.JSplitPane" name="jSplitPane1">
       <Properties>
-        <Property name="border" type="javax.swing.border.Border" editor="org.netbeans.modules.form.editors2.BorderEditor">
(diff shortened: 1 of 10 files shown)


sdedic:
Just to be sure I've not overlooked anything: no functional change, just layout & cleanup anonymous -> lambda ?

mbien:
> Just to be sure I've not overlooked anything: no functional change, just layout & cleanup anonymous -> lambda ?

right. this is primarily about UI layouts. The lamda cleanup is just a cleanup, no logic changed there.

mbien:
added another commit for the CSL HintsPanel. All three panels are now updated. They also have the same initial divider locations and look like the same component.

neilcsmith-net:
Weird thing is the issue isn't reproducible with a clean userdir until the options window is resized.

Makes sense, and as we're having another rc, let's merge.

PR #4907, opened 2022-11-03
Fixes for DialogNotifier in LSP

sdedic:
The main issue fixed is the binding between LSP `UIContext` and the **first client that initializes it** and then is stuck with it. The original code gave the 1st client the priority over the client that may have been present in the computational context `Lookup.getDefault()`. The changed code will use the current Lookup.getDefault() that is transferred through various RP.Tasks - this is already implemented in UIContext.find()

* fixes ADM report, so it uses fire-and-forge `notifyLater` instead `notify`. The LSP client may use modeless UI (vscode does) and the server thread is left blocked until the user dismisses the report (in vscode the user may just ignore it or hide it into status line without dismissing)
* minor fix for tests: the structure must gson-deserialize on JDK11, so default ctor + setters are needed


dbalek (approved):
Looks fine

ppisl (approved):

PR #4918, opened 2022-11-04
Fix mapping of dependencies described as key/value maps to the source

sdedic:
The implementation of `DependencyResult.getDeclarationRange` failed to find a dependency that was specified using a map, e.g.
```
implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.17.0'
```
parser was ok, but matching ignored such dependency text. Relevant tests for various dependency entry syntaxes were added.

Additional: a subtle bug in Oracle Cloud Vulnerability Audit feature was found that resulted in NPE in some cases, so a cached audit might not appear.

@MartinBalin might be also included in NB16, if it is not too late. Please decide, I'll eventually rebase.


MartinBalin (approved):

MartinBalin:
I will not add it to NB 16 at this moment as such syntax is rarely used. Can wait.

ppisl (approved):

PR #4919, opened 2022-11-04
Netbeans should be able to execute (it) test by maven in non-standard directory

Honza-cz:
I noticed sometime ago, that netbeans is not able to execute IT tests which were located in "src/it/java". This folder was added to maven by "build-helper-maven-plugin" as

> <plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>build-helper-maven-plugin</artifactId>
    <version>1.9.1</version>
    <executions>
        <execution>
            <id>add-test-source</id>
            <phase>generate-test-sources</phase>
            <goals>
                <goal>add-test-source</goal>
            </goals>
            <configuration>
                <sources>
                    <source>src/it/java</source>
                </sources>
            </configuration>
        </execution>
    </executions>
</plugin>

I tried to execute a test in that folder by netbeans, either by CTRL+F6 (or by mouse) but I noticed that netbeans is not able to correctly defined "-Dtest=" as:

`cd /.../repository; JAVA_HOME=/.../jdk-19.0.1 /.../maven/bin/mvn -Dtest=${packageClassName} surefire:test`

I searched in neatbeans source code and the root cause is, that the extra source folder is not consider as root one. It seems that only folders as /src/main/java and /src/test/java are supported. I was not able to find our how to add additional folder, but based on input FileObject, I was able to construct test names for
- 1 test file 
- n test files (even in different folder)
- package
as

After my changes, netbeans is able to resolve _packageClassName_:

`cd /.../repository; JAVA_HOME=/.../jdk-19.0.1 /.../maven/bin/mvn -Dtest=**.It3Test,some.it.pkg.It2Test surefire:test`

It is covered by UT for execution files individually (one as one liner + normal formatting) and by folder/package.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/execute/DefaultReplaceTokenProvider.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/DefaultReplaceTokenProvider.java
@@ -29,6 +29,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import static java.util.stream.Collectors.joining;
+import java.util.stream.Stream;
 import javax.swing.ActionMap;
 import org.netbeans.api.annotations.common.CheckForNull;
 import org.netbeans.api.java.project.JavaProjectConstants;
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
Can replicate the issue.  Not sure if the solution is correct.  When configuring the plugin I see the root listed in the project as `Generated Test Sources`. Therefore, should `addSelectedFiles(..)` also use the result of `prj.getGeneratedSourceRoots(..)`?

Another issue (although possibly due to file name in use?) is that this shouldn't be triggering `surefire:test` either but `failsafe:integration-test`.

Honza-cz:
You are right, the path to the extra test folder is present in `prj::getGeneratedSourceRoots`. It works fine for 1 test, 2 tests, only in case of package there is missing `.**` at the end  `-Dtest=a.b.c surefire:test`

About failsafe vs surefire plugin. I keep same maven config as we are using in our company, where IT tests are executed by surefire plugin. That's why I even noticed this.

PR #4922, opened 2022-11-05
try to figure out why building on 19 fails on gh.

mbien:
(it works locally)
more logging.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '11', '17', '19' ]
+        java: [ '19' ]
       fail-fast: false
     steps:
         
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -90,9 +90,15 @@ jobs:
           path: ~/.hgexternalcache
           key: ${{ runner.os }}-${{ hashFiles('*/external/binaries-list', '*/*/external/binaries-list') }}
           restore-keys: ${{ runner.os }}-
+        
+      - name: Setup Gradle properties
+        run: |
+          mkdir -p ~/.gradle
+          echo "org.gradle.java.home=$JAVA_HOME_11_X64" >> ~/.gradle/gradle.properties
+          cat ~/.gradle/gradle.properties
```
mbien, line 98: build is green.

gradle daemon didn't want to run on JDK 19 (`General error during conversion: Unsupported class file major version 63`)

looks like we can tell the gradle daemon which jdk to use. Similar like with mvnd.

Its interesting to me why this is working locally though. Since I never told it to use any JDK other than the one in JAVA_HOME.

Issue #4927, opened 2022-11-06
Maven EAR project fails to deploy on an app server (GlassFish)


OndroMih:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

When
* A Maven EAR project contains a WAR module
* The project uses maven-ear-plugin >= 3.0.0
* Deployment to GlassFish (I think the behavior is similar to any other app server)

Then:
* Deployment fails
* The exploded EAR generated by Netbeans in `target/gfdeploy` contains both an exploded directory and original JAR/WAR
* The exploded module directories in EAR don't match the expected path defined in `META-INF/application.xml`

![image](https://user-images.githubusercontent.com/2195988/200166534-3daed36f-3e8f-49d3-a2c7-12dab4f71ddb.png)

Expected:
* Deployment is successful
* The exploded EAR contains only epxloded module directories
* The names of the exploded module directories match what's defined in `application.xml` (match the names of the original JAR/WAR files)

### How to reproduce

Reproducer: 
* [ear-sample-maven-project.zip](https://github.com/apache/netbeans/files/9945637/ear-sample-maven-project.zip)
* Build the top level parent POM module
* Run the `ear-sample-ear` module with [Eclipse GlassFish 7.0.0-M9](https://github.com/eclipse-ee4j/glassfish/releases/tag/7.0.0-M9) or newer.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Ubuntu Linux 22.04.1 LTS

### JDK

openjdk version "17.0.4" (Zulu17.36+13-CA)

### Apache NetBeans packaging

Apache NetBeans Snap Package

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

PR #4929, opened 2022-11-06
[NETBEANS-4927] Maven EAR project fails to deploy on an app server

OndroMih:
Fixes https://github.com/apache/netbeans/issues/4927.

The strategy is to compute child module names based on the JAR and WAR files in the root directory of the exploded EAR instead of relying on the names computed from the Maven project configuration. Then the existing JAR and WAR files are not copied to the deployment package and exploded directories are created with names matching the existing JAR and WAR files. 

If there is any inconsistency (e.g. number of child modules detected from the exploded EAR is different to child modules detected from Maven project), the old behavior based on Maven configuration is preferred, just in case.

It's also possible to force the old behavior based on Maven configuration by setting the system property `glassfish.javaee..childModuleSource=mavenProject` (e.g. by `-J-Dglassfish.javaee...childModuleSource=mavenProject` in netbeans.conf or as a command line parameter for Netbeans). Setting `glassfish.javaee..childModuleSource=archive` will force the new behavior even if inconsistencies are detected.

### DIFF
--- a/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/impl/InitialServerFileDistributor.java
+++ b/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/impl/InitialServerFileDistributor.java
@@ -16,10 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
-
 package org.netbeans.modules.j2ee.deployment.impl;
-
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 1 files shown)


marcosvasquezc (approved):

matthiasblaesing:
@jGauravGupta @ehsavoie @juneau001 could one of you please review? Would be good to know whether or not this should be merged.

juneau001 (approved):
These modifications look good to me.  Thanks

matthiasblaesing:
@OndroMih thank you for your work, @juneau001 thank you for the review. Let's get this in.

Issue #4928, opened 2022-11-06
EAR deployment ignores a new property name introduced in Maven EAR plugin v3


OndroMih:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

Since version 3.0.0, the Maven EAR plugin renamed the `fileNameMapping` property to [outputFileNameMapping](https://maven.apache.org/plugins/maven-ear-plugin/ear-mojo.html#outputFileNameMapping). The old name is ignored but the EAR maven plugin.

Netbeans uses the `fileNameMapping` property of the EAR plugin to find out the name of the exploded module directories when deploying the EAR to an app server (e.g. Eclipse GlassFish). But it ignores the new name of the property if a new version of EAR plugin is used in the project.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Ubuntu 22.04.1 LTS

### JDK

openjdk version "17.0.4" (Zulu17.36+13-CA)

### Apache NetBeans packaging

Apache NetBeans Snap Package

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

PR #4930, opened 2022-11-06
[NETBEANS-4928] EAR deployment considers outputFileNameMapping Maven EAR plugin property

OndroMih:
If the old property isn't found, try the new one. They can't both be specified, the Maven EAR plugin 3+ will refuse the build if the old property is specified. Therefore it doesn't matter which property is looked up first.

Fixes https://github.com/apache/netbeans/issues/4928.


mbien (approved):
looks good. Thanks updating this.

Issue #4920, opened 2022-11-04
Build fails with UnknownHostException: www.w3.org


jgneff:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

The new file `jakartaee_9.xsd` in NetBeans 16 causes a direct DNS lookup during the build. No DNS lookups occurred in the builds for NetBeans 13, 14, or 15. The DNS request causes the build to fail when running behind a strict firewall that allows outbound traffic only through an HTTP or HTTPS proxy server.

The failure occurs while building the `nb.cluster.enterprise` modules. The build fails with the following messages:

```
     [echo] Building nb.cluster.enterprise modules
[...]

PR #4933, opened 2022-11-06
Import local schema to avoid DNS lookup in build

jgneff:
This pull request changes two files to fix [apache/netbeans#4920](https://github.com/apache/netbeans/issues/4920).

1. It modifies `jakartaee_9.xsd` to import the "xml:" namespace from a local copy of `xml.xsd` instead of importing it remotely from `http://www.w3.org/2001/xml.xsd`.
2. It replaces the local `xml.xsd` file with the latest [version dated 2009-01](https://www.w3.org/2009/01/xml.xsd) instead of the previous copy which seems to have been derived from the earliest [version dated 2001-03](https://www.w3.org/2001/03/xml.xsd).

Furthermore, I made the following changes to the `xml.xsd` file from the W3C:

* I removed the XML stylesheet because it causes a same-origin policy error in Firefox when viewing the file locally: "Error loading stylesheet: An unknown error has occurred (805303f4)."
* I added the current [W3C Software and Document Notice and License](https://www.w3.org/Consortium/Legal/copyright-software) as a comment formatted with an 80-character line length.
* I removed all trailing whitespace in the file.

How can I test this fix locally? I tried the following two test targets. The first one worked, but the second one failed.

```console
$ ant -quiet -Dmetabuild.branch=release160 -Dcluster.config=release commit-validation
$ ant -quiet -Dmetabuild.branch=release160 -Dcluster.config=release test-basic
```

Let's add this to NetBeans 16 only if someone far more knowledgeable than I am about JavaEE and Jakarta can review and approve it. Otherwise, I have a workaround and can wait for the fix in NetBeans 17.

### DIFF
--- a/enterprise/j2ee.dd/src/org/netbeans/modules/j2ee/dd/impl/resources/jakartaee_9.xsd
+++ b/enterprise/j2ee.dd/src/org/netbeans/modules/j2ee/dd/impl/resources/jakartaee_9.xsd
@@ -49,7 +49,7 @@
   </xsd:annotation>
 
   <xsd:import namespace="http://www.w3.org/XML/1998/namespace"
-              schemaLocation="http://www.w3.org/2001/xml.xsd"/>
+              schemaLocation="xml.xsd"/>
 
   <xsd:include schemaLocation="jakartaee_web_services_client_2_0.xsd"/>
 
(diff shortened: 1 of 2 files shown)


neilcsmith-net:
Requested reviews and we'll see if we have a 16-rc4.  Can you move the commit (rebase onto) delivery, so that it's possible to merge there if necessary.  Thanks!

juneau001 (approved):
Looks like a good remedy to prevent Maven from trying to obtain the xml.xsd from the schema source location.

mbien:
will we remember to do the same for the next xsd which is added? :)

jgneff:
> Actually, changed base is clean. Thanks!

Thanks, Neil. My local pull-request branch is also based off the *delivery* branch. I didn't notice that GitHub used the *master* branch as a base by default. Good to know.

PR #4817, opened 2022-10-19
CI: the final migration

mbien:
~~continues~~ finishes the great migration (#4431).

 * second attempt to migrate the groovy jobs (first: #4105). I have the suspicion that the tests can't run on anything > JDK 8. The test setup on travis is just testing the same thing three times in a matrix. This [comment](https://github.com/apache/netbeans/pull/2708#issuecomment-774009452) made a similar observation. -> going to leave the groovy tests on 8 for now.
 * migrated harness, ergonomics, nb modules and merged them into a single job
 * can't get the LSP tests to run so far (update:  #4975)
 * going to migrate the rest too since apache will end travis support end of this year

new labels to control jobs with:
 - `Groovy` runs the groovy job
 - `GraalVM` runs the GraalVM job - should this trigger on other labels too?
 - `LSP` runs the LSP (server) job, client is currently tested via the IDE job
 - `VSCode Extension` runs the VSCode job
 - `tests` enables a step which builds all tests, this will only run on PRs, not on the master branch
 - [list of all ci labels including the new ones](https://github.com/apache/netbeans/labels?q=%5Bci%5D)
 
Harness job runs with the `Platform` label - haven't found a better place to put it.

Each commit usually migrates a single job to make review easier.

The last commit adds some general improvements:
  - mac runners have ant installed, brew/cache isn't needed anymore.
  - added cache for platform-src build
  - bump JDK version of migrated jobs to 11 where possible
  - extract java distribution variable
  - increase pipeline artifact retention time from 1 to 2 days
  - added per-job timeouts (at least 2x of average runtime; 1h minimum)
  - always run the paperwork job
  - set continue-on-error for java.mx.project false in PRs
  - wrapped `platform/core.startup`, `java/java.editor` and `java.lsp.server` in retry script
  - `ci:dev-build` will now produce a dev-build zip artifact
  - setup all JDKs at the start of the job for consistent timestamp

NB can now also build on JDK 19:
 - build matrix is now 11, 17 and 19
 - gradle daemon will run on JDK 11

PR #4941, opened 2022-11-08
db.mysql: Bump connector/j version to prevent unittest failures ran against newer mysql server

matthiasblaesing:
https://github.com/apache/netbeans/pull/4817#issuecomment-1307814317


mbien (approved):
looks good - thanks

mbien:
everything is green, including the travis mysql job which means the new lib works fine -> merging so that I can update the other PR

Issue #4865, opened 2022-10-26
Gradle: can't debug composite build


errael:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

This issue discussed in dev,
https://lists.apache.org/thread/372jso34sfp42fotvfkhp2vj5r8vvq0h
The thread includes the suggestion to use a `composit build` with `includeBuild` and various suggestions on how to get debug to work. The problem is that although the `Run` button works on a composite build, the `Debug` button does not and results in the following error.

```
FAILURE: Build failed with an exception.

* What went wrong:
Problem configuring task :run from command line.
> Unknown command-line option '--debug-jvm'.
```

### How to reproduce

zip file has trivial projects, `lib, app, compose` that demo the issues:
https://github.com/apache/netbeans/files/9847401/gradle-projects-compose.zip

0.  clean userdir/cachedir
1.  jar -xf attached-zip-file; lib,app,compose; chmod +x */gradlew
2.  Open app,lib projects - trust yes, resolve problem
    Only lib can resolve, app depends on maven local
3.  lib:publishToMavenLocal
4.  app prime build, set as MainProject, run and see `=== Lib ===` output
6.  Modify Lib.java message (in lib project)
7.  Run app, note change to message not seen
8.  Open project compose, set compose as main project
9.  Run app, note change to message seen
10. Debug app
Observe:

```
FAILURE: Build failed with an exception.

* What went wrong:
Problem configuring task :run from command line.
> Unknown command-line option '--debug-jvm'.
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Linux harmony 5.19.16-76051916-generic #202210150742~1666053244~22.04~cf07008 SMP PREEMPT_DYNAMIC Tue O x86_64 x86_64 x86_64 GNU/Linux

### JDK

openjdk version "11.0.2" 2019-01-15 OpenJDK Runtime Environment 18.9 (build 11.0.2+9) OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

### Apache NetBeans packaging

Own source build

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

errael:
@neilcsmith-net asked
> Not sure if it's an issue or not looking at the list discussion?!

[...]

PR #4943, opened 2022-11-09
Prevent CCE NbGradleProjectImpl to DataObject

lkishalmi:
Playing with Ernies composite build in #4865 I've encountered  the following exception

```
java.lang.ClassCastException: class org.netbeans.modules.gradle.NbGradleProjectImpl cannot be cast to class org.openide.loaders.DataObject (org.netbeans.modules.gradle.NbGradleProjectImpl is in unnamed module of loader org.netbeans.StandardModule$OneModuleClassLoader @512bad4a; org.openide.loaders.DataObject is in unnamed module of loader org.netbeans.StandardModule$OneModuleClassLoader @657680c2)
	at org.netbeans.modules.editor.hints.HintsControllerImpl.setErrors(HintsControllerImpl.java:68)
	at org.netbeans.spi.editor.hints.HintsController$1.run(HintsController.java:59)
```
It seems some other NetBeans parts expecting a `DataObject` in a Document `Document.StreamDescriptionProperty` as Stream descriptor. I used that property ages ago probably seen that on some other editor. This PR just change the property name used for the code completion provider.

It seems if this exception does not happen, then the registration of the JDPA listener would g through, fixing #4865 entirely.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
@@ -60,6 +60,7 @@ import org.openide.util.Utilities;
  */
 public final class GradleCommandLine implements Serializable {
 
+    private static final String GRADLE_PROJECT_PROPERTY = "gradle-project"; //NOI18N
     private static final Logger LOGGER = Logger.getLogger(GradleCommandLine.class.getName());
     private static final String PROP_JVMARGS = "org.gradle.jvmargs"; // NOI18N
     /**
--- a/extide/gradle.editor/src/org/netbeans/modules/gradle/editor/cli/GradleCliCompletionProvider.java
+++ b/extide/gradle.editor/src/org/netbeans/modules/gradle/editor/cli/GradleCliCompletionProvider.java
@@ -68,6 +68,9 @@ public class GradleCliCompletionProvider implements CompletionProvider {
     private static final Pattern PROP_INPUT = Pattern.compile("\\$\\{([\\w.]*)$"); //NOI18N
     private static final String INPUT_TOKEN = "input:"; //NOI18N
     private static final Set<GradleCommandLine.GradleOptionItem> GRADLE_OPTIONS;
+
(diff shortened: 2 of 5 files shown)


mbien:
gh might be having issues since I can't label this PR (or other PRs). Unfortunately this means the Gradle tests won't run.

mbien (approved):
makes sense + looks like a fairly low risk change

lkishalmi:
> gh might be having issues since I can't label this PR (or other PRs). Unfortunately this means the Gradle tests won't run.
> 
> nevermind. was a temporary issue. Restarted the build.

Oh, I knew I forgot something! I'm sorry.

sdedic:
> 
> It seems some other NetBeans parts expecting a `DataObject` in a Document `Document.StreamDescriptionProperty` as Stream descriptor. I used that property ages ago probably seen that on some other editor. This PR just change the property

Yes,there's even an editor API for that.

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
```diff
@@ -60,6 +60,7 @@
  */
 public final class GradleCommandLine implements Serializable {
 
+    private static final String GRADLE_PROJECT_PROPERTY = "gradle-project"; //NOI18N
```
sdedic, line 63: Leftover ?
But maybe declare the property as `public` somewhere to have a shared constant.

sdedic (approved):
Seems that all occurrences are replaced.

extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleCommandLine.java
lkishalmi: Kind of. Did not want to change the API so late in NB16 cycle, so marked that private and made the two now public non apu one with a comment. Will pick that up for NB17

neilcsmith-net:
Great, thanks @lkishalmi .  I've seen that exception a couple of times recently testing 16, but could never reproduce what triggered it.  Should have opened an issue with it though.

PR #4947, opened 2022-11-09
[NETBEANS-6519] Maven dependency produces full tree with duplicates, avoids cycles.

sdedic:
See also [NETBEANS-6519](https://issues.apache.org/jira/browse/NETBEANS-6519)

The main issue with the implementation was that the 1st encountered Dependency from Maven might be truncated / stripped of its children - and marked as a duplicate. However in the dependency API all dependencies are "equal", we cannot assume that a client does depth-first or breadth-first traversal.
With the fix, the dependency tree will be enormous, as each common library dependency will introduce a full subtree at the library's place in the dependency tree.

I plan to add an additional API to mark duplicates (= subtrees with the same contents) and possibly resolution conflicts (an artifact dependency has been resolved by a different artifact, e.g. a different version of a library). But this fix is simple and allows clients to enumerate all dependencies. Clients can now filter the duplicates themselves, according to their needs.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/queries/MavenDependenciesImplementation.java
+++ b/java/maven/src/org/netbeans/modules/maven/queries/MavenDependenciesImplementation.java
@@ -207,7 +207,7 @@ public class MavenDependenciesImplementation implements ProjectDependenciesImple
         };
         
         return new MavenDependencyResult(nbMavenProject.getMavenProject(), 
-                convert2(n, compositeFiter, broken), new ArrayList<>(scopes), broken, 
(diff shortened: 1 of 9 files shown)


sdedic:
sorted dependency in the printout on each level. updated golden files.

sdedic:
Finally got the test golden file right -- local machine setup implied somehow activation of `graalvm` profile, controlled by `micronaut-parent` POM. 
The most tricky part was to disable Maven's global cache - NetBeans do the same at runtime.

dbalek (approved):

PR #4948, opened 2022-11-09
enable "sort document list by project" by default.

mbien:
as discussed on the [dev list](https://lists.apache.org/thread/g9w0038rdsksony6xyl4q42vnj4jrjk4), this proposes to tweak the default setting of the documents list dropdown of the editor.

option can be changed back in the main options window:
`tools -> options -> appearance -> document tabs -> "Sort opened documents list by default" checkbox`

This won't influence already made user settings, only the out-of-the-box default.

tab-list

This area got some fixes #3299 and pending improvements recently.#4792


eirikbakke:
This affects the appearance of tabs, not just the document list dropdown. Probably the UI tweaks that were done in https://github.com/apache/netbeans/pull/4286 should be adapted to the "Sort opened document list by project" mode as well before making this the default.

sortdoc

neilcsmith-net:
IIRC it uses the multi-row tab renderer in this case, and code for those needs looking at to work properly with #4286 

So, sorry, -1 to this change until we can fix that issue.

mbien:
converted it back to draft so that nobody can merge it by accident.

neilcsmith-net:
I think the problematic code is here - https://github.com/apache/netbeans/blob/master/platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java#L234  And I think it's currently not LAF-specific, and the position changes depending on where the tabs are located. Still, maybe as quick fix could just switch "under"line position for top and bottom tab positions?

neilcsmith-net:
And this seems to cover all the settings where they take over - https://github.com/apache/netbeans/blob/master/platform/core.multitabs/src/org/netbeans/core/multitabs/prefs/SettingsImpl.java#L57

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
```diff
@@ -238,10 +238,10 @@ public void paint( Graphics g ) {
                 switch (tabsLocation) {
                     default:
                     case JTabbedPane.TOP:
-                        g.fillRect(0, rect.height - underlineHeight, rect.width, underlineHeight);
+                        g.fillRect(0, 0, rect.width, underlineHeight);
                         break;
                     case JTabbedPane.BOTTOM:
-                        g.fillRect(0, 0, rect.width, underlineHeight);
+                        g.fillRect(0, rect.height - underlineHeight, rect.width, underlineHeight);
                         break;
```
mbien, line 245: @neilcsmith-net @eirikbakke  forgot to mention that I did swap the case ops as instructed.

It seems to have the desired effect.

neilcsmith-net: @mbien thanks, but thinking about this more, it would be better to swap the underline position based on the value of `EditorTab.underlineAtTop` as used at https://github.com/apache/netbeans/blob/master/platform/o.n.swing.laf.flatlaf/src/org/netbeans/swing/laf/flatlaf/ui/FlatEditorTabCellRenderer.java#L68 ?  That should keep it consistent.

For same reason, I'd probably swap the horizontal tabs line position on the same flag?

neilcsmith-net (changes requested):

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
mbien: roger. I *think* i know what you mean. Going to take another look.

eirikbakke:
I don't think multitabs is ready to become the default tab renderer implementation yet. There were a lot of improvements to the FlatLAF renderer that should be ported over to multitabs first:

* https://github.com/apache/netbeans/pull/4286 (besides the underlineAtTop part)
* https://github.com/apache/netbeans/pull/4335
* https://github.com/apache/netbeans/pull/4349

(Example adjustments that are missing on multitabs: left editor border on 150% HiDPI scaling, lighter text color on non-active tabs, vertical alignment of text and X buttons against Project/Files/Services sidebar and the document tab area, adjustments to horizontal margins around text and X buttons etc.)

I can take a whack at this at some point, though it may take me a while to get around to it...

mbien:
@eirikbakke ok, I convert this to draft so that nobody merges it by accident. 

If anyone wants to close this completely feel free to do so.

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
mbien: implemented it finally. When the tab is on the bottom it is rendered on the bottom if `underlineAtTop` is set since everything is bottom up.

mbien:
added dev build label so that it is easier to test, build is linked on the build summary page for 7 days.

@eirikbakke since you did the tweaks to the regular tabs, would you like to do them for those tabs here too?

platform/core.multitabs/src/org/netbeans/core/multitabs/impl/TabDataRenderer.java
mbien: I didn't touch left/right since the regular tabs don't change the line positioning there

eirikbakke:
@mbien I started looking at it. Applying the past improvements to multitabs is going to be a somewhat bigger project, as it's a completely different implementation/code path with more generic functionality (multiple rows, group by project, tabs on left/right/bottom etc.). I might get back to it later.

For reference, the logic for when multitabs is used is here:
[org.netbeans.core.multitabs.prefs.SettingsImpl.isEnabled](https://github.com/apache/netbeans/blob/4b7776a679963cd2d0797fd64fa11f8decda3c32/platform/core.multitabs/src/org/netbeans/core/multitabs/prefs/SettingsImpl.java#L57)

mbien:
I was wondering about that before: if multitabs can do everything what o.n.swing.tabcontrol can do, why are there still two implementations in use?

eirikbakke:
@mbien I don't know the history of the two implementations, but it seems like the regular tabcontrol implementation was much easier to customize for skinning purposes, so that's the implementation that was customized in all the old LAFs.

If at some point we get multitabs to support all the FlatLAF options that now exist for regular tabs, we could probably switch to multitabs in all cases for FlatLAF, and effectively deprecate the other one. (Unless multitabs turns out to have problems we aren't currently aware of.)

neilcsmith-net:
> If at some point we get multitabs to support all the FlatLAF options that now exist for regular tabs, we could probably switch to multitabs in all cases for FlatLAF, and effectively deprecate the other one. (Unless multitabs turns out to have problems we aren't currently aware of.)

Switching to multitabs in all cases within the IDE might be good.  But we need to be careful that it's an optional part of the platform so make sure we don't do anything that "breaks" cases where it's not included.

I'm not sure if we can make multitabs delegate better to the laf tabs so that most fixes can just go in there?

eirikbakke:
@neilcsmith-net Yeah, I think multitabs could only ever become the default for FlatLAF specifically. Regular tabs could be "deprecated" only in the sense that LAFs other than FlatLAF would be "deprecated".

> I'm not sure if we can make multitabs delegate better to the laf tabs so that most fixes can just go in there?

I suspect this might be difficult, and perhaps not necessary. If multitabs can be skinned via UIDefaults properties, like FlatLAF, then all future skinning of the tab components could be done this way, rather than by changing LAF-specific code.

I also wonder to what degree we hope to continue maintaining the old non-flatLAF LAFs. (e.g. GTK LAF, Windows LAF, MacOS/Aqua LAF). Myself, I've completely moved over to FlatLAF.

mbien:
moved to NB19

mbien:
removed milestone

mbien:
reading through old PRs to figure our what still needs to be done:

> Switching to multitabs in all cases within the IDE might be good. But we need to be careful that it's an optional part of the platform so make sure we don't do anything that "breaks" cases where it's not included.

we could probably set the default via branding so that it switches only for the IDE? Anything else what needs to be done here still before changing the default is possible?

PR #4950, opened 2022-11-09
Fix GradleCliEditorKit Registration

lkishalmi:
Neil, you are correct and right in #4946, though I'd address the issue using `@MimeRegistration` so that moves with the code in the future.


lkishalmi:
The generated value is:
```xml
<file name="org-netbeans-modules-gradle-editor-cli-GradleCliEditorKit.instance">
    <!--org.netbeans.modules.gradle.editor.cli.GradleCliEditorKit-->
    <attr name="instanceOf" stringvalue="javax.swing.text.EditorKit"/>
</file>
```
That's not the same, but seems to be enough for the IDE to pick the EditorKit up, the code completion is back.

sdedic (approved):

PR #4951, opened 2022-11-09
tweaks some versioning history settings.

mbien:
as discussed on the [dev list](https://lists.apache.org/thread/g9w0038rdsksony6xyl4q42vnj4jrjk4), this proposes some tweaks to the versioning UI

 - `git -> show history` will start with 50 entries loaded by default
 - file history will show the full history by default
 - search button will only show up if the advanced options are expanded since:
   - everything else searches at realtime
   - results will be shown on first open without the extra click since #3294
   - which means the button is only useful when the advanced options are expanded


matthiasblaesing (approved):
Looks sane to me

PR #4954, opened 2022-11-09
hide maven transfer progress via flag instead of regexp filter.

### LINKED ISSUE #4017: ASCII/UNICODE BUG K=75
### Apache NetBeans version

verified in 14rc4, 13... 12.0

### What happened

```java
for (int i=0; i<=127; i++){
    System.out.println(i + " " + (char) i);
} 
```
I was running this code above and received the following output:
(I was missing the value 75 of uppercase K) (Netbeans IDE version 12.6)
bugasiiunicode

### How to reproduce

see above

No

### Operating System

Windows

### JDK

17.0.1

### Apache NetBeans packaging

### Anything else

_No response_

Yes

### Code of Conduct

Yes

mbien:
- sets `--no-transfer-progress` as default global maven option
 - removes console regexp filter which fixes #4017
 - regexp filter didn't work with current maven anyway


lkishalmi (approved):
Cool!

Issue #4725, opened 2022-10-03
PHP 8.2 Support


junichi11:
### Description

Support for PHP 8.2

https://wiki.php.net/rfc#php_82

- [x] https://wiki.php.net/rfc/null-false-standalone-types
(#4888)
- [x] https://wiki.php.net/rfc/true-type
(#4888)
- [x] https://wiki.php.net/rfc/readonly_classes
(#4955)
- [x] https://wiki.php.net/rfc/constants_in_traits
(#5302)
- [x] https://wiki.php.net/rfc/fetch_property_in_const_expressions
(#4966)
- [x] https://wiki.php.net/rfc/dnf_types
(#5302 )

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
@junichi11 this issue looks done to me, should it be closed?

junichi11:
@mbien Will change to `NB20` because DNF types support is incomplete yet :) Thanks!

neilcsmith-net:
@junichi11 can we have a status update? Is this complete for NB20 or something that needs moving to NB21?

PR #4955, opened 2022-11-10
PHP 8.2 Support: Readonly classes

junichi11:
#4725
- https://wiki.php.net/rfc/readonly_classes

#### Part 1

- Fix the grammar file
- Add `PHP82UnhandledError`
- Add/Fix unit tests for the parser, the lexter, and the index

#### Part 2

- Fix hints for readonly classes
- Add the code template for a readonly class
- Add unit tests for hints, formatter, and CC

nb-php82-readonly-classes-syntax-and-hints

nb-php82-readonly-classes-syntax-and-hints2


junichi11:
@tmysik Could you please have a look at this when you have time? If there is no problem, let's merge it :) Thanks!

junichi11:
@tmysik Will increment the spec version.

php/php.editor/src/org/netbeans/modules/php/editor/model/nodes/ClassDeclarationInfo.java
```diff
@@ -93,14 +94,24 @@ public Set<QualifiedName> getInterfaceNames() {
...
+                    phpModifiers.add(PhpModifiers.ABSTRACT);
+                    break;
+                case FINAL:
+                    phpModifiers.add(PhpModifiers.FINAL);
+                    break;
+                case READONLY:
+                    phpModifiers.add(PhpModifiers.READONLY);
+                    break;
+                default:
+                    break;
```
tmysik, line 111: Don't we want to log/throw here?

tmysik (approved):

php/php.editor/src/org/netbeans/modules/php/editor/model/nodes/ClassDeclarationInfo.java
junichi11: Will add the following. Thanks!
```java
                case NONE:
                    // no-op
                    break;
                default:
                    assert false : "Handle " + modifier + " modifier";
                    break;
```

junichi11:
@tmysik Thank you for your review! I'll fix it with the next PR.

php/php.editor/src/org/netbeans/modules/php/editor/model/nodes/ClassDeclarationInfo.java
tmysik: 👍

tmysik:
@junichi11 Nothing to thank for, really. Plenty of tests really help us here. Thanks for your great work!

PR #4958, opened 2022-11-10
use grep -E instead of egrep

mbien:
gets rid of this warning:
`egrep: warning: egrep is obsolescent; using grep -E`

when using NetBeans from CLI eg.
`./netbeans --help`


lkishalmi (approved):
Looks good to me.

PR #4966, opened 2022-11-15
PHP 8.2 Support: Fetch properties of enums in const expressions

junichi11:
#### PHP 8.1 Support: New in initializers (Part 5) 

- https://issues.apache.org/jira/browse/NETBEANS-5599
- RFC:https://wiki.php.net/rfc/new_in_initializers
- Support for field and method access for constant (e.g. `const C = new A(); C->field; C->method();`)
- Fix the parser
- Fix PHP81UnhandledError

nb-php81-new-in-initializers-const-access

#### PHP 8.2 Support: Fetch properties of enums in const expressions (Part 1) 

- #4725
- https://wiki.php.net/rfc/fetch_property_in_const_expressions
- Fix the grammar file
- Fix CC: add the readonly name property to enums
- Add/Fix unit tests for the parser and CC

nb-php82-fetch-properties-of-enum-in-const-expressions1
nb-php82-fetch-properties-of-enum-in-const-expressions2

##### Note

The following code is valid but I don't think these are meaningful. So, I don't support these cases at the moment.

```php
const A = (null)?->test;

const B = (null)?->test->test;

const C = (null)->test?->test;

const D = (null)?->test['test'];

const E = (null)['test']?->test;

const F = (null)?->{new Printer};

const G = (null)?->test + (new Printer ? 1 : 0);

const H = (1)->prop;

const I = (1)?->prop;
````

### CODE REVIEW

File: php/php.editor/src/org/netbeans/modules/php/editor/model/nodes/ClassDeclarationInfo.java [hunk 1]
```diff
@@ -107,7 +107,11 @@ public PhpModifiers getAccessModifiers() {
                 case READONLY:
                     phpModifiers.add(PhpModifiers.READONLY);
                     break;
+                case NONE:
+                    // no-op
+                    break;
                 default:
+                    assert false : "Handle " + modifier + " modifier"; // NOI18N
```
junichi11, line 114: https://github.com/apache/netbeans/pull/4955#pullrequestreview-1176088975

junichi11:

php/php.editor/src/org/netbeans/modules/php/editor/model/nodes/ClassDeclarationInfo.java
tmysik: Thanks! 👍

tmysik (approved):
Again, great job! And thanks for all the unit tests.

junichi11:
@tmysik Thank you for your review and your time :)

tmysik:
@junichi11 Nothing to thank for, really.

PR #4971, opened 2022-11-16
use maven central search service for class name search.

mbien:
wip prototype to play around with.

how to test:
open/create maven project, type some class name, use the hint to do a maven search. Results should contain both central and local entries.

The maven central index does not contain class information anymore. As workaround we make a http request and merge the results with the local index.

augmented-central-search

performance is pretty bad compared to the local index, but this is to be expected from a web service which wasn't really made for this purpose:
```
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:logger with 234 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=223981, numDocs=200, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 4101
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:jfr with 50 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=47, numDocs=47, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 339
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:jfrlogger with 2 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=6, numDocs=6, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 1247
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:jfr with 50 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=47, numDocs=47, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 217
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:jfrlogger with 2 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=6, numDocs=6, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 463
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: passed on 1,024 clauses processing classnames:util with 1,203 hits
INFO [org.netbeans.modules.maven.centralsearch.MavenCentralClassesQuery]: Response{numFound=329503, numDocs=200, start=0}
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: findVersionsByClass time = 2320
```
It does some caching though, so repeated requests are quite fast.

### DIFF
--- a/java/maven.indexer/external/binaries-list
+++ b/java/maven.indexer/external/binaries-list
@@ -20,3 +20,6 @@
 5A44DF2CB26FA1E0E64BE53FE474C7F1D5A3E634 org.apache.lucene:lucene-queryparser:8.11.1
 E50AF506F271A3F7246DA054A2569B42FF73ABB2 org.apache.lucene:lucene-analyzers-common:8.11.1
 479C1E06DB31C432330183F5CAE684163F186146 javax.annotation:javax.annotation-api:1.2
+858C6CC78E1F08A885B1613E1D817C829DF70A6E com.fasterxml.jackson.core:jackson-annotations:2.13.4
+0CF934C681294B97EF6D80082FAEEFBE1EDADF56 com.fasterxml.jackson.core:jackson-core:2.13.4
+325C06BDFEB628CFB80EBAAF1A26CC1EB558A585 com.fasterxml.jackson.core:jackson-databind:2.13.4.2
--- a/java/maven.indexer/manifest.mf
+++ b/java/maven.indexer/manifest.mf
@@ -3,4 +3,4 @@ OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/maven/indexer/Bundle.prop
 AutoUpdate-Show-In-Client: false
 OpenIDE-Module-Specification-Version: 2.60
 OpenIDE-Module: org.netbeans.modules.maven.indexer/2
-
+OpenIDE-Module-Java-Dependencies: Java > 11
--- a/java/maven.indexer/nbproject/project.properties
+++ b/java/maven.indexer/nbproject/project.properties
@@ -17,7 +17,8 @@
 
 test.config.stableBTD.includes=**/*Test.class
 is.autoload=true
-javac.source=1.8
+javac.source=11
+javac.target=11
 javac.compilerargs=-Xlint -Xlint:-serial
 release.external/indexer-core-6.2.2.jar=modules/ext/maven/indexer-core-6.2.2.jar
 release.external/lucene-core-8.11.1.jar=modules/ext/maven/lucene-core-8.11.1.jar
@@ -25,6 +26,9 @@ release.external/lucene-highlighter-8.11.1.jar=modules/ext/maven/lucene-highligh
 release.external/lucene-queryparser-8.11.1.jar=modules/ext/maven/lucene-queryparser-8.11.1.jar
 release.external/lucene-analyzers-common-8.11.1.jar=modules/ext/maven/lucene-analyzers-common-8.11.1.jar
 release.external/javax.annotation-api-1.2.jar=modules/ext/maven/javax.annotation-api-1.2.jar
+release.external/jackson-annotations-2.13.4.jar=modules/ext/maven/jackson-annotations-2.13.4.jar
+release.external/jackson-core-2.13.4.jar=modules/ext/maven/jackson-core-2.13.4.jar
+release.external/jackson-databind-2.13.4.2.jar=modules/ext/maven/jackson-databind-2.13.4.2.jar
 # XXX so long as Lucene is bundled here:
 sigtest.gen.fail.on.error=false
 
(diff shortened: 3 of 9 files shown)


ebarboni:
@mbien we got informal slack discussion but do you have pointer why/when the removal of classname from central indexer ?

neilcsmith-net:
Nice! :smile: I've been wanting to look at using central search service for ages (for NB and a platform app).  I assume you've seen the search plugin?  https://github.com/emilianbold/maven.search.remote

Using JDK HTTP is great, but I wonder if requirement for JDK 11 will cause any concern?  This module is not excluded from VSCode I think?  Perhaps can be in a separate module?

Could this use pre-existing JSON libs in the IDE (like the plugin does)?

mbien:
> we got informal slack discussion but do you have pointer why/when the removal of classname from central indexer ?

@ebarboni I don't know. However it was a long time ago. I would say its years not months. I actually completely forgot about it and debugged deep into lucene code to figure out why we don't get any results :). Then I remembered reading [this](https://github.com/apache/maven-indexer/blob/6a0c6017794dca6ee1af151f7914ab3cd0e4aa25/indexer-examples/indexer-examples-basic/src/main/java/org/apache/maven/index/examples/BasicUsageExample.java#L274).


> Nice! smile I've been wanting to look at using central search service for ages (for NB and a platform app). I assume you've seen the search plugin? https://github.com/emilianbold/maven.search.remote

@neilcsmith-net I remember you mentioning it before but only looked at it now. My first attempt was similar, I did try to implement all queries of `RepositoryIndexQueryProvider` and put it into a module, but it showed that the search service is purely for artifacts and fairly limited. A prefix group ID search wouldn't be easily possible for example. The search is unfortunately not very well documented. I looked through source code to see what it is actually returning.

(btw the search can be also really slow. I sometimes got 10s response time in tests - it does cache though, so repeating the search is faster)

> 
> Using JDK HTTP is great, but I wonder if requirement for JDK 11 will cause any concern? This module is not excluded from VSCode I think? Perhaps can be in a separate module?

Well yes. This is the remainder of the first attempt to implement it, it was a separate optional module. We can borrow a bit time to use a different jdk 8 compatible client. But keep in mind that maven-indexer and lucene requires now JDK 11 too (maven-indexer has now a [nice multi threaded extraction implementation](https://github.com/apache/maven-indexer/pull/255)). So its not a if, its a when at this point. (we already backported a [whole class](https://github.com/apache/netbeans/blob/4a697ab128d7df8885a8479b5cc4212e6a56835e/java/maven.indexer/src/org/apache/lucene/search/MultiTermQuery.java#L334) due to a bugfix, this already is a fragile hack. If something would load lucene before we do, this will break)

So when will the VSCode extension switch to JDK 11?

> 
> Could this use pre-existing JSON libs in the IDE (like the plugin does)?

sure. Jackson (or any jaxb json equivalent) is just really convenient in this case and produces maintainable code. Anyone who looks at the POJOs knows what is going on there - if the service changes it can be quickly fixed. Streaming APIs require a few more brain cycles for maintainers :)

This is just an unfinished experiment to start a conversation. The method `find(final List<QueryField> fields, List<RepositoryInfo> repos)` is going to need a special case too if a class name appears.

`NexusRepositoryIndexerImpl` does too much right now IMO. I think we should check if we can split it into indexer and query provider, this might make it easier to work with.

matthiasblaesing:
> So when will the VSCode extension switch to JDK 11?

Why is VSCode tied to JDK 11? I can understand some of the concerns about the platform modules, but for the IDE I don't seen an argument. At least I would not think twice about creating a new JDK11 only module.

mbien:
> > So when will the VSCode extension switch to JDK 11?
> 
> Why is VSCode tied to JDK 11? I can understand some of the concerns about the platform modules, but for the IDE I don't seen an argument. At least I would not think twice about creating a new JDK11 only module.

This is in `java/maven.indexer` right now and only substitutes the class name query maven-indexer can't provide anymore. VSCode Extension is probably using this modules for maven support I suppose. As long the extension wants to run on 8, we can't bump the language level on this module. However, as previously mentioned, we can't delay that forever for reasons other than HTTPClient (there are enough JDK 8 replacements for it, even though its yet another extra step, is is really no big deal to swap this out).

Since this is a prototype to test things out, I didn't bother to add another dependency, I simply used JDK 11 HTTPClient.

mbien:
only a sync with minor changes

mbien:
closing in favor of #5984

PR #4921, opened 2022-11-05
temp test setup: try to migrate the LSP job

mbien:
this is just a test setup which tries to run the LSP job on github without running anything else.

not intended to be merged, only for debugging.

PR #4975, opened 2022-11-16
Test bugfixes: preparation for GH action migration

sdedic:
A follow-up to #4921. 

The most important fix is a fix to `nbjavac` task that prevents classes like `$JsCallback$` and similar to be deleted just before an incremental compilation because they are considered inner classes (without a matching outer class).

I've added additional logging to both server and client, because some thrown errors are wrapped into uninformative "Internal error". Added possibility to log-trace LSP messages (enabled by System property)

I've cleaned up the test a little:
- the exception about missing Groovy support turned into a WARNING line, if the client demands Groovy enabled, but it is simply not available. Other Reflective exceptions show stacktrace.
- fixed error that manifested by a print about 'not normalized' license template filename/path
- suppressed SocketClosed exceptions in test output
- fixed potential CNFE when running on JDK11: Truffle could not access `sun.misc.Unsafe` during script engine initialization

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
@@ -226,12 +226,16 @@ public class CustomJavac extends Javac {
                 continue;
             }
             int i = clazz.indexOf('$');
-            File enclosing = new File(d, clazz.substring(0, i) + ".class");
-            if (!enclosing.isFile()) {
-                File enclosed = new File(d, clazz);
-                log(clazz + " will be deleted since " + enclosing.getName() + " is missing", Project.MSG_VERBOSE);
-                if (!enclosed.delete()) {
-                    throw new BuildException("could not delete " + enclosed, getLocation());
+            // ignore filenames that start right with '$' (separatorChar preceded), these could not be inner classes.
+            if (i > 0 && clazz.charAt(i - 1) != File.separatorChar) {
+                File enclosing = new File(d, clazz.substring(0, i) + ".class");
+                // no inner class' filename may begin directly with '$', it must be preceded by an outer class' name.
+                if (!enclosing.isFile()) {
+                    File enclosed = new File(d, clazz);
+                    log(clazz + " will be deleted since " + enclosing.getName() + " is missing", Project.MSG_VERBOSE);
+                    if (!enclosed.delete()) {
+                        throw new BuildException("could not delete " + enclosed, getLocation());
+                    }
                 }
             }
         }
--- a/nbbuild/templates/projectized.xml
+++ b/nbbuild/templates/projectized.xml
@@ -123,7 +123,7 @@
         </condition>
     </target>
     <target name="-init-bootclasspath-prepend-run9" depends="-init-bootclasspath-prepend-compile" if="have-jdk-1.9">
-        <condition property="test.bootclasspath.prepend.args" value="--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming,jdk.jdi">
+        <condition property="test.bootclasspath.prepend.args" value="--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming,jdk.jdi,jdk.unsupported">
             <and>
                 <istrue value="${requires.nb.javac}"/>
                 <not>
--- a/ide/projectuiapi/src/org/netbeans/modules/project/uiapi/ProjectTemplateAttributesProvider.java
+++ b/ide/projectuiapi/src/org/netbeans/modules/project/uiapi/ProjectTemplateAttributesProvider.java
@@ -155,7 +155,10 @@ public final class ProjectTemplateAttributesProvider implements CreateFromTempla
             // in order to get through the freemarker, the path needs to "absolute" in freemarker terms - http://freemarker.sourceforge.net/docs/ref_directive_include.html
             // relative would mean relative to the template and we cannot be sure what the path from template to license template is..
             // it used to be, ../Licenses/ or ../../Licenses but can be anything similar, just based on where the template resides.
-            map.put(ATTR_LICENSE_PATH, "/" + url);
+            // Note: ensure reentrancy, so if 'url' starts with slash, do not prepend
+            if (!url.startsWith("/")) { // NOI18N
+                map.put(ATTR_LICENSE_PATH, "/" + url); // NOI18N
+            }
             //appears to cover both the new and old default value of the include path
         }  
         if (map.get(ATTR_ENCODING) == null) {
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ConnectionSpec.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/ConnectionSpec.java
@@ -25,6 +25,7 @@ import java.io.OutputStream;
 import java.net.Inet4Address;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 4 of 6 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
```diff
@@ -226,12 +226,16 @@ private void cleanUpDependDebris() {
...
-                log(clazz + " will be deleted since " + enclosing.getName() + " is missing", Project.MSG_VERBOSE);
-                if (!enclosed.delete()) {
-                    throw new BuildException("could not delete " + enclosed, getLocation());
+            // ignore filenames that start right with '$' (separatorChar preceded), these could not be inner classes.
+            if (i > 0 && clazz.charAt(i - 1) != File.separatorChar) {
+                File enclosing = new File(d, clazz.substring(0, i) + ".class");
+                // no inner class' filename may begin directly with '$', it must be preceded by an outer class' name.
+                if (!enclosing.isFile()) {
+                    File enclosed = new File(d, clazz);
+                    log(clazz + " will be deleted since " + enclosing.getName() + " is missing", Project.MSG_VERBOSE);
```
lbownik, line 235: since "verbose" level is almost always off, putting a guarding if statement will almost always avoid string concatenation of this message.

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
```diff
@@ -226,12 +226,16 @@ private void cleanUpDependDebris() {
...
+                    if (!enclosed.delete()) {
+                        throw new BuildException("could not delete " + enclosed, getLocation());
```
lbownik, line 237: "could not delete ".concat(enclosed.toString()) will avoid one allocation and one loop

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
mbien: this is a NB internal build tool. There are not really performance concerns here. Readability > efficiency in this case.

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
```diff
@@ -226,12 +226,16 @@ private void cleanUpDependDebris() {
...
             int i = clazz.indexOf('$');
-            File enclosing = new File(d, clazz.substring(0, i) + ".class");
-            if (!enclosing.isFile()) {
-                File enclosed = new File(d, clazz);
-                log(clazz + " will be deleted since " + enclosing.getName() + " is missing", Project.MSG_VERBOSE);
-                if (!enclosed.delete()) {
-                    throw new BuildException("could not delete " + enclosed, getLocation());
+            // ignore filenames that start right with '$' (separatorChar preceded), these could not be inner classes.
+            if (i > 0 && clazz.charAt(i - 1) != File.separatorChar) {
+                File enclosing = new File(d, clazz.substring(0, i) + ".class");
```
lbownik, line 231: clazz.substring(0, i).concat(".class") will avoid one allocation and one loop

sdedic: > clazz.substring(0, i).concat(".class") will avoid one allocation and one loop

I can't see that - could you explain, especially the "loop" part ?

nbbuild/templates/projectized.xml
```diff
@@ -123,7 +123,7 @@
         </condition>
     </target>
     <target name="-init-bootclasspath-prepend-run9" depends="-init-bootclasspath-prepend-compile" if="have-jdk-1.9">
-        <condition property="test.bootclasspath.prepend.args" value="--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming,jdk.jdi">
+        <condition property="test.bootclasspath.prepend.args" value="--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming,jdk.jdi,jdk.unsupported">
```
mbien, line 126: excellent! This means we have another job which could run on JDK 11+ (e.g in a matrix).

mbien (approved):
looks good to me. Appreciate the fixes to the exception logging too - this makes it easier to find issues in the log when something happens.

nbbuild/templates/projectized.xml
sdedic: Not really - there are still some failing tests, I asked @dbalek  for an assistance as the failure seems to be deeply burried in javac binding infrastructure. But it's definitely a step forward.

nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java
sdedic: Let's have some fun. I was curious, so I wrote a dumb testcase:
```
        long n1 = System.nanoTime();
        for (long i = 0; i < 100000000L; i++) {
            int index = S.indexOf('$');
            String x = S.substring(0, index) + ".class";
            A[0] = x;
        }
        long n2 = System.nanoTime();
        for (long i = 0; i < 100000000L; i++) {
            int index = S.indexOf('$');
            String x = S.substring(0, index).concat(".class");
            A[0] = x;
        }
        long n3 = System.nanoTime();
        
        System.err.println("First: " + (n2 - n1));
        System.err.println("Second: " + (n3 - n2));
```

What was funy was the result:
```
First:      3339608014
Second: 3703034326
```

mbien: i have also an older JMH benchmark to offer:
```java
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 5, time = 5, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 5, timeUnit = TimeUnit.SECONDS)
@Fork(1)
@State(Scope.Benchmark)
public class ConcatJMH {

    int i = 0;

    @Setup(Iteration)
    public void setup() {
        i = 0;
    }

    @Benchmark
    public void chain(Blackhole bh) {
        StringBuilder sb = new StringBuilder();
        sb.append("a");
        sb.append("[").append(i).append("]");
        sb.append("c");
        bh.consume(sb.toString());
    }

    @Benchmark
    public void mix(Blackhole bh) {
        StringBuilder sb = new StringBuilder();
        sb.append("a");
        sb.append("["+i+"]");
        sb.append("c");
        bh.consume(sb.toString());
    }

    @Benchmark
    public void plus(Blackhole bh) {
        bh.consume("a"+"["+i+"]"+"c");
    }

    @Benchmark
    public void concat(Blackhole bh) {
        bh.consume("a".concat("[").concat(String.valueOf(i)).concat("]").concat("c"));
    }

    public static void main(String[] args) throws RunnerException {
        
        Options opt = new OptionsBuilder()
                .include(ConcatJMH.class.getSimpleName())
                .resultFormat(ResultFormatType.JSON)
                .result("results/"+ConcatJMH.class.getSimpleName()+".json")
                .build();

        new Runner(opt).run();

    }
}
```
results on JDK 19:
```
    Benchmark          Mode  Cnt   Score   Error  Units
    ConcatJMH.chain    avgt    5  11.478 ± 0.266  ns/op
    ConcatJMH.mix      avgt    5  19.877 ± 0.301  ns/op
    ConcatJMH.concat   avgt    5   8.388 ± 0.179  ns/op
```

good old '+' wins hands down :)

lbownik: thanks for the feedback, I learned something new today :) 
it's good that the most compact form of concatenation is finally the most performant one
let's hope NetBeans doesn't get stuck with java8 forever ;)

mbien:
lets merge this this week so that we can finish up the migration PR and open it for review. Travis will end support for open source projects, apache projects have to migrate before the end of this year.

dbalek (approved):

Issue #4778, opened 2022-10-12
CSS: Show documentation (ctrl+shift+space) displays incorrect documentation


troizet:
### Apache NetBeans version

Apache NetBeans 15

### What happened

It looks like the documentation search for the display is not done by the full word, but by a few characters before the caret. Also the search is done for partial matches.

### How to reproduce

https://user-images.githubusercontent.com/9607501/195260539-1a57f8af-9b6a-42bd-8d17-83595b851a16.mp4

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Debian 10, Ubuntu 20.04.4 LTS

### JDK

OpenJDK 64-Bit Server VM 11.0.16+8-post-Ubuntu-0ubuntu120.04, OpenJDK Runtime Environment 11.0.16+8-post-Debian-1deb10u1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

The issue is similar to #4494.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
@troizet please have a look at the referenced PR, which should fix the issue. A testbinary is available there if needed.

troizet:
@matthiasblaesing Tried a test build. Everything works great. Thank you!

matthiasblaesing:
@troizet thanks for checking!

PR #4979, opened 2022-11-16
Optimize matching for CSS completion queries

matthiasblaesing:
There are two different issues:

- The documentation is fetched based on the completion result. This was only build based on the prefix. This change sorts the completion result using the node image where the cursor is currently located and sort by levenshtein distance between this image and the completion proposal. This result in preferring better matching entries.

- The documentation is looked up from the W3C specificiation HTML files. To do this the correct location is located using several regular expressions. To make this more precise a less flexible regexp is used. The other matches are retained and used as fallbacks.

Closes: #4778

### DIFF
--- a/ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java
+++ b/ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java
@@ -38,6 +38,7 @@ import javax.swing.ImageIcon;
 import javax.swing.text.Caret;
 import javax.swing.text.Document;
 import javax.swing.text.JTextComponent;
+import org.netbeans.api.annotations.common.NonNull;
 import org.netbeans.api.editor.EditorRegistry;
 import org.netbeans.api.lexer.Token;
 import org.netbeans.api.lexer.TokenHierarchy;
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
Testbuild is available here: https://doppel-helix.eu/NetBeans-dev-dev-3e58e3c60caa70f35929c6d2a98621cbc19623eb-release.zip

ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java
```diff
@@ -109,7 +109,7 @@ public class CssCompletion implements CodeCompletionHandler {
     @Override
     public CodeCompletionResult complete(CodeCompletionContext context) {
 
-        final List<CompletionProposal> completionProposals = new ArrayList<>();
+        List<CompletionProposal> completionProposals = new ArrayList<>();
 
         CssParserResult info = (CssParserResult) context.getParserResult();
```
lbownik, line 114: info variable seems redundant

ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java
```diff
@@ -1682,6 +1689,40 @@ public boolean visit(Node node) {
         } //switch
     }
 
+    @SuppressWarnings("AssignmentToMethodParameter")
+    public final int levenshteinDistance(
+            @NonNull String str1,
+            @NonNull String str2,
+            final boolean caseSensitive) {
```
lbownik, line 1696: putting blank line after { would make it more readable

matthiasblaesing: This is a verbatim copy, so I'll keep it this way.

ide/css.editor/src/org/netbeans/modules/css/editor/csl/CssCompletion.java
matthiasblaesing: a) When debugging it can be helpful to have the variable as local
b) This was not modified in this PR, so won't be touched as no improvment will be the result.

lbownik: sure

PR #4985, opened 2022-11-18
Use Project Dictated JVM for Gradle Discovery

lkishalmi:
There would be more commits after this one. Need to iron out the following things:

* Use the Project dictated Gradle Version in most cases. Revert back to the supplied one only if the project JVM is misaligned.
* Project problem provider should only report misaligned project JVM and Gradle version, also offer a fix to select project JVM
* Move Gradle Compile customizer options to Gradle Execution
* Offer better Gradle JVM Selection made it available on Root project only.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
@@ -20,7 +20,6 @@ package org.netbeans.modules.gradle.loaders;
 
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -81,7 +80,6 @@ import org.openide.util.Cancellable;
 import org.openide.util.NbBundle;
 
 import static org.netbeans.modules.gradle.loaders.Bundle.*;
-import org.netbeans.modules.gradle.spi.execute.GradleJavaPlatformProvider;
 import org.openide.filesystems.FileUtil;
 import org.openide.util.Exceptions;
 import org.openide.util.RequestProcessor;
@@ -474,20 +472,8 @@ public class LegacyProjectLoader extends AbstractProjectLoader {
             offline.addFlag(GradleCommandLine.Flag.OFFLINE);
         }
 
-        File javaHome = null;
-        GradleJavaPlatformProvider jvmProvider = projectImpl.getLookup().lookup(GradleJavaPlatformProvider.class);
-        try {
-            if (jvmProvider != null) {
-                javaHome = jvmProvider.getJavaHome();
-                LOG.log(Level.FINE, "Using JAVA_HOME=''{0}'' for project info load for: {1}", new Object[]{javaHome, projectImpl});
-            }
-        } catch (FileNotFoundException ex) {
-            LOG.log(Level.WARNING, "JAVA_HOME for project " + projectImpl + " not found.", ex);
-        }
-
         if (goOnline == GoOnline.NEVER || goOnline == GoOnline.ON_DEMAND) {
             BuildActionExecuter<NbProjectInfo> action = createInfoAction(pconn, offline, token, pl);
-            action.setJavaHome(javaHome);
             wasOnline.set(!offline.hasFlag(GradleCommandLine.Flag.OFFLINE));
             try {
                 ret = runInfoAction(action);
@@ -505,7 +491,6 @@ public class LegacyProjectLoader extends AbstractProjectLoader {
         }
         
         BuildActionExecuter<NbProjectInfo> action = createInfoAction(pconn, online, token, pl);
-        action.setJavaHome(javaHome);
         wasOnline.set(true);
         return runInfoAction(action);
     }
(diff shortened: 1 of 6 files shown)


extide/gradle/src/org/netbeans/modules/gradle/loaders/LegacyProjectLoader.java
```diff
@@ -472,8 +474,20 @@ private static NbProjectInfo retrieveProjectInfo(NbGradleProjectImpl projectImpl
...
+                javaHome = jvmProvider.getJavaHome();
+                LOG.log(Level.FINE, "Using JAVA_HOME=''{0}'' for project info load for: {1}", new Object[]{javaHome, projectImpl});
+            }
+        } catch (FileNotFoundException ex) {
+            LOG.log(Level.WARNING, "JAVA_HOME for project " + projectImpl + " not found.", ex);
+        }
+
         if (goOnline == GoOnline.NEVER || goOnline == GoOnline.ON_DEMAND) {
             BuildActionExecuter<NbProjectInfo> action = createInfoAction(pconn, offline, token, pl);
+            action.setJavaHome(javaHome);
```
lbownik, line 490: what if javaHome variable is null here?

lkishalmi: From: https://docs.gradle.org/current/javadoc/org/gradle/tooling/ConfigurableLauncher.html#setJavaHome-java.io.File-

> If not configured or null is passed, then the sensible default will be used.

sdedic (approved):

matthiasblaesing:
My personal worst-case is developing for liferay-portal. It uses a custom patched gradle version and requires JDK 8 to build. This seems to improve the situation a lot! This is not yet a comprehensive test, but this is what I did:

I build NetBeans from this branch. Additionally I had to adjust the gradle deamon settings like this (`~/.gradle/gradle.properties`):

```properties
org.gradle.jvmargs=-Xms512m -Xmx4g
liferay.home=/home/matthias/src/liferay-portal/bundles/
```

The first option give the gradle daemon enough space to handle the projects and the second is needed as the liferay build system sets this property from the ant invocation that starts the build. I have to restructure the build minimally, so that the grade wrapper is found. I build liferay once using `ant all`, then I copy the `gradle` directory from the main directory to `modules`. In the ` modules/gradle/wrapper/gradle-wrapper.properties`  I change the `distributionUrl` to `distributionUrl=../../../tools/gradle-6.9.2-LIFERAY-PATCHED1-bin.zip`. Finally the patched gradle needs to be renamed to fit the gradle versioning scheme from ` gradle-6.9.2.LIFERAY-PATCHED-1-bin.zip` to `gradle-6.9.2-LIFERAY-PATCHED1-bin.zip` .

Project loading takes time, but it feels nice to be able to open the projects in a sane manner. Thank you!

matthiasblaesing:
As an update: I noticed, that I tested running on OpenJDK 11, but it also works on JDK 17 if I patch the `GradleDistributionManager`

```diff
diff --git a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
index c468f0737a1e..366b0574a8b1 100644
--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
@@ -532,7 +532,8 @@ public final class GradleDistributionManager {
          * @return <code>true</code> if this version is supported with the runtime JDK.
          */
         public boolean isCompatibleWithSystemJava() {
-            return isCompatibleWithJava(JAVA_VERSION);
+            return true;
+//            return isCompatibleWithJava(JAVA_VERSION);
         }
 
         /**
```

lkishalmi:
Yes, that's a trivial patch for the first bulletpoint. It's on my radar, just got slammed with work in the last few weeks.

lkishalmi:
Well, this form is a good enough implementation, without major changes.

The `GradleProjectConnection` got rid of the "compatible" connection thing, So NetBeans does not try override the Gradle distribution in any case. The `javaHome` is set there so other places where `ProjectConnection` is needed, has the Java home and the Gradle Distribution set.

The Java vs Gradle version mismatch problem only triggered, if the Project Java setting is off.

I'm going to create a new update branch for NB16, where this and a few other fixes would go, as I think the Gradle support in base NB16 is not that great.

matthiasblaesing:
Works for me. Build from this branch, ensured JDK 17 is active and was able to load liferay project after setting its JDK to 8. Thank you!

sdedic:
Question: should be https://github.com/apache/netbeans/pull/4985/files#diff-3c58a0a0828b9926f6fc248e3020e6552233d2abe62b93fc0481dae7701dd1feR312 this still a warning message ?

errael:
@sdedic 
> Question: should <some_link> still a warning message ?

I think the link is messed up.

lkishalmi:
> Question: should be https://github.com/lkishalmi/netbeans/blob/gradle-use-project-java-for-discovery/extide/gradle/src/org/netbeans/modules/gradle/execute/GradleDaemonExecutor.java#L312 still a warning message ?
> 
> (link corrected)

I think yes, it should be a Warning message. Though it is now displayed if the JVM set on project is not compatible with the Gradle in use. In that case the IDE would still try to use the "built-in" Gradle. It would be elegant to have a resolver for this problem. That's for a later time though...

Issue #4861, opened 2022-10-26
CSS: Hint "unknown property" is displayed, in case the css property is written in uppercase letters.


troizet:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

Hint "unknown property" is displayed, in case the css property is written in uppercase letters.

![image](https://user-images.githubusercontent.com/9607501/197937097-86b7609a-5e8f-4694-9e97-23063a9a25df.png)

### How to reproduce

Write the css property in uppercase letters.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Debian 10

### JDK

OpenJDK Runtime Environment 11.0.16+8-post-Debian-1deb10u1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
@troizet could you please have a look at the referenced PR? It should fix this issue and a test build is available there.

matthiasblaesing:
@troizet thanks for checking!

PR #4986, opened 2022-11-18
Lookup CSS properties case-insensitive

matthiasblaesing:
Closes: #4861

### DIFF
--- a/ide/css.editor/src/org/netbeans/modules/css/editor/module/main/BrowserSupportModule.java
+++ b/ide/css.editor/src/org/netbeans/modules/css/editor/module/main/BrowserSupportModule.java
@@ -66,7 +66,12 @@ public class BrowserSupportModule extends CssEditorModule implements CssModule {
 
     @Override
     public PropertyDefinition getPropertyDefinition(String propertyName) {
-        return parser.getVendorSpecificProperties().get(propertyName);
+        PropertyDefinition pd = parser.getVendorSpecificProperties().get(propertyName);
+        if (pd != null || propertyName == null) {
+            return pd;
+        } else {
+            return parser.getVendorSpecificProperties().get(propertyName.toLowerCase());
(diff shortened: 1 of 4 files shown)


matthiasblaesing:
Testbuild: https://doppel-helix.eu/NetBeans-dev-dev-6f2056b4929fc2089418b487defec1b4150efefa-release.zip

ide/css.editor/src/org/netbeans/modules/css/editor/module/main/BrowserSupportModule.java
```diff
@@ -66,7 +66,12 @@ public Collection<String> getPropertyNames(FileObject file) {
 
     @Override
     public PropertyDefinition getPropertyDefinition(String propertyName) {
-        return parser.getVendorSpecificProperties().get(propertyName);
+        PropertyDefinition pd = parser.getVendorSpecificProperties().get(propertyName);
```
lbownik, line 69: what is the result of this execution if propertyName is null? can it ever be null?

matthiasblaesing: I don't understand what is the relevance of this question. If there was a problem before this change it is still there. This is not the point of the changeset.

To answer the question: `BrowserSpecificDefinitionParser#getVendorSpecificProperties` is backed by a `HashSet`. This allows `null` as key. The parser though will not yield `null` keys, so the `null` will not have a backing value, yielding `null` fro the `PropertyDefinition`. This in turn is caught in the new line 70, which explicitly checks for `null`.

lbownik: ok - good 
 just saw a lack of null check and though there might be some potential NullPointer there.

PR #4999, opened 2022-11-22
upgrade to maven-indexer 7.0.1 and improve index downloads

mbien:
upgrade to maven-indexer 7.0.1. (NetBeans entered the modern era! :-) )

 - module requires now jdk 11+
 - apache lucene 9.5.0 with memory mapped index
 - uses panama on JDK 19 ([PR1](https://github.com/apache/lucene/pull/912), [PR2](https://github.com/apache/lucene/pull/12033)) ([JDK 20 PR](https://github.com/apache/lucene/pull/12188))
   look out for: `INFO [org.apache.lucene.store.MemorySegmentIndexInputProvider]: Using MemorySegmentIndexInput with Java 19; to disable start with -Dorg.apache.lucene.store.MMapDirectory.enableMemorySegments=false`
 - multi-threaded remote index extraction (at least twice as fast as ST, disabled by default)
 - removed hack which backported a bugfix
 - no EOL libs anymore in this module
 - remote index download permission UI
 - ~~limited concurrent indexing~~ extracted and already [integrated](https://github.com/apache/netbeans/pull/5290) in NB 17

remote index (maven central) extraction takes ~6 minutes (post change) on my system (4 cores, hyperthreading off).
local `.m2` indexing speed is about the same

benchmark post-PR (old 4 core intel i7 and NVMe, JDK 19):
```
4 threads:
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of central took 383.24 s.
index folder: 6.4GB

3 threads:
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of central took 453.92 s.
index folder: 6,2 GB

2 threads:
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of central took 557.24 s.
index folder: 5,9 GB

1 thread:
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of central took 916.25 s.
index folder: 5,6 GB
```
thread count is capped to 4 since it incurs a on-disk file size penalty, furthermore the parallel extraction is always followed by a single threaded merge, which means more threads would have diminishing returns and move the bottle neck to the ST merge phase.

NB 17 on JDK 19 as comparison:
```
INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of central took 1118.49 s.
index folder: 5,8 GB
```

~~note: tmp files aren't cleaned up properly atm. This might fill up your tmp during testing. (edit: fixed via https://github.com/apache/maven-indexer/pull/273)~~ fixed in 7.0.1

https://issues.apache.org/jira/browse/NETBEANS-1268

UI:
repo index download permission notifications:

index-permission-bubble_v2
index-permissions-notificaiton

Options (permission table + multi threading check box):
indexer-options


java/maven.indexer/manifest.mf
```diff
@@ -3,4 +3,4 @@ OpenIDE-Module-Localizing-Bundle: org/netbeans/modules/maven/indexer/Bundle.prop
 AutoUpdate-Show-In-Client: false
 OpenIDE-Module-Specification-Version: 2.59
 OpenIDE-Module: org.netbeans.modules.maven.indexer/2
-
+penIDE-Module-Java-Dependencies: Java > 11
```
mbien, line 6: fixme

mbien:
bump to maven-indexer 7.0.1

mbien:
thanks to @cstamas for resolving the silo cleanup issue - works great!

mbien:
draft: https://github.com/apache/maven-indexer/pull/302 would give us more control over the temp storage location + give us the option to filter the index

mbien:
wip permission request notification
dl-permission-bubble_v1

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
```diff
@@ -534,12 +532,22 @@ private void indexLoadedRepo(final RepositoryInfo repo, boolean updateLocal) thr
                         httpwagon.setHttpHeaders(p);
                     }
 
-                    Path tmpStorage = Files.createTempDirectory("index-extraction");
+                    Path tmpStorage = Files.createTempDirectory(getIndexDirectory(), "extractor-");
```
mbien, line 535: note to self: this is extracting directly into cache right now. I am not convinced that this is a good idea since this requires the cleanup on failure/cancel/system.exit to work reliably.

The advantage of this approach is that only the compressed archive will land in tmp - which should fit even into small tmp folders.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
```diff
@@ -637,6 +644,10 @@ private void indexLoadedRepo(final RepositoryInfo repo, boolean updateLocal) thr
         }
     }
 
+    private static void pauseRemoteRepoIndexing() {
+        RepositoryPreferences.pauseIndexDownloadsFor(2, ChronoUnit.MINUTES); // pause a while  // TODO use reasonable duration
```
mbien, line 648: TODO: forgot to change this number

vieiro:
Hi @mbien,

I have no remote URLs to download from:

imagen

Is this as expected? How do I add one?

On other news, here's some stuff from my stdout (question: are we expected to System.out all artifacts in the local repo?)

```
...
     [exec] [Maven Repo Index Transfer/Scan] INFO org.eclipse.aether.internal.impl.DefaultArtifactResolver - Artifact org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-bom:pom:2.0.0 is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable from [central (https://repo.maven.apache.org/maven2, default, releases)]
     [exec] [Maven Repo Index Transfer/Scan] INFO org.eclipse.aether.internal.impl.DefaultArtifactResolver - Artifact org.jboss.shrinkwrap.resolver:shrinkwrap-resolver-bom:pom:3.1.4 is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable from [central (https://repo.maven.apache.org/maven2/, default, releases+snapshots)]
     [exec] [Maven Repo Index Transfer/Scan] INFO org.eclipse.aether.internal.impl.DefaultArtifactResolver - Artifact org.jboss.shrinkwrap.descriptors:shrinkwrap-descriptors-bom:pom:2.0.0 is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable from [central (https://repo.maven.apache.org/maven2/, default, releases+snapshots)]
     [exec] [Maven Repo Index Transfer/Scan] INFO org.eclipse.aether.internal.impl.DefaultArtifactResolver - Artifact org.jboss.shrinkwrap:shrinkwrap-bom:pom:1.2.6 is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable from [central (https://repo.maven.apache.org/maven2/, default, releases+snapshots)]
     [exec] INFO [org.netbeans.modules.maven.indexer.NexusRepositoryIndexerImpl]: Indexing of local took 18,09 s.

```

mbien:
@vieiro this is a list of user given permissions. Try opening a pom or something which triggers indexing, nb will then show a notification, based on what you answer it will show up in the list.

vieiro:
> @vieiro this is a list of user given permissions. Try opening a pom or something which triggers indexing (e.g add dependency to a project via UI), nb will then show a notification, based on what you answer it will show up in the list. You don't have to change anything on the list unless you want to change the permission in post.

Hi @mbien , thanks for the tip!

Works perfect now. I've tried to unpack the maven index from central, after a few minutes with fans at top speed and ~3.9Gb `var/cache/mavenindex/` I decided to cancel it, and it's now disabled forever! Thanks very much for this! 

imagen

matthiasblaesing (commented):
Thank you for your work. This is what I noticed:

- I canceled a running remote index download, then I have no option to restart that, because the "Index Now" button only triggers a local index, not an index of the remote repositories. I did not look deeply into the mechanisms that hold the "indexing was started, but aborted" logic, but somewhere this seems to be stored, because it did not retrigger, when asking for completion. IMHO either the "Index Now" button also covers the remote repositories or the listview for the known repository should allow requesting an update on an individual basis.
- The update frequency is still "Never", this contradicts my answer to allow downloading
- Multithreaded indexing should IMHO be enabled by default (we should at least pretend, that hardware manufactures know how to build thermally stable hardware)
- I suggest for the popup:
  - move text into bundle, so that it might be localiced
  - Text suggestion:
    ```java
        pane.setText(MessageFormat.format(
                "Maven index download requested for: {0}<br/>"
                + "<ul>"
                + "<li><a href=\"allow-always\">Allow always</a></li>"
                + "<li><a href=\"allow-once\">Allow once</a></li>"
                + "<li><a href=\"deny-once\">Deny once</a></li>"
                + "<li><a href=\"deny-always\">Deny always</a></li>"
                + "</ul>"
                + "A repository index contains artifact metadata which is useful for some NetBeans features.<br />"
                + "<a href=\"disable-downloads\">Disable all remote downloads</a>",
                repo.getRepositoryUrl()
        ));
    ```

    Leads to:

    Bildschirmfoto vom 2023-04-30 21-52-20

    At least Jan noticed, that maven download insisted on downloading while on the bus, that might be a situation where I'd like it to "Deny once".
- Looking at the indexing notification: Will this be a problem for LSP?

mbien:
>  I canceled a running remote index download, then I have no option to restart that, because the "Index Now" button only triggers a local index, not an index of the remote repositories.

right, canceling will add a delay before it will try to run again. This is intended. I am not sure how important the usecase is to start the indexing right after a manual cancel again. I agree that the "index now" button needs to be looked at since it is only for the local index and this is not clearly communicated in the UI. Restart after cancel is more of a dev testing issue for me and less of a common usecase.

> Multithreaded indexing should IMHO be enabled by default 

I don't want to do this before filtering is available since this has an index size overhead. This would be another PR. We also got feedback of notebook usage etc which I don't want to ignore either.

> I suggest for the popup: 

I will take a look at this. But I won't implement "run once" for now since this complicates things unnecessarily. Indexing is not a run-once feature, either you turn it on or off. Cancelling pauses it as mentioned before.

> Looking at the indexing notification: Will this be a problem for LSP?

likely. `IndexingNotificationProvider.java` has one method more now, if this isn't implemented and nobody sets the permission, indexing won't run now. Let me add a default implementation which simply allows all downloads to have the same behavior as before.

mbien (commented):
changes:
 - "update now" button will trigger updates for all repos (permission required, download pause is reset)
   - if a remote repo is already up2date, it won't update the repo
   - changed label text to "Check for Updates"
 - added a dummy impl for `IndexingNotificationProvider` to retain old behavior even when no proper implementation is found
 - extracted text to bundle, updated the notification a bit

index-permission-bubble_v2

matthiasblaesing:
I think this is mostly good to go. The only thing missing is IMHO restoring periodic updates. Given, that you now clear all downloads, there is IMHO no reason to still keep it on never and it IMHO contradicts the "allow" setting.

vieiro:
Great job, @mbien! 

There's a "gobally" -> "globally" typo you may want to take a look at before merging.

mbien:
> There's a "gobally" -> "globally" typo you may want to take a look at before merging.

thanks! I wouldn't have noticed this.


> The only thing missing is IMHO restoring periodic updates. Given, that you now clear all downloads, there is IMHO no reason to still keep it on never and it IMHO contradicts the "allow" setting.

good point. Probably better to remove the `never` option now. The usecase of `never` is also covered by the index downloads check box. Going to take a look at it.

mbien:
last commit removes the `never` update frequency. Typo is fixed too but already squashed.

planning to squash it into two commits before merge

matthiasblaesing (approved):

mbien:
last test run with all tests enabled. thanks for the reviews -> going to merge today

mbien:
merging earlier than planned since I got some time for testing and the next steps (search service/filters)

PR #5006, opened 2022-11-25
Gradle: Support for proxy detection / autoconfiguration.

sdedic:
On a corporate network or on a VPN the user often forgets about a network change - even worse because with PAC autoconfiguration, NetBeans IDE is able to dynamically change proxy / non-proxy, depending on PAC results.

But build tools like Gradle have to get explicit instructions. The user uses NetBeans, happilly browses online resources (with an PAC-configured browser), but build actions or even project info load itself fails mysteriously on missing on unreachable artifacts.

This PR adds a check, before taking an 'online' action in Gradle, if Gradle's configuration file matches IDE's own proxy settings. If it does not, the default is to suggest the user to reconfigure Gradle - and NetBeans will even do that for the user.
proxy2

The user may also choose to leave settings as they are (for whatever reason), but let the IDE to instruct Gradle to use proxy for IDE-initiated operations. In this scenario, commandline actions will likely fail - but no persistent change in files need to be made.

A global Gradle setting is introduced to specify the preferred handling:
- **ask the user (the default)**
- do not check
- display a notice and override
- override using system properties
- automatically update

Old `gradle.properties` are always saved as `gradle.properties.old` (`gradle.properties.old.1`, ....) to preserve potential user content, although `EditableProperties` API is used to retain eventual user comments in the file.

The current implementation contains some cut-through path, since because of gradle/gradle#22856 I am able to define a proxy, if one is not defined at all, but the bug prevents me from **changing** or **removing** proxy settings on-the-fly in case the proxy is configured in the file, but the config is wrong.

Note: because of Vscode handling, the option to pass system properties ("Override" option) can be disabled by branding API.

### DIFF
--- a/extide/gradle/apichanges.xml
+++ b/extide/gradle/apichanges.xml
@@ -91,7 +91,7 @@ is the proper place.
             <author login="sdedic"/>
             <compatibility semantic="compatible" addition="yes"/>
             <description>
-                Tasks declared by other projects are marked as 'external'. Tasjs can report their declaring project's path.
+                Tasks declared by other projects are marked as 'external'. Tasks can report their declaring project's path.
             </description>
             <class package="org.netbeans.modules.gradle.api" name="GradleTask"/>
         </change>
--- a/extide/gradle/arch.xml
+++ b/extide/gradle/arch.xml
@@ -130,6 +130,13 @@
 
 
  <answer id="exec-property">
+     <api category="devel" group="branding" name="org.netbeans.modules.gradle.api.execute.NetworkProxySettings.allowOverride" type="export">
+         Brand the <code>org.netbeans.modules.gradle.api.execute.NetworkProxySettings.allowOverride</code> key in a 
+         <code>org.netbeans.modules.gradle.api.execute.Bundle</code> file
+         with one of the values <code>true</code> or <code>false</code> to specify whether to offer override of proxies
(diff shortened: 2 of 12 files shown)


lkishalmi:
Would not be this better as a project problem provider with a resolution?

sdedic:
I would advise against that, but it is surely doable. In siuations where the DNS does not reject the proxy name immediately (timeout) or when the intranet simply drops (not rejects) outgoing packets, gradle app + network timeouts are huuuuuuge. It may take several to dozen minutes for the project load to fail. 
It seemed safer + better user experience to hook before any non-offline operation and check in advance.

sdedic:
I would appreciate if this PR could land in NB16u1-gradle (cc: @MartinBalin) -- in NB a misconfigured proxy is bad enough, but if the vscode environment is used with several lang servers all working against bad proxy/network configuration, it is a true hell on developer's machine.

sdedic:
@neilcsmith-net understood - but I'd like to get the code in here reviewed or sanity-checked first

dbalek (approved):
Useful functionality, Thanks.

sdedic:
@lkishalmi -- last call: there's sanity check from @dbalek, but this is your field of expertise, I'd feel more comfortable if I could get +-0 from you for this implementation.

lkishalmi (approved):
I like the idea, that this feature flagged!

PR #5009, opened 2022-11-26
adds a basic maven dependency updater hint.

mbien:
- marks artifacts which have a newer version available
 - provides a fix to bump the version
 - supports maven properties
   update-properties
 - ~ignores non-numerical versions~ discovered `ComparableVersion` :)
  update-deps
   * only suggests qualifiers like `-SNAPSHOT` if the used version is also using a qualifier
   * ignores some non-standard version formats. Like timestamps, since they are rarely used anymore.
   * can be restricted to not suggest major version upgrades
artifact-upgrade-options

I wasn't sure whether or not to enable this by default but I decided to enable it mostly because:
 - many new-project-poms are outdated, this mitigates the issue a little bit (this is also the original motivation for the hint)
 - poms don't have a lot of hints right now, so adding a little bit of noise probably doesn't hurt atm
 - users can turn it off / we can turn it off based on feedback later

Performance seems good. It checks large poms in about a second. Maven hints are not cancellable unfortunately.


mbien:
only rebased to latest master

matthiasblaesing (commented):
Implementation looks sane. Two thoughts:

1. (Out of scope for this changeset): Maven hints are missing "Show as warning on current line". With more hints the noise goes up and the hepfulness goes down. This is a candiate for to much noise. @work this would flag nearly all dependencies and thus I would deactivate it.
2. Would it make sense to provide an update to latest and "latest in same major version"? I associate bumping major version with incompatible change or at least "no stability guarentee".

mbien:
> Implementation looks sane. Two thoughts:
> 
>     1. (Out of scope for this changeset): Maven hints are missing "Show as warning on current line". With more hints the noise goes up and the hepfulness goes down. This is a candiate for to much noise. @work this would flag nearly all dependencies and thus I would deactivate it.

yeah would be nice to have feature parity with the java hints where it makes sense. There is also currently no way to run maven hints as inspections although this is probably only rarely needed. Projects can define project specific hints but this is also only possible for java hints I believe (I don't see anything else in the combo box).

> 
>     2. Would it make sense to provide an update to latest and "latest in same major version"? I associate bumping major version with incompatible change or at least "no stability guarentee".

Done. I added it as option in the hint settings. I actually thought about adding this too but dismissed it since I initially wanted to keep this really simple (since there are maven plugins for this). But I believe a simple "don't suggest major version upgrades" probably covers 95% of all typical cases, so why not.

The initial motivation for this was because many project pom templates are outdated, this gives the user a quick upgrade path even on first use once the IDE downloaded the maven index. The release version hint (and other IDE features too) require certain compiler plugin versions. This hint can basically unlock other hints by bumping this version.

java/maven.hints/src/org/netbeans/modules/maven/hints/pom/UpdateDependencyHint.java
```diff
@@ -0,0 +1,259 @@
...
+        return config.getPreferences().getBoolean(KEY_NO_MAJOR_UPGRADE, false);
+    }
+
+    @Override
+    public String getSavedValue(JComponent customizer, String key) {
+        if (KEY_NO_MAJOR_UPGRADE.equals(key) && customizer instanceof UpdateDependencyHintCustomizer) {
+            return Boolean.toString(((UpdateDependencyHintCustomizer)customizer).getSavedNoMajorUpgradeOption());
+        }
+        return null;
+    }
```
mbien, line 249: The maven hints option panel seems to be fairly broken atm. Cancel doesn't work for most hint options etc. The implementation is also a bit... unusual. It polls the `getSavedValue` method twice per second and it forgets polling again when cancelled and reopened.

It also expects the UI to remember the old ("saved") value for the options? I don't think it is right now possible to make the ok, cancel and apply button work at the same time. I chose ok.

mbien: i believe @neilcsmith-net mentioned this once, now I know what you meant :D

mbien: I think I have fixed the polling issue. This appears to be the root cause of options not cancelling properly and also the apply button not getting activated. Last commit contains the fix.

matthiasblaesing (approved):
Looks sane to me.

mbien:
awesome, squashed PR to:
 - first commit will be the maven option panel polling bugfix
 - second commit is the feature

this should hopefully reduce risks and allow to test things independently if something happens. (I try to avoid fixing bugs in feature commits)

java/maven.hints/src/org/netbeans/modules/maven/hints/pom/UpdateDependencyHint.java
mbien: filed #5165 for the remaining issues

mbien:
@matthiasblaesing sorry for changing the PR after approval. I fixed #5165 too. See first commit.

matthiasblaesing:
Could reproduce the problem and found it fixed with the update. Thank you.

PR #5014, opened 2022-11-27
Fixed some Gradle introspection problems wit pre-6.1 Gradle Versions

lkishalmi:
Yet another fix, restoring some pre-6.1 Gradle Compatibility.

Brief list of the changes:
 * `VersionNumber` -> `GradleVersion` (VersionNumber is deprecated for a while)
 * Also comparing on base version, so no comparisons to RCs are needed.
 * Identify plugin by name instead of class
 * Checking for MissingPropertyException-s in `getProperty` and returning `null` if that would happen.

I'm still unhappy about the issues caused by the mass auto-discovery feature.


sdedic (approved):
Many thanks. I still lack the cross-version expertise.

matthiasblaesing (approved):
I'm not an expert here, but looks sane. `GradleVersion` indeed seems to be around since a long time and looks like a save usage here.

PR #5021, opened 2022-11-28
Support directly native build plugin

sdedic:
The micronaut module supports 'nativeCompile' task indirectly - only if `io.micronaut.application` plugin was included. This PR reacts also on `org.graalvm.buildtools.native` plugin provided by GraalVM community  -- so it would serve native artifacts even for other configuration than just with `io.micronaut.application`.
The GraalVM plugin is also used by SpringBoot and possibly by other 4rd party plugins, too.


MartinBalin (approved):

jhorvath (approved):
Looks good

Issue #5015, opened 2022-11-27
Some Gradle Projects can't be Loaded due to NotSerializableException


lkishalmi:
### Apache NetBeans version

Apache NetBeans 16 release candidate

### What happened

I've tried to open the `netbeans-website/netbeans.apache.org` the IDE reports a  notification that it cannot be open due to NotSerializableException on `tasks.propertyValues`

### How to reproduce

1. Clone https://github.com/apache/netbeans-website
2. Open the netbeans.apace.org project in it

### Did this work correctly in an earlier version?

Apache NetBeans 15

### Operating System

Ubuntu 22.04 and 22.10

### JDK

Java 17

### Apache NetBeans packaging

Own source build

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

sdedic:
Will fix by tomorrow to get to the patch release. Thanks.

PR #4984, opened 2022-11-18
[#4865] Register JDPA and Java Compiler Output on Project level

lkishalmi:
This is a small change that registers the Gradle JDPA and JavaCompiler output listener on General Gradle project level instead of individual plugin level, so projects that non necessary Java project, but has Java sub-projects or deals with Java/JVM in some way can have their output processed. 

Fixes #4865

PR #5022, opened 2022-11-28
Fix Gradle/broken task dependency and serialization

sdedic:
During debugging of #5015, I've also encountered some broken task that depended on a non-existing class: so I wrapped task dependencies into `try-catch(LinkageError)`. 

The not serializable bug itself was caused by an improper test for primitive that passed on even some generated proxy with no package. The current version will still pass more values than just primitive wrappers (anything from java.lang).


lkishalmi (approved):
Looks good.

sdedic:
Note - the failing test is because of #4984, this PR should not have any effect on that test (negative or positive).

sdedic:
@lkishalmi can I merge this before #4984 fix ?

lkishalmi:
First this should go to master first. then it would be backported  (cherry-picked) for NB16u1 as a separate PR.

PR #5027, opened 2022-11-28
[NETBEANS-1914] Fix startup problems on Windows after moving NetBeans install directory

eirikbakke:
On Windows, renaming or moving the NetBeans installation directory caused errors on startup, due to references to old paths in all-layers.dat in the cache directory. The bug did not seem to happen on Linux or MacOS; I'm not sure why. See https://issues.apache.org/jira/browse/NETBEANS-1914 for an example stack trace.

The solution in this patch is to include the netbeans.home path in the all-checksum.txt file. This will invalidate the cache if the installation directory is changed.

### DIFF
--- a/platform/core.startup/test/unit/src/org/netbeans/core/startup/layers/CachingPreventsFileTouchesTest.java
+++ b/platform/core.startup/test/unit/src/org/netbeans/core/startup/layers/CachingPreventsFileTouchesTest.java
@@ -161,9 +161,18 @@ public class CachingPreventsFileTouchesTest extends NbTestCase {
         Collections.shuffle(Arrays.asList(arr));
         for (File f : arr) {
             if (!f.isDirectory()) {
-                System.err.println("checking " + f);
-                cnt++;
-                assertFileDoesNotContain(f, install);
+                if (f.getName().equals("all-checksum.txt")) {
+                    /* In org.netbeans.Stamps, we intentionally include the full netbeans.home path
+                    on Windows as a workaround for NETBEANS-1914. This is meant to invalidate the
+                    cache on changes to netbeans.home in case, despite the intentions of those who
+                    implemented the cache directory system, an absolute path to netbeas.home did end
+                    up in the cache directory. */
+                    System.err.println("skipping checksum file " + f);
+                } else {
+                    System.err.println("checking " + f);
+                    cnt++;
+                    assertFileDoesNotContain(f, install);
+                }
             }
         }
         assertTrue("Some cache files found", cnt > 4);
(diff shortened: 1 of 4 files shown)


mbien (dismissed):
elegant way of fixing this

edit: restarting tests since the platform jobs didn't run

mbien:
the tests don't want absolute paths in the file. 

```
at org.netbeans.core.startup.layers.CachingPreventsFileTouchesTest.assertFileDoesNotContain(CachingPreventsFileTouchesTest.java:200)
at org.netbeans.core.startup.layers.CachingPreventsFileTouchesTest.testCachesDontUseAbsolutePaths(CachingPreventsFileTouchesTest.java:166)

```

eirikbakke:
Hmm, I think that test is trying to catch the absolute paths that caused the problem in the first place (in files other than the checksum). Here was the discussion around the change that introduced this test:

https://bz.apache.org/netbeans/show_bug.cgi?id=208749

While getting rid of absolute paths from the cache is good, it seems like good hygiene to flush the cache when the installation directory changes in any case. Would be useful to get @jtulach's opinion here.

neilcsmith-net:
What's the status of this PR?  It's bumped through two milestones now - should either review / fix / merge, or close.

eirikbakke:
Recap of this PR:
1) On Windows, if the user moves the NetBeans install directory, NetBeans will fail to start, because the cache directory contains some absolute paths to files in the NetBeans install directory. This is a bug, because the cache directory is not supposed to contain absolute paths (we know this since @jtulach once wrote a testCachesDontUseAbsolutePaths test).
2) The fix to the bug in (1) would be to eliminate the absolute paths, but I looked into that and couldn't figure out how.
3) This PR takes a different approach: Just make sure the cache directory gets invalidated whenever the installation path changes. I think that's a good idea in any case; moving the install directory is a rare occurrence, except maybe if running from a USB drive where drive letters change or such.
4) As it happens, this PR itself fails the testCachesDontUseAbsolutePaths test because the invalidation is done by adding the absolute install path to the all-checksum.txt file. But I can just make it pass by allowing that particular occurrence of the absolute path, in all-checksum.txt only.

Unless @jtulach disagrees, my inclination here would be to fix the failing test by adjusting testCachesDontUseAbsolutePaths  to allow the absolute path in all-checksum.txt .

The main possible objection here would be in giving up the "feature" of being able to move the NetBeans install directory without invalidating the cache directory. (I think that's acceptable, in return for eliminating a whole class of bugs that make NetBeans permanently un-startable.)

neilcsmith-net:
> The main possible objection here would be in giving up the "feature" of being able to move the NetBeans install directory without invalidating the cache directory.

I think that might be OK, although it would be a pain for AppImages as the "install directory", well netbeans.home, moves every time you run.  As you mention Linux and macOS don't have this issue, so could we target only Windows for this temporary fix?  I say temporary, as 1. seems to be something that should be looked at - have seen that issue affect something else in a platform app.

eirikbakke:
I have now updated this patch to apply the workaround on Windows only, and to fix the resulting broken tests.

eirikbakke:
I'll merge this in, say, 14 days, if there are no objections.

PR #5038, opened 2022-12-01
Proxy autodetection + autoconfiguration for Maven.

sdedic:
Similar to #5006 but for Maven.  For "priming build" and maven online invocations, the proxy is checked first. If mismatch is found, the IDE asks the user (the default setting) other actions can be: ignore, auto-update maven settings.xml or override on-the-fly.

The override is done by generating a copy of the `~/.m2/settings.xml` with relevant parts changed/added, maven is then invoked with `--settings` option. Generated files are put into `$nbuserdir/var/cache`.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
+++ b/java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -84,6 +85,7 @@ import org.netbeans.modules.maven.runjar.MavenExecuteUtils;
 import org.netbeans.spi.project.ui.support.BuildExecutionSupport;
 import org.openide.LifecycleManager;
 import org.openide.awt.HtmlBrowser;
+import org.openide.awt.NotificationDisplayer;
 import org.openide.execution.ExecutionEngine;
 import org.openide.execution.ExecutorTask;
 import org.openide.filesystems.FileObject;
@@ -92,7 +94,9 @@ import org.openide.modules.InstalledFileLocator;
 import org.openide.modules.Places;
 import org.openide.modules.SpecificationVersion;
 import org.openide.util.Exceptions;
+import org.openide.util.ImageUtilities;
 import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
 import org.openide.util.RequestProcessor;
 import org.openide.util.Task;
 import org.openide.util.TaskListener;
@@ -143,6 +147,8 @@ public class MavenCommandLineExecutor extends AbstractMavenExecutor {
     private static final RequestProcessor RP = new RequestProcessor(MavenCommandLineExecutor.class.getName(),1);
 
     private static final RequestProcessor UPDATE_INDEX_RP = new RequestProcessor(RunUtils.class.getName(), 5);
+
+    private static final String ICON_MAVEN_PROJECT = "org/netbeans/modules/maven/resources/Maven2Icon.gif"; // NOI18N
     /**
      * Execute maven build in NetBeans execution engine.
      * Most callers should rather use {@link #run} as this variant does no (non-late-bound) prerequisite checks.
@@ -200,6 +206,11 @@ public class MavenCommandLineExecutor extends AbstractMavenExecutor {
         this.io = io;
     }
 
+    @NbBundle.Messages({
+        "# {0} - original message",
+        "ERR_CannotOverrideProxy=Could not override the proxy: {0}",
+        "ERR_BuildCancelled=Build cancelled by the user"
+    })
     /**
      * not to be called directly.. use execute();
      */
@@ -225,6 +236,49 @@ public class MavenCommandLineExecutor extends AbstractMavenExecutor {
         int executionresult = -10;
         final InputOutput ioput = getInputOutput();
 
+        MavenProxySupport mps = clonedConfig.getProject().getLookup().lookup(MavenProxySupport.class);
+        if (mps != null) {
+            MavenProxySupport.ProxyResult res;
+            boolean ok = false;
+            try {
+                res = mps.checkProxySettings().get();
+                
+                if (res != null) {
+                    res.configure(clonedConfig);
+                }
+                if (res.getStatus() == MavenProxySupport.Status.ABORT) {
+                    IOException ex = res.getException();
+                    
+                    if (ex == null) {
+                        ioput.getErr().append(Bundle.ERR_BuildCancelled());
+                    } else {
+                        throw ex;
+                    }
+                    
+                } else {
+                    ok = true;
+                }
+            } catch (IOException ex) {
+                NotificationDisplayer.getDefault().notify(Bundle.TITLE_ProxyUpdateFailed(),
+                        ImageUtilities.loadImageIcon(ICON_MAVEN_PROJECT, false),
+                        ex.getLocalizedMessage(), null, NotificationDisplayer.Priority.NORMAL, NotificationDisplayer.Category.ERROR);
+                ioput.getErr().append(Bundle.ERR_CannotOverrideProxy(ex.getLocalizedMessage()));
+                // FIXME: log exception
+            } catch (ExecutionException ex) {
+                // FIXME: log exception
+            } catch (InterruptedException ex) {
+                // FIXME: log exception
+            } finally {
+                if (!ok) {
+                    ioput.getOut().close();
+                    ioput.getErr().close();
+                    actionStatesAtFinish(null, null);
+                    markFreeTab();
+                    return;
+                }
+            }
+        }
+        
         final ProgressHandle handle = ProgressHandle.createHandle(clonedConfig.getTaskDisplayName(), this, new AbstractAction() {
             @Override
             public void actionPerformed(ActionEvent e) {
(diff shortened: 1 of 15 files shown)


MartinBalin (approved):
This has to go to 16.0.1 patch release along with Gradle fixes.

java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
```diff
@@ -225,6 +236,49 @@ public void run() {
         int executionresult = -10;
         final InputOutput ioput = getInputOutput();
 
+        MavenProxySupport mps = clonedConfig.getProject().getLookup().lookup(MavenProxySupport.class);
+        if (mps != null) {
+            MavenProxySupport.ProxyResult res;
```
lbownik, line 241: how about moving it to line 244 where it can be instanlty initialized?

java/maven/src/org/netbeans/modules/maven/options/NetworkProxySettings.java
```diff
@@ -0,0 +1,79 @@
...
+     * Ask the user for confirmation.
+     */
+    ASK,
+    /**
+     * Automatically override on execution, but do not change gradle.properties.
+     */
+    OVERRIDE;
+    
+    public String toString() {
+        switch (this) {
```
lbownik, line 58: wouldnt it be more elegant to introduce a proviate "name" or "str" field, initialize it properly it via constructor and then simply return it from toString()

java/maven/src/org/netbeans/modules/maven/problems/SanityBuildAction.java
```diff
@@ -97,6 +107,21 @@ public void run() {
                     String label = build_label(nbproject.getProjectDirectory().getNameExt());
                     config.setExecutionName(label);
                     config.setTaskDisplayName(label);
+                    
+                    MavenProxySupport mps = nbproject.getLookup().lookup(MavenProxySupport.class);
+                    if (mps != null) {
+                        ProxyResult res;
```
lbownik, line 113: this variable seems to be unused

sdedic: Hm, I'll better check the behaviour with 'cancel' in the dialog; thanks !

dbalek (approved):
Looks fine. Thanks.

java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java
```diff
@@ -826,7 +850,7 @@ public void valueChanged(javax.swing.event.ListSelectionEvent evt) {
                 .addComponent(lblCategory)
                 .addPreferredGap(javax.swing.LayoutStyle.ComponentPlacement.RELATED)
                 .addGroup(layout.createParallelGroup(javax.swing.GroupLayout.Alignment.LEADING)
-                    .addComponent(pnlCards, javax.swing.GroupLayout.DEFAULT_SIZE, javax.swing.GroupLayout.DEFAULT_SIZE, Short.MAX_VALUE)
+                    .addComponent(pnlCards, javax.swing.GroupLayout.DEFAULT_SIZE, 547, Short.MAX_VALUE)
```
mbien, line 853: Can you switch this back to `DEFAULT_SIZE`? This would set the component to a fixed size and layout managers won't resize it.

When I open the options now i see scroll bars even though everything would fit into the window.

To achieve this its often possible to open the panel in question via double click on the navigator. Then reduce the size till it snaps to default size.

mbien (commented):
this shouldn't be empty, right? ;)

empty_combo


The position right now is right between global exec options and global custom goals. It might look better if its moved one row up which would leave the other settings grouped together.

java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java
sdedic: argh ... stupid copy / cleanup error before commit: missing assignment. Thanks for catching.

sdedic:
> The position right now is right between global exec options and global custom goals. It might look better if its moved one row up which would leave the other settings grouped together.

You mean reorder to:
- network proxy
- global exec options
- the rest ?

If so then I'd separate the exec + network options (that affect the execution of maven) from the rest that affects behaviour of the tool

mbien:
> > The position right now is right between global exec options and global custom goals. It might look better if its moved one row up which would leave the other settings grouped together.
> 
> You mean reorder to:
> 
>     * network proxy
> 
>     * global exec options
> 
>     * the rest ?
> 
> 
> If so then I'd separate the exec + network options (that affect the execution of maven) from the rest that affects behaviour of the tool

maybe its fine. Its just that I usually wouldn't expect NW options between command line flags and maven goal definitions. If it should remain where it is right now, I would try to make the combo box the same width as the text field.

This panel might need a UI pass anyway, e.g maybe it would make sense to move everything what is below "edit global custom goal defs" to the appearance category which is almost empty ATM.

java/maven/src/org/netbeans/modules/maven/execute/MavenCommandLineExecutor.java
sdedic: fixed in 5d31b280c9d8

java/maven/src/org/netbeans/modules/maven/problems/SanityBuildAction.java
sdedic: fixed in 5d31b280c9d8

java/maven/src/org/netbeans/modules/maven/options/NetworkProxySettings.java
sdedic: fixed in 5d31b280c9d8 + 1126d7540e8a

sdedic:
> maybe its fine. Its just that I usually wouldn't expect NW options between command line flags and maven goal definitions. If it should remain where it is right now, I would try to make the combo box the same width as the text field.
> 

Kept order, made textfield + combo same size, resizable, aligned -- 1126d7540e8a

mbien:
looks much better to me now + can confirm that the combo box is working now

one more thing if you have the patience:
 - double click on all cards starting with the children on the navigator and shrink the height till it snaps to default size.
 - once done do the same for the main panel

this will remove the scrollbar since everything will be set to minimum size and the layout manager can do the rest.

here a diff if you want to apply it directly instead:
```diff
diff --git a/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.form b/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.form
index 595c94e..151e2a9 100644
--- a/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.form
+++ b/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.form
@@ -216,7 +216,7 @@
                       </Group>
                       <EmptySpace max="-2" attributes="0"/>
                       <Component id="jLabel3" min="-2" pref="44" max="-2" attributes="0"/>
-                      <EmptySpace pref="379" max="32767" attributes="0"/>
+                      <EmptySpace max="32767" attributes="0"/>
                   </Group>
               </Group>
             </DimensionLayout>
@@ -322,7 +322,7 @@
                       <Component id="cbDisableIndex" min="-2" max="-2" attributes="0"/>
                       <EmptySpace max="-2" attributes="0"/>
                       <Component id="jLabel5" min="-2" max="-2" attributes="0"/>
-                      <EmptySpace pref="450" max="32767" attributes="0"/>
+                      <EmptySpace max="32767" attributes="0"/>
                   </Group>
               </Group>
             </DimensionLayout>
@@ -422,7 +422,7 @@
                           <Component id="txtDirectory" alignment="3" min="-2" max="-2" attributes="0"/>
                           <Component id="btnDirectory" alignment="3" min="-2" max="-2" attributes="0"/>
                       </Group>
-                      <EmptySpace pref="332" max="32767" attributes="0"/>
+                      <EmptySpace max="32767" attributes="0"/>
                   </Group>
               </Group>
             </DimensionLayout>
diff --git a/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java b/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java
index 187de3b..b960bfc 100644
--- a/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java
+++ b/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java
@@ -525,7 +525,7 @@
                     .addComponent(comSource, javax.swing.GroupLayout.PREFERRED_SIZE, javax.swing.GroupLayout.DEFAULT_SIZE, javax.swing.GroupLayout.PREFERRED_SIZE))
                 .addPreferredGap(javax.swing.LayoutStyle.ComponentPlacement.RELATED)
                 .addComponent(jLabel3, javax.swing.GroupLayout.PREFERRED_SIZE, 44, javax.swing.GroupLayout.PREFERRED_SIZE)
-                .addContainerGap(379, Short.MAX_VALUE))
+                .addContainerGap(javax.swing.GroupLayout.DEFAULT_SIZE, Short.MAX_VALUE))
         );
 
         pnlCards.add(pnlDependencies, "dependencies");
@@ -585,7 +585,7 @@
                 .addComponent(cbDisableIndex)
                 .addPreferredGap(javax.swing.LayoutStyle.ComponentPlacement.RELATED)
                 .addComponent(jLabel5)
-                .addContainerGap(450, Short.MAX_VALUE))
+                .addContainerGap(javax.swing.GroupLayout.DEFAULT_SIZE, Short.MAX_VALUE))
         );
 
         pnlCards.add(pnlIndex, "index");
@@ -641,7 +641,7 @@
                     .addComponent(lblDirectory)
                     .addComponent(txtDirectory, javax.swing.GroupLayout.PREFERRED_SIZE, javax.swing.GroupLayout.DEFAULT_SIZE, javax.swing.GroupLayout.PREFERRED_SIZE)
                     .addComponent(btnDirectory))
-                .addContainerGap(332, Short.MAX_VALUE))
+                .addContainerGap(javax.swing.GroupLayout.DEFAULT_SIZE, Short.MAX_VALUE))
         );
 
         pnlCards.add(plnExperimental, "experimental");
```

matthiasblaesing:
I saw this dialog the first time today:

image

What is missing IMHO is a guide how to silence this. I had to come here to realize, that he solution is as easy as going to maven settings and setting the proxy selection to "Do not check".

Spelling this out for me ("Go to Options -> Java -> Maven -> Execution -> Network proxy to change the default action") would have helped me.

PR #5044, opened 2022-12-03
added more categories for the release notes generator.

mbien:
needs more categories since the `other` section is quite large still.

going to try to update older releases after this is merged by doing some boolean algebra on release notes :)

```java
public class RetroNoter2000 {

    public static void main(String[] args) throws IOException {

        Path base = Path.of("/home/mbien/NetBeansProjects/RetroNoter2000/");

        // second last file is the release of interest
        List<String> files = List.of(
            "HEAD_16.txt",
            "HEAD_15.txt"
//            "HEAD_14.txt",
//            "HEAD_13.txt",
//            "HEAD_126.txt",
//            "HEAD_125.txt"
        );

        Set<String> filter = new HashSet<>();
        for (int i = 0; i < files.size()-1; i++) {
            filter.addAll(readPRsAsSet(base.resolve(files.get(i))));
        }

        Path previous = base.resolve(files.get(files.size()-1)); // tag before release of interest

        try (Stream<String> lines = Files.lines(previous)) {
            List<String> notes = lines.filter((line) -> !filter.contains(line)).toList();
            notes = notes.subList(0, notes.indexOf("## New Contributors"));
            notes.forEach(System.out::println);
        }
    }

    // reads only the changes as Set
    private static Set<String> readPRsAsSet(Path path) throws IOException {
        try (Stream<String> lines = Files.lines(path)) {
            return lines.filter((l) -> l.startsWith("* "))
                        .collect(Collectors.toSet());
        }
    }

}
```

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 
+# doc: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options
 changelog:
 
   exclude:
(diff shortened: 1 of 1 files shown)


mbien:
updated NB 16 release:
https://github.com/apache/netbeans/releases/tag/16

older releases will need some manual re-labeling first since many PRs are not labeled.

Issue #4553, opened 2022-08-27
"Organize imports" adds record imports from same package


jirkabs:
### Apache NetBeans version

Apache NetBeans 14

### What happened

The function "Organize imports" adds imports for records from same package. The same behavior occurs on file save if "Organize imports on save" is enabled.

Example:
```
package demo;

import demo.RecordA;
import demo.RecordC;
import demo.sub.ClassD;

public class App {

    RecordA rA;
    RecordC rC;
    private ClassD cD;

}
```
The first two imports are needless and NB also shows hint "Import From The Same Package" on that rows.

### How to reproduce

The problem only occurs if there exists another import (not matter if class or record). If the `ClassD` was removed in example then "Organize imports" would not add any imports from same package.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Linux, Ubuntu

### JDK

17

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

jirkabs:
I tested it on the actual version 15 and imports from the same package are still wrongly inserted.

PR #5046, opened 2022-12-04
Fix organize imports to not add record from same package

jirkabs:
fixes apache/netbeans#4553


mbien:
thanks for taking a look at this. We should check if there are more places where the `RECORD` enum is missing, since there was a very similar issue not long ago: #3497

another possible candidate would be the switch in `importComments` in the same class.

mbien (approved):
tests are green (travis doesn't matter anymore). Can confirm that this is fixing the linked issue. merging.

Thanks for the bugfix.

going to take a look at missing RECORD enums in a followup PR, since its missing across many other modules too.

PR #5053, opened 2022-12-06
Make Java Platform available on non-java Gradle Projects

lkishalmi:
This is a workaround. It enables non-java project (root project, or others) to have a Java Platform specified and used during project discovery.

This should be provided by the base Gradle Project support, not the Gradlle Java Project support, but that's for the future.


lkishalmi:
FYI: The platform customizer is already always enabled.

sdedic (approved):
While I think this is fine as a workaround, I have some mixed feelings in general about more and more  java-related services exposed on generic root projects - see  JavaCompilerProcessorFactory, JPDAProcessorFactory: they will now apply to (e.g.) C++ Gradle builds as well.

Maybe some other approach (or registration mechanism) is needed for umbrella / parent / root projects in this case ?

Issue #4904, opened 2022-11-02
Many tests fail on JDK 11


mbien:
There are still many tests which can't be run on JDK 11. 

NB is currently leaning into the compatibility promise of OpenJDK - it would be good to repair the tests so that the project can say with confidence that things are actually working as intended on JDK 11+.

#4884 is a dummy PR which bumps everything to 11 to generate a current list of failed jobs

feel free to re trigger the workflow as needed when updates happen

remaining list of failing modules:

```
enterprise/websvc.editor.hints
enterprise/j2ee.dd.webservice
```

------
fixed in #5138:

```
java/classfile
java/whitelist
```

fixed in #5143:
```
webcommon/javascript2.jsdoc
```

fixed in #5144:
```
ide/db.dataview
ide/schema2beans
ide/xml.wsdl.model
```

fixed in #5658:
```
php/php.editor
```

fixed in #6194:
```
java/java.openjdk.project 
```

fixed in #6369:
```
java/debugger.jpda.truffle
```

fixed in #6617
```
platform/openide.filesystems
platform/options.api
platform/openide.awt
platform/core.multiview
platform/favorites
```

fixed in #7201
```
apisupport/apisupport.ant
```

fixed in #7415
```
groovy/groovy.editor
```

fixed in #7502
```
platform/lib.uihandler
platform/netbinox
platform/openide.text
platform/openide.util.ui
```

ebarboni:
@mbien jenkins [linux jenkins](https://ci-builds.apache.org/job/Netbeans/job/netbeans-linux/lastCompletedBuild/testReport/) build on jdk 11 shows junit failure too at the same place (some are from long time ago)
for example,
org.netbeans.core.multiview.MultiViewProcessorTest.testFailsWithoutAnIcon 

I'm not sure how to make the GA nice to review, very complicated to get the failure place.

mbien:
```
java/classfile
java/whitelist
```
are fixed

mbien:
`webcommon/javascript2.jsdoc` most likely too

ebresie:
Silly question, what needs to be called or done to trigger the tests failures locally?

lkishalmi:
I usually just copy the command from the failed job, and execute that locally. Not necessary all failures are reproducible locally, but it worth to give it a try.

KacerCZ:
Failing test in `php/php.editor` is `o.n.m.php.editor.csl.FoldingTest` in method `testPHPTags()`.

[...]

PR #5061, opened 2022-12-08
rewritten 2 tests of org.openide.util.TaskTest into more readable form

lbownik:
rewritten 2 tests of org.openide.util.TaskTest into more readable form
refactored one
removed tests obsoleted by https://github.com/apache/netbeans/pull/4112

### DIFF
--- a/platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
+++ b/platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
@@ -35,7 +35,8 @@ import org.junit.Ignore;
 
 
 public class TaskTest {
-    private final static long tenMiliseconds = 10000000; // in nanosecods
+    private final static long tenMiliseconds = 10000000; // in nanoseconds
+    private final static long fiveHundredMiliseconds = 500000000; // in nanoseconds
     
     private static final Logger LOG = Logger.getLogger("org.openide.util.TaskTest");
 
@@ -115,14 +116,14 @@ public class TaskTest {
             this.executedListenerTask = t;
         });
 
-        assertFalse(task.isFinished());
+        assertFalse("Task should not have been finished.", task.isFinished());
         assertNotEquals("null", task.debug());
 
         task.run();
 
-        assertTrue(this.runHasBeenExecuted);
+        assertTrue("Runnable should have been executed.", this.runHasBeenExecuted);
         assertSame(task, this.executedListenerTask);
-        assertTrue(task.isFinished());
+        assertTrue("Task should have finished.", task.isFinished());
     }
 
     //--------------------------------------------------------------------------
@@ -141,7 +142,7 @@ public class TaskTest {
 
         task.run();
 
-        assertTrue(task.isFinished());
+        assertTrue("Task should have finished.", task.isFinished());
         assertNull(this.executedListenerTask);
     }
 
@@ -156,20 +157,20 @@ public class TaskTest {
         task.run();
 
         assertNull(this.executedListenerTask);
-        assertTrue(task.isFinished());
+        assertTrue("Task should have finished.", task.isFinished());
 
         task.addTaskListener((t) -> {
             this.executedListenerTask = t;
         });
 
         assertSame(task, this.executedListenerTask);
-        assertTrue(task.isFinished()); // still finished
+        assertTrue("Task should have finished.", task.isFinished()); // still finished
     }
     
     //--------------------------------------------------------------------------
     @Ignore("Current implementation allows null listener but then Task.run throws NPE :(")
     @Test
-    public void addTaskListener_throwsNullPointer_whenGivenNullArgument() {
+    public void addTaskListener_throwsNullPointer_whenGivenNull() {
 
         Task task = new Task(() -> {
         });
@@ -183,7 +184,7 @@ public class TaskTest {
 
     //--------------------------------------------------------------------------
     @Test
-    public void removeTaskListener_doesNothing_whenGivenNullArgument() {
+    public void removeTaskListener_doesNothing_whenGivenNull() {
 
         this.executedListenerTask = null;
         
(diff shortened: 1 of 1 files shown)


platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
```diff
@@ -260,26 +303,32 @@ public void testWaitWithTimeOutReturnsAfterTimeOutWhenTheTaskIsNotComputedAtAll
...
             }
         }
-        
-        MyTask my = new MyTask ();
-        assertTrue ("The task thinks that he is finished", my.isFinished ());
-        assertTrue ("Ok, even with timeout we got the result", my.waitFinished (1000));
-        assertEquals ("But the old waitFinished is called", 1, my.values);
+
+        MyTask my = new MyTask(); 
+        assertTrue(my.isFinished()); //The task thinks that he is finished.
```
ehsavoie, line 329: Please put a message so that the test report is easier to read.

ehsavoie (changes requested):
Please use camelCase and not snakeCase :)

platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
lbownik: I'm confused - when the test fails just looking at this line makes is obvious what went wrong
Original messages in lines 280-282 were obviously wrong as according to https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertTrue(java.lang.String,%20boolean)
they get displayed only if the assertion fails.

could You propose an example?

desruisseaux: Opening source files is tedious if the tests were not run in the IDE. So it is useful to have some explanation in the console or test reports. Especially if many tests are failing, a quick look at the error messages help to chose on which test to focus first.

ehsavoie: The issue is that when the test fails you have something like : failure line 329 in the test report. Having failure line 329 "The task should have finished" gives more context.

lbownik: Is there any more generic policy concenring assertion massages I should follow or should I put this message only in this paricular case?

ehsavoie: More a rule of the thumb: assertEquals usually display both values which usually provides some context while assertNull / notNull/true / false don't provide such context.

lbownik: fixed in https://github.com/apache/netbeans/pull/5061/commits/4ec2cba5da4d0286cea8daef002e4e095c8e6f00

ehsavoie (approved):

lbownik:
squashed

mbien:
restarting the tests now that master is fixed. Hopefully this is getting picked up here.

platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
```diff
@@ -260,26 +304,32 @@ public void testWaitWithTimeOutReturnsAfterTimeOutWhenTheTaskIsNotComputedAtAll
         
         fail ("Something wrong happened the task should wait for 1000ms but it took: " + time + "\n" + log);
     }
+    
+    //--------------------------------------------------------------------------
     @Test
-    public void testWaitOnStrangeTaskThatStartsItsExecutionInOverridenWaitFinishedMethodLikeFolderInstancesDo () throws Exception {
+    public void waitOnTask_thatStartsItsExecutionWithOverridenWaitFinishedMethod()
```
mbien, line 310: the test case methods are getting _really_ long. I personally would prefer having waitOnTask1-x  while putting the rest into the javadoc for the method.

I believe junit 5 (which we don't use) has a description annotation. This would also make it easier one day to move the doc comment into an annotation if someone decides to upgrade the testing lib.

mbien (approved):
looks good to me thanks - some good tests in here.

left some comments for you to consider, other then that its ready to merge since the tests are now also green after playing a bit whack a mole.

platform/openide.util/test/unit/src/org/openide/util/TaskTest.java
lbownik: I try to follow the convention "what_behavesHow_underWhatConditions" which I adopted from [Kevlin Henney](https://youtu.be/3frRGQGiPuc?t=2829) - it sorts incely in IDEs class view.
yes - the test methods bacome longer but also more descriptive (though putting some of this description ind annotation might be good).
the reliefing factor is that nobody call these methods directly :)

mbien: yes but this is simply too long - it doesn't scale. They also do appear in test output windows etc and this makes it hard to read. I would just give up reading after the second compound word and look at the code instead. Esp since some of the names are longer then the test code - its just noise.

`waitOnTask` with a short description if necessary makes it more useful and is also closer to the style already in other tests. Please try to find some middle ground and don't use method name as a full sentence.

lbownik: I shoretened some method names in https://github.com/apache/netbeans/pull/5061/commits/a1d38f89757c486e5c8f9a2e57d28d0a75585068
I'll try to keep a soft 50 character length limit for test method names in the future.

mbien: ok thanks. looks better. Feel free to squash and force push. Lets get this finally in.

mbien: @matthiasblaesing @neilcsmith-net are you two also ok with this long-form method naming style for tests? I won't want to be the boo man here and block this for style reasons.

lbownik: squashed

matthiasblaesing:
The names seem to be ok, but in general I don't think PRs adding tests without fixing a problem are helpful. These bind time for review and discussion, without clear benefit.

mbien:
@matthiasblaesing I agree. I believe @lkishalmi already talked about this [here](https://github.com/apache/netbeans/pull/5115#issuecomment-1363182797). Things like this can be a big time sink for reviewers since the first concern are regressions caused by the refactoring of old tests. The benefit of having nicer looking tests is minimal.
 
The project has also no lack of tests, its more a lack of resources to run them all at once and also a test reliability problem of often failing tests in some areas.
 
There are commented out tests here which probably need fixes otherwise they wouldn't be commented out in the first place:
 https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml
 
There are a whole bunch of tests which can't run on JDK 11 which need fixes (but this one is hard since the easy ones are probably already fixed):
 https://github.com/apache/netbeans/issues/4904

So there is a lot of test related work available. Its just that refactoring already working tests isn't really solving anything right now.

That being said, I know that you only want to help which is appreciated - believe me. Merging this now. @lbownik thanks!

PR #5067, opened 2022-12-10
update remote platform properties, fixes #5065.

### LINKED ISSUE #5065: Upgrade from 15 to 16 does not transfer all required remote platform info
### Apache NetBeans version

Apache NetBeans 16

### What happened

When you 1st start v16 after install it sees you have been using a previous version and asks if you want to transfer settings. I check yes. Then I try to remote run a program that previously ran. I see the build finish then nothing. I check platforms in netbeans and my remote platforms are there.

### How to reproduce

Set up a remote platform in v15 and then upgrade to 16. If you compare build.properties files you can see the remote platforms defined at the end of v15's file but missing in v16 file. I manually copied the missing lines and then V16 works as it should. The odd thing is that in V16 you see the remote platforms before doing the copy but trying to run results in only a build, no errors. After updating V16 build.properties file. Builds and runs remotely as expected. This same problem happened when I upgraded from 12.5 to 15.

Apache NetBeans 12.6 or earlier

### Operating System

Win 10

### JDK

jdk1.7.0_15

### Apache NetBeans packaging

### Anything else

_No response_

Yes

### Code of Conduct

Yes

mbien:
The properties of remote platforms are currently only updated on new platform creation.
If NB is upgraded and the config is migrated, only the xmls are copied, build.properties however is never updated with RemotePlatforms.

This updates the properties once per NB session.

`org.netbeans.modules.java.j2seplatform.UpdateTask` has a similar task which updates regular platforms. It is used to update source levels however, the fact that it extracts the properties from the xml and updates everything else too is a side effect of that.


mbien:
dev build link (expires after 7 days):
https://github.com/apache/netbeans/suites/9791343345/artifacts/470932188

java/java.j2seembedded/src/org/netbeans/modules/java/j2seembedded/platform/UpdateLogger.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.tools.ant.module.spi.AntSession;
+import org.netbeans.api.java.platform.JavaPlatformManager;
+import org.netbeans.api.project.ProjectManager;
+import org.openide.util.Exceptions;
+import org.openide.util.lookup.ServiceProvider;
+
+@ServiceProvider(service = AntLogger.class, position = 55)
+public final class UpdateLogger extends AntLogger {
+
+    private boolean updated = false;
```
lbownik, line 33: is the intention of this variable to denote:
"already tried to update" (as it is now) or
 "actually updated" then "updated = true" should be put after line 46

mbien: its only there to make sure the code only runs once. This is getting initialized when the first ant project is built or run from the IDE. Nothing else reads this field so as long it is set from somewhere within the critical section it does its job.

lbownik: :)

mbien:
going remove the NB 17 target for this issue since nobody verified that the proposed patch is solving the issue.

PR #5091, opened 2022-12-12
update OutputUtils to support stacktrace links leading to JDK files.

### LINKED ISSUE #4516: Links in output window stacktraces do not work (when they point to JDK code)
### Apache NetBeans version

Apache NetBeans 14

### What happened

The links in stacktraces to JDK code do not work with JDK 9+.

### How to reproduce

Run any code causing an exception in JDK code in a Maven project, for example:

```
package com.acme.bugs.netbeans;

import java.io.FileNotFoundException;
import java.io.FileReader;

public class StacktraceLink {
    public static void main(final String[] args) throws FileNotFoundException {
        new FileReader("nonexistent");
    }
}
```

Output:
```
--- exec-maven-plugin:3.0.0:exec (default-cli) @ Test ---
Exception in thread "main" java.io.FileNotFoundException: nonexistent (The system cannot find the file specified)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:216)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:111)
	at java.base/java.io.FileReader.<init>(FileReader.java:60)
	at com.acme.bugs.netbeans.StacktraceLink.main(StacktraceLink.java:10)
```

Clicking on a link starting with `java.base/` (or any other Java module) does not open the file and a message `Source file not found for "FileReader"`/`Source file not found for "FileInputStream"` appears in the status bar.

No / Don't know

### Operating System

Windows 10 64-bit

### JDK

OpenJDK 18.0.2 Windows x64

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

Always happens with JDK 9+.

NetBeans is running on OpenJDK 18.0.2 and the project uses an OpenJDK 18.0.2 Maven toolchain. (Maven istself is running on Java 8, but that shouldn’t matter.)

NetBeans found the sources when adding the JDK 18.0.2 platform and these files can be opened in the Debugging tab when using the debugger. The links do not work even when the matching JDK file is opened.

I believe the links worked earlier when using Java 8 or older (without the module system).

The links work with own code and Maven dependencies (if the source JAR is present).

No

### Code of Conduct

Yes

mbien:
fixes #4516

 - add active JDK to class path
 - fix the packageName parsing of the line action when modules are involved
 - open class file if no source found -> allows user to download/attach sources

second commit is for cleanup, first has the changes.

test snippet
```java
    public static void main(String[] args) throws Exception {
        // for a JDK/application trace
        new FileReader("nonexistent");

        // for a trace in 3rd party deps (guava)
        ImmutableSet.of("a").iterator().remove();
    }
```

devbuild link (expires in 7 days):
https://github.com/apache/netbeans/suites/10002051531/artifacts/485721202

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
+++ b/java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
@@ -28,6 +28,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.netbeans.api.java.classpath.ClassPath;
 import org.netbeans.api.java.queries.SourceForBinaryQuery;
 import org.netbeans.api.project.Project;
@@ -35,6 +37,7 @@ import org.netbeans.modules.maven.api.classpath.ProjectSourcesClassPathProvider;
 import org.netbeans.spi.java.classpath.support.ClassPathSupport;
 import org.openide.awt.StatusDisplayer;
 import org.openide.cookies.EditorCookie;
+import org.openide.cookies.OpenCookie;
 import org.openide.filesystems.FileObject;
 import org.openide.filesystems.URLMapper;
 import org.openide.loaders.DataObject;
@@ -174,17 +177,21 @@ public final class OutputUtils {
                 Logger.getLogger(OutputUtils.class.getName()).log(Level.WARNING, "No file found for output line {0}", ev.getLine()); // NOI18N
                 StatusDisplayer.getDefault().setStatusText(Bundle.NoSource(ev.getLine()));
                 return;
-            } 
-            
-            ClassPath classPath = getClassPath();
-            int index = sa.method.indexOf(sa.file);
-            String packageName = sa.method.substring(0, index).replace('.', '/'); //NOI18N
+            }
+
+            // example: at java.base/java.io.FileReader.<init>(FileReader.java:60)
+            int start = sa.method.indexOf('/') + 1;
+            int end = sa.method.indexOf(sa.file);
+            String packageName = sa.method.substring(start, end).replace('.', '/'); //NOI18N
             String resourceName = packageName + sa.file + ".class"; //NOI18N
+
             // issue #258546; have to check all resources. javafx unpacks all classes to target,
             // SourceForBinaryQuery then fails to find the according java file ...            
+            ClassPath classPath = getClassPath();
             List<FileObject> resources = classPath.findAllResources(resourceName);
             if (resources != null) {
-                for (FileObject resource : resources) {                    
+                // find and open source file
+                for (FileObject resource : resources) {
                     FileObject root = classPath.findOwnerRoot(resource);
                     if (root != null) {
                         URL url = URLMapper.findURL(root, URLMapper.INTERNAL);
@@ -214,8 +221,18 @@ public final class OutputUtils {
                             }
                         }
                     }
-                }                
-                StatusDisplayer.getDefault().setStatusText(Bundle.NoSource(sa.file));
+                }
+                // open class file as fallback
+                try {
+                    FileObject resource = classPath.findResource(resourceName);
+                    if (resource != null) {
+                        DataObject dao = DataObject.find(resource);
+                        OpenCookie cookie = dao.getLookup().lookup(OpenCookie.class);
+                        if (cookie != null) {
+                            cookie.open();
+                        }
+                    }
+                } catch (DataObjectNotFoundException ignore) {}
             } else {
                 StatusDisplayer.getDefault().setStatusText(Bundle.NotFound(sa.file));
             }
@@ -227,6 +244,11 @@ public final class OutputUtils {
         @Override
         public void outputLineCleared(OutputEvent ev) {
         }
+    }
+
+    private static ClassPath createProxyClassPath(Stream<ClassPath> paths) {
+        List<ClassPath> cp = paths.collect(Collectors.toList());
+        return ClassPathSupport.createProxyClassPath(cp.toArray(new ClassPath[0]));
     }
     
     private static class ProjectStacktraceOutputListener extends StacktraceOutputListener {
@@ -240,8 +262,11 @@ public final class OutputUtils {
         protected ClassPath getClassPath() {
             Project prj = ref.get();
             if(prj != null) {
-                ClassPath[] cp = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class).getProjectClassPaths(ClassPath.EXECUTE);
-                return ClassPathSupport.createProxyClassPath(cp);
+                ProjectSourcesClassPathProvider prov = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class);
+                return createProxyClassPath(
+                    Stream.concat(Stream.of(prov.getProjectClassPaths(ClassPath.EXECUTE)),
+                                  Stream.of(prov.getProjectClassPaths(ClassPath.BOOT)))
+                );
             }
             return null;
         }        
@@ -258,7 +283,9 @@ public final class OutputUtils {
         protected ClassPath getClassPath() {
             FileObject fileObject = ref.get();
             if(fileObject != null) {
-                return ClassPath.getClassPath(fileObject, ClassPath.EXECUTE);
+                return createProxyClassPath(
+                    Stream.of(ClassPath.getClassPath(fileObject, ClassPath.EXECUTE),
+                              ClassPath.getClassPath(fileObject, ClassPath.BOOT)));
             }
             return null;
         }        
--- a/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
+++ b/java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
@@ -25,8 +25,8 @@ import java.beans.PropertyChangeSupport;
 import java.io.File;
 import java.net.URI;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Iterator;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
```diff
@@ -241,7 +244,8 @@ public ProjectStacktraceOutputListener(Project project) {
         protected ClassPath getClassPath() {
             Project prj = ref.get();
             if(prj != null) {
-                ClassPath[] cp = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class).getProjectClassPaths(ClassPath.EXECUTE);
+                ProjectSourcesClassPathProvider provider = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class);
+                ClassPath[] cp = provider.getProjectClassPaths(ClassPath.BOOT);
```
neilcsmith-net, line 248: Are you sure that still works for anything _not_ in the JDK?  I would expect to have `ClassPath.BOOT` added to `ClassPath.EXECUTE` in there?

mbien: first version i tested indeed merged both but the application is already included in the boot cp.

going to step through the lazy eval lambda hell of ClassPathProviderImpl later and check if this is just a bug or by design. Maven deps is another case i have to check

mbien: it doesn't contain maven deps. Going to revert to the original version.

java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
```diff
@@ -241,8 +258,12 @@ public ProjectStacktraceOutputListener(Project project) {
         protected ClassPath getClassPath() {
             Project prj = ref.get();
             if(prj != null) {
-                ClassPath[] cp = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class).getProjectClassPaths(ClassPath.EXECUTE);
-                return ClassPathSupport.createProxyClassPath(cp);
+                ProjectSourcesClassPathProvider prov = prj.getLookup().lookup(ProjectSourcesClassPathProvider.class);
+                List<ClassPath> cp = Stream.concat(
+                        Stream.of(prov.getProjectClassPaths(ClassPath.BOOT)),
+                        Stream.of(prov.getProjectClassPaths(ClassPath.EXECUTE)))
+                    .collect(Collectors.toList());
```
lbownik, line 265: it's a reallly overcomplicated (and not very efficient) way of cancatenating arrays :)
how about using something like this

static <T> T[] concat(T[] array1, T[] array2) {
    T[] result = Arrays.copyOf(array1, array1.length + array2.length);
    System.arraycopy(array2, 0, result, array1.length, array2.length);
    return result;
}

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
```diff
@@ -672,12 +603,14 @@ public ClassPath getActiveClassPath() {
...
-                boolean hasMainModuleDescriptor = mp.getCompileSourceRoots().stream().anyMatch((sourceRoot) -> (new File(sourceRoot, MODULE_INFO_JAVA).exists()));
+                boolean hasMainModuleDescriptor = mp.getCompileSourceRoots().stream()
+                        .anyMatch((sourceRoot) -> Files.exists(Paths.get(sourceRoot, MODULE_INFO_JAVA)));
                 if(hasMainModuleDescriptor) {
                     LOGGER.log(Level.FINER, "TestPathSelector {0} for project {1}: has main module-info.java", new Object [] {logDesc, proj.getProjectDirectory().getPath()}); // NOI18N
                 }
                 
-                boolean hasTestModuleDescriptor = mp.getTestCompileSourceRoots().stream().anyMatch((testSourceRoot) -> (new File(testSourceRoot, MODULE_INFO_JAVA).exists()));
+                boolean hasTestModuleDescriptor = mp.getTestCompileSourceRoots().stream()
+                        .anyMatch((testSourceRoot) -> Files.exists(Paths.get(testSourceRoot, MODULE_INFO_JAVA)));
```
lbownik, line 613: (testSourceRoot) -> Files.exists(Paths.get(testSourceRoot, MODULE_INFO_JAVA))
this code kind of asks to be a private method with a meaningful name like "moduleInfoJavaExsists" or something

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
```diff
@@ -672,12 +603,14 @@ public ClassPath getActiveClassPath() {
             ClassPath ret = active;
             if (ret == null) {
                 MavenProject mp = proj.getOriginalMavenProject();
-                boolean hasMainModuleDescriptor = mp.getCompileSourceRoots().stream().anyMatch((sourceRoot) -> (new File(sourceRoot, MODULE_INFO_JAVA).exists()));
+                boolean hasMainModuleDescriptor = mp.getCompileSourceRoots().stream()
+                        .anyMatch((sourceRoot) -> Files.exists(Paths.get(sourceRoot, MODULE_INFO_JAVA)));
                 if(hasMainModuleDescriptor) {
                     LOGGER.log(Level.FINER, "TestPathSelector {0} for project {1}: has main module-info.java", new Object [] {logDesc, proj.getProjectDirectory().getPath()}); // NOI18N
```
lbownik, line 609: if this code is being tuched anyway, why not removing this ugly "new Object [] {" and let the compiler do it for us?

java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
mbien: i rather not start playing with indices unless there is a good reason for that. I am more concerned about correctness and readability here. 
A proper implementation will probably have to retain the module info somehow. But this should work in many if not most cases.

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
mbien: yep thats a good idea since the same snippet is used twice which is another reason to extract it.

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
mbien: the advantage of this is to delay when the toString() method is called to the point the logging lib is building the message, it doesn't make a difference in this particular case but the whole class is using this way of logging. I don't really feel the need to rewrite it all.

java/maven/src/org/netbeans/modules/maven/api/output/OutputUtils.java
mbien: the result has also just a handful of classpath objects.

java/maven/src/org/netbeans/modules/maven/classpath/ClassPathProviderImpl.java
```diff
@@ -370,7 +305,13 @@ private ClassPath getRuntimeClasspath(int type) {
...
     private ClassPath getJava8RunTimeClassPath(int type) {
         final int ftype = type == TYPE_WEB ? TYPE_SRC : type;
-        return computeIfAbsent(JAVA8_RUNTIME_PATH + ftype, () -> ClassPathFactory.createClassPath(ftype == TYPE_SRC ? new RuntimeClassPathImpl(getNBMavenProject()) : new TestRuntimeClassPathImpl(getNBMavenProject(), false)));
+        return computeIfAbsent(JAVA8_RUNTIME_PATH + ftype,
+                () -> ClassPathFactory.createClassPath(
+                        ftype == TYPE_SRC
+                            ? new RuntimeClassPathImpl(getNBMavenProject())
+                            : new TestRuntimeClassPathImpl(getNBMavenProject(), false)
+                )
+        );
```
jglick, line 314: There is no change in code here IIUC? It would be helpful to have stylistic / reformatting changes clearly marked as such and be in separate PRs, rather than intermixed with behavioral changes or refactorings.

mbien: they are in a separate commit. I am not squashing PRs if they contain a code change commit and a cleanup commit.

I don't feel its needed to create separate PRs in many cases - this does also conserve CI resources.

mbien: added do-not-squash label

mbien:
rebased to latest master

matthiasblaesing (approved):
Looks sane to me and a quick test on JDK 8 and 11 showed, that it works both with and without modules.

One thing I'm not really sure about is the order of classpaths:

```
       return createProxyClassPath(
                    Stream.concat(
                        Stream.of(prov.getProjectClassPaths(ClassPath.BOOT)),
                        Stream.of(prov.getProjectClassPaths(ClassPath.EXECUTE)))
                );
```

consider nbjavac. When it is declared as a dependency it overrides the built-in javac although normal classloading behavior would be to prefer the boot classpath. So my assumption would be, that people having dependencies clashing with JDK classes would ensure that the classloaders are setup to prefer these. But that is a niche use-case, so feel free to disagree and merge as is.

mbien:
> consider nbjavac. When it is declared as a dependency it overrides the built-in javac although normal classloading behavior would be to prefer the boot classpath. So my assumption would be, that people having dependencies clashing with JDK classes would ensure that the classloaders are setup to prefer these. But that is a niche use-case, so feel free to disagree and merge as is.

@matthiasblaesing That is a good point. I thought about this before and wasn't sure either. But I think having project-cp before boot cp might be indeed the better default for stack trace linking. Swapped it now and rebased to latest master for a fresh build.

mbien:
all green. merging.

PR #5107, opened 2022-12-13
enable some important sounding java tests

mbien:
- java/maven.hints (build tools job)
 - java.lexer (always on since fast)
 - spi.java.hints (Java label)
 - refactoring.java (Java label)
 - form (Java label)
 - junit (always on)
 - javadoc (JavaDoc label)

this required to disable some tests in `java.lexer`, `javadoc` and a single test case in `spi.java.hints`. Everything else worked fine when run directly from the IDE with JDK 19. Lets see what CI says.

### DIFF
--- a/java/java.lexer/nbproject/project.properties
+++ b/java/java.lexer/nbproject/project.properties
@@ -21,6 +21,11 @@ javac.source=1.8
 javadoc.title=Java Lexer API
 javadoc.apichanges=${basedir}/apichanges.xml
 
+# failing or missing test files
+test.config.default.excludes=\
+    **/JavaLexerPerformanceTest.class,\
+    **/JavaFlyTokensTest.class
+
 test.config.stableBTD.includes=**/*Test.class
 test.config.stableBTD.excludes=\
     **/JavaLexerBatchTest.class,\
--- a/java/javadoc/nbproject/project.properties
+++ b/java/javadoc/nbproject/project.properties
@@ -25,3 +25,6 @@ spec.version.base=1.74.0
 test.config.stableBTD.includes=\
     **/hints/AnalyzerTest.class,\
     **/search/*Test.class
+
+# failing
+test.config.default.excludes=**/search/*Test.class
--- a/java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
+++ b/java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
@@ -366,7 +366,7 @@ public class UtilitiesTest extends TestBase {
         assertTrue(errors.isEmpty());
         assertPositions(result, positions[0], code, "assert true;", "true");
     }
-
+/* TODO: fails
     public void testCasePattern() throws Exception {
         prepareTest("test/Test.java", "package test; public class Test{}");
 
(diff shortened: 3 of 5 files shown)


java/java.lexer/nbproject/project.properties
```diff
@@ -21,6 +21,11 @@ javac.source=1.8
 javadoc.title=Java Lexer API
 javadoc.apichanges=${basedir}/apichanges.xml
 
+test.config.default.excludes=\
+    **/JavaLexerPerformanceTest.class,\
+    **/JavaFlyTokensTest.class,\
+    **/JavaTokenDumpTest.class
```
mbien, line 27: those `java.lexer` tests require a test file (looks like it was the source code of JComponent) which probably has been removed before the code was donated.

java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
```diff
@@ -366,7 +366,7 @@ public void testPositionsForCorrectStatement() throws Exception {
         assertTrue(errors.isEmpty());
         assertPositions(result, positions[0], code, "assert true;", "true");
     }
-
+/* TODO: fails
     public void testCasePattern() throws Exception {
```
mbien, line 370: this one failed

java/javadoc/nbproject/project.properties
```diff
@@ -25,3 +25,6 @@ spec.version.base=1.74.0
 test.config.stableBTD.includes=\
     **/hints/AnalyzerTest.class,\
     **/search/*Test.class
+
+# failing
+test.config.default.excludes=**/search/*Test.class
```
mbien, line 30: not sure about this one. Some cases failed, but it also appeared like there were missing test files.

matthiasblaesing: Indeed we are - I compared with the old `netbeans-releases`  repository and there the javadoc index files are present.

java/java.lexer/nbproject/project.properties
matthiasblaesing: The `JavaTokenDumpTest` can be fixed like this:

```diff
--- a/java/java.lexer/test/unit/data/testfiles/testInput.java.txt.tokens.txt
+++ b/java/java.lexer/test/unit/data/testfiles/testInput.java.txt.tokens.txt
@@ -41,7 +41,7 @@
 ----- EOF -----
 
 .t.e.s.t. String Literals
-STRING_LITERAL  """"
+STRING_LITERAL  """", la=1
 WHITESPACE      " ", la=1
 STRING_LITERAL  ""a""
 WHITESPACE      "\n", la=1
```

It also looks sane. When the lexer works on `"|" "a"` (| marks read position) and reads the next char `"`, it has to decide whether this is the end of an old style string (single quote) or the second quote char of a multiline string start `"""`. So with the addition of multiline strings look-ahead became necessary.

matthiasblaesing (commented):
Looks sane to me. For `java.lexer` I just think, that one of the tests can be trivially fixed (see inline comment).

java/java.lexer/nbproject/project.properties
mbien: awesome. done.

matthiasblaesing (approved):

PR #5116, opened 2022-12-15
globally enable maven stacktrace processing.

mbien:
stack trace processing is currently only enabled for the exec maven plugin. This PR enables it globally.

Unfortunately this is all public API so I don't think we can simply rename `ExecPluginOutputListenerProvider` and call it a day. So I created a copy and replaced the original provider with the global variant.

global-stacktrace-scan

before:
imagine no links and no happy face

(don't worry this is not my pw it is a integration test)

this works best when #5099 is also applied, since it fixed some of the patterns.

### DIFF
--- a/java/maven/src/org/netbeans/modules/maven/output/DefaultOutputProcessorFactory.java
+++ b/java/maven/src/org/netbeans/modules/maven/output/DefaultOutputProcessorFactory.java
@@ -33,13 +33,13 @@ import org.openide.util.lookup.ServiceProvider;
 public class DefaultOutputProcessorFactory implements ContextOutputProcessorFactory {
     
     @Override public Set<OutputProcessor> createProcessorsSet(Project project) {
-        Set<OutputProcessor> toReturn = new HashSet<OutputProcessor>();
+        Set<OutputProcessor> toReturn = new HashSet<>();
         if (project != null) {
             toReturn.add(new JavadocOutputProcessor());
             toReturn.add(new TestOutputListenerProvider());
             toReturn.add(new SiteOutputProcessor(project));
             NbMavenProjectImpl nbprj = project.getLookup().lookup(NbMavenProjectImpl.class);
-            toReturn.add(new ExecPluginOutputListenerProvider(nbprj));
+            toReturn.add(new JavaStacktraceOutputProcessor(nbprj));
             toReturn.add(new DependencyAnalyzeOutputProcessor(nbprj));
         }
         return toReturn;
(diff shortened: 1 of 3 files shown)


java/maven/src/org/netbeans/modules/maven/output/DefaultOutputProcessorFactory.java
```diff
@@ -33,13 +33,13 @@
...
     
     @Override public Set<OutputProcessor> createProcessorsSet(Project project) {
-        Set<OutputProcessor> toReturn = new HashSet<OutputProcessor>();
+        Set<OutputProcessor> toReturn = new HashSet<>();
         if (project != null) {
             toReturn.add(new JavadocOutputProcessor());
             toReturn.add(new TestOutputListenerProvider());
             toReturn.add(new SiteOutputProcessor(project));
             NbMavenProjectImpl nbprj = project.getLookup().lookup(NbMavenProjectImpl.class);
-            toReturn.add(new ExecPluginOutputListenerProvider(nbprj));
```
jglick, line 42: At least deprecate the original?

Was about to suggest making one extend the other, to reduce duplication, though the impl is pretty small anyway.

mbien:
> Beware that whereas maven-execute-plugin normally runs a process using a classpath taken from the project’s configured dependencies

right. There is actually another pending PR #5091 which fixes a few things there. E.g it adds the JDK to the path and fixes parsing for stack traces with module names in them.

I mostly got annoyed by NB not linking traces in tests (ant project did that if i remember correctly). Initially I only wanted to enable it in selected mojos, but I couldn't see a reason to not enable it everywhere. A stacktrace is a stacktrace. It does lose it's classloader information once its logged. The IDE can only resolve those links on a best-effort basis. Links leading into JDK or program code will be correct most of the time, but they don't have to be. The application can make up classes at runtime or load them from outside of the cp.

This could be potentially improved in future by letting the user pick a source path if the link can't be resolved or maybe make this hotkey activated (ctrl+click?) in case the user notices the link resolves to the wrong line.

java/maven/src/org/netbeans/modules/maven/output/DefaultOutputProcessorFactory.java
mbien: sure I can do that if you want me to

I didn't deprecate because:
 - the class still works perfectly fine
 - I usually don't ever deprecate anything unless it can be deprecated for removal and NB doesn't really have a process for removing code (unfortunately), although it happened before e.g. with Schlieman or CND cluster etc

To reuse the code of the method I could put it into an abstract class and let both extend it. The only difference between both would be the goal array.

What would you prefer?

mbien: i deprecated ExecPluginOutputListenerProvider and made the new JavaStacktraceOutputProcessor configurable. It uses global scope by default. Ran a project wide search and there were no usages of the deprecated class.

changed the hierarchy to make it DRY

jglick (approved):

PR #5118, opened 2022-12-16
Support for OCI profiles.

sdedic:
The OCI config (`~/.oci/config` by default) may contain multiple profiles, each providing a different auth token (= different effective permissions) and/or connecting to a different tenancy in specific region. The current implementation assumed just the default profile for all operations.

I've changed the implementation so that:
- adding a OCI cloud will allow the user to select from available profiles
- a profile becomes "connected" - it gets a record in NB configuration and will display in the Services tree
- the Services view honours different auth/region/tenancy settings for different profiles displayed. 
- all operations in Services view should now operate with the appropriate profile
- one of the profiles is active, initially the default one. Active profile selection is preserved in settings.

It is possible to temporarily set an active profile; the settings will be eventually propagated through RequestProcessors tasks initiated by the wrapped execution.

ADM vulnerability scan features now accept profile to run with, and the relevant LSP audit command accepts an optional profile setting.


dbalek (approved):
Looks fine

sdedic:
Some more changes coming, sorry -- @jhorvath suggested offline that the tenancy name should be shown instead of region, asi it probably makes more sense to the user. The tenancy name however need to be fetched from the remote, since OCI profile only contains tenancy OCID

mbien:
looks like this one causes the VSCode ext job to fail. We should probably trigger that one with the LSP label too, right?

PR #5121, opened 2022-12-18
Fixing pattern matching for binding patterns

matthiasblaesing:
---

### DIFF
--- a/java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
+++ b/java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
@@ -366,7 +366,7 @@ public class UtilitiesTest extends TestBase {
         assertTrue(errors.isEmpty());
         assertPositions(result, positions[0], code, "assert true;", "true");
     }
-/* TODO: fails
+
     public void testCasePattern() throws Exception {
         prepareTest("test/Test.java", "package test; public class Test{}");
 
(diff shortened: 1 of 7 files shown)


matthiasblaesing:
@jlahoda this is an updated version of your PR #3342. The fix was pending and began to bitrot. As I think I triggered it and it was already reviewed by @jtulach I had a look and checked if I could fix the problems. I think I succeeded. Could you please double check?

@mbien you touched the tests so would be nice if you could also have a look.

jlahoda (approved):
Looks good to me - thanks!

java/spi.java.hints/test/unit/src/org/netbeans/modules/java/hints/spiimpl/UtilitiesTest.java
```diff
@@ -380,9 +380,9 @@ public void testCasePattern() throws Exception {
         String golden = "case $expr: foo bar; $stmts$; ";
         assertEquals(golden.replaceAll("[ \n\r]+", " "), result.toString().replaceAll("[ \n\r]+", " "));
         assertDiagnostics(errors, "19-19:compiler.err.expected");
-        assertPositions(result, positions[0], code, "$expr", "$stmts$", "$stmts$;", "case $expr: foo bar $stmts$;", "foo", "foo bar ");
+        assertPositions(result, positions[0], code, "$expr", "$expr", "$stmts$", "$stmts$;", "case $expr: foo bar $stmts$;", "foo", "foo bar ");
     }
-*/
```
mbien, line 385: awesome. Good to see this test case enabled.
I wanted to do that in:
https://github.com/apache/netbeans/pull/5107#discussion_r1047673049
when we enabled tests for this module (and many others) which was completely turned off till now.

but I wasn't sure if the result was actually correct. So I commented it out instead of running the risk to commit a wrong test assumption.

mbien (approved):

matthiasblaesing:
Thank you for the review. Lets get this in.

PR #5122, opened 2022-12-19
JDK 20 preparations

mbien:
add JDK 20ea to build/test matrix

use 1.8 source/target as new baseline since JDK 20 removed support for 1.7 (14 removed support for 1.6).

had to add -Xlint:-options to modules which use -Werror since the compiler would print:
```
   [repeat] warning: [options] source value 8 is obsolete and will be removed in a future release
   [repeat] warning: [options] target value 8 is obsolete and will be removed in a future release
   [repeat] warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
   [repeat] error: warnings found and -Werror specified
```
It is unclear why since this should be *release* 8. nbjavac should not set source/target anymore (unless bootclasspath is set too).

tbh I would like to move `source/target=1.8` to `release=8` to switch to the new model without all the nb specific mapping rules (e.g #3278) in follow ups.


.github/workflows/main.yml
```diff
@@ -170,7 +170,6 @@ jobs:
           distribution: ${{ env.default_java_distribution }}
 
       - name: Setup Xvfb
-        if: ${{ matrix.java != '20-ea' }}  # see #4299
```
mbien, line 173: this problem disappeared for some reason, which is a good thing

ide/o.apache.xml.resolver/build.xml
```diff
@@ -49,8 +49,7 @@
             patchfile="external/xml-resolver-1.2-netbeans.patch" />
         <!-- ... 4. Compiles sources to build/external-patch/classes -->
         <mkdir dir="build/external-patch/classes" />
-        <javac srcdir="build/external-patch/sources"
-            destdir="build/external-patch/classes" source="${javac.source}" target="${javac.target}">
+        <javac srcdir="build/external-patch/sources" destdir="build/external-patch/classes" release="8">
```
mbien, line 52: this looks weird but I hope that we can do a simple search and replace from
`release="8"` to
`release="${javac.release}"` in future once we change the format to feature version (instead of `1.x`).

I also would like to set the property by default to 8 and remove it from all module `project.properties` so that it can be managed centrally. Modules can still set it if they diverge from the default, but they don't have to set the default.

mbien:
to make sure that PRs like this here don't cause unexpected issues of compiling on the wrong bytecode level I run this script:
```bash
#!/bin/bash
set -e

max=8

for c in `find . -name "*.class" -type f`; do

    info=$(file "$c")
    version=$(echo $info | cut -d " " -f 7)

    if [[ "$version" == "" ]]; then
        echo $info
        continue
    fi

    major=$(echo $version | cut -d "." -f 1)
    minor=$(echo $version | cut -d "." -f 2)

    if (( $major > 44+$max )); then
        echo $info
        continue
    fi
    if (( $minor != 0 )); then
        echo $info
    fi

done

```
sort the output and diff before/after PR. It had no changes here.

matthiasblaesing (approved):
Change looks sane to me. I had some doubts regarding `profiler/lib.profiler` as the code is loaded into the target VM. However I looked at the generated classes before your change and it is already 8, so there is no regression.

Regarding switching to `--release=8`: If I remember correctly, this might not work for all modules, when internal classes are accessed. I agree, that it would be good to do, but it might not work.

Regarding warning about `--release=8` triggering `--source` and `--target` warnings: I expect support for `release=8` to be removed at the same time `source=8`  and `target=8` are removed. Everything else makes no sense from my perpective, as the argument to remove support for old source and targets is based on making it easier to maintain javac, which would be invalidated, if you need to retain the code paths to support `release`

mbien:
> I expect support for release=8 to be removed at the same time source=8 and target=8 are removed. Everything else makes no sense from my perpective,

@matthiasblaesing agreed. my guess is that the warning isn't about the value `8`, it is about `source`/`target` itself. Since `release` is the new best practice which is stricter and safer (unless you have to specify bootcp, then you are forced to stick to source/target, but then javac probably links against the bootcp so that would be safe too I believe).

mbien:
squashed everything. No other changes.

going to merge tomorrow once green again.

PR #5124, opened 2022-12-19
Fixing project loading issues with micronaut complex project.

sdedic:
The bug that provoked the investigation was
```
    at java.base/java.util.ArrayList.(ArrayList.java:179)
    at org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder.detectSources(NbProjectInfoBuilder.java:1012)
    at org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder.lambda$runAndRegisterPerf$9(NbProjectInfoBuilder.java:408)
    at org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder.runAndRegisterPerf(NbProjectInfoBuilder.java:414)
    at org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder.runAndRegisterPerf(NbProjectInfoBuilder.java:408)
    at org.netbeans.modules.gradle.tooling.NbProjectInfoBuilder.buildAll(NbProjectInfoBuilder.java:211)
    at org.netbeans.modules.gradle.tooling.NetBeansToolingPlugin$NetBeansToolingModelBuilder.buildAll(NetBeansToolingPlugin.java:71)
```
that was thrown when the project contained `kotlin` sources. The bug was introduced by a change in `getProperty` that stopped throwing exceptions on an unknown property, and started to return `null` instead. The PR just adapts the `detectSource` code to the new error reporting style.

During testing, I've noticed some other bad behaviour I have not seen before on certain projects:
- the task Map is not cloned on return by Gradle (it seems), so I've got a ConcurrentModificationException - apparently some of the looped-over calls lazy-populated the collection
- `null` default property value was not handled well (changed to `Optional.ofNullable`).

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -186,7 +186,7 @@ class NbProjectInfoBuilder {
 
         public ValueAndType(Class type, Object value) {
             this.type = type;
-            this.value = Optional.of(value);
+            this.value = Optional.ofNullable(value);
         }
 
         public ValueAndType(Class type) {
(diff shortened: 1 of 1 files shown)


lkishalmi (approved):
Looks good to me.

MartinBalin (approved):

matthiasblaesing (approved):
Looks sane to me

PR #5126, opened 2022-12-20
readme update

mbien:
- remove travis badge
 - update to build notes and bug report link


README.md
```diff
@@ -54,21 +52,26 @@ Build with the default config (See the [cluster.config](https://github.com/apach
...
 ```
 $ ant -Dcluster.config=full
 ```
 Build the NetBeans Platform:
 ```
 $ ant -Dcluster.config=platform
 ```
+Cleanup:
+```
+$ ant -q clean
```
lkishalmi, line 69: I would also mention `git clean -xdf` which sometimes provides faster and "cleaner" result.

mbien: i thought about this. But this is more of a git trick and has not much to do with the NetBeans project IMO.

If we wanted to support it we should call it from ant, like: `ant reset-workspace` which would be also self explanatory that it is a quite dangerous ant goal.

git clean is no joke. It cleans your sources and everything else too. `ant clean` is for the build (assuming it works).

ebarboni: is there a way to clean the external ? I often had the issue of remaining jar file in external that pollute build.

ebarboni (approved):

README.md
mbien: don't think there is a target for that. I added a external clean target for nb-javac (#3392) since this would horribly break the build if it uses wrong jars (lost at least a hour on that trying to find the issue). 

But nothing else cleans the ext folder. Would be a good idea though.

PR #5127, opened 2022-12-20
Fix Javadoc 17 parsing

neilcsmith-net:
Attempt to fix parsing of Javadoc HTML generated by Java 17, which uses `<section>` tags, so that documentation shows up in code completion if no source available.

Second commit fixes external link format when linking to a constructor or method with multiple parameters.


java/java.sourceui/test/unit/src/org/netbeans/api/java/source/ui/HTMLJavadocParserTest.java
```diff
@@ -34,40 +33,88 @@ public HTMLJavadocParserTest(String testName) {
       super(testName);
   }
 
+//  /**
+//   * Test of getJavadocText method ised with HTML produced by standard doclet.
+//   */
+//  public void testGetJavadocText() throws MalformedURLException {
+//    URL url = HTMLJavadocParserTest.class.getResource("HTMLJavadocParser.html");
```
mbien, line 40: nitpick: can you comment out large blocks with:
```
/*
...
*/
```
this makes it git friendlier / reduces potential merge conflicts. We had it before in TreeShims.java which caused many PRs to conflict which could have been avoided.

mbien:
the tests are currently disabled:
https://github.com/apache/netbeans/blob/f49a2427f6bfc1919e7233e8834dbbc791db31f1/.github/workflows/main.yml#L1338-L1339

would be good to enable them and to move them down to the JDK 11 section

whatever doesn't run can be put on the exclude list, feel free to use #5107 as blueprint.

java/java.sourceui/test/unit/src/org/netbeans/api/java/source/ui/HTMLJavadocParserTest.java
neilcsmith-net: OK, done.  Got heavy handed with `CTRL+/`, but it's easier when there's Javadocs. :smile:

neilcsmith-net:
Thanks @mbien Yes, aware the tests were disabled.  Have added an extra commit to enable with two exclusions.  FastIndexTest is failing to access JDK sources.  MarkupTagProcessorTest doesn't seem like it ever could have worked - failing equality test on a class that doesn't implement equals(). :shrug:

mbien:
this PR triggered a bug in CI. If a job times out it counts as cancelled, which makes it think that the user cancelled the workflow and the cleanup job will run which means that we can't restart the timed out job since the workspace artifact is already gone. Sorry about that.

matthiasblaesing (approved):
Looks good to me. I did not totally follow the changes in JavadocHelper, but the new tests look sane and I verified, that the problem exists and is fixed by this change. Thank you.

neilcsmith-net:
Thanks for the reviews - merging.

> I did not totally follow the changes in JavadocHelper

Yes, at some point we could probably do with not parsing HTML5 with an HTML 3.2 parser! :smile:

t-gergely:
> Yes, at some point we could probably do with not parsing HTML5 with an HTML 3.2 parser! 😄

That would be great. The current HTMLJavadocParser.parseMember method doesn't work for a lot of methods of OpenJDK 21 that don't have “a” tags in their sections. I can't follow the quirks of the current state machine (eg. hrpos and A_OPEN/A_CLOSE), so I can't produce a reliable patch./PR Should I open an issue about it? Thanks.

PR #5138, opened 2022-12-23
move java modules job to JDK 11 and fix tests where needed

mbien:
move java modules job to JDK 11 and fix tests where needed.

 - put java module system flags into project properties instead of CI config.
   - allows running the tests out-of-the-box from IDE
   - has the benefit of not overriding other JVM flag defaults for CI
 - move whole job from 8 to 11 and remove (now) redundant steps which reduces max job time by >30mins
 - ModuleTest: fix JRTFS module root for JDK 9-12.
 - remove empty ConstantPoolTest

meta issue #4904


java/java.completion/nbproject/project.properties
```diff
@@ -31,3 +31,6 @@ test.timeout=1200000
 
 # requires nb.javac for compiling of tests on Mac
 requires.nb.javac=true
+
+test.jms.flags=--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming
+test.bootclasspath.prepend.args=-Dno.netbeans.bootclasspath.prepend.needed=true
```
mbien, line 36: this has almost the same effect as the old CI config line:
```
OPTS_11=-Dtest.run.args=--limit-modules=java.base,java.logging,java.xml,java.prefs,java.desktop,java.management,java.instrument,jdk.zipfs,java.scripting,java.naming -Dtest.bootclasspath.prepend.args=-Dno.netbeans.bootclasspath.prepend.needed=true
```
the difference is that the entire run args property isn't overwritten which is set up here:
https://github.com/apache/netbeans/blob/30a8cefa5a73f258b61939729e845672e66826fe/nbbuild/templates/projectized.xml#L293


this has however the side effect of warnings like:
`WARNING: Unknown module: jdk.compiler specified to --add-exports`

I am wondering if i should set `metabuild.jms-flags.jvm` to an empty string in the project configs?

mbien: actually, i got an idea while waiting for Santa.

we can simply unset the default flags of the compiler category:
```
# remove default compiler JMS flags so that we don't get "Unknown module" warnings due to --limit-modules
jms-compiler.flags.jvm=
```

mbien:
@jlahoda would be good if you could take a look at this one.

mbien:
@neilcsmith-net I have this one set for NB17 since this reduces the longest job by about 30mins, since releasing does require a lot of building I thought it would be good to get this one in.

mbien:
thanks for the review. Lets get this in.

PR #5143, opened 2022-12-23
move webcommon job to JDK 11 and fix tests where needed.

mbien:
only a single test failed:
`JsDocCodeCompletionTest` can't use `getTestFile()` in overwritten method since the method fails the test if the file does not exist. The Test file logic in `CslTestBase::assertDescriptionMatches` has to be able to call `exists()` on a potentially not existing file when it is trying to find JDK version specific (golden) test files.

everything else in this job should be ready for 11.

meta issue: #4904

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1591,7 +1591,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        java: [ '8' ]
+        java: [ '11' ]
     steps:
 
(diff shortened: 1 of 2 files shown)


matthiasblaesing (approved):
I had to look twice why this fails, but finally saw the problematic code path (scanning through version specific completion files on JDK 9+). Thank you, looks good to me.

mbien:
@matthiasblaesing thanks for review and happy holidays! :)

PR #5144, opened 2022-12-24
move ide modules job to JDK 11 and fix tests where needed.

mbien:
- don't test JDK APIs, lets assume that they work.
 - JDK 11 fixes: javac path and java.lang.Module classname conflict.
 - fix WSDLSchemaValidatorTest assertion, some xml validation errors seem to appear twice on newer JDKs.

meta issue: #4904

### DIFF
--- a/ide/xml.wsdl.model/test/unit/src/org/netbeans/modules/xml/wsdl/validator/WSDLSchemaValidatorTest.java
+++ b/ide/xml.wsdl.model/test/unit/src/org/netbeans/modules/xml/wsdl/validator/WSDLSchemaValidatorTest.java
@@ -35,6 +35,7 @@ import org.netbeans.modules.xml.wsdl.model.WSDLModelFactory;
 import org.netbeans.modules.xml.xam.ModelSource;
 import org.netbeans.modules.xml.xam.spi.Validation;
 import org.netbeans.modules.xml.xam.spi.ValidationResult;
+import org.netbeans.modules.xml.xam.spi.Validator;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


ide/xml.wsdl.model/test/unit/src/org/netbeans/modules/xml/wsdl/validator/WSDLSchemaValidatorTest.java
```diff
@@ -167,7 +168,13 @@ private void validate(URI uri, int expectedErrorCount)
         assertNotNull(vr.getValidationResult());
         
         ValidationHelper.dumpErrors(vr);
-        assertTrue("expect error " + expectedErrorCount,  vr.getValidationResult().size() == expectedErrorCount);
+        
+        // some missing element errors appear twice on newer JDKs
```
matthiasblaesing, line 172: Could reproduce the problem in JDK 19 (17 was ok).

matthiasblaesing (approved):
Thanks - look sane to me.

ide/xml.wsdl.model/test/unit/src/org/netbeans/modules/xml/wsdl/validator/WSDLSchemaValidatorTest.java
mbien: might depend on the update release version since it is also reproducible on JDK 11 with the zulu distribution:
https://github.com/apache/netbeans/actions/runs/3761176522/jobs/6394078954#step:94:2272
```
     [junit] Testcase: testValidateMessageNoName(org.netbeans.modules.xml.wsdl.validator.WSDLSchemaValidatorTest):	FAILED
    [junit] expect error 2
    [junit] junit.framework.AssertionFailedError: expect error 2
    [junit] 	at org.netbeans.modules.xml.wsdl.validator.WSDLSchemaValidatorTest.validate(WSDLSchemaValidatorTest.java:170)

```

i vaguely remember that xml libs got an update in OpenJDK, I just don't know the backporting rules for things like that.

PR #5156, opened 2022-12-27
Locate sources JAR if it is next to binary JAR

jtulach:
A year ago I was working on a project that integrates [JavaScript, JVM & etc. into Oracle DB](https://labs.oracle.com/pls/apex/f?p=LABS:project_details:0:15). Most of the colleagues were using Eclipse or IntelliJ, but I was very happy with NetBeans VSCode extension (btw. I needed needed VSCode remote development as the machine with all the sources was over ssh across the Atlantic ocean).

Things were working OK, but over the pair programming with colleagues I realized that NetBeans lacks better integration with sources. When their IDEs could jump to a source code, my just disassembled the `.class` file!

After a while I realized that the project is carefully structured and next to each JAR there is also `-sources.jar` and the other IDEs are able to pick it up. Just NetBeans was missing such functionality. This PR is my attempt to add it. I am putting this functionality next to _classpath_ support - e.g. not associated with any project as it is a general behavior not related to Maven or Gradle or etc. I am associating late `position` with the `@ServiceRegistration` so other registered mechanisms take precedence and this is used just as a fallback.

Btw. Right now I am [working with a project](https://github.com/enso-org/enso) that uses [Scala build tool](https://www.scala-sbt.org/) to assemble Scala and Java sources. It also uses the same convention! There is `$HOME/.cache/coursier/v1/https/repo1.maven.org/maven2/org/openjdk/jmh/jmh-core/1.35/jmh-core-1.35.jar` and next to it `$HOME/.cache/coursier/v1/https/repo1.maven.org/maven2/org/openjdk/jmh/jmh-core/1.35/jmh-core-1.35-sources.jar` - looks like this is a well known convention that is beneficial to support!

### DIFF
--- a/java/java.source.base/src/org/netbeans/modules/java/source/classpath/SourceNextToBinaryQueryImpl.java
+++ b/java/java.source.base/src/org/netbeans/modules/java/source/classpath/SourceNextToBinaryQueryImpl.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


lkishalmi (approved):
Looks good to me!
I really hope that SBT improved to be a good build tool over the last 8 years. At that time it was horrible...

jtulach:
`sbt` is still horrible ;-) But not that horrible to warrant rewrite.

Issue #5154, opened 2022-12-26
NetBeans 16 doesn't redeploy project on Tomcat 9.


KempesGit:
### Apache NetBeans version

Apache NetBeans 16

### What happened

During project start I got error:

```
Starting Tomcat process...
Waiting for Tomcat...
Tomcat server started.
Incrementally deploying http://localhost:8080/Generator_Best
Completed incremental distribution of http://localhost:8080/Generator_Best
Incrementally redeploying http://localhost:8080/Generator_Best
C:\Users\artur\OneDrive\Documents\NetBeansProjects\Generator_Best\nbproject\build-impl.xml:1063: 
java.lang.NoSuchMethodError: java.net.URLEncoder.encode(Ljava/lang/String;Ljava/nio/charset/Charset;)Ljava/lang/String;
	at org.netbeans.modules.tomcat5.deploy.TomcatManagerImpl.encodePath(TomcatManagerImpl.java:336)
	at org.netbeans.modules.tomcat5.deploy.TomcatManagerImpl.remove(TomcatManagerImpl.java:232)
	at org.netbeans.modules.tomcat5.optional.TomcatIncrementalDeployment.incrementalDeploy(TomcatIncrementalDeployment.java:99)
	at org.netbeans.modules.j2ee.deployment.impl.TargetServer.deploy(TargetServer.java:706)
	at org.netbeans.modules.j2ee.deployment.devmodules.api.Deployment.deploy(Deployment.java:221)
	at org.netbeans.modules.j2ee.deployment.devmodules.api.Deployment.deploy(Deployment.java:141)
	at org.netbeans.modules.j2ee.ant.Deploy.execute(Deploy.java:86)
	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
	at sun.reflect.GeneratedMethodAccessor56.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:99)
	at org.apache.tools.ant.Task.perform(Task.java:350)
	at org.apache.tools.ant.Target.execute(Target.java:449)
	at org.apache.tools.ant.Target.performTasks(Target.java:470)
	at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1391)
[...]

Issue #5134, opened 2022-12-22
Netbeans 16 - java.lang.NoSuchMethodError: java.net.URLEncoder.encode at org.netbeans.modules.tomcat5.deploy.TomcatManagerImpl.createTempContextXml(TomcatManagerImpl.java:361)


nemanjamilosevic98:
### Apache NetBeans version

Apache NetBeans 16

### What happened

Netbeans 16, when i run project, i got:
Deploying on Apache Tomcat or TomEE
    profile mode: false
    debug mode: false
    force redeploy: true
    In-place deployment at /home/nemanja/DSProjects/main2/bet-applications/ibet-apps/ibet-s4/target/ibet-s4
    and this exception
```
java.lang.NoSuchMethodError: java.net.URLEncoder.encode(Ljava/lang/String;Ljava/nio/charset/Charset;)Ljava/lang/String;
	at org.netbeans.modules.tomcat5.deploy.TomcatManagerImpl.createTempContextXml(TomcatManagerImpl.java:361)
	at org.netbeans.modules.tomcat5.deploy.TomcatManagerImpl.initialDeploy(TomcatManagerImpl.java:196)
	at org.netbeans.modules.tomcat5.optional.TomcatIncrementalDeployment.initialDeploy(TomcatIncrementalDeployment.java:126)
	at org.netbeans.modules.j2ee.deployment.impl.TargetServer.deploy(TargetServer.java:670)
	at org.netbeans.modules.j2ee.deployment.devmodules.api.Deployment.deploy(Deployment.java:221)
	at org.netbeans.modules.maven.j2ee.execution.DeploymentHelper.perform(DeploymentHelper.java:185)
	at org.netbeans.modules.maven.j2ee.execution.ExecutionChecker.executionResult(ExecutionChecker.java:70)
	at org.netbeans.modules.maven.execute.MavenCommandLineExecutor.run(MavenCommandLineExecutor.java:320)
	at org.netbeans.core.execution.RunClassThread.doRun(RunClassThread.java:132)
	at org.netbeans.modules.openide.util.GlobalLookup.execute(GlobalLookup.java:45)
	at org.openide.util.lookup.Lookups.executeWith(Lookups.java:278)
[catch] at org.netbeans.core.execution.RunClassThread.run(RunClassThread.java:81)
```

### How to reproduce

Run project, tomcat 8.5.37

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Ubuntu 20.04

### JDK

[...]

PR #5160, opened 2022-12-29
tomcat5: Enable usage of javac release and remove internal class use (fix JDK8 and JDK17+)

matthiasblaesing:
The tomcat5 module used the internal classes:

com.sun.org.apache.xml.internal.serialize.OutputFormat
com.sun.org.apache.xml.internal.serialize.XMLSerializer

to modify the tomcat configuration. For this to work the usage of the
javac release option had to be disabled as only exported classes are
then accessible. This is done by specifying bootclasspath.prepend.

Having this masked a different issue which rendered the tomcat module
JDK 9+. The method:

URLEncoder#encode(String s, Charset enc)

was used, which was introduced with JDK9. The alternative

URLEncoder#encode(String s, String enc)

can be used as a drop-in replacement.

Closes: https://github.com/apache/netbeans/issues/5134
Closes: https://github.com/apache/netbeans/issues/5154

### DIFF
--- a/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/deploy/TomcatManagerImpl.java
+++ b/enterprise/tomcat5/src/org/netbeans/modules/tomcat5/deploy/TomcatManagerImpl.java
@@ -333,10 +333,10 @@ public class TomcatManagerImpl implements ProgressObject, Runnable {
             }
             StringBuilder result = new StringBuilder();
             while (st.hasMoreTokens()) {
-                result.append("/").append(URLEncoder.encode(st.nextToken(), StandardCharsets.UTF_8)); // NOI18N
+                result.append("/").append(URLEncoder.encode(st.nextToken(), StandardCharsets.UTF_8.name()));
             }
             return result.toString();
-        } catch (Exception e) {
+        } catch (UnsupportedEncodingException e) {
             throw new RuntimeException(e); // this should never happen
         }
     }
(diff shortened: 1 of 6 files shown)


matthiasblaesing:
Test build is available here: https://github.com/apache/netbeans/suites/10100859651/artifacts/492515866

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/deploy/TomcatManagerImpl.java
```diff
@@ -333,7 +333,7 @@ private static String encodePath(String str) {
             }
             StringBuilder result = new StringBuilder();
             while (st.hasMoreTokens()) {
-                result.append("/").append(URLEncoder.encode(st.nextToken(), StandardCharsets.UTF_8)); // NOI18N
+                result.append("/").append(URLEncoder.encode(st.nextToken(), "UTF-8")); // NOI18N
```
mbien, line 336: I am sometimes using `StandardCharsets.UTF_8.name()` to make it explicit that the exception can't occur. But this is ok too of course since everyone knows that UTF is standard.

Maybe instead of catching `Exception` and rethrowing it as RT ex, it would be better to catch `UnsupportedEncodingException` a few lines down?

mbien (approved):
looks good. Good to see more tests getting enabled.

I started working on making the enterprise job testable on JDK 11 but got stuck and ran out of time. Going to open a draft later which partially fixes a few things.

enterprise/tomcat5/src/org/netbeans/modules/tomcat5/deploy/TomcatManagerImpl.java
matthiasblaesing: Valid points - pushed an update. Will merge, once green.

Issue #4228, opened 2022-06-16
NetBeans 14 text antialiasing doesn't respect system settings on KDE Plasma


NickLion:
### Apache NetBeans version

Apache NetBeans latest daily build

### What happened

NetBeans 14 run in KDE Plasma has grayscale antialiasing, though it's subpixel system-wide.

NB: I've chosen latest build, though I'm using official 14 version, but there is no such option in selector.

### How to reproduce

1. Enter KDE Plasma session.
2. Enable subpixel antialiasing.
3. Start NetBeans.
4. Check text antialias — it's graysacle.

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Linux — openSUSE Tumbleweed x86_64, KDE Plasma

### JDK

OpenJDK 18.0.1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

It seems this behaviour was introduced with 921e5c3df8e0137cb5f22c122b0435e289693007, where for KDE session is set
`-J-Dawt.useSystemAAFontSettings=on`
And `on` means — grayscale antialiasing.

Commenting out this line or replacing it with `-J-Dawt.useSystemAAFontSettings=lcd` fixes it for me, but `lcd` won't work for those who prefer grayscale.

What was the reason of introducing that change for antialiasing? NetBeans under KDE Plasma respected system text antialiasing before. Were there some problems for other people?

If this change is really necessary it's better to check for antialiasing mode, at least `Xft.rgba`.

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

mbien:
@eirikbakke something for you :)

@Nicklion My guess is that the JVM couldn't detect the system AA mode on some distributions, thats why the property is set. The commit msg you linked seems to confirm that.
old doc here:
https://docs.oracle.com/javase/7/docs/technotes/guides/2d/flags.html#aaFonts
also mentions KDE specifically. Maybe things changed and the JVM can now detect system settings there too?

eirikbakke:
Thanks for the report! @mbien is correct; on some KDE systems (including the one I tested), JVM could not detect the system AA mode, and reverted to no anti-aliasing at all. Greyscale antialiasing was the conservative choice, as subpixel antialiasing looks bad if the wrong mode is selected (HRGB vs. HBGR).

[...]

PR #5161, opened 2022-12-29
Linux launcher: Auto-detect subpixel text rendering mode on KDE

eirikbakke:
In a previous PR (https://github.com/apache/netbeans/pull/3113), the Linux launcher was modified to fix missing font anti-aliasing on some KDE installations. This enabled greyscale anti-aliasing only, however, including on KDE installations where subpixel antialiasing was previously seen to be working. See https://github.com/apache/netbeans/issues/4228 .

This PR further enhances the auto-configuration on KDE by taking the subpixel antialiasing setting from the Xft.rgba GTK setting where available, as suggested by [NickLion](https://github.com/NickLion) in the GitHub issue above.

Tested on KDE Plasma on Ubuntu 18.04.01 in VirtualBox (triggers the awt.useSystemAAFontSettings=lcd_hrgb setting), as well as in Gnome (confirming no change in behavior here). Confirmed visually that subpixel rendering got activated in both cases (and that it was previously not enabled on KDE).

### DIFF
--- a/nb/ide.launcher/unix/netbeans
+++ b/nb/ide.launcher/unix/netbeans
@@ -206,8 +206,8 @@ case "`uname`" in
         extra_automatic_options=""
 
         # Java/AWT/Swing will correctly detect text anti-aliasing settings on
-        # GNOME, but not on KDE. Force anti-aliasing on in this case using the
-        # "awt.useSystemAAFontSettings" property, as recommended in
+        # GNOME, but not (always) on KDE. Force anti-aliasing on in this case
+        # using the "awt.useSystemAAFontSettings" property, as recommended in
         # https://bugs.openjdk.java.net/browse/JDK-6408759 . Said bug is old,
         # but the relevant logic (in sun.awt.X11.XToolkit.initXSettingsIfNeeded ,
         # and fontpath.c ) seems not to have changed significantly since then.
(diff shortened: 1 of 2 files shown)


nb/ide.launcher/unix/netbeans
```diff
@@ -222,8 +222,29 @@ case "`uname`" in
...
+                    extra_automatic_options="-J-Dawt.useSystemAAFontSettings=lcd_vrgb"
+                    ;;
+                vbgr)
+                    extra_automatic_options="-J-Dawt.useSystemAAFontSettings=lcd_vbgr"
+                    ;;
+                *)
+                    extra_automatic_options="-J-Dawt.useSystemAAFontSettings=on"
+                    ;;
+            esac
+            echo "Detected KDE; use explicit setting for font antialiasing ($extra_automatic_options)"
```
mbien, line 247: do I understand this issue correctly that:
 - JVM doesn't enable AA at all on some KDE systems, thats why NB began to set `-J-Dawt.useSystemAAFontSettings=on` only on KDE
 - on other KDE systems, `-J-Dawt.useSystemAAFontSettings=on` would pick the wrong mode, so the JVM does need even more help there, so we end up to configure everything font AA related on KDE

If this works - great. However this probably should be upstreamed to OpenJDK if possible, looks like mode probing on KDE is lacking a bit behind. The Java 6 doc already mentioned that KDE won't be detected, interesting that this didn't change since then :)

eirikbakke: * First bullet: Correct.
* Second bullet: On these "other KDE systems", the JDK (reportedly) detects the subpixel antialiasing setting just fine. But the "on" option for awt.useSystemAAFontSettings actually refers specifically to the greyscale-type antialiasing. Which is a safe but not fully optimal setting.

> However this probably should be upstreamed to OpenJDK if possible, looks like mode probing on KDE is lacking a bit behind.

Yes, it's really an old JDK deficiency. Once NetBeans has run with this patch for a while, I can submit a JDK bug reporting our experience...

mbien: > Yes, it's really an old JDK deficiency. Once NetBeans has run with this patch for a while, I can submit a JDK bug reporting our experience...

awesome!

eirikbakke: (I also did one more experiment just now, confirming that the JDK does _not_ by itself detect AA settings just because Xft.rgba is present. So I don't know the exact conditions in which the JDK would be able to detect AA settings by itself on KDE, other than NickLion's report from openSUSE Tumbleweed with KDE Plasma.)

eirikbakke: I updated the comment to change "Java/AWT/Swing will correctly detect text anti-aliasing settings on GNOME, but not on KDE" to "not (always) on KDE".

mbien (approved):

mbien:
@eirikbakke if this is ready to merge, lets merge! :)

eirikbakke:
Sure, thank you!

PR #5162, opened 2022-12-29
Remove unnecessary bootclasspath.prepend and reenable CI/CD tests for html.editor

matthiasblaesing:


lkishalmi (approved):
Looks good as far as I can tell.
Just being curious, were do we get the JS engine nowadays?

matthiasblaesing:
Thank you for your review.

For JS parsing in the context of editing a fork of graaljs is used. At some point in the past graaljs as used in NetBeans and graaljs as in the JS engine of GraalVM split. In between the GraalVM version switched license to UPL and @jtulach created a backport of the NetBeans support to that version and I added more recent Ecmascript support. It is currently developed here: https://github.com/matthiasblaesing/graaljs-nb. The aim of that fork is not to create a JS engine, only the parser is needed.

For JS execution GraalJS (the full implementation based on truffle and the graal sdk) is present in webcommon/libs.graaljs.

lkishalmi:
Thank you for the info!

PR #5164, opened 2022-12-29
Add displayname for maven POM language and order languages in options dialog

matthiasblaesing:


matthiasblaesing:
@mbien your work on maven pom hints made me look at the options panel, which made me notice the missing display name and I noticed in the past, that ordering the languages could improve usability. So this was the moment to fix this. Could you have a look?

mbien (approved):
Code looks good, the result too!

.github/workflows/main.yml
```diff
@@ -1083,6 +1083,9 @@ jobs:
       - name: platform/options.api
         run: ant $OPTS -Dvanilla.javac.exists=true -f platform/options.api test-unit
 
+      - name: platform/options.editor
+        run: ant $OPTS -Dvanilla.javac.exists=true -f platform/options.editor test-unit
+
```
mbien, line 1088: oops. This should be `ide/options.editor` and should be moved to the `ide-modules-test` job.

mbien (changes requested):

mbien (approved):

PR #5170, opened 2022-12-31
upgrade bundled maven to 3.8.7

mbien:
https://maven.apache.org/docs/3.8.7/release-notes.html

maven dropped the `commons-io` dependency, other then that everything should stay the same.

some ignored overlaps files might need adjusting, lets see what CI says.


java/maven.embedder/external/binaries-list
```diff
@@ -16,4 +16,4 @@
 # under the License.
 A2AC1CD690AB4C80DEFE7F9BCE14D35934C35CEC jdom:jdom:1.0
 5D9CE6ADD7B714B8095F0E3E396C5E9F8C5DCFEF org.apache.maven.shared:maven-dependency-tree:2.2
-3ABB90E23BE975D397BBAA43773DD0861DAE26C6 org.apache.maven:apache-maven:3.8.6:bin@zip
+4CAAED06DFADFE54EA5C0DD8806EFC0B65812081 org.apache.maven:apache-maven:3.8.7:bin@zip
```
mbien, line 19: note for later: `maven-dependency-tree:2.2` is super old. Probably worth updating in separate PR

java/maven.embedder/external/binariesembedded-list
```diff
@@ -16,43 +16,41 @@
...
-4D22A3FAA8880EFEF2E960BB8A00C2A0B351C46A;org.apache.maven:maven-builder-support:3.8.6
-CAF3154F8E2DBBA1CAD6625919F795F19C50EA69;org.apache.maven:maven-compat:3.8.6
-F945F1F19452214D360D453D9357275313F1CFD9;org.apache.maven:maven-core:3.8.6
-9E13CAA4D04B3DE8E990E68FB955C65E9D2E0D8F;org.apache.maven:maven-embedder:3.8.6
-8AD31867C493C63AC60A3FA42356A72D20F8457B;org.apache.maven:maven-model:3.8.6
-8323A726C657BECBFFC8EA807970935DDE3E3903;org.apache.maven:maven-model-builder:3.8.6
-4138E2E9B39F364902AC263888FEB7B2407A298E;org.apache.maven:maven-plugin-api:3.8.6
-C60A7640E1B829E60F8791B5EDF3CF56A6556ACA;org.apache.maven:maven-repository-metadata:3.8.6
+1637B7E8FC392E389752E79B827B883629285626;org.apache.maven:maven-artifact:3.8.7
+4D22A3FAA8880EFEF2E960BB8A00C2A0B351C46A;org.apache.maven:maven-builder-support:3.8.7
```
mbien, line 27: btw I haven't updated a single hash in this file and everything is green for some reason. @ebarboni are some CI checks missing or did I forget something important?

mbien:
getting some warnings when running maven projects:
```
--- exec-maven-plugin:3.1.0:exec (default-cli) @ mavenproject1 ---
Hello World!
Failed to notify spy org.netbeans.modules.maven.event.NbEventSpy: org/json/simple/JSONObject
Failed to notify spy org.netbeans.modules.maven.event.NbEventSpy: org/json/simple/JSONObject
Failed to notify spy org.netbeans.modules.maven.event.NbEventSpy: org/json/simple/JSONObject
------------------------------------------------------------------------
BUILD SUCCESS
------------------------------------------------------------------------
```

matthiasblaesing:
@mbien you are seeing a problem in the build process (or better the result of) `netbeans-eventspy.jar`. The error message indicates, that `org.json.simple.JSONObject` is not found. But according to the build process, it should not even be there as it is shaded to `org.netbeans.shaded.json.simple.JSONObject`:

https://github.com/apache/netbeans/blob/604cf9d0babc9ef7ceef84fbe8082341d1c8ddf1/java/maven/build.xml#L58-L65

I decompiled the resulting `NbEventSpy` class and indeed it contained the original reference, not the shaded one. Some more googling showed, that the used `jarjar`  version we use fails to work with JDK 8+ class files. An updated `jarjar` version fixes that.

Please have a look here: https://github.com/matthiasblaesing/netbeans/tree/pr-5170. And/or here: https://github.com/matthiasblaesing/netbeans/commit/d9d130932f9edc0893b5c532b99f7fb3ebd91204.

Feel free to pull it in.

java/maven/external/asm-7.0-license.txt
```diff
@@ -0,0 +1,42 @@
+Name: ASM
+Description: ASM is an all purpose Java bytecode manipulation and analysis framework
+Origin: https://asm.ow2.io/
+Version: 7.0
+License: ASM
+URL: https://asm.ow2.io/
+Files: asm-7.0.jar, asm-commons-7.0.jar
+Type: compile-time
+Comment: Used to build ueber-jar bundling for example json library into NbEventSpy
```
mbien, line 9: @matthiasblaesing did you mean _simple_ json instead of _example_?

mbien:
@matthiasblaesing oh I get it. I was trying to find where the event spy is supposed to go, found it now:
`./nbbuild/netbeans/java/maven-nblib/netbeans-eventspy.jar`
I searched in the wrong module for `JSONObject`. It makes sense now.

Is it ok to squash and add you as co-author?

java/maven.embedder/external/binariesembedded-list
mbien: updated the hashes now after I noticed there is a manual `checkhash` target.

The zip is hashed and checked as regular build step since its in the `binaries-list`. I suppose the `binariesembedded-list` isn't checked because it would be redundant? I am wondering why it exists, just to list the jars?

java/maven/external/asm-7.0-license.txt
matthiasblaesing: Yes - and to answer the other question: feel free to squash.

java/maven.embedder/external/binariesembedded-list
matthiasblaesing: The file is used to get the maven coodinates for the binaries in the zip. See here:

https://github.com/apache/netbeans/blob/6afa3a9ce1cb474cbf084ab5e7f7071c4a6eca59/nbbuild/build.xml#L249-L268

That file is then further used by the build when the maven repository is generated.

mbien: yeah I was just wondering if there is any reason to update the hashes for us. If there is: we could hook the check into the build or call it directly from the paperwork job.

If there is no reason to update the hashes since the file is only used as list of jars, we could just leave a note to not bother with it and only update artifact name and version.

matthiasblaesing (approved):
Looks sane to me and a quick(!) test showed, that it seems to work.

ebarboni (approved):
+1 to merge thanks for doing this.  Maven 4 is on the way to not sure on how transition will be easy 😄

PR #5172, opened 2023-01-01
Improving indexing for erroneous sources.

jlahoda:
Recently, I ran into a few problems with indexing massively erroneous source:
- error cleanup does not work for record components. Error should be cleaned for record components.
- if there are multiple pattern switches (preview only feature, currently) only the first may be removed from the code when the errors are being cleared, and the rest might go to `TransPatterns`, failing there. Ensuring pattern switches are cleared while clearing errors if preview is disabled.
- member references referring to methods with erroneous parameters may sometimes be kept, crashing code generator later. Such member references should be removed from the code.
- in some cases, a new source might be parsed while desugaring another, and the current cleaning code fails for that. Need to avoid attempts to clean the code before it is being desugared. (Sadly, for this, I was not able to produce a test.)

---


mbien (commented):
even though this didn't get any reviews this looks like something we should merge for NB17.

lkishalmi (approved):
I've looked at the code. It does not look bad. I'm not qualified enough to check it all. It has test. Let's merge into NB18 early and see...
 
After all, this is from @jlahoda .

mbien:
setting target to NB 17 delivery

PR #5173, opened 2023-01-01
Adding an option to disable unused hint for package private elements,  ignoring elements that possibly are looked up using MethodHandles.Lookup.

jlahoda:
Recently, I ran into to issues with the unused element hint:
- the detection of unused package private elements sometimes causes problems, e.g. when the index is (for some reason) broken. Overall, it may not be possible to make this work 100%. Suggestion here is to add an option to disable the package private unused detection, with default "enabled".
- some code (OpenJDK, specifically), has private methods that are then looked up using `MethodHandles.Lookup`. The unused detection marks them as unused, which is not quite true. The idea here is to use the method names to (heuristically) try to find out which may be looked up using the `MH.Lookup`, and not mark them as unused.

---

### DIFF
--- a/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
+++ b/java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
@@ -22,15 +22,19 @@ import com.sun.source.tree.AssignmentTree;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.CompoundAssignmentTree;
 import com.sun.source.tree.EnhancedForLoopTree;
+import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.MemberReferenceTree;
 import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.Tree.Kind;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -76,11 +80,13 @@ public class UnusedDetector {
     public static class UnusedDescription {
         public final Element unusedElement;
         public final TreePath unusedElementPath;
+        public final boolean packagePrivate;
         public final UnusedReason reason;
 
-        public UnusedDescription(Element unusedElement, TreePath unusedElementPath, UnusedReason reason) {
+        public UnusedDescription(Element unusedElement, TreePath unusedElementPath, boolean packagePrivate, UnusedReason reason) {
             this.unusedElement = unusedElement;
             this.unusedElementPath = unusedElementPath;
+            this.packagePrivate = packagePrivate;
             this.reason = reason;
         }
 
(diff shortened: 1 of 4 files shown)


java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
```diff
@@ -597,5 +615,49 @@ private void handleDeclaration(TreePath path) {
...
+                ExpressionTree clazz = node.getArguments().get(0);
+                Element lookupType = null;
+                if (clazz.getKind() == Kind.MEMBER_SELECT) {
+                    MemberSelectTree mst = (MemberSelectTree) clazz;
+                    if (mst.getIdentifier().contentEquals("class")) {
+                        lookupType = info.getTrees().getElement(new TreePath(new TreePath(getCurrentPath(), clazz), mst.getExpression()));
+                    }
+                }
+                String lookupName = null;
+                if (node.getArguments().size() > 1) {
```
lbownik, line 644: lines 640 - 645 could be extracted to a separate method

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
```diff
@@ -597,5 +619,49 @@ private void handleDeclaration(TreePath path) {
                 }
             }
         }
+
+        @Override
+        public Void visitLiteral(LiteralTree node, Void p) {
+            if (node.getKind() == Kind.STRING_LITERAL) {
+                allStringLiterals.add((String) ((LiteralTree) node).getValue());
```
matthiasblaesing, line 626: The whole PR is conclusive to me, this part I don't understand. My initial thought was, that you are trying to get the strings used in lookups, but these are already covered by the explicit invocation in `visitMethodInvocation`. Or is this to find "indirect" lookups? If so, should this customizable?

If I'm right - does this only match the unqualified name of a  method or a field?

jlahoda: Sorry for late reply. Yes, this is for indirect lookups - if the lookup contains a method name, the `allStringLiterals` should not be used. This is all a heuristics (e.g. it never uses the `MethodType`, so if there are several private overloads, they will all be considered used, even those that will never be looked up). But, this is just a warning, I think it is fine to use sensible heuristics, and I am more fine with some false negatives (not found unused methods) than with false positives (used methods marked as unused).

If the code is in a recognized pattern (`$lookup.findSomething($className.class, ...)`, the class in which the lookup is performed will be taken into consideration. If the pattern is different, the type won't be considered, and more methods will be potentially marked as used. This is again mostly a heuristics that should find a relatively typical pattern.

This should all have zero effect on files which don't use `MethodHandles.Lookup`, which is a vast majority of files, and the proposed heuristics seemed like a reasonable tradeoff to me.

matthiasblaesing: Think you for the explanation. With that in mind it makes sense to me. I overlooked the combined effect of:

https://github.com/apache/netbeans/pull/5173/files/7c57e4b4d7faf2b094e67ffab43100182f4f45bb#diff-908c30e8d355bd77779d4bfffcd966667091f27f2ab26299889fcceb659478c6R643-R662

and

https://github.com/apache/netbeans/pull/5173/files/7c57e4b4d7faf2b094e67ffab43100182f4f45bb#diff-908c30e8d355bd77779d4bfffcd966667091f27f2ab26299889fcceb659478c6R297

matthiasblaesing (approved):

matthiasblaesing:
@jlahoda should this be merged?

java/java.editor.base/src/org/netbeans/modules/java/editor/base/semantic/UnusedDetector.java
```diff
@@ -597,5 +619,49 @@ private void handleDeclaration(TreePath path) {
...
+                String lookupName = null;
+                if (node.getArguments().size() > 1) {
+                    ExpressionTree name  = node.getArguments().get(1);
+                    if (name.getKind() == Kind.STRING_LITERAL) {
+                        lookupName = (String) ((LiteralTree) name).getValue();
+                    }
+                }
+                switch (invoked.getSimpleName().toString()) {
+                    case "findStatic": case "findVirtual": case "findSpecial":
+                        type2LookedUpMethods.computeIfAbsent(lookupType, t -> new HashSet<>()).add(lookupName);
```
mbien, line 652: nitpick:
`type2LookedUpMethods.computeIfAbsent(lookupType, t -> new HashSet<>()).add(lookupName);`
->
`type2LookedUpMethods.putIfAbsent(lookupType, new HashSet<>()).add(lookupName);`
potentially clearer?
(multiple occurrences in file)

mbien: nevermind, I believe the lambda is there to allocate the AL lazily. My mistake.

PR #5174, opened 2023-01-01
For nb.org projects, use the correct nbjavac prepend for the internal (boot)classpath.

jlahoda:
When building the bootclasspath for APISupport (nb.org modules in particular), lookup the correct nbjavac API jar in the particular repository, and prepend it to the bootclasspath.

---

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
@@ -42,6 +42,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import javax.swing.event.ChangeListener;
 import org.netbeans.api.java.classpath.ClassPath;
 import org.netbeans.api.java.platform.JavaPlatform;
@@ -423,9 +424,15 @@ public final class Evaluator implements PropertyEvaluator, PropertyChangeListene
             buildDefaults.put(CP, "${module.classpath}:${cp.extra}"); // NOI18N
             buildDefaults.put(RUN_CP, "${module.run.classpath}:${cp.extra}:${build.classes.dir}"); // NOI18N
             if (type == NbModuleType.NETBEANS_ORG && "true".equals(projectProperties.getProperties().get("requires.nb.javac"))) {
-                ModuleEntry javacapi = ml.getEntry("org.netbeans.libs.javacapi");
-                if (javacapi != null) {
-                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, javacapi.getClassPathExtensions());
+                ModuleEntry javacLibrary = ml.getEntry("org.netbeans.libs.javacapi");
+                if (javacLibrary != null) {
+                    List<String> bootstrapAPIs = new ArrayList<>();
+                    for (String ext : javacLibrary.getClassPathExtensions().split(Pattern.quote(File.pathSeparator))) {
+                        if (ext.endsWith("-api.jar")) {
+                            bootstrapAPIs.add(ext);
+                        }
+                    }
+                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, bootstrapAPIs.stream().collect(Collectors.joining(File.pathSeparator)));
                 }
             }
             
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/ClassPathProviderImpl.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/ClassPathProviderImpl.java
@@ -112,6 +112,7 @@ public final class ClassPathProviderImpl implements ClassPathProvider {
                     @Override
                     public ClassPath run() {
                         if (boot == null) {
+                            ClassPathImplementation prependCP = createPathFromProperty(BOOTCLASSPATH_PREPEND);
                             final String loc = project.evaluator().getProperty(Evaluator.NBJDK_BOOTCLASSPATH_MODULAR);
                             if(loc != null) {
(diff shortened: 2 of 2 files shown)


apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
```diff
@@ -423,9 +424,15 @@ private PropertyEvaluator createEvaluator(ModuleList ml) {
...
-                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, javacapi.getClassPathExtensions());
+                ModuleEntry javacLibrary = ml.getEntry("org.netbeans.libs.javacapi");
+                if (javacLibrary != null) {
+                    List<String> bootstrapAPIs = new ArrayList<>();
+                    for (String ext : javacLibrary.getClassPathExtensions().split(Pattern.quote(File.pathSeparator))) {
+                        if (ext.endsWith("-api.jar")) {
+                            bootstrapAPIs.add(ext);
+                        }
+                    }
+                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, bootstrapAPIs.stream().collect(Collectors.joining(File.pathSeparator)));
```
lbownik, line 435: Would it be more compact to say
String path = Stream.of(javacLibrary.getClassPathExtensions().split(Pattern.quote(File.pathSeparator))).
                      filter(ext -> ext.endsWith("-api.jar")).
                      collect(Collectors.joining(File.pathSeparator)));
buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, path);

or even extract path calculation into a private method with meaningful name

matthiasblaesing (commented):
If I understand the idea correctly, the indention is, that code completion and type resolution of the IDE will work correctly for projects using the bundled javac. This is very good @jlahoda. However I checked `java.source.base` module and noticed, that only a subset of problems is gone. The module uses classes from the `com.sun.tools.javac.tree` package, which is in `nb-javac-jdk-19+33.jar`. I tested with this change and all problems are gone:

```diff
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
@@ -426,15 +426,9 @@
             if (type == NbModuleType.NETBEANS_ORG && "true".equals(projectProperties.getProperties().get("requires.nb.javac"))) {
                 ModuleEntry javacLibrary = ml.getEntry("org.netbeans.libs.javacapi");
                 if (javacLibrary != null) {
-                    List<String> bootstrapAPIs = new ArrayList<>();
-                    for (String ext : javacLibrary.getClassPathExtensions().split(Pattern.quote(File.pathSeparator))) {
-                        if (ext.endsWith("-api.jar")) {
-                            bootstrapAPIs.add(ext);
+                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, javacLibrary.getClassPathExtensions());
                         }
                     }
-                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, bootstrapAPIs.stream().collect(Collectors.joining(File.pathSeparator)));
-                }
-            }
             
             baseEval = PropertyUtils.sequentialPropertyEvaluator(predefs, providers.toArray(new PropertyProvider[providers.size()]));
```

jlahoda:
> If I understand the idea correctly, the indention is, that code completion and type resolution of the IDE will work correctly for projects using the bundled javac. This is very good @jlahoda. However I checked `java.source.base` module and noticed, that only a subset of problems is gone. The module uses classes from the `com.sun.tools.javac.tree` package, which is in `nb-javac-jdk-19+33.jar`. I tested with this change and all problems are gone:
> 
> ```diff
> --- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
> +++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/Evaluator.java
> @@ -426,15 +426,9 @@
>              if (type == NbModuleType.NETBEANS_ORG && "true".equals(projectProperties.getProperties().get("requires.nb.javac"))) {
>                  ModuleEntry javacLibrary = ml.getEntry("org.netbeans.libs.javacapi");
>                  if (javacLibrary != null) {
> -                    List<String> bootstrapAPIs = new ArrayList<>();
> -                    for (String ext : javacLibrary.getClassPathExtensions().split(Pattern.quote(File.pathSeparator))) {
> -                        if (ext.endsWith("-api.jar")) {
> -                            bootstrapAPIs.add(ext);
> +                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, javacLibrary.getClassPathExtensions());
>                          }
>                      }
> -                    buildDefaults.put(ClassPathProviderImpl.BOOTCLASSPATH_PREPEND, bootstrapAPIs.stream().collect(Collectors.joining(File.pathSeparator)));
> -                }
> -            }
>              
>              baseEval = PropertyUtils.sequentialPropertyEvaluator(predefs, providers.toArray(new PropertyProvider[providers.size()]));
> ```

That's what I don't want to do blindly - that would mean that any module could seemingly use javac internals, which does not sound right. (There are many limitations on this, so some modules have an impl. dependency even though they should not, the impl might be partially visible in the editor anyway, and the build also uses everything including the impl. OTOH, runtime is strict, and I don't think we should make the situation worse.) But, we may be able to parse the javac dependency, and if it is an impl. dependency, use even the implementation. (Of course, the very correct way would be to only use exported packages, but we don't do that in the IDE for anything.) Please see https://github.com/apache/netbeans/pull/5174/commits/441cd888642002d9e52632d08aa31c08f318bc37.

matthiasblaesing (approved):
Approach looks good to me and seems to work. Thank you!

jtulach (approved):
It would be helpful to get this fixed. Otherwise working with `javac` sources is often a bit off for me.

matthiasblaesing:
Lets get this in.

PR #5180, opened 2023-01-02
[NETBEANS-1309] add support for @summary javadoc tag

jherkel:
original issue https://issues.apache.org/jira/browse/NETBEANS-1309


mbien:
NetBeans 16 already has auto completion for the summary tag. What is this PR adding to that?

jherkel:
@mbien  With this patch you can see content of "@summary" tag when javadoc popup is displayed and also reformatting of "@summary" tag should be working.

matthiasblaesing:
@mbien I think this comparison shows the fix. Compare the contents of the javaodoc comment and the rendering.

Before:

image

After:

image

matthiasblaesing (approved):
Change looks sane to me.

mbien (approved):
awesome - code looks good. Thanks a lot for the screenshots, I was testing it but must have overlooked that detail.

Screenshots are great in general since PRs are directly linked from the release notes.

mbien:
merging since we just fixed master :)
thanks @jherkel!

PR #5189, opened 2023-01-03
platform dir: added missing @Override annotations

tbw777:
@Override annotations are used to detect changing of interfaces at compile time For example. If parent method signature is changes then javac will stop compiling becouse of difference with method signature that marked by annotation and so on.


mbien (approved):
scrolled through and looks good to me. commit header is also fine.

@tbw777 thanks.

going to wait with merge till master is green again

tbw777:
replaced by new PR: https://github.com/apache/netbeans/pull/5203

PR #5195, opened 2023-01-04
Strip protocol part, present e.g. if in KDE + manual proxy mode.

sdedic:
I've noticed bad behaviour in KDE when the system proxy is specified manually. The `kioslaverc` then contains settings like
```
ftpProxy=ftp://proxy.acme.com:80
httpProxy=http://proxy.acme.com:80
httpsProxy=http://proxy.acme.com:80
socksProxy=socks://proxy.acme.com:80
```
and the proxy string then contains not only port specification, but the protocol part as well. The error can be seen in Tools | Options, when System proxy is selected (and KDE has a manual system proxy) - the "Test connection" button action fails with hostname prefixed by `http://`.

This PR strips the PR part from the host string.

### DIFF
--- a/platform/core.network/src/org/netbeans/core/network/proxy/NetworkProxySettings.java
+++ b/platform/core.network/src/org/netbeans/core/network/proxy/NetworkProxySettings.java
@@ -157,6 +157,10 @@ public final class NetworkProxySettings {
         if (string == null) {
             return EMPTY_STRING;
         } else {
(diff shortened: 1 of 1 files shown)


matthiasblaesing (approved):
Seems reasonable. The settings format seems a bit strange and this might be better in the KDE only part, but I can imagine people using that format in the `http_proxy` environment format. There might be an issue if people call their hosts like that, but then they will be in pain anyway.

PR #5199, opened 2023-01-04
Fix #5196: use precise sha1 instead of deleted tag vm-20.3.0

### LINKED ISSUE #5196: VSNetBeans build fails on java.mx.project missing tag in build.xml
### Apache NetBeans version

Apache NetBeans 16

### What happened

VSNetBeans builder failed at https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/1389/consoleText
at bottom there is following error:
```
-checkout-graalvm:
     [exec] Cloning into 'graal'...
     [exec] warning: Could not find remote branch vm-20.3.0 to clone.
     [exec] fatal: Remote branch vm-20.3.0 not found in upstream origin

BUILD FAILED
/home/jenkins/workspace/Netbeans/netbeans-vscode/java/java.mx.project/build.xml:30: exec returned: 128
```
Checking https://github.com/oracle/graal git repo there is no [`vm-20.3.0`](https://github.com/apache/netbeans/blob/6e18d439dbcf608714abe1f5b855900db9fd6216/java/java.mx.project/build.xml#L34) tag nor branch. Similarly there is also a reference to mx branch [`5.279.0`](https://github.com/apache/netbeans/blob/6e18d439dbcf608714abe1f5b855900db9fd6216/java/java.mx.project/build.xml#L44) at the same build.xml.
The failure happened with build #1389, VSIX itself is built.

### How to reproduce

See the build on jenkins https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/1389/

No / Don't know

### Operating System

ubuntu jenkins builder

### JDK

JDK11

### Apache NetBeans packaging

Apache VSNetBeans for VSCode

### Anything else

_No response_

Yes

### Code of Conduct

Yes

sdedic:
It seems that `graalvm` project deleted the tag we'd relied on, so I've just replaced the tag name with the tag's sha1 commit id. Since it is not possible to `git clone` a specific commit, I've used an alternative technique to avoid full (big) clone -- see for example https://www.techiedelight.com/clone-git-repository-with-specific-revision/

It will be eventually better to upgrade to newer graal / mx sources.

### DIFF
--- a/java/java.mx.project/build.xml
+++ b/java/java.mx.project/build.xml
@@ -26,15 +26,31 @@
 
     <target name="-checkout-graalvm" unless="graal.dir.available">
         <delete dir="${graal.dir}"/>
-        <mkdir dir="${graal.dir}/.."/>
-        <exec dir="${graal.dir}/.." executable="git" failonerror="true">
-            <arg value="clone"/>
-            <arg value="--depth=1"/>
-            <arg value="--branch"/>
-            <arg value="vm-20.3.0"/>
-            <arg value="--single-branch"/>
+        <mkdir dir="${graal.dir}"/>
+
+        <!-- Avoid git-clone of the full Graal repository -->
+        <exec dir="${graal.dir}" executable="git" failonerror="true">
+            <arg value="init"/>
+        </exec>
+        <exec dir="${graal.dir}" executable="git" failonerror="true">
+            <arg value="remote"/>
+            <arg value="add"/>
+            <arg value="origin"/>
             <arg value="https://github.com/oracle/graal"/>
         </exec>
+        <echo message="Fetching GraalVM revision from https://github.com/oracle/graal"/>
+        <exec dir="${graal.dir}" executable="git" failonerror="true">
+            <arg value="fetch"/>
+            <arg value="origin"/>
+            <!-- The commit ID corresponds to a removed tag vm-20.3.0, by Gilles Duboscq, message "Release 20.3.0", Thu Nov 12 22:49:14 -->
+            <arg value="c5ff5a5a091bfd974640aaf6fbe51c81bd080438"/>
+        </exec>
+        <exec dir="${graal.dir}" executable="git" failonerror="true">
+            <arg value="reset"/>
+            <arg value="--hard"/>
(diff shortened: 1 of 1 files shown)


java/java.mx.project/build.xml
```diff
@@ -26,15 +26,30 @@
...
+            <arg value="remote"/>
+            <arg value="add"/>
+            <arg value="origin"/>
             <arg value="https://github.com/oracle/graal"/>
         </exec>
+        <echo message="Fetching GraalVM revision from https://github.com/oracle/graal"/>
+        <exec dir="${graal.dir}" executable="git" failonerror="true">
+            <arg value="fetch"/>
+            <arg value="origin"/>
+            <arg value="c5ff5a5a091bfd974640aaf6fbe51c81bd080438"/>
```
mbien, line 45: could be potentially useful to add a comment which states what version of graal that is.

MartinBalin (approved):

java/java.mx.project/build.xml
sdedic: Added in 232bcdc

mbien:
lets get this in so that the tests can run again.

technically master would be red right now but the mx test step is ignored on master since it fails so often. It is only required to pass on PRs, as you can see on the conditional continue-on-error:
https://github.com/apache/netbeans/blob/9ea17f3470f002c589a8ecf216f25c65a40afab3/.github/workflows/main.yml#L714-L717

jtulach:
> It will be eventually better to upgrade to newer graal / mx sources.

That's certainly the idea! Yes, please move the testing to (at least) 22.3 version to make sure NetBeans/VSCode supports the latest GraalVM sources! CCing @MartinBalin  

Btw. there shall be (and certainly used to be) a gate in GraalVM CI verifying that the latest stable version of NetBeans/IGV supports the latest GraalVM sources. As far as I know the gate was disabled in Dec 2021 when support for JDK8 was dropped. Consider re-establishing that gate, @sdedic

mbien:
does this have to build graal from source? This is currently sidelining the CI cache, but graal src+bin would be also a bit too large to put into the cache so there is probably no way around it right now - just curious.

PR #5203, opened 2023-01-04
Added all missing @Override annotations

tbw777:
@Override annotations are used to detect changing of interfaces at compile time. For example. If parent method signature is changes then javac will stop compiling because of difference with method signature that marked by annotation and so on.


mbien:
this might cause a lot of merge conflicts. I am not sure if this is a good idea to integrate this as is - its also difficult to review.

I would have preferred the already reviewed platform PR which was in an area of less pending PRs.

tbw777:
I think that conflicts of other changes can be done by authors. The current commit has no any conflicts with master

mbien:
> I think that conflicts of other changes can be done by authors. The current commit has no any conflicts with master

thats the point. We should avoid that.

tbw777:
Other way is to split this commit by dirs. That do you think?

tbw777:
Conflicts is normal in git, not understand that is the problem

tbw777:
https://github.com/apache/netbeans/pull/5189 restored, it fully back compatible with this changes. But i think that the best way to merge this main 5203 PR commit single time and forget about

tbw777:
I think this commit willnt produce any conflicts because no one method changed

tbw777:
@vieiro 
What do i need to do to accomplish all changes here?

mbien:
we are going to merge 5189 first after 17rc3 is released.
see discussion on the dev list:
https://lists.apache.org/thread/f7mt7k3rb11sgjj1fo2vlflczvw3oycx

After that we take a look how to best integrate the other PRs. We might wait with this PR until NB 17 is released.

tbw777:
ty

PR #5205, opened 2023-01-05
Removed double null check with instance operator

tbw777:
Sometimes found usage of instance operator with variable superflous  check for null. But also there is not only variables. Additional null check performs with map.get() or method() and so on.

### DIFF
--- a/platform/openide.awt/src/org/openide/awt/SplittedPanel.java
+++ b/platform/openide.awt/src/org/openide/awt/SplittedPanel.java
@@ -779,10 +779,10 @@ public class SplittedPanel extends JComponent implements Accessible {
         getAccessibleContext().setAccessibleName(
             nameFormat.format(
                 new Object[] {
-                    ((firstComponent == null) || !(firstComponent instanceof Accessible)) ? null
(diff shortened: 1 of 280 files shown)


tbw777:
This can affect performance, it's not just for beauty

platform/openide.awt/src/org/openide/awt/SplittedPanel.java
```diff
@@ -779,10 +779,10 @@ private void initAccessible() {
         getAccessibleContext().setAccessibleName(
             nameFormat.format(
                 new Object[] {
-                    ((firstComponent == null) || !(firstComponent instanceof Accessible)) ? null
+                    (!(firstComponent instanceof Accessible)) ? null
                                                                                           : firstComponent.getAccessibleContext()
                                                                                                           .getAccessibleName(),
```
mbien, line 784: could you adjust the formatting here a little bit? this could be one line now.
```java
(!(firstComponent instanceof Accessible)) ? null : firstComponent.getAccessibleContext().getAccessibleName(),
```

tbw777:
Continues in https://github.com/apache/netbeans/pull/5224 PR

PR #5206, opened 2023-01-05
Initial implementation of ANTLR4 Lexer Support

lkishalmi:
Take number 2.

This is a bunch of collected knowledge, help to integrate ANTLR v4 based Lexers into NetBeans. This PR does not include the change of the existing ANTLR v4 based languages. If this one goes through, I'd provide separate PR-s for:
 - ANTLR Language Support
 - TOML Support
 - JSON Support

I could use this on my upcoming projects for:
 - Terraform, for two or more lexers
 - GoLang Lexer
 - Maybe: HTTP Lexer, may create a new Language Support tutorial featuring that simple grammar.


ide/lexer.antlr4/test/unit/src/org/antlr/grammars/dummy/DummyLexer.java
```diff
@@ -0,0 +1,175 @@
...
+
+	@Override
+	@Deprecated
+	public String[] getTokenNames() {
+		return tokenNames;
+	}
+
+	@Override
+
+	public Vocabulary getVocabulary() {
```
mbien, line 103: free floating annotation :)

ide/lexer.antlr4/test/unit/src/org/netbeans/spi/lexer/antlr4/AntlrTokenSequenceTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    public AntlrTokenSequenceTest() {
+    }
+
+    /**
+     * Test of seekTo method, of class AntlrTokenSequence.
+     */
+    @Test
+    public void testSeekTo1() {
+        System.out.println("seekTo");
```
mbien, line 41: are the printlns needed here? Consider printing more info on assert failure but less when its green.

mbien (approved):
looks good to me. Although, the last time i wrote a grammar was over a decade ago for the NetBeans [GL](https://github.com/mbien/netbeans-opengl-pack) and [CL](https://github.com/mbien/netbeans-opencl-pack) packs - so I am not going to be able to spot bugs just by looking at it :)

consider adding the test to the CI `ide-modules-test` job:
https://github.com/apache/netbeans/blob/master/.github/workflows/main.yml#L341
 since they won't start running automatically.

ide/lexer.antlr4/test/unit/src/org/antlr/grammars/dummy/DummyLexer.java
lkishalmi: Well, that code is generated from ANTLR. I would not change that.

ide/lexer.antlr4/test/unit/src/org/netbeans/spi/lexer/antlr4/AntlrTokenSequenceTest.java
lkishalmi: This one is generated by NetBeans. I can remove them.

lkishalmi:
Well, this one does not introduce something big/groundbreaking. It's only 4 classes. Two of them: `AbstractAntlrLexerBridge` and `LexerInputCharStream` are actually implementations of an existing API. The `AbstractAntlrLexerBridge.LexerState` duties and usage is strict and well defined, that leaves `AntlrTokenSequence` as new here. That one is also simple and straight forward enough.

I could imagine that there could be bugs in the implementation, though I'd not expect the API change without keeping backward compatibility any time soon. (Unless ANTLR forces us to do so...)

PR #5208, opened 2023-01-05
Include JS implementation for JDK16+ execution environment.

sdedic:
NBLS does not bundle GraalJS implementation. In JDK16+ Nashorn was removed from the JDK and without JS implementation NBLS is not able to process PAC proxy definitions. That affects all outbound connections from NBLS, such as JDBC or OCI ADM audit requests. 

It also affects Apache NB vscode extension: since PAC processing fails and no proxy is detected, NBLS gives misleading info to the user that maven/gradle proxy is misconfigured.


MartinBalin (approved):

MartinBalin:
We need to release it in a patch release of VSNetBeans (VSIX) 16.0.301 as soon as possible along with https://github.com/apache/netbeans/pull/5207 .

dbalek (approved):

PR #5209, opened 2023-01-05
Added missing @Deprecated annotations

tbw777:
These annotations helpfully to prevent usage of deprecated API. For example: developer can see crossed out method/field name if this item is deprecated.

### DIFF
--- a/enterprise/j2ee.sun.ddui/src/org/netbeans/modules/j2ee/sun/share/configbean/SunONEDeploymentConfiguration.java
+++ b/enterprise/j2ee.sun.ddui/src/org/netbeans/modules/j2ee/sun/share/configbean/SunONEDeploymentConfiguration.java
@@ -119,15 +119,16 @@ extends GlassfishConfiguration implements DeploymentConfiguration {
         super(module, J2eeModuleHelper.getWsModuleHelper(webServerDDName), version);
     }
 
-   /**
-     * Create an instance of SunONEDeploymentConfiguration for GF V2
-     * and earlier servers.
-     *
-     * @param module  J2eeModule instance for the project represented by this config.
-     * @param mySdmi  Sun deployment manager.
-     * @deprecated Use constructor with {@link GlassFishVersion}.
-     * @throws org.netbeans.modules.j2ee.deployment.common.api.ConfigurationException
-     */
+    /**
+      * Create an instance of SunONEDeploymentConfiguration for GF V2
+      * and earlier servers.
+      *
+      * @param module  J2eeModule instance for the project represented by this config.
+      * @param mySdmi  Sun deployment manager.
+      * @deprecated Use constructor with {@link GlassFishVersion}.
+      * @throws org.netbeans.modules.j2ee.deployment.common.api.ConfigurationException
+      */
+    @Deprecated
     public SunONEDeploymentConfiguration(
             final J2eeModule module, final SunDeploymentManagerInterface mySdmi
     ) throws ConfigurationException {
@@ -143,6 +144,7 @@ extends GlassfishConfiguration implements DeploymentConfiguration {
      * @deprecated Use constructor with {@link GlassFishVersion}.
      * @throws org.netbeans.modules.j2ee.deployment.common.api.ConfigurationException
      */
+    @Deprecated
     public SunONEDeploymentConfiguration(
             final J2eeModule module, final String webServerDDName
     ) throws ConfigurationException {
--- a/ide/xml.text/src/org/netbeans/modules/xml/text/completion/ElementResultItem.java
+++ b/ide/xml.text/src/org/netbeans/modules/xml/text/completion/ElementResultItem.java
@@ -118,11 +118,11 @@ class ElementResultItem extends XMLResultItem {
         return false;
     }
     
-    /**
-     * @deprecated we use startElement flag
-     */
+//    /**
+//     * @deprecated we use startElement flag
+//     */
 //    static class EndTag extends ElementResultItem {
 //    }
-    
+
     Color getPaintColor() { return LFCustoms.shiftColor(COLOR); }
 }
--- a/contrib/j2ee.jboss4/src/org/netbeans/modules/j2ee/jboss4/config/JBossDataSourceRefModifier.java
+++ b/contrib/j2ee.jboss4/src/org/netbeans/modules/j2ee/jboss4/config/JBossDataSourceRefModifier.java
@@ -188,6 +188,7 @@ final class JBossDataSourceRefModifier {
      * 
      * @deprecated
      */
+    @Deprecated
     static void modifyMsgDrv(Jboss modifiedJboss, String resRefName, Map beans, String jndiName) {
 
         assert(beans.size() > 0);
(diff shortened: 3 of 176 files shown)


mbien:
@neilcsmith-net this one might be a candidate for NB 17, what do you think?

neilcsmith-net:
@mbien in principle should be safe.  Will also add into the sig files / tests.  Your call!

If we're sure code cleanup changes have no behavioural changes (some of these PR do) then probably safe to merge even this close to branch.  Others I would prefer we push until after NB17 is released - some could cause some "fun" with sync PRs?

mbien:
@neilcsmith-net ok i haven't considered sig tests causing trouble - good that you mention that. I was only thinking about this PR here with the deprecation annotations - I just removed all other cleanup PRs from milestone 17.

Also.. the sig test seems to throw `[sigtest] java.lang.ClassFormatError: Index out of the constant pool bounds` which it probably shouldn't do.

mbien:
@ebarboni you did recently a lot of doc related work and know also a lot more about the NB build than me. You think adding the deprecation annotation would cause any issues anywhere? I think this would be an important PR to be merged.

mbien:
@matthiasblaesing do you see any potential issues caused by bulk adding the deprecated annotation for already in-doc deprecated items? (beside more compiler warnings)

I can't come up with a good reason to not merge this for NB17.

ide/xml.text/src/org/netbeans/modules/xml/text/completion/ElementResultItem.java
```diff
@@ -123,6 +123,7 @@ public boolean substituteText( JTextComponent c, int offset, int len, int modifi
      */
 //    static class EndTag extends ElementResultItem {
 //    }
-    
+
+    @Deprecated
     Color getPaintColor() { return LFCustoms.shiftColor(COLOR); }
```
matthiasblaesing, line 128: This is wrong. The @deprecated javadoc annotation was on the `EndTag` class, nut on getPaintColor. The author commenting out the class should have removed it.

java/java.source.base/src/org/netbeans/modules/java/source/pretty/VeryPretty.java
```diff
@@ -2223,6 +2223,7 @@ private boolean isLast(JCTree tree, List<? extends JCTree> list) {
      * <li>@deprecated (see How and When To Deprecate APIs)</li>
      * </ul>
      */
+    @Deprecated
```
matthiasblaesing, line 2226: This is wrong. The `@deprecated` string in the javadoc is part of a list of tags.

java/javadoc/src/org/netbeans/modules/javadoc/hints/AddTagFix.java
```diff
@@ -144,6 +144,7 @@ protected DocTree getNewTag(TreeMaker make) {
      * <li>@deprecated (see How and When To Deprecate APIs)</li>
      * </ul>
      */
+    @Deprecated
```
matthiasblaesing, line 147: This is wrong. The @deprecated string in the javadoc is part of a list of tags.

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/ChangeParamsTransformer.java
```diff
@@ -915,6 +915,7 @@ public Boolean reduce(Boolean r1, Boolean r2) {
      * <li>@deprecated (see How and When To Deprecate APIs)</li>
      * </ul>
      */
+    @Deprecated
```
matthiasblaesing, line 918: This is wrong. The @deprecated string in the javadoc is part of a list of tags.

matthiasblaesing (changes requested):
In general ok, but there are problems (see inline comments).

For the point Neil made, this is the result (example for `editor.mimelookup`):

```diff
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/platform/editor.mimelookup/nbproject/org-netbeans-modules-editor-mimelookup.sig
+++ b/platform/editor.mimelookup/nbproject/org-netbeans-modules-editor-mimelookup.sig
@@ -1,5 +1,5 @@
 #Signature file v4.1
-#Version 1.58
+#Version 1.59
 
 CLSS public abstract interface !annotation java.lang.Deprecated
  anno 0 java.lang.annotation.Documented()
@@ -49,9 +49,13 @@
 
 CLSS public final org.netbeans.api.editor.mimelookup.MimeLookup
 meth public <%0 extends java.lang.Object> org.openide.util.Lookup$Result<{%%0}> lookup(org.openide.util.Lookup$Template<{%%0}>)
+ anno 0 java.lang.Deprecated()
 meth public <%0 extends java.lang.Object> {%%0} lookup(java.lang.Class<{%%0}>)
+ anno 0 java.lang.Deprecated()
 meth public org.netbeans.api.editor.mimelookup.MimeLookup childLookup(java.lang.String)
+ anno 0 java.lang.Deprecated()
 meth public static org.netbeans.api.editor.mimelookup.MimeLookup getMimeLookup(java.lang.String)
+ anno 0 java.lang.Deprecated()
 meth public static org.openide.util.Lookup getLookup(java.lang.String)
 meth public static org.openide.util.Lookup getLookup(org.netbeans.api.editor.mimelookup.MimePath)
 supr org.openide.util.Lookup
```

The anntoation is recorded in the signature, but I don't see a problem with that. Even when branching that should not be an issue. The additional `@Deprecated` annotations might cause more merge conflicts, but that would be true for any large scale cleanup. The size here seems doable.

tbw777:
fixed

java/refactoring.java/src/org/netbeans/modules/refactoring/java/plugins/ChangeParamsTransformer.java
```diff
@@ -912,9 +912,11 @@ public Boolean reduce(Boolean r1, Boolean r2) {
      * <li>@see</li>
      * <li>@since</li>
      * <li>@serial (or @serialField or @serialData)</li>
-     * <li>@deprecated (see How and When To Deprecate APIs)</li>
      * </ul>
+     *
+     * @deprecated (see How and When To Deprecate APIs)
      */
+    @Deprecated
```
mbien, line 919: you made it worse. This is *not* deprecated. The annotation is just mentioned in the javadoc in a list.

tbw777: fixed

enterprise/j2ee.sun.ddui/src/org/netbeans/modules/j2ee/sun/share/configbean/SunONEDeploymentConfiguration.java
```diff
@@ -128,7 +128,8 @@ public SunONEDeploymentConfiguration(
      * @deprecated Use constructor with {@link GlassFishVersion}.
      * @throws org.netbeans.modules.j2ee.deployment.common.api.ConfigurationException
      */
-    public SunONEDeploymentConfiguration(
+   @Deprecated
+   public SunONEDeploymentConfiguration(
```
mbien, line 132: one space is missing. 4 spaces per indent.

mbien (commented):
marked one formatting adjustment. After that please squash everything to one commit.

If you need instructions how to squash please say so I can help.

mbien (approved):
thanks. looked through it again and it looks good to me.

matthiasblaesing (approved):
Looks sane to me.

PR #5210, opened 2023-01-05
Use default wpad.dat if autoproxy without explicit pacfile location.

sdedic:
On Windows it is possible to enable 'automatic proxy configuration', but leave PAC file location unset - this is even the Windows' default setting. NetBeans in 'system proxy' mode rely on pacfile location to be set, if it is blank, NB defaults to 'direct' connection, which may be wrong.
This PR just defaults to the standard `http://wpad/wpad.dat` if autoproxy is enabled in the system, but no pacfile location is set.


MartinBalin (approved):

MartinBalin:
Put into next VSNetBeans release.

dbalek (approved):

PR #5214, opened 2023-01-05
Removed redundant static modifier for interface

tbw777:
No needed to use static modifier for interfaces


vieiro (commented):
Hi @tbw777 ,

Thanks for your PR. Can you elaborate on the benefit of this PR to the NetBeans project?

Thanks!

tbw777:
Updated with changes from #5215

tbw777:
>Can you elaborate on the benefit of this PR to the NetBeans project?
The project will take up less disk space, less time to compile, less time to index in the ide, less distraction for the developer with the PR installed

mbien:
how much disk space does this save?

tbw777:
@mbien 
Less then sector size

vieiro:
Hi @tbw777 ,

The fact is that the Java compiler ignores "static" in inner interfaces and enums. The bytecode is exactly the same. You can test this yourself compiling these simple classes:

```java
package a.b.c;
public class A {
  public static interface AA {}
  public static enum AAA {}
}
```

and

```java
package a.b.c;
public class B {
  public interface BB {}
  public enum BBB {}
}
```

If you compile these classes with `javac` you'll see that the generated `A.class` and `B.class` have exactly the same size.

This does not affect performance, either. Since the generated bytecode is exactly the same the performance will be the same.

The fact is that this PR makes no difference to NetBeans. There will be no difference in performance, network consumption, etc.

We may approve it, since it does some cleaning of the source code, but I will have to review those 729 files, one by one, that you have changed, to make sure all your changes are correct. This is going to take some time to me, because I don't work for this. Maybe one week or two.

During these two weeks I won't be able to do anything else, I'm afraid.

Maybe next time you send a PR you can ask in the mailing list if the PR is worth the effort reviewing, because we would prefer adding new features to NetBeans instead of reviewing PRs that make no difference in performance. What say?

tbw777:
i mean source code of course )) Ok

mbien:
@vieiro I would like to close this one. I don't see the point of this. I hope you haven't started reviewing yet :)

vieiro:
@mbien +1 in closing: can't see any advantage in the PR. Compiling >500k LOC because of this is a waste of resources.

PR #4969, opened 2022-11-15
apidoc check and improvement

ebarboni:
This PR is a modification of apidoc generation to prepare the cleanup #4450

Compared to #4450, only do a full verification if parameter apidocfullcheck is set to true. 

Disallowedlinks is anticipated fixed #4450 generate txt file to list rejected/accepted to later processing (linkchecker)
Fixed xsl to new website.
linelength 240
check all html file not only arch.
anchor regexp should also scan for id
build.properties revised to match level stability and also master/release branding dependencies

Javadoc indexer changed to json and parse more information to get interface #4699

This PR should be put in each release branch later.

Hope this is shorter to review :D and so next PR

PR #5216, opened 2023-01-05
apidoc error/warning fix

ebarboni:
Hi, this is a split of  #4450 very limited as nb17 will be cut soon. 
Url are fixed to match the allowed link merged to master via #4969 

tag samp and q are not allowed to migrated to code
comment needed to type for generics
header should be used in order so pruned use strong

hope I did understand the api enough


mbien (approved):
good stuff, thanks for extracting it from the big one :)

ebarboni:
will merge if ok, needs some attention for the release process

PR #5219, opened 2023-01-05
Replaced usage of String.to[U/L]Case() with String.equalsIgnoreCase()

tbw777:
No needed to convert case to compare strings in different case. The better in memory and cpu way is to use equalsIgnoreCase() method.

JMH  tests showing that String.equalsIgnoreCase() method faster then String.to[U/L]Case().equals() about several times.

### DIFF
--- a/webcommon/javascript.cdnjs/src/org/netbeans/modules/javascript/cdnjs/ui/SearchPanel.java
+++ b/webcommon/javascript.cdnjs/src/org/netbeans/modules/javascript/cdnjs/ui/SearchPanel.java
@@ -182,9 +182,9 @@ class SearchPanel extends javax.swing.JPanel {
      */
     private void preSelectSearchedLibrary(Library[] libraries) {
         int index = 0;
-        String term = lastSearchTerm == null ? "" : lastSearchTerm.toLowerCase(); // NOI18N
+        String term = lastSearchTerm == null ? "" : lastSearchTerm; // NOI18N
         for (int i=0; i<libraries.length; i++) {
-            if (libraries[i].getName().toLowerCase().equals(term)) {
+            if (libraries[i].getName().equalsIgnoreCase(term)) {
                 index = i;
                 break;
             }
--- a/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/GlassFishSettings.java
+++ b/enterprise/glassfish.common/src/org/netbeans/modules/glassfish/common/GlassFishSettings.java
@@ -144,7 +144,7 @@ public class GlassFishSettings {
         String showProperty
                 = System.getProperty("org.netbeans.core.WindowSystem.show");
         return showProperty == null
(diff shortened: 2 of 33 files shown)


tbw777:
I think here is no user input or data input from external sources

neilcsmith-net:
> I think here is no user input or data input from external sources

?? What is this in reference to?

tbw777:
What the potencial source of different locales?

neilcsmith-net:
to[U/L]Case() are Locale sensitive, equalsIgnoreCase() is not - check Javadoc notes for the 3. Not having behaviour that is potentially different dependent on the Locale settings of the end user is probably a good thing anyway. But this PR is a shift (fix) in behaviour as well as a possible performance improvement.

webcommon/javascript.cdnjs/src/org/netbeans/modules/javascript/cdnjs/ui/SearchPanel.java
```diff
@@ -184,7 +184,7 @@ private void preSelectSearchedLibrary(Library[] libraries) {
         int index = 0;
         String term = lastSearchTerm == null ? "" : lastSearchTerm.toLowerCase(); // NOI18N
         for (int i=0; i<libraries.length; i++) {
-            if (libraries[i].getName().toLowerCase().equals(term)) {
+            if (libraries[i].getName().equalsIgnoreCase(term)) {
```
mbien, line 187: `toLowerCase()` in L185 can be removed too

mbien:
I looked through this and I believe this should be fine @neilcsmith-net. A lot of this are language identifier comparisons, file endings, SQL queries etc. Whenever there is a comparison with a constant like "dev" it shouldn't matter if the cases are compared with a specific local or not.

The only area I am not 100% sure about is the bugzilla package - maybe you could take a look at this one. If you aren't sure either we can skip it.

neilcsmith-net:
> I looked through this and I believe this should be fine @neilcsmith-net. A lot of this are language identifier comparisons, file endings, SQL queries etc.

I think it looks so too, although I haven't had time to look at all areas (like bugzilla) yet.

To be more explicit about what I was saying above, I think this should be evaluated as a bug fix.  I'm more interested in that aspect of it than any minor performance improvements.  And we should probably look for other uses of `to[U/L]Case()` that are not part of this but questionable, and should be using `Locale.ROOT` / `Locale.ENGLISH`.

> Whenever there is a comparison with a constant like "dev" it shouldn't matter if the cases are compared with a specific local or not.

That depends on what is being compared, and where the input string is from.  Consider -

```java
String input = "LIB";
boolean result = "lib".equals(input.toLowerCase());
```

The result might be true or false depending on [whether you're Turkish or not](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#toLowerCase())!

mbien (approved):
lets get this in

PR #5222, opened 2023-01-06
platform dir: fixed some javadoc discrepancies

tbw777:
Some discrepancies are from refactoring

### DIFF
--- a/platform/api.search/src/org/netbeans/api/search/provider/impl/SimpleSearchIterator.java
+++ b/platform/api.search/src/org/netbeans/api/search/provider/impl/SimpleSearchIterator.java
@@ -178,12 +178,10 @@ public class SimpleSearchIterator extends AbstractFileObjectIterator {
      * <code>the method returns
      * <code>null</code> and no further filters are checked. If a filter returns
      * <code>TRAVERSE_ALL_SUBFOLDERS</code>, the filter is removed from the base
-     * as it needs not be applied on the folder's children. The remaining list
-     * of filters is returned as a result.
+     * as it needs not be applied on the folder's children.
      *
      * @param folder folder to compute children filters for
-     * @return list of filters to be applied on the folder's children; or
-     * <code>null</code> if the folder should not be traversed
+     * @return True if directory can be searched, false if it cannot.
      */
     private boolean checkFolderFilters(final FileObject folder) {
         assert folder.isFolder();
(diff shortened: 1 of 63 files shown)


platform/api.io/src/org/netbeans/api/io/IOProvider.java
```diff
@@ -214,7 +214,8 @@ public abstract InputOutput getIO(@NonNull String name, boolean newIO,
      *
      * @param <IO>
      * @param <OW>
-     * @param <POS>
+     * @param <P>
+     * @param <F>
```
mbien, line 218: tags without doc are useless. Either they get a description or they can be all removed.

platform/api.search/src/org/netbeans/api/search/provider/impl/SimpleSearchIterator.java
```diff
@@ -220,7 +218,7 @@ public void remove() {
 
     /**
      * Sort enumeration by fila names;
-     *
+     * @param enm
```
mbien, line 221: can be removed

platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/Utilities.java
```diff
@@ -142,6 +142,9 @@ public static Collection<KeyStore> getKeyStore (TrustLevel trustLevel) {
      *
      * @param archiveCertPath
      * @param trustedCertificates
+     * @param trustedCACertificates
+     * @param validateCertificates
+     * @param validationCACertificates
      * @return
      */
```
mbien, line 149: same here. The whole comment serves no purpose.

platform/api.io/src/org/netbeans/api/io/IOProvider.java
tbw777: fixed below

platform/api.search/src/org/netbeans/api/search/provider/impl/SimpleSearchIterator.java
tbw777: fixed below

platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/Utilities.java
tbw777: fixed below

ebarboni:
seems ok to me, @mbien ?

mbien:
@ebarboni ok to me too. I looked through it only once and not in detail.

ebarboni:
merging

PR #5224, opened 2023-01-06
Removed double null check with instance operator

tbw777:
Sometimes found usage of instance operator with variable superflous  check for null. But also there is not only variables. Additional null check performs with map.get() or method() and so on.

Also fixed formatting in org.openide.awt.SplittedPanel.java

### DIFF
--- a/platform/openide.explorer/src/org/openide/explorer/propertysheet/PropertyDialogManager.java
+++ b/platform/openide.explorer/src/org/openide/explorer/propertysheet/PropertyDialogManager.java
@@ -120,7 +120,7 @@ final class PropertyDialogManager implements VetoableChangeListener, ActionListe
         if (env != null) {
             Object helpID = env.getFeatureDescriptor().getValue(ExPropertyEditor.PROPERTY_HELP_ID);
 
-            if ((helpID != null) && helpID instanceof String && (component != null) && component instanceof JComponent) {
+            if (helpID instanceof String && (component != null) && component instanceof JComponent) {
                 HelpCtx.setHelpIDString((JComponent) component, (String) helpID);
                 helpCtx = new HelpCtx((String) helpID);
             }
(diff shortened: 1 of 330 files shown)


tbw777:
https://github.com/apache/netbeans/pull/5205 reopened with code format fix

platform/openide.explorer/src/org/openide/explorer/propertysheet/PropertyDialogManager.java
```diff
@@ -120,7 +120,7 @@ public PropertyDialogManager(
         if (env != null) {
             Object helpID = env.getFeatureDescriptor().getValue(ExPropertyEditor.PROPERTY_HELP_ID);
 
-            if ((helpID != null) && helpID instanceof String && (component != null) && component instanceof JComponent) {
+            if (helpID instanceof String && (component != null) && component instanceof JComponent) {
```
sdedic, line 123: Yet another != null && instanceof :)

sdedic (commented):
A kind request: next time, please split such huge work into more commits or better more pul requests, e.g. one per cluster.  More reviewers can work in parallel, and review of a smaller chunk is less error prone.

Looks good except the one forgotten != null not converted.

platform/openide.explorer/src/org/openide/explorer/propertysheet/PropertyDialogManager.java
tbw777: Fixed.
Im entirely agree about task splitting

sdedic (approved):
Did just review against the last state 2nd time.  No need to further split this one, I see you've already partitioned the next round of code cleanup - thanks !

This is fine by me, but I'd advise that one other reviewer would scan through the changes. As there are so many same-type changes, eyes & brain get tired and may miss something.

tbw777:
@sdedic I dont see errors. Who can review?

matthiasblaesing (approved):
Looks sane to me. Thank you. I restarted the failed CI/CD run "NetBeans / Build Tools on Linux/JDK 8 (pull_request)", which failed in a module, that is known to be flaky.

matthiasblaesing:
The `java.mx.project` suite is failing. I ran it locally on JDK 8 and it works - so lets try the next retries (this is number 8).

matthiasblaesing:
Rerunning all tests finally fixed the tests.

PR #5227, opened 2023-01-06
Moving part of regexes to compiled Pattern

tbw777:
replaceAll() without Pattern works too slowly


mbien:
closing, reasons were given in #5315

PR #5228, opened 2023-01-06
Replaced a part of replaceAll() calls to replace()

tbw777:
replace() method is more fast then replaceAll() for non regex cases


mbien (approved):
changes look good to me.

I remember skipping some less important modules in #3724, this is probably what remained.

PR #5229, opened 2023-01-06
apisupport dir: removed unessesary old forms of Integer type

tbw777:
Changed boxing/autoboxing with Integer()/valueOf()/intValue() to raw int values or parseInt(some string)

PR #5230, opened 2023-01-06
contrib dir: removed unessesary old forms of Integer type

tbw777:
Changed boxing/autoboxing with Integer()/valueOf()/intValue() to raw int values or parseInt(some string)


vieiro:
Hi @tbw777 ,

Please add this as a single commit to PR https://github.com/apache/netbeans/pull/5229

Thanks!

PR #5240, opened 2023-01-07
Upgraded Gradle Tooling API to 8.0-rc-1

lkishalmi:
Well, I hope by the time of our release Gradle 8.0 will be available. In order to prepare for that this PR is to update of Gradle's Tooling API.

Later on when Gradle 8.0 will be available, I'll provide a PR for that one as well.


lkishalmi:
@sdedic I need your help with this one. A bunch of test cases are failing. I guess some could be easily fixed if the gradle version would be set on test execution. for that it would be sufficuent to create an appropriate gradle-wrapper.properties on the test projects.

mbien:
[Gradle 8.0-rc1 release notes ](https://docs.gradle.org/8.0-rc-1/release-notes.html)for the curious

lkishalmi:
Forced a couple of test project to use Gradle 7.5 for building. It seems The Gradle tests are running well on Java 11.

lkishalmi:
And now it's green!

extide/gradle/src/org/netbeans/modules/gradle/execute/Bundle.properties
```diff
@@ -18,7 +18,7 @@
 GradleExecutorOptionsPanel.lbTasks.text=Gradle Command:
 GradleExecutorOptionsPanel.jScrollPane1.TabConstraints.tabTitle=Edit
 GradleExecutorOptionsPanel.jScrollPane2.TabConstraints.tabTitle=Preview
-ExecutionOptionsPanel.cbNoRebuild.text=Do not Rebuild Project Dependencies
+ExecutionOptionsPanel.cbNoRebuild.text=<html><strike>Do not Rebuild Project Dependencies</strike>
```
mbien, line 21: deprecated checkbox - I love it :D

mbien (approved):
looks good to me. A quick hello world test worked too (JDK 19).

I think we should merge this before freeze to get some testing in and update to the GA version during RC which would be a smaller PR.

sdedic (approved):

java/gradle.java/test/unit/data/artifacts/shadowed/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
mbien, line 5: forgot to mention that I think this is fine to downgrade specific modules.

But we should not forget about this, since they could stay behind and cause problems later. Ideally we should try to keep the versions in sync or close together.

lkishalmi: Yes, we shall keep that in our mind.

Telling the truth, these test are on the border line of unit/integration, as they are using third-party plugins. Those plugins are not necessary support the current Gradle release.

PR #5242, opened 2023-01-08
Removed StringUtils unused object creation

tbw777:
Removed unused StringBuffer object without any side effects.


matthiasblaesing (approved):
Looks sane to me.

PR #5246, opened 2023-01-09
TOML uses ANTLR4 Lexer Support.

lkishalmi:
Just taking advantage of #5206,deleting duplicated code.


mbien (approved):
good cleanup here too!

PR #5249, opened 2023-01-09
Fix possible NPE in Tiny.java

lkishalmi:
Well, I've bumped into this one today:
```
java.lang.NullPointerException: Cannot invoke "java.util.List.isEmpty()" because "type" is null
	at org.netbeans.modules.java.hints.perf.Tiny.getDestinationType(Tiny.java:537)
	at org.netbeans.modules.java.hints.perf.Tiny.unnecessaryTempFromString(Tiny.java:488)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
```

This is a trivial patch for that. I've checked the invocations of `resolveType(...)` it seems that the return value is checked for `null` at other places.


mbien (approved):
thanks for fixing my code :)

PR #4100, opened 2022-05-10
Repair failing TrufleSuiteTest.testParseTruffleSourcesWithoutError().

entlicher:
Test [org.netbeans.modules.java.mx.project.TrufleSuiteTest.testParseTruffleSourcesWithoutError](https://ci-builds.apache.org/job/Netbeans/job/netbeans-vscode/1067/testReport/org.netbeans.modules.java.mx.project/TrufleSuiteTest/testParseTruffleSourcesWithoutError/) is failing regularly. The cause was a clean action run by `SuiteActionProviderTest`, which deleted build artifacts that were necessary to assemble the class path.

PR #5254, opened 2023-01-09
ci: add retry script to ide/options.editor tests and increase mac timeout.

mbien:
checked the CI runs of the last 18 merges to master to see if we can reduce the whack-a-mole factor a bit.

Most had `java.mx.project` failing which is already wrapped in a retry script and `continue-on-error`is set to true for merges. I couldn't reproduce the failure locally, it always passed, so this one might be tricky to fix (previews attempt #4100).

But `ide/options.editor` did also fail often. Those are fairly short tests which could benefit from a retry script - which I added with this PR.

I did also increase the job timeout on mac since the mac runners can sometimes have really high latencies or artifact download speeds.

The other tests failing were in `java/refactoring.java` and `java.source.base` which would be mitigated by #5138 if it makes it in before freeze.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -554,7 +554,7 @@ jobs:
         run: ant $OPTS -f ide/o.openidex.util test
 
       - name: ide/options.editor
-        run: ant $OPTS -f ide/options.editor test
+        run: .github/retry.sh ant $OPTS -f ide/options.editor test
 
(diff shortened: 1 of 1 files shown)


matthiasblaesing (approved):
Looks sane to me. If this is not enough, we might need to mask a subset of test with @RandomlyFails.

PR #5257, opened 2023-01-09
Do not expose LexerInputCharStream as an API.

lkishalmi:
This one is in response of https://github.com/apache/netbeans/pull/5206#issuecomment-1375911582

@matthiasblaesing Thanks for your feedback!


mbien (approved):

matthiasblaesing (approved):
I have no real strong feelings here. If really a breaking change needs to be introduced to the bridge, a second implementation can be added.

Issue #5253, opened 2023-01-09
NBM implementation versions not always matching build


neilcsmith-net:
[Noticed during 16-u1 process](https://lists.apache.org/thread/8zm62rbvvw2nlv4p4qn4sm2f4x617sf2) that a significant minority (~90) of NBM modules get a date based implementation / build version that differs from the git hash used elsewhere and in the main release.

Need to fix the version, possibly by ensuring the buildnumber is passed into the sub-ant task in build-nbms.

Ideally need to understand what is causing this.  Some modules being rebuilt when they shouldn't be?

cc @ebarboni - adding this as an issue so we don't forget for NB17!

neilcsmith-net:
Looking at the logs for the 16-u1 build, it looks like during the build-nbms phase, the problematic modules are treated as out-of-date and recompiled (see `-do-compile` output).  Presumably this is due to the `up-to-date` task reporting them as stale.  Not sure what they have in common.  Something on the classpath (`cp` ref)?

ebarboni:
https://github.com/apache/netbeans/blob/67393d2c6d8dc52aadb3477ac86b40bfc2dc4fef/nbbuild/default.xml#L160
seems that metabuid hash and rawversion are not know sometime and use the hudson(jenkins) fallback

neilcsmith-net:
@ebarboni I'm just testing a fix to pass `buildnumber` into `make-nbms` sub-task.

That's only going to patch over the issue - real issue is why those modules get rebuilt at all.

neilcsmith-net:
> seems that metabuild hash and rawversion are not know sometime and use the hudson(jenkins) fallback

The reason for this (and for passing in `buildnumber`) is that individual modules don't import `nbbuild/build.xml` but `nbbuild/templates/projectized.xml`, so running a sub-ant task without inheriting properties never runs the bootstrap tasks from the main build.

I just tried building the IDE, then commenting out the `cp` check in `up-to-date` before running `make-nbms`

https://github.com/apache/netbeans/blob/67393d2c6d8dc52aadb3477ac86b40bfc2dc4fef/nbbuild/templates/common.xml#L176-L178

No module rebuilds.  I think confirms my suspicion that the `cp` ref is differently calculated, presumably again due to something not being inherited in the sub-task?

jlahoda:
[...]

PR #5262, opened 2023-01-10
Pass buildnumber into NBM sub-ant target.

neilcsmith-net:
Make sure that the minority of modules that get rebuilt during make-nbms get the required implementation version. This patches over the issue - it does not address why these modules are being marked stale and rebuilt.

(Partial!) Fix for #5253


ebarboni (approved):
seems ok to me

Issue #5252, opened 2023-01-09
Latest upgrade of Gradle plugin fails to open project with Java 17 in Windows


epliskin:
### Apache NetBeans version

Apache NetBeans 16

### What happened

After upgrading the Gradle plugin to the latest version today I could not open my Gradle project running NetBeans on Java 17 on Windows. The following error message appears:

```
java.lang.StringIndexOutOfBoundsException: begin 1, end -1, length 4
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4601)
	at java.base/java.lang.String.substring(String.java:2704)
	at org.netbeans.modules.gradle.GradleJavaCompatProblemsProvider.getJavaMajorVersion(GradleJavaCompatProblemsProvider.java:147)
	at org.netbeans.modules.gradle.GradleJavaCompatProblemsProvider.getJavaVersion(GradleJavaCompatProblemsProvider.java:128)
	at org.netbeans.modules.gradle.GradleJavaCompatProblemsProvider.getProblems(GradleJavaCompatProblemsProvider.java:91)
	at org.netbeans.spi.project.ui.support.UILookupMergerSupport$ProjectProblemsProviderImpl.getProblems(UILookupMergerSupport.java:274)
	at org.netbeans.api.project.ui.ProjectProblems.isBroken(ProjectProblems.java:58)
	at org.netbeans.modules.gradle.NbGradleProjectImpl$ProjectOpenedHookImpl.lambda$projectOpened$0(NbGradleProjectImpl.java:689)
	at org.netbeans.modules.gradle.NbGradleProjectImpl$ProjectOpenedHookImpl.projectOpened(NbGradleProjectImpl.java:696)
	at org.netbeans.spi.project.ui.ProjectOpenedHook$1.projectOpened(ProjectOpenedHook.java:60)
	at org.netbeans.spi.project.ui.support.UILookupMergerSupport$OpenHookImpl.projectOpened(UILookupMergerSupport.java:219)
	at org.netbeans.spi.project.ui.ProjectOpenedHook$1.projectOpened(ProjectOpenedHook.java:60)
[catch] at org.netbeans.modules.project.ui.OpenProjectList.notifyOpened(OpenProjectList.java:1321)
	at org.netbeans.modules.project.ui.OpenProjectList.doOpenProject(OpenProjectList.java:1402)
	at org.netbeans.modules.project.ui.OpenProjectList.open(OpenProjectList.java:848)
[...]

PR #5271, opened 2023-01-11
Fix Java version detection in GradleJavaCompatProblemProvider

lkishalmi:
Fix for #5252.

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/GradleJavaCompatProblemsProvider.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/GradleJavaCompatProblemsProvider.java
@@ -143,8 +143,13 @@ public final class GradleJavaCompatProblemsProvider implements ProjectProblemsPr
 
             }
             String javaVersion = releasePros.getProperty("JAVA_VERSION"); //NOI18N
+            // This should look like "17" or "17.0.9"
+            //TODO: Use Runtime.Version (when we move to Java 11)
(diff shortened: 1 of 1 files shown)


extide/gradle/src/org/netbeans/modules/gradle/GradleJavaCompatProblemsProvider.java
```diff
@@ -143,8 +143,12 @@ private static int getJavaMajorVersion(File javaHome) {
 
             }
             String javaVersion = releasePros.getProperty("JAVA_VERSION"); //NOI18N
+            // This should look like "17" or "17.0.9"
             if ((javaVersion != null) && javaVersion.startsWith("\"") && javaVersion.endsWith("\"")) {
-                javaVersion = javaVersion.substring(1, javaVersion.indexOf('.'));
+                int dot = javaVersion.indexOf('.');
+                javaVersion = dot > 0
+                        ? javaVersion.substring(1, javaVersion.indexOf('.'))
+                        : javaVersion.substring(1, javaVersion.length() - 1);
```
mbien, line 151: this could be using:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Runtime.Version.html#parse(java.lang.String)

once the module moved to 11 so that we don't have to come up with so many ways how to parse java versions.
Maybe mark the spot with a TODO?

mbien (approved):
looks good!

PR #5274, opened 2023-01-11
Gradle options UI cleanup (Execution and Appearance panel).

mbien:
- consistent gaps and component anchors
 - moved a few options to the global exec group

please note that the before screenshots are actually NB 16 and not master. So they have 1-2 fields less.

before (checkboxes in upper panel are all over the place + panel gaps + some exec options not in the lower panel):
gradle-before

after:
gradle-after5

before (left gaps on checkboxes):
gradle-app-before

after (note: new Maven):
gradle-app-after


lkishalmi (changes requested):
@mbien Thanks for this PR. Generally it looks good. I'd leave out the Prefer Maven Projects check box out of Gradle Execution as that is really an IDE behavior.When a folder has `pom.xml` next to gradle build files, then the `GradleProjectFactory` let's `MavenProjectFactory` to take over.

I would put that as a distinct option at the bottom right corner.

mbien:
@lkishalmi tbh I don't quite like how this looks. Can I wrap this into an "Other Options" panel?
gradle-after2

other options panel:
gradle-after3

lkishalmi:
Well, I actually like the first new version. The second one looks too much for me.
I'd offer a third option then. Let's create a `Maven` specific panel and put that option there.

There might be other new options in that area...

mbien:
another proposal (all checkboxes aligned in left column):
gradle-after4

swpalmer:
The prefer Maven over Gradle option really isn't an option of Maven or Gradle.  It needs to be in a more general location, possibly on the general tab, even though (currently) these are Java-related.  Theoretically Gradle projects can support C++ as well so if that was ever pursued things could get messier still.
Also the tab order is a little odd, Ant, Gradle, and Maven are related build systems, they should be grouped together in the tab order.  I'm not sure if that is easy to do or if the order is somewhat ad-hoc because everything is a plugin or something like that.

mbien:
I only wanted to fix some layout issues in the UI, consistent gaps, resizing etc, Moving things around to entirely different option panels would be more than just layout adjustments and something for another PR.

lkishalmi:
@swpalmer That would be a major re-arch. We should create a Build Tools section in the Options Dialog.

mbien:
@lkishalmi updated the `after` image on the PR with the latest change.
 - added Maven card as suggested and moved the "Prefer Maven" checkbox over
 - kept the left aligned checkboxes of the screenshot above
 - some more tweaking for gap consistency between all cards

lkishalmi (approved):
Looks good now.

Issue #5266, opened 2023-01-10
"Extract source name from .class and use it to locate proper source" (#5152) breaks java source editing


matthiasblaesing:
### Body

Steps to reproduce:

1. Create a clone of the JNA repository. In a directory of your choosing run `git clone https://github.com/java-native-access/jna.git`
2. In your NetBeans Source Tree go to the last commit before your merge: `git checkout c07ee007fd18`
3. Ensure directory is clean: `git clean -f -x -d`
4. Build: `ant`
5. Run tryme: `ant tryme`
6. Open the JNA NetBeans project from the folder you cloned in step 1.
7. Open test file NativeLibraryTest.java (Package com.sun.jna), go to line 91, right click on `disposeAll`, choose  Navigate -> Goto Source
8. Observe, that you are taken to the Method `disposeAll` in `NativeLibrary.java`
9. Close NetBeans
10. Checkout your merge commit: `git checkout 2a9e190afaf799bd0c126f0ffc6c70c9f0f9a305`
11. Build: `ant`
12. Run tryme: `ant tryme`
13. Repeat Step 7
14. Observe, that you are **not** taken to the source, but an exception is reported:
```
java.lang.NullPointerException
	at org.netbeans.api.java.source.SourceUtils$Match.apply(SourceUtils.java:686)
	at org.netbeans.api.java.source.SourceUtils.findMatchingChild(SourceUtils.java:631)
	at org.netbeans.api.java.source.SourceUtils.findSourceForBinary(SourceUtils.java:614)
	at org.netbeans.api.java.source.SourceUtils.getFile(SourceUtils.java:582)
	at org.netbeans.api.java.source.ui.ElementOpen.getOpenInfo(ElementOpen.java:444)
	at org.netbeans.api.java.source.ui.ElementOpen.open(ElementOpen.java:125)
	at org.netbeans.api.java.source.ui.ElementOpen.open(ElementOpen.java:120)
	at org.netbeans.modules.editor.java.GoToSupport$7.open(GoToSupport.java:1263)
	at org.netbeans.modules.editor.java.GoToSupport.performGoToImpl(GoToSupport.java:285)
	at org.netbeans.modules.editor.java.GoToSupport.access$100(GoToSupport.java:125)
	at org.netbeans.modules.editor.java.GoToSupport$3.run(GoToSupport.java:224)
	at org.netbeans.modules.progress.ui.RunOffEDTImpl$1.run(RunOffEDTImpl.java:146)
[...]

PR #5152, opened 2022-12-26
Extract source name from .class and use it to locate proper source

jtulach:
I am working on a code base that is mixing [Scala and Java](https://github.com/enso-org/enso/tree/develop/engine/runtime). Often Java sources are referencing Scala classes. I'd like the NetBeans IDE to work with it a bit. Luckily there is information in every `.class` file about the name of the source file it comes from. This PR extracts that information and passes it to `ElementOpen.open` to use it in case regular search method used for Java sources fails.

With this PR I am able to jump from `.java` file to `.scala` file when Ctrl-Click on a class name.

PR #5277, opened 2023-01-11
Revert PR "Extract source name from .class and use it to locate proper source" (#5152)

matthiasblaesing:
Revert "Removing unused import"

This reverts commit 1ad59dc0a8856d36feb7e452c8e5135d1d139aec.

Revert "Introducing SourceUtils.findSourceFileName method that needs no reflection"

This reverts commit e7d543795366b9fc817b50973a698aa2e7ba4b26.

Revert "Use the suggested names also when navigating via LSP in VSCode"

This reverts commit 7e73d76b6b271b5662981231ecc58d56bcbd5e72.

Revert "Extracts source name from .class and uses it to locate proper source file in ElementOpen.open"

This reverts commit 93f6c381fd51f3c39e86def0e02f612921c16337.

Closes: #5266


matthiasblaesing:
This should only be merged if a proper solution can't be found before release branch forking.

matthiasblaesing:
Something went wrong when creating the revert. Will close for now, #5279 might be the real fix.

PR #5279, opened 2023-01-11
Prevent NullPointerException when matching filenames in SourceUtils

matthiasblaesing:
Closes: #5266

### DIFF
--- a/java/java.source.base/src/org/netbeans/api/java/source/SourceUtils.java
+++ b/java/java.source.base/src/org/netbeans/api/java/source/SourceUtils.java
@@ -683,7 +683,7 @@ public class SourceUtils {
         }
 
         final boolean apply(final FileObject fo) {
-            if (name.equals(fo.getNameExt())) {
+            if (fo.getNameExt().equals(name)) {
                 return true;
             }
             final String foName = fo.getName();


java/java.source.base/src/org/netbeans/api/java/source/SourceUtils.java
```diff
@@ -683,7 +683,7 @@ private abstract static class Match {
         }
 
         final boolean apply(final FileObject fo) {
-            if (name.equals(fo.getNameExt())) {
+            if (fo.getNameExt().equals(name)) {
```
mbien, line 686: good that getNameExt() returns empty Strings. This makes it better than the JDK 20 proposal of the Path API which was luckily dropped last minute because of it returning null:
https://bugs.openjdk.org/browse/JDK-8297814

I already had a [code inspection](https://github.com/mbien/jackpot-inspections/commit/89c3f68840351b745d16f9cc4a0b4191e6afe1d6) for that ready expecting it to be a common bug.

new version is the PR 11545 which might make it into JDK 21 following the same null safe logic

good choice back then @jtulach!

mbien (approved):
The Match class initialized with a null name would never match anything. This may or may not be intended. Hopefully this doesn't cause silent issues now which don't throw exceptions. cc @jlahoda 

but this is probably still a good hotfix so +1 from me.

matthiasblaesing:
Seems to work and is minimally invasive, so lets get this in.

PR #5280, opened 2023-01-11
Improve Dialog Placement

errael:
When placing a dialog through NetBeans' API, if no focused component is found then use NbMainWindow. Change fallback for Utilities.findDialogParent to use NbMainWindow.

@matthiasblaesing , @neilcsmith-net , @mbien alerting since you made comments on previous PR.

This PR is a follow on to https://github.com/apache/netbeans/pull/4739. Wanted to open a bigger change earlier in the cycle, but... So this is a minimal change, hopefully more to come next cycle with detailed examination of cases from https://github.com/apache/netbeans/discussions/4887.

There are situations where
```
null == KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusedWindow();
```
And also related methods such as `getFocusOwner()`, `getActiveWindow()` are `null`.

What happens is `NbPresenter.findFocusedWindow()` returns null and the logic in `NbPresenter.initBounds()` ends up doing `Utilities.findCenterBounds(size)` which does `Utilities.getCurrentGraphicsConfiguration()` and ends up on the default screen instead of a screen that NetBeans is on. The bigger change I've been thinking about addresses some issues with `Utilities.getCurrentGraphicsConfiguration()` and would put the dialog on the same screen as NetBeans is on, but even better is to have the dialog *over the Window that NetBeans is in*. This PR does that with a hack to `NbPresenter.findFocusedWindow()`; taken from Utilities.

In `NbPresenter.initBounds()`, can consider `Utilities.findDialogParent()` instead of the local `findFocusedWindow()`, and then make sure that whatever's returned is a window. In any event, with this change the else clause is unlikely to ever be executed; just as well, notice the comment for the else clause:
```
//just center the dialog on the screen and let's hope it'll be
//the correct one in multi-monitor setup
```

### notes on Utilities.findDialogParent
There was some discussion in the dialog parent PR https://github.com/apache/netbeans/pull/4739#discussion_r997414384 from last release about the `Utilities.findDialogParent()` fallback. Investigation for this PR showed a situation where a fallback is needed, and the current fallback is not good.

Frame.getFrames() comes from a Vector<WeakReference<Window>>; doing garbage collection shortens the array. Here's an example I saw, the fallback
```
parent = f.length == 0 ? null : f[f.length - 1];
```
selects the last item and is choosing something ripe for garbage collection. Notice in this case the 2nd entry is the main window.

```
f	Frame[]	#16829(length=6)	#16829(length=6)	
[0]	Frame	#16831	java.awt.Frame[frame0,2640,390,480x300,invalid,hidden,
[1]	JFrame	#16832	javax.swing.JFrame[NbMainWindow,2566,544,1274x505,inva
[2]	Popup$DefaultFrame	#16833	javax.swing.Popup$DefaultFrame[frame3,
[3]	Popup$DefaultFrame	#16778	javax.swing.Popup$DefaultFrame[frame2,
[4]	Popup$DefaultFrame	#16834	javax.swing.Popup$DefaultFrame[frame4,
[5]	Popup$DefaultFrame	#16835	javax.swing.Popup$DefaultFrame[frame5,
```

These entries are like:
```
javax.swing.Popup$DefaultFrame[frame5,0,32,0x0,invalid,hidden
```

This PR makes the change to use the main window as fallback. This was previously proposed, but was deferred pending clarification on whether a fallback was needed at all.

### DIFF
--- a/platform/core.windows/src/org/netbeans/core/windows/services/DialogDisplayerImpl.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/services/DialogDisplayerImpl.java
@@ -225,7 +225,7 @@ public class DialogDisplayerImpl extends DialogDisplayer {
             // if a modal dialog is active use it as parent
             // otherwise use the main window
 
-            NbPresenter presenter = null;
+            NbPresenter presenter;
             if (descriptor instanceof DialogDescriptor) {
                 if (NbPresenter.currentModalDialog != null) {
                     if (NbPresenter.currentModalDialog.isLeaf ()) {
@@ -252,11 +252,8 @@ public class DialogDisplayerImpl extends DialogDisplayer {
                         presenter = new NbPresenter(descriptor, NbPresenter.currentModalDialog, true);
                     }
                 } else {
-                    Frame f = KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow() 
-                        instanceof Frame ? 
-                        (Frame) KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow() 
-                        : WindowManager.getDefault().getMainWindow();
-
+                    Window w = KeyboardFocusManager.getCurrentKeyboardFocusManager ().getActiveWindow ();
+                    Frame f = w instanceof Frame ? (Frame) w : WindowManager.getDefault().getMainWindow();
                     if (noParent) {
                         f = null;
                     }
--- a/platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
+++ b/platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
@@ -1591,6 +1591,18 @@ implements PropertyChangeListener, WindowListener, Mutex.Action<Void>, Comparato
      */
     private Window findFocusedWindow() {
         Window w = KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusedWindow();
+        if( w == null ) {
+            // PR#5280
+            LOG.fine( () -> "No focused window, find mainWindow" );
+            for( Frame f01 : Frame.getFrames() ) {
+                if( "NbMainWindow".equals(f01.getName())) { //NOI18N
+                    if(f01.getWidth() != 0 || f01.getHeight() != 0) {
+                        w = f01;
+                    }
+                    break;
+                }
+            }
+        }
         while( null != w && !w.isShowing() ) {
             w = w.getOwner();
         }
--- a/platform/openide.util.ui/src/org/openide/util/Utilities.java
+++ b/platform/openide.util.ui/src/org/openide/util/Utilities.java
@@ -1157,11 +1157,9 @@ public final class Utilities {
                 return w.getGraphicsConfiguration();
             } else {
                 //#217737 - try to find the main window which could be placed in secondary screen
-                for( Frame f : Frame.getFrames() ) {
-                    if( "NbMainWindow".equals(f.getName())) { //NOI18N
-                        return f.getGraphicsConfiguration();
-                    }
-                }
+                Frame f = findMainWindow();
+                if(f != null)
+                    return f.getGraphicsConfiguration();
             }
         }
 
@@ -1326,6 +1324,25 @@ public final class Utilities {
     }
 
     /**
+     * Find the main NetBeans window; must have width or height.
+     * This is used locally to avoid dependency issues. 
+     * @return NetBeans' main window
+     */
+    private static Frame findMainWindow()
+    {
+        Frame f = null;
+        for( Frame f01 : Frame.getFrames() ) {
+            if( "NbMainWindow".equals(f01.getName())) { //NOI18N
+                if(f01.getWidth() != 0 || f01.getHeight() != 0) {
+                    f = f01;
+                }
+                break;
+            }
+        }
+        return f;
+    }
+
+    /**
      * This is for use in situations where a standard swing API,
      * such as {@linkplain JOptionPane.show*} or {@linkplain JFileChooser.show*},
      * is used to display a dialog. {@code null} should never be used
@@ -1345,8 +1362,8 @@ public final class Utilities {
             parent = KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow();
         }
         if (parent == null) {
-            Frame[] f = Frame.getFrames();
-            parent = f.length == 0 ? null : f[f.length - 1];
+            // PR#5280
+            parent = findMainWindow();
         }
         return parent;
     }


errael:
Doing the same steps, the problem (stemming from no focused component or window) is intermittent; I'm guessing timing.

The problem may be more likely to arise when using the jVi plugin. When the user does a vi colon command, a one line window is brought up for the command entry. Some jVi commands invoke actions which bring up a dialog. The one line window is disappearing while the dialog is coming up.

mbien:
not sure if I will have time to take a deeper look at this before feature freeze. But I found a situation where the direct parent is not the optimal answer:

the editor tab dropdown can close tabs. If the tab is unsaved a dialog will appear. This dialog's parent would be the popup which causes it open on the upper right area of the main window.

This doesn't look horrible but its something to check. Simple yes/no dialogs should probably always open centered on the main window which might be the better default.

platform/core.windows/src/org/netbeans/core/windows/services/DialogDisplayerImpl.java
```diff
@@ -252,11 +252,8 @@ public void showDialog () {
                         presenter = new NbPresenter(descriptor, NbPresenter.currentModalDialog, true);
                     }
                 } else {
-                    Frame f = KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow() 
-                        instanceof Frame ? 
-                        (Frame) KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow() 
-                        : WindowManager.getDefault().getMainWindow();
-
+                    Window w = KeyboardFocusManager.getCurrentKeyboardFocusManager ().getActiveWindow ();
+                    Frame f = w instanceof Frame ? (Frame) w : WindowManager.getDefault().getMainWindow();
```
errael, line 256: This is a cosmetic change. It's both more clear and allows easier comparison to what's going on in the
```
if (descriptor instanceof DialogDescriptor) {
```
case

errael:
> to take a deeper look at this

This is actually two tiny changes to handle corner cases where the "normal" situation doesn't apply. Adding some review comments to to clarify.

platform/openide.util.ui/src/org/openide/util/Utilities.java
```diff
@@ -1345,8 +1362,8 @@ public static Component findDialogParent() {
             parent = KeyboardFocusManager.getCurrentKeyboardFocusManager().getActiveWindow();
         }
         if (parent == null) {
-            Frame[] f = Frame.getFrames();
-            parent = f.length == 0 ? null : f[f.length - 1];
+            // PR#5280
+            parent = findMainWindow();
```
errael, line 1366: This is a change suggested by @matthiasblaesing in https://github.com/apache/netbeans/pull/4739#discussion_r997414384. The change was deferred pending concrete examples. There is an example detailed in the opening comment for this PR.

platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
```diff
@@ -1591,6 +1591,18 @@ private void initBounds() {
      */
     private Window findFocusedWindow() {
         Window w = KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusedWindow();
+        if( w == null ) {
+            // PR#5280
+            LOG.fine( () -> "No focused window, find mainWindow" );
```
errael, line 1596: If there is no focused window, then try the mainWindow rather than centering on the JDK's default screen.

errael:
> a situation where the direct parent is not the optimal answer:
> 
> the editor tab dropdown can close tabs. If the tab is unsaved a dialog will appear. This dialog's parent would be the popup which causes it open on the upper right area of the main window.

Is this intermittent for you? I can't reproduce what you're seeing. I thought it might be because I use "focus follows mouse" but disabling that didn't change it. The `Question`-`File xxx is modified. Save?` dialog always shows up over the main window. I tried detaching the editor, the dialog shows up over the detached editor. I'm on current linux, with
```
XDG_CURRENT_DESKTOP=pop:GNOME
```
Could be a OS/window manager issue.

The close sequence starts with the following and notice the comment about focusing the top component before continuing.
https://github.com/apache/netbeans/blob/cac7e658adf03d514dad62b29fab488b0ed18a99/platform/core.windows/src/org/netbeans/core/windows/actions/CloseWindowAction.java#L74-L89

The action does end up in `NbPresenter.findFocusedWindow()`, where I've got a change; in the cases I've looked at the code I've added doesn't come into play.

errael:
> > a situation where the direct parent is not the optimal answer:
> > the editor tab dropdown can close tabs. If the tab is unsaved a dialog will appear. This dialog's parent would be the popup which causes it open on the upper right area of the main window.
> 
> Is this intermittent for you? I can't reproduce what you're seeing. 
> ...
> 
> Could be a OS/window manager issue.

@mbien, If you have a chance to check what is returned by
```
Window w = KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusedWindow();
```

at the first line of
```
org.netbeans.core.windows.services.findFocusedWindow()
```

And it's something that doesn't make sense, something like
```
javax.swing.Popup
```
Then I could try to address it.

matthiasblaesing:
Choosing the close icon in the editor tabs drop down gives this strange placement:

Bildschirmfoto vom 2023-01-16 21-55-22
Bildschirmfoto vom 2023-01-16 21-55-03

The windows are neither in the right position, nor the in the center of the right screen.

errael:
@matthiasblaesing Thanks for the info. Could you and @mbien provide OS/window manager info.

I've been unable to reproduce this, see the last couple of messages I've posted. In particular, https://github.com/apache/netbeans/pull/5280#issuecomment-1382314612 indicates some information I'd like to see (for example, collected with the debugger) that might provide some insight into what's going on.

This behavior has been there for a long while (assuming it's the same thing Michael brought up) and doesn't have anything to do with this PR.

matthiasblaesing (approved):
My observation in the previous comment is reproducible, but it is also not caused by this change. The code in `org.netbeans.core.windows.services.DialogDisplayerImpl` is just not optimal, leading to the strange placement of the dialog. A quick test inserting:

```java
presenter.setLocationRelativeTo(presenter.getOwner());
```

before the `setVisible(true)` call, gives me sane centering over the main window. Not sure if that is desirable, but might be an option in the future to consider.

I checked what I got for the `Window w` / `Frame f` in line 255/256 and in both cases (editor pane in main window and editor pane in a separater window) I got the correect `Window`/`JFrame` objects I would have expected, not some popup.

errael:
> My observation in the previous comment is reproducible

@matthiasblaesing Wish I could reproduce it. Here's what the code does

Note that the placement code in NbPresenter _does not use the owner_,
_it uses the focused window_ (if found).

DialogDisplayerImpl.AWTQuery.showDialog
- presenter = new NbPresenter(desc, owner, modalTrue)
  - initBounds()
    - // assuming focused window
    - `setLocationRelativeTo( findFocusedWindow() )`
    - **XXX** code to change location if screen boundary overlaps dialog
- customizeDlg(presenter) // <<< does lookup, not found in my tests
- // matthias: `presenter.setLocationRelativeTo(presenter.getOwner())`

**XXX above could do what's being seen**

Here's what I tried, but can't reproduce

close

Under the debugger
- start up clean udir/cache dummy nbm project (no plugins)
- open existing ant java library project used for testing, several tiny files
- open a few java files
- Add a character to a file
- Context menu on editor tab, select close

Observe `save? dialog` centered on main window

platform/openide.util.ui/src/org/openide/util/Utilities.java
neilcsmith-net: @errael I would possibly fall back to the original behaviour if the search for the main window returns null.  Ideally it wouldn't use the name to find the right window, but that might be the easiest way to achieve.  I have no idea how many people (if any!) use `FileChoooserBuilder` outside of the standard window system, but this will change the behaviour of those cases.

errael: > @errael I would possibly fall back to the original behaviour if the search for the main window returns null. 

The original behavior is severely flawed. Take a look in the comment that opened this PR, under **notes on Utilities.findDialogParent**. And from the javadoc for Frame.getFrames()
```
Warning: this method may return system created frames, such as a shared, hidden frame which 
is used by Swing. Applications should not assume the existence of these frames, nor should an 
application assume anything about these frames such as component positions, LayoutManagers 
or serialization. 
```
> Ideally it wouldn't use the name to find the right window, but that might be the easiest way to achieve. 

Notice that this technique is already in use in this file. May have to do with dependencies. This PR encapsulates its use in a method so at least it's not buried.

> I have no idea how many people (if any!) use `FileChoooserBuilder` outside of the standard window system, but this will change the behaviour of those cases.

If someone explicitly wants random dialog placement there are probably more interesting ways. I believe this case rarely comes up AFAIK, I've just been lucky. 

A null owner is probably more consistent (better?).

If this is a strong concern for `FileChooserBuilder`, then suggest using a private version of `findDialogParent()` with the old code  rather than propagate buggy behavior in a new-ish public API.

(Aside: I've wondered about use of non standard window system; are there any known examples in "production"? )

neilcsmith-net: @errael yes, I read the doc. Let's stick with what you have and see if any complaints come out of the woodwork!

platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
```diff
@@ -1591,6 +1591,18 @@ private void initBounds() {
...
+            for( Frame f01 : Frame.getFrames() ) {
```
sdedic, line 1597: Nitpick: call `findMainWindow`  ?

errael: > call `findMainWindow` ?

@sdedic I'd like to do that; can you give me some guidance? Need to find a non-API file in `platform` that's OK to access from both `core.windows` and `openide.util.ui` where `findMainWindow` can be declared as a public method. Such a file must exist, any pointers?

errael:
@mbien @matthiasblaesing 

In https://github.com/apache/netbeans/pull/5280#issuecomment-1386241102 I said that the code in `NbPresenter.initBounds()` _code to change location if screen boundary overlaps dialog_, could be responsible for the behavior.  But it looks like there are other candidates for the problem.

I'm taking a look at `Window.setLocationRelativeTo()`, one comment in the javadoc is
```
Note: If the lower edge of the window is out of the screen, then the window is placed to the 
side of the Component that is closest to the center of the screen. So if the component is on 
the right part of the screen, the window is placed to its left, and vice versa.
```
which seems to echo the noted behavior.

In any event, knowing the arg in `setLocationRelativeTo( findFocusedWindow() )` in NbPresenter.initBounds() is important for getting to the bottom of what's going on. Also seeing `getLocation()` for the various components and the precise code path and intermediate values would be useful.

platform/openide.util.ui/src/org/openide/util/Utilities.java
```diff
@@ -1325,6 +1323,25 @@ private static Rectangle findCenterBounds(GraphicsConfiguration gconf, Dimension
...
+    /**
+     * Find the main NetBeans window; must have width or height.
+     * This is used locally to avoid dependency issues. 
+     * @return NetBeans' main window
+     */
+    private static Frame findMainWindow()
+    {
+        Frame f = null;
+        for( Frame f01 : Frame.getFrames() ) {
+            if( "NbMainWindow".equals(f01.getName())) { //NOI18N
```
eirikbakke, line 1335: Why doesn't WindowManager.getDefault().getMainWindow() work for this purpose? Is that in a module that is not available from here?

errael: > Why doesn't WindowManager.getDefault().getMainWindow() work for this purpose?

Since I didn't write the original code, I can only guess. BTW, the bug report that prompted the change is at https://bz.apache.org/netbeans/show_bug.cgi?id=217737 and the change was made by the window system guru at the time, AFAIK. I looked through emi's old repo, but didn't find the original commit.

I thought it was at least partly a module dependency issue, but I could be wrong. Side effects is another consideration. `Utilities.findMainWindow` looks for an existing window, may return null. `WindowManager...getMainWindow()` creates the window if not found.

errael: @eirikbakke Found it (was looking at the wrong Utilities)
https://github.com/emilianbold/netbeans-releases/commit/62e3746e8c1d0e7d3c033291bcc64c8dc8b075d8

But I saw no information about the change.

neilcsmith-net: Yes, `openide.windows` has a dependency on `openide.util.ui`. And theoretically `openide.util.ui` could be used in applications without a window system.

errael: IIUC, to summarize things

- `openide.util.ui` can not use `WindowManager.getDefault().getMainWindow()` due to dependencies
- The limited functionality based on finding an existing MainWindow can be accessed from any application using NetBeans modules, if the application names the main window "NbMainWindow"
- Don't want findMainWindow() public API
- Could make it internal API (but I don't know where to put it, or exactly the considerations)
- If, in the future, there is a problem with requiring "NbMainWindow" could handle a `NetBeansPlatformMainWindowName` service (or somesuch e.g. SystemProperty) (I'm tempted to put an emoji here, grimace perhaps?)

platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
sdedic: Ah ;) my bad, I overlooked the module boundary. No suggestion that 'feels' right then. An option would be to add a API/SPI method on e.g. DialogDisplayer or WindowManager.

BTW this method uses `KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusedWindow()` while `getCurrentGraphicsConfiguration()`  (in diff below) uses `KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusOwner()` then `getWindowAncestor()`.  Deliberate difference or inconsistency ?

platform/openide.util.ui/src/org/openide/util/Utilities.java
eirikbakke: OK, thanks for the detailed explanation!

eirikbakke (approved):
PR looks fine to me, though I don't have deep knowledge of what's going on here. But it sounds like it's been extensively tested and discussed. Thanks for this PR!

platform/openide.util.ui/src/org/openide/util/Utilities.java
errael: @eirikbakke hope you enjoyed it ;-) I'm not overly familiar with all the issues involved, especially the innards of focus system (not to mention NBs module system), the details are more for me, and to give those who ask "why?" things to consider. But most of all, for someone to point out flaws in my assumptions/reasoning.

platform/core.windows/src/org/netbeans/core/windows/services/NbPresenter.java
errael: I've wondered about that myself; are the two equivalent?

I'm copying your comment to #4887 which are notes for a follow on PR addressing some stuff in `Utilities`. The opening comment of this PR mentions issues for the next PR.

errael:
There are two approvals, and someone wearing a release manager's hat said "Let's stick with what you have and see if any complaints come out of the woodwork!". I'll count that as 2+ approvals, no negatives.

@mbien @sdedic (or @neilcsmith-net wearing a different hat) any further comments?

Can this get merged?

BTW, I think this PR will take care of several items I listed in #4887, I'll go through them all and see if they meet the criteria covered by this PR. The big revelation for me is that the layout decisions happen mostly in NbPresenter (with an occasional assist from openide.util.Utilities). I'll patch NB-17 release with this PR and keep my eyes open for follow on issues (I haven't been using NB lately).

errael:
@matthiasblaesing @mbien I'm willing to consider the problem (not related to this PR) both of you can reproduce (and I can't).

I think info gathered from `NbPresenter` as mentioned in https://github.com/apache/netbeans/pull/5280#issuecomment-1382314612 might be important in understanding what's going on.

There was some thought that it might be an issue with `DialogDisplayer` as mentioned in https://github.com/apache/netbeans/pull/5280#pullrequestreview-1252234566 but my guess is not.

And https://github.com/apache/netbeans/pull/5280#issuecomment-1397371539 might help in understanding what's going on.

If you want/can open up a bug with some info (especially reproducability) I'd like to know about it.

matthiasblaesing:
Lets get this in.

Issue #5250, opened 2023-01-09
Update java.mx.project to work with GraalVM 22.3.0 release


MartinBalin:
### Apache NetBeans version

Apache NetBeans 16

### What happened

java.mx.project has a dependency on MX project and GraalVM itself. Currently it depends on 20.3.0 release which is 2 years old. It will be good to update dependency on MX and Graal to version 22.3.0 from last Oct. Also new JDKs like 19 are supported over that time and JDK8 is being dropped of.

### How to reproduce

https://github.com/apache/netbeans/issues/5196 in the meanwhile tags were moved to commit sha but it is still 20.3

### Did this work correctly in an earlier version?

Apache NetBeans 15

### Operating System

OS independent

### JDK

JDK11

### Apache NetBeans packaging

Apache VSNetBeans for VSCode

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

PR #5285, opened 2023-01-12
Issue #5250: Update java.mx.project to work with GraalVM 22.3.0 release.

dbalek:
GraalVM used by `java.mx.project` updated to 22.3.0 release.

### DIFF
--- a/java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteActionProviderTest.java
+++ b/java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteActionProviderTest.java
@@ -91,7 +91,7 @@ public class SuiteActionProviderTest extends SuiteCheck {
         ap.invokeAction(ActionProvider.COMMAND_BUILD, ctx);
 
         assertTrue("Progress started", progress.started);
-        progress.finished.await(45, TimeUnit.SECONDS);
+        progress.finished.await(600, TimeUnit.SECONDS);
         assertNotNull("Progress finished", progress.success);
     }
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -692,7 +692,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        java: [ '8' ]
+        java: [ '11' ]
     steps:
 
       - name: Set up JDK ${{ matrix.java }}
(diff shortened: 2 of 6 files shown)


mbien:
wouldn't it be better to revert #5199 again to check out tags like before?

@jtulach mentions [here](https://github.com/apache/netbeans/issues/5196#issuecomment-1378342202) that tags will stay in the graal repo now.

MartinBalin (approved):
Nice! thank you

jtulach (approved):
Looks like a smooth upgrade.

java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteActionProviderTest.java
```diff
@@ -91,7 +91,7 @@ public void finished(boolean success) {
         ap.invokeAction(ActionProvider.COMMAND_BUILD, ctx);
 
         assertTrue("Progress started", progress.started);
-        progress.finished.await(45, TimeUnit.SECONDS);
+        progress.finished.await(600, TimeUnit.SECONDS);
```
mbien, line 94: i tried to reproduce the issues with this test locally a few times already. But it always passes. This test is also super fast if i remember correctly. It finishes in 700ms or so.

So I don't think this could be timeout related - but who knows.

mbien: also as note: the tests are wrapped in a retry script which retries twice on failure. So long timeout could lead to really long test times on failure.

dbalek: The last time it failed with the AssertionError thrown from [SuiteActionProviderTest.java#L95](https://github.com/apache/netbeans/blob/af5a1da6f7b5c826b301d0ac422bcaec5030ea0a/java/java.mx.project/test/unit/src/org/netbeans/modules/java/mx/project/SuiteActionProviderTest.java#L95)

mbien: yeah. that one fails very often. Yesterday I looked at the last 18 merges to master and wrote down all failures to wrap tests in retry scripts. This test was on top.

mbien (approved):

PR #5286, opened 2023-01-12
Open projects that correspond to workspace folders. Prevents "do you want to open project ?" questions.

sdedic:
If the LSP client added a folder to a workspace, then opened a file, NBLS would ask a question "File XXX is part of project Foo, do you want to open ... ?".  Since the containing folder **was** explicitly adeded the lsp client's user, this question is ... silly. 

It was caused by NBLS did not receive/process `didChangeWorkspaceFolders` notification from the LSP client. On the startup, workspace folders were 'registered' as known and NBLS does not ask "project open question" for files beneath those folders - but when the folders change during NBLS run, the added folders were not 'registered' in the same way.

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
@@ -50,7 +50,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


MartinBalin (approved):

MartinBalin:
Please add also to VSNetBeans_1603 branch

java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/WorkspaceServiceImpl.java
```diff
@@ -1170,6 +1173,54 @@ void updateJavaImportPreferences(FileObject fo, JsonObject configuration) {
...
+        if (!refreshProjectFolders.isEmpty()) {
+            server.asyncOpenSelectedProjects(refreshProjectFolders, true).thenAccept((projects) -> {
+                // report initialization of a project / projects
+                String msg;
+                if (projects.length == 0) {
+                    // this should happen immediately
+                    return;
+                } 
+                ProjectInformation pi = ProjectUtils.getInformation(projects[0]);
+                String n = pi.getDisplayName();
```
lbownik, line 1207: lines 1207 - 1213 look like they should belong to ProjectInformation class

sdedic: > lines 1207 - 1213 look like they should belong to ProjectInformation class

Rather to ProjectUtils; but yes, some "give-me-suitable-project-name-with-all-fallbacks' could be useful, I did this chain of null checks several times already. I'll address in a separate PR.

dbalek (approved):
Looks fine.

PR #5287, opened 2023-01-12
Added ANTLR based Go Parser module

lkishalmi:
This is a generated Lexer/Parser Library for Go.

I'm going to add another module `languages.go` soon. Since the discussion on the devops cluster has died off, I'm going to add these into the ide cluster, till we find something better.


lkishalmi:
I had to make this module show in auto updates, in order to satisfy the commit validation. I'm going to revert that in the upcoming languages.go PR.

mbien (commented):
although this is using a third party grammar it might be still worth to add a basic test.

lkishalmi:
@mbien Well, it is entirely third-party, just downloaded and generated code. I do not exactly know how, but it seems to be tested on these files: https://github.com/antlr/grammars-v4/tree/master/golang/examples

matthiasblaesing (approved):
Looks sane to me. I think it would have been better to get this in together with filetype wiring, but still ok.

PR #5290, opened 2023-01-13
allow concurrent local indexing while downloading remote indexes.

mbien:
extracted from #4999

This aims to improve the indexing experience by allowing remote index downloads while local `.m2` indices are processed.

concurrent_dl_while_indexing

CI should produce a dev build for this one for easier testing.

### DIFF
--- a/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
+++ b/java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
@@ -22,7 +22,6 @@
 
 package org.netbeans.modules.maven.indexer;
 
-import org.netbeans.modules.maven.indexer.api.RepositoryQueries;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


mbien:
all tests green. removing all-tests label for subsequent syncs.

java/maven.indexer/src/org/netbeans/modules/maven/indexer/NexusRepositoryIndexerImpl.java
```diff
@@ -152,7 +153,19 @@ public class NexusRepositoryIndexerImpl implements RepositoryIndexerImplementati
...
+     */
+    private static final RequestProcessor RP = new RequestProcessor("maven-indexing", 2);
+
+    /**
+     * Best effort attempt to prevent two _local_ repos to be indexed at the same time to avoid unnecessary IO contention.
+     * Concurrent local+remote indexing is allowed.
+     * This isn't perfect since a download can finish before local repos finished indexing, tests showed
+     * performance doesn't degrade significantly on solid state drives.
+     */
+    private static final ReentrantLock localIndexingPermit = new ReentrantLock();
```
matthiasblaesing, line 168: I think I see a problem here when two local repostories are being indexed and a new remote repository is requested. This will be stalled until the first local repository is done indexing. Basicly like this:

- Local Repository R1 is submitted into `RP` to be indexed and locks `localIndexingPermit`
- Local Repository R2 is submitted into `RP` to be indexed and tries to lock `localIndexingPermit` (this thread will block at Line 658)
- Remote Repository R3 is submitted into `RP` to be indexed - it does not execute, as R1 and R2 are "running" in the processor and R3 is queued.

Alternative idea: use two requestprocessors with a through put of 1. One for remote repos, one for local repos. If I understand your intention correctly, this should accomplish the same.

mbien: good idea!

mbien: I initially wanted to add the lock to the main `indexLoadedRepo` method, which would be the most correct implementation. However the method is already large enough and it deals with mutexes already. Probably worth splitting up one day.

So I opted to keep it as simple as possible - but your suggestion is even simpler. We can still switch to locks later if this doesn't work well enough.

matthiasblaesing (approved):
Looks sane to me.

PR #5292, opened 2023-01-13
Made ANTLR Token LexerInputCharStream Friendly

lkishalmi:
This fix makes ANTLR Token `getText()` and `toString()` methods usable when the lexer is backed by `LexerInputCharStream`. Unfortunately it seems that ANTLR does not really honor their contract when saying some methods can throw `UnsupportedOperationException`. I may file a PR for ANTLR as well, till then, this one makes debugging easier.

I also dared to change the `AbstractAntlrLexerBridge.mapToken(int antlrTokenType)` -> `AbstractAntlrLexerBridge.mapToken(org.antlr.v4.runtime.Token antlrToken)` till it's in this phase. Mappers could take advantage on the full token info as  the `getText()` method is working.

### DIFF
--- a/ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLexer.java
+++ b/ide/languages.toml/src/org/netbeans/modules/languages/toml/TomlLexer.java
@@ -37,8 +37,8 @@ public final class TomlLexer extends AbstractAntlrLexerBridge<org.tomlj.internal
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


matthiasblaesing (approved):
The change looks sane to me. The implementation of `getText` looks a bit strange, but I see, that it matches the implementation of `CommonToken`.

Issue #5284, opened 2023-01-12
PHP: Generates the return type boolean instead of bool for a function in phpDoc


troizet:
### Apache NetBeans version

Apache NetBeans 16

### What happened

Generates the return type `boolean` instead of `bool` for a function in phpDoc.
Php does not have a built-in `boolean` type: https://www.php.net/manual/en/language.types.intro.php
Also see the warning under "Scalar Types": https://www.php.net/manual/en/language.types.declarations.php

### How to reproduce

- Create a function that returns `true` or `false`:
```php
function test() 
{
    return true;
}
```
- Generate a phpDoc comment
- The `@return` tag was received with the wrong return type:
```php
/**
 * 
 * @return boolean
 */
function test() 
{
    return true;
}
```

instead of:
```php
/**
 * 
 * @return bool
 */
function test() 
{
    return true;
}
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Debian 10

### JDK

OpenJDK 64-Bit Server VM 11.0.16+8-post-Debian-1deb10u1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

KacerCZ:
Reproducible.
Thanks for your report.

Issue #5283, opened 2023-01-12
PHP: Generates the return type boolean instead of bool for a method when overriding a base class method with an autocomplete


troizet:
### Apache NetBeans version

Apache NetBeans 16

### What happened

Generates the return type `boolean` instead of `bool` for a method when overriding a base class method with an autocomplete.
Php does not have a built-in `boolean` type: https://www.php.net/manual/en/language.types.intro.php
Also see the warning under "Scalar Types": https://www.php.net/manual/en/language.types.declarations.php

### How to reproduce

- Create a base class with a test method that returns `true` or `false`:

```php
class Base
{
    
    public function test()
    {
        return true;
    }
    
}
```

- Create a subclass:

```php
class SubClass extends Base
{

}
```

- Press `Ctrl + Space` to autocomplete the subclass and select `test() - override`
- A method with the wrong return type was received:
``` php
    public function test(): boolean {
        return parent::test();
    }
```
instead of:
``` php
    public function test(): bool {
        return parent::test();
    }
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Debian 10

### JDK

OpenJDK 64-Bit Server VM 11.0.16+8-post-Debian-1deb10u1

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

KacerCZ:
Reproducible.

If method `Base::test()` has specified native return type `bool`, then generated method has also return type `bool`.
So it may have same cause as bug #5284 when NetBeans tries to guess return type.

Thanks for your report.

PR #5294, opened 2023-01-14
PHP: removed built-in type boolean

troizet:
What was done in this PR:
 - deleted the built-in type `boolean`
 ( php does not have a built-in `boolean` type: https://www.php.net/manual/en/language.types.intro.php
Also see the warning under "Scalar Types": https://www.php.net/manual/en/language.types.declarations.php)
 - fixed broken tests
 - fixed issues: 
   - #5283, 
   - #5284
 - new tests for fixed issues have been added

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/actions/UsedNamesCollector.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/actions/UsedNamesCollector.java
@@ -186,7 +186,7 @@ public class UsedNamesCollector {
         @Override
         public void visit(PHPDocTypeNode node) {
             UsedNamespaceName usedName = new UsedNamespaceName(node);
-            if (isValidTypeName(usedName.getName())) {
+            if (isValidTypeName(usedName.getName()) && isValidAliasTypeName(usedName.getName())) {
                 processUsedName(usedName);
             }
         }
@@ -194,6 +194,10 @@ public class UsedNamesCollector {
         private boolean isValidTypeName(final String typeName) {
             return !SPECIAL_NAMES.contains(typeName) && !Type.isPrimitive(typeName);
         }
+        
+        private boolean isValidAliasTypeName(final String typeName) {
+            return !SPECIAL_NAMES.contains(typeName) && !Type.isPrimitiveAlias(typeName);
+        }        
 
         private void processUsedName(final UsedNamespaceName usedName) {
             List<UsedNamespaceName> usedNames = existingNames.get(usedName.getName());
--- a/php/php.editor/src/org/netbeans/modules/php/editor/model/impl/Type.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/model/impl/Type.java
@@ -108,7 +108,7 @@ public final class Type {
 
     public static boolean isPrimitive(String typeName) {
         boolean retval = false;
-        if (BOOL.equals(typeName) || BOOLEAN.equals(typeName) || INT.equals(typeName)
+        if (BOOL.equals(typeName) || INT.equals(typeName)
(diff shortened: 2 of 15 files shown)


mbien:
added php label for full test suite. restarting CI.

KacerCZ:
Proposed change looks good to me.
Maybe then make another PR to remove other type aliases - integer, real, double.

Let's wait for @junichi11's opinion.

mbien:
just as reminder: feature freeze for NB 17 is on the 18th, there still is time but not a lot if this is intended for NB 17.

matthiasblaesing:
Just as a heads up. [phpDocumentator](https://docs.phpdoc.org/3.0/guide/guides/types.html) explicitly lists:

> - `int` and `integer`
> - `bool` and `boolean`

but also writes:

> The PHPDoc Standard, and thus phpDocumentor, can refer to all primitive types in PHP

This is contradictory and who is right? Can `PHPDoc` cover more than the "real" primitives or can't it?

KacerCZ:
PHPDoc types accept also type aliases but these aliases can't be used as native type.
See https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.base.scalar

This PR fixes case when NetBeans generate type "boolean" and use it as native return type for PHP 7.0 or newer.

troizet:
I seem to have rushed the PR. Sorry for wasting your time.
It would probably be better if we left the use of `boolean` type in phpDoc as it is, unchanged. Since `boolean` in phpDoc is used in most projects.
And remove the built-in `boolean` type only for type declarations and type guessing.
How do I convert a PR to a draft?

junichi11:
> It would probably be better if we left the use of `boolean` type in phpDoc as it is, unchanged. Since `boolean` in phpDoc is used in most projects.

Agree. It would be better.
As @matthiasblaesing wrote, I guess it is added for phpDocumenter.

tmysik:
If I get it right, the ideal situation here would be this one:

- in the PHP code, use `bool` everywhere; but
- keep support for `boolean`, but only in the PHPDoc comments (kind-of like a legacy code, we would not generate it, but still were able to read/understand it).

Does it make sense? Would it be possible to implement it?

troizet:
Updated the PR so that it does not affect the phpDoc

tmysik (approved):
The change looks good to me, but please, let @junichi11 approve it as well. Thank you.

junichi11:
@troizet Could you improve the commit message like the title and description of this PR? (We should add related issue numbers.)
If you would like to fix this into NB17, please rebase on `delivery` branch. (cc: @mbien)

neilcsmith-net:
@junichi11 I'm taking that as _you_ making the call that it belongs in delivery! :smile:  Adding milestone and do-not-merge pending rebasing of the PR so we can track.  17-rc3 is the last guaranteed release candidate unless we have urgent fixes, so ideally needs to be in a position I can merge within the next 24hrs.

troizet:
I changed the commit message and rebased it on delivery. @junichi11, @neilcsmith-net please check if I did it right?

junichi11 (approved):
Thank you for your work!

junichi11:
@neilcsmith-net Thank you for your help :)

@tmysik The release team (@neilcsmith-net ) merges this.

BTW, CI doesn't work, right?

neilcsmith-net:
Looks good, merging.

> BTW, CI doesn't work, right?

@junichi11 can be manually triggered by locking and unlocking the conversation as above if needed.  It's a "hidden" feature!

troizet:
Thanks!

junichi11:
> can be manually triggered by locking and unlocking the conversation as above if needed. It's a "hidden" feature!

@neilcsmith-net I see. Thank you!

PR #5298, opened 2023-01-16
Update FlatLaf from 2.6 to 3.0 and add macOS light and dark themes

DevCharly:
Update FlatLaf to v3.0.

Changes: https://github.com/JFormDesigner/FlatLaf/releases/tag/3.0

### macOS themes

FlatLaf 3 comes with two new themes, which look similar to macOS:

macOS themes

I've added both to the list of look and feels (on all platforms, screenshots are from Windows 11):

grafik

grafik

### Rounded outlined icons

FlatLaf 3 has new modern rounded outlined icons for `JFileChooser`, `JOptionPane`, `JPasswordField` and `JTree`:

image

They are e.g. used in project/file trees and in file choosers (in all FlatLaf themes):

grafik

### DIFF
--- a/platform/libs.flatlaf/external/binaries-list
+++ b/platform/libs.flatlaf/external/binaries-list
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-4DD2BA228E3C57EB3D80E3927B5A6A33265EB69B com.formdev:flatlaf:2.6
+B5410F3F9137FEBC7D916CA4E0A7E9F6DDEB5B9A com.formdev:flatlaf:3.0
--- a/platform/libs.flatlaf/external/flatlaf-3.0-license.txt
+++ b/platform/libs.flatlaf/external/flatlaf-3.0-license.txt
@@ -1,7 +1,7 @@
 Name: FlatLaf Look and Feel
 Description: FlatLaf Look and Feel
-Version: 2.6
-Files: flatlaf-2.6.jar
+Version: 3.0
+Files: flatlaf-3.0.jar
 License: Apache-2.0
 Origin: FormDev Software GmbH.
 URL: https://www.formdev.com/flatlaf/
--- a/platform/libs.flatlaf/manifest.mf
+++ b/platform/libs.flatlaf/manifest.mf
@@ -4,4 +4,4 @@ OpenIDE-Module: org.netbeans.libs.flatlaf/1
 OpenIDE-Module-Install: org/netbeans/libs/flatlaf/Installer.class
 OpenIDE-Module-Specification-Version: 1.13
 AutoUpdate-Show-In-Client: false
-OpenIDE-Module-Implementation-Version: 2.6
+OpenIDE-Module-Implementation-Version: 3.0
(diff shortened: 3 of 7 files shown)


DevCharly:
> I like the new icons.

Thx 😄 

> The dark mac theme might be interesting for those who prefer higher contrast themes.

Yes, definitely. That's why I've not limited those themes to mac platform 😉

mbien:
@DevCharly its off topic but since I see it in the screenshots. I feel like I am missing out on those nice shortcut buttons on the left side of the file chooser dialog as linux user :)

neilcsmith-net:
Looks great!  I was somewhat tempted to squeeze it in just before I branched, although hadn't had a chance to look properly so didn't want to make the call. :smile:

One concern would be whether we should be using the name `macOS` in the UI here?  Maybe something like `FlatLaf Cupertino Light`?

Are the mentioned fonts included in this change for NetBeans?

mbien:
> One concern would be whether we should be using the name macOS in the UI here? Maybe something like FlatLaf Cupertino Light?

agreed. Mac OS is also changing the theme periodically, so it would be better the not associate a Look and Feel with a OS theme to have more flexibility if it needs adjustments. I like Cupertino since it still alludes to the origin.

mbien:
@neilcsmith-net yes mine too. I already looked up bug species which like apples but they all have fairly complicated names :)

I only wanted to bring up a technical reason why naming it after an OS might not be a good idea. If it would start diverging from the mac look it might be perceived as a bug.

DevCharly:
> @DevCharly its off topic but since I see it in the screenshots. I feel like I am missing out on those nice shortcut buttons on the left side of the file chooser dialog as linux user :)

Yes, unfortunately the shortcuts are only available on Windows because the used Java API provides the information only on Windows 😞

DevCharly:
>One concern would be whether we should be using the name macOS in the UI here? Maybe something like FlatLaf Cupertino Light?

Good idea 👍 
I'm fine with Cupertino 

> Are the mentioned fonts included in this change for NetBeans?

No. They are in extra JARs.

eirikbakke (approved):
Looks great--applause!

(Also second the "Cupertino" name, for the trademark reasons and since it's a useful theme even on Windows.)

eirikbakke:
I tested this patch in its current form on my private NetBeans build; works as expected with no problems.

mbien:
if the `Cupertino` renaming makes it in time before RC2 we should rebase this PR on `delivery` and get this in for NB 17 in my opinion.

This is technically a bugfix since it removes 8 superfluous pixels on the file open dialog I _totally_ noticed while working with NB 16 :)

DevCharly:
I've renamed the new themes to Cupertino.

Please let me know if there is a need to rebase this PR to `delivery`.

But since the branch for this PR was created before the `delivery` branch was created, there is actually no need to rebase it.
It can be merged directly into the `delivery` branch. Rebasing makes no difference...

neilcsmith-net (approved):

Chris2011:
Great icons, but is it possible, that I can change it via the flatlaf properties file? I don't like this mixed stuff what we have. Most of the NetBeans Icons are not outlined and now it looks a bit odd with those all not outlined Icons in the Project window with the just folder outlined ones. As long as we can't switch all icons to either outline or not, I would like to change them by my own. Is this easy possible?

DevCharly:
@Chris2011 yes, this really looks odd.

Have created PR #5760 to replace FlatLaf outlined folder icons with NetBeans style icons.

Issue #4123, opened 2022-05-16
String Variables longer 100000 Characters are shown as <Unreadable> in Debug


hrotkogabor:
### Apache NetBeans version

Apache NetBeans 13

### What happened

String Variables longer 100000 Characters are shown in Debugger as <Unreadable>

instead of its content.

Tested with Netbeans 12.0 and 12.6 with Amazon Corretto open-jdk11.0.12_7 64 Bit and

Amazon Corretto open-jdk15.0.2_7 64 Bit and Oracle JDK 11.0.2

### How to reproduce

Debug a code, and give a String variable a big content. You cannot view it's content.

### Did this work correctly in an earlier version?

Apache NetBeans 12.3 or earlier

### Operating System

Linux

### JDK

11.0.15

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

Imported from https://issues.apache.org/jira/browse/NETBEANS-6382

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

lkishalmi:
Does that work with JDB? You might be stretching the limits of Java Debug protocol with that.

hrotkogabor:
At least a save to file button would help.

SirIntellegence:
I can confirm that this still happens with Netbeans 16

SirIntellegence:
#5299 Fixes this

SirIntellegence:
@matthiasblaesing This one is fixed and hasn't been closed yet.

matthiasblaesing:
@SirIntellegence thank you for the heads-up.

PR #5299, opened 2023-01-16
[NETBEANS-4123] Initial implementation of handling large strings

SirIntellegence:
in Java 9+ where the String class uses a byte[] instead of a char[]

Not a final implementation at the moment, but should be enough for now. I didn't see any tests related to this class, so I didn't write any. I have tested it and it works properly. May need more testing.

Handling UTF16 byte[] strings would require more effort since it becomes a nightmare.

### DIFF
--- a/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
+++ b/java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
@@ -21,22 +21,34 @@ package org.netbeans.modules.debugger.jpda.models;
 
 import com.sun.jdi.ArrayReference;
 import com.sun.jdi.ArrayType;
+import com.sun.jdi.ByteValue;
 import com.sun.jdi.CharValue;
 import com.sun.jdi.ClassNotLoadedException;
+import com.sun.jdi.ClassType;
 import com.sun.jdi.Field;
+import com.sun.jdi.IncompatibleThreadStateException;
+import com.sun.jdi.InvalidTypeException;
+import com.sun.jdi.InvocationException;
+import com.sun.jdi.Method;
+import com.sun.jdi.PrimitiveValue;
 import com.sun.jdi.ReferenceType;
 import com.sun.jdi.StringReference;
+import com.sun.jdi.ThreadReference;
 import com.sun.jdi.Type;
 import com.sun.jdi.Value;
+import com.sun.jdi.VirtualMachine;
 import java.io.IOException;
 import java.io.Reader;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.WeakHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.netbeans.api.debugger.DebuggerManager;
 import org.netbeans.api.debugger.DebuggerManagerAdapter;
 import org.netbeans.api.debugger.Session;
@@ -48,6 +60,7 @@ import org.netbeans.modules.debugger.jpda.jdi.ObjectReferenceWrapper;
 import org.netbeans.modules.debugger.jpda.jdi.ReferenceTypeWrapper;
 import org.netbeans.modules.debugger.jpda.jdi.StringReferenceWrapper;
 import org.netbeans.modules.debugger.jpda.jdi.VMDisconnectedExceptionWrapper;
+import org.openide.util.Exceptions;
 
 /**
  * A collector of shorted String values, that were too long.
@@ -61,6 +74,8 @@ public final class ShortenedStrings {
     private static final Map<String, StringInfo> infoStrings = new WeakHashMap<String, StringInfo>();
     private static final Map<StringReference, StringValueInfo> stringsCache = new WeakHashMap<StringReference, StringValueInfo>();
     private static final Set<StringReference> retrievingStrings = new HashSet<StringReference>();
+    private static final Map<VirtualMachine, Boolean> isLittleEndianCache =
+            new WeakHashMap<>();
 
     static {
         DebuggerManager.getDebuggerManager().addDebuggerListener(DebuggerManager.PROP_SESSIONS,
@@ -78,6 +93,9 @@ public final class ShortenedStrings {
                         stringsCache.clear();
                         retrievingStrings.clear();
                     }
+                    synchronized (isLittleEndianCache) {
+                        isLittleEndianCache.clear();
+                    }
                 }
             }
 
@@ -92,6 +110,61 @@ public final class ShortenedStrings {
         }
     }
 
+    private static boolean isLittleEndian(VirtualMachine virtualMachine) throws
+            InvalidTypeException, IncompatibleThreadStateException,
+            ClassNotLoadedException, InvocationException,
+            InternalExceptionWrapper, VMDisconnectedExceptionWrapper,
+            ObjectCollectedExceptionWrapper, ClassNotPreparedExceptionWrapper {
+        synchronized(isLittleEndianCache){
+            Boolean cached = isLittleEndianCache.get(virtualMachine);
+            if (cached != null){
+                return cached;
+            }
+            List<ReferenceType> possibleClasses = virtualMachine.classesByName(
+                    "java.lang.StringUTF16");
+            //If we don't know, we are going to assume little endian encoding.
+            //This should work for most architectures (x86, arm, riscv), but
+            //will result in bogus data on big endian architectures
+            final boolean defaultValue = true;
+            if (possibleClasses.isEmpty()){
+                ClassType ct = (ClassType) virtualMachine.classesByName(
+                        "java.lang.Class").iterator().next();
+                Method m = ct.concreteMethodByName("forName",
+                        "(Ljava/lang/String;)Ljava/lang/Class;");
+                StringReference referenceString = virtualMachine.mirrorOf(
+                        "java.lang.StringUTF16");
+                ThreadReference threadReference = virtualMachine.
+                        allThreads().get(0);
+                ct.invokeMethod(threadReference, m, Collections.
+                        singletonList(referenceString), 0);
+                possibleClasses = virtualMachine.classesByName(
+                        "java.lang.StringUTF16");
+            }
+            ReferenceType utf16;
+            if (possibleClasses.size() == 1){
+                utf16 = possibleClasses.get(0);
+            }
+            else {
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            Field hiByteShiftField = ReferenceTypeWrapper.fieldByName(utf16,
+                        "HI_BYTE_SHIFT");
+            if (hiByteShiftField == null){
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            Value hiByteShiftValue = utf16.getValue(hiByteShiftField);
+            if (!(hiByteShiftValue instanceof PrimitiveValue)){
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            boolean result = ((PrimitiveValue)hiByteShiftValue).intValue() == 0;
+            isLittleEndianCache.put(virtualMachine, result);
+            return result;
+        }
+    }
+
     private static void register(String shortedString, StringReference sr, int length, ArrayReference chars) {
         StringInfo si = new StringInfo(sr, shortedString.length() - 3, length, chars);
         synchronized (infoStrings) {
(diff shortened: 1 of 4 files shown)


SirIntellegence:
This is intended to (mostly) fix #4123

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -141,25 +148,59 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
...
+                        //is it compressed?
+                        final int LATIN1 = 0;
+                        Field coderField = ReferenceTypeWrapper.fieldByName(st,
+                                "coder");
+                        Value coderValue;
+                        if (coderField != null){
+                            coderValue = ObjectReferenceWrapper.getValue(sr,
+                                    coderField);
+                            if (coderValue instanceof IntegerValue &&
+                                    ((IntegerValue)coderValue).value() != LATIN1){
```
matthiasblaesing, line 185: This check failed on JDK 17 for me. I got a `ByteValue` at this point. Instead of getting to narrow just switch to `PrimitiveValue`, which has helpful accessors.

```suggestion
                            if (coderValue instanceof PrimitiveValue &&
                                    ((PrimitiveValue)coderValue).intValue() != LATIN1){
```

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -171,14 +212,43 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
             } else {
                 assert sa != null;
                 int l = AbstractObjectVariable.MAX_STRING_LENGTH;
+                if (isCompressedImpl && isUTF16){
+                    l *= 2;
+                }
```
matthiasblaesing, line 217: See below.

```suggestion
                List<Value> values = ArrayReferenceWrapper.getValues(sa, 0, isUTF16 ? (l * 2) : l);
```

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -171,14 +212,43 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
...
+                            characters[i] = c;
+                        }
+                    }
+                    else {
+                        //life is pain
+                        //We can't just inline code for this since... native
+                        //jazz and big/little endian stuff... so...
+                        //see StringUTF16.java
+                        //implement later!
+                        return "<Not Implemented>";
```
matthiasblaesing, line 241: To correctly decode a string we'd need to know the byte order of the target VM, but the following implemention should work correctly on little endian architectures (x86, arm, riscv), which should at least cover mainline onces.

```suggestion
                        // This assumes little endian encoding. This should work
                        // for most architectures (x86, arm, riscv), but will
                        // result in bogus data on big endian architectures
                        for (int i = 0; i < l; i++) {
                            int index = i * 2;
                            Value v = values.get(index);
                            if (!(v instanceof ByteValue)) {
                                return ERROR_RESULT;
                            }
                            Value v2 = values.get(index + 1);
                            if (!(v instanceof ByteValue)) {
                                return ERROR_RESULT;
                            }
                            char c1 = (char) ((ByteValue) v).byteValue();
                            char c2 = (char) ((ByteValue) v2).byteValue();
                            //remove the extended sign
                            c1 = (char) (0xFF & c1);
                            c2 = (char) (0xFF & c2);
                            // char bigEndianChar = (char) ((c1 << 8) | c2);
                            char litteEndianChar = (char) ((c2 << 8) | c1);
                            characters[i] = litteEndianChar;
                        }
```

SirIntellegence: Should we try to check the high and low byte shifts to determine which one it is or not worry about it?

matthiasblaesing: I  thought about that, but the problem is, that there is no indicator, what is "the right" shift. In my testcase, both shifts resulted in a valid code point. You could ask the target VM to encode a string, but then you can also directly determine the byteorder there.

matthiasblaesing:
Please add the necessary imports - the proposed changes also need corresponding import adjustments. Thank you.

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -130,7 +132,12 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
         try {
             ReferenceType st = ObjectReferenceWrapper.referenceType(sr);
             ArrayReference sa = null;
+            //only applicable if the string implementation uses a byte[] instead
+            //of a char[]
+            boolean isUTF16 = false;
+            boolean isCompressedImpl = false;
```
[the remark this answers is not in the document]
SirIntellegence: Aye! Variables have been renamed accordingly. I will hopefully get to testing it tomorrow? I had build issues not directly related to netbeans that prevented me from testing today.

SirIntellegence:
Ya. I am aware. I am still working in it. I did find out how to grab the hi byte field, so I added logic in for that. I will get to that tomorrow.

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -94,6 +95,42 @@ public static StringInfo getShortenedInfo(String s) {
         }
     }
 
+    private static boolean isSmallEndian(VirtualMachine virtualMachine) {
+        //TODO: Possibly cache this value
+        List<ReferenceType> possibleClasses = virtualMachine.classesByName(
+                "java.lang.StringUTF16");
```
matthiasblaesing, line 101: I think this should be cacheable with a WeakHashMap using the Virtualmachine as Key. This might be slow if the target VM is not local. I experimented a bit with this code and on JDK 17 initially `possibleClasses` was empty for me. Not sure how the VM handled this interernally, but I could make it work using this code:

```java
        if (possibleClasses.isEmpty()) {
            ClassType ct = (ClassType) virtualMachine.classesByName("java.lang.Class").iterator().next();
            Method m = ct.concreteMethodByName("forName", "(Ljava/lang/String;)Ljava/lang/Class;");
            StringReference referenceString = virtualMachine.mirrorOf("java.lang.StringUTF16");
            try {
                ThreadReference threadReference = virtualMachine.allThreads().get(0);
                ct.invokeMethod(threadReference, m, Collections.singletonList(referenceString), 0);
                possibleClasses = virtualMachine.classesByName("java.lang.StringUTF16");
            } catch (InvalidTypeException | ClassNotLoadedException | IncompatibleThreadStateException | InvocationException ex) {
            }
        }
```

SirIntellegence: Aye. Will get to that. Do we need to handle multiple threads or are we good without using concurrent structures here?

matthiasblaesing: In my test I had to make `isSmallEndian` synchronized - else I got into a `IncompatibleThreadStateException`. Not really sure whether this hammer just fixed the problem accidentally.

SirIntellegence: Ok. Should I slap a `synchronized` onto it, not worry about it, or find another solution?

matthiasblaesing: Works for me and looks sane. The caching you introduced works (at least in my trivial testcase).

One final idea would be to rework the synchronization. The maps are protected by individual `sychronized() {}` blocks. The pattern could be applied here two, which would also not leak the sychronization to the outside.

My suggestion is to move the `isLittleEndianCache.clear();` into a separate block `synchronized (isLittleEndianCache) {isLittleEndianCache.clear();}`, drop the `synchronized` from the signature of `isLittleEndian` and instead wrap the whole method contents into a `synchronized (isLittleEndianCache) {....}` block.

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -130,7 +213,14 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
         try {
             ReferenceType st = ObjectReferenceWrapper.referenceType(sr);
             ArrayReference sa = null;
+            //only applicable if the string implementation uses a byte[] instead
+            //of a char[]
+            boolean isUTF16 = false;
+            boolean isCompactImpl = false;
```
matthiasblaesing, line 219: Please add a comment like:
> JEP 254: Compact Strings after the boolean
that might help people to later understand the intention

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -171,21 +295,73 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
...
+                            }
+                            char c1 = (char) ((ByteValue) v).byteValue();
+                            char c2 = (char) ((ByteValue) v2).byteValue();
+                            //remove the extended sign
+                            c1 = (char) (0xFF & c1);
+                            c2 = (char) (0xFF & c2);
+                            char c = (char)(c1 << hiByteShift |
+                                    c2 << lowByteShift);
+                            // char bigEndianChar = (char) ((c1 << 8) | c2);
+                            //char litteEndianChar = (char) ((c2 << 8) | c1);
```
matthiasblaesing, line 344: Please remove the comment ed code

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -130,7 +213,14 @@ static String getStringWithLengthControl(StringReference sr) throws InternalExce
         try {
             ReferenceType st = ObjectReferenceWrapper.referenceType(sr);
             ArrayReference sa = null;
+            //only applicable if the string implementation uses a byte[] instead
+            //of a char[]
+            boolean isUTF16 = false;
+            boolean isCompactImpl = false;
+            int hiByteShift = 0;
+            int lowByteShift = 0;
```
matthiasblaesing, line 221: Please move these down to the use-site (directly before the if in 318) and remove the initializer. The if will initialize them.

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -92,6 +108,73 @@ public static StringInfo getShortenedInfo(String s) {
...
+            }
+            ReferenceType utf16;
+            if (possibleClasses.size() == 1){
+                utf16 = possibleClasses.get(0);
+            }
+            else {
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            Field hiByteShift;
```
matthiasblaesing, line 152: Please follow the naming convention you also used in the other method and make this `hiByteShiftField`

java/debugger.jpda/src/org/netbeans/modules/debugger/jpda/models/ShortenedStrings.java
```diff
@@ -92,6 +108,73 @@ public static StringInfo getShortenedInfo(String s) {
...
+                    ObjectCollectedExceptionWrapper | VMDisconnectedExceptionWrapper e){
+                Exceptions.printStackTrace(e);
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            if (hiByteShift == null){
+                isLittleEndianCache.put(virtualMachine, defaultValue);
+                return defaultValue;
+            }
+            Value val = utf16.getValue(hiByteShift);
```
matthiasblaesing, line 167: I suggest to switch variable to `hiByteShiftValue`.

matthiasblaesing (commented):
Final cleanup (see inline comments). Please also squash the commits into a single one.

SirIntellegence:
While applying those changes, noticed that `getStringWithLengthControl` is declared to throw some exceptions. Should `isLittleEndian` throw the internal exceptions as well instead of catching them?

SirIntellegence:
Pushing the exception lift change so it can be reviewed before squishing everything

matthiasblaesing:
To me this looks sane, thank you. It is an impressive list of Exceptions and we might want to refactor these calls later, but I think as a fix for the issue at hand this should be merged.

I requested a first CI run and the system is happy. So, I suggest that you squash the commit as you see fit. I would suggest to squash into a single commit as all commit cover the same area and there are not really smaller building blocks. If then GH actions is happy, I'll merge.

SirIntellegence:
And squashed

matthiasblaesing:
Thank you. As the unittests finally run cleanly through, lets get this in.

PR #5315, opened 2023-01-18
Moved Pattern.compile() stage to class level

tbw777:
Excluded creation of non variables Pattern objects at methods and cycles.
All patterns content (also with errors) are holded as is.


mbien (changes requested):
we can't do this kind of automated refactoring blindly
 - this has a memory impact and trades CPU time with memory, therefore it has to be a case by case decision
 - the fields need to have proper names

if you find patterns in utility methods or inner loops (aka "hot code") which are called very often, it is worth to extract them. But not as codebase wide automated refactoring.

tbw777:
@mbien 
How to recognize a utility method?

matthiasblaesing:
I agree with with @mbien without clear arguments where performance is improved this must not be merged.

This change hurts readability of the code as the declarations of the pattern and the use-site are hundreds of lines apart. The non-speaking names are an additional pain point, which make them a no-go. A declaration like `Pattern p = Pattern.compile` is fine a local context, where the use site is a few lines away and it is obvious what happens. With class level constants this needs to be better.

For the performance: @mbien pointed to the memory impact. A compiled pattern is held in memory, raising GC pressure as statics are not collected. Yes you will argue, that it is tiny and so I will argue, that most probably my CPU won't get a millikelvin warmer just because it has to recompile a `Pattern`. As @mbien already pointed out, this changes if you find situation where a something like this is done (this is a tight loop):

```java
for(int i = 0; i < LARGE_NUMER; i++) {
   Pattern p = Pattern.compile("CONSTANTSTRING"),
}
```

For this `p` can be pulled out of the loop without affecting readability, while improving performance. These cases are rare.

PR #5319, opened 2023-01-19
Simplified usage String.substring() method from two arguments to one …

tbw777:
…argument if applicable

One of String.substring() method implementation does require only one arg variant with detection of full length of the string inside. Therefore there is no needed to set length manually as second parametr.


tbw777:
@mbien this PR can be quick reviewed

mbien (approved):
looks good. tests are green too. lets merge.

PR #5328, opened 2023-01-20
Integrated try-with-resource with single style

tbw777:
Solved part of mem leaks. Reasons of leaks: unclosed streams and streams without closing if any exception occur


tbw777:
I havent errors on ide/bugtracking tests local

tbw777:
Also i found some places where transferTo can be used. Need to upgrade jvm bytecode version, also for collections too

matthiasblaesing:
> Also i found some places where transferTo can be used. Need to upgrade jvm bytecode version and this will helpfullty for for collections too

As mentioned now in multiple PR comments: Discuss these ideas on the mailing list. I know some developers who will not react kindly to bumping bytecode level, just to get access to new methods, if there is not a _very_ good reason.

_This_ is not the place to discuss ideas.

tbw777:
@matthiasblaesing 
ok, i entierly agree. Ty for fast feedback

PR #5333, opened 2023-01-20
Four small UI improvements, in various areas

eirikbakke:
Four small UI improvements (3 cosmetic + one focus handling adjustment), in various areas:
* Fix missing anti-aliasing for the Property Sheet's 'No Properties' message.
* Avoid the Navigator TopComponent itself grabbing focus when the user presses Tab from a focused component within the Navigator.
* Have SystemAction handle SVG icons properly.
* Make separator lines visible in the Database Explorer's 'connection properties' JTable.

### DIFF
--- a/ide/db/src/org/netbeans/modules/db/util/PropertyEditorPanel.form
+++ b/ide/db/src/org/netbeans/modules/db/util/PropertyEditorPanel.form
@@ -107,6 +107,7 @@
               </TableColumnModel>
             </Property>
             <Property name="columnSelectionAllowed" type="boolean" value="true"/>
+            <Property name="showGrid" type="boolean" value="true"/>
             <Property name="tableHeader" type="javax.swing.table.JTableHeader" editor="org.netbeans.modules.form.editors2.JTableHeaderEditor">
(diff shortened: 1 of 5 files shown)


mbien (commented):
i have no opinion on the navigator focus change but everything else makes sense to me.

a screenshot for the table tweak might be useful in case someone opens this PR via release notes.

eirikbakke:
1) To explain the focus adjustment thing a bit more: If you select e.g. a Java method in the Navigator, then the tree component gets focus. But if you then press tab a few times, you see that the focus indication "disappears" a few times during the focus cycle, due to components that are marked as focusable but which don't actually handle keyboard focus in any way. It's the child components of the NavigatorTC that should receive focus, but not the NavigatorTC itself (except when the Navigator is in the empty state, as alluded to in an earlier comment).

2) Sure, here's a before/after screenshot of the Connection Properties dialog:

gridlines

(The grid lines are rather light, but omitting them completely makes the dialog quite hard to understand.)

eirikbakke:
Thank you! Merging.

PR #5334, opened 2023-01-20
Added CompletionItem.shouldSingleClickInvokeDefaultAction to the Editor Completion API

eirikbakke:
This PR adds a method CompletionItem.shouldSingleClickInvokeDefaultAction() that can optionally be implemented by editor completion item implementations. If this method is overridden to return true, a completion item will be inserted after a single click, as opposed to the existing default behavior which is to require a double-click.

This allows platform applications to use the Editor Completion API to implement combobox-like behavior. E.g. in [Ultorg](https://www.ultorg.com/) below:

ComboBox

(The existing double-click behavior still makes sense as a default, because the user may be clicking the item once to show documentation.)

### DIFF
--- a/ide/editor.completion/src/org/netbeans/modules/editor/completion/CompletionLayout.java
+++ b/ide/editor.completion/src/org/netbeans/modules/editor/completion/CompletionLayout.java
@@ -367,48 +367,82 @@ public final class CompletionLayout {
                 completionScrollPane = new CompletionScrollPane(
                     editorComponent, listSelectionListener,
                     new MouseAdapter() {
+                        CompletionItem selectedItemOnPress;
+
+                        @Override
+                        public void mousePressed(MouseEvent evt) {
+                            if (!SwingUtilities.isLeftMouseButton(evt)) {
+                                return;
+                            }
+                            selectedItemOnPress = completionScrollPane.getSelectedCompletionItem();
+                        }
+
+                        @Override
+                        public void mouseReleased(MouseEvent evt) {
+                            /* Handle single-click via mouseReleased, rather than in mouseClick with
+                            getClickCount() == 1, since the latter kind of event will not be fired if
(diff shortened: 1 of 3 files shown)


Chris2011:
So as I understand it correctly, we can add now auto completion to JComboBox elements easily? I see some cases where I want to have this behavior.

ide/editor.completion/src/org/netbeans/modules/editor/completion/CompletionLayout.java
```diff
@@ -367,48 +367,82 @@ private void show(List data, String title, int anchorOffset, Point location,
                 completionScrollPane = new CompletionScrollPane(
                     editorComponent, listSelectionListener,
                     new MouseAdapter() {
+                        CompletionItem selectedItemOnPress;
+
+                        @Override
+                        public void mousePressed(MouseEvent evt) {
+                          if (!SwingUtilities.isLeftMouseButton(evt)) {
```
lbownik, line 374: wouldn't
if (SwingUtilities.isLeftMouseButton(evt)) {
        selectedItemOnPress = completionScrollPane.getSelectedCompletionItem();
}
be more readable?

matthiasblaesing (approved):
Looks sane to me.

eirikbakke:
Thank you! Will fix the indentation issue mentioned above and then merge after tests pass again.

mbien:
> So as I understand it correctly, we can add now auto completion to JComboBox elements easily? I see some cases where I want to have this behavior.

@Chris2011 this is off topic but you can actually do that fairly easy under some circumstances (simple completion). The maven settings panel for example has completion for a text field:
https://github.com/apache/netbeans/blob/4b7776a679963cd2d0797fd64fa11f8decda3c32/java/maven/src/org/netbeans/modules/maven/options/SettingsPanel.java#L186

mbien (approved):

PR #5336, opened 2023-01-20
Don't show the '+' separator in keyboard accelerators on MacOS

eirikbakke:
Keyboard accelerator strings, such as "Ctrl+R", are shown to the user in various places throughout NetBeans, e.g. in editor toolbar tooltips. On MacOS, the traditional display format for these is to use a special single-character symbol for modifier keys, and no "+" symbol. E.g. "⌘R" for Command+R. On NetBeans, though, we currently show this as "⌘+R".

This PR gets rid of the plus symbol on MacOS, to match the native convention. This is already how shortcuts are displayed in the main menu (where the accelerator display strings are generated by the OS).

### DIFF
--- a/platform/openide.awt/src/org/openide/awt/Actions.java
+++ b/platform/openide.awt/src/org/openide/awt/Actions.java
@@ -24,6 +24,7 @@ import java.awt.Dimension;
 import java.awt.EventQueue;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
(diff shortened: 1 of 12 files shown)


DevCharly:
`KeyEvent.getKeyModifiersText()` and `InputEvent.getModifiersExText()` on macOS have the problem that the order of the modifier symbols is different to the order used by macOS (left is Java, right is macOS:

Untitled Untitled 2 

That's why FlatLaf and Aqua L&F use own methods.
See: [FlatMenuItemRenderer.getMacOSModifiersExText()](https://github.com/JFormDesigner/FlatLaf/blob/546382e47170aa5b7132620b2cba32583d6011cc/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatMenuItemRenderer.java#L621-L644) and [AquaMenuPainter.getKeyModifiersUnicode()](https://github.com/openjdk/jdk/blob/92474f13f03afacc48541b0de17998998f70eb65/src/java.desktop/macosx/classes/com/apple/laf/AquaMenuPainter.java#L91-L126)

For right-to-left component orientation, Aqua L&F (and FlatLaf) reverses the order. E.g. Swing Aqua L&F:

Untitled 7

BTW would suggest to replace deprecated `KeyEvent.getKeyModifiersText()` with `InputEvent.getModifiersExText()`.

Tried this out. I see the change in the main toolbar, but not in editor/view toolbars and other areas. I fear that this kind of code is used in many places. It is also surprising that text "Ctrl" or "Shift" (instead of symbols) is used in some places on macOS.

Editor toolbar:

Untitled 3

Options Keymap:

Untitled 5

Editor macro shortcut:

Untitled 6

Editor popup:

Untitled 4

eirikbakke:
Thanks for these comments! I have pushed an improved version of this PR now, which gathers the required logic in a single new public API method Actions.keyStrokeToString, and uses it in more places. This includes the editor toolbar tooltips.

In the Keymap settings code, changing accelerator strings is a bit risky, since there may be parsing code that assumes the old format. So I've left the options.keymap module untouched for now.

There are also a few other places calling KeyEvent.getKeyText/getKeyModifiersText where presentation strings are generated in more complicated ways; I have left these for future work.

> KeyEvent.getKeyModifiersText() and InputEvent.getModifiersExText() on macOS have the problem that the order of the modifier symbols is different to the order used by macOS

Fixed, by using logic similar to the one you linked to in FlatLAF.

> BTW would suggest to replace deprecated KeyEvent.getKeyModifiersText() with InputEvent.getModifiersExText().

Done.

DevCharly (approved):
Works great 👍

platform/openide.awt/src/org/openide/awt/Actions.java
```diff
@@ -128,25 +129,100 @@ private static String findKey(Action action) {
...
-        String acceleratorText = ""; // NOI18N
+        return keyStrokeToString(stroke);
+    }
 
-        if (modifiers > 0) {
-            acceleratorText = KeyEvent.getKeyModifiersText(modifiers);
-            acceleratorText += "+"; // NOI18N
-        } else if (accelerator.getKeyCode() == KeyEvent.VK_UNDEFINED) {
-            return ""; // NOI18N
+    // Based on com.formdev.flatlaf.getMacOSModifiersExText.
```
eirikbakke, line 141: Meant com.formdev.flatlaf.ui.FlatMenuItemRenderer.getMacOSModifiersExText here in the comment... will fix before merging.

eirikbakke:
Thanks for reviewing, @DevCharly! Will wait a few days to see if there are any other comments since there is an API change involved. (One utility method added at Actions.keyStrokeToString.)

eirikbakke:
I pushed a change to fix the typo "com.formdev.flatlaf.getMacOSModifiersExText." to "com.formdev.flatlaf.FlatMenuItemRenderer.getMacOSModifiersExText" in a comment. Otherwise no changes.

mbien:
this has no milestone assigned. If you want to see this in NB 18, now would be a good time to merge :)

eirikbakke:
Great; merged!

PR #5337, opened 2023-01-20
[NETBEANS-5728] Fix spurious selections during drag and drop in ETable/OutlineView

eirikbakke:
In an OutlineView with multiple selection enabled, if the user tries to drag one row to a different position (e.g. to reorder rows when the underlying nodes support it), additional rows will be selected as the mouse moves over them during the drag operation. On MacOS, it happens 100% of the time. On Windows, I've seen this bug happen in the past as well, though much less frequently. On Windows the bug is merely annoying, but on MacOS the bug makes OutlineView's drag-and-drop feature unusable. This PR adds a workaround in ETable, the component used by OutlineView.

See my original ancient issue description from [2013](https://bz.apache.org/netbeans/show_bug.cgi?id=230690) and the more recent [NETBEANS-5728](https://issues.apache.org/jira/browse/NETBEANS-5728).

This PR also fixes a bug where dragging outside the ETable may cause Swing's autoscrolling events to continue firing indefinitely.

I don't know of any particular place in the IDE where ETable/OutlineView is used with drag-and-drop enabled, so this PR might be hard to test independently. But the patch in this PR has functioned successfully in the NetBeans Platform application [Ultorg](https://www.ultorg.com) for 1.5 years.

### DIFF
--- a/platform/o.n.swing.outline/src/org/netbeans/swing/etable/ETable.java
+++ b/platform/o.n.swing.outline/src/org/netbeans/swing/etable/ETable.java
@@ -41,6 +41,7 @@ import java.awt.event.KeyListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseListener;
+import java.awt.event.MouseMotionAdapter;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -1530,15 +1531,44 @@ public class ETable extends JTable {
         return header.getColumnModel().getColumn(columnIndex);
     }
 
+    private boolean scollingMouseMotionListenerAdded = false;
+
     /**
      * Adds mouse listener to the header for sorting and auto-sizing
-     * of the columns.
+     * of the columns. Also handle auto-scrolling during drag events.
      */
     private void updateMouseListener() {
         JTableHeader jth = getTableHeader();
         if (jth != null) {
             jth.removeMouseListener(headerMouseListener); // not to add it twice
             jth.addMouseListener(headerMouseListener);
+        }
+        if (!scollingMouseMotionListenerAdded) {
+            scollingMouseMotionListenerAdded = true;
+            /* Make sure auto-scrolling works even the selection does not change as a result of the
+            mouse dragging. See JComponent.setAutoscrolls. */
+            addMouseMotionListener(new MouseMotionAdapter() {
+                @Override
+                public void mouseDragged(MouseEvent evt) {
+                    Rectangle r = new Rectangle(evt.getX(), evt.getY(), 1, getRowHeight());
+                    scrollRectToVisible(r);
+                }
+            });
+            /* Under certain circumstances, at least on MacOS, the autoscroller can get "stuck",
+            continuing to issue new drag events after the user has released the mouse button. I
+            suspect it happens when the mouse button is released while the cursor is outside of the
+            root Java frame. Turning the autoscroller off and back on again will reset it. Do this
+            whenever the mouse enters the component, as this will usually resolve the situation
+            naturally. */
+            addMouseListener(new MouseAdapter() {
+                @Override
+                public void mouseEntered(MouseEvent e) {
+                    if (getAutoscrolls()) {
+                        setAutoscrolls(false);
+                        setAutoscrolls(true);
+                    }
+                }
+            });
         }
     }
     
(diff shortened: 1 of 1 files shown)


matthiasblaesing (commented):
I can reproduce the problem on linux and it seems to be fixed there too.

The method using the stack trace looks really ugly and here JDK 9+ stackwalker might be a nice use-case. I looked into the setup of the `TableUI`. There might be an alternative (wrap the UI and override the generated handler, but I did not really try it.

TL;DR it would be nice if someone with more Swing background could have a look.

eirikbakke:
Yeah, it's ugly. It's been a while since I worked on this, but I seem to remember trying various approaches, including subclassing BasicTableUI, and concluding the stack trace approach was the lowest-risk way to fix it.

I hadn't seen the StackWalker API before; does it have advantages over Exception.getStackTrace?

DevCharly:
Tried this. Works fine regarding "selection avoidance", but have the feeling that autoscrolling works not as good as before.

Have running two NB instances side-by-side (NB 16 from DMG and compiled master branch; on macOS).
Autoscrolling works pretty reliable in NB 16, but in the version with this PR, it often does not autoscroll.
Seems to depend on how fast I move the mouse out of the table...

Have tried a different solution:

- removed `mouseDragged` listener that invokes `scrollRectToVisible()`
- added following code from [JTable.changeSelection()](https://github.com/openjdk/jdk/blob/7dfe75cf553193faf709cff6b8b2505680d7cebc/src/java.desktop/share/classes/javax/swing/JTable.java#L2516-L2521) into "veto return" block in `ETable.changeSelection()`

I think this works a little bit better, but not as reliable as before. Don't know why 😕 

Anyway, for compatibility, I would move autoscrolling to `ETable.changeSelection()` as done in `super.changeSelection()`.
This also avoids an additional listener and a possible side-effect of always scrolling while dragging.

DevCharly:
> I hadn't seen the StackWalker API before; does it have advantages over Exception.getStackTrace?

It should be faster: https://www.sitepoint.com/deep-dive-into-java-9s-stack-walking-api/
But don't think that this would make a difference because it is not invoked very often here.

I'm using StackWalker in FlatLaf for similar purposes here: [StackUtilsImpl for Java 9+](https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/java9/com/formdev/flatlaf/ui/StackUtilsImpl.java)
Here is [StackUtilsImpl for Java 8](https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/StackUtilsImpl.java) and base class [StackUtils](https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/StackUtils.java)

DevCharly:
Update: just tried a DEV build without this PR, and autoscrolling is also not as reliable as in NB 16 😕 

Selection-while-dragging also behaves slightly different (in Search results view):

- in NB 16, rows under mouse are selected immediately
- in DEV build, there is some delay and not all rows are selected

Both NB versions use same Java 11 VM.
Maybe this is a issue on my machine.
Or NB DEV build is started with different options than installed NB 16.
Or something else changed in NB 17...
BTW I've not checked out the commit of this PR. I'm on latest master and cherry picked this PR

Is there a DMG of NB 17rc available anywhere

neilcsmith-net:
@DevCharly you'll have to use the zip at https://nightlies.apache.org/netbeans/candidate/netbeans/  There are unsigned installers for [other platforms](https://nightlies.apache.org/netbeans/candidate/installers/) but not a DMG.

platform/o.n.swing.outline/src/org/netbeans/swing/etable/ETable.java
```diff
@@ -1530,16 +1531,45 @@ private TableColumn getResizingColumn(Point p) {
...
      */
     private void updateMouseListener() {
         JTableHeader jth = getTableHeader();
         if (jth != null) {
             jth.removeMouseListener(headerMouseListener); // not to add it twice
             jth.addMouseListener(headerMouseListener);
         }
+        if (!scollingMouseMotionListenerAdded) {
+            scollingMouseMotionListenerAdded = true;
+            /* Make sure auto-scrolling works even the selection does not change as a result of the
```
eirikbakke, line 1548: Typo: Should insert "when" here ("even when").

PR #5338, opened 2023-01-20
Copy FlatLAF label positioning improvements into the Windows LAF

eirikbakke:
This PR copies the previously merged [text alignment improvements](https://github.com/apache/netbeans/pull/4349) in FlatLAF window system tabs back to the older Windows LAF.

This completes the tab component saga originally started in https://github.com/apache/netbeans/pull/3115 ; this PR corresponds to the last commit of the latter.

At some point we might want to move the FlatLAF tab control implementations into o.n.swing.tabcontrol so the code can be shared, but they have some dependencies on FlatLAF-specific UI scaling utilities and so on. But the Windows LAF will probably be obsoleted by FlatLAF soon anyway.

### DIFF
--- a/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/WinFlatEditorTabCellRenderer.java
+++ b/platform/o.n.swing.tabcontrol/src/org/netbeans/swing/tabcontrol/plaf/WinFlatEditorTabCellRenderer.java
@@ -21,13 +21,14 @@ package org.netbeans.swing.tabcontrol.plaf;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Dimension;
-import java.awt.FontMetrics;
+import java.awt.Font;
 import java.awt.GradientPaint;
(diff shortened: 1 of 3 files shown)


neilcsmith-net:
Sounds good.  I don't really use Windows, so possibly not the best person to review. Maybe you could add a before / after screenshot?

eirikbakke:
I did some before/after comparisons on the Windows LAF and decided a one-pixel adjustment in a LAF-specific Y offset was needed ("+ 1" instead of "+ 2" in WinFlatViewTabDisplayerUI and similarly "1 +" instead of "2 +" in WinFlatEditorTabCellRenderer).

After this change, the tabs in fact look identical before and after this PR... (tested on 100% HiDPI scaling, on 150% HiDPI scaling, and with "--fontsize 20" on 100% HiDPI scaling)

So the only benefit of this PR is harmonizing the copy/pasted between FlatLAF tabs and the old Windows LAF.

DevCharly (approved):

PR #5339, opened 2023-01-21
added missing record enums to editor module.

mbien:
based on delivery

checks for `ElementKind.RECORD` are missing all over the codebase. The lexer isn't even record aware. This PR only covers `java.editor`. I had a larger changeset which checked all usages of the CLASS enum and added RECORD on a case by case basis. But it is probably not a good idea to do this blindly without being able to validate it properly.

this PR fixes three small bugs which are easily verifiable:

 - tooltip on unused records will use the right bundle msg
 - javadoc completion for records
 - goto impl is now working for records

test code:

```java
package bugs;

public class RecordEnums {

    /**
     * @see bugs.RecordEnums.
     *      check completion ^^^ should have records
     */
    record ARecord() { }
    
    record UnusedRecord() { }
    // mouse over ^^^ should have tooltip
    
    class AClass { }
    
        ARecord r = new ARecord();
    //  ^^^ use navigate -> goto implementation
    }
}
```

took me a while to get to this. Wanted to enable the lexer tests first which required other CI related tasks, but its all done now :)

### DIFF
--- a/java/java.editor/src/org/netbeans/modules/java/editor/javadoc/JavadocCompletionTask.java
+++ b/java/java.editor/src/org/netbeans/modules/java/editor/javadoc/JavadocCompletionTask.java
@@ -102,7 +102,10 @@ public class JavadocCompletionTask<T> extends UserTask {
         T createPackageItem(String pkgFQN, int startOffset);
     }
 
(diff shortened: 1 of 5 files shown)


lkishalmi (approved):
Trivial, but important changes.
Looks good to me! Thank you!

mbien:
@lkishalmi thanks for reviewing!

matthiasblaesing (approved):
Looks sane to me.

PR #5343, opened 2023-01-22
UI: fixed conditional gradle version info label.

mbien:
targets delivery

#5274 overlooked an invisible label which is only visible if an unsupported version is selected in the combo box. This fixes the layout so that making the label visible does not disturb other UI components.

The label got moved above the combobox, since NB16 has the problem that the combobox can collapse to 0 width if the window is not large enough before the info label is made visible. This prevents this from happening, while retaining the same functionality. 

result:
unsupported-gradle-version


lkishalmi (approved):
Just tested the original code. That worked fine for me.
This one looks as odd as the other one.

I wrote that code 7-8 years ago (NetBeans 8.1 days). Both the daemon and wrapper were kind of new incubating technologies back then. Projects were debating on integrating the `gradle-wrapper.jar` into the source code, as that's a binary.

So for NB18 we could let the Gradle Distribution option go, just keep the Gradle Home.

mbien:
ah you did test it. I didn't even know that the label was there, it was a coincidence that it ended up at this place when I adjusted the other components in the first PR :)

PR #5344, opened 2023-01-22
Fixed detection of magic methods

KacerCZ:
In PHP method names are case insensitive.
Method isMagic() did not work for methods __callStatic, __toString and __debugInfo. Method isConstructor() should compare whole string even though this is already guarded by isMagic().


junichi11:
I think we can change the target to NB17.

junichi11 (approved):

KacerCZ:
OK, now PR targets NB17.
Thank you.

PR #5352, opened 2023-01-23
Springboot artifacts fix

sdedic:
There's a springboot module on the Plugin portal, but that one does not handle native-image compilation. Longer term, I'd like to create a dedicated native-image supporting module(s) and move the gradle+maven plugin support there. But for NBLS immediate release and NB17 release, I'd like to include a very trivial springboot support - Micronaut module is a good enterprise cluster candidate to make homeplace for it.

The PR adds ability to run native compilation from the NetBeans IDE based on springboot plugin support. In addition, springboot projects are configured differently to micronaut projects - the changes in `MicronautPackagingArtifactsImpl`  make the impl more compatible with various setups, as it now reacts on the plugin's goal regardless of whether the plugin is bound to the buildchain.

The maven module part is a necessary subtle API change - the friend API did allow to create `RunConfig`s for an explicit goal, but not for an (abstract) project action  which is needed here, and uses possible user's customizations for the action (and therefore the goal). It fits well to the `RunUtils` API function set.

I'd like to include this PR into NB17 release.

### DIFF
--- a/java/maven/nbproject/project.properties
+++ b/java/maven/nbproject/project.properties
@@ -22,7 +22,7 @@ javadoc.apichanges=${basedir}/apichanges.xml
 javadoc.arch=${basedir}/arch.xml
 javahelp.hs=maven.hs
 extra.module.files=maven-nblib/
-spec.version.base: 2.156
+spec.version.base=2.157.0
 
 # The CPExtender test fails in library processing (not randomly) since NetBeans 8.2; disabling.
 test.excludes=**/CPExtenderTest.class
--- a/enterprise/micronaut/nbproject/project.xml
+++ b/enterprise/micronaut/nbproject/project.xml
@@ -74,7 +74,7 @@
                     <compile-dependency/>
                     <run-dependency>
                         <release-version>2</release-version>
-                        <specification-version>2.156</specification-version>
+                        <specification-version>2.157</specification-version>
                     </run-dependency>
                 </dependency>
                 <dependency>
--- a/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/MicronautPackagingArtifactsImpl.java
+++ b/enterprise/micronaut/src/org/netbeans/modules/micronaut/maven/MicronautPackagingArtifactsImpl.java
@@ -40,6 +40,8 @@ import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.netbeans.api.project.Project;
 import org.netbeans.api.project.ProjectActionContext;
 import org.netbeans.modules.maven.api.NbMavenProject;
+import org.netbeans.modules.maven.api.execute.RunConfig;
+import org.netbeans.modules.maven.api.execute.RunUtils;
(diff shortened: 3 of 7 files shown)


dbalek (approved):
Looks fine as a temporary solution.

neilcsmith-net (dismissed):
In current state, -1 for NB17.  Haven't looked at the details of the implementation, but if this is to be included in NB17 it needs to be based on delivery and all spec version changes removed.

I also don't see the justification for inclusion in NB17.

MartinBalin:
I agree with what Neils says. But, without this fix which is needed for some of the GraalVM+SpringBoot support in VSNetBeans we will need to either diverge VSNetBeans code from NetBeans code or do a hacky solution elsewhere... I propose to do this and then fix it properly in master after NB17?

neilcsmith-net:
@MartinBalin nothing stopping the changes going into delivery and NB17 if you want them there, as long as the PR is updated to allow merging to delivery.

@sdedic assertion above about spec versions is wrong.  The Maven module is incorrectly configured here as far as I can see (comment inline with code).  It is certainly skipped when spec versions get incremented. So we won't get a merge conflict with master, but should fix it I think?  And once delivery is synced, increment again?

java/maven/nbproject/project.properties
```diff
@@ -22,7 +22,7 @@ javadoc.apichanges=${basedir}/apichanges.xml
 javadoc.arch=${basedir}/arch.xml
 javahelp.hs=maven.hs
 extra.module.files=maven-nblib/
-spec.version.base: 2.156
+spec.version.base: 2.157
```
neilcsmith-net, line 25: As far as I can see, this should actually be?

`spec.version.base=2.157.0`

Use of colon means it gets skipped in increment task, and if not skipped needs to end with `.0`?

sdedic: @neilcsmith-net - I am confused; so should I rebase against `delivery` plus adjust spec numbers to fit into `delivery` codebase ?

neilcsmith-net: @sdedic yes, it's confusing!  To get into NB17 it needs to be based on top of `delivery`.  Normally, spec versions are fixed from branch, because the first thing we do on `master` at branch is a spec version increment (eg. #5312 )

However, the `maven` (and `java.editor.lib`) modules use a colon rather than equals in the properties file, and are being missed out of that increment.  I've only just noticed that.

IMO, unless there's a reason those modules shouldn't be incrementing every release, we should change this to what I suggested in my previous comment, and keep your API comments as is, when rebasing on top of delivery.

sdedic:
Rebased onto `delivery`. Maven's spec version changed to use `=` and to `2.157.0`

neilcsmith-net:
Thanks!  That looks OK to me from perspective of merging to delivery.

I'll leave @MartinBalin to review in full and handle merging.

MartinBalin (approved):

PR #5358, opened 2023-01-24
performance improvements - no logic changes

lbownik:
Replaced some StringBuffers with StringBuilders
fixed some "concatenation in append" issues
removed some unused objects
fixed one potential file descriptor leak


tbw777:
Im save your time: jvm unseting sync flag for local StringBuffers and therefore no any speed changes

tbw777:
One of the most mysterious tasks is to choose the right starting size for StringBuilder() )

mbien (changes requested):
- `StringBuffer` -> `StringBuilder` changes are fine, looks like none of the objects escape.
- the ARM block change is also fine

- the `TreeSet` removal doesn't look correct since it was supposed to sort the array -> we have the first logic change which the title claims to avoid
- please revert the append changes since `+` concatenation will be faster in 95% of all cases and is easier to read. This code runs a wizard which iterates over a few items, not a game engine so it doesn't really matter.

also please update the commit message and PR title to code cleanup and mention what module it covers, since performance improvements would require some kind of proof as in measurements or a benchmark.

lbownik:
@mbien with things that need to be reverted what will be left is not really worth merging, closing this PR, sorry for waisting your time

P.S. using TreeSet to sort an array is ... unexpected :)

PR #5361, opened 2023-01-25
Various database fixes

sdedic:
"Refresh" action is available on OCI nodes including the leaves; but a leaf has no `factory` set, so refresh is likely to NPE. 

The `Connect` action implementation was written so that it blocked the UI thread with a modal dialog and/or sync with the background DB connection. It should have been rewritten probably more thoroughly, but at the least I've changed the action to `asynchronous=true` - I needed to restructure it to implement `ActionListener` + `ContextAwareAction` so that it properly reports DB connection changes through action's enable once instantiated. Asynchronous=true does not work well for `NodeActions` (default constructor must not be present).

Asynchronous processing is desirable in the headless mode, when the blocked "ui" thread is actually usually the message-processing one, which may block the client from all messages to/from the NBLS server. In particular progress messages are blocked, so if the connect operation takes some few seconds, the user is not given any feedback, although the connect action reports the progress.

As a result even Swing-bound parts now run asynchronously, so the UI parts were wrapped into `invokeLater`/`invokeAndWait`.

### DIFF
--- a/ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
+++ b/ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
@@ -23,13 +23,12 @@ package org.netbeans.modules.db.explorer.action;
 
 import java.awt.Dialog;
 import java.awt.Dimension;
-import java.awt.EventQueue;
 import java.awt.GraphicsEnvironment;
-import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.lang.reflect.InvocationTargetException;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -39,7 +38,10 @@ import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.swing.AbstractAction;
+import javax.swing.Action;
 import javax.swing.JComponent;
+import javax.swing.SwingUtilities;
 import org.netbeans.api.progress.ProgressHandle;
 import org.netbeans.api.progress.ProgressHandleFactory;
 import org.netbeans.lib.ddl.DDLException;
@@ -50,11 +52,9 @@ import org.netbeans.modules.db.explorer.DatabaseConnection.State;
 import org.netbeans.modules.db.explorer.DbUtilities;
 import org.netbeans.modules.db.explorer.dlg.ConnectPanel;
 import org.netbeans.modules.db.explorer.dlg.ConnectProgressDialog;
-import org.openide.nodes.Node;
 import org.netbeans.modules.db.explorer.dlg.ConnectionDialog;
 import org.netbeans.modules.db.explorer.dlg.ConnectionDialogMediator;
 import org.netbeans.modules.db.explorer.dlg.SchemaPanel;
-import org.netbeans.modules.db.explorer.node.ConnectionNode;
 import org.openide.DialogDescriptor;
 import org.openide.DialogDisplayer;
 import org.openide.NotifyDescriptor;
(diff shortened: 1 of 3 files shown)


sdedic:
@neilcsmith-net - we discussed with @MartinBalin  that we'd like to include this one (together with #5362) in `delivery` (and NB17) since it has some UI impact on vsnetbeans. If you'd agree (this PR is not that trivial), I will rebase this onto `delivery`.

neilcsmith-net:
That's your call, not mine.  Rebase on delivery, add the reasoning for targetting NB17, and open to re-review on that basis.  Would like to see that UI impact on the IDE has been considered in review.

Not an area I know well enough or use to want to get too involved in reviewing, although I have a couple of concerns about the approach (will add inline).

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
```diff
@@ -62,63 +63,63 @@
...
 import org.openide.util.Mutex;
 import org.openide.util.NbBundle;
+import org.openide.util.RequestProcessor;
+import org.openide.util.WeakListeners;
 
 @ActionRegistration(
         displayName = "#Connect", 
-        lazy = false,
+        lazy = true,
+        asynchronous = true,
```
neilcsmith-net, line 77: A pet hate pattern of mine - IMO actions should always be invoked on the EDT and then explicitly async when needed.

Are there implications to changing the behaviour of an existing action here?

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
```diff
@@ -62,63 +63,63 @@
...
         enabledOn = @ActionState(type = DatabaseConnection.class, useActionInstance = true)
 )
 @ActionID(category = "Database", id = "netbeans.db.explorer.action.Connect")
 @ActionReference(path = "Databases/Explorer/Connection/Actions", position = 100)
-public class ConnectAction extends BaseAction {
+// Note: the action ought to be asynchronous=true, but that does not work well with lazy = true (and with lazy = false does not work at all)
```
neilcsmith-net, line 82: This comment seems to say the opposite to the changes you've made?!

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
```diff
@@ -161,10 +162,18 @@ public void exceptionOccurred(Exception exc) {
...
-            showDialog(dbcon, showDialog);
+            // in headless mode EDT makes no sense; but we should not block the main thread.
+            // in EDT, the nested dialog will make a nested event loop, so OK.
+            if (GraphicsEnvironment.isHeadless()) {
+                RequestProcessor.getDefault().post(() -> showDialog(dbcon, showDialog));
+            } else {
+                showDialog(dbcon, showDialog);
+            }
         }
+        */
```
neilcsmith-net, line 176: Is this meant to be left in?

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
sdedic: Eh, thanks for catching the comment leftovers ... There was an 'interim' state when the 'asynchronous' seem NOT to work, but it was all because I've left the default parameter-less constructor in there, and the Annotation Processor ignored lazy attribute completely (I'll fix that in another PR, this time for NB18 ... some warning message from AP)

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
sdedic: Yes. In NB IDE it does only a little harm, as waiting on the dialog creates a nested event loop, and everything incl the 'blocked' EDT thread works (as the nested event loop services events). 
But in headless mode, if the [action's code blocks](https://github.com/apache/netbeans/pull/5361/files#diff-033fc81193b23efa97939feead127364383d694ff672ba7d5d1612c37cf499dfL331), the caller message handling thread blocks, so the LSP client gets no progress messages etc.

There were two options, actually: explicit offload the work to a RequestProcessor if headless, or this one. I'd slightly prefer `asynchronous = true` as it allows (eventually) the infrastructure to track what actions are running (not used now).

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
sdedic: Addressed in 73d57201fd3c

ide/db/src/org/netbeans/modules/db/explorer/action/ConnectAction.java
neilcsmith-net: Generally prefer the RequestProcessor approach, but hadn't considered the infrastructure tracking side.  Just been caught a few times (mainly in RCP work) by actions triggering chains of things unexpectedly off of the EDT.

I understand why you want to change it for the headless mode.  I assume there are no existing usages where the action not being completed by return is a problem?

Anyway, assuming no implications to changing the existing action, no concerns with it as it is.

jhorvath (approved):
Looks good

neilcsmith-net:
> @neilcsmith-net - we discussed with @MartinBalin that we'd like to include this one (together with #5362) in `delivery` (and NB17) since it has some UI impact on vsnetbeans. If you'd agree (this PR is not that trivial), I will rebase this onto `delivery`.

@sdedic @MartinBalin you still want to do this?  We'll sync for 17-rc3 tomorrow, so if it's not in by then and not flagged as a priority to trigger another rc, it won't make NB17.

eirikbakke:
I think the "Turned Connect action to asynchronous" commit may have broken the public DatabaseConnection.showConnectionDialog API method, which should work from the EDT, but now doesn't:

```
java.lang.Error: Cannot call invokeAndWait from the event dispatcher thread
	at java.desktop/java.awt.EventQueue.invokeAndWait(EventQueue.java:1351)
	at java.desktop/java.awt.EventQueue.invokeAndWait(EventQueue.java:1344)
	at java.desktop/javax.swing.SwingUtilities.invokeAndWait(SwingUtilities.java:1480)
	at org.netbeans.modules.db.explorer.action.ConnectAction$ConnectionDialogDisplayer.showDialog(ConnectAction.java:254)
	at org.netbeans.modules.db.explorer.DatabaseConnection$5.run(DatabaseConnection.java:1230)
	at org.netbeans.modules.openide.util.NbMutexEventProvider$Event.doEvent(NbMutexEventProvider.java:97)
	at org.netbeans.modules.openide.util.NbMutexEventProvider$Event.readAccess(NbMutexEventProvider.java:72)
	at org.netbeans.modules.openide.util.LazyMutexImplementation.readAccess(LazyMutexImplementation.java:66)
	at org.openide.util.Mutex.readAccess(Mutex.java:242)
	at org.netbeans.modules.db.explorer.DatabaseConnection.showConnectionDialog(DatabaseConnection.java:1227)
	at org.netbeans.api.db.explorer.ConnectionManager.showConnectionDialog(ConnectionManager.java:343)
```

I'll test a fix that simply changes invokeAndWait to "Mutex.EVENT.writeAccess((Mutex.Action<Void>) () -> { ... })", but if you can think of other potential problems like this, let me know...

PR #5362, opened 2023-01-25
Report progress start when handle immediately switches to indeterminate.

sdedic:
During testing of #5361, I've found out that the `LspInternalHandle` receives SWITCH event as the first one (as a result of the switch to indeterminate status), and does not report start event to the LSP client at all. As a result the progress indicator is never created in LSP client and the progress for long-running operation may remain unreported.

`sendProgress` will first send `start` event to the client, if not already started.


MartinBalin (approved):

dbalek (approved):

neilcsmith-net:
You still wanting to get #5361 into NB17? In which case, this should be too?

PR #5363, opened 2023-01-25
Proper escaping of TNS_ADMIN path. Select JDBC driver with defined implementation.

sdedic:
This PR fixes two issues with OCI database support:

 The `Download Wallet` action writes a JDBC connection string with a pathname in it, but the pathname is not properly escaped. This is not an issue on MacOS / Linux, but on Windows that uses backslashes as dir separators, the backslashes in string content are interpreted as escapes by the JDBC driver, and the resulting path will be broken.

Second, the action tries to find the appropriate JDBC driver, but `JDBCDriverManager.getDefault().getDrivers()` call may reeturn half-configured drivers, that have no JAR files associated with them. This PR makes the action to prefer such drivers that define jars with the actual driver code.
The original behaviour had the effect of selecting a wrong driver, i.e. OracleThin driver (without JARs) was persisted while there was full Oracle jdbc driver present (that was configured with appropriate code jars).

### DIFF
--- a/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
+++ b/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
@@ -19,14 +19,16 @@
 package org.netbeans.modules.cloud.oracle.actions;
 
 import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -45,6 +47,8 @@ import org.openide.awt.ActionReference;
 import org.openide.awt.ActionReferences;
 import org.openide.awt.ActionRegistration;
 import org.openide.awt.StatusDisplayer;
+import org.openide.filesystems.URLMapper;
+import org.openide.util.BaseUtilities;
 import org.openide.util.ContextAwareAction;
 import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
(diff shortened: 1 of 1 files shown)


MartinBalin (approved):

enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
```diff
@@ -140,6 +163,16 @@ public void actionPerformed(ActionEvent ev) {
         });
     }
     
+    static boolean isAvailable(JDBCDriver driver) {
+        URL[] urls = driver.getURLs();
+        for (int i = 0; i < urls.length; i++) {
```
lbownik, line 168: how about using for-each loop?

sdedic:
This will be also needed for vsnetbeans, I guess I should rebase to `delivery` (@neilcsmith-net) ?

neilcsmith-net:
If it needs to go into NB17, then yes - base branch of PR needs changing to delivery, and commit rebasing on top of delivery if necessary.

enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
sdedic: Addressed in c158a2ca0119

sdedic:
(marked as NB17, will rebase after @jhorvath's review, should be clean)

enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
```diff
@@ -105,7 +110,22 @@ public void actionPerformed(ActionEvent ev) {
...
+                            if (isAvailable(d)) {
+                                jarsPresent = d;
+                                break;
+                            }
+                        }
+                        if (jarsPresent == null) {
+                            jarsPresent = drivers[0];
+                            // PENDING: should be shown to the user ?
+                            LOG.log(Level.WARNING, "Unable to find driver JARs for wallet {0}, using fallback driver: {1}", new Object[] { walletPath, jarsPresent.getName() });
+                        }
```
jhorvath, line 128: Shouldn't we display message that the driver is not installed here? Driver with no jars is useless here.

jhorvath (dismissed):
looks good

enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
sdedic: For NB IDE usage, it's not an issue, as driver's jars can be configured when Connect... action is used. But yes, I can probably display an info dialog - but still create the connection.

sdedic: Addressed in 7ce696e

sdedic:
Rebased onto `delivery`

PR #5365, opened 2023-01-26
Bulk added some missing ElementKind.RECORD enums.

mbien:
#5339 had the bits which were easily verifiable as bugs.

This PR is the bulk version of it. I searched for `ElementKind.CLASS` usage and tried to find out on a case by case basis if `RECORD` was missing based on what the code does. I didn't actually try to reproduce issues in the IDE for most of this since this would be too time consuming for all changes.

I am not sure if this is the right approach of doing this, but most changes here seemed fairly obvious/trivial so I hope this doesn't break anything.

This doesn't investigate missing `RECORD_COMPONENT` enums or add the record token to the lexer.

### DIFF
--- a/platform/openide.util/src/org/netbeans/modules/openide/util/NbBundleProcessor.java
+++ b/platform/openide.util/src/org/netbeans/modules/openide/util/NbBundleProcessor.java
@@ -55,6 +55,8 @@ import org.openide.util.NbBundle;
 import org.openide.util.NbCollections;
 import org.openide.util.lookup.ServiceProvider;
 
+import static javax.lang.model.element.ElementKind.PACKAGE;
+
 @ServiceProvider(service = Processor.class)
 public class NbBundleProcessor extends AbstractProcessor {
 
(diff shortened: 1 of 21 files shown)


mbien:
this seems to cause some issues with the reformatter, haven't investigated any further yet.

singh-akhilesh (approved):
LGTM

platform/openide.util/src/org/netbeans/modules/openide/util/NbBundleProcessor.java
```diff
@@ -309,8 +311,12 @@ private String findCompilationUnitName(Element e) {
         case INTERFACE:
         case ENUM:
         case ANNOTATION_TYPE:
-            switch (e.getEnclosingElement().getKind()) {
-            case PACKAGE:
+            if (e.getEnclosingElement().getKind() == PACKAGE) {
+                return e.getSimpleName().toString();
+            }
+        }
+        if ("RECORD".equals(e.getKind().name())) {
```
lkishalmi, line 318: I guess the string matching is due to future compatibility, is it?
The indentation of the switch case block is bad, but I guess that how it was before this change, might fix that for the readability.

lkishalmi (approved):
Looks good. Thanks!

platform/openide.util/src/org/netbeans/modules/openide/util/NbBundleProcessor.java
mbien: The build fails without it since platform builds against 8 without nb-javac in the classpath.

Java modules can "cheat" and use the RECORD enum directly since nb-javac adds it to the classpath due to the fact that it is currently based on javac of JDK 19.

So its technically about backwards compatibility with JDK 8.

mbien: updated the formatting of the switch - I agree that this was confusing. Also added a comment.

mbien:
rebased to latest master

mbien:
merging. Btw I can't reproduce the [formatter issues](https://github.com/apache/netbeans/pull/5365#issuecomment-1405999583) anymore. I probably didn't do a full rebuild while working on #5374 in parallel and got confused.

PR #5332, opened 2023-01-20
Improved Map/Set usage

tbw777:
- keySet().contains() -> containsKey()
- values().contains() -> containsValue()
- keySet() -> entrySet() - both keys and values are used
- keySet() -> values() - keys used only with get()
- entrySet() -> values() - keys not used
- entrySet() -> keySet() - values not used
- remove() -> removeAll() - applicable
- put() -> putAll()

Targets: performance, structure, clean code

PR #5368, opened 2023-01-26
StringBuilder related performance improvements - no logic changes

lbownik:
fixed few "concatenation in append" issues
initialized String builders to larger size to mitigate internal realocations
removed one unnecessary StringBuilder creation


tbw777:
I have a branch where append is fixed for 2000+ files. But I'm not allowed to make that kind of change anymore

tbw777:
`JavaCodeTemplateProcessor.java:212: error: ')' expected`

neilcsmith-net:
Please at least check things build and run before opening a PR.  Thanks!

matthiasblaesing:
@lbownik  and @tbw777 
> But I'm not allowed to make that kind of changes anymore

This is nonsense. There is a small number of active reviewers. @BradWalker gives a pretty good advice here:

https://github.com/apache/netbeans/pull/5332#issuecomment-1399174804

If you did not already do so, I strongly advice you to subscribe to dev@netbeans.apache.org (https://netbeans.apache.org/community/mailing-lists.html#dev), where you can raise questions and read about current discussions.

I think it is worth reading.

PR #5372, opened 2023-01-26
Import local schema to avoid DNS lookup in build

jgneff:
This is the same fix as the one made to `jakartaee_9.xsd` in NetBeans 16 for #4920, but to the new file `jakartaee_10.xsd` in NetBeans 17. It allows NetBeans to be built in environments with very strict firewalls, such as Launchpad, by removing the only DNS lookup during the build.

I'd like to make sure NetBeans can build on Launchpad because that's the easiest way to publish Snap packages of NetBeans for all supported Linux architectures: amd64, arm64, armhf, and i386.

### DIFF
--- a/enterprise/j2ee.dd/src/org/netbeans/modules/j2ee/dd/impl/resources/jakartaee_10.xsd
+++ b/enterprise/j2ee.dd/src/org/netbeans/modules/j2ee/dd/impl/resources/jakartaee_10.xsd
@@ -39,7 +39,7 @@
 
 
   <xsd:import namespace="http://www.w3.org/XML/1998/namespace"
-              schemaLocation="http://www.w3.org/2001/xml.xsd"/>
+              schemaLocation="xml.xsd"/>
 
   <xsd:include schemaLocation="jakartaee_web_services_client_2_0.xsd"/>
 


matthiasblaesing:
Thanks for catching this, the change looks good. This needs to be cherry-picked onto delivery. Changes for NB17 need to come from branch `delivery`, changes in `master` go to NB18.

jgneff:
Thanks, Matthias. I wasn't sure whether I was too late for NetBeans 17. Is that something I should do (cherry-pick this onto `delivery`)?

matthiasblaesing:
There might or might not be a new RC, if there is this should be in. I added it to the NB17 milestone and changed the target branch to `delivery`, but you can see in the list of commits, that merging as is would cause multiple commits to be pulled in.

There are two options:

a) You create a new pull request that is based on `delivery` and opened with target `delivery` or
b) You feel happy with the git tools and do basicly this:

```bash
# Ensure the branch this PR is based on is active
git checkout import-local-schema
# Move it to a different name
git branch -m import-local-schema-on-master
# Create a new local branch with the old name
# (0e504e5911846f4f66b3d97cb5d1c884b83a50ab is current HEAD of `delivery`)
# maybe you need to `git fetch` from the github repository of apache netbeans
git checkout -b import-local-schema 0e504e5911846f4f66b3d97cb5d1c884b83a50ab
# Cherry pick your commit into the new branch
git cherry-pick 2dce1801e795731248679ca061315db7e2fcd6fa
# Force push this branch into the original branch that created this PR (this assumes you named your upstream origin)
git push -f -u origin import-local-schema
```

neilcsmith-net:
You can rebase using 

`git rebase HEAD~1 --onto delivery` then force push.

Label just until this is done.

matthiasblaesing:
The intermediate test failure was a flaky commit validation I saw in the past and is fixed after restarting.

jgneff:
The following appears to have worked!

```console
$ git switch -c delivery upstream/delivery
$ git switch import-local-schema
$ git rebase --onto delivery import-local-schema~ import-local-schema
$ git push --force
```

mbien:
added labels otherwise this would show up under the `do-not-merge` category on the release notes :P
looks like its on the right branch too now.

to avoid having to look up the hash in the log, i often use a graphical tool for it like `gitg`. Where i can simply right click on a commit and select cherry pick onto which is super convenient.

but the basic concept is simply:
```
git checkout copy_of_delivery_branch
git cherry-pick somehash
```
thats it.

PR #5379, opened 2023-01-28
Proposal to change  `==` comparing constants with `equals`

tbw777:
Fixed erroneous direct form of comparison ("==") for these object types: String, Integer, Boolean, BigDecimal

Direct comparsion with == operator may cause unpredictable effects for business logic because its skip content check. Single exclusion is using method intern() for String. Replaced with Objects.equals() where left side might be a null and equals() otherwise. No any test directory was changed.
This change covers occurrences of the == operator except intern() Also expressions like (== "") changed to IsEmpty()

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
@@ -146,7 +146,7 @@ public final class ModuleDependency implements Comparable<ModuleDependency> {
         if (result != 0) { return result; }
         
         // do not force resolution of SPEC_VERSION_LAZY for comparisons, unnecessary
-        if (specVersion != SPEC_VERSION_LAZY || other.specVersion != SPEC_VERSION_LAZY) {
+        if (!SPEC_VERSION_LAZY.equals(specVersion) || !SPEC_VERSION_LAZY.equals(other.specVersion)) {
             String otherSpec = other.getSpecificationVersion();
             String spec = getSpecificationVersion();
             result = spec == null // spec versions may be null
@@ -167,7 +167,7 @@ public final class ModuleDependency implements Comparable<ModuleDependency> {
             ModuleDependency other = (ModuleDependency) o;
             return getCodeNameBase().equals(other.getCodeNameBase()) &&
                     Utilities.compareObjects(getReleaseVersion(), other.getReleaseVersion()) &&
-                    ((specVersion == SPEC_VERSION_LAZY && other.specVersion == SPEC_VERSION_LAZY) ||
+                    ((SPEC_VERSION_LAZY.equals(specVersion) && SPEC_VERSION_LAZY.equals(other.specVersion)) ||
                     Utilities.compareObjects(getSpecificationVersion(), other.getSpecificationVersion())) &&
                     (hasImplementationDependency() == other.hasImplementationDependency()) &&
                     (hasCompileDependency() == other.hasCompileDependency());
--- a/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ImportantFilesNodeFactory.java
+++ b/apisupport/maven.apisupport/src/org/netbeans/modules/maven/apisupport/ImportantFilesNodeFactory.java
@@ -103,7 +103,7 @@ public class ImportantFilesNodeFactory implements NodeFactory {
 
         @Override
         public Node node(String key) {
-            assert key == IMPORTANT_FILES_NAME;
+            assert IMPORTANT_FILES_NAME.equals(key);
             return new ImportantFilesNode(project);
         }
 
--- a/contrib/j2ee.jboss4/src/org/netbeans/modules/j2ee/jboss4/customizer/CustomizerSupport.java
+++ b/contrib/j2ee.jboss4/src/org/netbeans/modules/j2ee/jboss4/customizer/CustomizerSupport.java
@@ -446,11 +446,11 @@ public final class CustomizerSupport {
             c.weighty = 1.0;
             ((GridBagLayout)this.getLayout()).setConstraints(spane,c);
             add(spane);
-            if (type == SOURCES || type == JAVADOC) {
+            if (SOURCES.equals(type) || JAVADOC.equals(type)) {
                 this.addButton = new JButton ();
                 String text;
                 char mne;
-                if (type == SOURCES) {
+                if (SOURCES.equals(type)) {
                     text = NbBundle.getMessage(CustomizerSupport.class, "CTL_Add");
                     mne = NbBundle.getMessage(CustomizerSupport.class, "MNE_Add").charAt(0);
                     ad = NbBundle.getMessage(CustomizerSupport.class, "AD_Add");
@@ -709,7 +709,7 @@ public final class CustomizerSupport {
         }
 
         private void selectionChanged () {
-            if (this.type == CLASSPATH) {
+            if (CLASSPATH.equals(this.type)) {
                 return;
             }
             int indices[] = this.resources.getSelectedIndices();
--- a/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/ui/LibrariesNodeFactory.java
+++ b/enterprise/j2ee.clientproject/src/org/netbeans/modules/j2ee/clientproject/ui/LibrariesNodeFactory.java
@@ -118,7 +118,7 @@ public final class LibrariesNodeFactory implements NodeFactory {
         }
 
         public Node node(String key) {
-            if (key == LIBRARIES) {
+            if (LIBRARIES.equals(key)) {
                 //Libraries Node
                 return  
                     new LibrariesNode(
@@ -141,7 +141,7 @@ public final class LibrariesNodeFactory implements NodeFactory {
                         cs,
                         new ExtraLibrariesNode(project, evaluator, AppClientProjectProperties.J2EE_SERVER_INSTANCE, cs)
                     );
-            } else if (key == TEST_LIBRARIES) {
+            } else if (TEST_LIBRARIES.equals(key)) {
                 return  
                     new LibrariesNode(
                         NbBundle.getMessage(LibrariesNodeFactory.class,"CTL_TestLibrariesNode"),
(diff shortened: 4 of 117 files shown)


BradWalker:
@tbw777,

Ask yourself the following questions:

1 - Did I solve a bug/defect/warning?
2 - Is there a reduction in warning messages?
3 - Is the code broken or just not "perfect" ?
4 - What am I trying to accomplish here?

tbw777:
@BradWalker 

1 - Did I solve a bug/defect/warning?
I think it solve bugs

2 - Is there a reduction in warning messages?
dont know

3 - Is the code broken or just not "perfect" ?
code broken

4 - What am I trying to accomplish here?
fix potentional errors

BradWalker:
> @BradWalker
> 
> 1 - Did I solve a bug/defect/warning? I think it solve bugs
> 
> 2 - Is there a reduction in warning messages? dont know
> 
> 3 - Is the code broken or just not "perfect" ? code broken
> 
> 4 - What am I trying to accomplish here? fix potentional errors

== – checks reference equality

equals() – checks the value equality

In this particular case, it doesn't really matter as they evaluate to the same answer.

tbw777:
There is no guarantee that the given primitive located in the pool, so they can be in different objects and == will return false

BradWalker:
Strings are "interned" as per [JLS 3.10.5](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.5)

So they evaluate to the same thing. Though generally, I would agree, it's better form to use the equals() method.

Again, does this accomplish something or is this "code beautification"?

tbw777:
@BradWalker
As far as I understand, [JLS 3.10.5](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.5) are talking about constants, not variables, and in this PR the changes concern not only of the strings

lkishalmi:
Well, I'm sorry this one is noise.

Many of the changes are touching ancient code written probably for Java 1.4 or earlier. Checking reference equality when the reference is a public constant was usual as these things were treated like enums and maybe for performance reasons,

The `Objects` class is available from Java 7. Replacing those would improve the readability, I see value in that. I'd create a separate PR with those, maybe that would not touch so many files.

Replacing the `==`-s in this one, I fear holds no value.

tbw777:
@lkishalmi 
This changes not concern to intern comparing, threfore they are required to used

vieiro:
Some questions:

> Fixed erroneous direct form of comparison ("==") for these object types: String, Integer, Boolean, BigDecimal

1. Of all the changes you're proposing, which of them do you think are really erroneous?
2. Let's assume NetBeans is working on erroneous code, as you say. Then if we apply all these changes to all these modules, NetBeans will work differently, right? Of all these modules you propose to change, where exactly is going NetBeans to work differently?

> There is no guarantee that the given type located in the pool, so they can be in different objects and == will return false while equals will return true

That is a good hint to evaluate the PR, and may help you answer point 2.  In which of the modules you propose to change is the type located in the pool? In the enterprise module? Web module? apisupport module? xml module? Java module? ant debugger module? Any other modules?

We have to evaluate the impact of changes in the behavior of NetBeans! What is your opinion?

Thanks,
Antonio

tbw777:
I mean data types pool.

If you run this code, for example:

>         String s1 = "test";
>         String s2 = new String("test");
> 
>         System.out.println(s1 == s2);
>         System.out.println(s1.equals(s2));

A result will:

> false
> true

vieiro:
> @vieiro I mean data types pool. Pool of JVM, not NB
> 
> If you run this code, for example:
> 
> > ```
> >     String s1 = "test";
> >     String s2 = new String("test");
> > 
> >     System.out.println(s1 == s2);
> >     System.out.println(s1.equals(s2));
> > ```
> 
> A result will:
> 
> > false
> > true
> 
> If somebody want to do pool comparing he must use the intern() method for Strings.

Exactly. The question is, where is this being done in all the modules you're changing, if any?

tbw777:
@vieiro 
This is done in all the modules that could be found. Most likely the coverage of this change is 95-100%. I have not yet found another way to detect such inconsistencies.

vieiro:
@tbw777 So all of these are bugs?

Let me ask the question again:

> Let's assume NetBeans is working on erroneous code, as you say. Then if we apply all these changes to all these modules, NetBeans will work differently, right? Of all these modules you propose to change, where exactly is going NetBeans to work differently?

tbw777:
@vieiro 
Yes, all these dirs are contains cmp bugs.
The main logic of code is correct, but sometimes may cant work if JVM place created new objects without reusage existing equals objects. Its depending on JVM version and JVM vendor.

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
```diff
@@ -146,7 +146,7 @@ public int compareTo(ModuleDependency other) {
         if (result != 0) { return result; }
         
         // do not force resolution of SPEC_VERSION_LAZY for comparisons, unnecessary
-        if (specVersion != SPEC_VERSION_LAZY || other.specVersion != SPEC_VERSION_LAZY) {
+        if (!SPEC_VERSION_LAZY.equals(specVersion) || !SPEC_VERSION_LAZY.equals(other.specVersion)) {
```
vieiro, line 149: Not a bug. SPEC_VERSION_LAZY is a class internal String that is used in one of the constructors. The comparison with `==` will detect if the class was created with this constructor.

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
```diff
@@ -167,7 +167,7 @@ public boolean equals(Object o) {
             ModuleDependency other = (ModuleDependency) o;
             return getCodeNameBase().equals(other.getCodeNameBase()) &&
                     Utilities.compareObjects(getReleaseVersion(), other.getReleaseVersion()) &&
-                    ((specVersion == SPEC_VERSION_LAZY && other.specVersion == SPEC_VERSION_LAZY) ||
+                    ((SPEC_VERSION_LAZY.equals(specVersion) && SPEC_VERSION_LAZY.equals(other.specVersion)) ||
```
vieiro, line 170: Same here.

vieiro (changes requested):
Comparison with `==` is indeed the way to go in this simple case. Comparing with `equals` will have different, unexpected behaviour.

vieiro:
Hi @tbw777 . There goes the first review of the first proposed change. The change you propose will generate a bug in the `ModuleDependency` class, which may have bad implications for NetBeans.

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
tbw777: specVersion is variable, not a compile time constant, therefore code its correct

apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ModuleDependency.java
tbw777: similar

tbw777:
> Hi @tbw777 . There goes the first review of the first proposed change. The change you propose will generate a bug in the `ModuleDependency` class, which may have bad implications for NetBeans.

What error do you mean? Explain please.

vieiro:
> I don't see an issue in principle with the change, but for each file here we either don't actually have a bug that needs fixing (most likely), or we find an issue where the behaviour does change, in which case we need to verify it's not reliant on the "wrong" behaviour.

I'm worried about these changes in behaviour, that can open a can of worms with bugs difficult to reproduce. I'm not for merging without reviewing the impact of each change.

vieiro:
> > Hi @tbw777 . There goes the first review of the first proposed change. The change you propose will generate a bug in the `ModuleDependency` class, which may have bad implications for NetBeans.
> 
> What error do you mean? Explain please.

It seems you haven't figured out the change in behaviour in NetBeans.

neilcsmith-net:
@vieiro I meant exactly that - not sure I worded it well.  Let's -1 and close this as is then?

A PR for this change, where it's provably a comparison on a String that's possibly not interned, and where that behaviour is wrong, would be good.  But then it also might be empty. :smile:

lkishalmi:
-1

We've already spent too much resources on this one. Closing.

tbw777:
What are you doing?

lkishalmi:
Well,

I made the call. We are having a discussion on code cleanup PRs here: https://lists.apache.org/thread/f7mt7k3rb11sgjj1fo2vlflczvw3oycx
A side discussion also got started here: https://lists.apache.org/thread/k4lhhvxv1jp2oyblgbbn67nghhr63zrr

You claim that this one is a safe bugfix, without naming a bug, that this one would actually fix. @vieiro also mentioned that would cause and error in `ModuleDependency` class. 

Moreover you seem to ignore the raised concerns multiple times. Be humble, do good! Not everything we make has the same goodness value for everyone. I have multiple PRs, that was closed without being merged. The people here, knows, that I can get eager sometimes, rushing for something, then I got pulled back for valid reasons. I'm grateful for that. Sitting back and re-evaluate what are we going for makes me a better programmer, and I hope a better person as well.

As I've said, I see value in some parts of this PR, but as a whole, this won't sail.

tbw777:
@lkishalmi 
@vieiro
@neilcsmith-net 
I cant catch this changes becouse it really fix bugs!
How much money do i need to donate (in crypto for example) to finish this PR correctly?
May be i need to make some tests?
Closing this PR is a mistake!

neilcsmith-net:
> I cant quit this changes becouse it really fix bugs! And I am sure that sometimes they appear in rare cases!

It's only a bug if it's a check against a non-interned String.  Please re-read [JLS-3.10.5](https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.10.5)

There are a few things in here where object identity rather than equality might be important.

Try concentrating on cases where object identity isn't needed and the check is a non-interned String - that's a real bug, _if you find one_.

Some of the other changes might be worth considering if identity is not important, if only to stop having this conversation in future rather than because there's a bug.  That probably does need splitting down into multiple small PRs, because it's not code cleanup as you say.

Offering money is pointless (and insulting)!  Offer your time, address the concerns, and realise we're all trying to make NetBeans better.  Locking this conversation.

Issue #4587, opened 2022-09-06
Unit test on single file is converted to integration test


mlvtito:
### Apache NetBeans version

Apache NetBeans 14

### What happened

I noticed that my Netesta plugin does not work anymore with Netbeans 14.
Each time it tries to apply `ActionProvider.COMMAND_TEST_SINGLE`, the action is replaced by the new `COMMAND_INTEGRATION_TEST_SINGLE`.

### How to reproduce

I use Netesta plugin to automatically trigger unit tests on a single file

### Did this work correctly in an earlier version?

Apache NetBeans 13

### Operating System

Ubuntu

### JDK

1.8, 11, 17 & 18

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

I have done some debugging, so I can tell where the problem occurs. 

In the class `DefaultReplaceTokenProvider`, in the method `isIntegrationTestTarget`, the variable `targetFiles` can be also empty (and not only `null`)

In this case, the `allMatch` method of the stream return `true` whereas there are no files.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

neilcsmith-net:
Looks like it should perhaps be using [extractFileObjectsfromLookup](https://github.com/apache/netbeans/blob/master/java/maven/src/org/netbeans/modules/maven/execute/DefaultReplaceTokenProvider.java#L86), although the extra check for empty would be good.  Feel free to file a PR.

Also, looks like your plugin should really be fixed to pass FileObject.

hemantsoni42:
@neilcsmith-net  can you please assign me this issue.
I already solve fork this and solved the issue.

hemantsoni42:
Fixed Unit test on single file is converted to integration test #4587 #5381

hemantsoni42:
@neilcsmith-net thank you sir for your help
I once again go through the code and file a different PR after adding an extra check.

neilcsmith-net:
This should be fixed in 18-rc3. Please test when that is available next week.

PR #5383, opened 2023-01-29
Fixed Unit test on single file is converted to integration test #4587

hemantsoni42:
adding an extra check in extractFileObjectsFromLookup method to making sure the result will not empty.


neilcsmith-net:
This isn't the issue and fix discussed on #4587 either.  Does this actually fix the issue?  Because part of the report covers the fact that `extractFileObjectsFromLookup` isn't used in that path.

As far as I remember, it's `isIntegrationTestTarget` that needs fixing to get the values from `extractFileObjectsFromLookup`, and always return false if the result is empty.

Issue #4054, opened 2022-04-29
Netbeans freezes due to "Organize Imports" on save and "Format" Tool if code contains specific syntax errors


Minekonst:
### Apache NetBeans version

Apache NetBeans 13, 15

### What happened

The "Format" Tool and "Organize Import" seem crash netbeans sometimes. This might be related to the code containg specific syntax errors, because I can reproduce the bug when inserting C++ code. This bug happend to me before but not as often in Netbeans 13 as in 12.5

### How to reproduce

This bug happens every once in a while to me, but I found a way to reproduce it for me:

I ported the code from https://github.com/Gregjksmith/Iterative-Closest-Point/blob/master/src/KdTree.cpp for a project to java and tried to save this file https://pastebin.com/NXWnF1Hu with "Organize Import" as on save action enabled. Netbeans did not respond anymore and did not save the file.
This specific code did not crash trying to use "Format" but others in th past did.

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10.0

### JDK

Oracle JDK 17 (17+35-LTS-2724)

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

albilu:
Hello, facing the same issue since a while. Its rely annoying. 
Seems the same reported here with logs: #4515

mbien:
can be reproduced with a short snippet:
```java
package main;

public class Test {
    void gs::KdTree::build(std::vector<Point*> &pointCloud, int start, int end, int sortOn) {
        if (end - start == 1) {
            __node = new Point(pointCloud[start]->pos[0], pointCloud[start]->pos[1], pointCloud[start]->pos[2]);
            return;
        }
    }
}
```
paste this into a java file and save while having "Organize Imports" on save enabled.

it will then enter an endless while-loop in (set breakpoint to that line to safely reproduce):

```
	at org.netbeans.modules.java.hints.spiimpl.batch.BatchSearch.getLocalVerifiedSpans(BatchSearch.java:233)
	at org.netbeans.modules.java.hints.spiimpl.batch.BatchSearch.access$400(BatchSearch.java:76)
	at org.netbeans.modules.java.hints.spiimpl.batch.BatchSearch$LocalIndexEnquirer.validateResource(BatchSearch.java:550)
	at org.netbeans.modules.java.hints.spiimpl.batch.BatchSearch.getVerifiedSpans(BatchSearch.java:183)
[...]

PR #5384, opened 2023-01-29
[NETBEANS-4054] Ensuring progress when javac crashes while batch evaluating hints.

jlahoda:
If hints are run on save, `BatchSearch` is called on a single file. Normally, if javac would crash during `toPhase`, an exception would be thrown. But, it may happen the existing parser is reused, and `toPhase` does not throw the exception, and this may be mistaken for an out-of-memory situation, to which the response is to restart the processing. This may then lead to an infinite loop.

The proposed change here is to force progress in this situation. javac crash(es) need to be handled separately, of course.

---

### DIFF
--- a/java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/batch/BatchSearch.java
+++ b/java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/batch/BatchSearch.java
@@ -234,8 +234,9 @@ public class BatchSearch {
                         if (cancel.get())
                             return;
                         final AtomicBoolean stop = new AtomicBoolean();
+                        List<FileObject> currentInputList = toProcess.subList(currentPointer.get(), toProcess.size());
 //                        JavaSource js = JavaSource.create(e.getKey(), f);
-                        JavaSource js = JavaSource.create(e.getKey(), toProcess.subList(currentPointer.get(), toProcess.size()));
+                        JavaSource js = JavaSource.create(e.getKey(), currentInputList);
 
                         js.runUserActionTask(new Task<CompilationController>() {
                             @Override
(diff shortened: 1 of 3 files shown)


java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/batch/BatchSearch.java
```diff
@@ -246,8 +247,15 @@ public void run(CompilationController parameter) throws Exception {
...
-                                    if (parameter.toPhase(Phase.RESOLVED).compareTo(Phase.RESOLVED) < 0)
+                                    if (parameter.toPhase(Phase.RESOLVED).compareTo(Phase.RESOLVED) < 0) {
+                                        if (currentInputList.size() == 1) {
+                                            //the javac crashed while processing the (single) file, we must ensure progress, otherwise infinite loop in processing would happen:
+                                            problems.add(new MessageImpl(MessageKind.WARNING, "An error occurred while processing file: " + FileUtil.getFileDisplayName(parameter.getFileObject()) + ", please see the IDE log for more information."));
+                                            currentPointer.incrementAndGet();
+                                        }
+
                                         return ;
+                                    }
```
mbien, line 258: this means that this can't happen when `BatchSearch` is used on multiple files because it wouldn't reuse javac instances there?

So the only reason it tries again with the same file when `Phase.RESOLVED` can't be reached is because of a possible OOM situation?

If that is the case we should add some kind of per-file retry counter. `while(true)` loops are always a bit of a danger zone.

mbien (approved):
looks good to me, this is actually somewhat similar to the hotfix I proposed [here](https://github.com/apache/netbeans/issues/4054#issuecomment-1263155197) it just limits it to single-file use only.

java/spi.java.hints/src/org/netbeans/modules/java/hints/spiimpl/batch/BatchSearch.java
mbien: sorry never mind this, I just noticed Jan posted a long explanation on the [issue](https://github.com/apache/netbeans/issues/4054#issuecomment-1407651558)

PR #5367, opened 2023-01-26
Sync delivery to master after 17-rc3

neilcsmith-net:

PR #5353, opened 2023-01-23
Use NotifyDescriptor.ComposedInput for Add ADB command

jhorvath:
Use NotifyDescriptor.ComposedInput for Add ADB command. It improves usability of multi step input.

PR #5387, opened 2023-01-30
Resolving conflicts between master, delivery and vsnetbeans_1603

sdedic:
This PR aims to solve conflict present in #5367, brings necessary changes from `master` to delivery first, allows `delivery` to be consistent with `vsnetbeans_1603` where the two conflicting commits were already merged and hopefully tested.

### DIFF
--- a/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
+++ b/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/DownloadWalletAction.java
@@ -62,8 +62,8 @@ import org.openide.util.NbBundle;
         category = "Tools",
         id = "org.netbeans.modules.cloud.oracle.actions.DownloadWalletAction"
 )
-@ActionRegistration( 
-        displayName = "#CTL_DownloadWalletAction", 
+@ActionRegistration(
+        displayName = "#CTL_DownloadWalletAction",
         asynchronous = true,
         lazy = true
 )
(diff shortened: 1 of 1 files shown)


MartinBalin (approved):

neilcsmith-net:
This doesn't seem the right way to handle this IMO.  Either we bring all of #5353 into delivery for NB17, or we handle the conflict in a PR to master and it never lands in delivery?  Do the changes in this one file really need to be in NB17 by themselves?  And if we don't bring in everything in VSCode 16.0.301, won't you get a regression in VSCode 17?

eg. one we've synced delivery to release170 and built rc3

```
git checkout -b master-sync17rc3 master
git merge delivery
# handle conflicts
# open new PR to master
```

neilcsmith-net:
> OK, the issue is that both changes of #5353 (now present in master and vsnetbeans_1603) AND #5363 (now present in delivery and vsnetbeans_1603) are needed IMHO in NB17 (= delivery). The correct file's content is the one in vsnetbeans_1603 branch.

Yes, but there are two more files changed in #5353 ?  Don't you need the new versions of `AddADBAction` and `DownloadWalletDialog` in NB17 as well?

> should that new PR replace #5367 ?

Yes.  But it can't be done until after we merge #5366 , close delivery for the rc, and build / announce 17-rc3.  As long as delivery has the right content before then, I can handle the sync side.  And also as long as no-one touches those 3 files in master this week!

neilcsmith-net:
Cherry pick the commits from #5353 on top of delivery, handling the conflicts, and open a PR.  I'll merge that to delivery.

As long as I know the file contents in delivery are what needs to be in master, that's easy enough to handle when we're ready to sync.

sdedic:
Closed in favour of #5389  -- @neilcsmith-net  thanks for the guidance.

PR #5390, opened 2023-01-30
Fix POM Graph colors for dark themes

johntor:
I noticed a problem using the Graph of a POM when a dark theme is selected.

The foreground color was requested from the Theme while backgrounds are hardcoded into the Class causing printing WHITE on Bright Colors.

 With this fix I added transparency to these hardcoded colors so the become darker on darker backgrounds.
 
 I also removed gradients that looked a bit outdated to my eyes.

pom_graph


neilcsmith-net:
Thanks for the fix!  Can you make your commit on top of the delivery branch instead of master, and force push an update?  Or open another PR if easier.  This is a potential fix to squeeze in the next release, which forked off a couple of weeks back.

johntor:
Moved to the delivery branch

PR #5391, opened 2023-01-30
Fix POM Graph colors for dark themes

johntor:
I noticed a problem using the Graph of a POM when a dark theme is selected.

The foreground color was requested from the Theme while backgrounds are hardcoded into the Class causing printing WHITE on Bright Colors.

With this fix I added transparency to these hardcoded colors so the become darker on darker backgrounds.

I also removed gradients that looked a bit outdated to my eyes.

215498346-c8cecdd1-e0a5-4354-9998-9b0bc4ad6ab9

### DIFF
--- a/java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
+++ b/java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
@@ -20,7 +20,6 @@ package org.netbeans.modules.java.graph;
 
 import java.awt.Color;
 import java.awt.Font;
-import java.awt.GradientPaint;
 import java.awt.Graphics2D;
 import java.awt.Point;
 import java.awt.Rectangle;
@@ -43,6 +42,8 @@ import org.netbeans.api.visual.widget.LabelWidget;
 import org.netbeans.api.visual.widget.LevelOfDetailsWidget;
 import org.netbeans.api.visual.widget.Scene;
 import org.netbeans.api.visual.widget.Widget;
+import org.netbeans.api.visual.widget.general.IconNodeWidget;
+import org.netbeans.api.visual.widget.general.IconNodeWidget.TextOrientation;
 import static org.netbeans.modules.java.graph.Bundle.ACT_FixVersionConflict;
 import static org.netbeans.modules.java.graph.Bundle.TIP_MultipleConflict;
 import static org.netbeans.modules.java.graph.Bundle.TIP_MultipleWarning;
@@ -50,8 +51,6 @@ import static org.netbeans.modules.java.graph.Bundle.TIP_SingleConflict;
 import static org.netbeans.modules.java.graph.Bundle.TIP_SingleWarning;
 import static org.netbeans.modules.java.graph.DependencyGraphScene.VersionProvider.VERSION_CONFLICT;
 import static org.netbeans.modules.java.graph.DependencyGraphScene.VersionProvider.VERSION_NO_CONFLICT;
-import org.netbeans.api.visual.widget.general.IconNodeWidget;
-import org.netbeans.api.visual.widget.general.IconNodeWidget.TextOrientation;
 import org.openide.util.ImageUtilities;
 import org.openide.util.NbBundle.Messages;
 import org.openide.util.Parameters;
@@ -67,8 +66,8 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
     static final Color DIRECTS_CONFLICT = new Color(235, 88, 194);
     static final Color DISABLE_HIGHTLIGHT = new Color(255, 255, 194);
     static final Color HIGHTLIGHT = new Color(255, 255, 129);
-    static final Color DISABLE_CONFLICT = new Color(219, 155, 153);
     static final Color CONFLICT = new Color(219, 11, 5);
+    static final Color DISABLE_CONFLICT = EdgeWidget.deriveColor(CONFLICT, 0.7f);
     static final Color MANAGED = new Color(30, 255, 150);
     static final Color WARNING = new Color(255, 150, 20);
     static final Color DISABLE_WARNING = EdgeWidget.deriveColor(WARNING, 0.7f);
@@ -85,6 +84,7 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
 
     private GraphNode<I> node;
     private boolean readable = false;
+    private boolean isHighlighted = false;
     private boolean enlargedFromHover = false;
 
     private Timer hoverTimer;
@@ -130,7 +130,7 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
             }
         }) : null;
         this.sceneHoverActionAction = sceneHoverActionAction;
-        
+
         setLayout(LayoutFactory.createVerticalFlowLayout());
 
         updateTooltip();
@@ -144,7 +144,7 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
             hoverBorderC = Color.GRAY;
         }
     }
-    
+
     @Messages({"TIP_Text=<html>{0}<br>{1}</html>",
                "TIP_SingleConflict=Conflict with <b>{0}</b> version required by <b>{1}</b>",
                "TIP_SingleWarning=Warning, older version <b>{0}</b> requested by <b>{1}</b>",
@@ -206,14 +206,15 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
 
     public void highlightText(String searchTerm) {
         if (searchTerm != null && node.getName().contains(searchTerm)) {
-            nodeW.setBackground(HIGHTLIGHT);
+            isHighlighted = true;
             nodeW.setOpaque(true);
             setPaintState(EdgeWidget.REGULAR);
             setReadable(true);
         } else {
             //reset
-            nodeW.setBackground(Color.WHITE);
+            isHighlighted = false;
             nodeW.setOpaque(false);
+            nodeW.getLabelWidget().setForeground(origForeground);
             setPaintState(EdgeWidget.GRAYED);
             setReadable(false);
         }
@@ -235,11 +236,11 @@ class NodeWidget<I extends GraphNodeImplementation> extends Widget implements Ac
     void setFixed(boolean fixed) {
         this.fixed = fixed;
     }
-    
+
     boolean isFixed() {
         return fixed;
     }
-    
+
     void updatePaintContent() {
         if (origForeground == null) {
             origForeground = getForeground();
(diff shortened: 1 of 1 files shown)


mbien:
hi @johntor, thanks for working on this.

it certainly does look much better and I agree the gradients have to go given that NetBeans is using a theme called **Flat** LAF  by default :)

However, I still see problems with FlatLAF dark, what theme are you using since your screenshorts show white font on dark background.

(FlatLaf dark, items in graph don't have focus)
before PR:
before-unfocussed


after PR:
after-unfocussed

mbien:
the scope markers are now also painted as solid blocks which is going to need some tweaks too:
scope-after
(Cupertino Dark, scopes set to Test)

johntor:
I did not notice all these! If you like the idea of the removed gradients we can optimize the code more.

mbien:
@johntor would be great. Simply force push into this PR if you want to make changes.

I believe the font color might need to be hard coded if the custom (non-standard) background fill is active - otherwise there gonna be always some theme which will clash.

johntor:
Yes I was thinking determining if the theme is light/dark then apply fixed black/white colors for fonts

mbien:
> Yes I was thinking determining if the theme is light/dark then apply fixed black/white colors for fonts

if the background is taken from the theme, its ok to take the font color from the theme too. But as soon the background is changed, for example when set to orange, the color can't be taken from the theme anymore - it has to fit to the orange background.

I wouldn't hardcode all font colors, only conditional when the background actually changes.

neilcsmith-net:
Yes, two hard coded colours for each pallette option based on value of `nb.dark.theme` boolean from UIManager might be way to go, at least for a quick fix.

mbien:
> Yes, two hard coded colours for each pallette option based on value of `nb.dark.theme` boolean from UIManager might be way to go, at least for a quick fix.

i don't think any of this needs to know if the dark theme is active or not if it follows the simple rule of changing both, foreground and background at the same time but never only one of them while leaving the other to the theme.

johntor:
> > Yes, two hard coded colours for each pallette option based on value of `nb.dark.theme` boolean from UIManager might be way to go, at least for a quick fix.
> 
> i don't think any of this needs to know if the dark theme is active or not if it follows the simple rule of changing both, foreground and background at the same time but never only one of them while leaving the other to the theme.

Then we have problems to apply your light-grey text over orange or yellow. I will go with the hardcoded solution and we see how it goes...

mbien:
> Then we have problems to apply your light-grey text over orange or yellow.

why would you do that? I am trying to say that if you change the background to orange, you have to change the text to a color which fits to orange. You can't leave the text color up to the theme.

johntor:
Hello again I think is fine now but how am I supposed to update the file and how I force my pull request? 

Image

Image (1)

Image (3)

mbien:
> Hello again I think is fine now but how am I supposed to update the file and how I force my pull request?

you have now two commits, I think the easiest option in this particular case is to remove the last commit again with:

```
git reset HEAD^
```
This will only remove the commit, the changes you made will remain in the source code.

Now you can simply commit with NetBeans again but check the `Amend last Commit` checkbox, this will add the changes to the first commit, instead of creating a second one again - basically updating the commit.

To replace the remote `delivery` branch with your local `delivery` branch you type in a console:
```
git push -f git@github.com:johntor/netbeans delivery:delivery
```

neilcsmith-net:
> i don't think any of this needs to know if the dark theme is active or not if it follows the simple rule of changing both, foreground and background at the same time but never only one of them while leaving the other to the theme.

You create a palette of background colours that works with black-ish text, and one that works with white-ish text.  Working on something similar with the visual library myself at the moment.

However, this fix looks good for now.  If we can get it amended and force pushed as above, we can merge later today for 17-rc3.

mbien:
> You create a palette of background colours that works with black-ish text, and one that works with white-ish text. Working on something similar with the visual library myself at the moment.

If the background would be theme aware - the foreground must be theme aware too, I agree.

But in this particular case the background fill colors (red/orange) already work with dark/light themes, additionally to that the boxes have an outline which is controlled by the theme - which makes it extra resistant to clash with theme colors. So the main problem here was that some boxes set the background fill without setting the font color. That is the reason I said I do not think the logic has to be aware of the theme at all - it only has to follow the rule to not pick the foreground from a theme color while hard coding the background. 

The visual for "Scopes" will need more work since it used to be a gradient but this is outside the scope of this PR :) (I don't think it will look good in any color the way it is currently rendered as simple box in a box)

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -78,13 +77,18 @@
...
-    private static final @StaticResource String BULB_ICON = "org/netbeans/modules/java/graph/resources/bulb.gif";
-    private static final @StaticResource String BULB_HIGHLIGHT_ICON = "org/netbeans/modules/java/graph/resources/bulb-highlight.gif";
+    private static final @StaticResource
+    String LOCK_ICON = "org/netbeans/modules/java/graph/resources/lock.png";
+    private static final @StaticResource
+    String LOCK_BROKEN_ICON = "org/netbeans/modules/java/graph/resources/lock-broken.png";
+    private static final @StaticResource
+    String BULB_ICON = "org/netbeans/modules/java/graph/resources/bulb.gif";
+    private static final @StaticResource
+    String BULB_HIGHLIGHT_ICON = "org/netbeans/modules/java/graph/resources/bulb-highlight.gif";
```
eirikbakke, line 87: These formatting changes can be left out of the patch.

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -102,25 +106,27 @@
...
-    
-    NodeWidget(@NonNull DependencyGraphScene scene, GraphNode<I> node, final Action fixConflictAction, final WidgetAction sceneHoverActionAction) {        
+    public double locX, locY, dispX, dispY;
+    private boolean fixed;
+
+    NodeWidget(@NonNull DependencyGraphScene scene, GraphNode<I> node, final Action fixConflictAction, final WidgetAction sceneHoverActionAction) {
         super(scene);
         Parameters.notNull("node", node);
-        if(fixConflictAction != null) {
+        if (fixConflictAction != null) {
```
eirikbakke, line 118: Nit: Lots of whitespace changes in this commit makes it hard to review the real changes. It's a good idea to put whitespace changes in a separate commit (though no need to change it now).

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -178,48 +184,52 @@ public String getConflictTooltip(GraphNode<I> node) {
...
-            for (I nd : node.getDuplicatesOrConflicts()) {
-                if (scene.isConflict(nd)) {
-                    tooltip.append("<tr><td>");
-                    tooltip.append(scene.getVersion(nd));
-                    tooltip.append("</td>");
-                    tooltip.append("<td>");
-                    GraphNodeImplementation parent = nd.getParent();
-                    if (parent != null) {
+        } else {
+            if (conflictCount > 1) {
```
eirikbakke, line 188: Why did the "else if" become a separate "if" here?

johntor:
I think current solution is simple and efficient for any color theme. Keep in mind the new feature of changing the colors of Flat Laf. 

This solution is adaptioning theme colors while keeping original colors with BLACK foreground to boxes with background yellow orange and red.

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -254,16 +264,34 @@ void updatePaintContent() {
...
-        contentW.setBorder(BorderFactory.createLineBorder(10, foreC));
         nodeW.setForeground(foreC);
-        if(versionW != null) {
+        DependencyGraphScene scene = getDependencyGraphScene();
+        int conflictType = scene.supportsVersions() ? node.getConflictType(scene::isConflict, scene::compareVersions) : VERSION_NO_CONFLICT;
+        if (isHighlighted) {
+            nodeW.getLabelWidget().setForeground(Color.BLACK);
+        } else {
+            nodeW.getLabelWidget().setForeground(origForeground);
+//      if (paintState != EdgeWidget.GRAYED) {
```
eirikbakke, line 277: Commented code here should be removed.

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -484,50 +511,55 @@ protected void notifyStateChanged(ObjectState previousState, ObjectState state)
...
         if (previousState.isSelected() != state.isSelected()) {
             updateNeeded = true;
         }
 
         if (updateNeeded) {
             updateContent();
-        } else if (repaintNeeded) {
-            repaint();
+        } else {
+            if (repaintNeeded) {
```
eirikbakke, line 522: Another case of an "else if" becoming a nested if...

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
```diff
@@ -484,50 +511,55 @@ protected void notifyStateChanged(ObjectState previousState, ObjectState state)
...
-    public GraphNode getNode () {
+    public GraphNode getNode() {
         return node;
     }
 
     /**
-     * readable widgets are calculated  based on scene zoom factor when zoom factor changes, the readable scope should too
+     * readable widgets are calculated based on scene zoom factor when zoom
+     * factor
+     * changes, the readable scope should too
```
eirikbakke, line 554: Unintended formatting change?

mbien (commented):
can you revert the formatting changes you did in the most recent commit We try to avoid unrelated changes in PRs which fix things.

eirikbakke (approved):
Thanks for this patch! I'm fine with this patch as long as it is tested well on both light and dark FlatLAF modes.

A general comment for future PRs is to try to keep the patches free of unnecessary formatting changes; this makes them easier to review. If you'd like to clean up formatting, it's good to do all the formatting changes in a single commit.

(A side note for @neilcsmith-net: aren't "release candidate" versions supposed to be very stable, with only critical bugfixes going into them? Patches like these benefit from being tested for a few weeks in dev environments before going into production.)

mbien:
> (A side note for @neilcsmith-net: aren't "release candidate" versions supposed to be very stable, with only critical bugfixes going into them? Patches like these benefit from being tested for a few weeks in dev environments before going into production.)

@eirikbakke well the colors and gradients were borderline unreadable before, so the bar is lower here for a hotfix. This is only a UI change which removes gradients and tweaks colors so it has low risk to cause trouble. I agree with the formatting changes which were added by the second commit - those should be reverted.

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
johntor: Fixed

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
johntor: My Netbeans autoformatter has different setting than yours!

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
johntor: My Netbeans autoformatter has different setting than yours!

java/java.graph/src/org/netbeans/modules/java/graph/NodeWidget.java
johntor: Fixed

eirikbakke:
@mbien Yeah, I trust your judgment here.

(I just remember having made "cosmetic-only" patches like this myself only to find I introduced a NullPointerException somewhere a week later...)

johntor:
I am sorry that I have no time to completed this Pull Request till the end right now...  (busy at work)
Can anybody else take care of it?
I promise next time I will complete the work all the way...
Thank you all!
Keep on your great job!

mbien:
@johntor no worries. I can take over. Will remove the formatting changes and force push into this PR. Thanks for your work!

This might add me as co author to the commit. Lets see how this works.

mbien:
squashed, reverted formatting and force pushed

mbien:
@eirikbakke @neilcsmith-net I am willing to accept this one as my side quest, if it causes followup problems assign issues to me.

mbien (approved):

neilcsmith-net:
Thanks @johntor and @mbien Will merge tomorrow.

@mbien given my comments on #5390 asking for it to be rebased, we can share any side quest 😆

mbien:
@johntor congrats on your first contribution.

johntor:
> @johntor congrats on your first contribution.

Thank you very much! I thought contributing to a complicated project like Netbeans is too difficult... I realized that I was wrong!

I will be back!

Take care!

PR #5397, opened 2023-01-31
chang to .yaml as per official YAML

featuriz:
As per YAML documentation file extension is .yaml not .yml 'The official recommended filename extension for YAML files has been .yaml since 2006'


featuriz:
Update as per YAML documentation

matthiasblaesing:
Needs to be updated with the additional change from #5397 and should also document how the change was checked.

While doing the update please also check that the author information is complete. You can check it by running `git log -n 1`  from the branch that holds the corresponding change. In this case the author email ends with `users.noreply.github.com`, which is obviously not correct.

Please also check the correct spelling of the commit message. I suggest:

> Change extension of YAML template file to .yaml as per official YAML specification

featuriz:
> Needs to be updated with the additional change from #5397 and should also document how the change was checked.
> 
> While doing the update please also check that the author information is complete. You can check it by running `git log -n 1` from the branch that holds the corresponding change. In this case the author email ends with `users.noreply.github.com`, which is obviously not correct.
> 
> Please also check the correct spelling of the commit message. I suggest:
> 
> > Change extension of YAML template file to .yaml as per official YAML specification

@matthiasblaesing I point out there is a problem. You are finding spelling mistake in my wording. Good work in finding spelling mistake. 


RIP Netbeans.

matthiasblaesing:
@featuriz not sure what is your problem. I reviewed your code and it needs to be fixed (as pointed out in the first sentence). Then I checked the author information. It is required to be correct and it was not. When both of these were done I _also_ mentioned, that you should update the commit message. I even suggested a replacement.

If you feel offended by that sorry, but it was not meant so.

You are welcome to reopen this PR with the required changes.

PR #5398, opened 2023-01-31
yml to yaml

featuriz:
Correct extension for YAML is .yaml not .yml
'The official recommended filename extension for YAML files has been .yaml since 2006'


matthiasblaesing:
This needs to be grouped with #5397 please join this change with that commit and force push there. Please also describe how you verified, that the change worked correctly. Thank you.

featuriz:
Okay

PR #5399, opened 2023-01-31
Cleanup by simplifying strings comparsion with JDK7 Objects.equals()

tbw777:
Objects.equals() implements normalized comparsion operations for two objects.

Different forms of comparison are reduced to the normalized.
No any changes for references check are made.


matthiasblaesing (approved):
Looks sane to me.

PR #5400, opened 2023-02-01
Cleanup by removing do nothing replaceAll

tbw777:
replaceAll("|", "") is do nothing code example

### DIFF
--- a/extide/gradle/src/org/netbeans/modules/gradle/loaders/ExtensionPropertiesExtractor.java
+++ b/extide/gradle/src/org/netbeans/modules/gradle/loaders/ExtensionPropertiesExtractor.java
@@ -170,7 +170,7 @@ public class ExtensionPropertiesExtractor implements ProjectInfoExtractor {
             if (c == null) {
                 return null;
             }
-            return Arrays.asList(c.split(";;")).stream().map(s -> s.replaceAll("\\;", ";")).collect(Collectors.toList());
+            return Arrays.asList(c.split(";;"));
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


BradWalker:
I NEVER touch tests. By definition, code might be executed that "doesn't make sense" just for the express purpose of executing a test.

Also, unless the test is provably wrong, it's not justification to claim that code does nothing. Even if code does nothing, by definition, it's a test. So the fact that it does nothing, means it is doing something.

Lastly, for example, the test ExtensionPropertiesExtractor.java, which you claim does nothing, is recent code as per the git log.  A better method would have been to reach out to the author of this code and ask if it is really the intent of the test.

Therefore, I'm closing this.

tbw777:
@BradWalker 
I wrote that only the method replaceAll does nothing, not the test

tbw777:
@sdedic Is this change of ExtensionPropertiesExtractor.java correct or you mean something different in test?

neilcsmith-net:
All this should probably be evaluated against whether the intention was do nothing or `replace("|", "");`?

The Gradle code should probably be moved from this PR and checked against what was intended - looks like replacing escaped semicolons?  In which case, both original and new code is wrong?

BradWalker:
> @BradWalker
> 
> > replaceAll("|", "")
> 
> | - here is control regexp symbol means OR. As a result this method nothing changed.

Using "|" as a regex pattern, also known as the alternation, is a common way to "match everything"..

BradWalker:
Reopening for others to code review.. My mistake..

tbw777:
@BradWalker 
You can check first test from here https://gist.github.com/tbw777/c11e8cb9b23d50d472177fda9f67170e

extide/gradle/src/org/netbeans/modules/gradle/loaders/ExtensionPropertiesExtractor.java
```diff
@@ -170,7 +170,7 @@ public Collection<String> keys(BuildPropertiesSupport.Property container) {
             if (c == null) {
                 return null;
             }
-            return Arrays.asList(c.split(";;")).stream().map(s -> s.replaceAll("\\;", ";")).collect(Collectors.toList());
+            return Arrays.asList(c.split(";;"));
```
sdedic, line 173: Thanks for spotting this - actually the replacement should be from "\;" ("escape sequence") to ;, since list delimiter is set to ";;". Will fix in a separate PR.

PR #5401, opened 2023-02-01
Speedup by removing non pattern replaceAll with constant arg

tbw777:
replaceAll("ABC", "") is non Pattern replace


BradWalker:
It's been reopened for others to review. My mistake.

tbw777:
Proof of changes https://gist.github.com/tbw777/2cec406eaafea974c2f702209bd8eb01

asbachb (commented):
Had a look into the changes and the gist. LGTM.

PR #5404, opened 2023-02-01
Sync delivery to master after 17-rc3 (merge conflict resolved)

neilcsmith-net:
Replacement for #5367 with merge conflict on DownloadWalletAction resolved in favour of version in delivery.

cc @sdedic


sdedic (approved):
DownloadWalletAction OK - just to be double sure, cc: @jhorvath , original author of the  wallet code.

neilcsmith-net:
I need to merge this now so that other aspects of the release process can proceed.  @jhorvath please double check master is as expected after merge and flag it up if not.  Thanks both!

PR #5405, opened 2023-02-01
Added missing @deprecated into existing javadocs

tbw777:
Syncing javadoc required


mbien:
* `@deprecated` javadoc tag is there to provide migration guidance for the API user - if the message is empty it serves no purpose.
* `@Deprecated` annotation is sufficient to make it show up as deprecated in the javadoc and produce a compiler warning

This means having an empty javadoc tag serves no purpose, it should be only added to provide a meaningful message.

PR #5413, opened 2023-02-02
Cleanup: removed dead/unused private methods. Part 1

tbw777:
Methods excluded throught commenting, no other types of changing are made.
After was excluded unused methods layer 1 there is next level unused methods was found that also exluded.
This type of changing is simple to review therefore contains a lot of diffs.
Its not automated change!

Profits:
- less IDE indexer load
- resulting version size
- JVM method search speed and memory usage
- less compile time
- improved code navigation (unused parents excluded)
- less overall % of methods without tests coverage
- less Strings pool size of literals

Excuse me for size. Its about 50% coverage.


sdedic:
Could you split the commit somewhat ? For example one commit per cluster ? It could be easier to review.

Please note that while removing dead code is good, we must be more careful in Platform or IDE clusters - it's possible that some other module hacked into the implementation using Reflection. It should not been done in the first place, I agree completely ... but this kind of ... things... happens in such long-lived codebase.

So please let's be careful.

tbw777:
Of course. For example i can split per parent dir.

>For example one commit per cluster ?
What do you mean? Explain please.

>it's possible that some other module hacked into the implementation using Reflection.
You are right, ANT can call private methods also. Nightmare.

You can mark this PR do no merge, i can create splitted versions later.

tbw777:
new PRs in #5420 and other

PR #5414, opened 2023-02-02
Cleanup: diamond operator

tbw777:
Easy to check step by step.

Good to use with #5214
Сan't look at it


mbien:
this is just noise. Lets not start with this kind of project wide stylistic changes.

PR #5417, opened 2023-02-03
harness/o.n.insane - added new tests

lbownik:
harness/o.n.insane - added new tests as the existing ones are just smoke tests which do not verify much
the intention is to remove existing ones in next PR

### DIFF
--- a/harness/o.n.insane/test/unit/src/org/netbeans/insane/FakeProgressModel.java
+++ b/harness/o.n.insane/test/unit/src/org/netbeans/insane/FakeProgressModel.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.insane;
+
+//------------------------------------------------------------------------------
+import java.util.ArrayList;
+import java.util.List;
+import javax.swing.BoundedRangeModel;
+import javax.swing.event.ChangeListener;
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


harness/o.n.insane/test/unit/src/org/netbeans/insane/FakeProgressModel.java
```diff
@@ -0,0 +1,141 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.netbeans.insane;
+
+//------------------------------------------------------------------------------
```
vieiro, line 21: We don't really decorate test cases with these lines, but it's ok.

vieiro (approved):
Looks fine to me. It's surprising we didn't have these tests!

vieiro:
@mbien I'm not sure which labels apply to this PR. I added the `platform` ci label, but maybe we want to run all tests here, right?

mbien:
hi @vieiro, platform should be enough here (harness stuff is hooked to the platform jobs). But your intuition is right, when in doubt -> all tests.

to make CI pick up the labeling change however we have to create a fresh workflow run via a PR sync or the "lock/unlock conversation" button on the right which we use as workaround in situations like here to get a new workflow. Pressing approve will start the *old* workflow run (!).

I sometimes go to https://github.com/apache/netbeans/actions directly and remove the old workflow run first to not cause confusion which also removes the button.

> added new tests as the existing ones are just smoke tests which do not verify much
the intention is to remove existing ones in next PR

hi @lbownik. Please leave the old tests there, you basically wrote tests for a facade while the other tests, although rudimentary on the surface, actually test the implementation.

lbownik:
"although rudimentary on the surface, actually test the implementation." - Isn't it an antipattern for tests? :)

Testing through the façade makes tests Structure-Insensitive .
https://medium.com/@kentbeck_7670/test-desiderata-94150638a4b3

harness/o.n.insane/test/unit/src/org/netbeans/insane/FakeProgressModel.java
vieiro: That's fine, but well, you're contributing to NetBeans, and I'm afraid we don't use those visual aids. Let's keep them as are, no more changes needed.

mbien:
> Testing through the façade makes tests Structure-Insensitive .

your tests test the overloaded methods of LiveReferences which is just a utility exposing the API, the exiting test LiveEngine directly which is much easier to do. Don't remove them please.

lbownik:
> > Testing through the façade makes tests Structure-Insensitive .
> 
> your tests test the overloaded methods of LiveReferences which is just a utility exposing the API, the exiting tests test LiveEngine directly which is much easier to do. Don't remove them please.

OK

PR #5424, opened 2023-02-04
Removed unessesary old forms of Integer type

tbw777:
Merged changes as requested at #5229

omg how big is a changes size


mbien:
you have to split this into multiple commits so that it is reviewable. 1-3 clusters per commit should be fine, depends how large the commits are.

tbw777:
i do this before already: 
#5239

mbien:
no you don't understand. One PR is ok. But split it into multiple COMMITS.

PR #5430, opened 2023-02-04
Cleanup: dead/unused private methods: [ide]

tbw777:
Target directories:
- ide

Methods excluded throught commenting, no other types of changing are made. After was excluded unused methods layer 1 there is next level unused methods was found that also exluded. This type of changing is simple to review therefore contains a lot of diffs. Its not automated change!

Profits:
- less IDE indexer load
- resulting version size
- JVM method search speed and memory usage
- less compile time
- improved code navigation if parent methods will searching (unused parents excluded)
- less overall % of methods without tests coverage
- less Strings pool size of literals

UPD: BUILD SUCCESSFUL

### DIFF
--- a/ide/api.debugger/src/org/netbeans/api/debugger/DebuggerManager.java
+++ b/ide/api.debugger/src/org/netbeans/api/debugger/DebuggerManager.java
@@ -785,18 +785,6 @@ public final class DebuggerManager implements ContextProvider {
         listeners.removeElement (l);
     }
 
-    /**
-     * Removes debugger listener.
-     * Does the same as {@link #removeDebuggerListener(org.netbeans.api.debugger.DebuggerManagerListener)},
-     * but uses the standard naming pattern, so that it can be called by
-     * {@link org.openide.util.WeakListeners}.
-     *
-     * @param l listener object.
-     */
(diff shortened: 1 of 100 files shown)


ide/api.debugger/src/org/netbeans/api/debugger/DebuggerManager.java
```diff
@@ -785,18 +785,6 @@ public void removeDebuggerListener (DebuggerManagerListener l) {
...
 
-    /**
-     * Removes debugger listener.
-     * Does the same as {@link #removeDebuggerListener(org.netbeans.api.debugger.DebuggerManagerListener)},
-     * but uses the standard naming pattern, so that it can be called by
-     * {@link org.openide.util.WeakListeners}.
-     *
-     * @param l listener object.
-     */
-    // DO NOT REMOVE, used by WeakListeners, by reflection!
```
vieiro, line 796: Can you please explain why you're deleting this that says **// DO NOT REMOVE, used by WeakListeners, by reflection!** in capital letters?

vieiro (changes requested):
There's a comment that says do not remove this, but you're removing it anyway. 

Please explain how many of these methods you're removing are still being used via reflection.

Issue #5402, opened 2023-02-01
Maven Dependency Graph fails to change pom.xml to fix conflict


max8github:
### Apache NetBeans version

Apache NetBeans 16

### What happened

I went to a pom file and clicked graph and then show graph.
After a while (and multiple attempts, but will leave that for another bug report) the graph is displayed.
I identify a dependency conflict.
I right-click on it to pull up the dialog to fix it.
I choose the fix and submit it (so that the pom changes with the correct excludes).
I get UnsupportedException.
File pom.xml is not changed.

### How to reproduce

Unsure how i can provide my code to you, as it is proprietary

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Mac OS Ventura 13.2

### JDK

(from sdkman 17.0.5-oracle) java version "17.0.5" 2022-10-18 LTS, Java HotSpot(TM) 64-Bit Server VM (build 17.0.5+9-LTS-191, mixed mode, sharing)

### Apache NetBeans packaging

Apache NetBeans binary zip

### Anything else

always happens

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

max8github:
[messages.log](https://github.com/apache/netbeans/files/10557437/messages.log)
[messages.1.log](https://github.com/apache/netbeans/files/10557442/messages.1.log)

max8github:
![Screenshot 2023-02-01 at 14 31 59](https://user-images.githubusercontent.com/2728740/216057295-4efa2810-c7da-4c63-95a7-447f4ea604a9.png)

max8github:
![Screenshot 2023-02-01 at 14 35 20](https://user-images.githubusercontent.com/2728740/216057772-9ca15dd2-2102-4b4a-b45d-5b369de058e7.png)

matthiasblaesing:
@max8github please have a look at the referenced PR and see if it fixes the issue for you. I did basic testing, but it would be good if you could verify. There is a prebuild binary available via the CI run summary or via this link: https://github.com/apache/netbeans/suites/10781960989/artifacts/541880163

PR #5431, opened 2023-02-04
UnsupportedOperationException in maven dependency graph "Exclude dependency" and "Fix Version Conflict" actions

matthiasblaesing:
The two actions:

org.netbeans.modules.maven.graph.FixVersionConflictAction
org.netbeans.modules.maven.graph.ExcludeDepAction

need to manipulate the duplicate or conflict list. The returned set is read-only, so a companion method to addDuplicateOrConflict removeDuplicateOrConflict is added to remove a node from the list.

Closes: #5402


mbien (approved):
good fix

matthiasblaesing:
@mbien thank you for the review.

Issue #5303, opened 2023-01-17
css feature :has does not recognize pseudo classes


mokraemer:
### Apache NetBeans version

Apache NetBeans 16

### What happened

just add 
body:has(#aCheck:checked) {
  background-color:red;
}

this is marked as syntax error, but this is valid. has can contain any selector, even pseudo classes

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Mageia 8 (Linux)

### JDK

java-11-openjdk-11.0.17.0.8

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
@mokraemer please have a look at the referenced PR and test it. If you need a prebuild binary, it is available from the CI summary or from this direct link: https://github.com/apache/netbeans/suites/10786014570/artifacts/542201388

mokraemer:
downloaded the referenced dev build.
It looks good for me. No more complains about the syntax.

Thanks!

matthiasblaesing:
@mokraemer thanks for the test. I merged a slightly updated version, based on review.

Issue #5025, opened 2022-11-28
[CSS] A warning annotation for nested CSS functions in an attribute


noelma:
### Apache NetBeans version

Apache NetBeans 15

### What happened

When the `var()` function is used in a `calc()` function for the declaration of a variable there is no problem.
But their uses in a CSS attribute, a warning annotation is displayed
![Capture d’écran du 2022-11-28 21-04-50](https://user-images.githubusercontent.com/10517497/204370499-3c5a10d4-1315-4471-b858-8b7f92a9a2fb.png)

### How to reproduce

Insert the following code in a CSS file: 
```css
.foo {
  --widthA: 100px;
  --widthB: calc(var(--widthA) + 2px);
  width: var(--widthB);
}

.bar {
  --widthC: 100px;
  width: calc(var(--widthC) + 2px);
}
```

### Did this work correctly in an earlier version?

Apache NetBeans 15

### Operating System

Linux version 5.15.0-53-generic running on amd64

### JDK

18.0.2-ea; OpenJDK 64-Bit Server VM 18.0.2-ea+9-Ubuntu-222.04; Private Build

### Apache NetBeans packaging

Apache NetBeans Snap Package

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

matthiasblaesing:
@noelma  please have a look at the referenced PR and test it. If you need a prebuild binary, it is available from the CI summary or from this direct link: https://github.com/apache/netbeans/suites/10786014570/artifacts/542201388

PR #5432, opened 2023-02-04
CSS: Support recursive calc definition, improve compatibility for custom properties, support :has pseudo-class

matthiasblaesing:
**Support recursive CSS property definitions (calc function with embedded function calls)**

The grammar for calc property values failed to parse embedded var and
calc calls correctly. The issue is the work-around introduces as
function-content, which basicly prevents nesting any expression that
contains braces.

This changes the grammar parsing result from a tree to a directed graph,
which can contain cycles. This needs to taken into account by useers of
the parsed grammar.

An example of such a cycle is:

```
calc-fn -> calc-sum -> calc-product -> calc-value -> calc-sum
              ^                                        |
              +----------------------------------------+
```

For example this is now supported:

```css
.header {
  height: calc(28 * var(--spacing-unit));
}
```

**Improve compatibility for custom properties**

Custom properties can hold more or less arbitrary content. This was not parsed and is now:

```css
:root {
  --grid-gutter: (var(--margins) * var(--spacing-unit));
  --cell-margin: (var(--grid-gutter) / 2);
}
```

**Support CSS :has pseudo-class**

This was not parsed and is now:

```css
body:has(+ #aCheck:checked) {
    background-color:red;
}
```

Closes: #5025
Closes: #5303


ide/css.lib/src/org/netbeans/modules/css/lib/Css3.g
```diff
@@ -36,6 +36,9 @@
 //    from the module directory (ide/css.lib)
 // 3. Rerun unittests
 // 4. Commit Css3.g together with generated Css3Lexer.java and Css3Parser.java
+//
```
vieiro, line 39: At the top of this file there's a line like this:
```
Author : Jim Idle, Temporal Wave LLC.
```
We may want to add an additional author there (be it Matthias or the NetBeans team or something). After all we don't want Jim Idle to be blamed by our sins... ;-)

matthiasblaesing: I'm inclined to agree, however I quote:

```
// This grammar is free to use providing you retain everyhting in this header comment
// section.
```

so I let it stand as is. Lines 19+ I think spell it out, that this is a modified version.

ide/css.lib/src/org/netbeans/modules/css/lib/api/properties/Properties.java
```diff
@@ -144,24 +146,26 @@ public static boolean isAggregatedProperty(FileObject context, PropertyDefinitio
         final GroupGrammarElement grammarElement = propertyDefinition.getGrammarElement(context);
         final AtomicBoolean isAggregated = new AtomicBoolean(false);
         grammarElement.accept(new GrammarElementVisitor() {
-            private boolean cancelled = false;
+            private final Map<GroupGrammarElement,Void> seen = new IdentityHashMap<>();
```
vieiro, line 149: It's ok. Maybe`Collections.newSetFromMap(new IdentityHashMap<>()) `  makes things clearer below (`seen.contains`, `seen.add`, etc.)?

ide/css.lib/src/org/netbeans/modules/css/lib/properties/GrammarParser.java
```diff
@@ -372,4 +386,59 @@ public CharSequence readText() {
...
+            return truncatedPath(this.parent).equals(truncatedPath(other.parent));
+        }
+
+        @SuppressWarnings("AssignmentToMethodParameter")
+        private String truncatedPath(GrammarElement parent) {
+            StringBuilder path = new StringBuilder();
+            while(parent != null) {
+                path.append(parent.getName());
+                path.append("/");
+                if(parent.getName() != null) {
```
vieiro, line 436: This `if (parent.getName() != null) {` shouldn't it be two lines above? Or do we want `path.append(parent.getName());` even if `parent.getName()` is null?

ide/css.visual/src/org/netbeans/modules/css/visual/RuleEditorNode.java
```diff
@@ -595,15 +596,29 @@ private PropertyValuesEditor createPropertyValueEditor(FileObject context, Prope
             GroupGrammarElement rootElement = pmodel.getGrammarElement(context);
 
             rootElement.accept(new GrammarElementVisitor() {
+                private Map<GroupGrammarElement,Void> seen = new IdentityHashMap<>();
+
```
vieiro, line 600: It's fine, as in comment above we could also use a map for cleaner semantics.

vieiro (approved):
Great job! Thanks, @matthiasblaesing

ide/css.visual/src/org/netbeans/modules/css/visual/RuleEditorNode.java
matthiasblaesing: Addressed via https://github.com/apache/netbeans/pull/5432/commits/e2dd93481e809985df55c138448cbaaf0ab75f4b

ide/css.lib/src/org/netbeans/modules/css/lib/api/properties/Properties.java
matthiasblaesing: Addressed via https://github.com/apache/netbeans/pull/5432/commits/e2dd93481e809985df55c138448cbaaf0ab75f4b
Learned something new here - thanks.

ide/css.lib/src/org/netbeans/modules/css/lib/properties/GrammarParser.java
matthiasblaesing: Yes implementation of `PropertyGrammarElementRef` was bogus. Addressed via:
https://github.com/apache/netbeans/pull/5432/commits/ecd64eee0b30d600967c751cae5281cf842daa1a

matthiasblaesing:
@vieiro thank you for the review. I pushed updates - could you have another look?

vieiro:
> @vieiro thank you for the review. I pushed updates - could you have another look?

@matthiasblaesing Great job. Thanks again for this!

Issue #4952, opened 2022-11-09
Remove Thread.stop() usage


mbien:
Those methods are deprecated since Java 1.2 and will throw UnsupportedOp exceptions from JDK 20+.

got reminded today again by getting this notification:
https://inside.java/2022/11/09/quality-heads-up/

matthiasblaesing:
I asked `jackpot` for help and it answered:

```
nbi/engine/src/org/netbeans/installer/downloader/dispatcher/impl/RoundRobinDispatcher.java:122:
nbi/engine/src/org/netbeans/installer/downloader/dispatcher/impl/RoundRobinDispatcher.java:168:
harness/o.n.insane/src/org/netbeans/insane/scanner/ScannerUtils.java:256:
harness/o.n.insane/src/org/netbeans/insane/scanner/ScannerUtils.java:267:
ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerModule.java:84:
ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerModule.java:151:
ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerModule.java:157:
ide/extbrowser/test/ExtBrowser/qa-functional/src/org/netbeans/test/gui/web/util/HttpRequestWaitable.java:117:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:78:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:112:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:147:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:181:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:217:
ide/editor.search/test/unit/src/org/netbeans/modules/editor/search/EditorFindSupportTest.java:253:
extide/o.apache.tools.ant.module/src-bridge/org/apache/tools/ant/module/bridge/impl/BridgeImpl.java:329:
java/lib.jshell.agent/agentsrc/org/netbeans/lib/jshell/agent/AgentWorker.java:493:
java/java.lsp.server/test/unit/src/org/netbeans/modules/java/lsp/server/protocol/ServerTest.java:263:
[...]

PR #5435, opened 2023-02-05
finalize() proposal

BradWalker:
Commit with prototype changes. Please see the conversation on the [discussion page](https://github.com/apache/netbeans/discussions/5436)

DO NOT INTEGRATE!

### DIFF
--- a/ide/editor.lib/nbproject/project.properties
+++ b/ide/editor.lib/nbproject/project.properties
@@ -16,7 +16,8 @@
 # under the License.
 
 javac.compilerargs=-Xlint:unchecked
-javac.source=1.8
+javac.source=9
+javac.target=9
 spec.version.base=4.27.0
 is.autoload=true
 
--- a/ide/editor.lib/src/org/netbeans/editor/LeafElement.java
+++ b/ide/editor.lib/src/org/netbeans/editor/LeafElement.java
@@ -63,6 +63,7 @@ public class LeafElement extends BaseElement {
         }
     }
 
+    @Deprecated(since="9", forRemoval=true)
     protected void finalize() throws Throwable {
         try {
             startMark.remove();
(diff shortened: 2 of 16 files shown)


BradWalker:
It was unclear how I should propose a "prototype" for discussion. So I recreated a discussion about it and submitted this pull request to help. It's not my intention to have this integrated.

If I didn't follow protocol, then just let me know. Also, if there is a better way, then also, please, let me know.

matthiasblaesing:
If this affects NetBeans development I think it belongs onto ` dev@`. I don't read discussions and I'm pretty sure I'm not the only developer.

neilcsmith-net:
I have no doubt we need to address the end of finalization, but I'm not sure we need to be changing the base release version of modules just to add the extended variety of the deprecated annotation?!

BradWalker:
> I have no doubt we need to address the end of finalization, but I'm not sure we need to be changing the base release version of modules just to add the extended variety of the deprecated annotation?!

Hey @neilcsmith-net , I mostly agree with your comments.

Here's the problem.

We are going to need to change the base release if and when we resolve the finalization issue. That much we all agree on. I'm simply proposing that we start the process to make the work easier later on.

The first step in that process, to me, would be to mark the `finalize()` methods as deprecated like I propose.

neilcsmith-net:
> We are going to need to change the base release if and when we resolve the finalization issue.

You can replace `finalize()` using JDK 8 if need be.  Admittedly not if we decide to use JDK Cleaner API, but that's not the only option.

We haven't yet agree to move the platform beyond JDK 8.  Some of these should upgrade to JDK 11 when we're ready to move the platform as a whole.  At least `o.n.bootstrap` will need to wait for that.

mbien:
we should track this as issue.
we have one for ThreadDeath: #4952
and one umbrella issue for all JDK 11 incompatible tests: #4904

added #5440 for the finalization mechaninsm

Issue #5427, opened 2023-02-04
PHP: The documentation does not show the property type unless the @var tag is specified for the property in phpDoc


troizet:
### Apache NetBeans version

Apache NetBeans 16

### What happened

The documentation does not show the property type unless the `@var` tag is specified for the property in phpDoc.

### How to reproduce

 - write a code

```php
class newPHPClass {

  //The type is displayed here. 
  public string $test;  
    
  /**
   * The type is not displayed here. 
   */
  public string $test2;
  
  /**
   * The type is displayed here. 
   * @var string
   */
  public string $test3; 

}
```
- use autocomplete or `Ctrl + hover` to display the documentation.

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Ubuntu 20.04.5 LTS

### JDK

OpenJDK 64-Bit Server VM 11.0.17+8-post-Ubuntu-1ubuntu220.04

### Apache NetBeans packaging

Apache NetBeans binary zip

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

PR #5443, opened 2023-02-06
[GH-5427] PHP: Fix for displaying property type in documentation if @…

troizet:
What was done in this PR:
- fixed issue #5427
- new tests for fixed issue have been added

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
@@ -439,8 +439,8 @@ final class DocRenderer {
                 }
             }
 
-            // field without phpdoc
-            if (phpDocBlock == null
+            // field without phpdoc or with but without @var tag
+            if ((phpDocBlock == null || !tagInTagsList(tags, PHPDocTag.Type.VAR))
                     && indexedElement instanceof FieldElement) {
                 FieldElement fieldElement = (FieldElement) indexedElement;
                 Set<TypeResolver> types = fieldElement.getInstanceTypes();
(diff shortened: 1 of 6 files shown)


matthiasblaesing (approved):
Looks sane to me.

matthiasblaesing:
@junichi11 could you also please have a look?

junichi11 (approved):
Thank you for your work!

junichi11:
@tmysik If there is no problem, let's merge it :) Thanks!

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
```diff
@@ -454,6 +454,20 @@ private void extractPHPDocBlock() {
                     others.toString()));
         }
 
+        private boolean tagInTagsList(List<PHPDocTag> tags, AnnotationParsedLine tagKind) {
```
tmysik, line 457: Likely not a problem here, but using a `Set` and simple lookup instead of `List` and iteration would be (much) better here. Would that be possible?

tmysik (approved):

tmysik:
A question added, let's wait for a while :) Then, let's make it. Thanks for this fix!

php/php.editor/src/org/netbeans/modules/php/editor/completion/DocRenderer.java
troizet: For the reason that I am a php developer, I know java only superficially :) Thanks for the tip. I will keep it in mind next time.

tmysik: OK, so maybe @junichi11 could try to improve it one day :)

troizet:
Thanks!

PR #5453, opened 2023-02-08
Wildfly server restart not possible

akronenw:
This PR fixes an issue with Wildfly servers. I noticed that, after successfully starting a Wildfly (JBoss) server, I was not able to restart or stop the instance in the `Services->Server->rightclick on server`-view. The respective menu items were grayed out, instead it was possible to start the server (although it was already running).
I investigated that and found out, that NetBeans did not take into account the port offset that can be specified in the `standalone.xml` server config file. Hence, in case a port offset is specified, NetBeans always checks against the wrong port and does not recognize a correctly started server.

What I did:
* I bundled hardcoded port numbers in a single place.
* I added a handler for port bindings from the config file, that also saves the port offset.
* When the server is set up, the config file is parsed and the offset added to the ports specified in the NetBeans server wizard.
* I refactored some code, mostly related to the files I touched anyway, hopefully not too much.

I tested this successfully on my machine, but I am not sure whether there are any side effects that I did not fathom.
Two things that I'd like to ask:
* Should NetBeans show a hint in the wizard when a port offset is detected, so the user knows the ports he enters are not the ones under which the running instance is reachable?
* While testing I got a bunch of warnings: `Parameter file was not normalized...` from `FileUtil#getFileObject`. After a clean build those warnings disappeared, they did not seem to do any harm. Why did that happen?

### DIFF
--- a/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/common/api/SocketBinding.java
+++ b/enterprise/j2eeserver/src/org/netbeans/modules/j2ee/deployment/common/api/SocketBinding.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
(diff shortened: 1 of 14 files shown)


mbien:
hi @akronenw, thanks for working on this!

> I refactored some code, mostly related to the files I touched anyway, hopefully not too much.

We usually try to avoid having cleanup and a bugfix in the same commit since it obfuscates the history esp when git blame is used. Would have been great if it would be kept separate here too - it makes review also easier.

akronenw:
Hello @mbien,
I understand. In the meantime I found out I solved this with too much effort, so I prepare a new PR and close this one. Sorry for the noise.
Alex

mbien:
@akronenw all good! Code cleanup in form of refactoring is perfectly fine btw, just keep it in a separate commit if possible and please avoid unnecessary formatting changes (e.g don't reformat whole files where you touched only a method).

Thanks a lot for fixing this issue!

akronenw:
@mbien just to be sure - I automatically type CMD-SHIFT-F for for code formatting before saving. Every missing or abundant whitespace etc is then added/removed in the file. Did I get you right, that such formatting should be avoided?

mbien:
@akronenw yes please don't do that :)

NetBeans' codebase is fairly old and has various styles. It is likely that formatting will create many changes which are just noise in the diff which makes git blame for example less useful.

You can format your own code of course since its new code.

PR #5455, opened 2023-02-09
use default javac release option for the NetBeans build

mbien:
follow up to #5122

attempt to remove project specific `javac.source/target/release` properties and sets them to `javac.default.release` in common.xml so that projects can still overwrite the default where required (example would be java/java.disco).

excluded some paths which use build-impl.xml instead of the NB infrastructure:
  - nbi/
  - webcommon/cordova/
  - java/java.j2seproject/copylibstask/
  - java/javawebstart/AntTasks/

full build is working locally and CV tests are green too (tested with JDK 19), ran only rudimentary manual tests.

this created an interesting problem I didn't anticipate:

NetBeans doesn't know anymore what javac level the projects have when they are opened. Have to find a fix for that.

but first lets see if a full test run is green.


jlahoda:
Sorry i misclicked and marked it as ready by accident.

mbien:
haha. looks like most tests except those few I ran locally is red. Going to take another look.

tbw777:
Which problem is fixed by?

neilcsmith-net:
> Which problems is fixed by this PR? Compile with new jdk can be done without change bytecode version.

Default bytecode / release version.

tbw777:
@neilcsmith-net 
No needed to change bytecode, only imports ))

neilcsmith-net:
@tbw777 that's off-topic to this PR.

mbien:
> Think we might have a little bit of work to do! smile I only searched for `javac.target`, so stupidly missed how many more modules just set `javac.source`! roll_eyes

the reason for that is because the custom javac ant task is mapping the properties in a non-standard way:
https://github.com/apache/netbeans/blob/master/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java#L68-L100

but we are getting there. The good thing is that all values are now basically the same.


i saw this in the log `The Java platform version 8 or higher was requested but only 1.8 is running.` This indicates that i clearly overlooked some usages of the target property. Because this sounds like a module system warning so the number must have made it into the manifest. https://github.com/apache/netbeans/blob/410710ec8669a6134b508d28136627e39cab8d27/platform/autoupdate.services/src/org/netbeans/modules/autoupdate/services/DependencyChecker.java#L116

I did set source/target to 1.8 for the current test run.

going to try to get the default release property in too so that we can use it for all the other javac calls, e.g
https://github.com/apache/netbeans/pull/3278/commits/bf8c26d91ea20cb7bf29b4510938f572f2202034

matthiasblaesing:
> > Think we might have a little bit of work to do! smile I only searched for `javac.target`, so stupidly missed how many more modules just set `javac.source`! roll_eyes
> 
> the reason for that is because the custom javac ant task is mapping the properties in a non-standard way: https://github.com/apache/netbeans/blob/master/nbbuild/antsrc/org/netbeans/nbbuild/CustomJavac.java#L68-L100

That is a n artifact of the NetBeans build itself, so the task is not set in stone and could be changed.

mbien:
lets try this first: https://github.com/apache/netbeans/pull/7188

PR #5463, opened 2023-02-10
Patterns normalization

tbw777:
java + xsd (separate commits)

Changes:

- [a-zA-Z0-9_] -> \w
- [0-9] -> \d
- [^\s] -> \S
- removed unnessasary escaping

BUILD SUCCESSFUL


BradWalker:
I have to say that regular expressions is definitely not my strong point. So I'm not inclined to review this, even if it is easy. Especially since a change, even if it is considered "simple" could possibly introduce a change in behavior.

Again, if it were me, I would  say to myself: "Self, do we have a problem here or are we changing this just to be technically pure?"

tbw777:
at your discretion

matthiasblaesing:
From the examples in the description I don't see a real benefit. Replacing explicit character groups with their symbol does not help readability or at least it highly depends on the location. A quick look at the code changes supports this. From a review perspective this a nightmare and from a benefit perspective I don't see it.

tbw777:
[8684e80](https://github.com/apache/netbeans/pull/5463/commits/8684e80daf35224eb23b0d420946196d84f0dc40)
Can we use this commit?

lkishalmi:
@tbw777 I've got a bit curious. Why are you doing these? What's your purpose with these kind of PRs?

PR #5464, opened 2023-02-10
Prevented unnessasary StringBuilder/StringBuffer creation

tbw777:
No need to create SB every time.
Its more better to reuse existing object via setLength(0) method call.

BUILD SUCCESSFUL

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ModuleList.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/universe/ModuleList.java
@@ -547,7 +547,7 @@ public final class ModuleList {
             }
         } catch(IllegalArgumentException e) {
             LOG.log(Level.WARNING, "Error getting subelements, malformed xml");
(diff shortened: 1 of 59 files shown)


lkishalmi:
Well, yet another set of ambivalent changes.

Some of these could be valid, some of them indifferent and there are a few places, when it hurts the readability of the code a little.

Object creation in Java is cheap, really cheap. Doing these optimization on exception handlers, test code, simply does not worth.

Also there are a lot of places where the new code is more complex than the original one including a null check. that's just bad.

lkishalmi:
And very low value/noise ratio.

mbien:
object reuse can be more expensive than allocating a new object. That is the whole point of a generational garbage collector: it is very good at cleaning short lived objects in chunks since it assumes most of them are short lived.

I am not in favor of blindly applying this everywhere since it is not obvious that it is of any benefit. If there are places identified which benefit from this - awesome. But lets not do this everywhere.

going to remove myself from the review list since I don't really want to review this unless it is reduces to a manageable size and proven that it is of any benefit.

PR #5465, opened 2023-02-10
Added ExecutorService close() where possible

tbw777:
try-with-resource for closing resources


matthiasblaesing:
This is invalid. `close` was added to `ExecutorService` in 19 and is just an alias for `shutdown`.

tbw777:
You are right

PR #5468, opened 2023-02-11
Moved operations closest to usage place

tbw777:
Changes was maked for next cases:

- Not a primitive links or types
- No side effects
- File not a test
- If memory allocation and returns exist
- If return exist
- If complex operation
- If increase readability
- Single place of usage - no need on other conditional cases
- If paint method
- If synchronized method body inside init
- If xml processing

Covarage:

- websvccommon
- webcommon
- platform
- nbi
- nbbuild
- nb
- apisupport

BUILD SUCCESSFUL

Example:
```java
@Override
public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, 
        boolean hasFocus, int row, int column) {

    /*moved below >>>*/
    //Component ret =  super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, column);

    String paramName = (String) tableModel.getValueAt(row, 0);

    if (value == null) {
    } else if (paramName.contains(Constants.PASSWORD)) {
        return new JPasswordField((String) value); //<<< no catch return found
    } 

    //return ret; //removed
    return /*>>>*/ super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, column);
}
```

```java
public CloseOperationState canCloseElement() {
    /*moved below >>>*/
    //final CloneableEditorSupport sup = getLookup().lookup(CloneableEditorSupport.class);

    Enumeration<CloneableTopComponent> en = getReference().getComponents();
    if (en.hasMoreElements()) {
        en.nextElement();
        if (en.hasMoreElements()) {
            // at least two is OK
            return CloseOperationState.STATE_OK;
        }
    }
    
    Savable sav = getLookup().lookup(Savable.class);
    if (sav != null) {
        AbstractAction save = new AbstractAction() {
            @Override
            public void actionPerformed(ActionEvent e) {
                try {
                    /*>>>*/ 
                    CloneableEditorSupport sup = getLookup().lookup(CloneableEditorSupport.class);
                    sup.saveDocument();
                } catch (IOException ex) {
                }
            }
        };
```
```java
@Override
public void actionPerformed (ActionEvent e) {
    WindowManagerImpl wmi = WindowManagerImpl.getInstance();
    ModeImpl contextMode = getMode2WorkWith();
    if( null == contextMode )
        return; //just being paranoid
    if( null != slidingMode ) {
        //WindowManagerImpl.getInstance().userRestoredMode( slidingMode, contextMode );
        wmi.userRestoredMode( slidingMode, contextMode ); //<<<
    } else {
        boolean isDocked = contextMode.getState() == Constants.MODE_STATE_JOINED;

        if (!isDocked) {
            wmi.userDockedMode(contextMode);
        }
    }
}
```
```java
private boolean processShortcut(KeyEvent ev) {
    //ignore shortcut keys when the IDE is shutting down
    if (NbLifecycleManager.isExiting()) {
        ev.consume();
        return true;
    }

    //KeyStroke ks = KeyStroke.getKeyStrokeForEvent(ev); //<<< removed
    Window w = SwingUtilities.windowForComponent(ev.getComponent());

    // don't process shortcuts if this is a help frame
    if ((w instanceof JFrame) && ((JFrame)w).getRootPane().getClientProperty("netbeans.helpframe") != null) // NOI18N
        return true;

    KeyStroke ks = KeyStroke.getKeyStrokeForEvent(ev); //<<< also synchronized here
```
```java
private void checkSemicolon(int offset) {
//...
//int originalOffset = ts.offset(); //<<< removed
CharSequence originalText = ts.token().text();

Token<? extends JsTokenId> previous = LexUtilities.findPrevious(ts, IGNORED);
if (previous != null && previous.id() == JsTokenId.OPERATOR_SEMICOLON) {
    return;
}
// we are probably at the end of file without the semicolon
int originalOffset = ts.offset(); //<<< also synchronized here
fileOffset = context.parserResult.getSnapshot().getOriginalOffset(originalOffset);
addMissingSemicolonHint(fileOffset, originalText.toString());
```
```java
@Override
public void run(JsParserResult result, SchedulerEvent event) {
    //remove the last occurrences - the CSL caches the last found occurences for us
    //range2Attribs = null; //<<< removed
    if (range2Attribs != null) {
        range2Attribs.clear();
    }

    if(cancelled) {
        cancelled = false;
        return ;
    }
    int offset = result.getSnapshot().getEmbeddedOffset(caretPosition);
    //Set<OffsetRange> ranges = findOccurrenceRanges(result, offset); //<<< removed
    //range2Attribs = new HashMap<>(); //<<< removed
    if(cancelled) {
        cancelled = false;
        return ;
    }

    if (range2Attribs == null) {
        range2Attribs = new HashMap<>();
    }

    if (range2Attribs == null) { //<<<
        range2Attribs = new HashMap<>(); //<<<
    } //<<<

    Set<OffsetRange> ranges = findOccurrenceRanges(result, offset); //<<< complex operation
    for (OffsetRange offsetRange : ranges) {
        range2Attribs.put(ModelUtils.documentOffsetRange(result, offsetRange.getStart(), offsetRange.getEnd()), ColoringAttributes.MARK_OCCURRENCES);
    }
}
```

### DIFF
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/JavadocForBinaryImpl.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/queries/JavadocForBinaryImpl.java
@@ -65,11 +65,11 @@ public final class JavadocForBinaryImpl implements JavadocForBinaryQueryImplemen
         }
         String cnbdashes = cnb.replace('.', '-');
         try {
-            final List<URL> candidates = new ArrayList<URL>();
             NbPlatform platform = project.getPlatform(false);
             if (platform == null) {
                 return null;
             }
+            final List<URL> candidates = new ArrayList<URL>();
             for (URL root : platform.getJavadocRoots()) {
                 candidates.add(new URL(root, cnbdashes + "/")); // NOI18N
             }
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/LibrariesNode.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/LibrariesNode.java
@@ -274,11 +274,11 @@ final class LibrariesNode extends AbstractNode {
         }
 
         private Node createLibraryPackageViewNode(FileObject jfo) {
-            Icon icon = getLibrariesIcon();
             FileObject root = FileUtil.getArchiveRoot(jfo);
             if (root == null) {
                 return Node.EMPTY;
             }
+            Icon icon = getLibrariesIcon();
             String name = String.format(getMessage("LBL_WrappedLibraryFmt"), FileUtil.toFile(jfo).getName());
             return ActionFilterNode.create(PackageView.createPackageView(new LibrariesSourceGroup(root, name, icon, icon)));
         }
--- a/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/SuiteCustomizerLibraries.java
+++ b/apisupport/apisupport.ant/src/org/netbeans/modules/apisupport/project/ui/customizer/SuiteCustomizerLibraries.java
@@ -401,11 +401,6 @@ public final class SuiteCustomizerLibraries extends NbPropertyPanel.Suite
     
     @Override
     public void store() {
-        // TODO C.P: disable buttons/show button "Upgrade"??? on old harness
-        Set<String> enabledClusters = new TreeSet<String>();
-        Set<String> disabledModules = new TreeSet<String>();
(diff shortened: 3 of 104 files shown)


matthiasblaesing:
Which problem is fixed by a change over 107 files? As now multiply stated: Beautification for its own sake is not welcome. This just adds noise to the codebases history.

tbw777:
To execute only needed code

lkishalmi:
Do you have statistical analysis, that with these changes the IDE performs any better? And if yes, how munch? Of course the overall performance of the IDE would be questionable "metric" alone.

-1 again.

Why are you doing these? What's your purpose with these kind of PRs?

PR #5469, opened 2023-02-11
Streams from java.nio.file.Files have to be closed

tbw777:
Some resources was in uncloused condition.

### DIFF
--- a/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CheckEmbeddedBinaries.java
+++ b/nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CheckEmbeddedBinaries.java
@@ -26,10 +26,13 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.math.BigInteger;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Stream;
+
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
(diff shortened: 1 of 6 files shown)


nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CheckEmbeddedBinaries.java
```diff
@@ -94,9 +97,11 @@ public void execute() throws BuildException {
             throw new BuildException("Could not open " + manifest + ": " + x, x, getLocation());
 
         }
-        try {
-            StringBuilder errorList = new StringBuilder();
-            Files.list(dir.toPath())
+
+        StringBuilder errorList = new StringBuilder();
```
matthiasblaesing, line 101: The `errorList` can be moved back into the `try` block.

matthiasblaesing (commented):
Looks sane to me. I left one inline comment.

lkishalmi:
@tbw777 I would not say this is a good PR, but one which actually could be merged. It creates relative small noise only 6 files affected. Though only 2 of them are actually in the IDE. Even those two are on some auxiliary execution path, used once or twice during one IDE execution.

The most value of this PR is an educative one, which is summarized succinctly by the new title.

nbbuild/antsrc/org/netbeans/nbbuild/extlibs/CheckEmbeddedBinaries.java
tbw777: Fixed, ty

tbw777:
@lkishalmi 
ty

lkishalmi (approved):
Looks good.

matthiasblaesing (approved):

matthiasblaesing:
Thank you.

Issue #5375, opened 2023-01-27
PHPDoc works only partially for class variables (ex: string vs array)


tanzanite777:
### Apache NetBeans version

Apache NetBeans 16

### What happened

Netbeans seemingly fails to parse class variable PHPDoc if it is 'array' type. Consequently, code completion and inspection (hold control key and hover - i believe it is the default binding) cannot show the content of the PHPDoc block comment.

Works correctly with everything else i can think of.

### How to reproduce

Create a php project (8.1) with this file:

```
<?php declare(strict_types=1);

class Test {
  /**
   * This text will show up correctly. Works with any other type, except array, that i can think of - not just 'string'.
   * @var string
   */
  static string $works;

  /**
   * This PHPDoc block comment is completely ignored and this text will not show up anywhere.
   * @var array
   */
  static array $snafu;
}
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

System: Windows 10 version 10.0 running on amd64; UTF-8; en_US (nb)

### JDK

Java: 18; Java HotSpot(TM) 64-Bit Server VM 18+36-2087

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

100% reproducible.

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

troizet:
Similar behavior for the intersection type. 
I can make a fix for the intersection type and a workaround for the array, but I can't make the correct tests until #5426, #5427 are fixed.

PR #5475, opened 2023-02-12
PHP: Changed the way to parse an array type by analogy with other types.

troizet:
What was done in this PR:
- Changed the way to parse an array type by analogy with other types
- Fixed broken tests
- Fixed issue #5375
- New tests for fixed issue have been added
- Fixed broken `UnusableTypesHintError`

The reason for issue #5375 is that the condition: https://github.com/apache/netbeans/blob/80ee045a8fb276323d6f756c8947950a07dbf540/php/php.editor/src/org/netbeans/modules/php/editor/parser/api/Utils.java#L106-L108 is not fulfilled, because the array type is parsed differently than the other type.
In this PR I tried to change the parser so that the array type is parsed similarly to other types.
Locally, all unit tests passed successfully.
Manual testing showed no problems either.


troizet:
Somehow missed the broken test. Sorry. I'll keep looking into it.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
```diff
@@ -590,7 +587,7 @@ private static boolean isCallableType(Identifier identifier) {
         private static boolean isVoidType(NamespaceName namespaceName) {
             return Type.VOID.equals(CodeUtils.extractUnqualifiedName(namespaceName));
         }
-
+
```
junichi11, line 590: Please remove whitespaces.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
```diff
@@ -576,11 +574,10 @@ private void createRedundantTypeCombinationError(ASTNode node, UnionType unionTy
         private void createIterableRedundantTypeCombinationError(UnionType unionType, IterableRedundantTypeCombination.RedundantType redundantType) {
             hints.add(new IterableRedundantTypeCombination(rule, fileObject, unionType.getStartOffset(), unionType.getEndOffset(), unionType, redundantType));
         }
-
-        private static boolean isArrayType(Identifier identifier) {
-            return !identifier.isKeyword()
-                    && Type.ARRAY.equals(identifier.getName().toLowerCase(Locale.ENGLISH));
-        }
+
```
junichi11, line 577: Please remove a whitespace.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
troizet: Done.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
troizet: Done.

junichi11 (approved):
Thanks!

php/php.editor/tools/ASTPHP5Parser.cup
```diff
@@ -2625,10 +2625,9 @@ fully_qualified_class_name:className
     RESULT = className;
 :}
 
-| T_ARRAY:array
+| T_ARRAY:name
 {:
-    Identifier classId = new Identifier(arrayleft, arrayright, "array");
-    RESULT = classId;
+    RESULT = NamespaceName.create(nameleft, nameright, name);
```
tmysik, line 2630: @junichi11 Is this really correct? I am sorry, but I simply cannot recall it... (yes, it is a long time ago, already :)

Thanks.

tmysik (approved):

tmysik:
@junichi11 Please, see my question above. If it is OK, feel free to merge this PR. Thank you.

tmysik:
@troizet Thanks a lot for this PR! It is really great to have it!

php/php.editor/tools/ASTPHP5Parser.cup
junichi11: `string` is recognized as a namespace name with #4037
https://github.com/apache/netbeans/blob/1d85a3e854fafef563a3b0986b88f4e9613a1d85/php/php.editor/tools/ASTPHP5Parser.cup#L4323-L4348

So, this should be OK, I think.
If we come across a problem, let's fix it. Thanks!

tmysik: Thanks!

PR #5476, opened 2023-02-12
PHP: Corrected a typo in the php parser tests

troizet:
Corrected a typo in the php parser tests.


KacerCZ:
Change looks sane.
Let's see result of CI.

KacerCZ (approved):

junichi11:
@mbien Doesn't work tests of php.editor?

mbien:
> @mbien Doesn't CI run tests of php.editor?

hi @junichi11, the label was added *after* the workflow request was built. Starting a fresh workflow will take new labels into account (restarting will not unfortunately).

This is the main reason we have the "secret" workaround of PR locking/unlocking which is used a trigger for a fresh workflow run (instead of restarting an old run which does not update its label information).

going to do this now and start a fresh workflow run here

junichi11:
@mbien OK. Thanks!

mbien:
@junichi11 thanks for checking that the correct tests run during review!

junichi11 (approved):

Issue #3936, opened 2022-04-06
Netbeans 13 integration with Wildfly 26.0.1 final Server


Kareemelagroudy:
### Apache NetBeans version

Apache NetBeans 13

### What happened

Hey, I'm creating some webapps that must be run on Wildfly server. The first time I run the project, It runs smoothly but it doesn't deploy afterwards (when modified or on 2nd, 3rd, .... times), I have to clean and build the project again in order to run it again
![Capture](https://user-images.githubusercontent.com/58118555/162054376-2f299e89-a49f-4929-bcca-efd04e6f8194.PNG)

### How to reproduce

The problem can be reproduced by modifying the project and running it again

### Did this work correctly in an earlier version?

No

### Operating System

Windows 10

### JDK

JDK 18

### Apache NetBeans packaging

Apache NetBeans provided installer

### Anything else

The problem occurs everytime I modify and rerun the project

### Are you willing to submit a pull request?

Yes

### Code of Conduct

Yes

asbachb:
@Kareemelagroudy I could not reproduce this on NetBeans 18

Could you give some feedback is the issue is still persistent on NetBeans 18? Otherwise please close the issue.

PR #5478, opened 2023-02-12
Support port offset for wildfly server

akronenw:
This solves the same issue as mentioned in #5453, but cleaner.
The problem: Wildfly servers are configured with XML files like `standalone.xml`. Those contain socket binding groups that comprise a bunch of ports and can have a port offset defined. One such group is the `standard-sockets` group in which for example the management port is defined. All variables in the config file can (must?) be written like `${jboss.socket.binding.port-offset:10000}`, that is, a variable and the fallback (or default) value separated by a `:`.
NetBeans does not set the port-offset variable, hence the default from the config file is used, but NetBeans does not parse it either so the Wildfly server runs on a different port as known by NetBeans. That leads to a correctly started server that can't be restarted, stopped or refreshed from the NetBeans server context view (but it is possible to start it regularly).

I fixed this by parsing the config and extracting the default port offset from the standard socket group. The offset is then added to the admin port value (which can be set in the add server wizard) in the deployment manager.

I did not add the offset to the regular port, because I have not yet found an issue with that (most likely I will, sooner or later) and that would mean touching a lot more files.

### DIFF
--- a/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/WildflyDeploymentManager.java
+++ b/enterprise/javaee.wildfly/src/org/netbeans/modules/javaee/wildfly/WildflyDeploymentManager.java
@@ -56,6 +56,7 @@ import org.netbeans.modules.javaee.wildfly.ide.commands.WildflyModule;
 import org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginProperties;
 import static org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginProperties.DEFAULT_ADMIN_PORT;
 import static org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginProperties.PROPERTY_ADMIN_PORT;
+import static org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginProperties.PROPERTY_CONFIG_FILE;
 import org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginUtils;
 import org.netbeans.modules.javaee.wildfly.ide.ui.WildflyPluginUtils.Version;
 import org.netbeans.modules.javaee.wildfly.util.WildFlyProperties;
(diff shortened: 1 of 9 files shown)


mbien:
@ehsavoie want to take a look at this?

would be good to check if this closes any issues, e.g description of #3936 sounds potentially related.

ehsavoie:
The issue is that this doesn't fix the problem: if the offset is set when starting WildFly using -Djboss.socket.binding.port-offset=100 then you are back on it since you would need to connect to the running server management API to get the effective value.
I'm a bit surprised because I thought that you could pass the management port in the plugin which is (in my opinion) the only viable solution:
"The offset is then added to the admin port value (which can be set in the add server wizard)"

I don't understand why you don't set the proper port value per server instance.

akronenw:
Hello @ehsavoie, 
thanks for checking this.

> if the offset is set when starting WildFly using -Djboss.socket.binding.port-offset=100 then you are back on it since you would need to connect to the running server management API to get the effective value

I don't understand that. Do you mean, starting Wildfly manually and passing the port-offset as parameter? If so, that would not make sense, as the whole point is that NetBeans starts the server and passes everything in.

> I don't understand why you don't set the proper port value per server instance.

Yes, I could do that. But the port offset in the `standalone.xml` applies to more ports than admin and HTTP. As NetBeans only allows to set admin and HTTP port, I'd have to remove the port offset in the config file and add it to the other ports by hand. That is inconvenient and makes the port offset variable pointless.
Unfortunately I have to deal with several Widlfy instances of different versions and just setting the port offset in the config makes it easier to have them running in parallel. 

Probably the best is to scrap my commits and add a new field in the server wizard where one can specify the port offset, which is then passed to Wildfly via `-Djboss.socket.binding.port-offset=Y`. I am not keen on doing Swing stuff so I tried to fix the problem with the commits you checked. 

What do you think?

ehsavoie:
Yes I'd like this idea . Also note that the plugin is able to connect to a running instance.

akronenw:
I will open a new PR for the new approach.

ehsavoie:
@akronenw maybe you would want to take a look at https://github.com/ehsavoie/netbeans/pull/new/support_port_offset_for_wildfly_server

PR #5487, opened 2023-02-13
Removed a lot of compiler warnings

tbw777:
Inserted diamond in constructions like this:
ArrayList<Strings> = new ArrayList();
No any interfaces was changed

### DIFF
--- a/apisupport/apisupport.refactoring/src/org/netbeans/modules/apisupport/refactoring/NbMoveRefactoringPlugin.java
+++ b/apisupport/apisupport.refactoring/src/org/netbeans/modules/apisupport/refactoring/NbMoveRefactoringPlugin.java
@@ -82,7 +82,7 @@ public class NbMoveRefactoringPlugin extends AbstractRefactoringPlugin {
     private EditableManifest targetManifest;
     
     private Map packagePostfix = new HashMap();
-    ArrayList<FileObject> filesToMove = new ArrayList();    
+    ArrayList<FileObject> filesToMove = new ArrayList<>();
     HashMap<FileObject,ElementHandle> classes;
     
     /**
@@ -647,7 +647,7 @@ public class NbMoveRefactoringPlugin extends AbstractRefactoringPlugin {
     }  
    
    private void initClasses() {
-        classes = new HashMap();
+        classes = new HashMap<>();
         for (int i=0;i<filesToMove.size();i++) {
             final int j = i;
             try {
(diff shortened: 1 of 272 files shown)


BradWalker (approved):
Nice work!

mbien:
please don't ever merge PRs which don't have tests run on them! This PR broke master. There is no need to hurry. 

This is a project wide cleanup which requires all tests to be run which is activated with the `ci:all-tests` label.

BradWalker:
> please don't ever merge PRs which don't have tests run on them! This PR broke master. There is no need to hurry.
> 
> This is a project wide cleanup which requires all tests to be run which is activated with the `ci:all-tests` label.

As someone who has previously worked on things in this area, I spent the time to look at each file that was changed. So I'll accept the responsibility for what happened. It appeared to me that the tests were being "skipped" by design. Previously when I've looked at the test runs, the system always showed as "failed". So it was clearly my mistake. 

My apologies to everyone! I'll be more diligent next time!

mbien:
the changes itself were fine, the problem was that some were in test data. Just don't be so quick with the merge button next time. It is perfectly fine to let a PR sit a few days - increases the probability that someone notices things like this, e.g missing label to run all tests.

BradWalker:
> the changes itself were fine, the problem was that some were in test data. Just don't be so quick with the merge button next time. It is perfectly fine to let a PR sit a few days - increases the probability that someone notices things like this, e.g missing label to run all tests.

Agreed. 

I looked at each change. But, making a change in test code is definitely a danger. 

Another good reason to keep these type of PRs smaller in size.

mbien:
> And don't worry, @BradWalker ! This is something that could happen to anyone!

@vieiro @BradWalker yes! This happens sometimes, so please don't worry about it. Its just that if you would have waited with the merge 1-2 days after review, I am certain someone would have noticed.

why?

because I for example try to look at https://github.com/apache/netbeans/actions on a daily basis to check if there is something to play whack-a-mole with, that is also how I noticed this in the first place. When someone syncs a new or old PR I would notice this too (since it appears on the all-knowing list) and probably check if its properly labeled. IMO the main lesson learned here should be to not rush PRs in, even if they seem trivial and were carefully reviewed.

This could have happened with a small PR too which only covers one module or less - it likely would have required some of the [[ci] labels](https://github.com/apache/netbeans/labels?q=%5Bci%5D) too. Java PRs need Java, PHP needs PHP otherwise the correct tests won't run etc etc.

mbien:
also regarding large cleanup PRs in general:
in my book they are fine as long they remain reviewable (of course the PR itself should have a purpose too - I am not saying that every possible automated cleanup is worth the review time or to be integrated). It helps a lot to split large PRs into commits - if someone would only have the time to review one commit, thats fine - add a comment to tell the next reviewer.

We can't split everything into PRs otherwise every project wide cleanup would mean 18 PRs, one for each cluster - lets not do that.

PR #5502, opened 2023-02-15
Removed unnecessary new Object calls on MessageFormat. Multicommit

tbw777:
Removed new operations from code like this:
MessageFormat.format(formater, new Object[]{ value });


vieiro:

As a consequence, you're not really "Removed unnecessary new Object calls on MessageFormat" , the compiler will add these for you.

I can't see enough value on this PR.

PR #5504, opened 2023-02-15
Removed unnecessary new Class[] {} calls. Multicommit

tbw777:


vieiro:

As a consequence, you're not really "Removed unnecessary new Class[] {} calls.", as the compiler will add those for you.

I can't see enough value on this PR.

PR #5505, opened 2023-02-15
Contributor has personal preferences on patterns, and introduces bugs.

tbw777:
jmh preformance proof: https://gist.github.com/tbw777/6dc75f03ef122e55477f883154bf856b
jmh tests: https://gist.github.com/tbw777/4e27229db015251088a9801aaa2bf2f0


matthiasblaesing:
I think this reduces readbility and the JMH benchmarks are incorrect. `\w` is not `[A-Za-z0-9_]` and the same is true for the antipattern. There are also replacements in messages, which are equally questionable.

From me: -1

tbw777:
\w is [A-Za-z0-9_]

matthiasblaesing:
There is no underscore:

https://github.com/openjdk/jdk/blob/jdk-17+35/src/java.base/share/classes/java/util/regex/Pattern.java#L178-L179

and also not here:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html#predef

> \w  A word character: [a-zA-Z_0-9]

tbw777:
You touch to underscore description, its not at the end (any position is applicable) in your examples.

matthiasblaesing:
A sorry, overlooked, still, not more readable from my perspective.

tbw777:
Yee, but with \w you cant be overlooked )

matthiasblaesing:
No, but I have to think what `\w` might mean. It might also change meaning depending on flags. It still leaves the question whether or not the messages should be changed and whether or not your JMH benchmarks have practical impact on real-life performance.

tbw777:
>but I have to think what \w might mean.
Im disagree. Also none \w ways arent normalized: 24 options.

>It still leaves the question whether or not the messages should be changed and whether or not your JMH benchmarks have practical impact on real-life performance.

extended tests:
https://gist.github.com/tbw777/8ce56d2cc3a0216012362d18b7262eb2

tbw777:
<script src="https://gist.github.com/tbw777/8ce56d2cc3a0216012362d18b7262eb2.js"></script>

asbachb:
I slightly agree with @matthiasblaesing 

If you don't know regex, you might still understand whats `[A-Za-z0-9_]` with `\w` you most likely need to look up which characters are included exactly.

@tbw777 if you run the same test per pattern twice is the output of both runs most likely equal?

e.g.

```
Regexps.testMatchLong         [A-Z_0-9a-z]  100000  thrpt       45397874,174          ops/s
Regexps.testMatchLong         [A-Z_0-9a-z]  100000  thrpt       45397874,174          ops/s
```

have you disabled jit for the benchmarks?

tbw777:
Also sometimes underscore symbol may go to different locations: [a-zA-Z0-9\Q-_.\\/:\E]+
Its unreadable

neilcsmith-net:
Well, vetoed at least twice already, and seem to be at least a couple of bugs introduced, so closing.

vieiro:
Readability is, [by definition](https://www.oxfordlearnersdictionaries.com/definition/english/readability) a subjective concept. We could be discussing forever on the different aspects that make code legible. NetBeans PRs are not the proper place to discuss readability concerns.

The definition of readability ("the fact of being clear and easy to read") clearly involves the _reader_ of a text, ant not the _writer_ of the text. At least two different readers in this thread have claimed this is less legible to them.

Again (for the third? fourth? time) the recommendation to @tbw777 is to discuss  these things in the mailing list _before_ submitting a PR, at the cost of wasted  contributor _and_ reviewer time.

PR #5507, opened 2023-02-15
Group and define order for project actions.

sdedic:
Other extension may need to add actions to project node. But without (any) group, the commands order alphabetically and after all defined groups. All project build-clean actions should form one group, but allow some prior groups to be present.


MartinBalin (approved):

dbalek (approved):

mbien:
@sdedic this PR is causing enterprise/micronaut tests to fail

sdedic:
> @sdedic this PR is causing enterprise/micronaut tests to fail

I wish I could point to the real cause ... but I am almost sure this change cannot affect anything except possibly `vscode` extension build and/or LSP server tests + vscode extension tests - the only change is a declaration in `package.json` that does not participate in the NB / NBLS java runtime at all...

PR #5512, opened 2023-02-16
New rt and rf abbrevations for return true; and return false;

japplis:
This is a pull request for 2 new code templates in NetBeans:

- rt for "return true;"
- rf for "return false;"

Reasons:

- Both are used quite often. For example NetBeans source code has more than 14,000 "return true;" and more than 20,000 "return false;"
- They are consistent with the following existing code template in NetBeans: "re" for "return" and "rn" for "return null;"
- Easy to remember: just like rn for return null; 't' for true and 'f' for false
- Easy to use: rt and rf are next each other on the keyboard
- I've been using these code templates for at least 2 years and they remain ones of the main code templates I use


lkishalmi (approved):
Looks sane to me.

Thanks for your contribution!

matthiasblaesing (approved):
Looks sane

PR #5524, opened 2023-02-17
Installed some internal generics to fix some compiler warnings

tbw777:
Only private simple non generic collections was refactored to generics if no out of class usage detected.

### DIFF
--- a/ide/editor/demosrc/base/org/netbeans/editor/example/Editor.java
+++ b/ide/editor/demosrc/base/org/netbeans/editor/example/Editor.java
@@ -75,7 +75,7 @@ public class Editor extends javax.swing.JFrame {
     private boolean safeSave;
 
     private int fileCounter = -1;
-    Map com2text = new HashMap();
+    private Map<Component, JTextComponent> com2text = new HashMap<>();
     
     private Impl impl = new Impl("org.netbeans.editor.Bundle"); // NOI18N
     
@@ -258,7 +258,7 @@ public class Editor extends javax.swing.JFrame {
     private boolean saveFile( Component comp, File file, boolean checkOverwrite ) {
         if( comp == null ) return false;
         tabPane.setSelectedComponent( comp );
-        JTextComponent edit = (JTextComponent)com2text.get( comp );
+        JTextComponent edit = com2text.get( comp );
         Document doc = edit.getDocument();
         
         if( checkOverwrite && file.exists() ) {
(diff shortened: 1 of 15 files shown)


ide/editor/demosrc/base/org/netbeans/editor/example/Editor.java
```diff
@@ -544,7 +544,7 @@ private String[] getOpenedFiles() {
             }
         }
 	
-        return (String []) opened.toArray(new String[opened.size()]);
+        return opened.toArray(new String[0]);
```
BradWalker, line 547: Shouldn't this be String[opened.size()] that way things are pre-allocated?

tbw777: toArray() method have autosize operation on jvm side. You can check inside.

BradWalker: > toArray() method have autosize operation on jvm side. You can check inside.

Sure it does.. But, the writer of the software expressly requested a pre-allocated String as an argument. This changes the intent of that work. Also, I don't really like to think about the internals. I like to think about what the author wanted done, which in this case is to expressly pre-allocate a String.

I would get rid of the cast and nothing more.

matthiasblaesing: @BradWalker this might be interesting: https://www.baeldung.com/java-collection-toarray-methods (the variant, that allocates a zero sized array wins)

mbien: @BradWalker this old PR has also some info about that "false friend": 3166

BradWalker (approved):

BradWalker:
@tbw777 , you need to take a look at the test results. there was failure.

I would by looking at: TestCaseDataFactory.java

BradWalker:
I reopened this because I did some work on it and I would like to see it get integrated. I understand people might get frustrated with the fact that we can't "keep up" with the volume of PRs being received. But, that's just how the world is sometimes.

BradWalker:
Hey @mbien , could you check the tests for me.. I'm unclear why Just the Profiler checks ran and not everything else. I would like to better understand this.

BradWalker:
Mistake on my part..

tbw777:
@BradWalker only for you

PR #5527, opened 2023-02-17
Removed unecessary creating new String[]

tbw777:


vieiro:

As a consequence, I can't see enough value on this PR.

PR #5529, opened 2023-02-17
Contributor changes public APIs without previous discussion.

tbw777:
runOffEventDispatchThread() and showProgressDialogAndRun()


BradWalker:
@tbw777 , as a friendly reminder, there are 23 open PRs that belong to you. I can't keep up and just generating PRs, even if they are valid, with no thought to effort required for review, is not really helpful. It's really starting to create churn for me.

tbw777:
This is my last directly PR. Now im going to search ways how other people can make this stupid but required PRs

sdedic:
Note that there's certain value in replacing the obsolete functions. There **is** even a way how to **remove** the deprecated API methods, while retaining backwards compatibility -- the same technique as it was used to remove for example `extractComponent` public method from the `InternalHandle` SPI can be useful here.

But it's definitely not just 'remove deprecated method'.  I'd suggest to create a project as an umbrella to this and other similar issues and / or PRs, and plan in advance for a 'cleanup release': each module affected by task/issue in that project should be carefully reviewed and cleaned up, since the overhead of compatibly-moving a method off the main distribution is high.

PR #5530, opened 2023-02-18
Remove Thread#stop from httpserver and update embedded tomcat

matthiasblaesing:
NetBeans carries an embedded httpserver, that is used for various use-cases:

- view javadoc in external browser
- preview result of XSLT transformation
- record tomcat traffic monitor

This changeset:

- updates the embedded tomcat from tomcat 3.2.4 and servlet-api 2.2 to tomcat 9.0.71 and servlet-api 4.0, which are the last versions supporting the `javax` namespace
- reworks the starting mechanism not to require a separate thread (partitial fix for #4952)
- makes the embedded webserver visible and stoppable in the UI: 
image
- fixes the tests and enables running them in CI/CD tests
- ensures access control is established for all resources and servlets

It was manually validated, that the referenced use-cases still work.


matthiasblaesing:
@BradWalker @mbien I saw your thumbs up. I assume you had a look at the changeset, so thank you.

I'll let this stay for a week to give others time to have a look and if no-one objects till then, I'll merge.

ide/httpserver/src/org/netbeans/modules/httpserver/GrantAccessPanel.java
```diff
@@ -55,7 +55,7 @@ private void initComponents() {
         localTopMessage.setDisabledTextColor (javax.swing.UIManager.getColor ("Label.foreground"));  // NOI18N
         localTopMessage.setFont (javax.swing.UIManager.getFont ("Label.font")); // NOI18N
 
-        StringBuffer lTopMessage = new StringBuffer();
+        StringBuilder lTopMessage = new StringBuilder();
         lTopMessage.append(msg);
         localTopMessage.setText(lTopMessage.toString());
```
lbownik, line 60: Is lTopMessage even needed here? 
Wouldn't
 
localTopMessage.setText(msg != null ? msg : "null"); 
suffice ?

ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerSettings.java
```diff
@@ -428,54 +382,6 @@ Set<String> getGrantedAddressesSet() {
...
 
-    Properties getMappedServlets() {
-        return mappedServlets;
-    }
-
-    /** Converts string into string that is usable in URL. 
-     *  This mangling changes some characters
-     */
-    static String mangle (String name) {
-        StringBuffer sb = new StringBuffer ();
```
lbownik, line 439: since this file is already touched why not changing this to StringBuilder?

ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerSettings.java
```diff
@@ -428,54 +382,6 @@ Set<String> getGrantedAddressesSet() {
...
-            }
-        }
-        return sb.toString ();
-    }
-    
-    /** Unconverts string from URL into old string. 
-     *  This mangling decodes '%xy'
-     */
-    static String demangle (String name) {
-        StringBuffer sb = new StringBuffer ();
```
lbownik, line 460: since this file is already touched why not changing this to StringBuilder?

matthiasblaesing: Because I stared at this for multiple hours and will only invest more time, if really necessary.

enterprise/web.monitor/serversrc/org/netbeans/modules/web/monitor/server/NotifyUtil.java
```diff
@@ -117,28 +119,32 @@ void setIDE(String server) throws MalformedURLException {
...
-		log("NotifyUtil: url is " + base.concat(queryStr)); //NOI18N
-	    if(!sendRecord(base.concat(queryStr), monData))
-		otherIDEs.remove(base); 
-	}
-	return; 
+        try {
+            if(debug) log("NotifyUtil::notifyServer");  //NOI18N
+
+            if(ideServer != null) {
+                    String urlStr = ideServer.concat(URLEncoder.encode(queryStr, "UTF-8"));
```
mbien, line 126: nitpick: the UTF String literals need all the `//NOI18N`, right?

or: `StandardCharsets.UTF_8.name()`

mbien (commented):
I like it. Code looks good to me. Left some minor remarks while going through it. 

Want to test it tomorrow since I can't do that right now. I do like that NB will now also show the  port tomcat uses in the background processes list.

ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerSettings.java
mbien: @lbownik red means removed

mbien:
> preview result of XSLT transformation

@matthiasblaesing this made me curious. When I press the blue xsl transformation button on a xml and transform it, it would open the transformed file directly in a browser (not from the webserver). This is not what you mean by preview, right?

Opening downloaded javadoc of maven deps works great, cancel/restart worked too.

matthiasblaesing:
> > preview result of XSLT transformation
> 
> @matthiasblaesing this made me curious. When I press the blue xsl transformation button on a xml and transform it, it would open the transformed file directly in a browser (not from the webserver). This is not what you mean by preview, right?

No, I meant this:

image

This allows you to preview, change the source, reload, repeat.

mbien:
@matthiasblaesing wow. Yeah this works too. I somehow never realized that this was a combo box with other options than file names - everything makes sense now.

I noticed that it is currently fairly easy to make NB throw exceptions while trying to open random resources in the web browser.

example:
 - http://127.0.0.1:8082/resource
```
java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "pi" is null
	org.netbeans.modules.httpserver.WrapperServlet.getRequestURL(WrapperServlet.java:111)
	org.netbeans.modules.httpserver.WrapperServlet.processRequest(WrapperServlet.java:66)
	org.netbeans.modules.httpserver.WrapperServlet.doGet(WrapperServlet.java:52)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:497)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:584)
...
```
 - http://127.0.0.1:8082/resource/
```
 java.lang.IllegalArgumentException: URI is not absolute
	java.base/java.net.URL.fromURI(URL.java:722)
	java.base/java.net.URI.toURL(URI.java:1136)
	org.netbeans.modules.httpserver.WrapperServlet.getRequestURL(WrapperServlet.java:115)
	org.netbeans.modules.httpserver.WrapperServlet.processRequest(WrapperServlet.java:66)
	org.netbeans.modules.httpserver.WrapperServlet.doGet(WrapperServlet.java:52)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:497)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:584)
...
```

ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerModule.java
```diff
@@ -75,60 +90,77 @@ static void initHTTPServer() {
...
+                        reloader.activate();
+                        progressHandle = ProgressHandle.createHandle(
+                                "Internal Webserver", () -> {
+                                    httpserverSettings().setRunning(false);
+                                    return true;
+                                }
+                        );
+                        progressHandle.start();
+                        progressHandle.progress("Running on port " + httpserverSettings().getPort());
+                        progressHandle.switchToDeterminate(1, 1);
```
mbien, line 124: instead of the two lines, this would work too I believe:
`progressHandle.suspend("Running on port " + httpserverSettings().getPort());`

mbien:
the progress handle is causing a UX issue during shutdown. The IDE will open the warning dialog when background processes are still active while the user is trying to close the IDE, you probably didn't notice this since during dev time 
`tryme.arg.hack=-J-Dnetbeans.full.hack=true` is set in `nbbuild/build.properties` and the dialog won't show up.


exit_warning


The problem is that this dialog is opened **before** `HttpServerModule.close()` is called so there doesn't seem to be a clean way to fix this. Maybe @neilcsmith-net knows a trick.

matthiasblaesing:
Thanks for testing. Would have been to easy. I'll think about alternatives.

matthiasblaesing:
@mbien @lbownik thank you for your inputs. I reworked this and it should fix your comments. I rebased the changeset onto master and added new commits, that hold the changes. The original commits are unchanged. I plan to selectively squash before merge (see commit comments for the target commits).

For controlling the server I dropped the progress handle approach and register the server in the `Services` tab:

image

image

image

mbien (approved):
tested again and it works great as a service. Shuts down nicely too. Awesome!

ide/httpserver/src/org/netbeans/modules/httpserver/HttpServerURLMapper.java
```diff
@@ -82,7 +90,12 @@ private URL decodeURL(String path) {
                 newPath.append(tok);
             } else {
                 try {
-                    newPath.append(URLDecoder.decode(tok, "UTF-8")); // NOI18N
+                    String decodedToken = URLDecoder.decode(tok, "UTF-8"); // NOI18N
+                    if(decodedToken.startsWith("/") && decodedToken.endsWith("/")) { // NOI18N
+                        newPath.append(decodedToken.substring(1, decodedToken.length() - 1));
```
lbownik, line 95: who about
newPath.append(decodedToken, 1, decodedTocken.length() b -1) ;

it avoids sub-string allocation and a loop
turning newPath into StringBuilder  in line 86 would also improve performance a bit.

matthiasblaesing:
And another iteration.

PR #5531, opened 2023-02-18
PHP Debugger - show NULL values by default

KacerCZ:
Changes default setting of filter for NULL values in PHP debugger.

Based on personal experience when I need to turn display of NULL values for almost every debugging session.


junichi11:
@KacerCZ Could you attach screenshots(before and after)?

KacerCZ:
Dialog "Local Variable Filters", after this patch "NULL type" is checked by default and variables with NULL value are shown in Variables pane.

Before:
filter-dialog-before

After:
filter-dialog-after

junichi11 (approved):
Thanks!

KacerCZ:
Thanks for the review.

PR #5532, opened 2023-02-18
Remove second set of Thread#suspend and Thread#resume usage

matthiasblaesing:
- INSANE referenced Thread#suspend and Thread#resume but only from commented code (57bb070662e3ab3f43efd3b907bc7750daf47eb4)
- extbrowser references them only from unused test code (492479216befc2fcd7dba8c86978580578458754)
- editor.search references them only from test code as last resort. That is switch to #interrupt (db16282f7000ec21e8b123e476862292e92ac209)

Related to #4952


lkishalmi (approved):
Looks good to me.

Bit sad about seeing tests go, but I agree. Old and unmaintained.

matthiasblaesing:
@lkishalmi thank you for the review.

In general I agree, that tests are good. That though needs to more comments: They need to work and be actually run. If that is not the case, they become a liability.

I'll let this stay for a week to give others time to have a look and if no-one objects till then, I'll merge.

mbien (approved):

Issue #5072, opened 2022-12-12
[PHP] Intersection types is not highlighted nor clickable on function return type and property type


the-liquid-metal:
### Apache NetBeans version

Apache NetBeans 16

### What happened

Netbeans already able to validate and highlight class name in context of intersection types. All accepted position: function argument type, function return type, property type are fine. Unfortunately this did not happen to interface. There are:
- missing validation and highlighting on function return type and property type.
- missing validation on function argument type.

### How to reproduce

```php
<?php
class Class1 {}
class Class2 {}
interface Iface1 {}
interface Iface2 {}

// displayed highlight: all accept null
// missing highlight: [none]
// displayed error: null and 2nd Class1
// missing error: 2nd Iface1
function funcX(Class1 & Class2 & Iface1 & Iface2 & null & Class1 & Iface1 $arg1) {
    echo $arg1;
}

// displayed highlight: Class1 and Class2
// missing highlight: Iface1 and Iface2
// displayed error: null and 2nd Class1
// missing error: 2nd Iface1
function funcY($arg1): Class1 & Class2 & Iface1 & Iface2 & null & Class1 & Iface1 {
    echo $arg1;
}

class Class3 {
    // displayed highlight: Class1 and Class2
    // missing highlight: Iface1 and Iface2
    // displayed error: null and 2nd Class1
    // missing error: 2nd Iface1
    public Class1 & Class2 & Iface1 & Iface2 & null & Class1 & Iface1 $prop1;
}
```

### Did this work correctly in an earlier version?

No / Don't know

### Operating System

Windows 10

### JDK

Java: 14.0.1; Java HotSpot(TM) 64-Bit Server VM 14.0.1+7 Runtime: Java(TM) SE Runtime Environment 14.0.1+7

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

KacerCZ:
Reproducible.
Thanks fro your report.

KacerCZ:
During testing I also found out, that "Find Usages" does not find all usages of `Iface1` and `Iface2`.

KacerCZ:
PR #5534 fixes reported errors.

PR #5534, opened 2023-02-19
Fixed detection of duplicate types

KacerCZ:
Goes through all types and does not return on first error.
Partial fix for #5072

### DIFF
--- a/php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
+++ b/php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
@@ -430,18 +430,15 @@ public class UnusableTypesHintError extends HintErrorRule {
                 }
                 if (checkedTypes.contains(name)) {
                     createDuplicateTypeError(type, typeName);
-                    return;
                 } else if (checkedTypes.contains(Type.BOOL)) {
                     // bool|false bool|true
                     if (Type.FALSE.equals(name) || Type.TRUE.equals(name)) {
                         createDuplicateTypeError(type, typeName);
-                        return;
                     }
                 } else if (checkedTypes.contains(Type.FALSE) || checkedTypes.contains(Type.TRUE)) {
                     // false|bool true|bool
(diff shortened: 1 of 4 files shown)


php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
```diff
@@ -430,18 +430,15 @@ private void checkDuplicateType(List<Expression> types) {
                 }
                 if (checkedTypes.contains(name)) {
                     createDuplicateTypeError(type, typeName);
-                    return;
```
junichi11, line 433: I've added it for a little performance (because the next error is shown if the error is fixed) but I don't think this fix has a big problem. So, it's OK.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
```diff
@@ -430,18 +430,15 @@ private void checkDuplicateType(List<Expression> types) {
...
                 } else if (checkedTypes.contains(Type.BOOL)) {
                     // bool|false bool|true
                     if (Type.FALSE.equals(name) || Type.TRUE.equals(name)) {
                         createDuplicateTypeError(type, typeName);
-                        return;
```
junichi11, line 438: Unit tests?

KacerCZ: I've added more test cases covering also combinations of boolean type error with duplicate class error.

php/php.editor/src/org/netbeans/modules/php/editor/verification/UnusableTypesHintError.java
KacerCZ: Right, original code with early exit finishes faster.
If there is no error then all types will be checked anyway.

junichi11 (approved):
Thanks.

KacerCZ:
Thanks for the review.

PR #5535, opened 2023-02-19
Fixed key escaping; fixed list/map property extraction.

sdedic:
This PR initially started from [reviewing this change](https://github.com/apache/netbeans/pull/5400/files/0054e511f349a3d4a075d0688ec7d040021a7b7e#diff-c214cb774bb9be2ee5aaaf45a8bbc5796aa98cdf86a045378bc9fd540c7ccc8fR173) as the intention was to un-escape '\;' sequence to ';'. Then as I attempted to write tests, I've found a few rough edges of the BuildProperties API ...

Some highlights here, minor commnents inline in the diff.
1/ Gradle defines an extensibility 'ext' container for most domain object, where extra properties can be defined. After definition, they become 'part of' the domain object itself and can be referenced as such. Those extra properties were not exported in the project info. Now they're exported into the domain object's namespace.

2/ Structures (structured things that are not Maps) get their keys enumerated. Allows for introspection in NB IDE

### DIFF
--- a/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
+++ b/extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
@@ -91,6 +91,7 @@ import org.gradle.api.logging.Logging;
 import org.gradle.api.plugins.ExtensionAware;
 import org.gradle.api.plugins.ExtensionContainer;
 import org.gradle.api.plugins.ExtensionsSchema.ExtensionSchema;
+import org.gradle.api.plugins.ExtraPropertiesExtension;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.reflect.HasPublicType;
 import org.gradle.api.reflect.TypeOf;
@@ -100,6 +101,7 @@ import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.TaskDependency;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.api.tasks.testing.Test;
+import org.gradle.internal.extensibility.DefaultExtraPropertiesExtension;
 import org.gradle.jvm.JvmLibrary;
 import org.gradle.language.base.artifact.SourcesArtifact;
 import org.gradle.language.java.artifact.JavadocArtifact;
@@ -121,6 +123,11 @@ class NbProjectInfoBuilder {
      * project loader is enabled to FINER level.
      */
     private static final Logger LOG =  Logging.getLogger(NbProjectInfoBuilder.class);
+
+    /**
+     * Name of the default extensibility point for various domain objects defined by Gradle.
+     */
+    private static final String DEFAULT_EXTENSION_NAME = "ext"; // NOI18N
     
     private static final String NB_PREFIX = "netbeans.";
     private static final Set<String> CONFIG_EXCLUDES = new HashSet<>(asList( new String[]{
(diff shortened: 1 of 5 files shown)


extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -538,32 +545,45 @@ private boolean isMutableType(Object potentialValue) {
...
                 return;
             }
         }
-        if (clazz == null || (excludes == null && ignoreClassesPattern.matcher(clazz.getName()).matches())) {
+        
+        if (clazz == null || clazz.isEnum()) {
             return;
         }
-        if (clazz.isEnum() || clazz.isArray()) {
+        if (clazz.isArray() || (excludes == null && ignoreClassesPattern.matcher(clazz.getName()).matches())) {
```
sdedic, line 577: The condition for shortcut path remain the same; but now the implementation tries to export Maps and Lists. Covers the case when `inspectObjectAndValues0` is called for e.g. java.util.HashMap object, which is otherwise excluded by filters - i.e. Map inside a Map.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -685,87 +708,10 @@ private void inspectObjectAndValues0(Class clazz, Object object, String prefix,
...
                     Map<String, ?> m = new HashMap<>(nc.getAsMap());
-                    List<String> ss = new ArrayList<>(m.keySet());
-                    propertyTypes.put(prefix + propName + COLLECTION_KEYS_MARKER, String.join(";;", ss));
-                    for (String k : m.keySet()) {
-                        newPrefix = prefix + propName + "." + k + "."; // NOI18N
-                        Object v = m.get(k);
-                        defaultValues.put(prefix + propName + "." + k, Objects.toString(v)); // NOI18N
-                        inspectObjectAndValues(v.getClass(), v, newPrefix, globalTypes, propertyTypes, defaultValues, null, false);
-                    }
+                    dumpContainerProperties(m, prefix + propName, defaultValues);
```
sdedic, line 711: Extracted into a method, to allow call from `inspectExtensions'.

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -685,87 +708,10 @@ private void inspectObjectAndValues0(Class clazz, Object object, String prefix,
...
-                            String k = o.toString();
-                            String newPrefix = prefix + propName + "[" + k + "]"; // NOI18N
-                            Object v = mvalue.get(o);
-                            defaultValues.put(newPrefix, Objects.toString(v)); // NOI18N
-                            inspectObjectAndValues(v.getClass(), v, newPrefix + ".", globalTypes, propertyTypes, defaultValues, null, itemClass == null);
-                        }
-                        dumped = true;
-                    }
+                } else {
+                    dumped = dumpValue(value, prefix + propName, propertyTypes, defaultValues, t, typeParameters);
```
sdedic, line 714: extracted into a method so that inspectObjectAndValues can call that for j.u.HashMap from line 579

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -811,14 +885,23 @@ private void inspectExtensions(String prefix, ExtensionContainer container) {
             
             Object ext;
             try {
-                ext = project.getExtensions().getByName(extName);
+                ext = container.getByName(extName);
```
sdedic, line 888: this was a bug !! The extension enumerated from a domain object was always looked up as extension of a project itself.

sdedic:
Rebased onto latest master; resolved conflicts with @tbw777 changes to `NbProjectInfoBuilder`

extide/gradle/netbeans-gradle-tooling/src/main/java/org/netbeans/modules/gradle/tooling/NbProjectInfoBuilder.java
```diff
@@ -534,32 +541,45 @@ private boolean isMutableType(Object potentialValue) {
         return adapter.isMutableType(potentialValue);
     }
     
+    private void addNestedKeys(String prefix, Map<String, String> propertyTypes, Collection<String> keys) {
+            propertyTypes.merge(prefix + COLLECTION_KEYS_MARKER, String.join(";;", keys), (old, add) -> {
+                List<String> oldList = new ArrayList<>(Arrays.asList(old.split(";;")));
+                List<String> newList = Arrays.asList(add.split(";;"));
+                oldList.addAll(newList);
```
lbownik, line 548: how about ?
Collections.addAll(oldList, add.split(";;"));

extide/gradle/test/unit/src/org/netbeans/modules/gradle/loaders/ExtensionPropertiesExtractorTest.java
```diff
@@ -164,6 +171,25 @@ public void testMapExtensionProperty() throws Exception {
...
+        BuildPropertiesSupport support = BuildPropertiesSupport.get(p);
+        assertNotNull(support);
+        
+        BuildPropertiesSupport.Property prop;
+        
+
+        prop = support.findExtensionProperty("graalvmNative", "binaries.main.buildArgs");
+        assertEquals(BuildPropertiesSupport.PropertyKind.LIST, prop.getKind());
+        Iterable<BuildPropertiesSupport.Property> it = support.items(prop, null);
+        assertNotNull(it);
```
lbownik, line 186: this seems redundant, next line will throw NPE if it == null and the test will fail

dbalek (approved):
Looks fine.

lkishalmi (approved):
Looks Ok.

extide/gradle/src/org/netbeans/modules/gradle/loaders/ExtensionPropertiesExtractor.java
```diff
@@ -170,7 +171,7 @@ public Collection<String> keys(BuildPropertiesSupport.Property container) {
             if (c == null) {
                 return null;
             }
-            return Arrays.asList(c.split(";;")).stream().map(s -> s.replaceAll("\\;", ";")).collect(Collectors.toList());
+            return Arrays.asList(c.split(";;")).stream().map(s -> s.replaceAll("\\\\;", ";")).collect(Collectors.toList());
```
mbien, line 174: `s.replace(..)` wouldn't need the double escape I believe - its also cheaper since it doesn't do regex.

sdedic: Apologies, completely forgot about this - fixed in 8f0cfa2c1683

PR #5536, opened 2023-02-19
Let the user choose main class, when none is open

jlahoda:
In the VS Code, when attempting to start a program using the Run pane, without any file opened, the Run fails, as no main class is found.

This patch proposes to show a dialog so that the user can choose a main class. In case only one main class is found, it is selected automatically.

The intent behind the `getServerLookup()` -> `getServerServicesLookup()` change is to ensure the content of the default lookup is present in `context.getLspSession().getLookup()` only once. It is added there once in the session lookup, and is also part of the server lookup. So, the patch is instead using a subset of the server lookup that does not include the default lookup.

---

### DIFF
--- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
+++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/debugging/launch/NbLaunchRequestHandler.java
@@ -23,13 +23,19 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


dbalek (approved):
Looks fine.

sdedic (approved):

mbien:
@dbalek @sdedic I added LSP label and going to restart tests so that LSP tests are run. Labels are also important for getting this PR into the release notes. Please check those things before approving to make it less likely that the master is red (which just happened a week ago, so this isn't theoretical - although I have no doubt that jan tested this PR properly).

see [wiki](https://cwiki.apache.org/confluence/display/NETBEANS/PRs+and+You+-+A+reviewer+Guide) for more details

matthiasblaesing:
@jlahoda this is reviewed - any reason not to merge?

jlahoda:
I realized yesterday I'd like one more tweak here (more appropriate place to set the lookup), I'm working on that.

mbien:
would recommend to squash before merge.

btw @jlahoda would you be ok with #5617, since we will probably turn off the github squash feature, which would require to squash locally as last step before pressing merge.

PR #5543, opened 2023-02-22
When adding an Autonomous DB, present profiles instead of tenancies

sdedic:
The user was presented with a selector for tenancies when asked to locate an autonomous DB from the OCI. But tenancy is not that descriptive and it's better when the user gets a selector for profiles, which usually have names meaningful to the user. Tenancy name and its region are still interesting as detail info.
This PR changes the chooser to present profiles, with tenancy+region as detail of the presented choice.

 In addition, two bugs were fixed:
- a progress detail message may be coalesced into the start event, but [was not sent to the client](https://github.com/apache/netbeans/pull/5543/files#diff-1027bccfaaf5f4f1d222a283acbbd61d93bf25b5512b88a24ca1a423781af527R156)
- the creation of next `NotifyDescriptor` step in `NotifyDescriptor.ComposedInput` may block when gathering information. We need to leave the main LSP thread unblocked

### DIFF
--- a/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/AddADBAction.java
+++ b/enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/AddADBAction.java
@@ -18,18 +18,21 @@
  */
 package org.netbeans.modules.cloud.oracle.actions;
 
+import com.oracle.bmc.identity.model.Tenancy;
 import com.oracle.bmc.model.BmcException;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import org.netbeans.api.progress.ProgressHandle;
 import org.netbeans.modules.cloud.oracle.OCIManager;
 import org.netbeans.modules.cloud.oracle.OCIProfile;
 import org.netbeans.modules.cloud.oracle.actions.DownloadWalletDialog.WalletInfo;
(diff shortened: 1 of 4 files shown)


enterprise/cloud.oracle/src/org/netbeans/modules/cloud/oracle/actions/AddADBAction.java
```diff
@@ -83,47 +89,82 @@ public class AddADBAction implements ActionListener {
...
-                        p.getTenancy().ifPresent(tenancies::add);
+                    ProgressHandle h = ProgressHandle.createHandle(Bundle.MSG_CollectingProfiles());
+                    h.start();
+                    h.progress(Bundle.MSG_CollectingProfiles_Text());
+        
+                    Map<OCIProfile, Tenancy> profiles = new LinkedHashMap<>();
+                    Map<String, TenancyItem> tenancyItems = new LinkedHashMap<>();
+                    try {
+                        for (OCIProfile p : OCIManager.getDefault().getConnectedProfiles()) {
+                            TenancyItem t = p.getTenancy().orElse(null);
```
lbownik, line 116: wouldn't Optional.ifPresent be more apropriate?

sdedic: Acutally, I don't think so. OCIProfile IMHO misuses `Optional`, where a simple `null` indicating that value is not available could be sufficient. I don't find any value in replacing `if` with `Optional` in a simple 
imperative code - except that it looks fancy, and is harder to debug, and consumes more resources at runtime. 
The `orElse` here is a way how to convert the unnecessary Optional into value that can be used. The other options would be an temporary var and `ifPresent` test (possible) and change of `getTenancy()` (which would require me to analyse **actuall** usages whether the Optional is generally useful or not for clients).

sdedic: @neilcsmith-net  - We should argue about that on the mailing list :) good arguments make better coding style.

jhorvath (approved):
Looks good

PR #5548, opened 2023-02-22
Maven EventSpy not compatible < JDK 1.8

graben:


mbien:
hi @graben, thanks for the update.

But before we can merge this you would have to update the commit to a valid email address, we would like to avoid having noreply mails in the git history.

to fix the commit you can run:
`git commit --amend --author="Foo Bar <mail@foobar.com>" --no-edit`

after that simply force push again into this PR.

here is also a guide how to change this globally in your git settings:
https://cwiki.apache.org/confluence/display/NETBEANS/Submitting+Pull+Requests

graben:
@mbien : Email address should be set, now.

mbien:
@graben I just noticed, do you want to update the comment at L400 so that it fits to the new code?
https://github.com/apache/netbeans/pull/5548/files#diff-b60bb1906d8d7eb62f0b4fa7925e1c8e360c8bba3c4af8beaf5c05fe33b32bcdR400

github doesn't let me mark it since it is not part of the changeset

graben:
@mbien : Sure, missed that comment the first time

mbien (approved):
Excellent. Thanks again. Going to merge once tests ran.

Issue #5546, opened 2023-02-22
Try with resources with "var" formatted with extra space before "var"


rturner-edjuster:
### Apache NetBeans version

Apache NetBeans 16

### What happened

This is almost the same issue as https://github.com/apache/netbeans/issues/3720.

When Java code is automatically formatting a try-with-resources statement, if the resource is specified with the type `var` instead of a specific type, an additional space is injected before the `var` keyword.

### How to reproduce

With a clean install of NetBeans 16 on MacOS 13.1, with the default Java code formatting settings (no changes to defaults after install, and no importing of previous version settings), and using a "new file", and no existing project:

1. Create a new Java file:
2. Replace all code with the following:
```
import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Paths;

public class NewClass {

    public void test() throws IOException {
        try (var br = Files.newBufferedReader(Paths.get("test"))) {

        }
    }
}
```
3. Select "Format" from the "Source" menu and the code will be reformatted as:
```
import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Paths;

public class NewClass {

    public void test() throws IOException {
        try ( var br = Files.newBufferedReader(Paths.get("test"))) {

        }
    }
}
```

Note the additional space in `try ( var`.

If you replace `var` with `BufferedReader`, the problem does not occur.

### Did this work correctly in an earlier version?

Apache NetBeans 12.6 or earlier

### Operating System

Mac OS X version 13.1 running on aarch64; UTF-8; en_CA (nb)

### JDK

18.0.2; OpenJDK 64-Bit Server VM 18.0.2+9-FR

### Apache NetBeans packaging

Apache NetBeans provided installer

### Are you willing to submit a pull request?

No

### Code of Conduct

Yes

rturner-edjuster:
@neilcsmith-net I will also try with version 17 when the release is made available for general download -- seems imminent based on voting thread. I will provide an update here once I've done that.

neilcsmith-net:
@rturner-edjuster we're just waiting for the webpages to be updated now.  The downloads are already published on ASF infrastructure at https://downloads.apache.org/netbeans/netbeans-installers/17/  And the community installers are also up at https://www.codelerity.com/netbeans/

rturner-edjuster:
Looks like this issue is resolved in v17. My test case above, along with testing it on other code seems to work as expected. Closing.

neilcsmith-net:
[...]

PR #5374, opened 2023-01-27
Annotated record reformatting fix

mbien:
fixes #4399 targets delivery

had a hard time figuring out how the `Reformatter` works since not much is documented there, But the fix turned out to be only a small change. Comparing `scanRecord` with other methods helped a bit.

PR #5550, opened 2023-02-22
Adding support for JDK 20 javac.

jlahoda:
This adds the separately released nb-javac based on JDK 20, and asdjusts NetBeans to it. One of the biggest changes is the drop of source level 7.

---


mbien (commented):
awesome! 

Since this is a major version upgrade, we have to add a few more tweaks:
 - version check to 20 analog to #3953 line: https://github.com/apache/netbeans/pull/3953/files#diff-4acb756ec7378fd25f5bc06c331506df4d42d67d21b4356f8c90ff35dfaa1f75R37
 - module description (and spec version?) update like in [#4010](https://github.com/apache/netbeans/pull/4010/files)

i *believe* this would be everything - but would be good that a second set of eyes verifies this

added https://github.com/apache/netbeans/labels/ci%3Adev-build label in case someone would like to run manual tests. 

This nb-javac build also removes the javac backports (https://github.com/JaroslavTulach/nb-javac/pull/16/files). Since arm block formatting tests didn't fail it appears everything is still working :)

neilcsmith-net:
> This nb-javac build also removes the javac backports (https://github.com/JaroslavTulach/nb-javac/pull/16/files). Since arm block formatting tests didn't fail it appears everything is still working :)

Well, we don't have an arm block formatting test that's sensitive to the fix AFAIK - possibly one with `var` judging by #5546 ?

mbien:
@neilcsmith-net thats the main reason I added the dev build label. Wanted to look up the old issue and test that manually later. I only took a quick look so far - this was no review.

neilcsmith-net:
Thanks @jlahoda Yes, my test wasn't sensitive to this because of the workaround in place.  `var` might have been judging from reports.  I also wonder if the other reported issue in https://github.com/apache/netbeans/issues/3720#issuecomment-1248804306 is still there - it was also triggered by the overlap in reported positions, but I couldn't find a useful workaround for that.

It's such a small issues, but it didn't half bring in some complaints! :smile:

Also looks like there's a conflict since I merged the updated sigfiles from NB17.

java/java.hints/test/unit/src/org/netbeans/modules/java/hints/errors/AddFinalModifierTest.java
```diff
@@ -105,4 +107,12 @@ protected String toDebugString(CompilationInfo info, Fix f) {
...
 	return AddOrRemoveFinalModifier.createAddFinalModifier().getCodes();
     }
+    public static Test suite() {
+        //javac no longer supports any source level that would produce this error
+        return new TestCase("noop") {
+            @Override
+            public void runBare() throws Throwable {
+            }
+        };
+    }
```
mbien, line 117: i suppose that is because variables are now implicitly final if they are only read so that they can be used in lambdas etc. Is it still useful to keep this test around?

mbien (approved):
looks good to me, left one comment inline.

I manually tested the arm blocks a bit and formatting works as expected (no extra space -> no regression). Good to see that nb-javac is now again backport-free.

https://github.com/apache/netbeans/issues/3720#issuecomment-1248804306 is still reproducible but this might have different causes. But this reminded me that I have a wip branch locally which tries to fix formatter issues with annotated records.

jtulach (approved):
I am glad to see that the `nb-javac` [jdk-20+34](https://github.com/jaroslavtulach/nb-javac/tree/jdk-20+34) is working well for NetBeans purposes. Everything seems to be green - merge it!

Once the `jdk-20+34` `nb-javac` is in, I'd like to proceed with a _"dog food experiement"_ - I'd like the try to build NetBeans with it! How else can we demonstrate the `nb-javac` is good enough than by building such a huge project like NetBeans?

mbien:
> But this reminded me that I have a wip branch locally which tries to fix formatter issues with annotated records.

oops this is already merged and released #5374. I forgot to remove the local branch and thought I still have to create the PR for this :)

mbien:
there were a few changes in javac this update:
https://github.com/openjdk/jdk20u/compare/jdk-20+34...jdk-20.0.1-ga

https://foojay.io/java-20/?version=20.0.1&quarter=042023&tab=component (set filter to tools)

I see @jlahoda was involved in most of them, would you recommend to try to get a fresh nb-javac build in? There would be still some time since we are at RC1 atm.

There is always the option to run without nb-javac if someone is hit by an edge case, so if those fixes are not important we could stick to 20+34.

PR #5560, opened 2023-02-24
Vscode ext should update default connection

sdedic:
The vscode "Database" view queried NBLS for the default DB just on startup, then it retained the connection node's ID, and only allowed vscode > NBLS change propagation. We need - for example when defining a new DB - that the NBLS's default connection is reflected back in the vscode. That will help if no connection is present and a new (first) one is defined that becomes the default. May also help in cases when other command(s) redefine the default connection internally.

I've changed DB module since there was a bug in that it offered the 1st connection as a default if none was specifically set - but was willing to change the default 'on the fly' if a new connection appeared, so the default could shift unexpectedly. Case when there's a preferred connection name stored, but the definition is somehow missing was also mishandled.

### DIFF
--- a/ide/db/src/org/netbeans/modules/db/explorer/ConnectionList.java
+++ b/ide/db/src/org/netbeans/modules/db/explorer/ConnectionList.java
@@ -24,9 +24,9 @@ import java.beans.PropertyChangeListener;
 import java.io.IOException;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 import java.util.WeakHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.prefs.PreferenceChangeEvent;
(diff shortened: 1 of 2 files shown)


ide/db/src/org/netbeans/modules/db/explorer/ConnectionList.java
```diff
@@ -273,32 +278,57 @@ public void setPreferredConnection(DatabaseConnection c) {
...
+            }
+        }
+        if (prefChanged) {
+            fireListeners();
+        }
+        return p;
+    }
+    
+    private static DatabaseConnection findConnection(String name, DatabaseConnection[] conns) {
         for (int i = 0; i < conns.length; i++) {
```
lbownik, line 327: for-each loop maybe?

jhorvath: I don't see the problem, it's a matter of personal preference.

jhorvath (approved):
Looks good

ide/db/src/org/netbeans/modules/db/explorer/ConnectionList.java
sdedic: I usually use for-loops in these situations ... this part was actually extracted from the older code that is obviously no excuse for not updating the style. I usually incorporate 'style issues' whenever some func/perf ("real") issue is found and a new commit has to be made.

PR #5565, opened 2023-02-26
Regenerate PHP signature files

junichi11:
PHP Documentation: Date: 21 Feb 2023

I've verified that there is no exception when I open a PHP file.


tmysik (approved):

tmysik:
@junichi11 Junichi, just to be sure - we can merge this PR, right? If so, feel free to do so. Thanks.

junichi11:
@tmysik Yes, we can merge it :) Thanks!

